PR #21, opened 2022-07-22
add nip22: unacceptable event created_at field

jeffthibault:
Marked as draft because looking for input on the time limit. Should it be 1 minute, 10 minutes, 1 hour, etc...?

### DIFF
--- a/22.md
+++ b/22.md
@@ -6,18 +6,18 @@ Unacceptable Event `created_at` time
 
 `draft` `optional` `author:jeffthibault`
 
-Relays may support notifying clients that the event they published has an unacceptable `created_at` time. A relay will will consider the `created_at` time unacceptable if the `created_at` time is more than **[limit]** before the event was received by the relay. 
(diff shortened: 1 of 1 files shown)


fiatjaf:
Can we let the relay decide that time threshold maybe? Does that defeat the purpose of this NIP?
What about events with time in the future?

jeffthibault:
Good point on time in the future. I think the relay should reject any event with a `created_at` that is later than when it was published. I can't think of a valid reason to do that.

I suppose the relays could determine the limit themselves but I was hoping to establish a standard value for this NIP. The relays can do whatever they want but I was thinking that if a relay says it supports NIP-22, then the client knows what the limit is.

fiatjaf:
Yes, I agree that would be the purpose of this NIP.

One valid reason for events in the future:
  - scheduled posts.
Three valid reasons for events in the past:
  - migrating your own posts from one relay to another;
  - spreading the word from someone else who isn't publishing in this specific relay but may have followers that are listening to it;
  - "retweeting" posts that aren't present in one relay by sending first the event you're retweeting then another event, from yourself, referencing the other.

jeffthibault:
accidently closed this. my aplogies. new PR is #22

PR #22, opened 2022-07-22
NIP-22: event created_at limits

jeffthibault:
For relays to set limits on acceptable created_at times for events

Formatted: https://github.com/jeffthibault/nips/blob/nip22-unacceptable-event-time/22.md

Previous: #21

### DIFF
--- a/22.md
+++ b/22.md
@@ -1,31 +1,38 @@
 NIP-22
 ======
 
-Unacceptable Event `created_at` time
+Event `created_at` Limits
 ---------------------------
 
 `draft` `optional` `author:jeffthibault`
 
-Relays may support notifying clients that the event they published has an unacceptable `created_at` time. A relay will consider the `created_at` time unacceptable if the `created_at` time is more than **[limit]** before the event was received by the relay (in the past) OR if the `created_at` time is later than the time the event was received by the relay (in the future). 
+Relays may define both upper and lower limits for which they will consider an event's `created_at` time to be acceptable if it is within those limits. Both the upper and lower limits MUST be unix timestamps to match the format of the event's `created_at` field. The upper limit restricts how far into the future an event's `created_at` time can be set to and the lower limit restricts how far into the past an event's `created_at` time can be set to.  
 
-If a relay supports this NIP, the relay SHOULD send the client a `NOTICE` message saying the event was not stored because the `created_at` time was unacceptable.
+If a relay supports this NIP, the relay SHOULD send the client a `NOTICE` message saying the event was not stored when the `created_at` time is not within the upper and lower limits.
 
 Client Behavior
 ---------------
 
-Clients SHOULD use the `supported_nips` field to learn if a relay supports event `created_at` time checks.
+Clients SHOULD use the `supported_nips` field to learn if a relay uses event `created_at` time limits as defined by this NIP.
 
 Motivation
 ----------
 
-The motivation for this NIP is to prevent clients from saying they published an event *significantly* earlier than they actually did or saying they published an event in the future.
+The motivation for this NIP is to formalize a way for relays to restrict event timestamps to times they deem to be acceptable and allow clients to be aware of relays that have these restrictions.
 
-The event `created_at` field is just a unix timestamp (integer) so one could set it to a time in the past or future. For example, the `created_at` field could be set to a time 10 years ago even though it was created today and it could still be a valid event. One could also set the `created_at` field to a time 10 years in the future and it could still be a valid event. This NIP aims to set a maximum amount of time elapsed between when an event was created and when it was *actually* published and prevent events from being from the future.
+The event `created_at` field is just a unix timestamp and can be set to a time in the past or future. For example, the `created_at` field can be set to a time 20 years ago even though it was created today and still be a valid event. It can also be set to a time 20 years in the future and still be a valid event. This NIP aims to define a way for relays that do not want to store events with *any* timestamp to set their own restrictions.
 
-Relay Logic
------------
+A wide adoption of this could create a better UX on clients as well because it would decrease the liklihood of the user seeing events from dates such as 1984 or 2084, which could be confusing.
 
-```
-if time.now - event.created_at > limit OR event.created_at > time.now:
-  send NOTICE
+Python Example
+--------------
+
+```python
+import time
+
+UPPER_LIMIT = int(time.now) + 900 # Define upper limit as 15 minutes into the future
+LOWER_LIMIT = int(time.now) - 86400 # Define lower limit as 1 day into the past
+
+if event.created_at < LOWER_LIMIT or event.created_at > UPPER_LIMIT:
+  ws.send('["NOTICE", "The event created_at field is out of the acceptable range for this relay and was not stored."]')
 ```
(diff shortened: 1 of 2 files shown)


jeffthibault:
@fiatjaf in ref to https://github.com/nostr-protocol/nips/pull/21#issuecomment-1192758758

How would scheduled posts work? Let's say the `created_at` time for an event is in the future. A client could still fetch it before it was ready to be published, right? Example: pubkey A creates a scheduled post event with `created_at` 10 minutes in the future. 2 minutes later, pubkey B requests all events from pubkey A. pubkey B will see the scheduled event in the present but the `created_at` time is still 8 minutes in the future.

jb55:
> The motivation for this NIP is to prevent clients from saying they published an event *significantly* earlier than they actually did.

I like this feature, as it allows you to replay all your old posts to a new account. I don't see why it would be a problem for relays?

fiatjaf:
> How would scheduled posts work?

Some relays can have a policy to not return future posts ever. If you want a scheduled post you send it only to these relays.

jeffthibault:
> I like this feature, as it allows you to replay all your old posts to a new account.

True, it is good to be able to do this. And @fiatjaf gave other valid uses in #21.

So I wrote this NIP because I realized someone could exploit this and create revisionist history by making "old" posts in the present and try to benefit from it. Maybe it's not a big deal and I'm overthinking it.

jb55:
timestamps by themselves are not reliable on nostr. if you really cared if the timestamp is accurate you should use a client that implements [NIP03][nip03].

https://damus.io will likely implement this sometime in the future...

[nip03]: https://github.com/nostr-protocol/nips/blob/master/03.md

Giszmo:
Users migrating from a private relay to such a relay would be unable to push their old events.
If you want proof of time, use opentimestamps as jb55 said.
Clients should treat bogus timestamps differently. Relays not.

Scheduled events are actually a cool idea but I'm afraid what would happen if we are not more explicit about these events being scheduled is that clients that have set their clocks wrong will complain about peers not receiving their events quickly.

Cameri (approved):
Love this, I was implementing this on nostr-ts-relay. Glad to see there's a NIP!

RandyMcMillan:
### Weeble Wobble: Modulus Time derived from Bitcoin block height and UTC_secs time. 

I was using a time stamp mechanism I called "Weeble Wobble" it is basically:

<UTC_secs>%<bitcoin_block_height>

UTC_secs mod BLOCK_HEIGHT

<Weeble> = floor(UTC_secs/BLOCK_HEIGHT)

<Wobble> = Remainder 

<Weeble>:<Wobble>

The target pub time would be <Weeble>:

<Wobble> is useful as a fizzy network time between blocks.

Since <Weeble> is derived from floor(<UTC_secs>%<bitcoin_block_height>) it is rooted in a known truth.

*NOTE <UTC_millis> is applicable if more granularity is needed.

jeffthibault:
@Cameri Nice! What are you using as time limit for "past" events? I'd like to establish a standard limit before I move this out of Draft status.

Cameri:
> @Cameri Nice! What are you using as the time limit for "past" events? I'd like to establish a standard limit before I move this out of Draft status.

I was going to leave it to the relay operator to set these in a settings file because it depends on the type of relay. The limits should be in seconds since `created_at` is in seconds, otherwise we need to do math. 

Archival relays may set this to the day Nostr was born.

Public relays may allow events to be backfilled up to a day. This allows clients to be offline for up to a day and then briefly connect to send all events.

Some relays may choose disallow backfilling so they would set it to allow for clock skew (how long does it take an event to travel around the world when there's network congestion?).

There's also a talk about scheduled events so an upper limit may need to be configurable.

In my relay implementation I'm calling out both positive and negative difference maximum limits separately.

I'd suggest the following defaults: a maximum negative difference of a day in seconds. And a maximum positive difference of 15 minutes (until there's support for scheduled events).

jeffthibault:
@Cameri I re-wrote the NIP to be more generic after thinking about your comment. Let me know if you still approve.

@fiatjaf @jb55 Do you guys have any thoughts on the updated NIP?

22.md
```diff
@@ -0,0 +1,38 @@
...
+--------------
+
+```python
+import time
+
+UPPER_LIMIT = int(time.now) + 900 # Define upper limit as 15 minutes into the future
+LOWER_LIMIT = int(time.now) - 86400 # Define lower limit as 1 day into the past
+
+if event.created_at < LOWER_LIMIT or event.created_at > UPPER_LIMIT:
+  ws.send('["NOTICE", "The event created_at field is out of the acceptable range for this relay and was not stored."]')
```
Cameri, line 37: Should the limits be mentioned here? Or should there be a different message depending on the limit exceeded?

Giszmo: A notice is human readable stuff, so while putting the limits into the example would be nice, it would not be part of the nip's specification.

Giszmo:
It's worth mentioning that `created_at` is very relevant for relays, too. I screwed up and used milliseconds in my created_at timestamps, so that a test user now has his kind-0 meta data stored in the year 54460. November 14th to be precise. Standard compliant relays would reject any sane replacement.

Other than that - other than the prevention of such accidents - I see no big benefit in this nip, especially for the back-dating part.

Giszmo:
So I reworded the nip a bit. For your consideration: https://github.com/jeffthibault/nips/pull/1

22.md
Cameri: @Giszmo  So this notice message is just a recommendation or is the exact message that should be sent?
I would advice against saying what the relay does with the event because different kinds are handled differently, that means we should not say `and was not stored` in the notice.

jeffthibault: @Cameri My intention with the example was just to show one way a relay could notify the client about this. Of course, any relay implementor or relay operator could change this notice message to be worded any way they want. I can add a comment in the example to clarify that.

Cameri: Ready to approve this when you are ready for reviews

Cameri (approved):

Giszmo (approved):

jb55:
On Sat, Aug 13, 2022 at 10:59:29AM -0700, Jeff Thibault wrote:
***@***.*** @jb55 Do you guys have any thoughts on the updated NIP?

eh I'm indifferent because I wouldn't want to encourage such a feature
and probably wouldn't connect to relays that supported this. Replaying
old notes is one of the only way to migrate accounts atm if you screw up
and leak your key.

fiatjaf:
My thoughts are that this shouldn't be encouraged as being somehow "part of the protocol", but I see no harm in some relays deciding to enforce these rules (although I hope they doesn't become very widespread).

fiatjaf:
Is this implemented anywhere? Can I merge?

jb55:
> My thoughts are that this shouldn't be encouraged as being somehow "part of the protocol", but I see no harm in some relays deciding to enforce these rules (although I hope they doesn't become very widespread).

yeah same, I'm of course fine standardizing this if relays want to implement it. LGTM!

jeffthibault:
> Is this implemented anywhere already? Can I merge?

@Cameri mentioned he was implementing it in nostr-ts-relay.

It's ready to be merged.

Cameri:
> On Sat, Aug 13, 2022 at 10:59:29AM -0700, Jeff Thibault wrote:
> ***@***.*** @jb55 Do you guys have any thoughts on the updated NIP?
> 
> eh I'm indifferent because I wouldn't want to encourage such a feature
> and probably wouldn't connect to relays that supported this. Replaying
> old notes is one of the only way to migrate accounts atm if you screw up
> and leak your key.
> 

I'm implementing the ability to set these limits. Relay operators can choose to not use it.

Giszmo:
NostrPostr probably will limit forward dating of replaceable events. I see little use in limiting back-dating.

PR #23, opened 2022-07-26
nip16: small fix

Semisol:
Yes, I did modify NIP-12 because I couldn't come up a template by myself. Fixing an artifact left from that


jb55:
ACK

PR #24, opened 2022-07-28
NIP-23: Supported Feature Signaling

Giszmo:


fiatjaf:
This is made to be read by humans only? Who decides the names of the features?

Giszmo:
This nip defines two arrays. The first is meant to be clear commitments about well defined features that have generally agreed upon definitions but features front-run specs always, so client devs need a way to communicate features - also to find consensus.

Lets say Astral had implemented this nip already. When visiting Damus user profiles I would see some form of

> Unknown features supported by this user are 'boost text notes' and 'quoted boost text notes'.

Now if Astral were quick to respond, it could render those two lines differently:

> This user uses a client supporting reactions to posts. Contribute [here]() if you want to see reactions in Astral, too!

The other way around, Damus could detect that user doesn't support this and whenever its users want to like some text note it could pop up a hint for 2s: "Giszmo probably can't see your :heart:. Tell them to switch to Damus!".

staab:
Closing due to inactivity

PR #25, opened 2022-07-29
NIP-24: Prevent Data Deletion in Replaceable Events

Giszmo:


Giszmo:
In the nip itself I did not mention the ugliness of actually ensuring to have user interaction before deleting stuff. If I developed a chess client that stored the user profile in 10101 and now a user happens to have some encrypted blob of data there, how should my app react? To just replace it as if it were a normal update of the user profile is what apps are doing now. Advertising to support this app means that this should never happen but what else should happen?

* Tell the user that there is some weird data and ask for permission to overwrite it, resulting in an edit war between two apps if the other app is still in use?
* Use the next higher `kind` until one is found that has no bogus data?
* Should overwriting other apps' data an option at all?

At the very least I hope with this nip to get devs to stop dropping properties of events they generally handle as in "Alice" case.

fiatjaf:
I like the name "Protect the Unknown" and I would very much love to see this feature implemented, but I think it is a tricky one to get right when implementing things.

Giszmo:
> I like the name "Protect the Unknown" 

I'm for changing the nip name, too. Also for reasons you get into below ...

> and I would very much love to see this feature implemented, but I think it is a tricky one to get right when implementing things.
> 
> That is to say I shouldn't trust a client to not replace my events even if it claimed it wouldn't. Ultimately I would have to not use any client unless I really knew what it would do and was happy with that, which requires manual inspection.
> 
> But I like this NIP nonetheless, even if it only serves as a way to point to client makers how they are bad.

I agree completely. There can be very subtle details about data. If the client dev doesn't think of sorting being relevant, he might add tags that are referenced by index in the content to mention an obvious one.

Changing the name to "Protect the Unknown" would emphasise how this is not all to serious about implementation details but more a reminder to be kind to other projects. Client/Relay devs come here and build their checklists from nips and it's easy to forget that we are building stuff that should be interoperable in many ways. I'm not sure if this nip will end up as more than a reminder but a reminder nevertheless is important as existing projects already destroy the unknown a lot.

fiatjaf:
@Giszmo do you want to revisit this?

mikedilger:
It seems that (ideally) this should be `mandatory`, however...

What if there is an event out there that I am replacing, but I didn't know it was out there?  All I can do is make a best effort.

This might be secure scuttlebutt linked events in a chain [Also, because I can never know if I have all the past events].

Other than that the NIP looks good to me.

fiatjaf:
I think this is ok as just a recommendation we can point to. Nothing can ever be guaranteed.

Another mitigation technique that just came up to me is: to have relays (or some relays that decide to, or maybe just relays designed to serve as personal storage servers) store multiple versions of kind-0 events and then some clients can present all these to the user and allow the user to pick one (then republish that with the current date) or even merge the data from multiple.

Giszmo:
So clients hopefully consider this issue but it will ultimately be users and third party listings that will judge if a client plays nicely with other clients.

@Cameri I fixed the NIO typo. If you found others, please code-review and suggest fixes.

24.md
```diff
@@ -0,0 +1,57 @@
...
+
+### Different extensions to a feature
+
+Clients differ in their feature sets and as nips expand kinds, clients that do
+not consider unsupported parameters remove those when "replacing" events,
+breaking the clients' compatibility.
+
+Alice uses Branle at home and Damus on mobile. She configured `nip05` on Branle
+but when storing her `screen_name` on Damus using the kind `0` metadata, her
+`nip05` configuration disappeard. When configuring this on Damus again, her
```
eskema, line 29: this phrasing confuses me. Is Damus the intended client here again? by configuring what again on Damus makes her loose the screen_name? is it supposed to be Branle again instead?

24.md
```diff
@@ -0,0 +1,57 @@
...
+======
+
+Protect the Unknown
+-------------------
+
+`draft` `optional` `client` `author:giszmo`
+
+## Replaceable Events
+
+A *replaceable event* as defined in [NIP-01](01.md) and [NIP-16](16.md).
```
monlovesmango, line 11: are kind 0 and kind 3 considered a replaceable event? If so should NIP-16 be amended to state that? If not, should we explicitly state these are also included in this NIP?

it seems that this does cover kind 0 and kind 3, but the definition in NIP-16 of replaceable events doesn't include these.

24.md
```diff
@@ -0,0 +1,57 @@
...
+Implementation
+--------------
+
+Nostr does not provide a means by which a client may know it has seen the event
+it is replacing, resulting in a theoretical impossibility of guaranteeing this
+NIP's promise but by advertising the client to support this NIP, it promises to
+at least attempt to avoid interference with other apps using the same `kind`s.
+
+At the very minimum, a client should not drop data that itself does not use, so
+instead of serializing the own data structure, it should update the generic
```
monlovesmango, line 55: "instead of serializing the own data structure" doesn't seem clear to me. maybe "instead of only serializing its own client data" or something similar would be better?

24.md
```diff
@@ -0,0 +1,57 @@
...
+JSON object found in `body` or Array<Array<String>> in `tags`.
```
monlovesmango, line 56: do you mean 'content' (instead of 'body') here?

monlovesmango (commented):
good NIP to have. 

in your opinion, if a client is making best effort to abide by this NIP would that include first querying for the kind they are saving? If so should that be recommended in this NIP as well?

24.md
Giszmo: kind 0 and 3 are covered by nip-01 mentioned above

24.md
Giszmo: fixed

24.md
```diff
@@ -0,0 +1,57 @@
...
+Clients differ in their feature sets and as nips expand kinds, clients that do
+not consider unsupported parameters remove those when "replacing" events,
+breaking the clients' compatibility.
+
+Alice uses Branle at home and Damus on mobile. She configured `nip05` on Branle
+but when storing her `screen_name` on Damus using the kind `0` metadata, her
+`nip05` configuration disappeard. When configuring this on Damus again, her
+`screen_name` disappeared from her metadata.
+
+Bob's client invited him to define [NIO-02](02.md) petnames. And so he did. For
```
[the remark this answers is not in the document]
Giszmo: fixed

24.md
monlovesmango: yes but nip-01 does not define them as a replaceable event. was pretty confusing to me but just my opinion. i know kind 0 and kind 3 are replaceable, but nip-16 explicitly defines what we consider to be a "replaceable event". if we want to use terms like this we should be clear on the definition and not use it in different contexts in different nips.

monlovesmango:
@Giszmo in your opinion, if a client is making best effort to abide by this NIP would that include first querying for the kind they are saving? If so should that be recommended in this NIP as well?

Giszmo:
> @Giszmo in your opinion, if a client is making best effort to abide by this NIP would that include first querying for the kind they are saving? If so should that be recommended in this NIP as well?

I would suspect the client to be subscribed to that kind anyway as it's relevant to the client.

24.md
```diff
@@ -0,0 +1,57 @@
...
+
+Implementation
+--------------
+
+Nostr does not provide a means by which a client may know it has seen the event
+it is replacing, resulting in a theoretical impossibility of guaranteeing this
+NIP's promise but by advertising the client to support this NIP, it promises to
+at least attempt to avoid interference with other apps using the same `kind`s.
+
+At the very minimum, a client should not drop data that itself does not use, so
```
monlovesmango, line 55: could we add subscribing to existing events to the minimum requirements? like:

At the very minimum, a client should:
- subscribe to existing events of the relevant kind
- not drop data that itself does not use, so
instead of only serializing its own client data, it should update the generic
JSON object found in `content` or Array<Array<String>> in `tags`

mikedilger (approved):
Looks good and I think this is needed.

24.md
```diff
@@ -0,0 +1,57 @@
...
+
+Protect the Unknown
+-------------------
+
+`draft` `optional` `client` `author:giszmo`
+
+## Replaceable Events
+
+A *replaceable event* like `kind: 0`, `kind: 3` and as defined in
+[NIP-16](16.md).
```
monlovesmango, line 12: maybe update to include nip 33 too?

A *replaceable event* like `kind: 0`, `kind: 3` and as defined in
[NIP-16](16.md) and [NIP-33](33.md).

monlovesmango (approved):
lgtm

Giszmo:
Absent vetos within 24h I would merge this.

monlovesmango:
does this cover nip 33?

Giszmo:
> does this cover nip 33?

@monlovesmango in spirit it did. Now it does explicitly. The nip is a reminder so I hope updating it in the future doesn't turn into a burden.

staab:
I think this goes more under "best practices" than "protocol definition". There's been some discussion about how to design events going forward to avoid too many conflicts like this, but I think at the end of the day client developers will figure this out or gain a bad reputation as a result of stomping people's data, I don't think it's something for the protocol.

Closing due to inactivity, re-open if desired.

PR #26, opened 2022-07-30
NIP-25: Reactions

jb55:
:+1: 

https://github.com/nostr-protocol/nips/blob/reactions/25.md

### DIFF
--- a/25.md
+++ b/25.md
@@ -9,8 +9,8 @@ Reactions
 
 A reaction is a `kind 7` note that is used to react to `kind 1` text notes.
 
-The generic reaction, represented by an empty string, SHOULD be interpreted as
-a "like".
+The generic reaction, represented by the `content` set to a `+` string, SHOULD
+be interpreted as a "like" or "upvote".
 
 A reaction with `content` set to `-` SHOULD be interepreted as a "dislike" or
 "downvote". It SHOULD NOT be counted as a "like", and MAY be displayed as a
--- a/25.md
+++ b/25.md
@@ -1,0 +1,43 @@
+
+NIP-25
+======
+
+Reactions
+---------
+
+`draft` `optional` `author:jb55`
+
+A reaction is a `kind 7` note that is used to react to `kind 1` text notes.
+
+The generic reaction, represented by an empty string, SHOULD be interpreted as
+a "like".
+
+The `content` MAY be an emoji, in this case it MAY be interpreted as a "like",
+or the client MAY display this emoji reaction on the post.
+
+Tags
+----
+
+The reaction event SHOULD include `e` and `p` tags from the note the user is
+reacting to. This allows users to be notified of reactions to posts they were
+mentioned in. Including the `e` tags enables clients to pull all the reactions
+associated with individual posts or all the posts in a thread.
+
+The last `e` tag MUST be the `id` of the note that is being reacted to. 
+
+The last `p` tag MUST be the `pubkey` of the event being reacted to.
+
+Example code
+
+```swift
+func make_like_event(pubkey: String, privkey: String, liked: NostrEvent) -> NostrEvent {
+    var tags: [[String]] = liked.tags.filter { 
+    	tag in tag.count >= 2 && (tag[0] == "e" || tag[0] == "p") 
+    }
+    tags.append(["e", liked.id])
+    tags.append(["p", liked.pubkey])
+    let ev = NostrEvent(content: "", pubkey: pubkey, kind: 7, tags: tags)
+    ev.calculate_id()
+    ev.sign(privkey: privkey)
+    return ev
+}
--- a/25.md
+++ b/25.md
@@ -12,6 +12,12 @@ A reaction is a `kind 7` note that is used to react to `kind 1` text notes.
 The generic reaction, represented by an empty string, SHOULD be interpreted as
 a "like".
 
+A reaction with `content` set to `-` SHOULD be interepreted as a "dislike" or
+"downvote". It SHOULD NOT be counted as a "like", and MAY be displayed as a
+downvote or dislike on a post. A client MAY also choose to tally likes against
+dislikes in a reddit-like system of upvotes and downvotes, or display them as
+separate tallys.
+
 The `content` MAY be an emoji, in this case it MAY be interpreted as a "like",
 or the client MAY display this emoji reaction on the post.
 
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@ NIPs stand for **Nostr Implementation Possibilities**. They exist to document wh
 - [NIP-14: Subject tag in text events.](14.md)
 - [NIP-15: End of Stored Events Notice](15.md)
 - [NIP-16: Event Treatment](16.md)
+- [NIP-25: Reactions](25.md)
 
 ## Event Kinds
 
@@ -29,6 +30,7 @@ NIPs stand for **Nostr Implementation Possibilities**. They exist to document wh
 | 3    | Contacts                  | 2    |
 | 4    | Encrypted Direct Messages | 4    |
 | 5    | Event Deletion            | 9    |
+| 7    | Reaction                  | 25   |
 
 Please update this list when proposing NIPs introducing new event kinds.
 


jb55:
Wasn't sure if it was worth defining an explicit generic "downvote", perhaps content could be `+` `-` for something more explicit.

jb55:
How about:

A reaction with `content` set to `-` SHOULD be interepreted as a "dislike" or
"downvote". It SHOULD NOT be counted as a "like", and MAY be displayed as a
downvote or dislike on a post. A client MAY also choose to tally likes against
dislikes in a reddit-like system of upvotes and downvotes, or display them as
separate tallys.

Cameri (approved):

fiatjaf:
Looks simple good enough to me.

I like the `-` thing in the content. And then should we assume that anything else is a `+`? Or just an empty content is a `+`?

Maybe forcing a `+` (and discarding anything else) is clearer, and opens space for people adding other contents with other meanings in the future (like emojis or, who knows, sound and animation reactions or whatever the young people like), not that I would be encouraging such things.

fiatjaf:
The idea I had in my head for actually showing reactions in practice without having to list them all was to have the relays optionally include a `"likes"` field in the main event JSON object.

If the client trusts that aggregated number from the relay it could just display that number of likes and call it a day.

(This is in response to @jb55's comments [elsewhere](https://t.me/nostr_protocol/27123) about adding a `COUNT` websocket request.)

unclebob:
Just for the record, there are many things I don't like about this.  

1. The concept in general was invented to feed commercial algorithms on twitter, facebook, etc.  It doesn't actually serve a good purpose in a non-commercial setting.
2. Likes are cheap.  They don't really mean anything other than "yeah, cool, whatever."  If you want to react to something, take the time and thought to react with WORDS.  (what a concept!) (@jb55 I thought your idea of tipping was far better than zero cost likes)
3. I hate the idea that nostr relays are updating fields (e.g. counters for likes).  Updates mean transactions and locks, and concurrent update issues and...  I would much prefer nostr relays be simple event streams with no updates.  Events can be written once, read many times, but never updated, and never deleted.  The clients can work out the state of things.

Anyway, that's my rant.

fiatjaf:
Just for the record, here are my answers to @unclebob:

1. I agree, I also personally dislike likes and reactions, but this would have to happen at some point. Luckily it will have less meaning in a decentralized protocol and maybe overtime it can die.
2. I also agree with this.
3. The idea of relays updating counters was a storage optimization idea, completely optional. The Nostr concept in my mind was always about relays and clients being free to optimize and implement new optional features, always over a common standardized base which is NIP-01. Don't you think, for example, that relays should delete old "metadata" events? They don't need to, of course, but it would bring me pain and headache to run a relay that would store thousands of gigabytes of completely useless data no one wants (old, replaced, metadata events) -- of course you can disagree.

RandyMcMillan:
It would be great to have a neutral count for certain reactions such as 👀

Each platform may have their own choice of emojis but they would be weighted as such:

like/👍 = +10

+1 thru +9

neutral/👀 = N (effectively zero but counted as a neutral)

-1 thru -9

dislike/👎 = -10

By using +/- 10 we can proactively setup support for a gradient reaction scheme to support such things as sad/😔 = -3 & happy/🤗 = +5 etc...

unclebob:
...and just to answer @fiatjaf point 3.  What's a gigabyte here and there between friends?  A terabyte here, a terabyte there, and after awhile you might be talking about real storage capacity.  

In any case, there is real history in all those old metadata events.  History that (to borrow a phrase) deserves to be remembered.   <grin>

Giszmo:
I want "likes" in principle. And dislikes. Yes @fiatjaf @unclebob, "likes" are cheap but they are also saying "I hear you", which a social network is about. People want to be heard and users who share 20 events and get no reply will not be satisfied and getting that one :eyes: will be their kicker of the day. Take that from them and they will not use the platform. Sure, this is a sad example but we are all humans and not being heard hurts. Let's provide the cheapest way possible to say "you have been heard" so that all that get heard also get that feedback.

(All the big platforms don't like dislikes. They give you a cheap way to send love but for hate, you really have to use words. While on the surface, the only positive reactions look like harmony, I'm convinced they are a reason for controversy as now all negativity gets expressed with sarcasm, ridicule and insults instead of a simple :-1:.)

Now the implementation does not address the problem with multiple likes per person and whoever does the tally should avoid giving users an incentive to spam the platform with 1000 likes per event but if any popular client screws up and tallies favourable to such trolls, relays will suffer (or flourish if they can charge for the queries).

Clients might want to query "likes" of follows in order to discover new content the user might want to see just like Twitter does.

a5an0:
empty string being a “like”, - being “dislike”, and anything else being application specific or a like seems to be special-casing “likes” above other reaction types. I think the more general here is that you want to be able to increment or decrement counters that have context-specific meaning. Maybe “-“ can be generalized? My application might want to send “like” or “-like”, but it might also want to send “🪙” when a payment is sent and “-🪙” when a payment is canceled, or in a social game you might want something like “point” when a point is scored and “-point” when a point is lost. This is all just “likes and dislikes” but for things other than “likes”.

I think the literal “-“ prefix is probably not the way to go. Maybe a field for increment/decrement?

Giszmo:
I like the use of emojis that can be displayed in notes "as is" and slightly dislike the implicit `"" == "👍"` but Damus does it this way and it's not too far off, so I'd roll with it. How clients might tally reactions is maybe not necessarily subject to this nip.

GitHub for example allows each users to add multiple different reactions to comments. Do we want this on nostr social? Do we want to prescribe certain limits? Are we concerned about relays getting flooded with likes if some client shows posts as having "1000 likes" after the author himself liked his own post that much? Or might users fake engagement by using all the reactions supported by UTF8 if each was limited by 1 per user but not the number of reactions of a user per event in the displaying clients or tally logic on relays?

jb55:
> Maybe forcing a `+` (and discarding anything else) is clearer, and opens space for people adding other contents with other meanings in the future (like emojis or, who knows, sound and animation reactions or whatever the young people like)(not that I personally would appreciate such things).

I will update it to this. damus creates empty "" but I should probably just update it to do `+` instead.

jb55:
> The concept in general was invented to feed commercial algorithms on twitter, facebook, etc. It doesn't actually serve a good purpose in a non-commercial setting.

This is bizarre viewpoint to me, sending emojis is a huge part of social internet culture, regardless if companies use it for data science purposes.

> I hate the idea that nostr relays are updating fields (e.g. counters for likes). Updates mean transactions and locks, and concurrent update issues and... I would much prefer nostr relays be simple event streams with no updates. Events can be written once, read many times, but never updated, and never deleted. The clients can work out the state of things.

agreed, not sure how talks about stateful counters came up here, and idea I never have suggested. You can always aggregate event streams to do totals which I think is the correct way to do this.

I'm surprised by the the amount of discussion of subjective opinion of likes. how does that matter at all? There will be types of notes on nostr that I disagree with, but I don't have to query those notes so it doesn't matter. all I'm doing is spec'ing how damus implements them and if other clients want to interact with likes they can implement this, if not they can ignore it.

Reactions are everywhere. They are in basically every social app that exists. I can't imagine not speccing this because it's going to happen and there already is an implementation in the wild (damus).

As for dislikes, it is a think that exists in the wild (reddit) so I wanted to support it in this NIP even if damus will probably not implement it. damus might have a small set of emojis for reactions in the future though.

As for the +10 idea to indicate like strength, I think this overcomplicates the spec for little gain.

> empty string being a “like”, - being “dislike”, and anything else being application specific or a like seems to be special-casing “likes” above other reaction types. I think the more general here is that you want to be able to increment or decrement counters that have context-specific meaning. Maybe “-“ can be generalized? My application might want to send “like” or “-like”, but it might also want to send “coin” when a payment is sent and “-coin” when a payment is canceled, or in a social game you might want something like “point” when a point is scored and “-point” when a point is lost. This is all just “likes and dislikes” but for things other than “likes”.

applications can just create a new kind if they need more application specific logic. I dislike the idea of overloading a NIP with additional functionality, which just makes it harder for other clients to integrate with it. In the future there will be many NIPs, the simpler each individual NIP is the better imo.

jb55:
Updated spec to make + explicit:

```diff
--- a/25.md
+++ b/25.md
@@ -9,8 +9,8 @@ Reactions
 
 A reaction is a `kind 7` note that is used to react to `kind 1` text notes.
 
-The generic reaction, represented by an empty string, SHOULD be interpreted as
-a "like".
+The generic reaction, represented by the `content` set to a `+` string, SHOULD
+be interpreted as a "like" or "upvote".
 
 A reaction with `content` set to `-` SHOULD be interepreted as a "dislike" or
 "downvote". It SHOULD NOT be counted as a "like", and MAY be displayed as a
```

Giszmo:
ACK

(as jb55 said, it formalizes what's already implemented in Damus. All questions and doubt were about the merit of reactions in the first place and potentially relays getting spammed. Client and relay devs will figure out how to optimize the communication of these reactions. For now we have no obvious way of achieving the suggested feature in a fundamentally better way.)

jb55:
> ACK
> 
> (as jb55 said, it formalizes what's already implemented in Damus. All questions and doubt were about the merit of reactions in the first place and potentially relays getting spammed. Client and relay devs will figure out how to optimize the communication of these reactions. For now we have no obvious way of achieving the suggested feature in a fundamentally better way.)

spam is a problem, I have created 100 fake likes on a post before using nostril. that's outside the scope of this as I can't think of a good way to not count these. I felt this problen falls in line with every other spam issue on nostr, and maybe likes won't matter as much in the future like @fiatjaf said if they are easily faked anyways. This is why I'm looking forward to working on lightning tipping next.

WRT optimizations... I expect in the future there will be a NIP that aggregates notes based on type and perhaps content, so you don't have to fetch all 10,000 likes (which would be expensive).

Giszmo:
I imagine clients could consider only stuff referenced by follows (f) or follows of their follows (f2) etc. This way, the total amount of relevant events shrinks. Also such clients would not count likes from pubkeys outside their f2 or f3 for example. This could be very effective against spam, making the sphere of influence of each user an invite-only thing.

Users that want more content, could request f4 or f5 but spammers would have a hard time to break into my experience if they need to be part of my web of follows.

PR #30, opened 2022-08-10
nip-01.md: kind: 0 - set_metadata conform to a standard event

RandyMcMillan:
added an example

### DIFF
--- a/01.md
+++ b/01.md
@@ -97,10 +97,65 @@ This NIP defines no rules for how `NOTICE` messages should be sent or treated.
 
 ## Basic Event Kinds
 
-  - `0`: `set_metadata`: the `content` is set to a stringified JSON object `{name: <username>, about: <string>, picture: <url, string>}` describing the user who created the event. A relay may delete past `set_metadata` events once it gets a new one for the same pubkey.
-   * Where `<username>` is a string that matches the pattern: `[\w+\-]` (java regular epression).  Or, in other words, a sequence of the following
+  - `0`: `set_metadata`: Structure
+
+
+
+##### `[ "EVENT", {<event id>, <pubkey>, <created_at>, <kind>, ...`
+
+##### `"tags": [["name", "<regular_expression>"], ["about", "<string>"], ["picture", "<url>"], ... `
+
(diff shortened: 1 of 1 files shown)


fiatjaf:
This actually does look good, but we can't do this at this point. The benefits are not worth the cost.

We could however add a recommendation to NIP-01 for future event kinds that need different kinds of metadata to add them as tags instead of as stringified JSON on content.

RandyMcMillan:
Yes detecting a "legacy" and/or an upgraded kind: 0 is reasonable.

RandyMcMillan:
Being able to add an OPENPGP: "<fingerprint>" detection would be an added value - and enable alternate encryption schemes etc...

RandyMcMillan:
@fiatjaf - if you want to post a patch to the wording I will amend the PR.

fiatjaf:
No, what I was saying is that this PR can't be merged.

What is the goal you have in mind for this? Being able to add an openpgp fingerprint to the metadata event? You can do it already today if you want, either in the JSON body or as a tag.

RandyMcMillan:
The OPENPGP mention was just an example use case - the idea is to expand kind: 0 set_metadata to enable additional profile information as well as to simply conform to a standard EVENT. Maybe I am missing something (my apologize). The structure on the set_metadata is unclear. More generally these partial examples in the documentation are a little too terse - they should include example JSON in context of an entire EVENT (some do - but not all). This would clarify a lot for me and other people trying to implement NIPs. Thanks for your patience. NOTE: This is probably why relay implementations are diverging and/or are slow to add support for NIPs.

RandyMcMillan:
This PR was started initially because I attempted to push a set_metadata EVENT to update my profile and it broke my profile metadata. Please post an example of an entire EVENT to do this. Thanks.

Cameri:
@RandyMcMillan Just add that to include extra metadata you must do it in the tags. The content property should not be changed because that would break every single client out there.

jb55:
I closed this since it looks like this stemmed from a misunderstanding about the contents of kind0. feel free to reopen if you feel like it's still worth continuing the discussion.

PR #31, opened 2022-08-11
Update Readme for NIP 20

MTG2000:
I added what we discussed in a little more details to the ReadMe.
Please take a look, and update however you see fit.

### DIFF
--- a/20.md
+++ b/20.md
@@ -8,22 +8,42 @@ Web Comments
 
 A special event with kind `34` is defined, meaning "web comment".
 
-Events of this kind should have at least one `r` tag with the value set to the normalized URL of the webpage they're commenting in.
+Events of this kind should have both or at least one of the following tags:
 
-The content should be the plaintext of the comment.
+1- a `r` tag with the value that represents some kind of identifier for the object being commented on. 
+No specific format is enforced, but it preferably should be something that won't change later.  
+A possible format could be: `"<MY WEBSITE NAME> <POST TYPE> <POST ID>"`  
+So when used it'll look something like: `"Github Pull_Request 19"`.  
+This identifier can be placed in a `<meta name='identifier' content='...' />` tag on its page.  
+This will help later if someone wants to create a nostr commenting extension or widget that automatically injects comments on each page.  
+So this way this widget will be able to detect which filter it should be using in the query.  
 
-For example:
+2- a `u` tag with the value set to the normalized URL of the webpage they're commenting in.  
+This is meant to be used in case the identifier wasn't found in the page for whatever reason, so it fallbacks to the page normalized URL.  
 
+### **Why not use only the URL filter?? 🤔**  
+While this might work most of the time, depending only on the URL means that if the URL of the original webpage changes in the future for whatever reason, mean that all the old comments will not appear on the page anymore. So that's why we suggest using another tag that should be more constant.
+
+## Usage
+When you **post** a comment, the object should preferably contain **Both** tags:
 ```json
 {
   "kind": 34,
   "tags": [
-    ["r", "https://random.blog/article"]
+    ["r", "Github Pull_Request 19"],
+    ["u", "https://github.com/nostr-protocol/nips/pull/19"]
   ],
-  "content": "I didn't like this article.",
+  "content": "What is this PR about??",
   ...other fields
 ```
 
+And when **querying** for comments, you can use 2 filters with **OR** to get all the events that either has the `r` identifier, or the `u` url.
+```js
+sub({ cb: ()=>{...} filter: [ { "#r": "..."}, {"#u": "..."} ] })
+```
+
+
+ 
 URL Normalization
 -----------------
 


huumn:
I don't know how helpful this is, but would it be beneficial to add a standard for nested comments?

Afaict this would allow for nested comments, ie threads, where `POST TYPE` in `<MY WEBSITE NAME> <POST TYPE> <POST ID>` is perhaps a `Reply`, but without standardizing this it might leave things a little too open to interpretation given how common/desirable threads are. 

[Fiatjaf asked me to review](https://stacker.news/items/57844) and this is the only deficiency for an SN use case that stands out. It's also apparently doable without such a change though.

Note: I'm unfamiliar how NIPs handle cases like these typically and its probably best to conform to the level of under/over specifying of use cases that is typical.

jb55:
On Thu, Aug 11, 2022 at 01:13:05PM -0700, Keyan wrote:
>I don't know how helpful this is, but would it be beneficial to add a standard for nested comments?
>
>Afaict this would allow for nested comments, ie threads, where `POST TYPE` in `<MY WEBSITE NAME> <POST TYPE> <POST ID>` is perhaps a `Reply`, but without standardizing this it might leave things a little too open to interpretation given how common/desirable threads are.

nested comments/threads is just nip10. Any threaded discussion note type should just implement that.

20.md
```diff
@@ -8,22 +8,42 @@ Web Comments
...
+While this might work most of the time, depending only on the URL means that if the URL of the original webpage changes in the future for whatever reason, mean that all the old comments will not appear on the page anymore. So that's why we suggest using another tag that should be more constant.
+
+## Usage
+When you **post** a comment, the object should preferably contain **Both** tags:
 ```json
 {
   "kind": 34,
   "tags": [
-    ["r", "https://random.blog/article"]
+    ["r", "Github Pull_Request 19"],
```
jb55, line 33: single-letter tags should be reserved for things that will be indexed, it doens't look like this needs to be indexed.

20.md
```diff
@@ -8,22 +8,42 @@ Web Comments
...
+    ["u", "https://github.com/nostr-protocol/nips/pull/19"]
```
jb55, line 34: we should just keep the url as `r` and just use nip-14 for the subject.

jb55 (changes requested):

jb55:
On Thu, Aug 11, 2022 at 08:28:23PM -0700, William Casarin wrote:
>On Thu, Aug 11, 2022 at 01:13:05PM -0700, Keyan wrote:
>>I don't know how helpful this is, but would it be beneficial to add a standard for nested comments?
>>
>>Afaict this would allow for nested comments, ie threads, where `POST TYPE` in `<MY WEBSITE NAME> <POST TYPE> <POST ID>` is perhaps a `Reply`, but without standardizing this it might leave things a little too open to interpretation given how common/desirable threads are.
>
>nested comments/threads is just nip10. Any threaded discussion note type should just implement that.

After taking a look at nip20 it looks like the `r` tag could be
considered the root tag, so it's a bit simpler than nip10. Each reply
just needs to include the r tag + e tag of the note you're replying to.
It's still probably a good idea to add a `reply` marker to the e tag
though.

Cheers,

20.md
MTG2000: You mean to use the `subject` field to put the identifier that is currently suggested for the `r` tag??

jb55:
>You mean to use the `subject` field to put the identifier that is currently suggested for the `r` tag??

["r", "github.com/nostr-protocol/nips/pull/19"]
["subject", "Github Pull_Request 19"]

MTG2000:
Okay.
Btw, the `subject` field can be included in the query filter, right??
meaning I can construct a query like:
`sub({ cb: ()=>{...} filter: [ {"subject": "..."} ] })`

MTG2000:
> meaning I can construct a query like:
> `sub({ cb: ()=>{...} filter: [ {"subject": "..."} ] })`

Maybe I did that wrong, but when I tried using this, it seemed that the `subject` field isn't used when filtering the events returned currently.

eskema:
> 



> Maybe I did that wrong, but when I tried using this, it seemed that the `subject` field isn't used when filtering the events returned currently.

you need to add "#" on the query, so use filter: `{"#subject": "..."}`

jb55:
> Okay.
> Btw, the `subject` field can be included in the query filter, right??
> meaning I can construct a query like:
> `sub({ cb: ()=>{...} filter: [ {"subject": "..."} ] })`

sure but usually you wouldn't want to search on a subject, these are more useful for things like hashtags `#t` and urls `#r` not arbitrary text content. what's your use case here?

MTG2000:
> what's your use case here?

As I mentioned in the NIP, the reason why I'm suggesting the use of something more "stable\constant" such as this, is because the URLs can change. 
And although this might not be something that happens very commonly or frequently, but it happens.
And if we are only depending on the url tag to query these comments, then that means we will all the comments if the URL change.

But if we manage to generate/extract a unique identifier for a page, that won't change even if we change the URL, and we put it for example in a `<meta>` tag in the page headers, then we can almost guarantee that we will be able to get all the comments for this post/article/question/page/whatever.

And we can also keep the URL tag for the cases where the comments client knows the URL but can't get this identifier, so it'll still be able to get them.

MTG2000:
> you need to add "#" on the query, so use filter: `{"#subject": "..."}`

Ohh, good to know.
It's working now correctly 👍

jb55:
> As I mentioned in the NIP, the reason why I'm suggesting the use of something more "stable\constant" such as this, is because the URLs can change. And although this might not be something that happens very commonly or frequently, but it happens. And if we are only depending on the url tag to query these comments, then that means we will all the comments if the URL change.

I don't think there's a way to solve this is in general. web clients are going to be using the url as the root id anyways. I don't see how they would could know about this unique identifier? If it's created from the first note that comments, then you might as well just use nip10 + the first note id as the root id.

If you want something like a chatroom with a stable channel id you can just use that chatroom NIP that people are working on.

MTG2000:
> I don't see how they would could know about this unique identifier

Yeah...
I agree with you on this, that there isn't really a perfect way for other clients to get this identifier.
But at least there is one page (the original webpage) that has access to this identifier where all the comments can be seen.
For other external clients who don't have access to this identifier, they can use the URL, keeping in mind that they might be missing some comments.
So I think it's still useful.

And if the external clients want to access this identifier, there might be a way to expose it:
1- Including it in the URL as a query parameter. 
2- Fetching the page then scraping the `<meta/>`  tag, but this most likely needs to be done on a server to avoid CORS issues, and the identifier be cached somewhere so that it's not fetched each time. 


But still, even considering that external clients won't be able to get the identifier and only the URL, at least the original client has access to all the comments.

20.md
```diff
@@ -8,22 +8,42 @@ Web Comments
 
 A special event with kind `34` is defined, meaning "web comment".
 
-Events of this kind should have at least one `r` tag with the value set to the normalized URL of the webpage they're commenting in.
+Events of this kind should have both or at least one of the following tags:
 
-The content should be the plaintext of the comment.
+1- a `r` tag with the value that represents some kind of identifier for the object being commented on.
```
Giszmo, line 13: Please use markdown here and below.

```suggestion
1. An `r` tag with the value that represents some kind of identifier for the object being commented on. 
```

20.md
```diff
@@ -8,22 +8,42 @@ Web Comments
...
+1- a `r` tag with the value that represents some kind of identifier for the object being commented on. 
+No specific format is enforced, but it preferably should be something that won't change later.  
+A possible format could be: `"<MY WEBSITE NAME> <POST TYPE> <POST ID>"`  
+So when used it'll look something like: `"Github Pull_Request 19"`.  
+This identifier can be placed in a `<meta name='identifier' content='...' />` tag on its page.  
+This will help later if someone wants to create a nostr commenting extension or widget that automatically injects comments on each page.  
+So this way this widget will be able to detect which filter it should be using in the query.  
 
-For example:
+2- a `u` tag with the value set to the normalized URL of the webpage they're commenting in.
```
Giszmo, line 21: The empty line above needs to be removed, too for markdown to enumerate correctly.

```suggestion
2. A `u` tag with the value set to the normalized URL of the webpage they're commenting in.  
```

20.md
Giszmo: Why should this not need an index? It will get queried for.

Giszmo (changes requested):

PR #32, opened 2022-08-13
add nip 23: relays list.

fiatjaf:
For sharing relays lists between clients in a reasonable and extensible manner.

https://github.com/nostr-protocol/nips/blob/26-relays-list/26.md


26.md
```diff
@@ -0,0 +1,29 @@
...
+Relays List
+-----------
+
+`draft` `optional` `author:fiatjaf`
+
+A special event with kind `10001`, meaning "relay list" is defined as having a list of tags, one for each relay the author uses.
+
+The content is not used.
+
+The tags consist of arrays of 3 elements: the first is the relay URL, the second is the "read" condition -- which if set `"true"` means the client is reading events from that relay --, the third is the "write" condition -- which if set to `"true"` means the user is writing events to that relay.
```
Cameri, line 13: Instead of using "true" or "false", I think it's better to use a sequence of two strings: either `read` or `write` followed by `scope` which is runes-like.

- If you only want to read notes:
  `["wss://rss.relay.com", "read", "kind=1"]`

- If you only want to communicate with `aabb` over this relay:
  `["wss://private.relay.com", "read", "kind=4 & pubkey^aabb", "write", "kind=4 & p^aabb"]`

- If you only want to set metadata on this relay:
  `["wss://rss.relay.com", "write", "kind=0"]`

Read and write can repeat many times to add more scopes. The allow everything scope is an empty string.

fiatjaf: I like that.

fiatjaf: OK, I tried to implement your proposal and I think it is considerably more annoying to implement than just `["wss://relay.com", "kinds^4", "kind=4"]` (the first condition is the "read" one, the second is the "write" one). What is the downside of this?

Cameri: No downside. I think it's good.

fiatjaf: The downside is that there is no way to specify a condition that evaluates to `false` (while on your proposal you can just omit that attribute).

fiatjaf: 2022-08-13-194848_579x366_scrot

Actually according to this the `#` can be used for `true` and then we can use `""` for `false`.

fiatjaf: Hmm, but that is no good as it conflicts with tag queries that begin with `#` very badly.
Also it means "a comment" which implies an empty string should be the same as a string with a comment on it, so it makes no sense to evaluate `""` differently than `"#"`.

We should probably change so the naked operator `"!"` means `false` and then we can use `""` to mean `true`.

fiatjaf: See my next proposal: https://github.com/nostr-protocol/nips/pull/32/commits/418cd859fcb12dd6d4ffc5a1e00880b6039b6c09

Cameri: `kind<0` seems hacky but would result in false

26.md
```diff
@@ -4,25 +4,61 @@ NIP-26
...
 
-In the future this NIP will be modified with more fine-grained conditions for writing and reading from relays, based on properties of the filter or of the event respectively, written as strings in a simple language.
+The "read" condition consists of a string containing a rule that follows a subset of the [runes](https://pypi.org/project/runes/) language. Specifically only the `|`, `&`, `!`, `>`, `<`, and `=` operators are allowed. When the rule is an empty string it evaluates to `true`. When it consists of a single naked `"!"` it evaluates to `false`. The `#` specifically must not be interpreted as a comment, but always as a tag attribute. All the operatores must be tested against all possible values in the case of filter values that consist of lists of values and also in the case of event tags that can have multiple values for the same tag -- in other words, the `=` may be interpreted as a `values.any(v => v == runeValue)` instead of an `values == runeValue` operator; the `<` may be interpreted as a `values.any(v => v < runeValue)` and so on. In the case of events tags must also be identified by the `#` prefix, just like in filters.
```
Cameri, line 15: Let's please include `^` aka begins with. With it it's possible to use prefixes for pubkeys and event IDs.

Cameri: Would it be easier if we drop the "!" evaluating to false and just pass some condition that will always evaluate to false, like `kind<0`?

Cameri: Ah, `\` is not allowed in FIELDNAME. I thought I was getting somewhere!

Cameri: Any fieldname other than `kind`, `pubkey`, `created_at`, `id` could mean you are searching in tags so there's no need to include the `#` to prefix them.

26.md
```diff
@@ -4,25 +4,61 @@ NIP-26
...
+      - `<empty>`: `true`
+      - `!`: `false`
+      - `authors=7890`: `false`
+      - `authors=7890|authors=1234`: `true`
+      - `authors=7890&authors=1234`: `false`
+      - `#e!`: `true`
+      - `#e=5555`: `false`
+      - `kinds=1|kinds=4`: `true`
+      - `kinds<2`: `true`
+      - `kinds>7`: `false`
```
Cameri, line 40: - `""`: `true`
      - `"!"`: `false`
      - `"authors=7890"`: `false`
      - `"authors=7890|authors=1234"`: `true`
      - `"authors=7890&authors=1234"`: `false`
      - `"e!"`: `true`
      - `"e=5555"`: `false`
      - `"kinds=1|kinds=4"`: `true`
      - `"kinds<2"`: `true`
      - `"kinds>7"`: `false`

26.md
```diff
@@ -4,25 +4,61 @@ NIP-26
...
+  - **write**
+
+    - for the event `{"kind": 7, "content": "banana", "tags": ["p", "6677"], "created_at": 123456789, "pubkey": "e3e3"}`
+
+      - `<empty>`: `true`
+      - `!`: `false`
+      - `pubkey=7890`: `false`
+      - `pubkey=e3e3`: `true`
+      - `kind=7&#p=6677`: `true`
+      - `created_at>999999999|#e=5a5a`: `false`
```
Cameri, line 51: - `""`: `true`
      - `"kind<0"`: `false`
      - `"pubkey=7890"`: `false`
      - `"pubkey=e3e3"`: `true`
      - `"kind=7&p=6677"`: `true`
      - `"created_at>999999999|e=5a5a"`: `false`

Giszmo:
So is this meant to help with sharding later on? This relay for kinds x to y, this relay for nwiki, ... ? But any pubkey could advertise any such list of relays and their limitations? What should clients do with those? They will not be consistent between different pubkeys and who would share them based on what? Based on relays self-reported filters?

I see this nip as benign but don't see the benefit yet.

Cameri:
> I see this nip as benign but don't see the benefit yet.

@Giszmo  There's currently no NIP specifying how a client should store a list of relays. Branle stores the list of relays in the `content` field of kind 3 events but that's not covered in NIP-02. This NIP fixes that.

Giszmo:
I find the rules quite complex and doubt many clients will implement all of it. I agree with @Cameri to not use `#`. (I don't agree with dropping `!` and adding `^`.)

Cameri:
^ is already part of the runes-like spec and means starts with or begins with. It can be skipped for now since adding support for it won't break existing rules.
I think it is fine to implement a subset, just what we need.
But our usage of # and ! are not part of the original runes-like spec and would make it non-compliant for whatever it's worth, breaking compatibility with existing libraries.
Anyone wanting to implement this NIP will have no choice but to implement the custom spec, not a hard problem true, but they could have just leveraged the existing python library for example.
Either way this not a deal breaker for me but I don't find it ideal.
------- Original Message -------

AnonymousGeekDev:
Hello everyone. This spec is interesting(but complex, at least to me). Though I have partially understood, thanks to the examples used, and the event structure, I fail to understand the use  evaluation criteria in the examples(where conditions such as pubkey=(a value), etc are determined). How are they used? Let us imagine a scenario between 2 Nostr users. One person shares the list. Does the other person get the list with the values defined, or with the evaluation criteria? 
Thanks.

Cameri (approved):

README.md
```diff
@@ -19,7 +19,7 @@ NIPs stand for **Nostr Implementation Possibilities**. They exist to document wh
 - [NIP-15: End of Stored Events Notice](15.md)
 - [NIP-16: Event Treatment](16.md)
 - [NIP-25: Reactions](25.md)
-- [NIP-26: Relays List](26.md)
+- [NIP-23: Relays List](23.md)
```
Cameri, line 22: this should probably be sorted

23.md
```diff
@@ -0,0 +1,85 @@
...
+
+When a `read` rule evaluates to `true` for a given **filter** the client SHOULD send a `REQ` message to the relay, otherwise it SHOULD NOT. When a `write` rule evaluates to `true` for a given **event** the client SHOULD send that event (in a `EVENT` message) to the relay, otherwise it MUST NOT.
+
+When a rule is malformed or the client is unable to parse it for any reason (for example, for not having implemented all the operators) it SHOULD treat it as `true` if it is a _read_ rule and `false` if it is a _write_ rule.
+
+### Purposes
+
+This NIP serves two purposes: (1) backup and interoperability of relay lists and relay specific rules between clients; and (2) sharing of relay URLs between users.
+
+The first use case is meant to make it so users can open their client -- or different clients -- in different devices and have their list of relays automatically fetched from a coon global relay and start using their relays list without having to set everything up again.
```
jeffthibault, line 27: What is a "coon" global relay? Typo?

23.md
```diff
@@ -0,0 +1,85 @@
...
+
+  - "do not use this relay for DMs": sets _write_  to `kind/4`
+  - "only use this relay for DMs": sets _write_ to `kind=4` and _read_ to `kinds=4`
+  - "this is Bob's personal relay, only use it to fetch Bob's events": sets _write_ to `!` and _read_ to `authors=<bob-pubkey>`
+  - "this relay is full of spambots, do not get note replies from this relay": sets _read_ to `(kinds=1&e!)|kinds/1`
+  - "this is my personal relay, only store my stuff in it": sets _read_ to `authors=<my-pubkey>` _write_ to `pubkey=<my-pubkey>`
+
+### Examples
+
+(Public keys are shortened to 3 characters for readability.)
```
jeffthibault, line 43: Looks like you used 4 characters

23.md
```diff
@@ -0,0 +1,85 @@
...
+
+A special event with kind `10001`, meaning "relay list" is defined as having a list of tags, one for each relay the author uses.
+
+The content is not used.
+
+The tags consist of arrays of 3 elements: the first is the relay URL, the second is the _read_ condition, the third is the _write_ condition.
+
+The _read_ condition consists of a string containing a rule that follows a subset of the [runes](https://pypi.org/project/runes/) language. Specifically only the `|`, `&`, `!`, `>`, `<`, `/` and `=` operators are allowed. When the rule is an empty string it evaluates to `true`. When it consists of a single naked `"!"` it evaluates to `false`. All the operatores must be tested against all possible values in the case of filter values that consist of lists of values and also in the case of event tags that can have multiple values for the same tag -- in other words, the `=` may be interpreted as a `values.any(v => v == runeValue)` instead of an `values == runeValue` operator; the `<` may be interpreted as a `values.any(v => v < runeValue)` and so on. Event tags are identified just by their tag key (for example, `e` or `p`) without the `#` prefix used in filters.
```
Cameri, line 15:
```suggestion
The _read_ condition consists of a string containing a rule that follows a subset of the [runes](https://pypi.org/project/runes/) language. Specifically only the `|`, `&`, `!`, `>`, `<`, `/` and `=` operators are allowed. When the rule is an empty string it evaluates to `true`. When it consists of a single naked `"!"` it evaluates to `false`. All the operators must be tested against all possible values in the case of filter values that consist of lists of values and also in the case of event tags that can have multiple values for the same tag -- in other words, the `=` may be interpreted as a `values.any(v => v == runeValue)` instead of an `values == runeValue` operator; the `<` may be interpreted as a `values.any(v => v < runeValue)` and so on. Event tags are identified just by their tag key (for example, `e` or `p`) without the `#` prefix used in filters.
```

23.md
fiatjaf: ahahah I have no idea

jeffthibault: 😆

23.md
```diff
@@ -0,0 +1,91 @@
+NIP-23
+======
+
+Relays List
+-----------
+
+`draft` `optional` `author:fiatjaf` `author:cameri` `author:monlovesango` `author:giszmo`
```
Giszmo, line 7:
```suggestion
`draft` `optional` `author:fiatjaf` `author:cameri` `author:monlovesmango` `author:giszmo`
```

23.md
```diff
@@ -0,0 +1,91 @@
...
+
+The `read` rule operates on the values of the [NIP-01](01.md) **filter** object, while the `write` rule operates on the values of the **event** object.
+
+When a `read` rule evaluates to `true` for a given **filter** the client SHOULD send a `REQ` message to the relay, otherwise it SHOULD NOT. When a `write` rule evaluates to `true` for a given **event** the client SHOULD send that event (in a `EVENT` message) to the relay, otherwise it MUST NOT.
```
Giszmo, line 19: Why be prescriptive in the `true` cases at all? Why "MUST NOT" in the false case for write? I see the "Don't re-send my DMs to any relay not compliant with nip-21" but then you also wouldn't want the client to REQ those events from a non-nip-21 relay? You would also not want to use more than one nip-21 relay for your DMs with any particular correspondence but would you want to put into your public event a list of your conspirators by relay?

```suggestion
When a `read` rule evaluates to `false` for a given **filter** the client SHOULD NOT send that filter in a `REQ` message to that relay.

When a `write` rule evaluates to `false` for a given **event** the client SHOULD NOT send that event in an `EVENT` message to that relay.
```

at the very least this typo and formatting:

```suggestion
When a `read` rule evaluates to `true` for a given **filter** the client SHOULD send that filter in a `REQ` message to the relay, otherwise it SHOULD NOT.

When a `write` rule evaluates to `true` for a given **event** the client SHOULD send that event in an `EVENT` message to the relay, otherwise it MUST NOT.
```

23.md
```diff
@@ -0,0 +1,91 @@
...
+
+When a rule is malformed or the client is unable to parse it for any reason (for example, for not having implemented all the operators) it SHOULD treat it as `true` if it is a _read_ rule and `false` if it is a _write_ rule.
```
Giszmo, line 21: This needs some justification I think.

23.md
```diff
@@ -0,0 +1,91 @@
...
+
+A special event with kind `10001`, meaning "relay list" is defined as having a list of tags, one for each relay the author uses.
+
+The content is not used.
+
+The tags consist of arrays of 3 elements: the first is the relay URL, the second is the _read_ condition, the third is the _write_ condition.
+
+The _read_ condition consists of a string containing a rule that follows a subset of the [runes](https://pypi.org/project/runes/) language. Specifically only the `|`, `&`, `!`, `>`, `<`, `/` and `=` operators are allowed. When the rule is an empty string it evaluates to `true`. All the operators must be tested against all possible values in the case of filter values that consist of lists of values and also in the case of event tags that can have multiple values for the same tag -- in other words, the `=` may be interpreted as a `values.any(v => v == runeValue)` instead of an `values == runeValue` operator; the `<` may be interpreted as a `values.any(v => v < runeValue)` and so on. Event tags are identified just by their tag key (for example, `e` or `p`) without the `#` prefix used in filters.
```
Giszmo, line 15: Why not include starts-with and ends-with from runes? Would make rules more compact with hardly any false positives and using libraries, implementers wouldn't care.

Giszmo (changes requested):
some minor issues

23.md
Cameri: I agree, please add starts with. For Ends with I can't think of a use case right now.

fiatjaf: Sure.

23.md
fiatjaf: I can't justify it to myself even. It's probably better to just default to `false` or leave it for the client to decide. I don't know.

jb55:
> The first use case is meant to make it so users can open their client -- or different clients -- in different devices and have their list of relays automatically fetched from a default relay and start using their relays list without having to set everything up again.

isn't this is what the contents of kind3 is for?

eskema:
> > The first use case is meant to make it so users can open their client -- or different clients -- in different devices and have their list of relays automatically fetched from a default relay and start using their relays list without having to set everything up again.
> 
> isn't this is what the contents of kind3 is for?

well, that was a rogue undocumented implementation in Branle, we kinda discussed this in the chat group that a new NIP dedicated for relays would be better.

fiatjaf:
Branle is now implementing a draft of this proposal that only recognizes empty strings as `true`.

23.md
```diff
@@ -0,0 +1,93 @@
...
+
+This NIP serves two purposes: (1) backup and interoperability of relay lists and relay specific rules between clients; and (2) sharing of relay URLs between users.
+
+The first use case is meant to make it so users can open their client -- or different clients -- in different devices and have their list of relays automatically fetched from a default relay and start using their relays list without having to set everything up again.
+
+For the second purpose, if any client decides to, they can show to the user what relays other users are using, suggest that or automatically add these to the user's relay list, this can take into account the rules or more likely not. The possibility of sharing a list of relays in standardized format is good for spreading information about relays and contributes to the censorship-resistance of the network.
+
+### Use cases
+
+A client can expose to the user a set of premade rule templates (the user doesn't have to see the rules) for common relay policies, for example:
```
jb55, line 35: I'm not sure what damus would do with these... I should trust my friends to make filtering decisions for me or something?

jb55:
On Sat, Aug 13, 2022 at 08:20:18PM -0700, Ricardo Arturo Cabral Mejía wrote:
>> I see this nip as benign but don't see the benefit yet.
>
***@***.***  There's currently no NIP specifying how a client should store a list of relays. Branle stores the list of relays in the `content` field of kind 3 events but that's not covered in NIP-02. This NIP fixes that.

I see the benefit of a spec as a standardization of nip-3 content but I'm not exactly sure what I would do with the rules as a client.

23.md
Giszmo: @Cameri `endsWith` because proof of work in event IDs makes the start less attractive to use as a short identifier.

23.md
Giszmo: Can we delete this phrase then?

Cameri:
> On Sat, Aug 13, 2022 at 08:20:18PM -0700, Ricardo Arturo Cabral Mejía wrote:
> >> I see this nip as benign but don't see the benefit yet.
> >
> ***@***.***  There's currently no NIP specifying how a client should store a list of relays. Branle stores the list of relays in the `content` field of kind 3 events but that's not covered in NIP-02. This NIP fixes that.
> 
> I see the benefit of a spec as a standardization of nip-3 content but I'm not exactly sure what I would do with the rules as a client.
> 

Instead of clients blasting events to all relays the write rule lets users choose which relays to send to per event basis. Instead of subscribing to all relays with the same filters, the read rule lets users choose which relays can be subscribed to (the filter must match the read rule).

jb55:
couldn't we just use the `r` tag for this? It seems like a similar usecase. Thinking about single-char economy here.

23.md
Cameri: should we mention here that the underscore `_` is an allowed identifier? It's needed for `created_at`

23.md
Cameri: I think we should remove this. If a client implements this NIP it must handle rules properly.

Cameri (approved):
LGTM. Would recommend removing the paragraph @Giszmo pointed out.

Cameri:
What's missing here to move forward?

fiatjaf:
Check if these rules make sense from the client point-of-view or determining what rules make sense.

I think the current form is not great. Maybe we could just specify a JSON object of relay URLs with an object for each like

```
{
  "wss://relay.nosotros.com": {"read": true, "write": true},
  "wss://personal-relay.x.com": {"read": false, "write": true, "archive-everything-i-like": true}
}
```

and then clients interpret that in no strict manner.

23.md
dskvr: @fiatjaf bumping this

dskvr:
> Maybe we could just specify an arbitrary JSON object of relay URLs with an object for each like

stringified JSON in `content`?

fiatjaf:
Yes. Very bad?

dskvr:
Nope, just wanted clarification.

KotlinGeekDev:
Please, @fiatjaf would it not better to make it a JSON array of  relays? Something like [ { "url": " wss://... ", "read": true, "write": true}, { "url": " wss://...", ...}, ...]
I think it would make it easier to parse for clients.

fiatjaf:
@KotlinGeekDev that is true, but then you would have to handle duplicates in your code. What do you prefer?

dskvr:
@KotlinGeekDev 

Using JS as an example
```
let relays = JSON.parse(event.content)
for(relay in relays) {
...
}
```
vs
```
let content = JSON.parse(event.content)
let uniques = new Set(content.relays)
uniques.forEach( relay => {
...
})
```

Not much different, except that not every language has easy uniques from `Array`, and understanding that the result may not be only uniques is something that many people will forget to implement.

Objects have unique keys by default.

Doesn't completely erradicate the problem, however, anomalies such as trailing slashes will still require sanitization. Which brings up another point, it needs to be established whether or not relays should contain or not contain a trailing slash.

KotlinGeekDev:
@fiatjaf I already handle duplicates in code when it comes to obtaining notes from several relays. So it is not a big deal.

KotlinGeekDev:
But @dskvr raises some interesting points here as well.:thinking:

dskvr:
@KotlinGeekDev Made some logical errors, updated. Unlike an object, it would be difficult to easily/efficiently dedupe an array of objects. You would either need to filter the array on each iteration or iterate the array to dedupe.

dskvr:
>  I already handle duplicates in code when it comes to obtaining notes from several relays. So it is not a big deal.

I think we're more concerned what it means at scale, on a case by case basis anything is solvable.

dskvr:
On second look some of the code examples I gave were grossly incorrect (probably still are, they are untested), but they're much closer correct now. Deduping an array of objects is objectively more cumbersome. Only real benefit is that well-formed objects provide support for programmatic schemata (`json-schema`, `protobuf`, etc)

KotlinGeekDev:
Right @dskvr , I see your point. Though, it will be complicated to implement that parsing logic.:sweat_smile:

dskvr:
@fiatjaf As modified this is basically `kind:3` but in the replaceable-kind space. If you like your proposed, it drastically simplifies your proposal. Coracle appears to be using `kind:3` to track user relays, though, doing so is quite messy as it's really a user state variable, not a recommendation. I would love to see some traction made here and for this replaceabl-kind to see the light of day. Let me know if there's anything I can/cannot do to help.

fiatjaf:
I am also interested in seeing some traction here. I am not sure how to proceed since I am not implementing this myself right now, so maybe we should again ask for the input of the clients who are currently using these other approaches and what would be best for them.

staab:
I think the array syntax as proposed is fine, a dictionary with a list of policies would also work. I'm not sure the use cases hold up though, they seem to imply that notes are distributed across relays based on use case, rather than sphere of contacts. I think we're ultimately going to end up with a network topology divided by social graph (like mastodon) rather than by function (microservices).

For that reason, the DM use cases don't seem to make sense to me — if you're limiting the relays you send DMs to without reference to which relays the person you're trying to contact is connected to, you're making deliverability worse. Clients should be smart enough to send DMs only to relays the recipient is known to participate in instead.

For the "Bob's/my personal relay" use cases, I would expect that to be implemented on the relay's end instead. Event publishing is not very high-volume, a personal relay can simply drop anything its not interested in (and it'll have to anyway to reject spam, unauthorized pubkeys, etc).

If a relay is full of spambots, I can't think of why you would want to ignore replies but not top-level notes without an additional filter based on web of trust.

So I'm not convinced by any of the filter use cases, much more useful would be to specify an index generated by relays to help clients make smarter social graph decisions: follow/follower/reaction/reply counts, and intersection of pubkeys/relays. I intend to write a NIP for that, but haven't had an opportunity yet.

staab:
After a little more thought, I do think this NIP could be useful as a hint to _other_ users' clients on how to interact with the user publishing their list. IOW advertising a specific relay as their onbox/outbox could be useful as an optimization so clients aren't just spraying and praying.

And I do think this is an improvement over hijacking kind 3.

1440000bytes:
Concept ACK

1440000bytes:
> Kind 2 should be deprecated

why?

dskvr:
> I vote we split it out into a separate PR.

I also exchanged this thought with @fiatjaf in my RFC.

eskema:
> I was just going to implement some smarter relay traversal in Coracle and came across the problem this NIP solves, so I'm looking forward to it being merged. I think we should do some cleanup too, since there are some flaws in how relays are recommended currently (see my comment [here](https://github.com/nostr-protocol/nips/issues/121#issuecomment-1370505023)):
> 
> * Kind 2 should be deprecated
> * NIP 2 should be modified to specify that kind 3 should include duplicate tags based on the target pubkey's lates relay list. Or some other fix, but a single relay url seems to be a weak point here.
> 
> If the filters part of this NIP is still controversial (and remains un-implemented), I vote we split it out into a separate PR.

kind-2 is recommend a relay. it's just a note with a recommendation. why would you deprecate it? how is it not working as intended? kind-3 is your contact list, why would you have duplicates in there?

mikedilger:
I would rather this be a serialized JSON object in the content as suggested above.

But if it stays as tags, I would very much like those tags to have a new tagname first, rather than starting with a relay URL, because it really screws up my generic tag serialization/deserialization if it doesn't have a known tag name to code branch from.

dskvr:
I've submitted a PR that modifies NIP-23. These revisions simplify NIP-23 to look more like a migration from `kind 2` to `kind 10001`. It encourages and supports backwards compatible extension by default, such as advanced filter rules; which should absolutely be explored. 

Ideally, historical participants would give this a look

blakejakopovic:
Sorry, travelling so can’t ingest all this PRs content. 

I prefer JSON in content. I know it’s up to clients, however ideally clients allow a local relays list and a publish/sync relays list (I.e. create a new 10001 kind event)

Justification is not all relays you use in a client should be published/broadcast as the ones you use/want in all clients. A good example is dev relays. Another example is app specific relays I may not want as my defaults for all apps (a timeline like app, vs corporate Nostr, vs private group app, etc. 

Basically, clients should support a synced relay list and a local relay list. The unique set is then what gets connected to in apps. Local can be persisted in iCloud or whatever, but just not as Nostr events.

KJsRepo:
I really like the future-proofing this would allow.  Programmability in how relays are used will open up countless possibilities.

That said, @blakejakopovic is right in that private relays are absolutely necessary.  If we are discussing formalizing relay lists, it only makes sense to see if encryption can play a part in this.

That said, relay lists and *secure* relay lists are two different problems.

1:  The find-me beacon (which kind 3 / NIP 2 is currently taking the place of) is required for proper operation of the network, this needs to be quick and easy to process since all sorts of relays and clients will be using it constantly.

2:  Storing relays for ease of use / transportability.  This can be slow and a bit bloated, as only one user will be pulling it once in a while.  This should allow users to securely store their relays.  I would much prefer JSON in content, as Blake stated, for the same reason.  A relay "object" contains the relay, the filter rules, an "encrypted" flag, and read/write permissions (perhaps we can attach most of these to the group).

So, with all of this in mind, perhaps a hybrid of each approach is worthy of discussion.  Kind 3 data can be moved to the tags to satisfy the need for efficiency and openness, and the relay lists can move to JSON content, allowing clients a bit of leeway to do cool things with non-protocol-critical lists (and above all, **encrypted and secure** lists).

monlovesmango:
leaving my comments as it was requested of me a couple weeks ago (sorry for delay). I like this NIP and am glad its still being discussed. I read it a while ago and didn't have any concerns/questions.

concept ack

mikedilger:
Is this NIP happening?  I hope so.

If people spread their relay lists far and wide, gossip will use that to find them instead of the massive mess of data it currently tries to collect and rank.

IMHO people should spread their relay list widely so they can be easily found (1 replaceable event per relay), but put their normal posts on just a few (maybe even just 1) relays.

KJsRepo:
I have a full implementation (runes rules and kind 10001) in nostr-tools that's working in unit-testing, but the integration with the client I was testing with is finicky.  I may just PR what I have for nostr-tools this weekend and hope for the best.

Cameri:
> I have a full implementation (runes rules and kind 10001) in nostr-tools that's working in unit-testing, but the integration with the client I was testing with is finicky.  I may just push what I have for nostr-tools this weekend and hope for the best.

Runes are also implemented on Nostream. The only difference is I allow underscores to support the created_at field.

fiatjaf:
Closing this on behalf of https://github.com/nostr-protocol/nips/pull/218.

PR #34, opened 2022-08-14
added examples

monlovesmango:
added a few small changes and a couple examples (mainly to validate my understanding)

### DIFF
--- a/26.md
+++ b/26.md
@@ -50,7 +50,7 @@ When a `read` rule evaluates to `true` for a given **filter** the client SHOULD 
       - `pubkey=e3e3`: `true`
       - `kind=7&#p=6677`: `true`
       - `created_at>999999999|#e=5a5a`: `false`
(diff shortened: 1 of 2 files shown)


26.md
```diff
@@ -38,6 +38,7 @@ When a `read` rule evaluates to `true` for a given **filter** the client SHOULD
       - `kinds=1|kinds=4`: `true`
       - `kinds<2`: `true`
       - `kinds>7`: `false`
+      - `kinds=1|kinds=7&authors=8543|authors=1234`: `true`
```
Giszmo, line 41: This is an important example to explain the `&` behaviour.

Giszmo:
ACK if @fiatjaf confirms the rule I commented on above is correct. In rune, `&` is the separator in a list of rules that all have to evaluate to true. It's not the logical `&&` that would have precedence in Java for example.

fiatjaf:
I think I can confirm that.

PR #35, opened 2022-08-14
a warning for future nip designers

Giszmo:
https://github.com/Giszmo/nips/blob/amendNip1/01.md


fiatjaf:
I agree with this but it is a little too opinionated and harsh, maybe it would be a better fit in a standalone article with your name, not in the canonical and sacred home of the higher NIPs.

Giszmo:
Well, I wasted enough time on this and will point to this PR - merged or not - if others want to embark on this endeavour of making events smaller.

fiatjaf:
Oh, you've wasted your time trying to think about how to make events smaller?

I think we should push some more written explanations for these things not very related to the NIPs to the github.com/nostr-protocol/nostr repository. Maybe this could be a document there called "warning for future nip designers".

jb55:
hmm it's not super clear what this is suggesting. is it a warning to not work on compressed formats?

Giszmo:
@jb55 yes, kind of. It is as fiatjaf said "opinionated" and saying that space optimizations are plenty possible but so far were discarded as premature optimization in favour of remaining accessible for developers to work with easily.

Giszmo:
I withdraw this PR for not being really actionable.

PR #38, opened 2022-08-24
NIP-28: Public Chat

AtlanteanOne:
Let's also disrupt Telegram, FB Messenger, WhatsApp, WeChat, et al.

[Link to NIP-28](https://github.com/nostr-protocol/nips/blob/af6893145f9a4a63be3d90beffbcfd4d90e872ae/28.md)

### DIFF
--- a/28.md
+++ b/28.md
@@ -50,7 +50,7 @@ Clients MAY add additional metadata fields.
 
 ```json
 {
-    "content": <string>,
+    "content": "{\"name\": \"Updated Demo Channel\", \"about\": \"Updating a test channel.\", \"picture\": \"https://placekitten.com/201/201\"}",,
     "tags": ["e", <channel_create_event_id>],
     ...
 }
--- a/28.md
+++ b/28.md
@@ -9,16 +9,17 @@ Public Chat
 
 This NIP defines new event kinds for public chat channels, channel messages, and basic client-side moderation.
 
-It reserves new four event kinds (40-43) for immediate use and six event kinds (44-49) for future use.
+It reserves five event kinds (40-44) for immediate use and five event kinds (45-49) for future use.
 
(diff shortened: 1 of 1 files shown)


fiatjaf:
I would say bring back channel metadata to kind 41 since it doesn't fit the concept of replaceable event as defined by NIP-16 anyway. Relays will just have to add a custom handler for kind 41 events just like they do already for kind 0 and kind 3.

Aside from that all looks great.

28.md
```diff
@@ -0,0 +1,128 @@
...
+}
+```
+
+## Kind 43: Hide message
+
+User no longer wants to see a certain message.
+
+The `content` may optionally include metadata such as a `reason`.
+
+Clients SHOULD hide event 42s shown to a given user, if there is an event 43 from that user matching the event 42 `id`.
```
jb55, line 79: this could work for kind1 messages as well, it's something I've wanted for hiding replies to posts in damus.

28.md
```diff
@@ -0,0 +1,128 @@
...
+}
+```
+
+## Kind 44: Mute user
+
+User no longer wants to see messages from another user.
+
+The `content` may optionally include metadata such as a `reason`.
+
+Clients SHOULD hide event 42s shown to a given user, if there is an event 44 from that user matching the event 42 `pubkey`.
```
jb55, line 99: same for this one

28.md
vishalxl: This could be done by a special reaction, like '!' or something like that, which if emitted should signify that the comment replied to should not be shown to the given user. Its a sort of 'hide' indicator.

Cameri (approved):

fiatjaf:
Can someone update the README with the new kinds and a link to the NIP?

PR #39, opened 2022-08-24
NIP-22: add note about lower limit

jeffthibault:
For consideration about how a lower limit would effect user's ability to migrate events.

### DIFF
--- a/22.md
+++ b/22.md
@@ -40,3 +40,5 @@ if event.created_at not in range(LOWER_LIMIT, UPPER_LIMIT):
   # NOTE: This is one example of a notice message. Relays can change this to notify clients however they like.
   ws.send('["NOTICE", "The event created_at field is out of the acceptable range (-24h, +15min) for this relay and was not stored."]')
 ```
+
+**Note**: A relay that enforces a small lower limit will prevent users from migrating their old events to that relay, which is bad because the ability to migrate events is a great feature of Nostr. A recommendation would be to use the unix timestamp for 2020-11-08, the date the [commit](https://github.com/nostr-protocol/nostr/commit/bdeb03aeaf2d44db954e0eb1356a3f585ee7c50d) containing the Nostr readme was published, because no Nostr events existed before then.


fiatjaf:
But I want to publish things I wrote many years ago to Nostr!

jeffthibault:
Then you will not be publishing to relays that use NIP-22!

fiatjaf:
I thought NIP-22 was for relays to not allow anything from 1 day ago. What is this recommendation for the Nostr genesis timestamp about then? I'm confused.

jeffthibault:
I just used 1 day in the example. The NIP says a relay can choose any date range within which they will allow events.

jeffthibault:
The relay should send a notice when they don't accept an event though

fiatjaf:
In that case this NIP is completely useless, counterproductive and confusing, since relays already could choose any date range without it. I thought this NIP specified a fixed date range.

jeffthibault:
Yes, you are right. Although it would be good if relays enforcing a date range send a NOTICE that specifically says they rejected an event because it is not in the date range.

I will close this PR. Feel free to nuke NIP-22.

fiatjaf:
To be honest I didn't read NIP-22 carefully or fully, and there were a bunch of people discussing about it and liking it, so I thought it had a purpose. Do you really think it is that bad?

jeffthibault:
Well my reasoning for NIP-22 was to define a way for relays to restrict what events they store based on when they were created. For example, maybe a relay operator does not want to run an "archival" (for lack of a better term) relay, so they would set a lower limit to reject old events and would probably also delete stored events after an amount of time passed. Also, a relay operator running a relay today might not want to store events from the year 3000 (or any date in the future) so they would set an upper limit. Whatever the case, the NIP defines a way for clients to find out if relays use date limits and also defines how relays should notify clients about the limits.

In my mind, this seems reasonable to have a NIP for but maybe without defining a specific range, it is too general and therefore useless.

fiatjaf:
Well, I do not understand, but let's leave it there anyway.

PR #40, opened 2022-08-25
Add specific replaceable events to nip16

Kukks:
Adds specific replaceable events at a new range of kinds, where now it takes additional criteria to match and replace: the first tag and its values.

One use-case for this is if you are posting replaceable notes and need to anticipate a greater amount of notes to the allowed kind range for a single author.

### DIFF
--- a/16.md
+++ b/16.md
@@ -11,12 +11,17 @@ Relays may decide to allow replaceable and/or ephemeral events.
 Replaceable Events
 ------------------
 A *replaceable event* is defined as an event with a kind `10000 <= n < 20000`.
-Upon a replaceable event with a newer timestamp than the currently known latest replaceable event with the same kind, the old event SHOULD be discarded and replaced with the newer event.
+Upon a replaceable event with a newer timestamp than the currently known latest replaceable event with the same kind and author, the old event SHOULD be discarded and replaced with the newer event.
 
 Ephemeral Events
 ----------------
 An *ephemeral event* is defined as an event with a kind `20000 <= n < 30000`.
 Upon an ephemeral event being received, the relay SHOULD send it to all clients with a matching filter, and MUST NOT store it.
+
+Specifically Replaceable Events
+------------------
+A *specific replaceable event* is defined as an event with a kind `30000 <= n < 40000`.
+Upon a replaceable event with a newer timestamp than the currently known latest replaceable event with the same kind, author, and the first tag (and its values), the old event SHOULD be discarded and replaced with the newer event.
 
 Client Behavior
 ---------------
--- a/16.md
+++ b/16.md
@@ -18,9 +18,9 @@ Ephemeral Events
 An *ephemeral event* is defined as an event with a kind `20000 <= n < 30000`.
 Upon an ephemeral event being received, the relay SHOULD send it to all clients with a matching filter, and MUST NOT store it.
 
-Specifically Replaceable Events
+Parameterized Replaceable Events
 ------------------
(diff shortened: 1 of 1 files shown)


16.md
```diff
@@ -11,13 +11,18 @@ Relays may decide to allow replaceable and/or ephemeral events.
...
 
 Ephemeral Events
 ----------------
 An *ephemeral event* is defined as an event with a kind `20000 <= n < 30000`.
 Upon an ephemeral event being received, the relay SHOULD send it to all clients with a matching filter, and MUST NOT store it.
 
+Specifically Replaceable Events
+------------------
+A *specific replaceable event* is defined as an event with a kind `30000 <= n < 40000`.
+Upon a replaceable event with a newer timestamp than the currently known latest replaceable event with the same kind, author, and the first tag (and its values), the old event SHOULD be discarded and replaced with the newer event.
```
fiatjaf, line 24: Should it say "its value" instead as everywhere we only deal with the first value of tags, i.e. the second element in the array?

Cameri: Agree, but I wouldn't make this match the first tag, I would mark the tag similarly to how we do it in nip 10 or use a tag called `deduplication`. Clients have had problems in the past because of ordering of tags had a meaning but was not consistent.

Giszmo: I agree with disregarding anything beyond the second element of a tag.

@Cameri the "lets use the first tag" was born from the idea to use tags that already have a meaning otherwise. In the chat it was about e-tags to define properties of group chats with that e-tag ID. I agree with you that it would be desirable to use something else. `unique` over `deduplication` though.

16.md
```diff
@@ -11,13 +11,18 @@ Relays may decide to allow replaceable and/or ephemeral events.
...
+Upon a replaceable event with a newer timestamp than the currently known latest replaceable event with the same kind and author, the old event SHOULD be discarded and replaced with the newer event.
 
 Ephemeral Events
 ----------------
 An *ephemeral event* is defined as an event with a kind `20000 <= n < 30000`.
 Upon an ephemeral event being received, the relay SHOULD send it to all clients with a matching filter, and MUST NOT store it.
 
+Specifically Replaceable Events
+------------------
+A *specific replaceable event* is defined as an event with a kind `30000 <= n < 40000`.
```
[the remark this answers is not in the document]
Cameri: parameterized*

Cameri (approved):
LGTM

16.md
```diff
@@ -11,13 +11,18 @@ Relays may decide to allow replaceable and/or ephemeral events.
 Replaceable Events
 ------------------
 A *replaceable event* is defined as an event with a kind `10000 <= n < 20000`.
-Upon a replaceable event with a newer timestamp than the currently known latest replaceable event with the same kind, the old event SHOULD be discarded and replaced with the newer event.
+Upon a replaceable event with a newer timestamp than the currently known latest replaceable event with the same kind and author, the old event SHOULD be discarded and replaced with the newer event.
```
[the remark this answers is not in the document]
Giszmo: I wouldn't mind adding this although it's tangential to the PR.

Giszmo:
Kukks has other stuff on his mind. Cameri is in a hurry. I want this feature eventually and would implement it in NostrPostrRelay but I agree with Semisol that it deserves a different nip as relay feature support signalling would break by slapping it on nip16.

@fiatjaf which number? I could create the PR.

Giszmo:
What? (looking at @Cameri )

Cameri:
> What? (looking at @Cameri )

https://github.com/nostr-protocol/nips/pull/54

PR #41, opened 2022-08-28
NostrAuth easily readable UIN address derived from pubkey 

saranshisatgit:
User identification number , inspired from Bitauth and SIN.

### DIFF
--- a/26.md
+++ b/26.md
@@ -1,5 +1,5 @@
 
-NIP-26
+NIP-27
 ======
 
 NostrAuth
@@ -7,7 +7,7 @@ NostrAuth
 
 `draft` `optional` `author:saranshisatgit`
 
-The current approach for signed events is good but problematic on user side. Proposing a signed event with a `UIN` (User Identification Number). These UIN 
+The current approach for signed events is good. Proposing a new signed event with a `UIN` (User Identification Number). These UIN 
 can be stored in public database somewhere like namecoin, spacechains, where users can claim unique names. These UIN can be fetched by the client and relays on demand and could be useful
 in determinng certain attributes say providing specific service to the user or a role or permissions.
 
@@ -16,24 +16,57 @@ This makes identification easier , user preference can be thus stored retrieved 
 My motivation is to make the user namespace and identification easier across clients and relays so that we can build things like blogging, relationships. This would be a good starting point for the reputation.
 
 Generating UIN could be done like this, since keypair is already generated we can use the `NIP-06`  or as client needs.
+
 **Generate**
 
 ```
 nostrAuth.generateUIN()
 ```
 
+The follwing function that will return a UIN from the public key;
+
+```
+function(pubkey) {
+  let pubkeyBuffer;
+  if (!Buffer.isBuffer(pubkey)) {
+    pubkeyBuffer = Buffer.from(pubkey, 'hex');
+  } else {
+    pubkeyBuffer = pubkey;
+  }
+
+  // sha256 hash the pubkey
+  const pubHash = crypto.createHash('sha256').update(pubkeyBuffer).digest();
+
+  // get the ripemd160 hash of the pubkey
+  const pubRipe = crypto.createHash('rmd160').update(pubHash).digest();
+
+  // add the version
+  const pubPrefixed = Buffer.concat([Buffer.from('0f02', 'hex');, pubRipe]);
+
+  // two rounds of hashing to generate the checksum
+  const hash1 = crypto.createHash('sha256').update(pubPrefixed).digest();
+  const checksumTotal = crypto.createHash('sha256').update(hash1).digest();
+
+  // slice the hash to arrive at the checksum
+  const checksum = checksumTotal.slice(0, 4);
+
+  // add the checksum to the ripemd160 pubkey
+  const pubWithChecksum = Buffer.concat([pubPrefixed, checksum]);
+
+  // encode into base58 to get SIN
+  return bs58.encode(pubWithChecksum);
+};
+```
+
 This will produce  
 
 ```
 {
-  created: 1661671608,
-  priv: '5e7d9dde9c2df88152c7e7a7791d416079e6d72d8de48cb6688e5f00b88b1077',
-  pub: '066185612bc5b79079ffa8acc86bbd26a24ea811161740bd70e7f39e400a9d06',
   uin: 'TfGeAbsDxTKB9VEHX9X6EpMXbEB25Jh27Nj'
 }
 ```
 
-If a user choose to generate UIN and does not want to register the UIN publically , then we can derive the UIN from PublicKey also.
+If a user choose to generate UIN and does not want to register the UIN publically , then we can derive the UIN from PublicKey also as we demonstrated.
 
 
 Example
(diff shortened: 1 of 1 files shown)


fiatjaf:
I don't understand what is problematic about using public keys directly and what is the point of this UIN. I am probably missing something, but it looks you are replacing everything that is related to Nostr with new things and making a cryptographic protocol from scratch completely unrelated to Nostr.

saranshisatgit:
No, not replacing, just adding. UIN is just a user identification number;

1. Unique namespaces that could be UIN can be stored on public DB like namecoin or other, associated to keypair.
2. Auth using these UIN if they exist and probably an open plethora of use cases.

Bitauth is using Bitcoin identity protocol for Authentication without password.

References are here https://en.bitcoin.it/wiki/Identity_protocol_v1.

26.md
```diff
@@ -0,0 +1,57 @@
+
+NIP-26
+======
+
+NostrAuth
+---------
+
+`draft` `optional` `author:saranshisatgit`
+
+The current approach for signed events is good but problematic on user side. Proposing a signed event with a `UIN` (User Identification Number). These UIN
```
Cameri, line 10: I disagree that the current approach for signing events is problematic. On the contrary, it's very simple to use Schnorr signatures for creating identities.
There's many things left to interpretation because this NIP glosses over so many details.
This NIP pushes trust out of Nostr and to a third party, coupling the protocol to some blockchain which is wholly undesirable.
To me it sounds like you are trying to build a different protocol which isn't Nostr.

26.md
```diff
@@ -0,0 +1,57 @@
+
+NIP-26
```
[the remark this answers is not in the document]
saranshisatgit: ok

26.md
saranshisatgit: The current approach is suitable but unsuitable for building use cases like reputation and relationships. This could be achieved if we use UIN.

This is not at all third party it uses an existing solution that is schnorr , it does not require any coupling with blockchain, user can store it on their website or any where as long its publically accessible

fiatjaf:
I still have no clue at all of what is being proposed here.

saranshisatgit:
> I still have no clue at all of what is being proposed here.

I have made some changes , probably that will help

26.md
```diff
@@ -1,13 +1,13 @@
 
-NIP-26
+NIP-27
```
Cameri, line 2: Please take a look at the [list of PRs](https://github.com/nostr-protocol/nips/pulls), this one is taken too.

saranshisatgit:
> I don't understand what is problematic about using public keys directly and what is the point of this UIN. I am probably missing something, but it looks you are replacing everything that is related to Nostr with new things and making a cryptographic protocol from scratch completely unrelated to Nostr.
> 
> What are BitAuth and SIN? Maybe provide some links to explanations because it is impossible to search for "SIN" and from the BitAuth supposedly-official webpages I couldn't get any explanation about how it works.
https://github.com/bitpay/bitauth

jb55:
If you want to associate extra key information with a nostr pubkey you could just put it in the user profile (kind-0). having a generic format for tags that could be displayed on a profile (like in mastodon) would be interesting. could be used for gpg keys, website, etc. I assume this is all you really want to do?

saranshisatgit:
> If you want to associate extra key information with a nostr pubkey you could just put it in the user profile (kind-0). having a generic format for tags that could be displayed on a profile (like in mastodon) would be interesting. could be used for gpg keys, website, etc. I assume this is all you really want to do?

, the UIN is more like address derived from pubkey , like bitcoin address . The association is close but since it’s derived out of the pubkey it would be more suitable if user could just simply relay them and probably sign data using the UIN address type so that one knows that type x address is sending information. Other use cases can be built like identity reputations from here , since users can publish these address anywhere .

third party may choose to attest these UIN.

my main motivation is to work out a better address system , where they are easily readable

eskema:
sorry for my lack of understanding, but why not use the pubkey directly instead of this UIN? what properties does the UIN have that a pubkey doesn't? Can you create more than one per pubkey?

Cameri:
> , the UIN is more like address derived from pubkey , like bitcoin address . The association is close but since it’s derived out of the pubkey it would be more suitable if user could just simply relay them and probably sign data using the UIN address type so that one knows that type x address is sending information. Other use cases can be built like identity reputations from here , since users can publish these address anywhere .

a pubkey is just a 64 byte number IIRC. so is it a matter of representing it in a bitcoin address format that you need?

fiatjaf:
> my main motivation is to work out a better address system , where they are easily readable

I want that too, but I don't see the point of this UIN thing. Can't you get your Namecoin name to point to your Nostr key directly?

saranshisatgit:
> sorry for my lack of understanding, but why not use the pubkey directly instead of this UIN? what properties does the UIN have that a pubkey doesn't? Can you create more than one per pubkey?

Sorry for not presenting idea properly some key things were left out from the proposal, 1) Better representation of the key, you can generate one single unique key out of the pubkey, 2) If you want the UIN to be persistent then one may use Lightning or sats to pay some fees. 

UIN properties are (referenced from Bitcoin Identity v1 Protocol)

```
Ownership may be digitally proven
Attach sequence of key-value pairs (public proof) and hashes (private proof) to your SIN record.
A merkle root exists in each record, for even more private proofs.
Start as anonymous; opt out of anonymity by attaching identifying key-value pairs (real.name = "John Smith", gov.us.ssn = "123-45-6789").
Disposable
All key-value pair updates digitally signed by SIN owner (key holder)
Third parties may offer digital attestions:
Identity Verification, Inc. digitally signs a SIN as passing their Not A Criminal/Level-1 check.
Big Auction Provider, Inc. digitally signs a SIN as having a certain reputation score, on their website.
Decentralized market users digitally sign one another's SINs, building a decentralized reputation
Type-1 UINs: have some creation cost, deterring spam.
Type-1 UINs: Sacrifice may be digitally proven, bootstrapping root of trust from blockchain data
```

eskema:
ok, I may still misunderstand some concepts, but in the case of nostr, you already have an identity tied to you, and you can prove you control it by posting using a pubkey / privkey. It's different from bitcoin because in bitcoin, addresses are not identities, they're disposable recipients not tied to any other identity. In that case, a UIN makes sense. In nostr, the pubkey is already an identity so all this makes no sense here.

saranshisatgit:
Agreed , pubkey is a form of identity, but same with nostr pubkey they are not tied to any other identity also  , since I can generate thousands of pubkey and sign events from them and yet claiming that it’s not mine. But if Make  it expensive (question is do we really need at the nostr end ) to acquire UIN for these pubkey that  work in eliminating spam and allows user recognition with reputation that makes sense. 

Some events say like , claiming a statement X is true from a well known UIN versus a pubkey publishing the same event ! 

Maybe 🤔, I need to rethink the whole thing again!

Cameri (changes requested):
I don't agree this should be part of the protocol. For starters the PR is too hand wavy so it isn't clear how anything really works. You can already add your unique external identifier (UIN) to events as a tag.  Expecting relays and clients to verify these third-party identities in an open set of namespaces (namecoins, space chains, oauth) is wishful thinking.

PR #44, opened 2022-08-31
NIP-31 Pinned Events

plantimals:
NIP-31 describes pinned events. It wasn't entirely clear whether to propose in-place edits to NIP-01 or create new NIP.

Put a field in your `kind 0` metadata content to indicate which event you want to pin. 

[This event](https://nostr.io/api/v1/events/71412d5cdb7e675b9806d23bda3d7c4878ca7f8e72ae3477c7179938e857cb9b) is an example of a `kind 0` metadata event with the `pinned_event` field specified. An implementation of a client that interprets the pinned event as an event that should appear at the top of a user's profile page, before other more recent events can be found [here](https://nostr.io/p/dd81a8bacbab0b5c3007d1672fb8301383b4e9583d431835985057223eb298a5).

### DIFF
--- a/31.md
+++ b/31.md
@@ -1,0 +1,28 @@
+NIP-31
+======
+
+Pinned Events
+-------------
+
+`draft` `optional` `author:plantimals`
+
+NIP-31 amends the content metadata for `Kind 0` as described in [NIP-01](01.md).
+
+Each pubkey MAY indicate a pinned event by setting the field `"pinned_event"`
+in the json content of one's `Kind 0` metadata event to point to an
+event `ID`.
+
+Example json content from [this event](https://nostr.io/api/v1/events/71412d5cdb7e675b9806d23bda3d7c4878ca7f8e72ae3477c7179938e857cb9b):
+
+```json
+{
+    "name":"plantimals",
+    "picture":"https://plantimals.org/img/avatar.png","pinned_event":"c76166d9384355e8f55abfc071294fb815ac5f4050e2b9373b997b4bed66c669",
+    "about":"[plantimals](https://plantimals.org)",
+    "nip05":"_@plantimals.org"
+}
+```
+
+The `pinned_event` field is meant to provide a link to additional content for
+someone wishing to supplement their `about` text. Users SHOULD ONLY pin events
+that they have authored in order to avoid mistaken identity.
--- a/README.md
+++ b/README.md
@@ -20,6 +20,7 @@ NIPs stand for **Nostr Implementation Possibilities**. They exist to document wh
 - [NIP-16: Event Treatment](16.md)
 - [NIP-22: Event created_at Limits](22.md)
 - [NIP-25: Reactions](25.md)
+- [NIP-31: Pinned Events](31.md)
 
 ## Event Kinds
 


31.md
```diff
@@ -0,0 +1,28 @@
...
+Each pubkey MAY indicate a pinned event by setting the field `"pinned_event"`
+in the json content of one's `Kind 0` metadata event to point to an
+event `ID`.
+
+Example json content from [this event](https://nostr.io/api/v1/events/71412d5cdb7e675b9806d23bda3d7c4878ca7f8e72ae3477c7179938e857cb9b):
+
+```json
+{
+    "name":"plantimals",
+    "picture":"https://plantimals.org/img/avatar.png","pinned_event":"c76166d9384355e8f55abfc071294fb815ac5f4050e2b9373b997b4bed66c669",
```
Giszmo, line 20:
```suggestion
    "picture":"https://plantimals.org/img/avatar.png",
    "pinned_event":"c76166d9384355e8f55abfc071294fb815ac5f4050e2b9373b997b4bed66c669",
```

31.md
```diff
@@ -0,0 +1,28 @@
...
+
+`draft` `optional` `author:plantimals`
+
+NIP-31 amends the content metadata for `Kind 0` as described in [NIP-01](01.md).
+
+Each pubkey MAY indicate a pinned event by setting the field `"pinned_event"`
+in the json content of one's `Kind 0` metadata event to point to an
+event `ID`.
+
+Example json content from [this event](https://nostr.io/api/v1/events/71412d5cdb7e675b9806d23bda3d7c4878ca7f8e72ae3477c7179938e857cb9b):
```
Giszmo, line 15: The link doesn't work right now.

```suggestion
Example json content:
```

31.md
```diff
@@ -0,0 +1,28 @@
...
+    "name":"plantimals",
+    "picture":"https://plantimals.org/img/avatar.png","pinned_event":"c76166d9384355e8f55abfc071294fb815ac5f4050e2b9373b997b4bed66c669",
+    "about":"[plantimals](https://plantimals.org)",
+    "nip05":"_@plantimals.org"
+}
+```
+
+The `pinned_event` field is meant to provide a link to additional content for
+someone wishing to supplement their `about` text. Users SHOULD ONLY pin events
+that they have authored in order to avoid mistaken identity.
```
Giszmo, line 28:
```suggestion
someone wishing to supplement their `about` text. Clients SHOULD ONLY show pinned
events as such if they were authored by the same account as the `kind 0` event.
```

Giszmo (approved):
Concep ACK. Pending implementation.

31.md
fiatjaf: I agree, let's pin stuff from others, why not? Clients are still free to ignore these if that makes things harder for them.

plantimals: I'm not opposed to it. in fact, I added the language based on my incorrect reading of Vishal's comments here: https://github.com/nostr-protocol/nips/pull/42#issuecomment-1229578286

blakejakopovic:
For discussion, are multiple pinned events something to consider? Twitter limits to one, Subreddits two, however three for profiles (unconfirmed) - and I'm unsure about other platforms. 

The most common use-cases I can see are:
- A profile/identity posting a status-like message, announcement, or favourite post of theirs (often 10 days, or even 1-2 years)
- A forum/group post with getting started, mod/group rules, etc (typically long lived)
- A forum/group posting an announcement (like an upcoming or active group event or mega thread usually on recent/active news) that usually has an expiry
- A forum/group pointing to external group places (e.g. discord/telegram/etc)
- Often a pinned post content gets updated (e.g. new rules, or links to more resources or external content). No a problem, but something to consider if using immutable nostr events.

I don't have a strong opinion either way.. just trying to consider known use-cases today and if a single pinned event is too limited by design.

fernandolguevara:
@plantimals can we change to `pinned_events: ["<event-id>"]`, just to allow us to have N pinned events, I don't see the reason to limit only one... clients could choose to show 1 o the entire list

monlovesmango:
hey all, know I am joining the conversation late but wanted to get your thoughts on using a new replaceable event kind to store pinned events instead? this might allow more flexibility like specifying a pinned note for different note kinds. would also allow more space for suggested relays.

31.md
```diff
@@ -0,0 +1,28 @@
...
+======
+
+Pinned Events
+-------------
+
+`draft` `optional` `author:plantimals`
+
+NIP-31 amends the content metadata for `Kind 0` as described in [NIP-01](01.md).
+
+Each pubkey MAY indicate a pinned event by setting the field `"pinned_event"`
```
Cameri, line 11: This should use `e` tags instead with a `pinned` marker, otherwise only 1 pinned event is allowed. Clients can choose to display 1 or more.
e.g.
```
{
  ...
  "tags": [["e","<eventId>", "<relayUrl>", "pinned"]]
  ...
}
```

Cameri:
> hey all, know I am joining the conversation late but wanted to get your thoughts on using a new replaceable event kind to store pinned events instead? this might allow more flexibility like specifying a pinned note for different note kinds. would also allow more space for suggested relays.

Agree. I prefer a separate replaceable event with `e` tags instead of pushing this into `kind 0`

PR #46, opened 2022-09-08
NIP-32: Reputation

ArcadeCityMayor:
This NIP defines a specification for key-based reputation.

Any [NIP-01](https://github.com/nostr-protocol/nips/blob/master/01.md) keypair may assign reputation events to other keys based on user-defined schemas.

(wip - see latest version in [Files changed](https://github.com/nostr-protocol/nips/pull/46/files))

### DIFF
--- a/32.md
+++ b/32.md
@@ -10,3 +10,14 @@ Reputation
 This NIP defines a specification for key-based reputation.
 
 Any [NIP-01](https://github.com/nostr-protocol/nips/blob/master/01.md) keypair may assign reputation events to other keys based on user-defined schemas.
+
+## Kind 32: Assign Reputation
+
+**`tags`** MUST contain an entry identifying the recipient key, in the form `["p", "<pubkey, as a hex string>"]`.
+
+
+Additional Info
+---------------
+
+- [Reputation data-type conversation, Sept 2021](https://github.com/nostr-protocol/nostr/issues/20)
+- [Previous reputation NIP, not adopted](https://github.com/nostr-protocol/nostr/pull/72/files)
--- a/32.md
+++ b/32.md
@@ -8,3 +8,5 @@ Reputation
 `draft` `optional` `author:ChristopherDavid`
 
 This NIP defines a specification for key-based reputation.
+
+Users controlling keypairs may assign reputation events to other keys based on user-defined schemas.
--- a/32.md
+++ b/32.md
@@ -1,0 +1,10 @@
+
+NIP-32
+======
+
+Reputation
+-----------
+
+`draft` `optional` `author:ChristopherDavid`
+
+This NIP defines a specification for key-based reputation.
--- a/32.md
+++ b/32.md
@@ -9,4 +9,4 @@ Reputation
 
 This NIP defines a specification for key-based reputation.
 
-Users controlling keypairs may assign reputation events to other keys based on user-defined schemas.
+Any [NIP-01](https://github.com/nostr-protocol/nips/blob/master/01.md) keypair may assign reputation events to other keys based on user-defined schemas.
--- a/32.md
+++ b/32.md
@@ -15,6 +15,9 @@ Any [NIP-01](https://github.com/nostr-protocol/nips/blob/master/01.md) keypair m
 
 **`tags`** MUST contain an entry identifying the recipient key, in the form `["p", "<pubkey, as a hex string>"]`.
 
+**`tags`** MAY contain entries referencing previous events, in the form `["e", "<event_id, as a hex string>"]`.
+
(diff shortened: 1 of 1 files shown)


pseudozach:
I'll be implementing this approach soon for nostr.directory: https://github.com/nostr-protocol/nostr/issues/20#issuecomment-913027389

Any feedback is appreciated, perhaps we can resurrect and formalize this NIP with datatypes as mentioned in the linked comment.

fiatjaf:
I don't know how you're going to do that, but I'm for it, sounds great.

Experiment with the best ways to format structure, once you reach a point that looks good we can NIP it.

neilck:
Regarding [reputation datatypes](https://github.com/nostr-protocol/nostr/issues/20#issuecomment-913027389).

I think this should be a rating NIP, not a reputation NIP.

If it is one person's opinion about another person at a particular point in time it's a rating. Using the same structure (tags with domain-specific keys and 0-100 integer values), but specifying the pubkey of a restaurant vs a person in "p" tag, or even an event in a "e" tag covers multiple use cases. Rating a person, rating a restaurant, rating a recent post. 

For me, reputation is the current state of how "trustworthy" or "good of a driver", or any other characteristic associated with a person. Reputation can be calculated from ratings from multiple people, or even a single authority if I trust them. This is closer to a Kind 0 metadata event in how a relay should handle it.

As an example of the relationship between rating and reputation: the movie  [Venom](https://www.rottentomatoes.com/m/venom_2018) has a "movie review critic" score of 30% based on 364 professional reviews, but an audience score of 80% based on 25K+ user ratings. I trust the audience score more, because for me personally, the reputation of movie critics is to favour movies that are "high-concept" and serious, which I don't enjoy as much.

earonesty:
i would like a special "reputation type" for "identity verification".  where you are asserting you have personally verified that a public key "is honest about what they claim to be".  obviously you can lie.   but a reputation system is a web-of-trust and liars eventually are ejected from the system

the idea is to prevent centralized identity systems from being invented or needed in the nostr ecosystem,

i'm thinking only 2 things:
 - claims attached to a pubkey (part of profile data)
 - verifications of those claims (signed message from anyone)
 - revocation of claims (signed message from anyone)

example:
 - i claim to be the official pubkey for the libertarian party of nh
 - verification service with a high reputation signs a message confirming that
 - other people can choose to use that service or not
 
probably this should be its own nip?

arkin0x:
Recommended changes:

All "Assign Reputation" events MUST have [NIP-13 proof-of-work](https://github.com/nostr-protocol/nips/blob/master/13.md).

The amount of valid proof-of-work **is the rating**.

The `content` MAY contain a string comment.

The new `rating` tag MUST contain a string identifying ONLY 1 arbitrary category, and it SHOULD be prefixed with a plus (+) or minus (-) to indicate whether the rating is good or bad, respectively. If the +/- is omitted, the rating is assumed to be good.

The category SHOULD be all lowercase and have as few non-alphanumeric characters as possible to eliminate inconsistencies which may break rating aggregation.

For consistency in categorization, I recommend the following specs as well:

If the category is a verb, it MUST be the [bare infinitive of a verb](https://www.grammarly.com/blog/infinitives/).

`drive`, not `driving`

`cook`, not `cooking`

If the category is a noun, it MUST be the singular noun.

If you are rating your friend's vehicle, you use `car` as the category, not `cars`.

If you are rating your friend's footwear, you use `shoe`, not `shoes`.

Other advice for categories:

Avoid abbreviations and acronyms: Labels should be written in their full form to prevent confusion and to maintain consistency. For example, use 'television' instead of 'TV'.

Use common and easily understood terms: Choose labels that are widely recognized and understood to prevent ambiguity. Avoid jargon, slang, or overly technical terms.

Adopt a standard format for compound words: Decide on a consistent format for compound words, such as using hyphens, spaces, or merging the words. For example, choose either 'ice-cream', 'ice cream', or 'icecream' and use it consistently.

Interpreting reputation is left up to clients: web-of-trust, fancy mathematics for weighting, etc.

Idea: Category suggestion feature based on categories already used in the wild, encouraging them to choose from established categories and reducing the likelihood of creating new, inconsistent categories.

Also, AI could be used to aggregate semantically similar categories.

earonesty:
> proof-of-work

perhaps er can use a proof-of-burn lightning address (channel with endpoint of 1btcthisisafakepubkey or something) instead?   ZAPS to this address are unrecoverable, and can be proven to exist and contain the associated reputation.   also it increases the value of everyone's bitcoin every time you score someone, and the cost, even if it's just 1 sat, can be accumulated, just like proof of work.

ArcadeCityMayor:
Closing as superceded by #532

PR #47, opened 2022-09-11
add NIP-28 link and kinds in README

scsibug:
Simply links to the now-merged NIP-28 (Public Chat) in the main README.


scsibug:
Also documents the kinds reserved by NIP-28.

PR #49, opened 2022-09-24
NIP 33: Verifiable Secret Sharing

1440000bytes:
This NIP introduces secret sharing between nostr relays and clients which is required for coinjoin rounds and coordination in [joinstr](https://github.com/1440000bytes/joinstr).

Still in draft mode as its my first NIP and not sure if it has all the required information. Feel free to ask any questions or suggest changes.

### DIFF
--- a/33.md
+++ b/33.md
@@ -1,0 +1,134 @@
+NIP-33
+======
+
+Verifiable Secret Sharing
+---------------
+
+`draft` `optional` `author:1440000bytes`
+
+Relays may decide to allow secret sharing.
+
+Motivation
+----------
+
+This NIP allows relays to share a random secret with some clients that are trying to coordinate a coinjoin round. Coinjoin is the primary usecase, however it can be used for other things as well that require coordination between peers.
+
+Gregory Maxwell had mentioned chaumian blind signatures in a bitcointalk post in 2013: https://bitcointalk.org/index.php?topic=279249
+
+> Using chaum blind signatures: The users connect and provide inputs (and change addresses) and a cryptographically-blinded version of the address they want their private coins to go to; the server signs the tokens and returns them. The users anonymously reconnect, unblind their output addresses, and return them to the server. The server can see that all the outputs were signed by it and so all the outputs had to come from valid participants. Later people reconnect and sign.
+
+This NIP is an alternative to blind signatures and introduces a simple mechanism to verify if the users registering outputs are the ones that registered inputs.
+
+Sequence Diagram
+----------
+
+```mermaid
+sequenceDiagram
+    participant Client
+    participant Relay
+
+    Client->>+Relay: 1. Publish event with kind N and peers M [EVENT]
+    Relay->>+Client: 2. Respond with a random number X and does the same for next M-1 similar events  [NOTICE]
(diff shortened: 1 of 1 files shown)


fiatjaf:
I don't fully get it, but looks good to me.

However I can't help but think: is this really Nostr? Or is it just a separate protocol that wants to reuse Nostr relays? I guess it is fine to reuse Nostr infrastructure like that.

1440000bytes:
> I don't fully get it, but looks good to me -- except maybe instead of using the NOTICE message you could specify a separate message and instead of using normal events use something else (I didn't get if you really need these events).

Using a new message type instead of NOTICE makes sense although I was trying to somehow fit things in to messages already available. Events would be required for step 1 and step 3 else something similar that gets published and everyone connected to relay can read it.

> However I can't help but think: is this really Nostr? Or is it just a separate protocol that wants to reuse Nostr relays? I guess it is fine to reuse Nostr infrastructure like that or it is confusing?

Everything else works fine in joinstr with nostr relays but anyone can read those events and publish their output to steal money from users trying to do coinjoin. It won't work in practice if everyone verifies PSBT before signing, however adds a DoS vector in the coinjoin implementation. Relays should also not be aware of the links between inputs and outputs. So, joinstr is just a proof of concept without this NIP and cannot be used on mainnet.

I am not trying to add too much complexity in nostr protocol but if 2-3 relays implement this NIP it would help joinstr and other similar multi party contracts that want to use joinstr. Development is easy with joinstr compared to other distributed networks and libraries/tooling is helpful even if the protocol is still evolving.

> What if instead of people running "Nostr relays that support NIP 33" they instead ran "Joinstr relays that know nothing about Nostr"?

Building a network of relays from scratch will be difficult and there are few other differences:

|  | nostr relay w/ NIP 33 | joinstr relay |
| :---:         |     :---:      |          :---: |
| Network effect  | :white_check_mark:     | :x:    |
| Censorship Resistance     | HIGH     |  LOW    |
| Tooling     | :white_check_mark:      |  :x:     |
| Community     | :white_check_mark:      |  :x:     |

vishalxl:
In step 2 of the secret diagram, you seem to expect the relay to respond with a 'random number X'.

This will likely never happen in nostr ( or in other words is not desirable), because the relay's can't be depended on to provide a true random number. They are supposed to be trustless relayers of information, and cannot be depended on to provide any critical service or information ( which is not cryptographically secured).

So the participants will need to find some other source of random number.

1440000bytes:
@vishalxl 

> This will likely never happen in nostr ( or in other words is not desirable), because the relay's can't be depended on to provide a true random number. They are supposed to be trustless relayers of information, and cannot be depended on to provide any critical service or information ( which is not cryptographically secured).

- Clients can use 2 or more relays to avoid trusting one relay and this won't affect the process.
- Relays are incentivized to not cheat as they will be caught easily while signing the tx. There won't be any loss of funds and clients will stop using that relay for coinjoin.

> So the participants will need to find some other source of random number.

This will add unnecessary complexity and dependency in joinstr.

@Semisol 

> I don't fully get it but this may be able to be hijacked by a relay from what I see. Can we instead try to use other methods to achieve the same result?

I will think of better solutions and closing this pull request for now.

PR #52, opened 2022-10-13
NIP-24: Private Messages

jeffthibault:
A proposal for more private direct messages. Please provide any feedback.

https://github.com/jeffthibault/nips/blob/private-messages/24.md

### DIFF
--- a/24.md
+++ b/24.md
@@ -25,7 +25,7 @@ Alice and Bob want to send private messages to each other.
 **Step 1:**<br>
 Note: If either Alice or Bob do not know the other's public key, then one of them MUST send a standard NIP-04 message to the other. If, and only if, Alice and Bob *both* know the other's public key, then Step 1 can be ignored.
 
-Upon sending or receiving the first direct message, each client SHOULD add the other to their [NIP-02](https://github.com/nostr-protocol/nips/blob/master/02.md) contact list, but MUST at least save the other's public key locally if they want to continue messaging.
+Upon sending or receiving the first direct message, each client SHOULD add the other to their [NIP-02](https://github.com/nostr-protocol/nips/blob/master/02.md) contact list, but MUST at least save the other's public key locally if they want to continue messaging privately.
 
 **Step 2:**<br>
 Alice and Bob establish `Decoy Keys`, with which they will send and receive all subsequent messages to and from.
--- a/24.md
+++ b/24.md
@@ -1,0 +1,68 @@
+NIP-24
+======
+
+Private, Encrypted Direct Messages
+----------------------------------
+
+`draft` `optional` `author:jeffthibault`
+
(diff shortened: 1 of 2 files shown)


jb55:
This would allow either user to forge messages with each others decoy keys since they are both composed of information shared by both parties, is this intended?

jeffthibault:
@jb55 I guess if one of the users signed a message with the decoy key, it wouldn't matter because it's neither one's actual private key. Can you give an example of how they could forge. I might be misunderstanding.

jb55:
ah I think I misunderstood, the decoy keys are not used as the new secret keys which is what I thought in the first reading of this... interesting approach!

jeffthibault:
Yeah, the decoy keys are kinda like P.O. boxes that only the two participants know about. The relay is the post office in that it knows the two participants have the P.O. boxes (decoy keys) because you are filtering for them but it can't see what's inside.

monlovesmango:
I really like this. was envisioning something similar but don't know cryptography (yet)

curious if its possible to take this a step further to cover the author? would it be too insecure to do something like this? I believe this would allow you to author notes from decoy public key and sign with decoy private key, so no one would know who the message is from or to (except for the people messaging). if we're trying to remove metadata from DM's (and a lot of people seem to want this) author is probably worth trying to obfuscate as well.

A[sk] = Alice's private key
A[pk] = Alice's public key
A[dsk] = Alice's decoy private key
A[dpk] = Alice's decoy public key
B[sk] = Bob's private key
B[pk] = Bob's public key
B[dsk] = Bob's decoy private key
B[dpk] = Bob's decoy public key
AB[ss] = Alice and Bob's shared secret, derived using ECDH(A[sk], B[pk]) or ECDH(B[sk], A[pk])

Alice's decoy private key is defined as `A[dsk] = SHA256(AB[ss]) * A[sk]`
Bob's decoy private key is defined as `B[dsk] = SHA256(AB[ss]) * B[sk]`
Alice's decoy public key is defined as `A[dpk] = SHA256(AB[ss]) * A[pk]`
Bob's decoy public key is defined as `B[dpk] = SHA256(AB[ss]) * B[pk]`

apologies if I'm way off with any of this, just trying to learn and was fun to ponder.

jeffthibault:
Thanks for the feedback! I'm glad you like the proposal. I am not a cryptography person and just learning as well.

You make a great point that the author should be covered as well. I am working on an update for this NIP to address that.

Cameri (commented):
@jeffthibault have you implemented this? PR looks good but I'd like to see a working implementation

jb55:
I may take a stab at implementing this in damus since I am intrigued...

jeffthibault:
@Cameri @jb55 I am working on an update to this NIP which allows the sender's pubkey to be obfuscated as well. I will push an updated spec and POC code in a few days.

Cameri:
@RandyMcMillan had this to say: https://twitter.com/RandyMcMillan/status/1582842087409881088

PR #55, opened 2022-10-18
NIP-34 Relay Supported Features

Semisol:
Very simple NIP, adds `supported_features` to the relay information document that is an array of strings such as:
- `nip1.limit`
- `dev.semisol.search`

### DIFF
--- a/34.md
+++ b/34.md
@@ -1,0 +1,10 @@
+NIP-34
+======
+
+Relay Supported Features
+------------------------
+
+`draft` `optional` `author:Semisol`
+
+This NIP adds a field to the NIP-11 relay information document, `supported_features`, which is an array of strings containing a list of supported features.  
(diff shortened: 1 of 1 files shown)


fiatjaf:
I understand why you want this, but I don't think the argument is sound enough.

Ideally we should do more with less, and NIP-11 has already a ton of unused features. I think they could be used for your purposes -- or, alternatively we should remove stuff from it and replace with these newer things here.

I would also want to see an actual concrete real-world use case.

jb55:
I'm confused, nip-11 already defines `supported_nips`, isn't this the same thing being proposed here?

fiatjaf:
I understand the arguments, but these are not actual concrete real-world use cases. Or maybe I chose the wrong words. I want an _actual situation_.

PR #57, opened 2022-10-31
add NIP-19: bech32-encoding of stuff.

fiatjaf:
With an extra section for TLV-encoded arbitrary metadata. Let me know what you think.

I think this will be very much necessary, for example, for Minds-like integrations through which some profiles and events will generally be only accessible from one master relay from each platform (e.g. the Minds relay). @ottman @harding what do you think?

And also when the network grows and we start to see more segregation of contents between relays, this makes it easier for people to share their profiles, without having to say: "my pubkey is X, and please make sure to add relay Z to your client".

https://github.com/nostr-protocol/nips/blob/ce4869b9404b9c588ef92353999cbe787cb679c3/19.md

Is this correct @jb55? What else are you bech32-encoding? Event ids?

### DIFF
--- a/19.md
+++ b/19.md
@@ -4,21 +4,38 @@ NIP-19
 bech32-encoded entities
 -----------------------
 
-`draft` `optional` `author:jb55` `author:fiatjaf`
+`draft` `optional` `author:jb55` `author:fiatjaf` `author:Semisol`
+
+This NIP species all bech32-encoded entities.
+
+## Bare keys
 
 To prevent confusion and mixing between private keys, public keys and event ids, which are all 32 byte strings. bech32-(not-m) encoding with different prefixes can be used for each of these entities.
 
-For public keys, the prefix is `npub`, so, for example, `3bf0c63fcb93463407af97a5e5ee64fa883d107ef9e558472c4eb9aaaefa459d` translates to `npub180cvv07tjdrrgpa0j7j7tmnyl2yr6yr7l8j4s3evf6u64th6gkwsyjh6w6`.
+These are the possible bech32 prefixes:
 
-For private keys, the prefix is `nsec`.
+  - `npub`: public keys
+  - `nsec`: private keys
+  - `note`: note ids
 
-For kind-1 notes, the prefix is `note`.
+Example: the hex public key `3bf0c63fcb93463407af97a5e5ee64fa883d107ef9e558472c4eb9aaaefa459d` translates to `npub180cvv07tjdrrgpa0j7j7tmnyl2yr6yr7l8j4s3evf6u64th6gkwsyjh6w6`.
 
-## Optional extra metadata in bech32-encoded blobs
+## Shareable identifiers with extra metadata
 
-Parsers must expect to decode extra metadata that may exist after the initial 32 bytes. Any metadata must be added as a `TLV` (type-length-value) item, with `T` and `L` being 1 byte (`uint8`, range 0-255) each, and `V` being of the size indicated by `L`.
+When sharing a profile or an event, an app may decide to include relay information and other metadata such that other apps can locate and display these entities more easily.
 
-The possible standardized types are indicated here:
+For these events, the contents are a binary-encoded list of `TLV` (type-length-value), with `T` and `L` being 1 byte each (`uint8`, i.e. a number in the range of 0-255), and `V` being a sequence of bytes of the size indicated by `L`.
 
+These are the possible bech32 prefixes:
+
+  - `nprofile`: a nostr profile
+  - `nevent`: a nostr event
+
+These possible standardized `TLV` types are indicated here:
+
+- `0`: `special`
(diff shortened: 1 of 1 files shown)


harding:
What is bech32-2?

jb55:
balas makes a good point that the TLV might seem confusing on npubs since it changes your identifying key which people might partially memorize (i do this already for my npub). Also this screws up npub rendering in damus and I would have to figure out another way to display it, but both of these issues aren't too bad for me to deal with.

jb55:
On Mon, Oct 31, 2022 at 08:07:52AM -0700, William Casarin wrote:
>damus uses npub,nsec,note
>
>That way there is a uniform way of mentioning entities when you @ then in posts. Damus also will detect when you accidentally do @nsec and convert it to an npub mention.

I'll probably add nchan as well to reference channel ids

fiatjaf:
Well, the first characters of an npub with or without TLVs will always be the same, and you can shrink it when displaying, it is too big to be comfortable anyway even without TLVs.

jb55:
>Parsers must expect to decode extra metadata that may exist after the initial 32 bytes. Any metadata must be added as a `TLV` (type-length-value) item, with `T` and `L` being 2 bytes (`uint16` big-endian) each, and `V` being of the size indicated by `L`.

You mentioned you prefer 1 byte instead of 2? I'm happy with either.

fiatjaf:
Yes, I wrote this just copying from the BOLT spec, but I can change to 1 byte.

fiatjaf:
Are we going forward with this? 

I was right now going to say I was liking this design again more than the `nshare1` idea because I also think it would be interesting to have relays inside `note1` strings, so you could be sure that note would show up on other clients when you shared it.

Althougn that is not a direct argument against `nshare1` and might as well be an argument for `nshare1`.

fiatjaf:
I've revamped this PR. Please agree now.

19.md
```diff
@@ -4,21 +4,38 @@ NIP-19
 bech32-encoded entities
 -----------------------
 
-`draft` `optional` `author:jb55` `author:fiatjaf`
+`draft` `optional` `author:jb55` `author:fiatjaf` `author:Semisol`
+
+This NIP species all bech32-encoded entities.
```
eskema, line 9: This NIP specifies all bech32-encoded entities.

19.md
```diff
@@ -4,21 +4,38 @@ NIP-19
...
+## Shareable identifiers with extra metadata
 
-For private keys, the prefix is `nsec`.
+When sharing a profile or an event, an app may decide to include relay information and other metadata such that other apps can locate and display these entities more easily.
 
-For kind-1 notes, the prefix is `note`.
+For these events, the contents are a binary-encoded list of `TLV` (type-length-value), with `T` and `L` being 1 byte each (`uint8`, i.e. a number in the range of 0-255), and `V` being a sequence of bytes of the size indicated by `L`.
 
-## Optional extra metadata in bech32-encoded blobs
+These are the possible bech32 prefixes:
```
eskema, line 29: These are the possible bech32 prefixes with `TLV`:

I think it makes sense to speficy that these are the ones with a TLV, so it's a different line than from above..

19.md
```diff
@@ -4,21 +4,38 @@ NIP-19
...
-The possible standardized types are indicated here:
+These possible standardized `TLV` types are indicated here:
 
+- `0`: `special`
+  - depends on the bech32 prefix:
+    - for `nprofile` it will be the 32 bytes of the profile public key
+    - for `nevent` it will be the 32 bytes of the event id
 - `1`: `relay`
-  - in case the _npub_ writer wants to make sure its stuff will be found by the reader they can append this data so the reader will know where to look for basic metadata and events for that public key. This may be be included multiple times.
+  - A relay in which the entity (profile or event) is more likely to be found. This may be be included multiple times.
```
eskema, line 41: - A relay in which the entity (profile or event) is more likely to be found. This may be included multiple times.

eskema (commented):
minor stuff

Giszmo (changes requested):
The NIP should clarify that bech32 is only a representation format. Serialized events may not use these.

I would also suggest clients to only talk about for example "public key":

```
Enter your public key: 
```

but accept legacy hex, too.

The nip could suggest client apps to warn users when a hex key could also be a private key but that's tricky.

monlovesmango:
I agree. but my opinion matters very little bc I don't know much about this encoding stuff. will be cool to be able to share relay with a pubkey.

monlovesmango (approved):

ok300:
Just curious, why bech32 and not bech32m?

AFAIK the former has a checksum vulnerability when it ends with a p, which is fixed in the later.

fiatjaf:
Because that's how it was implemented first and it wasn't worth the trouble changing. I believe the vulnerability isn't very meaningful in the context of Nostr.

PR #58, opened 2022-11-04
Fix regex for SetMetadata in NIP1

w3irdrobot:
It appears the regex given in NIP1 for setting the username in the setmetadata event was slightly off. I think the fix here is what was intended. Though I think what was meant here was pretty obvious, to make it easier on future developers, I updated the regex to something that should work with just copying and pasting.


eskema:
I think that whole part is to be removed, there was discussion to not restrict what can be used as a username, so it can be any string

eskema:
because it's already used in the wild like that.. so no point in trying to dissuade people from doing it, might as well just go with the implementations and if needed use a different field for machine username

jb55:
On Fri, Nov 04, 2022 at 07:47:53AM -0700, Tiago Balas wrote:
>because it's already used in the wild like that.. so no point in trying to dissuade people from doing it, might as well just go with the implementations and if needed use a different field for machine username

It seems like most clients don't give a shit and are happy with complete nonsense in this field

PR #59, opened 2022-11-06
NIP-38 - Encrypted Group Chat using a single shared secret

vishalxl:
NIP-38 is modeled on NIP-28, group chat. Main difference is use of a shared secret, and using p-tags to mention members in channel create and update events. Simplicity is one of the main goals.

Implementation for the events mentioned ( 140-142) is in Nostr Console 0.0.9 beta release. 

Note 1: One important aspect of this NIP-38 ( kind 14x) is that they have a participant list. The 4x group chat is public by design, and it is unlikely that it will have a participant list anytime in future ( though it can add a PoW condition to participate easily). So the 14x public chat, whether it has secret shared secret, or even in case where its 'shared secret' is public, can still serve well because it allows the channel creator to mention as p-tags in 140/141 events who are the participants of the group. 

Edit: pull request made more for discussion.

### DIFF
--- a/38.md
+++ b/38.md
@@ -117,13 +117,23 @@ Clients that support this NIP will read that and know that they're invited to jo
 
 Other clients will decrypt this message, and ignore it, but the user can read that they have been invited to such a channel.
 
-
-
 ### Encryption
 
-All participants should ideally have the shared secret sent to them by the creator of the channel. If such a message is present, the client should get the shared secret from that messages, and then encrypt the 142 kind message using that.
+All participants should ideally have the shared secret sent to them by the creator of the channel. If such a message is present, the client should read the shared secret from that message, and then encrypt any 142 kind message they want to send using it.
 
 Only the given shared-secret private and public keys are used to encrypt or decrypt the messages. 
+
+## Adding and Removing Participants 
+
+### Adding Participants In Same Group
+
+To add participants to the group, the admin or creator of the group emits a kind 141 message with the updated p tags, which also include the pubkeys of new members. Along with this, the admin shall also communicate to the newly added members the group shared-secret, using the same mechanism as mentioned in section for kind 104 above. 
+
+### Removing Participants
+
+Once a group is created, all participants know the shared-secret, so to remove a participant, we need to change the shared-secret of the remaining or new group. The remaining group can now be considered a candidate for a new channel, and there will be established a logical relationship between the new channel and the old one. This is done by use of a tag in the creation event of the new channel. 
+
+The applications on looking at that tag can find out which was the precursor channel to this new channel, and display both the old and new channel seamlessly to the user, if that's needed.
 
 
 ## NIP-10 relay recommendations
--- a/38.md
+++ b/38.md
@@ -1,0 +1,142 @@
+
+NIP-38
+======
+
+Encrypted Group Chat
+--------------------
+
+`draft` `optional` `author:vishalxl` 
+
+This NIP defines new event kinds for encrypted group chat. 
+
+It works by creating a shared secret that is used to encrypt the group or channel messages. Only the participants know that secret, so only such members can read/write to this group chat.
+
+It reserves three event kinds (140-142) for immediate use and 7 event kinds (143-149) for future use.
+
+- `140 - create encrypted channel`
+- `141 - change channel metadata including participants`
+- `142 - send encrypted message`
+
+
+## Kind 140: Create Encrypted channel
+
+Create a Encrypted chat channel.
+
+In the channel creation `content` field, Client SHOULD include basic channel metadata (`name`, `about`, `picture` as specified in kind 41).
+
+```json
+{
+    "content": "{\"name\": \"Demo Channel\", \"about\": \"A test channel.\", \"picture\": \"https://placekitten.com/200/200\"}",
+    ...
+}
+```
+
+The members of the group ( people who are allowed to post there) should be added as p tags to this event.
+
+Further, the creator of the channel should send a Direct Message to each of the other participants, which will have the channel's shared-secret private/public keys. 
+
+The format of that message is:
+
+`App Encrypted Channels: inviting you to encrypted channel <channel id> encrypted using private public keys <shared private key> <corresponding shared public key>`
+
+The three id's mentioned should each be 64 byte long hex encoded strings. No quotes are used.
+
+Clients that support this NIP will read that and know that they're invited to join the given encrypted channel, and they can use the given secret key to send/read messages in that channel.
+Other clients will decrypt this message, and ignore it, but the user can read that they have been invited to such a channel.
+
+## Kind 141: Set Encrypted channel metadata
+
+Update an encrypted channel's public metadata including its particpants.
+
+Clients and relays SHOULD handle kind 141 events similar to kind 0 `metadata` events.
+
+Clients SHOULD ignore kind 141s from pubkeys other than the kind 140 pubkey.
+
+Clients SHOULD support basic metadata fields:
+
+- `name` - string - Channel name
+- `about` - string - Channel description
+- `picture` - string - URL of channel picture
+
+Clients MAY add additional metadata fields.
+
+Clients SHOULD use [NIP-10](10.md) marked "e" tags to recommend a relay.
+
+```json
+{
+    "content": "{\"name\": \"Updated Demo Channel\", \"about\": \"Updating a test channel.\", \"picture\": \"https://placekitten.com/201/201\"}",
+    "tags": [["e", <channel_create_event_id> <relay-url>]],
+    ...
+}
+```
+
+
+## Kind 142: Create encrypted channel message
+
+Send a text message to an encrypted channel.
+
+Clients SHOULD use [NIP-10](10.md) marked "e" tags to recommend a relay and specify whether it is a reply or root message.
+
+Clients SHOULD append [NIP-10](10.md) "p" tags to replies.
+
+### Format
+
+Root message:
+
+```json
+{
+    "content": <string>,
+    "tags": [["e", <kind_40_event_id> <relay-url> "root"]],
+    ...
+}
+```
+
+Reply to another message:
+
+```json
+{
+    "content": <string>,
+    "tags": [
+        ["e", <kind_42_event_id> <relay-url> "reply"],
+        ["p", <pubkey> <relay-url>],
+        ...
+    ],
+    ...
+}
+```
+
+### Encryption
+
+All participants should ideally have the shared secret sent to them by the creator of the channel. If such a message is present, the client should get the shared secret from that messages, and then encrypt the 142 kind message using that.
+
+Only the given shared-secret private and public keys are used to encrypt or decrypt the messages. 
+
+
+## NIP-10 relay recommendations
+
+For [NIP-10](10.md) relay recommendations, clients generally SHOULD use the relay URL of the original (oldest) kind 40 event.
+
+Clients MAY recommend any relay URL. For example, if a relay hosting the original kind 40 event for a channel goes offline, clients could instead fetch channel data from a backup relay, or a relay that clients trust more than the original relay.
+
+
+Future extensibility
+--------------------
+
+We reserve event kinds 143-149 for other events related to chat, to potentially include new types of media (photo/video), moderation, etc.
+
+
+Motivation
+----------
+This is one step to make some private channels. 
+
+Drawbacks
+---------
+The meta-data for any message is practically much public in nostr. So such group chat should not be considered private. Anybody can see who is sending message(s) to what channel, and when.
+
+Further, if any single participant of the group chat leaks the shared secret ( whether intentionally or by accident), all the messages (past or future) can then be decrypted by others.
+
+
+
+Additional info
+---------------
+
--- a/38.md
+++ b/38.md
@@ -33,7 +33,7 @@ In the channel creation `content` field, Client SHOULD include basic channel met
 
 The members of the group ( people who are allowed to post there) should be added as p tags to this event.
 
-Further, the creator of the channel should send a Direct Message to each of the other participants, which will have the channel's shared-secret private/public keys. 
+Further, the creator of the channel should send a Direct Message to each of the other participants, which will have the channel's shared-secret private/public keys. The Direct message should be sent by a kind other than 4; but presently kind 4 direct messages are used. 
 
 The format of that message is:
 
--- a/38.md
+++ b/38.md
@@ -11,8 +11,9 @@ This NIP defines new event kinds for encrypted group chat.
 
 It works by creating a shared secret that is used to encrypt the group or channel messages. Only the participants know that secret, so only such members can read/write to this group chat.
 
-It reserves three event kinds (140-142) for immediate use and 7 event kinds (143-149) for future use.
+It reserves 4 event kinds (104 and 140-142) for immediate use and 7 event kinds (143-149) for future use.
 
+- `104 - used to communicate the shared secret`
 - `140 - create encrypted channel`
 - `141 - change channel metadata including participants`
 - `142 - send encrypted message`
@@ -33,17 +34,7 @@ In the channel creation `content` field, Client SHOULD include basic channel met
 
 The members of the group ( people who are allowed to post there) should be added as p tags to this event. The creator should also add themselves as p tag, if they want to participate there too.
 
-Further, the creator of the channel should send a Direct Message to each of the other participants, which will have the channel's shared-secret private/public keys. The Direct message should be sent by a kind other than 4; but presently kind 4 direct messages are used. 
-
-The format of that message is:
-
-`App Encrypted Channels: inviting you to encrypted channel <channel id> encrypted using private public keys <shared private key> <corresponding shared public key>`
-
-The three id's mentioned should each be 64 byte long hex encoded strings. No quotes are used.
-
-Clients that support this NIP will read that and know that they're invited to join the given encrypted channel, and they can use the given secret key to send/read messages in that channel.
-
-Other clients will decrypt this message, and ignore it, but the user can read that they have been invited to such a channel.
+On creating a channel, the creator should also generate a unique/new private-public key pair which will serve as the `shared-secret` for a given channel. The creator of the channel should then communicate this secret to other members as described in section for `kind 104`.
 
 Note: Clients for a user may create a channel where the only participant is the creator of the channel. Then new members can be added by sending a kind 141 message, as discussed further. 
 
@@ -73,7 +64,7 @@ Clients SHOULD use [NIP-10](10.md) marked "e" tags to recommend a relay.
 }
 ```
 
-Further, the main use case of kind 141 event is to update the participants list. New p tags may be added, and old ones removed with the 141 event for an encrypted channel. Client will internally then manage participants. It should be noted that since the shared-secret is not changeable,  the 'removed' members can continue to read all channel messages. So effectively, it may be said that removal does not really work. However, addition of new members just shares the shared-secret with new members, and will prove useful when required.
+Further, the main use case of kind 141 event is to update the participants list. New p tags may be added, and old ones removed with the 141 event for an encrypted channel. Client will internally then manage participants. It should be noted that since the shared-secret is not changeable,  the 'removed' members can continue to read all channel messages.
 
 If new members are added, then its the responsibility of the creator's client to share the shared-secret with them as mentioned in kind 140 section. 
 
@@ -91,7 +82,7 @@ Root message:
 ```json
 {
     "content": <string>,
-    "tags": [["e", <kind_40_event_id> <relay-url> "root"]],
+    "tags": [["e", <kind_140_event_id>],
     ...
 }
 ```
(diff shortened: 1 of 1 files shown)


38.md
```diff
@@ -0,0 +1,147 @@
...
+    "content": "{\"name\": \"Demo Channel\", \"about\": \"A test channel.\", \"picture\": \"https://placekitten.com/200/200\"}",
+    ...
+}
+```
+
+The members of the group ( people who are allowed to post there) should be added as p tags to this event. The creator should also add themselves as p tag, if they want to participate there too.
+
+Further, the creator of the channel should send a Direct Message to each of the other participants, which will have the channel's shared-secret private/public keys. The Direct message should be sent by a kind other than 4; but presently kind 4 direct messages are used. 
+
+The format of that message is:
```
jb55, line 38: we really shouldn't do this over kind 4. this is really hacky. ideally it would be a protocol message that clients can handle without parsing, display an invite properly, etc. Users shouldn't have to see a shared secret at all.

38.md
```diff
@@ -0,0 +1,147 @@
...
+
+The three id's mentioned should each be 64 byte long hex encoded strings. No quotes are used.
+
+Clients that support this NIP will read that and know that they're invited to join the given encrypted channel, and they can use the given secret key to send/read messages in that channel.
+
+Other clients will decrypt this message, and ignore it, but the user can read that they have been invited to such a channel.
+
+Note: Clients for a user may create a channel where the only participant is the creator of the channel. Then new members can be added by sending a kind 141 message, as discussed further. 
+
+## Kind 141: Update metadata and Add participants
```
jb55, line 50: Something me and @Semisol talked about awhile back was creating a new shared secret whenever a participant is removed, and this is encrypted into tags for each participant. 

The client would then query all the update messages and add all of the session keys to their current set. Then for older messages it would try to decrypt with each key in the set until a successful decryption worked.

This would enable you to re-key channels once participants are removed with a pretty simple algo. Also you don't need to send invites in this case either, the key is automatically encrypted to the user in the channel creation/update tags.

jb55 (changes requested):

vishalxl:
A related note. We may need to have in total three types of "channels":

a) totally public kind 4x channels 
b) encrypted channels with membership (like this proposed NIP 38 with kind 14x) 
c) public channels with membership ( where content is public but only allowed people are considered in), could be kind 24x

38.md
vishalxl: Agree with this that there is no good reason to use kind 4 here. 

Invite suggestions on what kind it should be. I think it would be good to have a special kind where different apps can store or send their information from where they can pick it up. So that user can know what app is sending what message ( given that its the user itself who is sending these messages, or the client is doing so on user's behalf) 

Also then the format of the message, or its 'header' or first few words etc, should help differentiate between different apps.

vishalxl: Edit the proposed NIP 38 and mentioned kind 104 rather than kind 4 to communicate the shared-secret.

vishalxl: There is not much parsing involved for shared-secret message. Once the app matches the first few words, which are "App Encrypted Channels: ", then the required information is at well-known points in the message. If strInvite is the given message as a string then the required values are always at the same place:

Encrypted Channel id : strInvite.substring(58, 58 + 64) 
Secret private key        : strInvite.substring(159, 159 + 64); 
Secret public key        : strInvite:substring(224, 224 + 64);

And the length of strInvite should be 288 chars.

cmdruid:
IME, encrypted group chat is a rabbit-hole of complexity.

A lot of folks use double ratcheting as a standard. All the requirements are found in WebCrypto: 
https://en.wikipedia.org//wiki/Double_Ratchet_Algorithm

I have seen this method scale (and fail) gracefully on rooms with 100+ members. So maybe that is a good option.

Otherwise, I would suggest a protocol that uses forward secrecy by default, so that updating the shared secret is easy. A chat room has to broadcast frequently, so updating the secret is no big deal. Everyone has to agree on the state of channel members though.

I also think that if there are plans to use group encryption and membership, you may as well add some role / admin features as well. Matrix uses a concept of "power levels" that works really well.

Anyway, my 2 sats on the discussion. :-D

1440000bytes:
Thanks for writing this NIP. I think it can be used as an alternative to https://github.com/nostr-protocol/nips/pull/49 for joinstr.

I have not tested it yet in nostr console. Will do it today.

vishalxl:
> Thanks for writing this NIP. I think it can be used as an alternative to #49 for joinstr.
> 
> I have not tested it yet in nostr console. Will do it today.

I think the requirement of verifiability that everybody is getting the same secret can be done by mandating that everybody make a (encrypted) comment in the group chat. Once that is done, then its verified that everyone knows the same secret. If that's what you need.

38.md
vishalxl: Removing of participant ( or cases where a new secret is needed for all ) can be catered to by this method, where we allow the creator  (or anyone for that matter) to "fork" a group, so that the new fork will have a new id, new secret. We mark the new group as a 'fork' by adding some specific tag info. 

Then internally in clients, user can be seamlessly shown messages from what are two channels ( if you go by channel id itself), but are logically the same cause the client knows which channels are connected logically as same. 

This is the simplest way I think we can implement removal: it also has the logical consistency ( which will make coding easier imo), that each channel id is only ever connected with one shared secret. But using extra tags, we can establish a logical connection ( that 2+ channels are same channel as far as user knows).

fiatjaf:
I think this may be a superior alternative, since it would be fully stateless like NIP-04: https://github.com/nostr-protocol/nips/issues/72#issuecomment-1353592661

38.md
vishalxl: I now think the above message should ideally be in json format. will work on it further in that direction. 

Unless you have some other idea there?

1440000bytes:
@fiatjaf 

> I think this may be a superior alternative, since it would be fully stateless like NIP-04: https://github.com/nostr-protocol/nips/issues/72#issuecomment-1353592661

Can you explain the stateless part in detail or what are the advantages of this method over the one shared in this PR?

@vishalxl 

> The meta-data for any message is practically much public in nostr. So such group chat should not be considered private. Anybody can see who is sending message(s) to what channel, and when.

This might not be perfect for all use cases however it should work for some. As long as shared secret and messages are private it will work for 2 of my projects: joinstr and p2p exchange

Not sure if any other client and library has already implemented it apart from nostr-console.

> Once members are 'removed' by using the kind 141 event for a channel, they can continue to read the new messages posted in the group, because currently there is no provision to change the shared secret of a single group. The remaining users, may howerver choose to create a new group where the recently-removed member is not a member, and which would have a new shared-secret.

This should not be a concern if clients implement it correctly and create a new group chat once someone leaves with the rest of the members.

fiatjaf:
> Can you explain the stateless part in detail or what are the advantages of this method over the one shared in this PR?

Each event contains its own decryption keys, just like NIP-04, that is what I called "stateless".

vishalxl:
> aB, aC and aD

The three secrets that Alice generates in that scheme you mentioned @fiatjaf , I can't tell what happens to them, in that where do they go after steps 4 and 5, as they are not mentioned after 5? I am having trouble understanding what is happening here. 

So in step 4 Alice generates three secrets, one each for 3 receivers. In step 5 she uses them to encrypt x. Then Later are those secrets never needed by Alice or any other people? Since x was encrypted with those secrets, wont' they be needed again to decrypt x? And if they will be needed, by whom are they needed? Are they given to respective recipients?

cmdruid:
What about a replaceable event that simply provides each member with the latest shared secret for a channel, encrypted to their pubkey?

If you are a member on the list, then there will be a shared secret available for you to decrypt. Otherwise, you get nothing.

Whenever the channel owner wants to update the member list, they simply update the event with a new shared secret, and update the encrypted keys for each member. Members can subscribe to the replaceable event to get the latest key.

You could also throw other channel metadata in there as well, just use the whole content field as a metadata store for the channel.

vishalxl:
> What about a replaceable event that simply provides each member with the latest shared secret for a channel, encrypted to their pubkey?

Pretty much this is what is being done in this NIP 38, where the single shared secret is being shared using kind 104, which is encrypted to the recipeients respective pubkey. So member m1 gets from creator C a kind 104 which is like a DM, which mentions the shared secret. Now whether that's replaceable or not is up for discussion. 

> Whenever the channel owner wants to update the member list, they simply update the event with a new shared secret, and update the encrypted keys for each member. Members can subscribe to the replaceable event to get the latest key.

Updation is propsed differently - the NIP 38 proposes that a single channel is always only associated with a single secret. To remove members, for example, a new channel is created with new secret. This keeps the logic simple , as in one channel, one secret, always. 

> You could also throw other channel metadata in there as well, just use the whole content field as a metadata store for the channel.

I am thinking of having a json for kind 104, with details about group also thrown in, if that's what you mean too.

cmdruid:
> Pretty much this is what is being done in this NIP 38, where the single shared secret is being shared using kind 104, which is encrypted to the recipeients respective pubkey. So member m1 gets from creator C a kind 104 which is like a DM, which mentions the shared secret. Now whether that's replaceable or not is up for discussion.

Can you elaborate on this more? Are you encrypting the private key using the public key and ECDH?

cmdruid:
Is it something like this:

## Both Parties
```ts
// Generate keypair for channel owner and channel member.
const { ownerPubKey,  ownerPrivKey  } = KeyPair.generate(32)
const { memberPubKey, memberPrivKey } = KeyPair.generate(32)

// Both parties compute shared secret between owner <-> member.
const sharedSecret = 
     ECDH(memberPubKey, ownerPrivKey)
  || ECDH(ownerPubKey, memberPrivKey)

// Use shared secret to derive a shared keypair.
const { sharedPubKey, sharedPrivKey } = KeyPair.from(sharedSecret)
```
## Channel Owner
```ts
// The channel owner create a secret for the channel.
const channelSecret = Random.bytes(32)

// To share the channel secret with a select member,
// owner encrypts the secret using sharedPubKey
const encryptionKey = ECDH(ownerPrvKey, sharedPubKey)
const encryptedString = encrypt(encryptionKey, channelSecret)

// Use tag field as a hash map for distributing member keys.
const memberTag : HashMap = [
  sharedPubKey,    // Hex-encoded compressed public key.
  encryptedString  // AES-encrypted using member key and ECDH.
]

// Example of a channel event.
const channelEvent = {
  id      : 'eventId'
  kind    : 10000
  tags    : [ memberTag ]
  content : encryptedMetaData
  pub     : ownerPublicKey
  sig     : ownerSignature
}
```
## Channel Member
```ts
// Check if shared pubKey exists in tags hash map.
if (channel.tags.has(sharedPubKey)) {
  const encryptedString = tags.get(sharedPubKey)
  // Use private key to derive secret between shared <-> owner.
  const encryptionKey = ECDH(sharedPrvKey, ownerPubKey)
  // Decrypt channel secret using shared encryption key.
  const channelSecret = decrypt(encryptionKey, encryptedString)
  // Decrypt the channel meta data using the channel secret.
  const channelMetaData = decrypt(channelSecret, encryptedMetaData)
}
```

vishalxl:
@cmdruid that's pretty much the outline of this NIP 38 in code/pesudo-code. I have added some mention of kind 104, and otherwise edited to make it more in sync: 


## Both Parties

```
// Generate keypair for channel owner and channel member.
const { ownerPubKey,  ownerPrivKey  } = KeyPair.generate(32)
const { memberPubKey, memberPrivKey } = KeyPair.generate(32)

// Both parties compute shared secret between owner <-> member.
const sharedSecret = 
     ECDH(memberPubKey, ownerPrivKey)
  || ECDH(ownerPubKey, memberPrivKey)

// Use shared secret to derive a shared keypair.
const { sharedPubKey, sharedPrivKey } = KeyPair.from(sharedSecret)
```

Note: this shared secret is also used in kind 4 messages. 

## Channel Owner

When creating the channel, the channel creator will create a random 'shared secret' and then create two messages

a) First is a kind 104 message using the following logic, which is sent to EACH of the members ( it contains the shared secret, encrypted using the newly generated key mentioned in part "both parties")

```
// The channel owner create a secret for the channel.
const channelSecret = Random.bytes(32)

// To share the channel secret with a select member,
// owner encrypts the secret using sharedPubKey
const encryptionKey = ecdh(ownerPrvKey, sharedPubKey)
const encryptedString = encrypt(encryptionKey, channelSecret)

// cmd: Use tag field as a hash map for distributing member keys. 
// my comment: this is just application logic. The application will have generated different secret for each intended member/recipient, so it will send them to them using their specific secret. This is the kind 104 , which is modeled on kind 4.
const memberTag : HashMap = [
  sharedPubKey,    // Hex-encoded compressed public key.
  encryptedString  // AES-encrypted using member key and ECDH.
]


```

b) second is a kind 140 event with following format, which also has member tags so people can fetch it with #p REQ: 

```
// Example of a channel event.
const channelEvent = {
  id      : 'eventId',
  kind    : 10000,
  tags    : [ memberTag ],
  content : encryptedMetaData,
  pub     : ownerPublicKey,
  sig     : ownerSignature
}
```

Note: encryptedMetaData is in specific 'one line' format in this NIP 38 as of now, but should ideally be in json to add more information. 

## Channel Member

```
// Check if shared pubKey exists in tags hash map.
if (channel.tags.has(sharedPubKey)) {
  const encryptedString = tags.get(sharedPubKey)
  // Use private key to derive secret between shared <-> owner.
  const encryptionKey = ecdh(sharedPrvKey, ownerPubKey)
  // Decrypt channel secret using shared encryption key.
  const channelSecret = decrypt(encryptionKey, encryptedString)
  // Decrypt the channel meta data using the channel secret.
  const channelMetaData = decrypt(channelSecret, encryptedMetaData)
}
```

cmdruid:
> NB I am assuming kind 4 works like this ECDH logic you mentioned, without verifying. Because kind 104 follows just how kind 4 does.

Yeah ECDH meaning Elliptic-Curve Diffe-Hellman, used to derive the shared secret. NIP-04 uses that and then encrypts using AES with random IV.

Does the current proposal send DMs to all members on each channelSecret update? If that's the case, would it scale better for members to subscribe to the channel memberlist for updates, while using DMs sparingly to send out invites?

Also side note: Since interim sharedPubKey obfuscates the real participating pubkeys, I don't believe the memberlist has any sensitive information? I think it would be neat if channelEvents are used to set up an encrypted group chat, but without being publicly linked the group chat itself, or any of its members. Then it would simply be a rendezvous for members to get the current state of the encrypted group.

I like to use a hash of the channelSecret to tag the group chat, as publishing the hash does not reveal the secret, and members can subscribe to the tag as a group feed. You could also use the encryptedMetaData to store the real list of member pubkeys, so the sharedKeys aren't used outside of delivering the channelSecret.

vishalxl:
> Does the current proposal send DMs to all members on each channelSecret update? If that's the case, would it scale better for members to subscribe to the channel memberlist for updates, while using DMs sparingly to send out invites?

DM's ( of kind 104) need to be sent only when the secret is shared. Then no DM's need to be sent.

> I like to use a hash of the channelSecret to tag the group chat, as publishing the hash does not reveal the secret, and members can subscribe to the tag as a group feed. You could also use the encryptedMetaData to store the real list of member pubkeys, so the sharedKeys aren't used outside of delivering the channelSecret.

So then this will happen: creator gives shared secret key in DM to members. From that the members can derive its corresponding public key, and when they want to send messages to the group, they use that 'public key' as an e tag, lets call it channel-ID-tag. 

So when a member m1 sends a group chat message, all that outsiders will see is some random e tag is attached to the message. They would have no idea what that e tag is. They will also see other people using that e tag, so people can guess its one group. 

And since the member list is published as encrypted data, there is no real member list. In fact, there can be MULTIPLE channel-ID-tag, and the group would be told, as encrypted message, all the channel-ID-tag's of a group. The members can send message using any of the tags, and all the receivers can fetch all the tags. And since the set of channel-ID-tag for any channel was encrypted from start, no one outside would have any clue who is messaging as a group, as long as enough number of channel-ID-tag 's are used for each channel. 

Along with encrypted membership list as you mentioned too ( which can be updaed with 141 kind as encrypted data too), this would give a lot of 'privacy' to users. Or some amount of anonymity, compared to the very base case scenario.

38.md
```diff
@@ -0,0 +1,170 @@
...
+
+Only the given shared-secret private and public keys are used to encrypt or decrypt the messages. 
+
+## Adding and Removing Participants 
+
+### Adding Participants In Same Group
+
+To add participants to the group, the admin or creator of the group emits a kind 141 message with the updated p tags, which also include the pubkeys of new members. Along with this, the admin shall also communicate to the newly added members the group shared-secret, using the same mechanism as mentioned in section for kind 104 above. 
+
+### Removing Participants
```
itslenny, line 132: What if a user wants to remove themself? Obviously the client can just ignore messages, but it seems like a good client should also notify the relay that they left the room, and the admin should remove them from the list and roll the keys.

38.md
```diff
@@ -0,0 +1,170 @@
...
+        ["e", <kind_142_event_id_replied_to>], 
+        ["p", <pubkey> <relay-url>],
+        
+        ...
+    ],
+    ...
+}
+```
+
+## Kind 104: Communicate the Shared-Secret
```
itslenny, line 106: This part could handle encryption the way it is done on https://github.com/nostr-protocol/nips/pull/168 instead. This creates a single DM that can be read by multiple people by encrypting it with a unique secret that is included in the tags for every recipient.

This avoids the channel owner needing to send N messages every time someone is removed from the group.

38.md
```diff
@@ -0,0 +1,170 @@
...
+
+Upon adding a new member to a encrypted group of kind 140, the creator of the channel should send a direct message ( of kind 104) to each of the other participants, which will have the channel's shared-secret private/public keys. The direct message should of `kind` 104; it must follow the message construction rules as are there for `kind 4` messages, except for the difference in value of `kind`.
+
+The format of that message is:
+
+`App Encrypted Channels: inviting you to encrypted channel <channel id> encrypted using private public keys <shared private key> <corresponding shared public key>`
```
itslenny, line 112: This message feels easy for humans to read and hard for machines to read. We should optimize for the opposite. Also, the public key can be derived from the public key so it does not need to be included.

My suggestion would be `content="<shared private key>"`

Then, include the <channel id> as a tag of some sort.

guilhermegps:
Any news about it?

MTG2000:
What's the state of this NIP??
The last update is a couple of months ago.

So is this still relevant?

vishalxl:
Status as of May 2023: This is implemented, the way its mentioned right now, in nostr console. 

There are some improvements needed, or suggested, such as using json format to share initial shared secret etc, and there are many other improvements possible, most mentioned here in comments. I will like to implement them at least in nostr console ( but nothing certain about when that may happen). 

Others may also take up and implement these ideas in their/other clients.

38.md
```diff
@@ -0,0 +1,170 @@
...
+
+```json
+{
+    "content": "{\"name\": \"Updated Demo Channel\", \"about\": \"Updating a test channel.\", \"picture\": \"https://placekitten.com/201/201\"}",
+    "tags": [["e", <channel_create_event_id> <relay-url>]],
+    ...
+}
+```
+
+Further, the main use case of kind 141 event is to update the participants list. New p tags may be added, and old ones removed with the 141 event for an encrypted channel. Client will internally then manage participants. It should be noted that since the shared-secret is not changeable,  the 'removed' members can continue to read all channel messages.
```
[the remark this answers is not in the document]
earonesty: equivalent to rolling a new room and saying that's where we're talking now

38.md
```diff
@@ -0,0 +1,170 @@
...
+
+Clients SHOULD use [NIP-10](10.md) marked "e" tags to recommend a relay.
+
+```json
+{
+    "content": "{\"name\": \"Updated Demo Channel\", \"about\": \"Updating a test channel.\", \"picture\": \"https://placekitten.com/201/201\"}",
+    "tags": [["e", <channel_create_event_id> <relay-url>]],
+    ...
+}
+```
```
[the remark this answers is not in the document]
earonesty: tags should be in the envelope, yes.

38.md
earonesty: this should not use a nip04 style encryption, instead it should use directional encryption 

encrypt({
sender pub,
sender sig of this payload,
content=....
tags=...
}, ephemeral pub),

no tags outside the envelope, pubkey is ephemeral.   recipient verifies the sender's provenance after decryption

*all* messages should work this way in this protocol

38.md
earonesty: that's a good idea

38.md
earonesty: no need for public key, it can be derived

i'll throw up a nip for "more secret" dm's

38.md
```diff
@@ -0,0 +1,170 @@
...
+Create a Encrypted chat channel.
+
+In the channel creation `content` field, Client SHOULD include basic channel metadata (`name`, `about` and `picture`).
+
+```json
+{
+    "content": "{\"name\": \"Demo Channel\", \"about\": \"A test channel.\", \"picture\": \"https://placekitten.com/200/200\"}",
+    ...
+}
+```
```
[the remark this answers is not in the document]
earonesty: yes, we can avoid this by using ephemeral public keys listed here:https://github.com/nostr-protocol/nips/discussions/570#discussioncomment-6032245

38.md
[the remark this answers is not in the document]
earonesty: see also: https://github.com/nostr-protocol/nips/discussions/570#discussioncomment-6032245  for a nice replacement that encrypts everything in a channel

38.md
earonesty: @paulmillr

staab:
Closing in favor of #574, #686 and possibly #875

PR #62, opened 2022-11-10
NIP-20: Command Results

jb55:
When submitting events to relays, clients currently have no way to know if an event was successfully committed to the database. This NIP introduces the concept of command results which are like NOTICE's except provide more information about if an event was accepted or rejected.

[Doc](https://github.com/nostr-protocol/nips/blob/c510e646d0fd5bfa56a044756182cc10ad7e494d/20.md)

An earlier draft of this proposal is implemented on `nostr-rs-relay` on this branch:

https://git.sr.ht/~gheartsfield/nostr-rs-relay/log/feat-notice-like-responses

Will update it soon to match the NIP once this is reviewed.

### DIFF
--- a/20.md
+++ b/20.md
@@ -1,0 +1,63 @@
+NIP-20
+======
+
+
+Command Results
+---------------
+
+`draft` `optional` `author:jb55`
+
+When submitting events to relays, clients currently have no way to know if an event was successfully committed to the database. This NIP introduces the concept of command results which are like NOTICE's except provide more information about if an event was accepted or rejected.
+
+A command result is a JSON object with the following structure that is returned when an event is successfully saved to the database or rejected:
+
+    ["OK", <event_id>, <true|false>, <message>]
+
+Relays MUST return `true` when the event is a duplicate and has already been saved. The `message` SHOULD start with `duplicate:` in this case.
+
+Relays MUST return `false` when the event was rejected and not saved.
+
+The `message` SHOULD provide additional information as to why the command succeeded or failed.
+
+The `message` SHOULD start with `blocked:` if the pubkey or network address has been blocked, banned or is not on a whitelist.
+
+The `message` SHOULD start with `rate-limited:` if the event was rejected due to rate limiting techniques.
+
+Ephemeral events are not acknowledged with OK responses, unless there is a failure.
+
+If the event is malformed and could not be parsed, a NOTICE message SHOULD be used instead of a command result.
+
+
+Examples
+--------
+
+Event successfully written to the database:
+
+    ["OK", "b1a649ebe8b435ec71d3784793f3bbf4b93e64e17568a741aecd4c7ddeafce30", true, ""]
+
+Event successfully written to the database because of a reason:
+
+    ["OK", "b1a649ebe8b435ec71d3784793f3bbf4b93e64e17568a741aecd4c7ddeafce30", true, "pow: difficulty 25>=24"]
+
+Event blocked due to ip filter
+
+    ["OK", "b1a649ebe8...", false, "blocked: tor exit nodes not allowed"]
+
+Event blocked due to pubkey ban
+
+    ["OK", "b1a649ebe8...", false, "blocked: you are banned from posting here"]
+
+Event blocked, pubkey not registered
+
+    ["OK", "b1a649ebe8...", false, "blocked: please register your pubkey at https://my-expensive-relay.example.com"]
+
+Event rejected, rate limited
+
+    ["OK", "b1a649ebe8...", false, "rate-limited: slow down there chief"]
+
+
+
+Future Extensions
+-----------------
+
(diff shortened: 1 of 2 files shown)


fiatjaf:
Looks good to me.

0xtlt:
We could also think of an error message for a required POW level 👍

jb55:
added `pow` and `invalid` event message suggestions

jb55:
> an optional json object should be added after `message` which can be extended based on NIPs (such as extended content, which could say link to a website) and the command in question (not everything can be inserted into the text string prefix)

the messages are meant for humans, with prefixes so that clients can be slightly more intelligent with what to do with them. For instance, for the `rate-limit` type: the client may not show anything and simply try again with a longer timeout.

for the `pow:` prefix it may query relay metadata to get the updated difficulty requirement and try again in the background.

for the `invalid` and `blocked` prefix the client may wish to show these as styled error popups.

I should probably include all this in the NIP...

jb55:
> It's not clear that any additional info is required, so maybe we can defer the json object thing for when we need that?

actually the `pow` thing may eventually want per-user pow requirements, so in that case it would make sense to return extra metadata. maybe that can be it's own NIP though...

fiatjaf:
I agree, the object can be useful, but it's better to deal with it when it is actually needed.

x1ddos:
> If the event is malformed and could not be parsed, a NOTICE message SHOULD be used instead of a command result.

curious why this distinction for malformed messages. why not respond to the client with something like:

```
["OK", "b1a649ebe8...", false, "malformed: ..."]
```

jb55:
> > If the event is malformed and could not be parsed, a NOTICE message SHOULD be used instead of a command result.
> 
> curious why this distinction for malformed messages. why not respond to the client with something like:
> 
> ```
> ["OK", "b1a649ebe8...", false, "malformed: ..."]
> ```

what ID would be returned?

x1ddos:
> > > If the event is malformed and could not be parsed, a NOTICE message SHOULD be used instead of a command result.
> > 
> > 
> > curious why this distinction for malformed messages. why not respond to the client with something like:
> > ```
> > ["OK", "b1a649ebe8...", false, "malformed: ..."]
> > ```
> 
> what ID would be returned?

ah, of course! makes sense.

> We could perhaps set the ID to nothing and maybe that's ok?

nip-10 says:

> This NIP defines no rules for how NOTICE messages should be sent or treated.

but maybe it's defined elsewhere. sending back a response with an empty ID also doesn't sound too bad.

i'm too knew to this protocol - can't really say which is better - but to me it seems like responding back directly with an empty ID is more consistent. if i'm implementing a client compatible with nip-20, it's probably easier to code it while always expecting a response back.

also, if i'm implementing the nip in a relay, this part:

> Relays MUST return false when the event was rejected and not saved.

makes me thing i must return something even on malformed events.

jb55:
> makes me feel compelled to always return something, even on malformed events.

I think the confusion is the fact that a malformed event is not an event, it's just garbage. This nip only applies to valid events that otherwise may be rejected for it's contents or sender.

Keep in mind that you can still return `invalid:` if your id or signature is wrong, assuming the basic json structure is valid.

jeffthibault:
Maybe a dumb question but why use "OK" when rejecting an event?

x1ddos:
> > makes me feel compelled to always return something, even on malformed events.
> 
> I think the confusion is the fact that a malformed event is not an event, it's just garbage. This nip only applies to valid events that otherwise may be rejected for it's contents or sender.
> 
> Keep in mind that you can still return `invalid:` if your id or signature is wrong, assuming the basic json structure is valid.

yeah, makes sense. i'd suggest extending the part about NOTICE in the nip spec a little bit, just to be explicit in describing the reasoning.

Cameri:
If it's always OK, but you still need to read the third argument to know whether it was a fail or pass it makes OK kind of redundant so I suggest we change OK for ACK and NACK and drop the boolean. Everything else stays the same.

If event is malformed, clients should not expect a response but if the relay is nice it should be a NOTICE.
The reason is that an event ID can only be produced/obtained if the event is valid as per NIP-01.

Cameri (approved):
LG, please consider `ACK/NACK` instead of `OK,...,false`

fiatjaf:
I think it is better to use `"OK"` so you can have a single `case` handler for `"OK"` and read the id and the boolean from there. The word `"OK"` doesn't have to mean anything, it could be `"RESULT"` but that is bigger and boring.

fiatjaf:
Implemented here: https://github.com/fiatjaf/relayer/blob/d69479974c26d95dd4b626d9deea8e47653fa73f/handlers.go#L119-L149

jb55:
On Thu, Nov 10, 2022 at 06:05:52PM -0800, Ricardo Arturo Cabral Mejía wrote:
***@***.*** approved this pull request.
>
>LG, please consider `ACK/NACK` instead of `OK,...,false`

I just wanted to have a single term to represent this feature. I think
OK is fine because I see it as "Is everything OK?" which could be true
or false.

PR #63, opened 2022-11-10
nip16: clarify about the signers of replaceable events

x1ddos:
A newer replaceable event must be signed by the same key.

### DIFF
--- a/16.md
+++ b/16.md
@@ -11,7 +11,7 @@ Relays may decide to allow replaceable and/or ephemeral events.
 Replaceable Events
 ------------------
 A *replaceable event* is defined as an event with a kind `10000 <= n < 20000`.
-Upon a replaceable event with a newer timestamp than the currently known latest replaceable event with the same kind being received, the old event SHOULD be discarded and replaced with the newer event.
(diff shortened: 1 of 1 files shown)


jb55:
LGTM

Cameri (changes requested):
I don't think this belongs here. Please ddd this clarification to NIP-26 instead of NIP-16 otherwise people seeking to implement NIP-16 need to care about NIP-26 even if they don't implement it.

x1ddos:
changed the part about nip-26 to:

> or a delegated keypair if NIP-26 is implemented

@Cameri if a relay doesn't implement nip-26, it shouldn't care about it at all.

Cameri:
> changed the part about nip-26 to:
> 
> > or a delegated keypair if NIP-26 is implemented
> 
> @Cameri if a relay doesn't implement nip-26, it shouldn't care about it at all.

exactly the reason why we should add this to NIP-26 instead

x1ddos:
ok, on a second thought maybe it is simpler. i removed nip-26 from here.

Cameri (approved):

PR #73, opened 2022-11-18
NIP-35 User Discovery

mikedilger:
Draft for comments.

It will need to be assigned a number before being merged.

### DIFF
--- a/35.md
+++ b/35.md
@@ -32,7 +32,7 @@ This NIP proposes an optional additional key like this:
   "names": {
     "bob": "b0635d6a9851d3aed0cd6c495b282167acf761729078d975fc341b22650b07b9"
   },
-  "relays" {
+  "relays": {
     "b0635d6a9851d3aed0cd6c495b282167acf761729078d975fc341b22650b07b9": [ "wss://relay.example.com", "wss://relay2.example.com" ]
   },
 }
--- a/35.md
+++ b/35.md
@@ -25,7 +25,7 @@ NIP-05 specifies a `nostr.json` file with contents like this (refer to NIP-05):
 }
 ```
 
-This NIP proposes an optional additional key lke this:
+This NIP proposes an optional additional key like this:
 
 ````
 {
--- a/35.md
+++ b/35.md
@@ -1,5 +1,5 @@
 
-NIP-NN
+NIP-35
 ======
 
 User Discovery
--- a/35.md
+++ b/35.md
@@ -32,7 +32,7 @@ This NIP proposes an optional additional key like this:
   "names": {
     "bob": "b0635d6a9851d3aed0cd6c495b282167acf761729078d975fc341b22650b07b9"
   },
-  "relays' {
+  "relays" {
     "b0635d6a9851d3aed0cd6c495b282167acf761729078d975fc341b22650b07b9": [ "wss://relay.example.com", "wss://relay2.example.com" ]
   },
 }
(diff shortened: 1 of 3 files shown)


fiatjaf:
Looks good to me.

I've consulted the oracle and she has decided that this should be NIP-35.

35.md
```diff
@@ -0,0 +1,41 @@
...
+```
+
+This NIP proposes an optional additional key like this:
+
+````
+{
+  "names": {
+    "bob": "b0635d6a9851d3aed0cd6c495b282167acf761729078d975fc341b22650b07b9"
+  },
+  "relays' {
```
jb55, line 35:
```suggestion
  "relays": {
```

jb55 (approved):
LGTM

35.md
mikedilger: I haven't used this "resolve conversation" before. I just pushed commits to change this (at first I thought it was just the quote)

mikedilger:
I just thought of something.  I could list somebody elses keys in my nostr.json, and suggest relays that they do not operate under.  A client might be misdirected to believe that and then this other person would effectively be silenced.

To prevent this kind of thing, here are some options
1. Sign the relay information with the public key, proving it was you who put it there,
2. Specify ways in which clients can avoid such tricks by considering the relay information only as hints, discarding it if a public key NIP-05 points to a different domain, and treating all in-protocol relay information as superseding these hints.

Thoughts?

eskema:
altough I find that having a signature might be good on some scenarios, the client needs to do the opposite workflow of what you're thinking. It searches for a nip-05 on the user and then goes and query that, so you know the user signed that nip-05 on it's kind-0 or because it just manually entered it. you don't go asking servers for user info.. does that make sense?

eskema:
with this we need to make 2 queries right? the first to get the pubkey, and the second to query the relays from the pubkey. or can we do it with in one go?

mikedilger:
Yes, the workflow you're describing @eskema of NIP-05 doesn't need a signature in nostr.json because it trusts the kind-0 data most and because it started there. It can just ignore the signature if it wants to (if we go with a signature).  The workflow I'm describing (start with nostr.json and discover users from it) would want that data signed, or at the very least should consider the data as advisory and not trusted.

The nostr.json file is returned with both parts.  If you are doing NIP-05, you just get the part for the user you queried, but also his/her relays.  If you are doing user discovery, you don't need the query string and you can get the whole file. Maybe that should be specified in this NIP.

jb55:
This is now implemented in damus-ios: 

http://git.jb55.com/damus/commit/749364189676eda0465b0a9799227b0fdf31d226.html#h1-5-67

and it's used to populate the bootstrap relays if logging in this way.

jb55:
> with this we need to make 2 queries right? the first to get the pubkey, and the second to query the relays from the pubkey. or can we do it in one go? thinking if there was a way to return a single package that contains whatever info is available for that key

you only need to do one. it's just an additional relays field in the json object that maps usernames to relays.

fiatjaf:
I think the signature thing is not necessary and too cumbersome to implement anyway.

All this relay information you may get for any pubkey from any source _must_ **always** be considered advisory and not trusted. Signatures do not change that because you may sign a thing today and get banned from the relay tomorrow.

fiatjaf (approved):

jb55:
https://jb55.com/.well-known/nostr.json

jb55:
I just realized mine is wrong and has usernames instead of pubkeys... fixing

fiatjaf:
https://gitlab.com/minds/engine/-/merge_requests/1182/diffs

PR #74, opened 2022-11-22
finalize some NIPs we will not change anymore

fiatjaf:


mikedilger:
What does final mean? That implementations can expect it will not change? That's my preference.  Or that we never change it at all? Some of these could use improvements in wording, can we tweak wording or add clarifications after it is final?

I'm so new to nostr that I haven't even really carefully read much less implemented some of these. So I should probably bow out of this review.  But I wanted to ask what final meant.

Giszmo (approved):
In the sense that the implementation details won't change, I agree with marking them final.

fiatjaf:
Yes, we can still change wording and so on, just not their essence.

PR #80, opened 2022-11-27
Update NIP-10: add 'mention' marker

monlovesmango:
I think that adding a mention marker would eliminate ambiguity for clients supporting both the deprecated and preferred conventions. I also think that this would allow for extensibility in adding new types of event mentions (for example if we want to add context for a note).

just thought I would throw this out there, let me know what you think.

### DIFF
--- a/10.md
+++ b/10.md
@@ -44,9 +44,9 @@ Where:
 
  * `<event-id>` is the id of the event being referenced.
  * `<relay-url>` is the URL of a recommended relay associated with the reference.  It is NOT optional.
- * `<marker>` is optional and if present is one of `"reply"` or `"root"`
+ * `<marker>` is optional and if present is one of `"reply"`, `"root"`, or `"mention"`
 
-**The order of marked "e" tags is not relevant.**  Those marked with `"reply"` denote the `<reply-id>`.  Those marked with `"root"` denote the root id of the reply thread.
+**The order of marked "e" tags is not relevant.**  Those marked with `"reply"` denote the `<reply-id>`.  Those marked with `"root"` denote the root id of the reply thread. Those marked `"mention"` denote id of events that are tagged and are not the `"reply"` nor `"root"`.
 
 >This scheme is preferred because it allows events to mention others without confusing them with `<reply-id>` or `<root-id>`.  
 
--- a/10.md
+++ b/10.md
@@ -46,7 +46,7 @@ Where:
  * `<relay-url>` is the URL of a recommended relay associated with the reference.  It is NOT optional.
  * `<marker>` is optional and if present is one of `"reply"`, `"root"`, or `"mention"`
 
-**The order of marked "e" tags is not relevant.**  Those marked with `"reply"` denote the `<reply-id>`.  Those marked with `"root"` denote the root id of the reply thread. Those marked `"mention"` denote id of events that are tagged and are not the `"reply"` nor `"root"`.
+**The order of marked "e" tags is not relevant.**  Those marked with `"reply"` denote the `<reply-id>`.  Those marked with `"root"` denote the root id of the reply thread. Those marked with `"mention"` denote a quoted or reposted event id.
 
 >This scheme is preferred because it allows events to mention others without confusing them with `<reply-id>` or `<root-id>`.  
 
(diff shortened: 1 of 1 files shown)


jb55:
fwiw damus ios's deprecated-nip10 mention parsing code always had a mention marker so this makes sense from the perspective of the damus ios' codebase.

fiatjaf:
Wait, when I first read this PR's title I thought `mention` was for the `p` tags like @eskema had suggested, not the `e` tags. I don't think it makes sense to include `e` tags for _all_ the events in a thread, because that will be a giant number of tags that will be unnecessary and confusing for the most part.

I can't prevent you from creating events with dozens of `e` tags, but I also don't want to encourage people to include that and I don't want newcomers to think this is _the proven way_ to deal with threads.

For me threads can be done with just two `e` tags: the "root" and the "reply".

eskema:
My suggestion is that ALL `p` and `e` tags be marked. `root`, `reply` or `mention`

monlovesmango:
I don't plan to have more than one 'reply' e tag. there may be multiple 'mention' e tags.

mostly want this for scenarios where there is no 'reply' or 'root' e tag, but there is a 'mention' e tag. having the marker 'mention' means there is no ambiguity in this scenario.

monlovesmango:
yes, the 'mention' marker would be for quoted or reposted events. 'mention' marker would not be used for constructing threads.

fiatjaf:
Then we should write that on the NIP.

fiatjaf:
Looks good to me now.

vishalxl:
Overall, this does solve the problem of ambiguity that it seeks to solve.

Also, you could add a line or two, explaining how many of root,reply and mention tags are recommended/acceptable for an event. Since there can be only one root, we should just mention that here, that there can't be more than 1 root tags. And also mention that since a reply can be made to only one event at a time, there can't be more than 1 reply tags. and that mentions can be more than one, but they should ideally be kept at minimum required.

monlovesmango:
I can add that. IMO it doesn't make sense to have both 'root' and 'reply' marker for top level reply. why would you tag the root event as a 'reply', when it is not a reply to anything?

vishalxl:
> I can add that. IMO it doesn't make sense to have both 'root' and 'reply' marker for top level reply. why would you tag the root event as a 'reply', when it is not a reply to anything?

yeah, its good to clarify it.

eskema (approved):
yes

offbyn:
I am confused why there could be multiple mentions, if so is the order important again?

>  and that mentions can be more than one, but they should ideally be kept at minimum required.

what is the usecase of more than 1 mention/quoted event?

just for clarification, by "quoted events" do you mean smth like this?

Screen Shot 2023-01-14 at 10 58 23 AM

monlovesmango:
yes that is an example of a quoted event but in twitter. use case for multiple mentions is when you want to embed multiple events in your note (like you can do on twitter)

vishalxl:
As I understand, in this mention system as proposed here, a single event can be mentioned by adding it to the tag, and making content something like #[0] where 0 would refer to the mentioned event. 

That is basically a repost of the mentioned event; then the [NIP 18](https://github.com/nostr-protocol/nips/blob/master/18.md)  which was included with pull request #140 becomes a duplicate of this feature; or they are duplicate features doing same things with two different ways. 

Just wanted to state this explicitly, in case I am missing something. 

As it is right now, NIP 18 is just duplicating this feature, so it should be deprecated as it is. No need to have two ways of doing same thing.

But Kind 6 can be improved; ( NIP 18) right now says the content field should be empty. If kind 6 content field could contain a whole event json as payload, then kind 6 could become a way to share events when someone is not able to find events from relays, which is going to happen a lot. 

Let say your follower can't see the event you reply to, they ask you for it, you reply with kind 6 with content set to that event.

eskema:
> But Kind 6 can be improved; ( NIP 18) right now says the content field should be empty. If kind 6 content field could contain a whole event json as payload, then kind 6 could become a way to share events when someone is not able to find events from relays, which is going to happen a lot.
> 
> Let say your follower can't see the event you reply to, they ask you for it, you reply with kind 6 with content set to that event.

i think it would be simpler to simply re-broadcast the original event to your relays where you post the mention / kind-6, so it's also there where they fetch your note referencing it. i get it that it's simple to just re-wrap it, but I feel that it hijacks the original post and keeps a discussion separate from it, which could be a good thing if that's what's expected, so you can talk about an event outside of it, but I don't think that that's the current intent when people use this feature now.

Cameri:
Quoting a note and repost are NOT the same.

fiatjaf:
Having these two things be different kinds is good for clients that want to filter out reposts from original notes of a pubkey, stuff like that.

PR #81, opened 2022-11-28
NIP-12: add 'delegation' exception and 'l' for language

monlovesmango:
two parts to this:
1. explicitly calling out 'delegation' as an exception to only single letter tags being queryable
2. adding use case of 'l' being used to identify language

I think #1 should definitely be added. happy to drop #2 if people disagree.


fiatjaf:
1. Why?

monlovesmango:
oops sorry, my bad. I guess Im still not thinking too clearly on delegation. but yeah, delegation is not a generic tag query and would be supported distinctly from this NIP.

Issue #83, opened 2022-11-30
NIP5: Add some field for specifying the kind of direct message 


arejula27:
The direct message is a way to intereact with other person, however it could be in many ways: direct chat, mail or whatever u want. I think it would be appropiated to add a field for specifying the kind o encrypted message it is, so if you are building a chat app u can filter this kind of message or if you are doing a mail app (my case) you can filter them.

arejula27:
Another option could be separate them on different event kinds, i dont think it is the best option because both are the same event (encrypted for 2 people), the only difference is the type of communication it is, one more instantaneous and other more polite.

arejula27:
In my opinion with a tag will be enough, however i wrote this issue for discuss a standardized tag, so every app could use it, not only mine.

eskema:
I don't underestand what you're trying to do, is this to know what is the user preferred way to get contacted? like kind-1 vs kind-4?

arejula27:
The idea is to specify the structure of the content, it could be a simple sentence or a JSON {title:string, text:string} or whatever the client prefer. So with this tag the clients would decode correctly.

arejula27:
Maybe my explanation was terrible at the begining :\

eskema:
encodings and structures are better suited for specific kinds, otherwise you pollute clients with unreadable data where it shouldn't. If you want a specific format, you write a new kind for it and use that instead of forcing clients to use your own.

arejula27:
Okay, how i select a new number? I will write a pr then

eskema:
write the nip without a number, the number will reveal itself after

PR #84, opened 2022-11-30
Create NIP for mail events

arejula27:
I write this NIP for creating a new kind of structure direct message as i was suggested in the  issue #83 . I make this proposal for discuss the best way of doing this structured events


eskema:
so the only difference with nip4 is the encrypted subject?

arejula27:
yes

eskema:
we already have a subject tag, why not simply add that but encrypted like the content, then you decrypt both, you don't need the structure..

arejula27:
The problem is that chat apps will show mail messages and the mail clients  wills how chat mssages. I feel that they are not the same, but if it is what i should do i will do it.

eskema:
> The problem is that chat apps will show mail messages and the mail clients wills how chat mssages. I feel that they are not the same, but if it is what i should do i will do it.

why are they not the same? what's the difference other than the UI abstraction?

arejula27:
for me these messages has not the same context, i dont want to receive on my gmail small messages as good morning, how are you etc. For that u use whatsapp or signal. A mail is used in other context, more polites maybe. For me is more than just a UI asbtraction, is a way of comunication, if they were the same people would just use gmail as a message app.

arejula27:
Another option will be have keys for chat apps and keys for mail apps, maybe it a bit annoying for users idk. What do u think?

eskema:
> for me these messages has not the same context, i dont want to receive on my gmail small messages as good morning, how are you etc. For that u use whatsapp or signal. A mail is used in other context, more polites maybe. For me is more than just a UI asbtraction, is a way of comunication, if they were the same people would just use gmail as a message app.

what prevents anyone from sending you events iun this format but like it was a chat? the format doesn't change anything if it doesn't impose a structure, you can also receive a bunch of these saying `hello sir`

arejula27:
Ofc the format didnt change anything, that the reason i ask for an extension of the nip 5. For setting the intention of the messages. People can send small mail for saying `hello sir` but they dont use to, the prefer other ways, thats the idea. Dont make a mess of messages with different scopes.

arejula27:
The main difference is not the structre, is the intention and the scope of the message, it is not a tech dif, is a language context diff.

eskema:
kind-4 already goes out of the way and makes it a direct communication avenue, it's exactly what you want. the intention is irrelevant if sending a kind-4 or kind-44 if the rest in unchanged, but that's just my perspective, feel free to use a different kind for that, but the key points of nostr is interoperability, so you can levarage existing apps that already talk the same language, if you're creating a new kind just for yourself, you can, but what would be the readon for anyoine else to use your different format if they can already talk to each other?

arejula27:
Sure, i understand your point and see the advantages of it. We can let this pr open for a week to see more peolpe opinions and then we can close it.

arejula27:
I will also ask on anigma, but i thinking your point is really good

arejula27:
I make some updates on the nip, i think now is different from the dm, cuz it has more functionalities (set category, mark if read, etc) If someone want to add something will be helpful.

arejula27:
Im also valuing using shared secrets  #59. The idea would be have a shared secret or group chat as mail thread, but maybe a user will have a lot of keys to manage.

jb55:
something I've always wanted is a more general nip04 that has encrypted tags and maybe even encrypted kinds. you could even retain queryability by hashing the shared secret into the tag value: {kinds: [444], #k: sha256(shared_secret, 7)}  that way you could create reactions to dms, etc.

arejula27:
I like the idea

fiatjaf:
Maybe add email-specific metadata (`subject`, for example) as a tag, then your email client will only show messages containing that?

arejula27:
That was mi initial idea with issue #83, I propose to set a few tags for specifying some kind of encrypted messages. I explained the idea  bad :(

jb55:
> something I've always wanted is a more general nip04 that has encrypted tags and maybe even encrypted kinds. you could even retain queryability by hashing the shared secret into the tag value: {kinds: [444], #k: sha256(shared_secret, 7)} that way you could create reactions to dms, etc.

even more general: 

  `{kinds:[444], #g: [sha256(shared_secret + "kinds" + 4), sha256(shared_secret + "#p" + "pubkey")]}`

but this doesn't work because it is eqivalent to `[REQ,"",{#p: pubkey}, {kinds: [4]}]`, which is an OR query and not what we want. It would be nice if we had a way to do AND in generic tag queries.

jb55:
waiting for @Semisol to take this idea https://github.com/nostr-protocol/nips/pull/84#issuecomment-1332799070 and make it work

arejula27:
I think I will be implement the client as kind 4, cuz every one looks agree with that. 
 @Cameri the pubkey on the content is redundant, there is a reason for it or just following the format? Do you think is convenient?

PR #86, opened 2022-12-03
docs: add NIP-xx reposts

cameri:
We've had kind 6 reposts for a while. Let's make it official.

View here: https://github.com/nostr-protocol/nips/blob/c875b97c60d39534d343deb4b230e1c92601c804/39.md

### DIFF
--- a/39.md
+++ b/39.md
@@ -2,23 +2,22 @@
 NIP-39
 ======
 
-Reposts
--------
+Boosts
+------
 
 `draft` `optional` `author:cameri` `author:jb55`
 
-A repost is a `kind 6` note that is used to share other notes.
+A boost is a `kind 6` note that is used to refer to other notes.
 
 The `content` SHOULD be an empty string.
 
 Tags
 ----
 
-The repost event SHOULD include `e` and `p` tags from the note the user is
-sharing. This allows users to be notified when their notes are reposted.
-Including the `e` tags enables clients to pull all the reposts associated
-with individual posts or all the posts in a thread.
+The boost event SHOULD include `e` and `p` tags referencing the boosted note
+and the pubkey of the boosted note, respectively.
+This allows users to be notified when their notes are boosted.
+Including the `e` tags enables clients to pull all the boosts associated
+with individual notes.
 
-The last `e` tag MUST be the `id` of the note that is being reposted.
-
-The last `p` tag MUST be the `pubkey` of the event being reposted.
+The tags should be marked with `boost` as explained in [NIP-10](10.md).
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ NIPs stand for **Nostr Implementation Possibilities**. They exist to document wh
 - [NIP-28: Public Chat](28.md)
 - [NIP-35: User Discovery](35.md)
 - [NIP-36: Sensitive Content](36.md)
-- [NIP-39: Reposts](39.md)
+- [NIP-39: Boosts](39.md)
 
 ## Event Kinds
 
@@ -37,7 +37,7 @@ NIPs stand for **Nostr Implementation Possibilities**. They exist to document wh
 | 3           | Contacts                    | [2](02.md)             |
 | 4           | Encrypted Direct Messages   | [4](04.md)             |
 | 5           | Event Deletion              | [9](09.md)             |
-| 6           | Reposts                     | [39](39.md)            |
+| 6           | Boosts                      | [39](39.md)            |
 | 7           | Reaction                    | [25](25.md)            |
 | 40          | Channel Creation            | [28](28.md)            |
 | 41          | Channel Metadata            | [28](28.md)            |


fiatjaf:
Should we really make the "boost" name official? I think maybe call it by a more generic name like "repost" (as you had before) and let clients call these boosts or anything else.

eskema:
honestly, I think kind 6 is not needed. we could achieve the same using only a tag `mention` in a kind1 with an empty string, and we could also write something if we wanted (repost with comment), tagging the original pubkey would be optional as you could simply repost to your audience without notifying the op (if desired). 
If we really need this kind, then an option to also comment on the boost would be good to have.

eskema:
my reasoning is that reposts are at the essence, the same as sharing a webpage (url), but when we do that, we use kind-1, but for sharing nostr content we use a different kind. doesn't make much sense to me..

fiatjaf:
I think the rationale for a new kind should be: what will a client that doesn't implement this do?

If we make a "repost" event as `{ kind=1 tag=[e:abcdef] content=<blank> }` what will the nonimplementing client do? It show a blank and possibly broken event that does nothing and appears as a reply to the original event.

If we make it as `{ kind=6 tag=[e:abcdef] content=<blank> }` then it just won't show for the nonimplementing client.

fiatjaf:
Using the same rationale above, the NIP-10 `"mention"` tags should be eliminated and included in this NIP and use kind 6, that would make more sense.

fiatjaf:
Another (relatively weak but still important I believe) argument in defense of a new kind is: when a client is displaying a note `A` it will query relays for all events that have `#e = A, kind = 1` expecting to get only events that are replying to `A`, but if we're using kind 1 for reposts that query will return irrelevant results.

(Of course if a client wants to get all the reposts for a given event it can still use `kind = 6` in the query.)

fiatjaf:
> Embedding the original post means the repost stands on it's own

I don't like this, by this rationale we should also send the original event whenever we're replying to something, so our reply stands on its own -- but we don't do that.

More efficient and more "nostric" solutions to me are: to send the event you're reposting to your relays; and/or to include the address of the relay where you found it in your repost event tags, so receivers can know for sure where to grab the original event from.

fiatjaf:
> Do we want interoperability with reposts/boosts or should clients play it by ear?

We definitely want interoperability! @jb55 has been playing with kind 6 for a while and I've been waiting for him to make a NIP. I was happy to learn that he gave up embedding the full original event in the content and decided to just send tags. That is also extra good because it opens space for the reposter to attach its own comment as the content.

eskema:
> Another (relatively weak but still important I believe) argument in defense of a new kind is: when a client is displaying a note `A` it will query relays for all events that have `#e = A, kind = 1` expecting to get only events that are replying to `A`, but if we're using kind 1 for reposts that query will return irrelevant results.
> 
> (Of course if a client wants to get all the reposts for a given event it can still use `kind = 6` in the query.)

in this case, it would also return mentioned events, that also fall into the irrelevant category no? so you want to eliminate mentions altogether? I don't quite follow that part.. I think maybe a new tag ['share', 'url'] could be used, and that could also include nostr events if we do it `["share", "nostr:npub543734abcf"]`, `["share", "nostr:e:543734abcf"]` or `["share", "https://twitter.com/post/whatever"]`. dunno, I'm fine either way...

fiatjaf:
> in this case, it would also return mentioned events, that also fall into the irrelevant category no?

Yes, see: https://github.com/nostr-protocol/nips/pull/86#issuecomment-1336188686, I think it's better to do the mentions using kind 6. (But I'm just thinking this live and hoping my thoughts make sense.)

>  I feel like if kind-1 is not good to share nostr but is good to share non nostr content, something is not right.

I think I agree with this but at the same time I don't agree. When we share URLs and images we post a link to them, so it is just text and anyone can click on the link, but when we share a Nostr event how can we do that?

eskema:
> I think I agree with this but at the same time I don't agree. When we share URLs and images we post a link to them, so it is just text and anyone can click on the link, but when we share a Nostr event how can we do that?

I think this is what we should solve, because it would also solve more things. the talk we already keep having about the nostr uri scheme should be brought into this IMO, because it should be as simple as pasting a link to share nostr content..

eskema:
but yes, let's make the kind-6 official and we'll solve this other thing later... 
so kind-6 is share nostr event, with or without comment as content, tag with e and p

fiatjaf:
> I think I agree with this but at the same time I don't agree. When we share URLs and images we post a link to them, so it is just text and anyone can click on the link, but when we share a Nostr event how can we do that?

Oh, wait, I get this now. We just need to share a link to a Nostr event like you said. No need for a tag -- although the client _may_ automatically include a tag.

This is probably the most backwards-compatible way, but we need to develop a standard for Nostr event URIs first.

@Semisol @jb55 we need the `nshare` / `nevent` thing here now!

If we have a standard for a Nostr URL like `nostr:nevent1abc...` thing it can include a relay URL of the event being shared and we can just include that anywhere in the `content` and supporting clients can read that and replace it with a preview of the event being shared. This seems to be exactly what Twitter does too. Nonsupporting clients will the full `nostr:...` URI and users will have to resolve that manually. Partially-supporting clients can easily be made to recognize `nostr:...` and turn those into clickable `<a href=...>` links so they can be resolved by some registered link handler or even resolved by the app itself.

eskema:
for example, take this event from astral.ninja: fc66a0d6d67f8b12bd6a404d613e00e2053cec849872f57372efcd7ff37cf1dd

it is basically a kind-6, but better because every client that reads kind-1 should display these as well. 
Now, even better would be if we took that content out of context but also retained a link that even something outside of nostr would recognize, like our brand new soon to be nostr uri scheme.

monlovesmango:
hey all, was holding off on commenting to see what others thoughts are and I don't really feel strongly one way or the other. I personally lean towards kind 6 being redundant (as we can easily accomplish this with kind 1 and a 'mention' marker as @eskema said). however I also see some benefit in making these their own kind.

one part I am unclear on. for any event that is a reply and also mentions another event, should this also be kind 6? if so, would we still use the 'root' and 'reply' markers for kind 6? and any 'e' tag without a marker should be assumed to be the mentioned events?

overall I would be happy with whatever is decided around this. I also think we should call this something more generic (like mentions? reposts?) rather than boosts.

monlovesmango:
I really like this, as it gives us a dedicated event kind to use for automatically embedding nostr content.

So can you confirm that kind 6 can also use the 'root' and 'reply' e tag markers (for when you want to reply to a thread, but also want to mention a separate event unrelated to the thread)?

maybe we can name it 'embedded' and have this NIP be for embedded nostr content generally? seems like this will expand to cover nostr uri scheme from the discussion.

fiatjaf:
> - For kind 1, parse any Nostr URIs (nostr:... To be determined) into clickable links. If there's any mention tag, show it as a link as well. Do not automatically embed.
> - For kind 6, use the mention tag to embed the mentioned event. The content field is a comment by the user reposting.

This is good but the problem is that it is 3 ways of achieving the same result. Can we reduce that to at least 2?

fiatjaf:
Yes, I like that more.

I think we should also specify that kind 6 reposts are not expected to appear as replies to other events.

monlovesmango:
what about replies that are embedding/reposting a different and unrelated event (NOT the event/thread they are replying to)? this should still be kind 6?

eskema:
the more I think about reposts the more it makes sense them to be simply a kind-1 and we define a new tag ["re","event_id"], this way it doesn't confuse and doesn't nest in threads unless when replying which would work fine.. you could easily query them using the tag query "#re"

PR #87, opened 2022-12-04
NIP-40 Expiration Timestamp

0xtlt:


fiatjaf:
Looks good to me.

eskema (approved):
I like this. I think we can expand it later with other rune like conditions for the expiration but for now it's perfect.

fiatjaf:
@eskema What would be these other conditions? I can't imagine anything.

fiatjaf:
Unrelated question: why are you prefixing these PRs with "docs: "? That makes no sense to me. NIPs are not docs, they are specs, and there is nothing else in this repo besides them.

eskema:
> 

I don't know, can't think of anything on the top of my head that would make much sense, but just because an expiration could have a timestamp as last resort but could be trigerred by something else before that.. but ignore that line, this is perfect as is.

0xtlt:
Just a mistake with my autocomplete 😅

40.md
```diff
@@ -0,0 +1,46 @@
+NIP-40
+======
+
+Expiration Timestamp
+-----------------------------------
+
+`draft` `optional` `author:0xtlt`
+
+The `expiration` tag enables users to specify a timestamp at which the message should be considered expired and deleted by relays. This helps ensure that messages are not retained by the network after they are no longer relevant, which helps maintain the privacy of the messages. This is especially useful in cases where a replacement event is not available or feasible.
```
Giszmo, line 9: I would not justify this nip with privacy. We all know that events once broadcast will get stored for eternity.

```suggestion
The `expiration` tag enables users to specify a timestamp at which the message should be considered expired and deleted by relays. This helps reduce the burden of the network around events with relevance limited in time. This is especially useful in cases where a replacement event is not available or feasible.
```

40.md
```diff
@@ -0,0 +1,46 @@
...
+
+#### Spec
+
+```
+tag: expiration
+options:
+ - [timestamp]: required
```
Giszmo, line 16:
```suggestion
 - [UNIX timestamp in seconds]: required
```

40.md
```diff
@@ -0,0 +1,46 @@
...
+    "id": "<event-id>"
+}
+```
+
+Note: The timestamp should be in the same format as the created_at timestamp and should be interpreted as the time at which the message should be deleted by relays.
+
+Client Behavior
+---------------
+
+Clients SHOULD use the `supported_nips` field to learn if a relay supports this NIP. Clients SHOULD NOT send expiration events to relays that do not support this NIP.
```
Giszmo, line 39: Relays oblivious to this nip will not care about such desires and just mirror these events. A nip can only impose stuff on software supporting the nip, not others. I wouldn't even imply there were any effort to keep these events away from relays that don't support the nip.

40.md
```diff
@@ -0,0 +1,46 @@
...
+
+Suggested Use Cases
+-------------------
+
+* Temporary announcements - This tag can be used to make temporary announcements that are automatically deleted from the network after a certain amount of time. For example, an event organizer could use this tag to post announcements about an upcoming event, and the announcements will be automatically deleted once the event is over.
+* Limited-time offers - This tag can be used by businesses to make limited-time offers that expire after a certain amount of time. For example, a business could use this tag to make a special offer that is only available for a limited time, and the offer will be automatically deleted from the network once the time has passed.
+* Ephemeral messages - This tag can be used by individuals to send ephemeral messages that are automatically deleted from the network after a certain amount of time. For example, someone could use this tag to send a message that is only visible to the recipient for a short time, and the message will be automatically deleted once the time has passed. This can be useful for sending sensitive information that should not be retained on the network indefinitely.
```
Giszmo, line 46: Don't broadcast sensitive information! Ever!

40.md
0xtlt: Was already "implemented" for nip16 : https://github.com/nostr-protocol/nips/blob/master/16.md

But I understand what you have in mind.

It's not a requirement but a recommendation here

40.md
0xtlt: So I am also responding here with the example of a signal-like application with custom relays

ok300:
> The `expiration` tag enables users to specify a timestamp at which the message should be considered expired and deleted by relays.

I can see the usefulness of this, but:

First, for this to work, relays (and clients) need to have NTP. Otherwise things can get messy and out of sync due to time drift, especially at a granularity of seconds. Since there's no way to enforce that, some relays could advertise NIP-40 but do it wrong, intentionally or not. This is very hard to detect and troubleshoot.

Second: "should be ... deleted by relays" implies that relays that follow the spec *will* delete these messages. This makes it seem like timed deletion is a primitive clients can reasonably rely on. This is dangerous, because even if a relay would want to do it, a simple NTP outage or misconfiguration can lead to unexpected event deletion (or non-deletion when deletion is expected, depending on the direction of the time drift).

AFAIK the protocol makes no guarantees about the accuracy or meaning of `created_at` (is it client time? relay time? neither, it's a client-chosen timestamp mostly close to UTC because of mostly well-behaved and NTP-enabled clients). Timestamping is generally a problem in distributed decentralized systems. It therefore seems reasonable the spec shouldn't prescribe any action, like message deletion, based on a timestamp.

---

IMO the time drift issue can be avoided by changing from an absolute `timestamp` to a relative `TTL`. The deletion primitive issue can be solved by changing from _"..should be deleted.."_ to a softer _"..clients may choose to hide it"_.

> The `TTL` tag indicates a duration for which this event can be considered fresh. The duration is in seconds and relative to the `created_at` of the event. When `created_at + TTL` is reached as per the client's time, the event can be considered expired and the client may choose to ignore it or hide it.

This bypasses any dependency on NTP because it's relative to a timestamp already embedded in the message.

It also doesn't create any false expectation that messages are deleted, while preserving the benefit of "expirable" messages.

40.md
Giszmo: In a group chat, in a DM, don't ever send information you require to not be available after a certain expiry date. It simply can't be enforced and Signal pretending it could is a disservice. With the very varied client landscape in nostr there is zero chance of all client agreeing to delete expired events. If information being linkable to your account after time X is a problem, don't sign and send it before time X neither.

Giszmo:
I mostly agree with @ok300 but instead of trying to fix it by prescribing clock synchronization I would just soften the promises made by the nip. NTP-enabled is a double-edged sword as NTP can also be manipulated to cause disruption. More or less accurate local clocks, some of which do randomly use NTP are certainly more decentralized.

ok300:
Where did I prescribe clock synchronization? :) My suggestion was entirely around how to avoid it.

40.md
mikedilger: I concur with Giszmo and ok300. I cannot imagine how nostr could create any reasonable expectation of privacy outside of anonymity. Once a post is sent relays can do anything with it, and clients can cache it forever (as mine currently does).

But i ♥ this nip for a different reason: relays will fill up if they don't eventually drop irrelevant events. I've seen relay code that just drops everything after a certain time period (a relay that I cannot use for a blog), and this NIP offers a useful hint to the relay.  I would expire my 'tweets' after a month, and expire my blog posts after 3 years, for example.

0xtlt:
I have read all your feedback, I have updated the NIP with it accordingly.
Let me know what you think.

40.md
```diff
@@ -0,0 +1,53 @@
...
+
+Clients SHOULD use the `supported_nips` field to learn if a relay supports this NIP. Clients SHOULD NOT send expiration events to relays that do not support this NIP.
+
+Clients MUST ignore events that have expired.
+
+Suggested Use Cases
+-------------------
+
+* Temporary announcements - This tag can be used to make temporary announcements that are automatically deleted from the network after a certain amount of time. For example, an event organizer could use this tag to post announcements about an upcoming event, and the announcements will be automatically deleted once the event is over.
+* Limited-time offers - This tag can be used by businesses to make limited-time offers that expire after a certain amount of time. For example, a business could use this tag to make a special offer that is only available for a limited time, and the offer will be automatically deleted from the network once the time has passed.
```
ok300, line 48:
```suggestion
* Limited-time offers - This tag can be used by businesses to make limited-time offers that expire after a certain amount of time. For example, a business could use this tag to make a special offer that is only available for a limited time.
```

40.md
```diff
@@ -0,0 +1,53 @@
...
+* Ephemeral messages - This tag can be used by individuals to send ephemeral messages that are automatically deleted from the network after a certain amount of time. For example, someone could use this tag to send a message that is only visible to the recipient for a short time, and the message will be automatically deleted by relays or/and ignored by clients once the time has passed. This can be useful for sensitive information that should not be retained on the network indefinitely.
+
+#### One last time, be careful!
+The events could be downloaded by third parties as they are publicly accessible all the time on the relays.
+So don't consider ephemeral messages as a security for your conversations or other uses.
```
ok300, line 53:
```suggestion
So don't consider expiring messages as a security feature for your conversations or other uses.
```

ok300 (commented):
Added some comments around simplifying the field, and clarifications on client and relay behavior.

40.md
```diff
@@ -0,0 +1,53 @@
...
+
+Note: The timestamp should be in the same format as the created_at timestamp and should be interpreted as the time at which the message should be deleted by relays.
+
+Client Behavior
+---------------
+
+Clients SHOULD use the `supported_nips` field to learn if a relay supports this NIP. Clients SHOULD NOT send expiration events to relays that do not support this NIP.
+
+Clients MUST ignore events that have expired.
+
```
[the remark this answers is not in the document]
0xtlt: As the expiration time must be at UTC-0, the local times is not a problem here

40.md
```diff
@@ -0,0 +1,53 @@
...
+    "id": "<event-id>"
+}
+```
+
+Note: The timestamp should be in the same format as the created_at timestamp and should be interpreted as the time at which the message should be deleted by relays.
+
+Client Behavior
+---------------
+
+Clients SHOULD use the `supported_nips` field to learn if a relay supports this NIP. Clients SHOULD NOT send expiration events to relays that do not support this NIP.
```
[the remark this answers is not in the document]
0xtlt: Even if the relay can store the message indefinitely and/or transfer it to relays that store it infinitely, I prefer to leave this line in the PIN.

This will not be the first nip to include a rule like this

40.md
```diff
@@ -0,0 +1,53 @@
...
+#### Example
+
+```json
+{
+    "pubkey": "<pub-key>",
+    "created_at": 1000000000,
+    "kind": 1,
+    "tags": [
+      ["expiration", 1600000000]
+    ],
```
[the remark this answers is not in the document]
0xtlt: Good idea the ttl, however, the advantage of the timestamp here is that the relay can easily locate and delete / not send or treat differently a message with an outdated timestamp.
It is also possible to do this by adding the timestamp of the creation date + the ttl but I am for a Nostr philosophy where relays are very simple programs and clients are explicit in their requests.

It would avoid a relay to make a big loop + calculation in the messages to delete them for example.

40.md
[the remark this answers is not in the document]
ok300: @x1ddos said
> As the expiration time must be at UTC-0, the local times is not a problem here

There is no such thing as UTC-0 in a decentralized network.

Services that rely on an accurate time, like 2FA or TOTP, can run into issues because of time drift:

* https://blog.gaborszathmari.me/tricking-google-authenticator-totp-with-ntp/
* https://www.reddit.com/r/synology/comments/f9lu53/2fa_and_ntp_woes/
* https://forum.matomo.org/t/2fa-authentication-suddenly-fails/45829

Basing message deletion on that might be asking for trouble :)

ok300:
@x1ddos said
> As the expiration time must be at UTC-0, the local times is not a problem here

There is no such thing as UTC-0 in a decentralized network.

Services that rely on an accurate time, like 2FA or TOTP, can run into issues because of time drift:

* https://blog.gaborszathmari.me/tricking-google-authenticator-totp-with-ntp/
* https://www.reddit.com/r/synology/comments/f9lu53/2fa_and_ntp_woes/
* https://forum.matomo.org/t/2fa-authentication-suddenly-fails/45829

Basing message deletion on that might be asking for trouble :)

mikedilger:
I offer my suggestions here: https://github.com/mikedilger/nips/tree/nip40

Among the offered changes:

- I removed the initial motivating sentences, making it more declarative, considering the Suggested Use Cases covers that better.
- I changed "clients MUST ignore" to "clients SHOULD ignore".  I think MUST should be reserved for things that, if you don't do it, causes incompatibility problems.  Not ignoring an expired message doesn't cause incompatibility problems.

mikedilger:
On my tree I also took out the NIP-11 check. Relays that don't support this tag won't delete the event, but they can still relay the event and client's can honor the tag.

Once you send out an event you should think of it as permanent, and this tag as 'advisory' because you can't control other people/nodes/relays. I just don't see the benefit of checking NIP-11.

0xtlt:
> @x1ddos said
> 
> > As the expiration time must be at UTC-0, the local times is not a problem here
> 
> 
> 
> There is no such thing as UTC-0 in a decentralized network.
> 
> 
> 
> Services that rely on an accurate time, like 2FA or TOTP, can run into issues because of time drift:
> 
> 
> 
> * https://blog.gaborszathmari.me/tricking-google-authenticator-totp-with-ntp/
> 
> * https://www.reddit.com/r/synology/comments/f9lu53/2fa_and_ntp_woes/
> 
> * https://forum.matomo.org/t/2fa-authentication-suddenly-fails/45829
> 
> 
> 
> Basing message deletion on that might be asking for trouble :)

As the nip has gone from an « obligation » to a non-mandatory recommendation, this should no longer be a problem now

Any idea would still be welcome

0xtlt:
> I offer my suggestions here: https://github.com/mikedilger/nips/tree/nip40
> 
> Among the offered changes:
> 
> * I removed the initial motivating sentences, making it more declarative, considering the Suggested Use Cases covers that better.
> * I changed "clients MUST ignore" to "clients SHOULD ignore".  I think MUST should be reserved for things that, if you don't do it, causes incompatibility problems.  Not ignoring an expired message doesn't cause incompatibility problems.

I agree with both points, I have changed accordingly

0xtlt:
> On my tree I also took out the NIP-11 check. Relays that don't support this tag won't delete the event, but they can still relay the event and client's can honor the tag.
> 
> Once you send out an event you should think of it as permanent, and this tag as 'advisory' because you can't control other people/nodes/relays. I just don't see the benefit of checking NIP-11.

I agree on the "consider the message to be permanent" aspect, but that should be a recommendation.

in this case, I use "SHOULD" to define the recommendation.

Not an obligation here, and it seems important to me to keep this recommendation.

0xtlt:
Good ideas, now updated @Semisol

40.md
```diff
@@ -0,0 +1,59 @@
...
+
+Clients SHOULD ignore events that have expired.
+
+Relay Behavior
+--------------
+
+Relays MAY NOT delete an expired message immediately on expiration and MAY persist them indefinitely.  
+Relays SHOULD NOT send expired events to clients, even if they are stored.  
+Relays SHOULD drop any events that are published to them if they are expired.  
+Relays SHOULD persist ephemeral events until their expiration timestamp if an ephemeral event has one, and relays MAY limit the amount of time an ephemeral event is persisted.
```
Giszmo, line 49: Ephemeral events currently do not get persisted. They get relaid to interested parties and dropped. With this change, they turn into non-ephemeral events. I'm against expiration for ephemeral events. If anything I would make this explicit:

```suggestion
An expiration timestamp does not affect storage of ephemeral events.
```

Giszmo (approved):

x1ddos:
> @x1ddos said
> 
> > As the expiration time must be at UTC-0, the local times is not a problem here
> 
> There is no such thing as UTC-0 in a decentralized network.

@ok300 that was someone else :)
but yes, time drifts are always present.

image

on a slightly different topic, i'm not sure what the meaning of "MAY NOT" in here is:

> Relays MAY NOT delete an expired message immediately on expiration and MAY persist them indefinitely.

are the keywords in uppercase based on some rfc? i know only about "MAY" from rfc2119. the "MAY NOT" reads to me like "MUST NOT" or "SHALL NOT" as in, prohibited. quite strong and i wonder why? also, confusing when reading together with the next line:

> Relays SHOULD drop any events that are published to them if they are expired.

is "drop" the same as "delete" or something else? if same meaning, it feels like the two contradict each other: first one prohibits deletion, the second recommends it. unless of course "immediately on expiration" and "if they are expired" are a hint on a time drift.

i'd recommend clarifying what "MAY NOT" means exactly, and the difference between "immediately expired" vs ...not "immediately"?

mleku:
resurrecting this one because not having this query be accessible through a standard filter query is annoying for forcing the extension of an event store API in the case of writing event stores that use other event stores

it should be `x` IMO

in the meantime, i will be adding an API function for finding and returning these events that won't be accessible via a filter query but just want to necrobump this because it should be discussed... this is a very useful feature that i think should be commonly implemented, and moved from "optional" to "recommended" or whatever the in between to "mandatory" is

PR #92, opened 2022-12-15
NIP-41: Authentication

Semisol:


41.md
```diff
@@ -0,0 +1,69 @@
...
+
+Relays SHOULD send [`EOSE`](https://github.com/nostr-protocol/nips/blob/master/15.md) when an authentication request is triggered by a `REQ` command, and not send stored events after the `EOSE` when authentication is completed.
+Relays SHOULD send `OK` as a response when a command triggers authentication with the reason starting with `auth:`.
+
+Clients SHOULD retry the action (resending event, resubscribing) after they authenticate if they receive an `AUTH` request.
+
+## Signed event verification
+Relays when processing `AUTH` responses SHOULD verify the following before accepting the response:
+- that the `kind` is `22223`
+- that the event was recently signed (~10 minutes, by `created_at`)
```
earonesty, line 68: note: this requires that the client and the server use time sync of some kind

fernandolguevara:
auth can be handled at HTTP before upgrades the connection to ws

41.md
gkbrk: The server can send its timestamp inside the `<data object>`, so the client knows what created_at values it signs will be considered valid.

gkbrk:
Should this be a separate event kind? If there is any relay recommendations, chat or tweeting going on at the Ephemeral Events Reserved kind 22241, that can potentially be used for authenticating to relays. I think either this event number should be excluded from the Ephemeral Events Reserved range, or this NIP should use a different kind.

Example chat with kind=22241 events.

```
A: Hey man, which relay do you use these days?
B: wss://relay.example.com/
A: Thanks.
```

Now we have a valid AUTH signature from B that can be used for 10 minutes.

41.md
```diff
@@ -0,0 +1,71 @@
...
+`data object` MUST be a JSON object. It currently has one defined field, but may be extended by amendments to this NIP or other NIPs:
+- `subscription_id`: The subscription ID that triggered the `AUTH` request.
+
+A client may send one of the following to the relay:
+- `["AUTH", <signed event>]` to indicate it has accepted the request. This may also be sent without an authentication request.
+- `["AUTH", null]` to indicate it has rejected the request.
+
+A relay SHOULD send the [`OK`](https://github.com/nostr-protocol/nips/blob/master/20.md) command after they receive a
+non-rejecting authentication response, and use one of the following `message` prefixes if the event sent cannot be verified:
+- `too-old:`: The event signed has a too old `created_at` date.
```
gkbrk, line 55: Should a relay accept events that are signed too much into the future? Maybe -/+ 10 minutes instead of just allowing up to 10 minutes in the past.

41.md
```diff
@@ -0,0 +1,71 @@
...
+  This can be used for where only the signer of an event can publish it, or a pay-as-you-go relay allowing for you to publish others' events 
+- `private`: The client has attempted to access a restricted set of events (example: kind 4 DMs) and should authenticate with the relay to receive them.
+- `other`: Any reason not defined here.
+
+`data object` MUST be a JSON object. It currently has one defined field, but may be extended by amendments to this NIP or other NIPs:
+- `subscription_id`: The subscription ID that triggered the `AUTH` request.
+
+A client may send one of the following to the relay:
+- `["AUTH", <signed event>]` to indicate it has accepted the request. This may also be sent without an authentication request.
+- `["AUTH", null]` to indicate it has rejected the request.
```
fiatjaf, line 51: Can we remove this? It's an unnecessary complication. Clients gain nothing from sending this, and relays gain nothing from expecting it.

41.md
```diff
@@ -0,0 +1,71 @@
...
+A relay SHOULD send the [`OK`](https://github.com/nostr-protocol/nips/blob/master/20.md) command after they receive a
+non-rejecting authentication response, and use one of the following `message` prefixes if the event sent cannot be verified:
+- `too-old:`: The event signed has a too old `created_at` date.
+- `invalid-url:`: The URL in `content` is not matching.
+- `already-used:`: This event was already used to authenticate to this relay.
+- `bad-signature:`: The event has a bad signature.
+
+Please note that the `OK` message should only be sent as a response to other commands that the client sends instead of the `AUTH` command if the issue is not related to the authentication event being incorrectly signed (example: not on whitelist).
+
+Relays SHOULD send [`EOSE`](https://github.com/nostr-protocol/nips/blob/master/15.md) when an authentication request is triggered by a `REQ` command, and not send stored events after the `EOSE` when authentication is completed.
```
fiatjaf, line 62: I don't understand this.

41.md
```diff
@@ -0,0 +1,71 @@
...
+- `["AUTH", null]` to indicate it has rejected the request.
+
+A relay SHOULD send the [`OK`](https://github.com/nostr-protocol/nips/blob/master/20.md) command after they receive a
+non-rejecting authentication response, and use one of the following `message` prefixes if the event sent cannot be verified:
+- `too-old:`: The event signed has a too old `created_at` date.
+- `invalid-url:`: The URL in `content` is not matching.
+- `already-used:`: This event was already used to authenticate to this relay.
+- `bad-signature:`: The event has a bad signature.
+
+Please note that the `OK` message should only be sent as a response to other commands that the client sends instead of the `AUTH` command if the issue is not related to the authentication event being incorrectly signed (example: not on whitelist).
```
fiatjaf, line 60: I don't get the `OK` stuff. NIP-20 `OK` messages have an event id in them. Here there is no event id to use. I think this entire thing will be very much ok without the `OK`s. Much simpler to understand and implement. The client knows when it is signing stuff correctly, it doesn't need the relay to tell them.

If it is signing stuff wrong it doesn't have to handle that error message, it must necessarily fix the code that is producing bad signatures. Once it is fixed, the signatures will be good forever. There is nothing for the user to do here at runtime.

fiatjaf:
I may be missing something, but to me the flow to me should be:

1. If the client knows the relay requires auth already, so it sends an auth request before doing anything and then does whatever it wants to do. The websocket messages ordered, so it knows the relay will receive his `AUTH` before the other commands.
2. If the client doesn't know and tries to do something it can't, the relay can send an auth request, then the client can send an `AUTH` or not. At this point the client doesn't know for sure _what_ triggered the auth request, so it will probably have to restart all its subscriptions or do something else. In either case the `OK` or `EOSE` messages do not help very much, so it is better to eliminate them and then this NIP becomes super simple.

41.md
```diff
@@ -0,0 +1,71 @@
+NIP-41
+======
+
+Authentication of clients to relays
```
Giszmo, line 4: The authentication is not of a client but of a pubkey.
The nip as is contains a commitment to the relay and the pubkey but not to the client.

In theory, a proxy that gets to read the communication could take over the authenticated connection and run queries using the authentication.

Alice (Client) <--WS--> Eve (proxy) <--WS--> Bob (Relay)

Should we make sure of end-to-end encryption before trusting anything that reaches the relay via this websocket?

```suggestion
Pubkey authentication to relays
```

41.md
```diff
@@ -0,0 +1,71 @@
...
+
+This NIP defines a way for clients to authenticate to relays by signing an ephemeral event.
+
+## Event format
+
+An event should be signed with `kind: 22241` and `content` being the WebSockets URL of the relay.
+The URL MUST:
+- Have a trailing slash if the path is `/`
+- Not have a port number if protocol is `ws` and port is `80` or protocol is `wss` and port is `443`
+- Not include the search/query
```
Giszmo, line 17: This part is not clear. What would the "search/query" even be in this context? Does this line really resolve any real issue?

41.md
```diff
@@ -0,0 +1,71 @@
...
+
+Relays SHOULD treat authenticaiton events with a valid delegation as if it was the delegator authenticating.
```
Giszmo, line 19: typo

```suggestion
Relays SHOULD treat authentication events with a valid delegation as if it was the delegator authenticating.
```

41.md
fiatjaf: Query string.

Giszmo: "querystring" as in `wss://some.relay.com/?query=string` ? Is that a thing? How about `wss://some.relay.com/#tag=string` ?

fiatjaf: Yes, why not? I think relay URLs with querystring arguments can be used in some interesting ways. The # stuff I don't know. Is that sent to the server? I thought it was a client thing.

Giszmo: Browsers interpret `#something` client side, only but of course, links can contain them, maybe to signal nip support: `wss://some.relay.nostr/#nips=1,2,3,4`

To identify towards the relay, it should be omitted for normalization.

On the other hand, `...?mirror=2` could connect you to a different machine behind the same load balancer maybe? Or, ... I don't know. Hence my question: Why do we exclude the query thing? Does this solve a known problem? I think this rule could get in the way of creative use cases and probably is confusing else.

Giszmo: So you want to limit it to `protocol://domain[:port]/`, with port only being allowed for non-standard ports. Why consider the path if query isn't to be considered?

`wss://relay.minds.com/nostr/v1/ws` indeed uses a path but what's the qualitative difference to using a query string? Why allow one but not the other?

fiatjaf: I don't know. I think we should either do everything including query string or just the domain without scheme or port.

fiatjaf:
I've proposed NIP-42 at https://github.com/nostr-protocol/nips/pull/141 that just does the basic of the basic and is already super powerful. I think this NIP was already too complicated and `EXECAS` stuff would triple that complication. It is an interesting idea, but I think we should take smaller steps. Auth is needed today.

PR #98, opened 2022-12-17
nip-07: add the missing functions

x1ddos:
as per conversation in t.me/nostr_protcol

getRelays, nip04.encrypt and nip04.decrypt - these are already implemented by nos2x and getalby.

### DIFF
--- a/07.md
+++ b/07.md
@@ -13,6 +13,9 @@ That object must define the following methods:
 ```
 async window.nostr.getPublicKey(): string // returns a public key as hex
 async window.nostr.signEvent(event: Event): Event // takes an event object and returns it with the `sig`
+async window.nostr.getRelays(): { [url: string]: {read: boolean, write: boolean} } // returns a map of relay urls to relay policies
+async window.nostr.nip04.encrypt(pubkey, plaintext): string // returns ciphertext+iv as specified in nip-04
+async window.nostr.nip04.decrypt(pubkey, ciphertext): string // takes ciphertext+iv as specified in nip-04
 ```
 
 ### Implementation


Giszmo:
Where would I put multi-key-management? New nip?

```
async window.nostr.getPublicKeys(): string[] // returns public keys as hex
```

07.md
```diff
@@ -13,6 +13,9 @@ That object must define the following methods:
 ```
 async window.nostr.getPublicKey(): string // returns a public key as hex
 async window.nostr.signEvent(event: Event): Event // takes an event object and returns it with the `sig`
+async window.nostr.getRelays(): { [url: string]: RelayPolicy } // returns a map of relays
```
Giszmo, line 16: RelayPolicy is not defined. Maybe this is clearer?

```suggestion
async window.nostr.getRelays(): { [url: string]: {read: boolean, write: boolean} } // returns a map of relay urls to relayPolicies
```

x1ddos: good idea. thanks! done.

x1ddos:
On Sat, Dec 17, 2022 at 05:47:40PM -0800, Leo Wandersleb wrote:
> Where would I put multi-key-management? New nip?
> 
> ```
> async window.nostr.getPublicKeys(): string[] // returns public keys as hex
> ```

my guess would be yes, a new nip. it sounds like a multi-key management
might also require some kind of an extra API for a client to pick which
key to sign an event with?

or be able to "pin" a key to a client/website from the browser
extension. but if so, why expose `getPublicKeys` at all. a website
could just always call `getPublicKey` and the user picks the key from an
extension popup.

Giszmo:
@x1ddos as signing already contains an unsigned event, this information is contained in the current API and only the plugin would have to manage permissions per pubkey.

The plugin would have to return multiple public keys though and the encrypt/decrypt need to specify which to use.

Issue #103, opened 2022-12-20
Stateless key rotation using a series of hidden commitments


fiatjaf:
So the idea here is that Nostr apps could generate a series of keys for each user, all based on an initial seed. And they would show the seed and the first key to the user. The seed must be kept safe at maximum security, while the user can be a little more relaxed with that first key.

Then if for any reason the first key was compromised the user could use the seed, maybe in a very safe fully-offline device in a vault inside a nuclear bunker, and use it to get the second key, then use that second key to sign an event telling all interested parties that the first key was compromised.

The event from the second key alone would be sufficient for all supporting clients to verify that that second key was indeed owned by the same person who owned the first key, and that the first key was irrevocably compromised. Upon seeing this event all clients could easily automatically stop following the first key and start following the second, and (if this is applicable) even move their internal state (chats, metadata and so on) from the first key profile to the second key profile.

Once upon a time I [asked](https://t.me/spacechains/14509) @RubenSomsen to come up with a way to do this and he invented the following [scheme](https://t.me/spacechains/14565):

> So what if the secondary key is simply committed inside the primary key similar to a tapscript. Opening the commitment proves the relationship.
>
> It's basically:
>
> A' = A + hash(A||B')*G
> B' = B + hash(B||C')*G
> C' = C + hash(C||D')*G
> 
> etc.
> 
> A' can be overruled by B' by revealing A and B', which allows the commitment to B' inside of A' to be verified.
> A, B, and C have completely independent entropy

A more naïve scheme would be to just make `A = hash(B)`, `B = hash(C)` and so on (or something like that), but this would require revealing the compromise private key, which we don't want to do since that could increase the damage. Other, safer methods would involve doing zero-knowledge proofs, but the scheme above seems to be much better since it is simple and clients can probably implement it easily.

Even for clients that do not implement it, a third-party "compromised keys directory" web app could be created so users could manually go there and use that to verify if someone was compromised or not.

fiatjaf:
I believe the series of keys `A`, `B`, `C` and so on could be generated using something like NIP-06 (i.e. BIP-32) and from the first, say, 1000 keys (or less, I don't know what is the ideal number here), the sequence `A'`, `B'`, `C'` and so on could be generated.

kevinsmith:
[...]

PR #105, opened 2022-12-20
Create nip-hierarchically-deterministic-authentication.md

kngako:
Hello, 

I have a NIP that will allow users to share temporary xprivs with the nostr clients they use. 

This should also allow nostr client services to be authenticated and scale by also having unique xprivs on each server working on behalf of the nostr client service.

https://github.com/kngako/nips/blob/master/nip-hierarchically-deterministic-authentication.md

### DIFF
--- a/nip-hierarchically-deterministic-authentication.md
+++ b/nip-hierarchically-deterministic-authentication.md
@@ -1,0 +1,80 @@
+# NIP-Hierarchically-Deterministic-Authentication
+
+`draft` `mandatory` `author:kngako`
+
+## Summary
+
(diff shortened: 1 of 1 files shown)


fiatjaf:
I thought about this possibility some time ago and asked around if it would work. My very superficial and probably wrong understanding of BIP-32 after that is that you can either use

- non-hardened keys, in which case the knowledge of the xpub plus that of any derived private key is sufficient to compute the xpriv; or
- hardened keys, which the problem above doesn't exist, but then you cannot prove that a given public key is derived from a given xpub.

Or something like that, which makes this technique impossible.

fiatjaf:
Related: https://github.com/nostr-protocol/nips/issues/103

staab:
Is anyone still interested in this?

PR #108, opened 2022-12-22
Add error specification for unknown name to NIP-05

brianoflondon:
There should be some reliable response indicating a given name does not exist on the endpoint.

I've implemented this here: 

Works:
[https://v4v.app/.well-known/nostr.json?name=brianoflondon](https://v4v.app/.well-known/nostr.json?name=brianoflondon)

404:
[https://v4v.app/.well-known/nostr.json?name=123bad](https://v4v.app/.well-known/nostr.json?name=123bad)

### DIFF
--- a/05.md
+++ b/05.md
@@ -37,6 +37,14 @@ It will make a GET request to `https://example.com/.well-known/nostr.json?name=b
 
 That will mean everything is alright.
 
+### Unknown Name
+
+For an unknown `name` the end point should return `404` status code and a suitable message:
+
+```json
+{"detail":"user jack not found"}
+```
+
 ## Notes
 
 ### User Discovery implementation suggestion
--- a/05.md
+++ b/05.md
@@ -37,12 +37,15 @@ It will make a GET request to `https://example.com/.well-known/nostr.json?name=b
 
 That will mean everything is alright.
 
-### Unknown Name
+### Unknown Name for dynamic servers
 
-For an unknown `name` the end point should return `404` status code and a suitable message:
+For a site serving many names (not a static file), the api should return a `404` status code and the following result
 
 ```json
-{"detail":"user jack not found"}
+{
+  "names": {},
+  "detail" : "user bob not found"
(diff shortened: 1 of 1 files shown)


gkbrk:
Isn't that reliable response not seeing the name you queried in the "names" dictionary? That NIP is structured in this way to allow both static and dynamic servers to provide this data. So even if some servers return 404 for nonexistent names, clients still need to check that dictionary to know if the name is there or not.

mutatrum:
This doesn't work for static files indeed. I just serve a nostr.json from my domain and the webserver doesn't even look at the name query parameter.

brianoflondon:
If Nostr is successful I could end up serving 1000's of names.

I've updated the edit to be more explicit that this is for dynamic servers. I've amended [my own API](https://v4v.app/.well-known/nostr.json?name=bob) now to serve up a response like this with a `404`:

```json
{"names":{},"detail":"user bob not found"}
```

gkbrk:
If you serve 1000s of names with a dynamic server, you can return only the one that was queried. And since not all servers will check the query field, clients need to check if a name is in there. I don't see the point of both returning an empty json dict for names, including that detail and giving a 404 status code. They all indicate the same thing it looks like.

Is there a situation where names would be empty, but the detail would not say "user bob not found"?

Or detail: "user bob not found", but names has a "bob"?

I don't think we should represent the same condition with multiple ways that can all contradict with each other. If there is a field for name->key mapping like this even with a 404, do I need to check that? If that will always be empty, why do we serve it?

fiatjaf:
@gkbrk has spoken.

brianoflondon:
So make this instruction explicit in the docs.

All over lightning implementation, things that could be explicitly explained: what to do when a name doesn't exist are left vague.

I'm sorry I wasted my time trying to clarify anything in this eco system.

fiatjaf:
I am sorry you wasted your time, that wasn't my intention.

Issue #65, opened 2022-11-11
Surveys/polls


fernandolguevara:
we could add a new to tag to allow creation of surveys/polls on nostr protocol  

```meta
// create
tag: surv
options:
 - <multi|single> allow others to reply with one or multiple options
 - <ttl> TTL (in seconds|timestamp) when surv expires
 - [<choice>]: up to 4 choices each limited to 25 chars  

// reply 
tag: surv-resp
options:
 - [<choice>]: based on surv type it can have one or multi
```

```json
{
    "pubkey": "<pub-key>",
    "created_at": 1000000000,
    "kind": 1,
    "tags": [
      ["t", "hastag"],
      ["surv", "<multi|single>", "<ttl>", "choice 1", "choice 2"]
    ],
    "content": "#hastag what is your favorite ...?\n",
    "id": "<event-id>"
  },
  {
    "pubkey": "<pub-key>",
    "created_at": 1000000000,
    "kind": 1,
    "tags": [
      ["p", "<pub-key-root>", "wss://..."],
      ["e", "<event-id-root>", "wss://...", "root"],
      ["t", "tag"],
      ["surv-resp", "choice 1", ...] // based on root event surv type it can have one or multi 
    ],
    "content": "hello #tag\n",
    "id": "<event-id>"
  }
```

monlovesmango:
I think polls would be really nice to have. small comment, i think any response to the poll should list the poll event as the "root" and not "reply" in the event tag

tcheeric:
A few thoughts:
You would need to assign a different `kind` number (?) 

I would split `surv` and `choice`, and introduce `choice` as an additional separate tag, with two attributes, `index` (optional) and  `text` (mandatory) 

examples:
 ["choice", 0, "Option # 1"] 
 ["choice", "Option # 2" ] 

Minimum of two `choice` tags. 

In the reply, we put the index of the selected `choice` tag (s) in the `content` <comma-separated) , or alternatively, we include the selected `choice` tags (minimum 1) in the tag list (probably better, since `index` is optional). 

One vote per key. The relay needs to discard all subsequent votes, as well as invalid replies.

monlovesmango:
@tcheeric for #2 it should be 'root' not 'reply'. the marker identifies the event that is tagged. the event that is tagged is a root event, not a reply event. I have a pr open to clarify top level replies for nip 10.

https://github.com/nostr-protocol/nips/pull/80

PR #111, opened 2022-12-22
feat: add surveys / polls

fernandolguevara:
#65

@fiatjaf what number could we use for this nip ?


fernandolguevara:
@fiatjaf `t` and `p` tags are on the `json examples` just to show how this feature extend kind `1` events.

I will check if it is possible to use parametrized replaceable events

0xtlt:
I just saw this NIP proposal, I made another #148, I think we can try to merge our 2 ideas :)

So...

41.md
```diff
@@ -0,0 +1,69 @@
...
+#### Spec
+
+##### Message
+
+```
+tag: surv
+options:
+ - <multi|single> allow others to reply with one or multiple options
+ - <ttl> TTL (in seconds|timestamp) when surv expires, 0 TTL don't expire
+ - [<choice>]: up to 4 choices each limited to 25 chars
```
0xtlt, line 21: Because relays are dumb by design, we can't really force them to have a maximum chars which can be différents from one to another, same for the number of choices.

But we can add a recommendation to help the integration

```suggestion
 - [<choice>]: array of string
```

41.md
```diff
@@ -0,0 +1,69 @@
...
+##### Message
+
+```
+tag: surv
+options:
+ - <multi|single> allow others to reply with one or multiple options
+ - <ttl> TTL (in seconds|timestamp) when surv expires, 0 TTL don't expire
+ - [<choice>]: up to 4 choices each limited to 25 chars  
+````
+
```
0xtlt, line 23:
```suggestion

###### Relay and client recommendations

- Choices should be limited to 4 choices
- Each choices should be 25 chars limited

```

41.md
```diff
@@ -0,0 +1,69 @@
...
+```
+tag: surv-resp
+options:
+ - [<choice>]: based on surv type it can have one or multi
```
0xtlt, line 27: {
    "pubkey": "<pub-key>",
    "created_at": 1000000000,
    "kind": 8,
    "tags": [
      ["e", "<surv-event-id>", "wss://..."], // Should be the only e event which is the surv event
      ["surv-resp", "1", "2"...] // based on root event surv type it can have one or multi 
    ],
    "content": "hello #tag\n",
    "id": "<event-id>"
  }

```suggestion
tag: surv-resp
kind: 8
options:
-  [<e tag poll/survey event id>]: Should be the only e event which is the surv event
 - [<choice>]: based on surv type it can have one or multi, They represent the indexes of the answers
```

0xtlt, line 27:
```suggestion
tag: surv-resp
kind: 8
options:
-  [<e tag poll/survey event id>]: Should be the only e event which is the surv event
 - [<choice>]: based on surv type it can have one or multi, They represent the indexes of the answers
```

0xtlt (changes requested):

41.md
fernandolguevara: we can rephrase the sentence to suggest clients implement the char limit, what do you think?

dfralan (approved):

vitorpamplona:
Replaced by NIP-88

PR #112, opened 2022-12-22
Partial-Content - Transmitting file chunks over nostr

Egge7:
This draft introduces

- Two new kinds (11,12)
- Three new Tags (Content-Type, Content-Length, Content-Range)
- A general flow for transmitting files over nostr that is based on HTTP's 206 Partial Content

So far there are no clients or relays implementing these. I have written an article about a proof-of-concept that I did: https://starbackr.com/sending-images-and-files-on-nostr/

### DIFF
--- a/41.md
+++ b/41.md
@@ -1,4 +1,4 @@
-NIP-XX
+NIP-41
 ======
 
 Partial-Content
@@ -16,7 +16,7 @@ In order to handle files in the most decentralised and convenient way, we implem
 
 #### Spec
 
-This NIP introduces two new kinds `X`, `Y`, as well as a couple of tags.
+This NIP introduces two new kinds `11`, `12`, as well as a couple of tags.
 
 ```
 tags:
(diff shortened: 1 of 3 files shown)


fiatjaf:
Please don't do this.

Egge7:
@fiatjaf care to elaborate what problems you see if it is optional and Relays can decide wether to offer it or not?

Egge7:
We could avoid query-hell by introducing a file-id tag that would be consistent over all chunks. Relays could index file-ids to allow a fast retrieval of all required events from database and would not need to look up all event-ids.

Entry points would no longer need to specify all ids and it would allow transmitting live content, as chunks of data could be added while transmission has already begun.

fernandolguevara:
agreed with @fiatjaf this is a bad idea
@Egge7 u can use `http request` to upload files to the server

Egge7:
@fernandolguevara of course you can, but how would you share it with other clients without having a single point of failure? File servers and URLs in notes is what people use today, but it is not censorship resistant.

Maybe we could start to mirror content on different servers and include URLs to all mirrors in a note...

fiatjaf:
IPFS was created to solve this problem specifically.

If it works, you can use that, if it doesn't work then we must make a new protocol to solve that problem. The problem is too big in itself to be solved with an unsustainable hack like this.

Egge7:
@fiatjaf I am mostly aligned when it comes to long term storage of files, especially because of the inefficiencies with the approach proposed here. We will continue working on it. 

Now where Partial-Content shines IMO (especially if we replace the event-is-list proposal with a static file-id as proposed here: https://github.com/nostr-protocol/nips/pull/112#issuecomment-1363338850) is live content that is ephemeral.

We could have nostr spaces where audio is streamed to a relay and then to clients listening to the file-id. Of course not every Relay will be able to handle the load, so Relays can choose wether to support it or pay-gate it.

fiatjaf:
But this is so inefficient.

Considering that clients will already have to implement a very custom way of transmitting audio or files and the same for relays, wouldn't it be better to spec a more efficient protocol that used binary streams over WebSocket directly? Then relays that wanted to support it would have an easier time adding support, and client integration would be simpler too.

jb55:
NACK

adamritter:
I have thought about implementing it as well, and I think the best way would be to send an epheremal DM with file metadata and hash (or magnet hash) and a websocket address that the other party can connect to. 

After establishing a direct connection, it's rasy to send the file.

adamritter:
Just a bit of a followup: nostr already has deletion, so whenever the websocket address changes, the client could just delete the older dm.

1440000bytes:
> After establishing a direct connection, it's rasy to send the file.

But that won't be using nostr to save the file on relays. So does that require a NIP?

PR #113, opened 2022-12-22
NIP-42 - External Embed Sizes

deanpress:
This is a draft NIP to look into the issue of auto-embeds from external URLs in notes wasting user bandwidth.

### DIFF
--- a/42.md
+++ b/42.md
@@ -1,0 +1,44 @@
+NIP-42
+======
+
+External Embed Sizing
+---------------
+
+`draft` `optional` `author:deanpress`
+
+An `embed-sizes` array in notes that allows relays to store information about the filesizes of external embeddable media (i.e. URLs to images and other media in posts and avatars).
+
+Clients can auto-embed small media within a configured limit, and not auto-embed larger media by default (and give the user an option to click "View Embedded Media"), to save user bandwidth and prevent abuse.
+
+Spec
+---------------
+```
+tag: embed-sizes
+values:
+ - [array of strings of filesizes in bytes, in order of urls in note content]: required
+```
+
+Example
+---------------
+```json
+{
+    "pubkey": "<pub-key>",
+    "created_at": 1000000000,
+    "kind": 1,
+    "tags": [
+      ["embed-sizes", ["500000", "1500000", "0"]]
+    ],
+    "content": "https://imgur.com/1.png and https://imgur.com/2.png and https://imgu2.com/3.png.",
+    "id": "<event-id>"
+}
+```
+In this example the `1.png` embed is 500kb, `2.png` is 1.5MB, and `3.png` was not estimated (due to no `Content-Length` header, or the header doesn't match the real content's size).
+
+Client Behavior
+---------------
+Clients SHOULD check the `embed-sizes` array for each externally linked embeddable media url in a note and only auto-embed the media if it's within the client's (or user's) configured max allowed auto-embeddable filesize. If the `embed-sizes` value of the media embed is `"0"`, also don't auto-embed the media.
+
+Relay Behavior
+---------------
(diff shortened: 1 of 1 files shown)


deanpress:
Like Leo mentioned in the nostr telegram, there also needs to be a way to prevent this from being worked around by a server replacing an image with a new, larger image after the filesize is set.

42.md
```diff
@@ -0,0 +1,44 @@
...
+
+Example
+---------------
+```json
+{
+    "pubkey": "<pub-key>",
+    "created_at": 1000000000,
+    "kind": 1,
+    "tags": [
+      ["embed-sizes", ["500000", "1500000", "0"]]
```
gkbrk, line 29: How does this determine which URLs are merely links and which are media. Does it check the Content-Type header? Or look for a .png extension in the URL?

gkbrk, line 29: We only save the embed size, but is that all we want to save? Might be better to embed the hash of the content as well.

gkbrk: Or is it all URLs, and non-media don't have content sizes?

42.md
```diff
@@ -0,0 +1,44 @@
...
+    ],
+    "content": "https://imgur.com/1.png and https://imgur.com/2.png and https://imgu2.com/3.png.",
+    "id": "<event-id>"
+}
+```
+In this example the `1.png` embed is 500kb, `2.png` is 1.5MB, and `3.png` was not estimated (due to no `Content-Length` header, or the header doesn't match the real content's size).
+
+Client Behavior
+---------------
+Clients SHOULD check the `embed-sizes` array for each externally linked embeddable media url in a note and only auto-embed the media if it's within the client's (or user's) configured max allowed auto-embeddable filesize. If the `embed-sizes` value of the media embed is `"0"`, also don't auto-embed the media.
```
gkbrk, line 39: Why does the client trust these values? If the user wishes to avoid excessive data usage, they will need to implement checks of these file size limits even without this embedded data.

fiatjaf:
I don't understand the purpose of this.

deanpress:
@fiatjaf Like @gkbrk mentioned it's not solid yet and needs some major changes. The purpose is to prevent clients from auto-embedding extremely large files, causing major bandwidth use.

42.md
deanpress: If a relay deliberately provides false data then they can simply be removed by the user and/or the client as a reliable relay. The idea is that a good integration of this NIP helps with abuse/spam of extremely large external auto-embedded media.

42.md
deanpress: Agree we probably need more. But how does a client utilize the hash of an external embed without having downloaded the file first?

deanpress: I would say Content-Type header for media services that don't use extensions (e.g. https://placekitten.com/512/512).

42.md
```diff
@@ -0,0 +1,44 @@
...
+
+Relay Behavior
+---------------
+* Relays SHOULD check external URL embeds for a `Content-Length` header. If `Content-Length` exists and is within the Relay's configured max filesize limit for embeddable content, verify the size is correct by attempting to download the file to a temporary cache (up to the `Content-Length`) and storing the filesize (in bytes) in the note's `embed-sizes` array if it matches. If the `Content-Length` header does not exist, or the file is not fully downloaded after reaching the `Content-Length` limit, set the value belonging to the URL's array key in `embed-sizes` to `"0"`.
```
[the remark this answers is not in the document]
deanpress: My mistake. I suppose this would have to be a separate WS call that does not involve note objects directly?

deanpress:
@1440000bytes 
> 1. Why only size? If this is trying to improve security
> 2. Does it require a NIP or each client needs to manage such things for security irrespective of specs?

1. This NIP is specifically for helping large media from being auto-embedded in the user's front-end causing excessive bandwidth use.
2. If relays are to store/provide estimate data of external embeddable media, there needs to be a standard for them to provide it, so that clients can all integrate the same standard.

deanpress:
I think like @gkbrk suggested in Telegram, it might be best to leave this up to clients (e.g. streaming image data into the client and stopping the download after a certain size limit), and doesn't need to be an NIP. Closed.

PR #114, opened 2022-12-22
NIP-42 Lists

monlovesmango:
hey all,

wanted to get this NIP out there for encrypted lists as I think we should start supporting blocking and muting sooner rather than later with the influx of users. I believe I did the encryption right but would love for someone to double check.

also this was just my initial take but would love feedback to make it better.

### DIFF
--- a/28.md
+++ b/28.md
@@ -54,7 +54,7 @@ Clients SHOULD use [NIP-10](10.md) marked "e" tags to recommend a relay.
 ```json
 {
     "content": "{\"name\": \"Updated Demo Channel\", \"about\": \"Updating a test channel.\", \"picture\": \"https://placekitten.com/201/201\"}",
-    "tags": [["e", <channel_create_event_id> <relay-url>]],
+    "tags": [["e", <channel_create_event_id>, <relay-url>]],
     ...
 }
 ```
@@ -73,7 +73,7 @@ Root message:
 ```json
 {
     "content": <string>,
-    "tags": [["e", <kind_40_event_id> <relay-url> "root"]],
+    "tags": [["e", <kind_40_event_id>, <relay-url>, "root"]],
     ...
 }
 ```
@@ -84,8 +84,8 @@ Reply to another message:
 {
     "content": <string>,
     "tags": [
-        ["e", <kind_42_event_id> <relay-url> "reply"],
-        ["p", <pubkey> <relay-url>],
+        ["e", <kind_42_event_id>, <relay-url>, "reply"],
+        ["p", <pubkey>, <relay-url>],
         ...
     ],
     ...


vishalxl:
This is creating two kind 30000 events if the users wants private/secret block list and a mute list.

Why not combine the two or all such lists into one event, which will include the name of list too. 

So the json will look something like: 

`
{
  {
    "listname":"block",
    [
      ["p", "2df69cd0c6ab95e08f466abe7b39bb64e744ee31ffc3041f270bdfec2a37ec06"],
      ["p", "3bf0c63fcb93463407af97a5e5ee64fa883d107ef9e558472c4eb9aaaefa459d"],
    ]
  },
  {
    "listname":"mute",
    [
      ["p", "2df69cd0c6ab95e08f466abe7b39bb64e744ee31ffc3041f270bdfec2a37ec06"],
      ["p", "3bf0c63fcb93463407af97a5e5ee64fa883d107ef9e558472c4eb9aaaefa459d"],
    ]
  }
  #rest of such lists
}
`
( I may need to edit the exact json) 

You encrypt it and put it in content field or even a special tag field if you don't want to put encrypted stuff in content, and that won't tell anyone whether user has mute/block list even. 

Also, each list can have some other payload associated with it. For example, this allows for an 'about' field for each list, which will also remain encrypted. Other payloads or properties of lists can be come up with. Which may be standard even as we find use cases.

jb55:
I was thinking these could be different lists in case it gets too big and hit relay message size limits.

42
```diff
@@ -0,0 +1,45 @@
+NIP-42
+======
+
+Encrypted Lists
+-------------------------
+
+`draft` `optional` `author:monlovesmango`  `author:fiatjaf` `author:arcbtc`
+
+A special event with kind `30000`, meaning "encrypted list" is defined as having a list of tags encryped withen the event content, one for each of the list items that are to be included. This event kind would follow NIP-33 - Parameterized Replaceable Events. The tag list SHOULD ONLY contain single character tags that are queriable by relays. each encrypted list can only contain tags of a single type (ie. 'p' tags can't be included in the same list as 'e' tags)
```
Cameri, line 9:
```suggestion
A special event with kind `30000`, meaning "encrypted list" is defined as having a list of tags encrypted and stored within the content field. This event kind would follow [NIP-33 - Parameterized Replaceable Events](33.md). The tag list SHOULD ONLY contain [generic tags](12.md) that are queryable by relays in lowercase HEX format. Encrypted lists SHOULD not mix different generic tags (ie. 'p' tags can't be included in the same list as 'e' tags)
```

1440000bytes:
Concept ACK

> Commonly established parameters for this kind are:
> - `block` - an encrypted list of 'p' tags (or pubkeys) that the user wants to block
> - `mute` - an encrypted list of 'p' tags (or pubkeys) that the user wants to mute
> - `follow` - an encrypted list of 'p' tags (or pubkeys) that the user wants to follow secretly
> - `hashtag` - an encrypted list of `t` tags (or hashtag) that the user wants to follow
> - `event` - an encrypted list of `e` tags (or event id) that the user wants to follow

How does the first encrypted list work and block someone?

Cameri (approved):
LGTM

monlovesmango:
> How does the first encrypted list work and block someone?

I guess it wouldn't work now that I think about it. block lists would have to be public and they would have to have the users in the event tag so that it is queryable. I will remove this.

leesalminen:
I'm working on a feature for Astral right now [0] that allows users to arbitrarily group sets of relays together. When posting a new note you can select relay group(s) to post to. And when viewing a feed you can select relay group(s) to filter by.

Wondering if this can be incorporated in NIP-42? 

[0] https://github.com/monlovesmango/astral/pull/93

monlovesmango:
from @eskema on nostr (note1uazzw78ws6mz2wq9c74vz5y6lf24pj3aqnh9lc7frc94seax8p6qma7pvs):
>just a thought looking at this, why not keep everything in the same event and have the public be in tags and private in encrypted content? would that simplify or complicate?

I like this idea and am gonna make this a generalized lists pr for both public and private lists. I also changed the parameter for this kind to start with the type of list (rather than just being a list name)

monlovesmango:
@leesalminen I think for the relay groups we could likely just make a relay list for each group..? might be a bit hacky bc there isn't a convention for tagging for relays. but as long as it is kept private it shouldn't interfere with anything...

monlovesmango:
fml I accidentally closed this pr (will never create pr on my master branch again). new pr is here: https://github.com/nostr-protocol/nips/pull/183

Giszmo:
Why did you not rescue this PR? You can edit the source branch of your PR

monlovesmango:
I would prefer that too but I could not find a way to retrieve the original commits. I know with git generally you can do this. but couldn't find a way with github.

PR #118, opened 2022-12-24
LUD-01: fix typo for the desc of event kind 2

hsjoberg:
Surely must be a typo? Event 2 kind should recommend a websocket URI/URL for a relay.


fiatjaf:
You're probably the first person to read this spec.

PR #125, opened 2022-12-27
Typo in kind of Regular Events

CoolLamer:
Regular events are all events from 0 to 10 000. Not from 1 000 to 10 000.

### DIFF
--- a/16.md
+++ b/16.md
@@ -13,6 +13,10 @@ Regular Events
 A *regular event* is defined as an event with a kind in range `0 <= n < 10000` without events with kinds `0, 3, and 41` which are replaceable events.
 Upon a regular event being received, the relay SHOULD send it to all clients with a matching filter, and SHOULD store it. New events of the same kind do not affect previous events in any way.
 
+- plaintext notes should use kind `0 <= n < 999`
+- Non-plaintext notes should use kind `1000 <= n < 10000`
+- Events kinds `0, 3 and 41` are Replaceable Events defined in Previous NIP's
+
 Replaceable Events
 ------------------
 A *replaceable event* is defined as an event with a kind `0, 3, 41` and range `10000 <= n < 20000`.
--- a/16.md
+++ b/16.md
@@ -10,7 +10,7 @@ Relays may decide to allow replaceable and/or ephemeral events.
 
 Regular Events
 ------------------
-A *regular event* is defined as an event with a kind `1000 <= n < 10000`.
+A *regular event* is defined as an event with a kind `0 <= n < 10000`.
 Upon a regular event being received, the relay SHOULD send it to all clients with a matching filter, and SHOULD store it. New events of the same kind do not affect previous events in any way.
 
 Replaceable Events
--- a/16.md
+++ b/16.md
@@ -13,9 +13,8 @@ Regular Events
 A *regular event* is defined as an event with a kind in range `0 <= n < 10000` without events with kinds `0, 3, and 41` which are replaceable events.
 Upon a regular event being received, the relay SHOULD send it to all clients with a matching filter, and SHOULD store it. New events of the same kind do not affect previous events in any way.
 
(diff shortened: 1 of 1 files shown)


fiatjaf:
I think this is not a typo. Is it a typo, @Semisol?

16.md
```diff
@@ -10,7 +10,7 @@ Relays may decide to allow replaceable and/or ephemeral events.
 
 Regular Events
 ------------------
-A *regular event* is defined as an event with a kind `1000 <= n < 10000`.
+A *regular event* is defined as an event with a kind `0 <= n < 10000`.
```
Cameri, line 13: This should mention that 0, 3, and 41 are replaceable events.

CoolLamer: I added mention of those events.

Cameri (approved):
LGTM

CoolLamer:
I Added more info about events what i found in NIP-01 - Plain, non plain text regular events

16.md
```diff
@@ -13,6 +13,10 @@ Regular Events
 A *regular event* is defined as an event with a kind in range `0 <= n < 10000` without events with kinds `0, 3, and 41` which are replaceable events.
 Upon a regular event being received, the relay SHOULD send it to all clients with a matching filter, and SHOULD store it. New events of the same kind do not affect previous events in any way.
 
+- plaintext notes should use kind `0 <= n < 999`
+- Non-plaintext notes should use kind `1000 <= n < 10000`
+- Events kinds `0, 3 and 41` are Replaceable Events defined in Previous NIP's
```
Cameri, line 18: Already mentioned below.

16.md
```diff
@@ -13,6 +13,10 @@ Regular Events
 A *regular event* is defined as an event with a kind in range `0 <= n < 10000` without events with kinds `0, 3, and 41` which are replaceable events.
 Upon a regular event being received, the relay SHOULD send it to all clients with a matching filter, and SHOULD store it. New events of the same kind do not affect previous events in any way.
 
+- plaintext notes should use kind `0 <= n < 999`
```
Cameri, line 16: Not true, DMs are kind 4 which are cyphertext.

16.md
```diff
@@ -10,12 +10,15 @@ Relays may decide to allow replaceable and/or ephemeral events.
 
 Regular Events
 ------------------
-A *regular event* is defined as an event with a kind `1000 <= n < 10000`.
+A *regular event* is defined as an event with a kind in range `0 <= n < 10000` without events with kinds `0, 3, and 41` which are replaceable events.
 Upon a regular event being received, the relay SHOULD send it to all clients with a matching filter, and SHOULD store it. New events of the same kind do not affect previous events in any way.
 
+### Additional rules
+- Non-plaintext notes should use kind `1000 <= n < 10000`
+
```
Giszmo, line 18: This is not a typo but the addition of a whole new rule?? Could we elaborate on this please? What is a non-plaintext note? What is a note?

CoolLamer: It was based on this row from NIP-01:
https://github.com/nostr-protocol/nips/blob/d41834fa5169d15ffd1519a773ee495c6076ecd3/01.md?plain=1#L101

CoolLamer: It was based on this row from NIP-01:
https://github.com/nostr-protocol/nips/blob/d41834fa5169d15ffd1519a773ee495c6076ecd3/01.md?plain=1#L101

Giszmo: Just saying. The PR is not about typos but yeah, it's not a new rule. I had not known the rule from nip-01 and think it should be clearer. And I'm not sure if it should be maintained or dropped entirely as now kind ranges are about treatment, so where do I put my ephemeral binary message? Probably in the ephemeral range, not the binary range.

staab:
Closing due to inactivity. I also think these changes make things more complicated.

PR #126, opened 2022-12-28
Adds the possibility of NIP05 identifiers in a user's contact list (NIP02).

vitorpamplona:
Simple extension to NIP02-Contact List and Petnames to allow users to follow NIP 05 identifiers instead of direct keys. 

The goal is to automatically migrate followers to new keys when the user rotates them. 

Since most UIs already hide the pub key when NIP05 is verified, it makes sense that users only get to know and see NIP05 IDs and thus choose to follow them as opposed to keys. NIP05 is the only identifier they trust.

### DIFF
--- a/02.md
+++ b/02.md
@@ -6,9 +6,9 @@ Contact List and Petnames
 
 `final` `optional` `author:fiatjaf` `author:arcbtc`
 
-A special event with kind `3`, meaning "contact list" is defined as having a list of `p` tags, one for each of the followed/known profiles one is following.
+A special event with kind `3`, meaning "contact list" is defined as having a list of `p` and `i` tags, one for each of the followed/known profiles one is following.
 
-Each tag entry should contain the key for the profile, a relay URL where events from that key can be found (can be set to an empty string if not needed), and a local name (or "petname") for that profile (can also be set to an empty string or not provided), i.e., `["p", <32-bytes hex key>, <main relay URL>, <petname>]`. The `content` can be anything and should be ignored.
+Each tag entry should contain the key (`p`) for the profile or an `i` for a NIP-05 identifier where the key can be found, a relay URL where events from that key can be found (can be set to an empty string if not needed), and a local name (or "petname") for that profile (can also be set to an empty string or not provided), i.e., `["p", <32-bytes hex key>, <main relay URL>, <petname>]`. The `content` can be anything and should be ignored.
 
 For example:
 
@@ -19,6 +19,7 @@ For example:
     ["p", "91cf9..4e5ca", "wss://alicerelay.com/", "alice"],
     ["p", "14aeb..8dad4", "wss://bobrelay.com/nostr", "bob"],
     ["p", "612ae..e610f", "ws://carolrelay.com/ws", "carol"]
+	["i", "jeff@domain.com", "ws://jeffrelay.com/ws", "jeff"]
   ],
   "content": "",
   ...other fields
--- a/02.md
+++ b/02.md
@@ -19,7 +19,7 @@ For example:
     ["p", "91cf9..4e5ca", "wss://alicerelay.com/", "alice"],
     ["p", "14aeb..8dad4", "wss://bobrelay.com/nostr", "bob"],
     ["p", "612ae..e610f", "ws://carolrelay.com/ws", "carol"]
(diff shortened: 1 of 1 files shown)


fiatjaf:
This breaks all the things in the world.

vitorpamplona:
> This breaks all the things in the world.

Why?

gkbrk:
> This breaks all the things in the world.

Nostr apps already allow finding and following users by NIP-05 identifiers, but the contact list event loses this information when backing up users found by NIP-05 identifiers. Next time you read this contact list event, and query for the profile metadata, you might see a completely different NIP-05 identifier, or the relays might not have a profile metadata event to restore this from.

fiatjaf:
> Why? What am I missing?

The entire purpose of Nostr is that one would follow public keys, not DNS identifiers. This breaks everything and introduces two kinds of follows which wil probably cause incompatibility between all clients.

You could argue that Nostr would be better off with just the DNS identifiers -- and then you wouldn't need keys at all, just let your NIP05 server produce events in your name and as long as they are coming from that server they are authentic. This is Mastodon.

vitorpamplona:
> The entire purpose of Nostr is that one would follow public keys, not DNS identifiers. 

I agree it's about the keys. But this PR does not affect the core purpose or the protocol of Nostr. Apps will still work with public keys in the background. This change doesn't affect relays at all. The API is still public-key based. Everything stays the same. 

The change simply allows users to CHOOSE to follow the identifier, in case they trust the identifier MORE than the key itself. 

> You could argue that Nostr would be better off with just the DNS identifiers

But I am not arguing that. Otherwise, I would propose killing the pub key field entirely.  

>  introduces two kinds of follows which wil probably cause incompatibility between all clients.

Yes, apps will need to adjust (as they need to adjust with any NIP change). They are already fully adjusted to NIP05 (many don't even show the public key anymore). It shouldn't be very hard to add this peace of mind feature to their users. 

Again, this just adds an option. If you are right that pubkey is the way to go at all times, I am sure the free market will prefer it. If users choose to rely on NIP-05, I don't think it breaks anything.

gkbrk:
> The entire purpose of Nostr is that one would follow public keys, not DNS identifiers. This breaks that principle and introduces two kinds of follows which wil probably cause incompatibility between all clients.

There are already two kinds of follows, because clients allow finding users by their NIP-05 identifiers. This is even a suggestion written under NIP-05. If a user types `user@example.com` to an application, and clicks follow on the NIP-05 verified result that pops up, the NIP-05 identifier is what the user followed regardless of how much the app likes to pretend the user actually wanted to follow the pubkey that was resolved through NIP-05.

If the only thing someone told me is they are `user@example.com`, it makes sense for my client to save that as the identifier I am following, and potentially even auto-update the pubkey if the NIP-05 pubkey changes. Because that person told me they are `user@example.com`, that is what I wanted to follow, that is what I wrote on the application. I clicked a follow button because the app told me "this user is verified to be user@example.com".

If we don't agree that there are two kinds of follows (pubkey and NIP-05), maybe that suggestion should not be listed on the NIP, or even _not_ doing that should be recommended for apps.

fiatjaf:
> If the only thing someone told me is they are user@example.com, it makes sense for my client to save that as the identifier I am following, and potentially even auto-update the pubkey if the NIP-05 pubkey changes.

If this is your understanding please help me rewrite NIP-05 such that it doesn't give room for this kind of interpretation. Clients should really not follow DNS identifiers and if any is doing it is a very bad client.

If someone told you they "are" `user@domain.com` that is because they cannot spell a 32-byte random public key and they trust your client to follow the Nostr specs and fetch their public key and follow that. If you're going to argue on the basis of misunderstandings like that you can also argue that if they said they are `user@domain.com` then it is because they are telling your that is their email address, not Nostr address, or anything like that. All understanding must be done within a context.

I will modify NIP-05 to make this clearer.

fiatjaf:
> Again, this just adds an option. If you are right that pubkey is the way to go at all times, I am sure the free market will prefer it. If users choose to rely on NIP-05, I don't think it breaks anything.

As my free-market act as maintainer of this repository, I beg to disagree and close this issue as pernicious to the principles and openness of the protocol.

vitorpamplona:
So, NIP05s are secure enough to do user discovery, but not secure enough to do key rotation. It doesn't make any sense.

tmathews:
This adds unneeded complexity to every client. No thanks.

Issue #127, opened 2022-12-28
NIP-05 reflectivity


majestrate:
a common issue with activity pub i have seen in the wild is in practice it turns into a giant XSRF sending network that you can get thousands of servers to do a GET request to a single url. activity pub has no way to authenticate if this is something that the server wants.

in NIP-05, this problem exists too. a malicious actor would be able to very trivially use a server that serves the `/.well-known/nostr.js` endpoint to redirect to an arbitrary other location and CORS lets it go cross origin again. one could flood with new bogus users to redirect the thousands of GET requests to whatever url they want, as many times as they want.

NIP-05 should explicitly disallow any kind of redirects to prevent this.

fiatjaf:
Can you edit NIP-05 to add a line disallowing redirects? Redirects are a cancer of the internet.

majestrate:
yup. there should also be clear requirements backoff with refetches here too to prevent servers that arent serving this endpoint from being spammed hard.

majestrate:
or... perhaps use a DNS record to opt out of this that admins can set to prevent their server being hit. without notice.

fiatjaf:
No, I don't think we should do the DNS stuff.

majestrate:
fair.

PR #128, opened 2022-12-28
update NIP-05 addressing reflectivity.

majestrate:
address issue #127

add security constraints to mitigate XSRF properties in NIP-05

### DIFF
--- a/05.md
+++ b/05.md
@@ -66,9 +66,6 @@ Users should ensure that their `/.well-known/nostr.json` is served with the HTTP
 
 ### Security Constraints
 
-
-If a TXT record for `_ignoreme._nostr.<domain>` exists the fetcher MUST NOT send NIP-05 requests to this domain.
-
 The `/.well-known/nostr.json` endpoint MUST NOT redirect cross origin. 
 
 Fetchers MUST ignore any cross origin redirects on the `/.well-known/nostr.json` endpoint.
--- a/05.md
+++ b/05.md
@@ -67,7 +67,7 @@ Users should ensure that their `/.well-known/nostr.json` is served with the HTTP
 ### Security Constraints
 
 
-If a TXT record for `_ignoreme._nostr.<domain>` exists the fetcher MUST NOT send NIP-05 requests from this domain.
+If a TXT record for `_ignoreme._nostr.<domain>` exists the fetcher MUST NOT send NIP-05 requests to this domain.
 
 The `/.well-known/nostr.json` endpoint MUST NOT redirect cross origin. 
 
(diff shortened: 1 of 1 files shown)


gkbrk:
Are we really asking every domain on the web to add a DNS record to opt out, instead of asking for the comparatively minuscule number of NIP-05 serving domains to opt in?

majestrate:
in general, i find that having to configure another dns record to be the biggest PITA when setting a service up. 

i am okay with opt in, i just thought that an explicit opt out would help with regards to negative caching of dns. e.g. removing the old record but it remains in the cache for a few days.

majestrate:
also, the dns record is out of scope of that issue. it is now omitted.

gkbrk:
Looks good, if clients are recommended to ignore redirects, there would be no point in bad servers trying to do it either.

If someone really wants to delegate Nostr lookups to different servers, they can reverse-proxy the requests instead of using an HTTP Redirect.

majestrate:
>Looks good, if clients are recommended to ignore redirects, there would be no point in bad servers trying to do it either.

this uses RFC 2119 language. MUST is not a recommendation.

gkbrk:
> this uses RFC 2119 language. MUST is not a recommendation.

Haha, I'm sure any client that follows redirects anyway will feel terrible about not listening to RFC-2119 language.

majestrate:
>Haha, I'm sure any client that follows redirects anyway will feel terrible about not listening to RFC-2119 language.

my understanding is that following these redirects is the default behavior on a browser hitting an endpoint with wide open CORS.

gkbrk:
> my understanding is that following these redirects is the default behavior on a browser hitting an endpoint with wide open CORS.

Any decent HTTP client can configure following redirects or not, and the browser is no exception. The `fetch` API can be considered to not follow redirects, or even raise errors if a redirect is asked.

https://developer.mozilla.org/en-US/docs/Web/API/fetch#redirect

majestrate:
> > my understanding is that following these redirects is the default behavior on a browser hitting an endpoint with wide open CORS.
> 
> Any decent HTTP client can configure following redirects or not, and the browser is no exception. The `fetch` API can be considered to not follow redirects, or even raise errors if a redirect is asked.
> 
> https://developer.mozilla.org/en-US/docs/Web/API/fetch#redirect

correct, and a naive implementation will use the default:

```
redirect
How to handle a redirect response:

  follow: Automatically follow redirects. Unless otherwise stated the redirect mode is set to follow.
```

hence the note here about this being explicitly not allowed. the defaults of the fetch api are not going to do that.

fiatjaf:
I think they must not redirect _at all_, not only cross-origin.

majestrate:
updated to disallow all redirects

PR #129, opened 2022-12-29
NIP-10: Clarify marked e-tags for direct replies

ok300:
AFAIK different clients implement this differently. Some add both marker tags (`root` and `reply`) for direct replies, some add one, some add the other.

This PR suggests a way to clarify that confusion.

### DIFF
--- a/10.md
+++ b/10.md
@@ -48,7 +48,7 @@ Where:
 
 **The order of marked "e" tags is not relevant.**  Those marked with `"reply"` denote the `<reply-id>`.  Those marked with `"root"` denote the root id of the reply thread.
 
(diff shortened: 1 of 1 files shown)


10.md
```diff
@@ -48,6 +48,8 @@ Where:
 
 **The order of marked "e" tags is not relevant.**  Those marked with `"reply"` denote the `<reply-id>`.  Those marked with `"root"` denote the root id of the reply thread.
 
+A direct reply to the root of a thread should have a single marked "e" tag of type "reply".
+
```
eskema, line 52: it's the opposite actually, a single reply should contain only an `e` marked root

ok300: Updated PR.

PR #132, opened 2022-12-29
Make it clear that NIP-05 Keys should be in Hex

lylepratt:
There has been some confusion about whether npub keys are supported by this spec. According to @fiatjaf only Hex keys are supported. https://twitter.com/fiatjaf/status/1608606752987316224?s=20&t=6fJLD3077byuoTm96kva1g


fiatjaf:
Thank you very much.

PR #138, opened 2023-01-01
NIP-0X: Deterministic Aliases derived from private keys

RandyMcMillan:
NIP-0X: Deterministic Aliases derived from private keys

### DIFF
--- a/0X-simple-deterministic-aliases.md
+++ b/0X-simple-deterministic-aliases.md
@@ -1,0 +1,141 @@
+NIP-0X
+======
+
+Simple Deterministic Aliases
+----------------------------
+
+`draft` `optional` `author:@RandyMcMillan`
+
+### Generating new profiles deterministically from an existing secure privkey or entropy using sha256
+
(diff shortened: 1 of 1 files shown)


RandyMcMillan:
Screen Shot 2023-01-01 at 12 57 21 AM

fiatjaf:
I don't understand, is this a technique for generating multiple keypairs from the same seed? Why not use BIP32 as in NIP06?

RandyMcMillan:
@fiatjaf - Did you actually read the proposal? I imagine you are busy so I won't hold it against you... but part of the objective here is to have a scheme that can both be implemented easy in software... but can also be done manually by the user. People seem to like autonomy and "self sovereignty" - so it would follow that - there is a simple scheme for user's to generate keys - while also having the benefits of a bip-0085 like scheme.

As time moves forward... people are going to realize that they are trusting the implementation of a particular client or website to generate their keys. All the common concerns about retirement attacks/etc. will come to the fore... I don't see any reason to poo poo anticipating this stuff. It is not like we haven't seen/heard it before.

We both know that even if a cryptographic library is well scrutinized - that doesn't remove the problem of implementation by a developer. Or worse - a malicious developer - that generates keys with faux entropy.

**Maybe you don't realize how easy it is to fake secure entropy? I assure it is quite easy to do...**

PR #140, opened 2023-01-02
NIP18: Reposts

fiatjaf:


18.md
```diff
@@ -0,0 +1,19 @@
+NIP-18
+======
+
+Reposts
+-------
+
+`draft` `optional` `author:jb55`
+
+A repost is a `kind 6` note that is used to signal that you want your followers
+that another event is worth reading.
```
Giszmo, line 10:
```suggestion
A repost is a `kind 6` note that is used to signal to followers
that another event is worth reading.
```

18.md
```diff
@@ -0,0 +1,19 @@
...
+
+The `content` of a repost event is empty.
+
+The repost event MUST include an `e` tag with the `id` of the note that is
+being reposted. That tag SHOULD include a relay URL from where as its third entry
+that note can be fetched as its third entry.
```
Giszmo, line 16:
```suggestion
being reposted. That tag SHOULD include a relay URL as its third entry
to indicate where it can be fetched.
```

Giszmo (changes requested):
fix these wordings please!

Also I would add an example event although it's quite clear.

You could add some request to not add a million more p- and e-tags.

18.md
```diff
@@ -0,0 +1,19 @@
...
+A repost is a `kind 6` note that is used to signal to followers
+that another event is worth reading.
+
+The `content` of a repost event is empty.
+
+The repost event MUST include an `e` tag with the `id` of the note that is
+being reposted. That tag SHOULD include a relay URL as its third entry
+to indicate where it can be fetched.
+
+The reaction SHOULD include a `p` tag with the `pubkey` of the event being
```
jb55, line 18: reaction?

jb55:
LGTM, I plan on having a quoted repost which is like this but with something in the content. I do want to it be a different kind because damus puts crap in the content of kind6 atm.

fiatjaf:
If I understand it correctly kind 1 probably the best place for quoted reposts.

fiatjaf:
@Giszmo examples can come later, I just wanted to write this here somewhere. We do need some more examples or test vectors and a general cleanup on all NIPs, but I hope eventually these things will happen by magic.

fiatjaf:
Yeah, maybe kind is not the best? We've had this discussion in some other NIP proposal, from @monlovesmango I think, but we got nowhere.

I think it is easy though. You add a tag and mark it as "quote". Then clients show that as reposts. If clients don't understand a thing they may show your repost as a reply to the quoted event, which won't be bad.

jb55:
btw these kinds of reposts will not show up in damus, which is unfortunate.

jb55:
> If I understand it correctly kind 1 probably the best place for quoted reposts.

damus is fast because it is able to display everything it needs to display on the first query. These types of specs don't really work in damus. for quoted reposts I will have something like damus' current repost spec that encodes the event that is being reposted in the event itself.

fiatjaf:
> btw these kinds of reposts will not show up in damus, which is unfortunate.

I don't understand. What kinds? the ones described in this NIP? My only intention with this NIP was to describe what Damus was doing.

jb55:
this current spec is not how damus currently implements reposts.

Cameri:
I'm wondering what was so wrong with https://github.com/nostr-protocol/nips/pull/86/commits/c875b97c60d39534d343deb4b230e1c92601c804

jb55:
I never spec'd it because I figured a lot of people would think its dumb and wouldn't want to do it the way I'm doing. I encode the event I'm reposting as json into the content so I don't have to fetch it on the initial timeline load.

jb55:
snort.social is also planning on implementing it damus style as well afaik

fiatjaf:
I know you used to do that, but you said you were going to change, then I saw @Cameri and others explaining to others how it must be done, which is how I wrote in this spec.

jb55:
I tried to change it but it made damus crappier so I didn't go with the change

fiatjaf:
Would be nice if everybody could be more cooperative and if we could agree on a "Nostr spec" for this, not a "Damus spec", "Snort spec", "Astral spec" and so on. I actually have no idea of what each client is doing now, but I think Astral is doing an entirely different thing. Of course the Damus way will always win since Damus has so many more users and mindshare.

fiatjaf:
> I tried to change it but it made damus crappier so I didn't go with the change

I feel like the basic Nostr relay querying model is not good enough for these use cases. It makes it hard to show threads with replies in the home feed too, and other things probably.

Maybe we should fix that.

jb55:
yeah I didn't want my dumb implementations to dictate the protocol, which is why I stayed out of it. I can make my implementation handle the less efficient repost spec in the future.

jb55:
right now astral reposts are just posting an event link in the timeline. this looks horrible on damus. I'm very much against using kind1 for everything.

eskema:
> > I tried to change it but it made damus crappier so I didn't go with the change
> 
> I feel like the basic Nostr relay querying model is not good enough for these use cases. It makes it hard to show threads with replies in the home feed too, and other things probably.
> 
> Maybe we should try fix that.

maybe some value on the request to also include events in tags referenced in the notes it sends?

jb55:
> > > I tried to change it but it made damus crappier so I didn't go with the change
> > 
> > 
> > I feel like the basic Nostr relay querying model is not good enough for these use cases. It makes it hard to show threads with replies in the home feed too, and other things probably.
> > Maybe we should try fix that.
> 
> maybe some value on the request to also include events in tags referenced in the notes it sends?

this would be bandwidth-inefficient if the client already has the events cached... I'm not sure it makes sense for most things.

fiatjaf:
> maybe some value on the request to also include events in tags referenced in the notes it sends?

Sounds like a good idea, but I wonder how bad this will be for relays. @scsibug @atdixon

jb55:
I was thinking specifically for reposts, having clients try to hunt down the source of it would be really annoying. encoding it in the event allows you to get around all that complexity. Trying to hunt down posts in a thread that you're missing is also annoying, but out of the scope of what I was trying to do with the damus repost spec.

jb55:
@fiatjaf has always said clients should be connecting to relays all the time and following relay hints to track down event. There is no easy way to do that in the damus client. It only ever talks to the relays it has configured. I don't even know if its a good idea for clients to connect to random untrusted relays all the time just because someone suggested to find something there. These are deep issues which affect the design of a lot of these things.

jb55:
There are many issues I have had when connecting to random relays. its very easy for them to crash and confuse damus with garbage. Sure damus should probably be hardened against that, but it's still a concern that I get around by making sure the client doesn't try to connect to random buggy relays.

fiatjaf:
Should we replace the entire WebSocket thing with one-off HTTP calls that can be easily sent to a bunch of random relays?

fiatjaf:
> I don't even know if its a good idea for clients to connect to random untrusted relays all the time just because someone suggested to find something there.

But don't you think your client should connect to a random relay if someone you follow is only publishing to that relay? What if you click on a person's profile because you saw it a reposted event from them?

How do you see Nostr working without some form of relay hints and clients following them.

You even said on a podcast that you imagine a world with a handful of big supercrowded relays where you can basically find everyone, but then smaller and more restricted relays that are safer places you can visit later. How will your client find these safer places if not by following some form of hints?

jb55:
> But don't you think your client should connect to a random relay if someone you follow is only publishing to that relay? What if you click on a person's profile because you saw it a reposted event from them?

relays are displayed on user profiles, its a users manual decision to become well connected to someone, not an automatic decision by the client.

fiatjaf:
> > But don't you think your client should connect to a random relay if someone you follow is only publishing to that relay? What if you click on a person's profile because you saw it a reposted event from them?
> 
> relays are displayed on user profiles, its a users manual decision to become well connected to someone, not an automatic decision by the client.

I personally like manual decisions, but still doesn't prevent someone from living in their own small set of relays but still being able to get signal from the external world?

In my mind one should be able to just query posts from their followers from anywhere, then accept to see replies to events from a restricted set of relays that don't usually have spam (so they can see replies from people they don't follow). And then they're still able to see posts from people they do not follow and are not in their set of "safe" relays by following hints from
  - reactions and replies from their followers to others in the external world
  - reposts from their followers from posts from the external world

I should probably draw this.

fiatjaf:
By the way, all this could be done either automatically by the client or only after manual action from the client: "load this note from relay wss://danger.zone? yes / no`

jb55:
yes this is what I'm thinking for missing notes in threads. the ux might be terrible so I will have to think about it. but at least it will give the user hints that they should be connected to that relay if they see a blank event with that message a lot.

jb55:
another option is just to have a temporary section of the relay pool that is only active during the course of a session or thread, then it could use that to follow hints. I don't know, maybe that's better.

fiatjaf:
```
["REQ", "_", {"authors":["A","B","C",...], "kinds":[1,6,7], "with_references": true}]
```

```
["EVENT", {"id":"xyz", ...}]
["EVENT", {"id":"abc", "tags":["e", "foo"]}]
["REFERENCE", "abc", {"id":"foo"}]
["EVENT", {...}]
```

eskema:
> ```
> ["REQ", "_", {"authors":["A",...], "kinds":[1,6,7], "with_refs": true}]
> ```
> 
> ```
> ["EVENT", {"id":"xyz", ...}]
> ["EVENT", {"id":"abc", "tags":["e", "foo"]}]
> ["REF", "abc", {"id":"foo"}]
> ["EVENT", {...}]
> ```

i like this. @Cameri @scsibug what's your opinion from the relay side? this should also relieve some of the load from inneficient queries, so the drawbacks and gains should also have that in consideration..

one thing i see there is you include one event per ref, which may not be desirable because i think it may lead to multiple duplicates or refs missing because they were returned first from another one. so maybe instead of the id, it could have the subscription id, so it only returns unique ones

atdixon:
> > ```
> > ["REQ", "_", {"authors":["A",...], "kinds":[1,6,7], "with_refs": true}]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```
> > ["EVENT", {"id":"xyz", ...}]
> > ["EVENT", {"id":"abc", "tags":["e", "foo"]}]
> > ["REF", "abc", {"id":"foo"}]
> > ["EVENT", {...}]
> > ```
> 
> i like this. @Cameri @scsibug what's your opinion from the relay side? this should also relieve some of the load from inneficient queries, so the drawbacks and gains should also have that in consideration..
> 
> one thing i see there is you include one event per ref, which may not be desirable because i think it may lead to multiple duplicates or refs missing because they were returned first from another one. so maybe instead of the id, it could have the subscription id, so it only returns unique ones
> 
> `["REF", "_", {"id":"foo"}]`

is indicating REF important? i agree things get complicated with it... duplicates and indicating things that are both in the result set and count as a ref..

what is primary goal here? to avoid two queries from client?
in current world, can't clients come back with a batch request (#e=[...,...]) for any needed reposted/ref'd events? can't clients can render on first query results and then progressively enhance reposts etc after second? 

(i'm assuming with_refs wouldn't be asking relays to _recursively_ resolve refs? or would the spec make that up to relay? starts to get complicated in this case... also for non-repost use cases where number of tagged events could be arbitrary, even w/o recursive resolve, there's a question of arbitrarily large ref results and clients won't really know what they're getting into..... things seem to stay simple if client just leverages current protocols and pick & choose what resolves they want and i don't see that as a huge burden on clients .. but i might be misisng something)  (forgive typos..banging out thoughts while on the run)

eskema:
> (i'm assuming with_refs wouldn't be asking relays to _recursively_ resolve refs?

right, only from EVENT, hence I think the REF distinction, so it doesn't trigger tags for the ref..

eskema:
maybe it should only send the refs after all events are done but before eose

jb55:
@atdixon damus' repost spec serves two purposes: to avoid two queries and potentially many more expensive actions if it can't find the event on any relays the client is connected to

atdixon:
> > (i'm assuming with_refs wouldn't be asking relays to _recursively_ resolve refs?
> 
> right, only from EVENT, hence I think the REF distinction, so it doesn't trigger tags for the ref..

it seems to me to be useful to the client the ref has to come pretty much right with the event - otherwise client is going to have to hang the user experience anyway until it gets the refs it likes. this does put a burden on relays who either have to take care to avoid sending duplicates later on in the result set)

as i think more -- i see just two sensible options
1- relay gives back ref effectively at the same time as it delivers the event (which is what damus is doing by hijacking the content)
2- or client has to progressively enhance (i.e., show ux before refs resolve)

1 is always going to be bandwidth inefficient
2 seems the only sensible path to me (prove me wrong) ... and if that is the case i would keep nostr simple right now and leave it to clients to do a second-batch request (or not like astral is choosing to) ... this path seems simplest, better for bandwidth, and most flexible for clients

(i'm trying to argue _against_ anything new here - definitely playing devil's advocate - pls dont shoot me)

atdixon:
> @atdixon damus' repost spec serves two purposes: to avoid two queries and potentially many more expensive actions if it can't find the event on any relays the client is connected to

makes sense, at the expense of bandwidth ... but you are still delivering _some_ kind of ux for reposts from non-damus clients, right? (what does that look like?)

jb55:
I probably won't show them for now, or at least check to see if the id is in the local cache and display it if so.

atdixon:
makes sense - it's bandwidth but it's simplest approach

brainstorming - let's say my query results are
```
["EVENT", "{kind: 6, #e: A}", "{id: A, ...}", ...]
...
["EVENTID", A]
...
["EVENT", "{kind: X, #e: [A,B]}", "{id: B, ...}"]
```
(nip01 is extended to allow ref'd events to get appended to "EVENT" results)

however it's optional AND smart relays will try not to duplicate events when it knows its already served it for the subscription

so in this example the first event includes A
later -- in this example -- A is _actually part of the result set, as well_ so relay sends EVENTID
also later -- the last event refers to A and B and relay simple does not include A because it knows its already sent it

relays don't have to support this, of course

atdixon:
the one thing i don't like is that when _new_ reposts come in, relays will now have to hit disk/cache to serve the refs - this is new demand on relay not currently present and has new perf implication

scsibug:
I think this is a lot of complexity for supporting a minor client side feature.  If I was going to do this from scratch; I'd probably have a tag for kind-1 notes called "repost" that referred to an event ID the user was trying to share.  The content would still be whatever the user wanted.  The client would issue a new REQ for one or a batch of several of the IDs mentioned in the repost tags, and fill in the content if possible, otherwise leaving a placeholder to indicate a referenced-but-not-yet-retrieved note.  This isn't super fast to load, but i wonder how much we need to optimize this use case.

Relays dereferencing and including events seems like a recipe for poor performance.  At worst issuing N queries to satisfy a single filter that returned N results.  At best, joining on another table and increasing the number of events returned (my bigger concern).  At some point, this just turns into a graph query language....

I think we should probably have multiple really compelling use cases before changing the core protocol, and an implementation; and stick to something less performant until then.

Relays could just send extra events that they think a client would want, but it's not clear to me that isn't mostly a waste of bandwidth.

atdixon:
> I think this is a lot of complexity for supporting a minor client side feature. If I was going to do this from scratch; I'd probably have a tag for kind-1 notes called "repost" that referred to an event ID the user was trying to share. The content would still be whatever the user wanted. The client would issue a new REQ for one or a batch of several of the IDs mentioned in the repost tags, and fill in the content if possible, otherwise leaving a placeholder to indicate a referenced-but-not-yet-retrieved note. This isn't super fast to load, but i wonder how much we need to optimize this use case.
> 
> Relays dereferencing and including events seems like a recipe for poor performance. At worst issuing N queries to satisfy a single filter that returned N results. At best, joining on another table and increasing the number of events returned (my bigger concern). At some point, this just turns into a graph query language....
> 
> I think we should probably have multiple really compelling use cases before changing the core protocol, and an implementation; and stick to something less performant until then.
> 
> Relays could just send extra events that they think a client would want, but it's not clear to me that isn't mostly a waste of bandwidth.

💯  perfectly stated

Cameri:
> I think this is a lot of complexity for supporting a minor client side feature. If I was going to do this from scratch; I'd probably have a tag for kind-1 notes called "repost" that referred to an event ID the user was trying to share. The content would still be whatever the user wanted. The client would issue a new REQ for one or a batch of several of the IDs mentioned in the repost tags, and fill in the content if possible, otherwise leaving a placeholder to indicate a referenced-but-not-yet-retrieved note. This isn't super fast to load, but i wonder how much we need to optimize this use case.
> 
> Relays dereferencing and including events seems like a recipe for poor performance. At worst issuing N queries to satisfy a single filter that returned N results. At best, joining on another table and increasing the number of events returned (my bigger concern). At some point, this just turns into a graph query language....
> 
> I think we should probably have multiple really compelling use cases before changing the core protocol, and an implementation; and stick to something less performant until then.
> 
> Relays could just send extra events that they think a client would want, but it's not clear to me that isn't mostly a waste of bandwidth.

Agreed.

cxplay:
Today, I inevitably ran into the same problem while exploring nostr: https://github.com/scsibug/nostr-rs-relay/issues/122

But this time I'm just an average nostr user who is not technically proficient in programming, but I still want to post some of my thoughts. And since I am not a native English speaker, feel free to discuss with me.

Before that, it is important to mention "repost", "comment", "reply", "quote" and "ref.", which I often encounter when writing, so it is important to understand that these five concepts are not identical.

1. "repost": republish an event to your own timeline, either an event that you have already posted, or an event that someone else has posted. Also, these events can be any kind of events, such as kind-30023 as defined by NIP-23, or regular events in kind-1. And "repost" does not require the existence of new content (relative to the target event being "repost").
2. "comment": A comment on an event (note, not "reply"), which is often confused with "reply", this is an event that requires a clear context, but its not required in a "reply".
3. "reply": This should only be used in IM, but since nostr can also do IM, it should be added to this list. Since IM is not always linear, it doesn't really need a context at all, and if it does it becomes a separate one: "quote".
4. "quote": "repost" another event and a separate event containing the "comment" of this event, which differs from "repost" in that it must contain new content, otherwise it does not constitute a "quote". It is often confused with "ref.". In essence, to implement "quote" is to "comment" and "repost" an event at the same time.
5. "ref.": contains the content of another event content within an event, and is used to form part of the content of this event. Since the Internet is composed of URLs, in this case, the content of "ref." can also be the URL of another event. There is also a context here, but it is the content or URL of another event that the user manually extracts as part of the content of this event.

1. repost = Post an existing event.
2. comment = Post content that points to the context.
3. reply = Post content but no context(Specifically in IM).
4. quote = Post an event containing already existing events and comments.
5. ref. = Part of the event content, but this part is extracted from other event content.

---

So, what is the most important thing for a repost? It has to contain an event that already exists before it can evolve into a repost and quote. And unlike a comment, it has to be a separate event.

The performance burden for such a repost should be on the client side, because the relay should only provide the ID(?) of which existing event is contained in the repost, and the client will further retrieve this existing event and display it.

The repost does not need to contain the content of that existing event, if it does it is more like embedding an event in the long content of NIP-23, at which point he should be using the definition of NIP-23.

PR #141, opened 2023-01-02
add nip-42: authentication

fiatjaf:
Readable: https://github.com/nostr-protocol/nips/blob/auth/42.md


fiatjaf:
Who else is implementing chat stuff so I can tag them here? @ArcadeCityMayor

42.md
```diff
@@ -0,0 +1,68 @@
...
+
+```
+["NOTICE", "restricted: we can't serve DMs to unauthenticated users, does your client implement NIP-42?"]
+```
+
+or
+
+```
+["NOTICE", "restricted: we do not accept events from unauthenticated users, please sign up at https://example.com/"]
+```
```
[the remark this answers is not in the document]
fiatjaf: Please tone down.

42.md
```diff
@@ -0,0 +1,68 @@
...
+## Protocol flow
+
+This NIP defines a new message, `AUTH`, which clients can send to relays when they want to authenticate. The message is of the following form:
+
+```
+["AUTH", <signed-event>]
+```
+
+The signed event is an ephemeral event not meant to be published or queried, it must be of `kind: 22242` and content must be set to the
+WebSocket URL of the relay. `created_at` should be the current time. Example:
```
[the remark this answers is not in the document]
fiatjaf: @Giszmo and I were discussion about the rationale behind this on the other PR, but you said nothing. Is there a strong reason to have these things specified here?

42.md
```diff
@@ -0,0 +1,68 @@
...
+
+```json
+{
+  "id": "...",
+  "pubkey": "...",
+  "created_at": 1669695536,
+  "kind": 22242,
+  "tags": [],
+  "content": "wss://relay.example.com/",
```
eskema, line 40: just a question: should the trailing slash be enforced to guarantee that the url matches the one the relay expects? if one expects origin (without trailing slash) would that collide with the verification?

fiatjaf: I may be missing something, but don't think that is necessary at all, see: https://github.com/nostr-protocol/nips/blob/4a5202646a86ce63a85a50b560d1a2fbce277df6/42.md?plain=1#L67-L68

The relay must decode the URL _anyway_ and check if it is acceptable or not according to hostname, path, querystring etc. It doesn't have to match an exact pattern.

ismyhc:
Im for this nip 100%. I believe that a simple method to auth (prove ownership of keys) was missing from the protocol. I think this solves that as well as lends the ability for a relay to restrict access to events. Of course the relay still has access, but it's the clients choice to store where they see fit.

This gives me more hope for Seer :)

42.md
```diff
@@ -0,0 +1,68 @@
...
+```
+["OK", "b1a649ebe8...", false, "restricted: we do not accept events from unauthenticated users, please sign up at https://example.com/"]
+```
+
+## Signed Event Verification
+
+To verify `AUTH` messages, relays must ensure:
+
+  - that the `kind` is `22242`
+  - that the event was recently signed (~10 minutes, by `created_at`)
```
monlovesmango, line 66: can we be more explicit about how long we want authentication to last for to avoid confusion? or are we saying about 10 minutes bc it should be up to the relays how long auth lasts for?

fiatjaf: No, this is not for how long it will last. This is for when the event must be created. I think it is somewhat implied that the auth lasts for the duration of the websocket connection, but we can make that more explicit.

monlovesmango: ah ok that makes a lot more sense. yeah might be good to make it a bit more explicit as that implication went right over my head.

1440000bytes:
Concept ACK

42.md
phyro: Can different websocket connections reuse the same auth message?

monlovesmango: yes i would think so. as long as it is recent enough. but pretty trivial to sign a new auth message.

monlovesmango (approved):

42.md
```diff
@@ -0,0 +1,69 @@
...
+  - A relay may limit access to `kind: 4` DMs to only the parties involved in the chat exchange, and for that it may require authentication
+    before clients can query for that kind.
+  - A relay may limit subscriptions of any kind to paying users or users whitelisted through any other means, and require authentication.
+
+## Protocol flow
+
+This NIP defines a new message, `AUTH`, which clients can send to relays when they want to authenticate. The message is of the following form:
+
+```
+["AUTH", <signed-event>]
```
Giszmo, line 27: How do unexpecting relays react to this unknown command? They will probably reply with a notice?

So do I understand correctly that we are using AUTH and not EVENT to avoid this being sent to others who then could use it to auth? Which could only happen if the relay did only occasionally support this nip by maybe having different relay implementations behind some round-robin?

I think we could spare us this new command but tell me please what could go wrong ;)

Giszmo (approved):

42.md
fiatjaf: Yes, you should be able to send this command to unexpecting relays and no harm be done. I don't think they will answer with a notice, most likely they will ignore it, but if they do so be it, it is better than having your auth attempt be broadcasted to the world.

I don't see any advantage of using `EVENT` instead of `AUTH`. The relay will already have to have a very specific custom logic for handling auth events anyway, so I think it makes things clearer to have this handled in a separate place already from the get go.

fiatjaf: I think they shouldn't. Mine doesn't. Or maybe mine does, I don't remember, but I think it doesn't.

staab (approved):

jblachly:
Sorry to be a late entrant to the conversation. What is the rationale behind the combination of (~10 minute time window, URL of relay in content), versus a traditional challenge-response?

Challenge-response requires current possession of the key, whereas an unlimited number of these authentication messages could be pre-generated with temporary possession of a key (admittedly less likely than simply exfiltrating the key itself).

fiatjaf:
Challenge-response is more complex and there are no benefits whatsoever, or at least I am not aware of any. What are the benefits? "temporary possession of a key", as you noticed yourself, is not a real thing that exists.

davestgermain:
I've implemented NIP-42 in [nostr-relay](https://code.pobblelabs.org/fossil/nostr_relay/doc/tip/docs/authentication.md). Feel free to try it out.

jblachly:
> Challenge-response is harder to implement and there are no benefits whatsoever, or at least I am not aware of any. What are the benefits? "temporary possession of a key", as you noticed yourself, is not a real thing that exists.

Thanks for responding. In my view, a key benefit is that it (challenge-response) is widely understood among security professionals and widely deployed with a track record of safety.  The mechanism proposed in NIP-42 is uncommon, and has theoretical drawbacks. For example, in the future when keys are stored in hardware tokens (perhaps a TPM or Passkey on a phone), blind-signing could produce authentication tokens with future validity. I think it is a bad idea to design new security schemes when existing schemes are established. 

Kind regards

Cameri:
NACK
Any client could pre-sign with or without your knowledge a bunch of auth events and sign in as you in the future with or without your knowledge. A client could leak your signed AUTH events to an impersonator and they could authenticate as you.
I would avoid implementing this spec without some challenge-response that turns this flow into causal.

phyro:
I'll write down some differences:
1. **2-step vs 3-step protocol** - challenge/response is a 3-step protocol where where the user requests AUTH, gets challenge, proves challenge (note, it may be possible to do a 2-step auth if the relay sends a challenge automatically when connection is initialized). This nip is a 2-step protocol.
2. **auth context** - challenge/response is per connection, this nip is per pubkey. Reconnecting requires re-auth in both cases.
3. **security** - challenge/response comes with an unpredictable challenge while timestamps are predictable due to being a counter (while I can't see how this would be less secure, I wouldn't bet my life the security models are the same)
4. **ws connection MITM** - if you have a ws and not wss connection (nostr allows ws), a MITM AUTH is possible with this proposal because the AUTH can be reused with a different connection. Challenge/response is per connection so it doesn't have this problem afaict. Or does it?
5. **generalization** - an AUTH message has a clearly defined purpose. This has some tradeoffs. It's very predictable and simple to understand, but also potentially limiting. A more abstract scheme i.e. with PROVE/CHALLENGE/PROOF messages might allow for more creativity. This depends on the creativity though and whether we want the protocol to go in this direction.

I don't necessarily think all of the points above are worth considering. This is just a very quick breakdown to make the potential differences more clear. It's possible it contains mistakes, if you spot any, don't hesitate to correct me.

Cameri (changes requested):
Sorry, I don't think this can be trusted for the purpose of authenticating users unequivocally.

Cameri:
This should probably be:

1. Client connects to relay
2. Upon connection or if client sends restricted message: Server sends `[AUTH, challenge]` to client
3. Client creates HMAC with challenge as secret
4. Client hashes `relay_url:pubkey`, this is the response
5. Client sends a kind: 22242 ephemeral message with the content as the `response` to the challenge.
6. Server responds with `[OK, event, true, string]` if authenticated, otherwise `[OK, event, false, 'rejected: invalid auth response']`

davestgermain:
The challenge-response could be even simpler:

1. upon connection, relay sends `[“NOTICE”, “authentication required 12345”]` — the message can be human readable, with a random or timestamp component. 
2. When client wants to authenticate, make the content of the auth event the text of the NOTICE. 

No extra hashing required, and it’s resistant to replays. 

And clients can ignore the notice if they don’t support auth.

42.md
```diff
@@ -0,0 +1,68 @@
...
+  "content": "wss://relay.example.com/",
+  "sig": "..."
+}
+```
+
+The client may send an auth message right before performing an action for which it knows authentication will be required -- for example, right
+before requesting `kind: 4` chat messages --, or it may do right on connection start or at some other moment it deems best.
+
+Upon receiving a message from an unauthenticated user it can't fulfill without authentication, a relay may choose to notify the client. For
+that it can use a `NOTICE` message with a standard prefix `"restricted: "` that is readable both by humans and machines, for example:
```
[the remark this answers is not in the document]
fiatjaf: Yes.

42.md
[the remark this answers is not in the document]
fiatjaf: Yes, that is written somewhere else above after @monlovesmango's questioned the same thing.

fiatjaf:
> Sorry, I don't think this can be trusted for the purpose of authenticating users unequivocally.

Why?

Cameri:
> > Sorry, I don't think this can be trusted for the purpose of authenticating users unequivocally.
> 
> Why?

https://github.com/nostr-protocol/nips/pull/141#issuecomment-1374331848

There's no reason why Nostr relays should accept a weaker form of AUTH when we could have something similar to #85

fiatjaf:
> Any client could pre-sign with or without your knowledge a bunch of auth events and sign in as you in the future with or without your knowledge. A client could leak your signed AUTH events to an impersonator and they could authenticate as you.

If a client is malicious and has your private keys they can just store your private keys and sign whatever they want whenever they want. They don't have presign a bunch of AUTH events with different timestamps. What am I missing here?

fiatjaf:
> There's no reason why Nostr relays should accept a weaker form of AUTH

First you need an argument on why it is weaker. And yes, there is a reason to do this and not request-response which is the "best practice": simplicity of implementation.

But sure, if someone can come up with a reason why the current proposal scheme is flawed I will not insist.

Cameri:
> > There's no reason why Nostr relays should accept a weaker form of AUTH
> 
> First you need an argument on why it is weaker. And yes, there is a reason to do this and not request-response which is the "best practice": simplicity of implementation.
> 
> But sure, if someone can come up with a reason why the current proposal scheme is flawed I will not insist.

I've commented above why it is flawed but TL;DR: signed auth events without a challenge/response can be replayed by a third-party.

The client does not need your private key, it just needs to forward your signed auth event to someone else. That someone else can auth as you.

fiatjaf:
> The client does not need your private key, it just needs to forward your signed auth event to someone else. That someone else can auth as you.

OK, that makes sense.

What about we just tell relays to keep track of event ids used to sign, then? So they can't be replayed. Relays can keep track of them for ~10 minutes, after which their `created_at`s will become invalid.

Sure, relays may not do that, but that would be the same as relays issuing a challenge but not keeping track of it and then not checking the challenge in the body of the signed event they receive afterwards.

This provides the same level of security, but without requiring one extra step in the auth flow.

davestgermain:
> Sure, relays may not do that, but that would be the same as relays issuing a challenge but not keeping track of it and then not checking the challenge in the body of the signed event they receive afterwards.

It's much easier for a relay to send a NOTICE challenge per connection and remember that string, than it is to maintain a list of potentially thousands of recent event ids.

fiatjaf:
> It's much easier for a relay to send a NOTICE challenge per connection and remember that string, than it is to maintain a list of potentially thousands of recent event ids.

OK, you won.

monlovesmango:
I guess I am not so much worried about replay attacks. but I do think not having challenge response creates an incentive for attackers to create malicious apps that don't steal your keys but create a bunch of future auth events that can be used to request data that is supposed to be protected behind an auth request. obviously this incentive also depends upon how valuable the data protected by auth request is, which is hard to say.

I just think this is a bad incentive to have as it opens up an avenue for abusing access to event signing without the user being able to tell. with events the user can tell when something they didn't send goes out. with auth they have no way to know.

fiatjaf:
I tihnk I was overreacting. Challenges do not add as much complexity as I thought they would since websocket connections are already as stateful as it can get.

Please check my latest changes, patient gentlemen.

davestgermain:
> I tihnk I was overreacting. Challenges do not add as much complexity as I thought they would since websocket connections are already as stateful as it can get.

Nice. I don't understand the rationale for using tags rather than the content field, but no big deal.

I've already updated the python [implementation](https://code.pobblelabs.org/fossil/nostr_relay/info/830acdd97cf7d026)

phyro:
> At any moment the relay may send an `AUTH` message

This may add some uncertainty on the client side i.e. how does the client know when to expect the message? If I want to query my events, should I wait a minute for the relay to send the AUTH or should I try to query them? Since the client is the one doing the requests that may require AUTH, it may make sense to consider having AUTH message done when the connection is established, or allow the client to ask for the AUTH challenge.

fiatjaf:
The client doesn't know if the relay supports AUTH at all. So it should just query for stuff without AUTH if the relay hasn't sent anything yet. Doesn't that make sense?

I imagine all relays will send AUTH as soon as a connection is established, but clients shouldn't count on that being the case.

What will clients do differently if the challenge comes 2 seconds after the connection is initiated as opposed to 2 milliseconds?

phyro:
I guess it does. This makes me wonder if the relay should communicate the capabilities when connection is established i.e. a bitmap 111101110110... which describes the nip support. This way the client would at least not need to guess the future behaviour.

Another question. Can a single connection have a multi-auth by sending `["AUTH", <signed-event-json>]` multiple times with a different pubkey?

Cameri:
> I guess it does. This makes me wonder if the relay should communicate the capabilities when connection is established i.e. a bitmap 111101110110... which describes the nip support. This way the client would at least not need to guess the future behaviour.
> 
> Another question. Can a single connection have a multi-auth by sending `["AUTH", <signed-event-json>]` multiple times with a different pubkey?

I'm not a fan of this capability sharing and also sounds like a conversation for a different PR. We have NIP-11 that clients can already use with most relays to request capabilities out-of-band before they even initiate a connection.

AUTH as discussed here does not allow for multiple authenticated users to reuse the same connection, each will need their own. I'd recommend avoiding solving this now and push it to the future when it is actually needed. No single client supports multiple accounts today and is not something users are asking for.

42.md
```diff
@@ -0,0 +1,86 @@
...
+```
+
+And, when sent by clients, of the following form:
+
+```
+["AUTH", <signed-event-json>]
+```
+
+The signed event is an ephemeral event not meant to be published or queried, it must be of `kind: 22242` and it should have at least two tags,
+one for the relay URL and one for the challenge string as received from the relay. `created_at` should be the current time. Example:
```
Cameri, line 38:
```suggestion
The signed event is an ephemeral event not meant to be published or queried, it must be of `kind: 22242` and it should have at least two tags,
one for the relay URL and one for the challenge string as received from the relay.
Relays MUST exclude `kind: 22242` events from being broadcasted to any client.
`created_at` should be the current time. Example:
```

Cameri (approved):
LGTM. Left a explicit comment on what relays should do when a subscription filter matches `kind: 22242` events.

42.md
1440000bytes, line 38:
```suggestion
one for the relay URL and one for the challenge string as received from the relay. `created_at` should be the current time. Relays should use a maximum acceptable clock skew between its own clock and the client's clock, and to reject signed events with `created_at` timestamps that are outside of this acceptable range. Example:
```

1440000bytes (commented):
NIP looks better with the recent changes. Added one suggestion if there is a difference between client and relay clocks.

jblachly:
> I tihnk I was overreacting. Challenges do not add as much complexity as I thought they would since websocket connections are already as stateful as it can get.
> 
> 
> 
> Please check my latest changes, patient gentlemen.

I really appreciate you taking the time to consider my challenge, especially as this looked like a done deal when I stepped in. Salut

jmanero:
Hello All. I want to propose a question for consideration; apologies if this was discussed elsewhere:

Does the protocol _need_ to implement its own session authentication scheme in-band? Considering that WebSocket is an upgraded HTTP req/rep, would it make more sense to define usage of an existing HTTP signing authentication scheme (e.g. some usage of JWT) on the initiating HTTP request before upgrade?

Among other considerations, this could lower the complexity of authenticated relay implementation and operation by enabling use of  OSS libraries and/or third-party authorization gateways.

42.md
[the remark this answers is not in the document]
Giszmo: I don't understand Cameri's comment. Is this resolved? It's marked outdated but I see no REQ or CLOSE in the new diff.

42.md
```diff
@@ -0,0 +1,88 @@
...
+`created_at` should be the current time. Example:
+
+```json
+{
+  "id": "...",
+  "pubkey": "...",
+  "created_at": 1669695536,
+  "kind": 22242,
+  "tags": [
+    ["relay", "wss://relay.example.com/"],
```
Giszmo, line 49: If it's challenge-response, do we really need the relay here?

42.md
```diff
@@ -0,0 +1,88 @@
...
+
+## Signed Event Verification
+
+To verify `AUTH` messages, relays must ensure:
+
+  - that the `kind` is `22242`;
+  - that the event `created_at` is close (e.g. within ~10 minutes) of the current time;
+  - that the `"challenge"` tag matches the challenge sent before;
+  - that the `"relay"` tag matches the relay URL:
+    - URL normalization techniques can be applied. For most cases just checking if the domain name is correct should be enough.
```
Giszmo, line 88: With challenge-response we could omit prescribing all details about the event but the challenge-tag. Relays must not re-use nonces and all the rest doesn't need checking.

Lets leave it to the relays to limit the event to a reasonable size but it should really only care about the challenge-tag.

42.md
```diff
@@ -19,35 +19,51 @@ A relay may want to require clients to authenticate to access restricted resourc
...
 
-## Protocol flow
+## Definitions
+
+This NIP defines a new message, `AUTH`, which relays can send when they support authentication and clients can send to relays when they want
+to authenticate. When sent by relays, the message is of the following form:
+
+```
+["AUTH", <challenge-string>]
+```
```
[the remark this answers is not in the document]
fiatjaf: We can add that later if necessary.

42.md
fiatjaf: Yes, otherwise one malicious relay can reuse your auth event with it to login on a different relay.

42.md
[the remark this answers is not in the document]
Giszmo: So I would implement the challenge upon a socket being established. If the client somehow lost the nonce, it could always re-connect.

The nip is vague about when the relay might challenge indeed. Also a client might want to switch authentication, so while authed as Alice it might want to switch to auth as Bob now, so both for un-authed and authed clients there is a need to request a new nonce. Semisol's empty AUTH message triggering this has my support.

42.md
Giszmo, line 88: With challenge-response we could omit prescribing all details about the event but the challenge-tag. Relays must not re-use nonces and all the rest doesn't need checking.

Lets leave it to the relays to limit the event to a reasonable size but it should really only care about the challenge-tag.

```suggestion
To verify `AUTH` messages, relays must ensure that the event's `"challenge"` tag matches the challenge sent before.
```

42.md
```diff
@@ -0,0 +1,88 @@
...
+  "created_at": 1669695536,
+  "kind": 22242,
+  "tags": [
+    ["relay", "wss://relay.example.com/"],
+    ["challenge", "challengestringhere"]
+  ],
+  "content": "",
+  "sig": "..."
+}
+```
```
Giszmo, line 55: If it's challenge-response, do we really need to be so prescriptive about all these details?

```suggestion
The signed event is an ephemeral event not meant to be published or queried. It must have a tag with the challenge string as received from the relay. Example:

```json
{
  "id": "...",
  "pubkey": "...",
  "created_at": 1669695536,
  "kind": 22242,
  "tags": [
    ["challenge", "challengestringhere"]
  ],
  "content": "",
  "sig": "..."
}
```
```

File: 42.md [hunk 12]
```diff
@@ -0,0 +1,88 @@
...
+  "pubkey": "...",
+  "created_at": 1669695536,
+  "kind": 22242,
+  "tags": [
+    ["relay", "wss://relay.example.com/"],
+    ["challenge", "challengestringhere"]
+  ],
+  "content": "",
+  "sig": "..."
+}
```
Giszmo, line 54: If it's challenge-response, do we really need to be so prescriptive about all these details?

```suggestion
The signed event is an ephemeral event not meant to be published or queried. It must have a tag with the challenge string as received from the relay. Example:

```json
{
  "id": "...",
  "pubkey": "...",
  "created_at": 1669695536,
  "kind": 22242,
  "tags": [
    ["challenge", "challengestringhere"]
  ],
  "content": "",
  "sig": "..."
}
```

42.md
```diff
@@ -0,0 +1,88 @@
...
+```
+
```
Giszmo, line 56: A proposal for what @Semisol suggested.

```suggestion

Clients can also send an empty `["AUTH"]` message. With this, relays should end any existing authentication of the websocket and must offer a new authentication challenge.

```

Giszmo:
> Hello All. I want to propose a question for consideration; apologies if this was discussed elsewhere:
> 
> Does the protocol _need_ to implement its own session authentication scheme in-band? Considering that WebSocket is an upgraded HTTP req/rep, would it make more sense to define usage of an existing HTTP signing authentication scheme (e.g. some usage of JWT) on the initiating HTTP request before upgrade?
> 
> Among other considerations, this could lower the complexity of authenticated relay implementation and operation by enabling use of OSS libraries and/or third-party authorization gateways.

@jmanero I think we want to work with nostr identities. I also just suggested a way to switch identities mid-session. I want multi-account clients to be able to connect billing Alice and later Bob without re-connecting.

While doing it after the Websocket is established might be more expensive for the server hardware and thus leaves more potential for DOS attacks, we probably also need this. Maybe a later nip will be required for DOS protection that prevents the HTTP upgrade if not authed.

42.md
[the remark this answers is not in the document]
Cameri: I agree and it's so simple to implement we should try to have it from the start.

42.md
Giszmo: What is this a reply to? My suggestion to drop the commitment to a relay name?

If your relay doesn't re-use challenge nonces, it will associate each nonce with only one websocket connection ever. That websocket connection implicitly commits to the relay already, so naming the relay is not necessary.

arthurfranca:
TLDR: What if no challenge is involved and we use url query parameter instead of AUTH command for *initial* authentication and then if needed for switching account use client initiated AUTH command?
--
Hello, im a bit confused with the flow.

in a future that possibly the majority of relays will require authentication (e.g.: rate limiting by authenticated pubkey)
i imagine most relays will want to send AUTH+challengeX on connection

the clients will expect that the majority of relays would require authentication
so to not waste time by sending an EVENT/REQ that will most likely fail it will have two ways of dealing with it:
1. poke the relay by sending an empty (no arguments) AUTH message onopen which would make relay respond with AUTH+challengeY then client will send the 22242 event to authenticate
(but hey how much it will wait for the challenge that may never come and problem is maybe the AUTH+challengeX from relay on connection will arrive soon, which the client will have to handle by sending a new 22242 event)
2. wait for the relay AUTH+challengeX (problem is it may never come - so it will have to waste like 500ms before sending empty AUTH.
The 500ms may not be enough, meaning the same problem above may happen).

If we consider that the challenge is not that much needed (Use the _"What about we just tell relays to keep track of event ids used to sign"_ approach),
the flow could be by default the client will send the authentication AUTH message with 22242 event onopen (no previous client empty AUTH message onopen nor relay AUTH message on connection), which would just be ignored by relays without support.
Then it can make EVENT/REQs freely (hoping that the previous, maybe concurrent, message was completely handled first so to not receive a "restricted: " notice/commandResult. Maybe it should be better to add the 22242 event on a query paramenter on ws url like new WebSocket("wss://...?protocol=nostr&authorization=stringified_22242_event") to avoid the concurrency problem)
To switch account, the client can send AUTH message with 22242 event whenever it wants. The relay doesn't send AUTH messages at all.

42.md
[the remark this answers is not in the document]
fiatjaf: I think it is unnecessary bloat. How does that work on the relay side? The relay is emitting events for identity A at a given websocket, then that websocket changes to identity B, now the relay has to shove that inside the code it might have for handling the subscriptions, maybe cancel a subscription that was valid?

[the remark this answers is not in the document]
fiatjaf: The client should just open a new websocket if it wants a new identity -- which is also an edge case I see no use for and we shouldn't be complicating this for 99.9% of the cases.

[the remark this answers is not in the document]
fiatjaf: > The nip is vague about when the relay might challenge indeed

This is because it is irrelevant. All clients would just set up a listener like `when("AUTH") => send("AUTH")` and be done with it. Most relays will send the `AUTH` challenge on connect, but why specify that if it changes nothing on the code side (the client has to wait anyway)?

Also, if you think the client sending an empty `AUTH` request fixes it, consider that the relay may still take 5 minutes to reply to that with a challenge -- or may never reply, so really nothing is changed.

fiatjaf:
@arthurfranca I agree with you. Authentication with no challenge would work very well indeed, and the flows I had in mind were exactly these, but the argument on https://github.com/nostr-protocol/nips/pull/141#issuecomment-1374601500 is strong.

Giszmo:
The server could briefly cache auth events and reject replays if replays become an issue.

brugeman:
Sorry if I'm bringing up this almost approved NIP again.

Some thoughts:
1. Nostr protocol is async - client does not wait for responses and can send several requests, relay can reorder replies, and sometimes replies aren't even expected. Any number of messages from either party can be in flight.
2. It seems we're discussing two things: requiring auth to enforce rules (i.e. DM access, whitelist), and requiring auth to prevent abuse. The first one requires an identity, the second could be solved with an anonymous LN payment. 
3. Since this nip is about auth, I think abuse prevention should be out of scope of this nip. In fact, an authed identity might start misbehaving and then should either be dropped, banned or forced to pay, so we should plan accordingly for a separate payment-required nip. 
4. I think we should allow for non-authed connections to exist as much as possible for privacy and simplicity, and we should allow for relays to request auth only when it's absolutely needed, so auth using http-headers/query-string doesn't fit.

Suggestions, mainly concerning the "Protocol flow" section of the current nip text:

1. Clients can open a connection and start sending REQ/EVENT messages w/o expecting that relay requests an AUTH (like they do today). Even if client supports AUTH it can't know if relay would require it and when, so it should start sending messages right away, no need to wait for AUTH and guess when it might come. No changes for clients needed. 

2. When relay decides to require auth (either on conn open, or on receiving some type of request, or on high load, etc), relay must: 
- put all incoming auth-requiring messages from this conn into a buffer, 
- reply with AUTH+challenge, and may proceed serving non-auth-requiring requests from this conn, 
- wait for say 60 seconds for AUTH+reply to come, connection might be overloaded so let's not make this threshold too small
- after valid AUTH+reply is received relay must process all buffered messages and send proper replies
- after timer expires relay must drop the buffered messages, and reply with "OK:false:blocked" on all dropped EVENT messages, and EOSE on all dropped REQ messages. 

This should allow non-supporting clients to just keep working as they do. This should also allow for a simple all-conns-require-auth strategy for relays - send AUTH+challenge on conn open and buffer everything until AUTH+reply arrives.

3. If client was sending a bunch of EVENT messages and then relay decided it needs auth and sends AUTH+challenge, relay should not buffer _all_ incoming EVENT messages until AUTH+reply is received, instead if should buffer some of them and then if buffer gets full it should start dropping further messages and reply with OK:false:blocked:. Clients w/o auth support will work as they do, clients w/ support for auth will reply with AUTH+reply on AUTH+challenge and then may resend _once_ all events that were blocked after AUTH+challenge was received. Suggested buffer size for EVENT messages is 100kb. Since most clients don't send bulks of EVENT messages, they don't have to add the re-sending complexity, but limited buffering is a must for relays.

4. Relay should buffer all REQ messages while waiting for AUTH+reply, but up to a limit that should already be in place in all existing relays.

5. If client receives AUTH message and supports it then it replies w/ AUTH+reply. If client does not support or does not wish to auth then it ignores the AUTH+challenge and proceeds as usual. If clients choose to drop a connection on unsupported AUTH it's fine, this relay can't serve them anyway. 

6. If relay does not receive a valid AUTH+reply in time, relay must reject buffered messages as per p.2, and should proceed replying with EOSE and OK:false:blocked on all further auth-requiring requests. Relay may choose to disconnect clients that send too much of auth-requiring events. This way non-supporting or privacy-loving clients can just keep working as they do.

7. If client wants to switch identities, it sends AUTH+"". Relay should reset the connection's identity upon receiving an empty auth. This way multi-acc clients can terminate the previous session and just start requesting stuff for the next acc, and when relay requires an auth for stuff of the next acc - new AUTH+challenge will arrive.
 
8. If client already authed itself but then sends a message that requires a different identity, relay must send NOTICE: forbidden, then should send EOSE and OK:false:blocked. If client receives NOTICE:forbidden it may show that to the user and if user switches accs then client will send AUTH:"" and proceed. Relay may drop a connection if too many rejected messages arrive from the client.

I included a lot of details to explain the rationale, final text would be much shorter.

The main ideas:
- buffering is necessary if clients are not to be required to wait/request for AUTH+challenge
- this is flexible wrt whitelisting strategies for relays, and is privacy preserving and backward-compatible for clients
- it allows to switch identities for a multi-acc clients which is useful in the browser (desktop/mobile could just open several connections)

Does this make sense?

fiatjaf:
To the extent that I followed you it did make sense, @brugeman. I didn't see what exactly you're proposing though, it seems you're in agreement with the current text of this NIP, since it fits your protocol description flow already if I'm not mistaken.

brugeman:
Indeed I'm mostly in agreement.

I'm proposing:
- describe that buffering by relays is required btw AUTH+challenge and AUTH+reply, without it the flow will break
- AUTH:"" is added to reset the identity
- remove "The client may send an auth message right before performing an action for which it knows authentication will be required" - client can't send auth message unless it receives a challenge first
- change "Upon receiving a message from an unauthenticated user it can't fulfill without authentication, a relay may choose to notify the client. For that it can use a NOTICE" - instead, relay should _then_ send AUTH+challenge and buffer the message until client sends AUTH+reply. It could send NOTICE:restricted to force unsupporting clients to notify user, but it shouldn't reject the message it can't fulfill - it should buffer it and wait. 
- specify behavior if AUTH+reply does not arrive
- specify behavior if authed client requests stuff that requires a different identity

Giszmo:
> * specify behavior if authed client requests stuff that requires a different identity

I don't see the need for that. If you request the DMs for somebody else, receiving the same "auth required" should work.

But that brings me to another issue: If I need multiple auths for example because I want to watch the DMs of my 25 accounts, I'd either need many websockets or rotate a lot between accounts. Maybe we should not describe in the nip that AUTH+"" resets the auth but only that it requests a new challenge.

Best would be though to allow multiple signatures to not need a sequence of many challenge-response cycles.

brugeman:
> > * specify behavior if authed client requests stuff that requires a different identity
> 
> I don't see the need for that. If you request the DMs for somebody else, receiving the same "auth required" should work.

You might be right, that might be a minor difference not worth specifying.

> 
> But that brings me to another issue: If I need multiple auths for example because I want to watch the DMs of my 25 accounts, I'd either need many websockets or rotate a lot between accounts. Maybe we should not describe in the nip that AUTH+"" resets the auth but only that it requests a new challenge.

What if I want to switch my connection back to 'anon' mode? Besides, in this scheme it's the REQ/EVENT messages that might trigger AUTH+challenge by relay, why add another way?

> 
> Best would be though to allow multiple signatures to not need a sequence of many challenge-response cycles.

Why not allow several signed events in AUTH:reply? Several keys could sign the same challenge, then relay assigns several identities to this conn and only rejects stuff not one of them fits.

Giszmo:
> > But that brings me to another issue: If I need multiple auths for example because I want to watch the DMs of my 25 accounts, I'd either need many websockets or rotate a lot between accounts. Maybe we should not describe in the nip that AUTH+"" resets the auth but only that it requests a new challenge.
> 
> What if I want to switch my connection back to 'anon' mode? Besides, in this scheme it's the REQ/EVENT messages that might trigger AUTH+challenge by relay, why add another way?

You can't undo authentication. You might want to logout. That doesn't make you anon. The relay would still know who you authed with.

> > Best would be though to allow multiple signatures to not need a sequence of many challenge-response cycles.
> 
> Why not allow several signed events in AUTH:reply? Several keys could sign the same challenge, then relay assigns several identities to this conn and only rejects stuff if not one of them fits.

agreed

Cameri:
I think this PR is getting into excessive bike-shedding territory.

As a relay I wouldn't care what relay URL you pass or if you pass the wrong one or any other decorative artifact, the only thing that matters is that you respond with the challenge the relay sent you, that is the only proof needed.

We should not be talking here about multi-auth, let's keep it simple people. If you want multi-auth for the time being just open a new connection and be aware that relays may also limit how many connections come from the same IP.
Multi-auth over the same connection makes little sense as the users will just get throttled/rate limited.

Mandating relays to buffer pending AUTH should not be part of the protocol, let alone something that any relay should be doing. Why should we occupy resources for an unknown user is beyond my comprehension, sounds like a recipe for DoSing.

I suggest we proceed with the NIP as-is.

jmanero:
I think @brugeman has pointed out a lot of inherent complexity that the current state of this proposal could add to relay implementations, and the non-determinism that a client AUTH message would add to the protocol.

An alternative to the proposed buffering scheme and asynchronous AUTH exchanges that would also simplify support for multiple identities would be to revisit adding signing support to REQ messages (#20). This could be achieved by adding an optional `auth` top-level key to the `filter` object which would contain one or more `authentication` objects for `author` identities in the respective filter:

```json
{
  "authors": [...],
  "kinds": [4, ...],
  "auth": [
    {
      "id": "<32-bytes sha256 of the the serialized auth data>",
      "pubkey": "<32-bytes hex-encoded public key of the author>",
      "challenge": "<challenge string associated with WS session from server>",
      "sig": "<64-bytes signature of the sha256 id hash>"
    }
  ]
}
```

- `challenge` is a string produced by the relay that is unique to a single WebSocket connection. It may be reused by the client in multiple `authentication` objects sent on the respective WebSocket. The relay may rotate a WebSocket session's `challenge` value periodically, as defined below.

Initial Flow:

1. Client connects to really and sends one or more NIP-01 `REQ` messages
2. For each `REQ` message that the relay determines requires authentication, an `AUTH` message is sent to the client:

    ```json
    ["AUTH", "<subscription-id>", "<challenge>", { "authors": ["<pubkey of author that requires authentication>"]}, /* additional, possibly empty, objects for each `filter` in respective REQ */]
    ```

  - `subscription-id` correlates to the `REQ` message that triggered the authentication challenge
  - `challenge` is the replay-prevention value assigned to the WebSocket session by the relay. This should be the same for multiple `AUTH` messages sent to the same WebSocket in close succession, e.g. within a several-minute window.
  - Additional entries in the message are objects that correlate to the set of one-or-more `filter` entries in the respective `REQ` message.
    - `authors` is the set of pubkeys for the respective `filter` that need to be authenticated
    - The protocol could be extended in the future to support additional parameters

3. Some options...
  - A permissive relay _COULD_ begin streaming `EVENT` messages matching a `REQ`'s filter(s) that are not in-scope for authentication (e.g. types `1` and `2`, but not `4`) to support non-auth-aware clients. The relay would simply reject entries in the `filter` object, but otherwise carry on as as usual as far as the client can tell. In this case, the relay _SHOULD_ send a human-readable `NOTICE` message in addition to the `AUTH` message informing an older client that the `REQ` has been altered and authentication may be required.
  - Alternatively, a strict relay _COULD_ reject the `REQ` with an `OK:false` message in addition to the `AUTH` message. This approach wouldn't create any new state on the relay.
4. An auth-aware client, upon receiving an `AUTH` message, _MUST_ re-generate the respective `REQ` message with `authentication` entries for each `author` in the `AUTH`, and re-send the `REQ`.
  - The client _SHOULD_ use the same `subscription-id` to replace an existing partial-req in the case of the "permissive relay", above.
  - Clients _MAY_ send new filters in the `REQ` with `authentication` entries using the same `challenge`.
6. The relay should re-process the `REQ`, validating `authentication` entries on each `filter` entry, and replace any existing filters for the `REQ`'s `subscription-id`, per NIP-01. 
7. The relay _MUST_ send an `OK` message with appropriate `true`/`false` status for all of the validity of the `authentication` entries in the `REQ`


At this point, the client _COULD_ send more `REQ` messages with `authentication` entries reusing the `challenge`, and the relay _MAY_ reply with `AUTH` messages with a new `challenge`, and/or for `filters` missing required `authentication` entries.

Clients _MAY_ safely cache and reuse `authentication` objects on the same WebSocket session until an `AUTH` message provides a new `challenge` value.

fiatjaf:
I agree with everything @cameri said, except

> As a relay I wouldn't care what relay URL you pass or if you pass the wrong one or any other decorative artifact, the only thing that matters is that you respond with the challenge the relay sent you, that is the only proof needed.

You should care, because otherwise another relay can impersonate a user and login to your relay using an event provided by them. See:

1. User X connects to relay A
2. Immediately relay A poses as a user and connects to relay B
3. Relay B sends a challenge to user A
4. Relay A sends the same challenge to user X
5. X signs the challenge and sends to relay A
6. User A sends the same event to relay B
7. Now A can read X's messages from relay B and other stuff

fiatjaf:
But to your bikeshedding point, I think we just need people to implement this. No one has had the time or will to do that yet, but once an implementation happens we can stop the discussions.

Or should we merge this as is?

I think we should probably merge even without implementations, just so people feel safer to implement.

Cameri:
> I agree with everything @Cameri said, except
> 
> > As a relay I wouldn't care what relay URL you pass or if you pass the wrong one or any other decorative artifact, the only thing that matters is that you respond with the challenge the relay sent you, that is the only proof needed.
> 
> You should care, because otherwise another relay can impersonate a user and login to your relay using an event provided by them. See:
> 
>     1. User X connects to relay A
> 
>     2. Immediately relay A poses as a user and connects to relay B
> 
>     3. Relay B sends a challenge to user A
> 
>     4. Relay A sends the same challenge to user X
> 
>     5. X signs the challenge and sends to relay A
> 
>     6. User A sends the same event to relay B
> 
>     7. Now A can read X's messages from relay B and other stuff

Thanks for explaining it. That makes total sense.

arthurfranca:
> > Then it can make EVENT/REQs freely (hoping that the previous, maybe concurrent, message was completely handled first so to not receive a "restricted: " notice/commandResult. Maybe it should be better to add the 22242 event on a query paramenter on ws url like new WebSocket("wss://...?protocol=nostr&authorization=stringified_22242_event") to avoid the concurrency problem)
> 
> I believe there is an expectation that the AUTH command is not async.

@Semisol i think that the received AUTH command may be handled by an async onmessage function (e.g. an async schnorr.verify step to validate the 22242 event or any other async validation in place will make the function async). So a REQ command may arrive and be blocked because the authentication hasn't finished yet.

> The server could briefly cache auth events and reject replays if replays become an issue.

@Giszmo @fiatjaf doesn't this make the challenge unnecessary? again, if it is the case, the client could try to authenticate upon connection using an url query parameter, then the client will go on with REQ/EVENT commands if the connection isn't dropped by the relay.

phyro:
> and it should have at least two tags, one for the relay URL and one for the challenge string as received from the relay

Do we gain anything by having also the relay url in the tag? If the challenge is random 32 bytes from the relay, do we need the url? If we do believe committing to the url makes sense, is it different if we commit to the url in the challenge i.e.
```
challenge = H(relay_url || random)
```

fiatjaf:
Yes, see here: https://github.com/nostr-protocol/nips/pull/141#issuecomment-1382937209

phyro:
Should've read the comments, good point. It seems the client needs to commit to a connection it is authorizing either by adding the url to the tag or the challenge.

brugeman:
> I think this PR is getting into excessive bike-shedding territory.
> 
> We should not be talking here about multi-auth, let's keep it simple people. If you want multi-auth for the time being just open a new connection and be aware that relays may also limit how many connections come from the same IP. Multi-auth over the same connection makes little sense as the users will just get throttled/rate limited.

Agreed.

> 
> Mandating relays to buffer pending AUTH should not be part of the protocol, let alone something that any relay should be doing. Why should we occupy resources for an unknown user is beyond my comprehension, sounds like a recipe for DoSing.

Without buffering, this is how work flow looks for a client in this PR:
- connect
- send a bunch of REQs or EVENTs
- get AUTH+challenge from relay
- reply AUTH+reply 
- auth request from relay implies that every message I sent before it was dropped - I don't really know which one caused auth request, well maybe relay sends OK:false or a NOTICE, but even that might not happen
- so, I need to resend all the messages that I send before the AUTH+challenge arrived, hoping that now they will work fine

So either server buffers, or clients are all forced to implement resending their reqs. 

I'm not saying that buffering is great for relay - it's indeed a DoS vector. But without it this PR forces complex behavior on clients.

> 
> I suggest we proceed with the NIP as-is.

This statement in PR makes no sense - client needs challenge before it can auth:
> The client may send an auth message right before performing an action for which it knows authentication will be required

brugeman:
> But to your bikeshedding point, I think we just need people to implement this. No one has had the time or will to do that yet, but once an implementation happens we can stop the discussions.
> 
> Or should we merge this as is?
> 
> I think we should probably merge even without implementations, just so people feel safer to implement.

I implemented this at wss://relay.nostr.band/restricted (it will send AUTH+challenge on any incoming message), any client willing to try it?

Giszmo:
> > The server could briefly cache auth events and reject replays if replays become an issue.
> 
> @Giszmo @fiatjaf doesn't this make the challenge unnecessary? again, if it is the case, the client could try to authenticate upon connection using an url query parameter, then the client will go on with REQ/EVENT commands if the connection isn't dropped by the relay.

That's a very good point. No back-and-forth needed thanks to the commitment. Relays would have to store event IDs (if there is no malleability) until 10 (or whatever the time constraints we define) minutes after AUTH was received. Shouldn't have much DOS potential.

fiatjaf:
@Cameri should we go back to the version without challenge and ask relays to store used events? That would simplify things on the client and the logic on the relay wouldn't be very different.

Cameri:
@brugeman 
> I'm not saying that buffering is great for relay - it's indeed a DoS vector. But without it this PR forces complex behavior on clients.

Keeping relays simple and clients complex is the appropriate trade off here. Relays need to handle thousands of clients and requests already and every microsecond matters on their end. Clients can simply send the last request or event again once authed

> > I suggest we proceed with the NIP as-is.

> This statement in PR makes no sense - client needs challenge before it can auth:

Agreed. It doesn't make sense. @fiatjaf explained why on a subsequent comment.
.

Cameri:
> @Cameri should we go back to the version without challenge and ask relays to store used events? That would simplify things on the client and the logic on the relay wouldn't be very different.

Should clients be simple now and relays complex? Seems like we all forgot something here hehe

jmanero:
> I think the people just proposing a challenge are missing the point that they can forward the challenge to someone on another relay and impersonate them. Anyway, a few requests:

Isn't the challenge supposed to be scoped to a single WebSocket session, which is by definition limited to a single relay instance?

Cameri:
@Semisol 
> I think the people just proposing a challenge are missing the point that they can forward the challenge to someone on another relay and impersonate them. Anyway, a few requests:

Challenge + Relay URL solves that problem.

davestgermain:
> I think the people just proposing a challenge are missing the point that they can forward the challenge to someone on another relay and impersonate them. Anyway, a few requests:

That would be a malicious client.   
A malicious client could also read your decrypted DMs and forward them to the NSA.

Cameri:
> > I think the people just proposing a challenge are missing the point that they can forward the challenge to someone on another relay and impersonate them. Anyway, a few requests:
> 
> That would be a malicious client. A malicious client could also read your decrypted DMs and forward them to the NSA.

Cameri:
Sorry, misclicked!

fiatjaf:
> Should clients be simple now and relays complex? Seems like we all forgot something here hehe

I don't follow. Do you think relays should be simple and clients complex or vice-versa?

I'm just curious, because regardless of your answer, I don't think keeping track challenges issued by the relay to each client is less complex than keeping track of used challenges in the last 10 minutes, or am I wrong?

To keep track of used events one can just use a global like

```javascript
var challenges = []
setInterval(() => {
  challenges = challenges.slice(challenges.findIndex(([d]) => d < Date.now() - 60 * 10 * 1000))
}, 60 * 3 * 1000)

// then when a client sends an AUTH message
let challenge = authEvent.tags.find(([t, v]) => t === 'challenge')[1]
if (challenges.find(([_, c]) => c === challenge)) throw new Error('bad auth: reused challenge')
challenges.push([Date.now(), challenge])
```

Now to keep track of challenges issued to each client it would be something like

```javascript
var challenges = {}

// when client connects
let challenge = somehowGenerateRandomString()
challenges[ws] = challenge

// then when client sends an AUTH message
let challenge = authEvent.tags.find(([t, v]) => t === 'challenge')[1]
let expected = challenges[ws]
if (!expected || challenge !== expected) throw new Error('bad auth: reused challenge')

// then when client disconnects
ws.on('close', () => {
  delete challenges[ws]
})
```

davestgermain:
> I'm just curious, because regardless of your answer, I don't think keeping track challenges issued by the relay to each client is less complex than keeping track of used challenges in the last 10 minutes, or am I wrong?

What if your relay is multi-process? 

What happened to the idea of keeping the challenge scoped to the websocket connection? There's no need to remember "used challenges". The relay just needs to keep track of one challenge per connection. For instance, my python relay is issuing challenges in the form of '<client_ip_address>:<random 4 bytes>'.

Cameri:
@fiatjaf this wouldn't work on Nostream since new connections are handled by a different process and the memory isn't shared. The challenges would have to be cached using Redis or on a primary process and it could lead to race conditions if locks are not used.

fiatjaf:
You mean each connection has a different process? Don't you already use Redis in this case? With Redis this is much easier. Just save the challenge on Redis and make it expire after 10 minutes.

fiatjaf:
> What if your relay is multi-process?

Then you store the used challenges on Redis and that is super easy.

> What happened to the idea of keeping the challenge scoped to the websocket > connection? There's no need to remember "used challenges".

This is for the version of this NIP that doesn't require the relay to send a challenge to the client.

Cameri:
Nostream would need to use a lock to make sure only one authentication attempt can use the challenge (which is now a shared resource) so it will slow down considerably. If locks are not used then a race condition is theoretically possible where more than 1 client can auth if timed correctly.

davestgermain:
> You're getting it wrong. Relay A could connect to relay B, get a challenge, send that to you, and you sign it, which then allows relay A to impersonate you on relay B

True, and that would be a malicious relay...

So, keep the relay domain in the tags, as the current proposal states.

Cameri:
> > Nostream would need to use a lock to make sure only one authentication attempt can use the challenge (which is now a shared resource) so it will slow down considerably. If locks are not used then a race condition is theoretically possible where more than 1 client can auth if timed correctly.
> 
> ... each socket has its own challenge and auth state. there is no one challenge that is shared across all clients

I'm responding to Fiatjaf's suggestion on caching challenges so they are not re-used, it introduces a "double-spend" problem that needs solving. If the challenge is per socket we don't need to check any cache, cause every connection has it's own challenge.

fiatjaf:
https://github.com/nbd-wtf/go-nostr/commit/9775016bf19c871d515bec293c163f9f27adef77

fiatjaf:
https://github.com/fiatjaf/relayer/commit/fe91197d45578e227995b52fe1ea9b9b56ff0eb3

fiatjaf:
I'm going to merge this if no one objects until tomorrow.

Cameri:
> > (Except @Semisol's objections that we must have a way for the client to initiate the process, which I don't think is reasonable as I think I've explained before and it can always be added if it proves necessary.)
> 
> It's just a handler for the AUTH command that starts the authentication flow. It's simple so why not have it in the initial spec? Most relays will probably have a dedicated function to call for doing authentication anyway so the implementation should be at most a few lines.

I'll support it. It's good for testing as well.

fiatjaf:
"Why not?" is the source of all protocol bloat.

alexgleason:
This NIP could be expanded to let relays reject creating events from a pubkey you didn't authenticate with. Currently nothing prevents me from taking someone else's signed event from one relay and putting it onto another relay.

fiatjaf:
> This NIP could be expanded to let relays reject creating events from a pubkey you didn't authenticate with.

They can already do this if they want. It's a relay policy.

PR #142, opened 2023-01-03
NIP 1: Hashtags

dkobia:


fiatjaf:
This is what is being done by multiple clients already, right? It is good, but must be another NIP.

dkobia:
Makes sense @fiatjaf

PR #157, opened 2023-01-07
NIP-26: Change example condition to expire at a future date

kdmukai:
I'm not sure what the current `created_at>1640995200` condition is accomplishing (unless there's some worry that relays might accept back-dated events?).

It seems more useful to demonstrate that the delegation token can expire after X time (e.g. a 30-day authorization).

### DIFF
--- a/26.md
+++ b/26.md
@@ -34,7 +34,7 @@ nostr:delegation:<pubkey of publisher (delegatee)>:<conditions query string>
 For example, the token `c33c88ba78ec3c760e49db591ac5f7b129e3887c8af7729795e85a0588007e5ac89b46549232d8f918eefd73e726cb450135314bfda419c030d0b6affe401ec1` is signed by `86f0689bd48dcd19c67a19d994f938ee34f251d8c39976290955ff585f2db42e` and consists of:
 
 ```json
-nostr:delegation:62903b1ff41559daf9ee98ef1ae67cc52f301bb5ce26d14baba3052f649c3f49:kind=1&created_at>1640995200
+nostr:delegation:62903b1ff41559daf9ee98ef1ae67cc52f301bb5ce26d14baba3052f649c3f49:kind=1&created_at<1700995200
 ```
 
 #### Example
@@ -51,7 +51,7 @@ Below is an example of an event published by `62903b1ff41559daf9ee98ef1ae67cc52f
     [
       "delegation",
       "86f0689bd48dcd19c67a19d994f938ee34f251d8c39976290955ff585f2db42e",
-      "kind=1&created_at>1640995200",
+      "kind=1&created_at<1700995200",
       "c33c88ba78ec3c760e49db591ac5f7b129e3887c8af7729795e85a0588007e5ac89b46549232d8f918eefd73e726cb450135314bfda419c030d0b6affe401ec1"
     ]
   ],
--- a/26.md
+++ b/26.md
@@ -25,40 +25,58 @@ This NIP introduces a new tag: `delegation` which is formatted as follows:
 
 ##### Delegation Token
 
-The **delegation token** should be a 64-bytes schnorr signature of the sha256 hash of the following string:
+The **delegation token** should be a 64-byte Schnorr signature of the sha256 hash of the following string:
 
 ```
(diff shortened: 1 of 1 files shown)


fiatjaf:
I agree with this change, but you must change the signature too to be valid.

kdmukai:
The NIP doesn't include any of the PKs for the example delegator or delegatee pubkeys, so I can't redo any of the sigs.

I guess I'll just generate new example keys and update everything.

fiatjaf:
Sorry for asking that. If you can publish all the keys such that it serves as a poor man's test vector that will be great.

kdmukai:
@fiatjaf, done! Pretty satisfying to add the PKs and tighten up the presentation a bit!

fiatjaf:
Thank you very much!

Aspie96:
I feel that the way in which conditions can be specified (i.e. the possible operators) should be more specific. I assume it's "=" for "kind" and "<" or ">" for "created_at", but the specification doesn't specify this.

The specification should also specify whether multiple conditions are allowed on the same field (they should be, for instance to specify a validity range for "created_at").

fiatjaf:
I agree, @Aspie96. The spec is not very specific. If you can contribute with clarifications that will be great.

zackwy:
It's worth noting that this example would now allow the delegator to publish events on your behalf retroactively. We could avoid that by specifying that we allow multiple conditions and include both a `created_at<` and `created_at>` condition.

The conditions are definitely pretty vague at present, but the rules that @Aspie96 have suggested are essentially what we've been operating under.

kdmukai:
> It's worth noting that this example would now allow the delegator to publish events on your behalf retroactively.

Should relays accept back-dated events? I realize we can't guarantee all relays' behavior, but it seems like accepting back-dated events (within some reasonable fudge factor) should be discouraged.

Or is there a use case for back-dated events? Some kind of contingent publishing (I can prove I signed it X months ago but have been sitting on it... for some reason)?

Aspie96:
> Should relays accept back-dated events?

I think there are cases in which it absolutely makes sense to accept back-dated events.

First, not all clocks are perfectly synchronized, so there must be at least a little bit of tolerance in this sense.

Secondly, a relay may allow a user to mirror its past events. This possibility is actually important to prevent old events from being lost if old relays fall out of use in favor of newer ones.

However, this is irrelevant for this specific issue, because the delegate party and the relay may be the same party. And in fact, they are in the case of Minds.

Minds could definitely publish back-dated events on itself, regardless of what Nostr specifications say about it. But if the certificate is invalid for back-dated events, Nostr clients will not validate such back-dated events, which is the intention.

26.md
```diff
@@ -25,41 +25,59 @@ This NIP introduces a new tag: `delegation` which is formatted as follows:
...
-  "content": "Hello world",
-  "sig": "cd4a3cd20dc61dcbc98324de561a07fd23b3d9702115920c0814b5fb822cc5b7c5bcdaf3fa326d24ed50c5b9c8214d66c75bae34e3a84c25e4d122afccb66eb6"
+  "content": "Hello, world!",
+  "sig": "55ed9a78d6449b8c189b6dbc34bc4bcd34dcc79e6da6c9078268fe3d7c0cbe62b1b907ffb76ba591e83895b1329bf2e6e16f3b0cd5827272e420d419c6f0f0b5"
 }
 ```
 
+The event should be considered a valid delegation if the conditions are satisfied (`kind=1` and `created_at<1675721813` in this example) and, upon validation of the delegation token, are found to be unchanged from the conditions in the original delegation string.
+
+Clients should display the delegated note as if it was published directly by the delegator (8e0d3d3e).
```
Aspie96, line 79: I'm commenting here even though this has been merged because I don't know if this is worth an issue in itself.

> Clients should display the delegated note as if it was published directly by the delegator (8e0d3d3e).

I don't necessarily completely agree with this. While displaying the note exactly as if it had been published by the delegator is a perfectly acceptable behavior, I think there other reasonable options that a client could chose (and it should be up to the implementation).

A client could chose to notify the user, in some way, for instance by highlighting it, or by displaying an icon, that the event has been delegated and also mention the public key of the delegatee. A client could also chose not to trust certain user-selected delegatees and highlight those events more prominently or hide them.

Issue #162, opened 2023-01-09
Decentralizing the NIPs registry


fiatjaf:
Would be great if people could agree on NIPs and publish them on Nostr itself, without having to request permission from a cabal of maintainers of a specific GitHub repository.

Then people could see NIP proposals and say they endorse them or not, or point to implemented NIPs by their ids from client and relay implementations.

This could be chaotic at the beginning but I think it would be healthier to the protocol. It would also be nice to not lend all the NIPs an equal status of "officialness" just by them being in this repository.

The sad thing would be that we would lose the cool-sounding NIP numbers.

fiatjaf:
To be honest there are some NIP proposals that I don't like at all and don't want to merge, and while I try to say to myself that I should merge them if there are some people wanting them and they're not completely absurd, I feel like when merging I immediately award them with a special status when they shouldn't have that -- no NIPs should.

tmathews:
I agree, developers should feel free  to experiment and try things. They can release their own NIP for what they have created and if it catches on great. The repo/pr/issue process is acting as a gateway rather than a documentation.

andrerfneves:
Yes to this. Make NIPs great again

mikedilger:
I'm worried.

I get the argument @fiatjaf  that you don't want to be gatekeeper and don't want to be seen as blessing these NIPs. But I think if we don't have some kind of partial consensus or partial centralization, interoperablity will be at risk.

So long as whatever replaces this serves the purpose of extending nostr and also allows us to continue to be interoperable, then my concerns are addressed.

1440000bytes:
Concept ACK

Bitcoin developers tried doing it but failed<sup>1<sup> although some proposals for Bitcoin exist outside https://github.com/bitcoin/bips repository and implemented in different projects.

Would be a great if nostr protocol has no official registry for NIPs however proposals still get discussed. Instead of NIP numbers, developers can use present year and a unique name for it. Example: `NIP 2023-Ostrich`

<sup>1<sup>Related emails: 

https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2021-April/018859.html

https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2021-April/018868.html

https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2021-April/018869.html

https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2021-April/018871.html

eskema:
> developers can use present year and a unique name for it

[...]

PR #164, opened 2023-01-10
NIP-23 Revisions (relay list)

dskvr:
This is a PR into #32 that simplifies the specification, while encouraging the extension of this NIP, such as advanced rules.  These modifications are made in light of comments from several people since August 2022, which seemed to lead to [this comment](https://github.com/nostr-protocol/nips/pull/32#issuecomment-1365394029)

Will refer to this as NIP-23b for now. 

The changes made in this PR reduce the original spec drastically, resulting in a specification that more accurately described as a transition from using recommended relays (kind 2) as relay lists to using a more appropriate kind in the _replaceable kind-space_. This should make adoption and transition for clients extremely simple. 

The only differences between `kind 10001` [NIP-23b] and `kind 2` are as follows
- the values of `read` and `write` go from type `Boolean` to type `String`. The reason is to support extension without type handling (this is also different from [fiatjaf's comment linked above](https://github.com/nostr-protocol/nips/pull/32#issuecomment-1365394029) )
- `kind 10001` is a replacable kind, so clients can much more easily and confidently persist and retrieve changes

@fiatjaf please review.

### DIFF
--- a/23.md
+++ b/23.md
@@ -40,7 +40,7 @@ _Invalid event.content: missing `write` member_
 "{\"wss://relayer.somewhere.xyz\":{\"read\": \"true\"}}"
 ```
 
-### Purposes
+### Use Cases
 
 This NIP serves two purposes: (1) backup and interoperability of relay lists and relay specific rules between clients; and (2) sharing of relay URLs between users.
 
--- a/23.md
+++ b/23.md
@@ -30,10 +30,6 @@ _Valid event.content_
 "{\"wss://relayer.somewhere.xyz\":{\"read\": \"true\",\"write\": \"true\"}}"
 ```
 
-_Invalid event.content: trailing slash_
-```
-"{\"wss://relayer.somewhere.xyz/\":{\"read\": \"true\",\"write\": \"true\"}}"
-```
 
 _Invalid event.content: missing `write` member_
 ```
--- a/23.md
+++ b/23.md
@@ -12,6 +12,8 @@ This kind is identical to kind 2, except that it is a transition to the replacab
 
 The tags array is empty.
 
+The object member key, being the relay URL, **must** omit the trailing slash. Good: `wss://nostr.someone.xyz`, Bad: `wss://nostr.someone.xyz/` 
+
 The `read` value operates on the values of the [NIP-01](01.md) **filter** object, while the `write` rule operates on the values of the **event** object. 
 
 When the `read` value equals `false` for a given **filter** the client SHOULD NOT send that filter in a `REQ` message to that relay.
--- a/23.md
+++ b/23.md
@@ -46,7 +46,7 @@ This NIP serves two purposes: (1) backup and interoperability of relay lists and
 
 The first use case is meant to make it so users can open their client -- or different clients -- in different devices and have their list of relays automatically fetched from a default relay and start using their relays list without having to set everything up again.
 
-For the second purpose, if any client decides to, they can show to the user what relays other users are using, suggest that or automatically add these to the user's relay list. The possibility of sharing a list of relays in standardized format is good for spreading information about relays and contributes to the censorship-resistance of the network.
+Another use case is to display which relays other users are using, suggest connecting to them or automatically add these to the user's relay list. The possibility of sharing a list of relays in standardized format is good for spreading information about relays and contributes to the censorship-resistance of the network.
 
 ### Use cases
   - User selects relays from any number of sources and manages their relays' read and write capabilities.
--- a/23.md
+++ b/23.md
@@ -8,7 +8,7 @@ Relays List
 
 A special event with kind `10001`, meaning "relay list" is defined as a stringified content field containing an object where each member represents a relay the author uses.
 
-This kind is identical to kind 2, except that it is a transition to the replacable-kind-space [kind range: 10000-29999]
+This kind is similar to kind 2, but migrates towards kind 10001 to discourage relay bloat and encourage more consistent implementation. 
 
 The tags array is empty.
 
--- a/23.md
+++ b/23.md
@@ -37,9 +37,11 @@ This NIP can be extended at a later date to include more advanced rules while re
 ### Schema
 
 _literal_
-
 ```json
 {
+  "id": "<hash>",
+  "signature": "<signature>",
+  "created_at": "1673363256", 
   "kind": 10001,
   "tags": [],
   "content": {
@@ -54,6 +56,9 @@ _literal_
 _stringified_ 
 ```json
 {
+  "id": "<hash>",
+  "signature": "<signature>",
+  "created_at": "1673363256", 
   "kind": 10001,
   "tags": [],
   "content": "{\"wss://relayer.somewhere.xyz\":{\"read\": \"true\",\"write\": \"true\"}}"
--- a/23.md
+++ b/23.md
@@ -8,7 +8,7 @@ Relays List
 
 A special event with kind `10001`, meaning "relay list" is defined as a stringified content field containing an object where each member represents a relay the author uses.
 
-This kind is identical to kind 3, except that it is a transition to the replacable-kind-space [kind range: 10000-29999]
+This kind is identical to kind 2, except that it is a transition to the replacable-kind-space [kind range: 10000-29999]
 
 The tags array is empty.
 
@@ -32,7 +32,7 @@ For the second purpose, if any client decides to, they can show to the user what
 
 
 ### Extension possiblities 
-This NIP can be extended at a later date to include more advanced rulesets and remain backwards compatible.
+This NIP can be extended at a later date to include more advanced rules while remaining backwards compatible.
 
 ### Schema
 
--- a/23.md
+++ b/23.md
@@ -10,8 +10,6 @@ A special event with kind `10001`, meaning "relay list" is defined as a stringif
 
 This kind is similar to kind 2 in format, but would live at kind 10001 to discourage relay bloat and encourage more consistent implementation. 
 
-The key for each object member is the `relay URL` and **must** omit the trailing slash. 
-
 The tags array is empty.
 
 The `read` value operates on the values of the [NIP-01](01.md) **filter** object, while the `write` rule operates on the values of the **event** object. 
@@ -22,18 +20,11 @@ Both the `read` and `write` values are strings, to enable extensibility and type
 
 When the `read` value equals the string `true`  then read from that relay, otherwise don't. 
 
-When the `write` value equals the string `true`  then write to that relay, otherwise don't.
-
+When the `write` value equals the string `true`  then write to that relay, otherwise don't
 
 _Valid event.content_
 ```
 "{\"wss://relayer.somewhere.xyz\":{\"read\": \"true\",\"write\": \"true\"}}"
-```
-
-
-_Invalid event.content: missing `write` member_
-```
-"{\"wss://relayer.somewhere.xyz\":{\"read\": \"true\"}}"
 ```
 
 ### Use Cases
--- a/23.md
+++ b/23.md
@@ -16,37 +16,23 @@ The key, being the relay URL, **must** omit the trailing slash.
 
 The `read` value operates on the values of the [NIP-01](01.md) **filter** object, while the `write` rule operates on the values of the **event** object. 
 
-When the `read` value equals `false` for a given **filter** the client SHOULD NOT send that filter in a `REQ` message to that relay.
+When the `read` value equals the string `false` for a given **filter** the client SHOULD NOT send that filter in a `REQ` message to that relay.
 
-When the `write` value equals `false` for a given **event** the client SHOULD NOT send that event in an `EVENT` message to that relay.
+When the `write` value equals the string `false` for a given **event** the client SHOULD NOT send that event in an `EVENT` message to that relay.
 
-_Valid_
+_Valid event.content_
 ```
-"content": {
-  "wss://relayer.somewhere.xyz": {
-    "read": "true",
-    "write": "true"
-  }
-}
+"{\"wss://relayer.somewhere.xyz\":{\"read\": \"true\",\"write\": \"true\"}}"
 ```
 
-_Invalid_
+_Invalid event.content_
 ```
-"content": {
-  "wss://relayer.somewhere.xyz/": {
-    "read": "true",
-    "write": "true"
-  }
-}
+"{\"wss://relayer.somewhere.xyz/\":{\"read\": \"true\",\"write\": \"true\"}}"
 ```
 
-_Invalid_
+_Invalid event.content_
 ```
-"content": {
-  "wss://relayer.somewhere.xyz": {
-    "read": "true",
-  }
-}
+"{\"wss://relayer.somewhere.xyz/\":{\"read\": \"true\"}}"
 ```
 
 ### Purposes
@@ -67,28 +53,12 @@ This NIP can be extended at a later date to include more advanced rules while re
 
 ### Schema
 
-_literal_
-```json
-{
-  "id": "<hash>",
-  "signature": "<signature>",
-  "created_at": "1673363256", 
-  "kind": 10001,
-  "tags": [],
-  "content": {
-    "wss://relayer.somewhere.xyz": {
-      "read": "true",
-      "write": "true"
-    }
-  }
-}
-```
-
 _stringified_ 
 ```json
 {
   "id": "<hash>",
   "signature": "<signature>",
+  "pubkey": "<pubkey>",
   "created_at": "1673363256", 
   "kind": 10001,
   "tags": [],
--- a/23.md
+++ b/23.md
@@ -54,7 +54,6 @@ This NIP can be extended at a later date to include more advanced rules while re
 
 ### Schema
 
-_stringified_ 
 ```json
 {
   "id": "<hash>",
--- a/23.md
+++ b/23.md
@@ -16,9 +16,9 @@ The key, being the relay URL, **must** omit the trailing slash.
 
 The `read` value operates on the values of the [NIP-01](01.md) **filter** object, while the `write` rule operates on the values of the **event** object. 
 
-When the `read` value equals the string `false` for a given **filter** the client SHOULD NOT send that filter in a `REQ` message to that relay.
+When the `read` value equals the string `true`  then read from that relay, otherwise don't
 
-When the `write` value equals the string `false` for a given **event** the client SHOULD NOT send that event in an `EVENT` message to that relay.
+When the `write` value equals the string `true`  then write to that relay, otherwise don't
 
 _Valid event.content_
 ```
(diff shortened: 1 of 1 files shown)


fiatjaf:
> When the read value equals false for a given filter the client SHOULD NOT send that filter in a REQ message to that relay.
> When the write value equals false for a given event the client SHOULD NOT send that event in an EVENT message to that relay.

Sir, you've removed my giant brain computation spec with properties of filters and events, therefore these lines make no sense anymore.

Without the fancy computation, I think this could be just `if the value equals the string "true" then send/read from that relay, otherwise don't` so that later we can reintroduce the fancy computation by changing the value from `"true"` to some other string that represents a fancy program.

fiatjaf:
Also the event format `{"content": {"relayurl": ...}}` is invalid, the `.content` must stringified always. I know you know this, but the way the NIP displays it is very confusing.

It is probably better to only include the inner JSON that will be stringified in the examples (i.e. just `{"relayurl": ...}`, without the parent `"content") and then in the end give the full event example with the stringified JSON inside `"content"`.

dskvr:
> Sir, you've removed my giant brain computation spec with properties of filters and events, therefore these lines make no sense anymore.

Hey! I actually modified those lines but maybe they required adjustments. Given 

```
"content": {
  "wss://relayer.somewhere.xyz": {
    "read": "true",
    "write": "true"
  }
}
```
`event.content.write` **equals** false
`event.content.read` **equals** false

Was worded differently with respect to the rules. 

I will modify to include `string` as that is definitely more explicit.

dskvr:
@fiatjaf Updated. Modified examples to be stringified only and clarified as per suggestions. 

Completely missed that the read/write evaluations no longer made sense, thanks for pointing it out!

staab:
I like this a lot more

dskvr:
I have a rudimentary and buggy af implementation up 

Preview link and usage here (there are edge cases in demo if you don't follow usage): https://github.com/dskvr/nostr-watch/pull/185

dskvr:
> I won't be using the "read" attributes though. As some of you are aware, I think clients should read from wherever the people they are following happen to post to...

In the case of a microblogging implementation on nostr, you're probably correct as in the end you'll require a social graph to get where you're trying to go. But not everything built on nostr will be a microblog or even require a social graph.

> not from a specific fixed set of relays they configure.

You might consider the possiblity that some groups/authors may have private archival nodes that may not be expensive, and they have explicitly set it to "write only" in order to reduce read requests to their relay. Just a thought.

> If neither are provided, what would that mean?

It wasn't explicitly specified, but implied via the `invalid` stringified content.

It could go one of two ways:
1. Strict, must be set.
2. Non-strict, default to false or true

I'm defaulting to strict because then the standard can be extended with fancy things like the runes by @fiatjaf and is easier to implement and maintain.

dskvr:
Also @fiatjaf the motivation behind a PR into #32 is that everything here is basically a compilation of the discussion, and ultimately you're the author. If you would prefer to disassociate from this NIP (maybe subtly implied in #162) or don't like these changes, I can change the base branch.

mikedilger:
> In the case of a microblogging implementation on nostr, you're probably correct as in the end you'll require a social graph to get where you're trying to go. But I find it highly likely that not everything built on nostr will be a microblog or even require a social graph. (I'm just trying to avoid using the "T" word)

I'm aware people use nostr in different ways. I don't think having read relays is wrong nor that they shouldn't be there. If you see nostr as a community you want to interact with, then you want read relays and a general feed. That model is widespread and I'm not saying it needs to go away.  It's just that I won't be using them because I don't like centralization. I want to use those hundred small relays that a few people post to and nobody reads from currently because they have such low populations. I want to follow those outliers, nomads, pariahs outside of the central circles of damus and wellorder. And to do that, I need to find out where they post.

I'm not sure I would call following pubkey P at relay R a "social graph". That makes it sound difficult to implement, like graph algorithms would be necessary or something.

I don't understand the "T" word you are thinking of, but my creative mind imagines you are calling me a troll.  I am not trolling, I am seriously interested in improving nostr and we have to work this issue out.

> You might consider the possiblity that some groups/authors may have private archival nodes that may not be expensive, and they have explicitly set it to "write only" in order to reduce read requests to their relay. Granted, this is a stretch. I'm sure there are better counter-arguments. Just a thought.

This is the first good counter argument I've heard. It clarifies the issue for me.

The list of relays someone wants to remember _for themselves_ to move from client to client is different from the list of relays that someone is advertising that they post to.

That just makes me want to write a new NIP to create such a second list explicitly. Or to modify this one so it can have a dual-use.

dskvr:
I think we got off on the wrong foot, the "T" word I was referring to was "Twitter." Additionally, I am not critical of your rationale in the slightest, in the context of a microblog (twitter-like client), your rationale makes perfect sense. If you limited yourself to a relay list you would encumber UX, good for nobody. We're on the same page there. 

> That just makes me want to write a new NIP to create such a second list explicitly. Or to modify this one so it can have a dual-use.

That's a good point, any ideas on how to move forward?

mikedilger:
> I think we got off on the wrong foot, the "T" word I was referring to was "Twitter." 

Understood. I'm too creative by half.

> > That just makes me want to write a new NIP to create such a second list explicitly. Or to modify this one so it can have a dual-use.
> 
> That's a good point, any ideas on how to move forward?

I thought maybe a third field "follow" also being "true" or "false".  Then if you don't want to be followed at a place you write to, set "follow":"false", and if it wasn't there I'd just presume it is okay to read from your write relays.

dskvr:
> The list of relays someone wants to remember for themselves to move from client to client is different from the list of relays that someone is advertising that they post to.

So upon re-reading this, I'm now having a hard time understanding how they are different. Unless I'm missing something...

Case A: (remember for themselves) 

Client reads read the value of 'read' and 'write' or none from the list. Though, by reading these values it could assist the client with finding relevant relays faster without traversal

Case B: (advertise that they post to) 

Client (of any kind) reads the value of 'write' from the list. 

I must be missing something. 

> Then if you don't want to be followed at a place you write to, set "follow":"false", and if it wasn't there I'd just presume it is okay to read from your write relays.

What is "follow" in this context?

mikedilger:
I'm trying to figure out how not to read from a relay that you write to when you don't want me to read from it, but you still want to remember for yourself that you are writing to it.  It's an edge case for sure.

dskvr:
> I'm trying to figure out how not to read from a relay that you write to when you don't want me to read from it, but you still want to remember for yourself that you are writing to it.

This sentence took me on ride, I'll mull it over.

mikedilger:
It was your edge case:

> You might consider the possiblity that some groups/authors may have private archival nodes that may not be expensive, and they have explicitly set it to "write only" in order to reduce read requests to their relay. Granted, this is a stretch. I'm sure there are better counter-arguments. Just a thought.

I'm not really concerned myself, relays can just refuse to serve my REQ request if they don't want it.  We should just climb out of this rathole because I think it's derailing the main thing, and my point that I'm not using the read fields is not material to this NIP.

dskvr:
@Cameri just informed me that Kind 2 is replaceable already?! Is this documented anywhere?

Cameri:
> @Cameri just informed me that Kind 2 is replaceable already?! Is this documented anywhere? 😞

I fixed my initial issue and corrected it to use `kind 3`. `kind 2` are not replaceable. Sorry for the confusion.

23.md
```diff
@@ -6,88 +6,65 @@ Relays List
...
+_Valid event.content_
+```
+"{\"wss://relayer.somewhere.xyz\":{\"read\": \"true\",\"write\": \"true\"}}"
+```
 
-(Public keys are shortened to 4 characters for readability.)
+_Invalid event.content: trailing slash_
+```
+"{\"wss://relayer.somewhere.xyz/\":{\"read\": \"true\",\"write\": \"true\"}}"
+```
```
Cameri, line 36:
```suggestion
```
I would drop this altogether. Let's not restrict which relay URLs are valid or not. If you can connect it is valid, if not, it isn't.

23.md
```diff
@@ -6,88 +6,65 @@ Relays List
...
 
-- Rule evaluation examples:
+_Invalid event.content: missing `write` member_
+```
+"{\"wss://relayer.somewhere.xyz\":{\"read\": \"true\"}}"
+```
 
-  - _read_
+### Purposes
```
Cameri, line 43:
```suggestion
### Use Cases
```

23.md
```diff
@@ -6,88 +6,65 @@ Relays List
...
-      - `e!`: `true`
-      - `e=5555`: `false`
-      - `kinds=1|kinds=4`: `true`
-      - `kinds<2`: `true`
-      - `kinds>7`: `false`
-      - `kinds=1|kinds=7&authors=8543|authors=1234`: `true`
+The first use case is meant to make it so users can open their client -- or different clients -- in different devices and have their list of relays automatically fetched from a default relay and start using their relays list without having to set everything up again.
 
-  - _write_
+For the second purpose, if any client decides to, they can show to the user what relays other users are using, suggest that or automatically add these to the user's relay list. The possibility of sharing a list of relays in standardized format is good for spreading information about relays and contributes to the censorship-resistance of the network.
```
Cameri, line 49:
```suggestion
Another use case is to display which relays other users are using, suggest connecting to them or automatically add these to the user's relay list. The possibility of sharing a list of relays in standardized format is good for spreading information about relays and contributes to the censorship-resistance of the network.
```

23.md
dskvr: The problem with that, is that then you will have duplicate relays in the list if one client adds trailing slashes, and another client does not.

Cameri: I think you should make your app tolerant to these by treating the trailing slash and non slash urls the same instead of forcing this validation. I expect clients to send even false relays that you can't ever connect to because the ip or host is not public.

dskvr: Yes, absolutely, regardless of the standard things will happen sometimes. But by not explicitly stating trailing slash requirements, it will happen a lot more, and there will be no means to resolve. With a canonical NIP, resolution is inherent. 

Given the scenario

```
{
  "wss://somerelay/" : { "read": "true", write: "false" }
  "wss://somerela": { "read": "false" }
}
```

Now not only does a dev need to dedupe, but they also now need to merge these two arrays, and decide which one to listen to.

Given a strict standard, the compliant client would just remove the invalid record, because they know due to a canonical standard which is "valid".

Cameri: We could normalize URLs using https://developer.mozilla.org/en-US/docs/Web/API/URL
```
var a = new URL('wss://relay')
var b = new URL('wss://relay/')

a.toString() === b.toString() // true
```

dskvr: @Cameri Let's wait for some others to weigh in.

dskvr: > We could normalize URLs using https://developer.mozilla.org/en-US/docs/Web/API/URL

How does this resolve the problem with **subjective validity?**

dskvr: Out of curiousity, what are the core concerns of the standard imposing trailing slashes? I can only think of one, which is that:

1. A server config may be configured with a backslash, and sanitizing the backslash could present issues (or vice versa if it were strict in the other direction) 

I'm sure you have a valid reason to be against it, but to me it is unclear.

Cameri: I think it is not necessary to mandate slashes (or not) but I don't have a strong desire to block this NIP because of it. Please disregard my comments and proceed with your requirement.

dskvr: Would be good to get some more input from others. I'm going to reopen and we can revisit tomorrow. I wouldn't get started on my implementation until next week anyways.

dskvr: @mikedilger @staab Any thoughts?

Cameri: Should we have the same restriction everywhere a relay is mentioned or just this particular NIP?

staab: I do not think we should validate relay urls, any string is "valid", it's the client's responsibility to deduplicate, validate, catch errors, etc.

dskvr: @staab To be clear it's not validation, it would be part of the standard to avoid collisions. Have you considered the scenario I presented above and the implications on the user experience?

staab: Nostr is just going to have all kinds of weird subjective validity problems due to its open nature, mandating a trailing slash is only whacking one of a million moles. I've had to be many times more defensive in my coding with Coracle than on any other project I've worked on. Permissiveness isn't easy, but the simplicity is worth the pain of implementation.

In your merging example, a client does indeed have to just pick a strategy, because there's no truly correct way to dedupe/merge. But I also think any method, including dropping both, picking first/last, merging forward, and merging backward are acceptable approaches because they approximate the intent, which is all any client can do in a distributed system. But I also think that for the most part it's just not going to be a problem because a) no one would do that on purpose, so a spec prohibiting it would be redundant with common sense, and b) users have access to their relay list and can manually de-dupe or correct things. Clients could also notify the user that something's wacky with their relays, which would solve the subjective validity issue, as users are the only authority on what relay settings should be.

dskvr: > Should we have the same restriction everywhere a relay is mentioned or just this particular NIP? 

@Cameri I'm not sure tbh, didn't think this one would be contentious, but it clearly is. At the end of the day my motivations are to get something through by consensus, even if I personally disagree with it. 

The rationale behind the relay format is holistic. Eliminate code bloat for devs and prevent poor UX with a clear out-route. 

I do see that it might not be necessary in the purest sense of the word, but it will probably result in a much cleaner implementation.

staab: IOW, you're asking clients to be smart enough to contradict a user when they enter a trailing slash, rather than be smart enough to notice problems and as the user for help.

dskvr: > But I also think any method, including dropping both, picking first/last
Object keys can change position

> merging forward
This is a replaceable kind, so if I understood you correctly, merging forward isn't really an option here (unless by merging forward you mean overwriting user data with your client data) 

>merging backward are acceptable approaches because they approximate the inten which is all any client can do in a distributed system. 
I absolutely agree, and clarifying in a standard which format is "valid" and defensive programming are not mutually exclusive. 

> as users are the only authority on what relay settings should be.
Yes, but 99% of them will enter the relay through a tool/client, which can sanitize in a one liner.

staab: By merging I was referring to https://github.com/nostr-protocol/nips/pull/164#discussion_r1068294779

With specs, I think they should be minimal or exhaustive. Mandating no trailing slash is neither, and nostr's ethos is one of minimalism.

dskvr: > By merging I was referring to https://github.com/nostr-protocol/nips/pull/164#discussion_r1068294779

There's no way to merge that non-destructively in a way that is logical. If there is guidance on a trailing slash, then dropping one of them in this scenario, is objectively justified. Maybe it starts a trailing vs no-trailing slash war, maybe the standard is amended to remove the guidance. 

> With specs, I think they should be minimal or exhaustive. Mandating no trailing slash is neither, and nostr's ethos is one of minimalism

Again, this isn't a mandate, it's technically a guidance. Virtually every other NIP has the language **must**, it doesn't detract from it's minimalism.

dskvr: > IOW, you're asking clients to be smart enough to contradict a user when they enter a trailing slash, rather than be smart enough to notice problems and as the user for help.

I'm actually asking the opposite, to handle the above scenario without guidance requires the client to be incredibly smart. To handle the scenario where the is guidance on the correct format is simply a rejection (1-2 lines of code). The client would also technically protecting the user from potential data loss

The only case that's been outlined, which I think @Cameri and you are touching on, is that enforcing trailing vs no trailing can result an edge case where a server is configured to only accept a trailing slash. 

And it is a completely valid case.

dskvr: Would be really good to get some more feedback from others. At this juncture, based on feedback it looks to me like this conversation is leaning towards removing relay URL formatting from the standard. Also, thanks for your input @staab!

dskvr:
@Cameri In advance, in no world would I want this merged into someone else's NIP without their blessing. In the event it's rejected or author is unresponsive, I will change the base branch and we can update the number.

23.md
Cameri: Your last sentence on your first reply to my initial suggestion was that you wouldn't die on this hill. Are you suuuuure?
🤪
On a serious note: are relays with obvious path like /ws/ also follow this rule? Keeping in mind that /ws and /ws/ are distinct paths. Is the restriction only when the path is just /?

Relays can't reject these events because they don't inspect the content field. This is a client-only rule.

Cameri:
> @Cameri In advance, in no world would I want this merged into someone else's NIP without their blessing. In the event it's rejected or author is unresponsive, I will change the base branch and we can update the number. 

I'm not suggesting that. That said, I agree with you on principle.

dskvr:
@Cameri Sorry, didn't mean to imply you did! Just wanted to make sure it was clear.

23.md
dskvr: Part of the process is arguing any points, hashing things out and clarifying any misconceptions. My last comment reads 

> At this juncture, based on feedback it is clear we are leaning towards removing relay URL formatting from the spec.

Unless new information comes up, I probably don't have anything left to say on the topic of a relay URL format

> On a serious note: are relays with obvious path like /ws/ also follow this rule? Keeping in mind that /ws and /ws/ are distinct paths. Is the restriction only when the path is just /?

If this rule somehow gets through, which at this point it does not appear it will, it would apply to any url. 

> Relays can't reject these events because they don't inspect the content field. This is a client-only rule.

Yes, this was expressed above.

dskvr: I was going to leave this unresolved until tomorrow, if there's no further feedback, I'll happily merge in your changes :D

dskvr:
@jb55 do you have any bandwidth for this?

fiatjaf:
Do we have any implementations? I would say we're living in chaos already so instead of debating more we should find someone with free time to implement this and see if it works.

staab:
I'm in the process of adding this to coracle, but I'm currently using the tags version and it's still a wip.

eskema:
I can try to implement this, but I don't think the url trailing slash should be enforced as we already need to make our own sanitization and logic. for example, I normalize my relay urls to have trailing slashes because in my case, since I use javascript, it's what the websocket object returns as ws.url, so to simplify my side I chose the other way around, but don't think it should be enforced because I know someone is not going to enforce it, so I need to prepare for it anyway...
regarding which version to implement, the content one is just a simpler switch to the current kind-3 implementation and is probably simpler to try, but I can simply go for the tags one. which one should I do?

eskema:
I think I'm going with the original one, makes sense to me. not sure what rules I'll do, probably only "" and "!" for starters.

dskvr:
@fiatjaf
> Do we have any implementations? I would say we're living in chaos already so instead of debating more we should find someone with free time to implement this and see if it works.

I have a rough demo, but am doing some work on the data layer first so that it can be implemented correctly. 

**That's two implementations of NIP-23a!** 

Music to my ears. I'll leave this open for a few days, but if there are implementations being build on the original spec, I'm also going to implement that one. 

Thanks for for the feedback @fiatjaf, @staab, @Cameri, @mikedilger, @eskema!

dskvr:
After implementing 3 cases:
1. NIP23a
2. NIP23b (without url formats)
3. NiP23c (with url formats) 

The only one that produced a clean implementation that will yield consistent results was the one that nobody seems to support. Therefor, I'm going to close this and simply not implement relay lists in my use case until either there is adequate demand or a sane NIP has been accepted. 

cheers

PR #165, opened 2023-01-10
NIP-47: BIP-47 payment code notification via Nostr DM [Work-in-progress]

kdmukai:
Squatting on this NIP number (for obvious reasons!) while I flesh out this concept.

For new or existing BIP-47 payment codes (and/or PayNyms), any user can derive the payment code's associated Nostr `npub1` pubkey. That opens up the possibility to DM the payer's notification to the recipient that is normally done as an onchain `OP_RETURN` in BIP-47.

The onchain "pollution" of using `OP_RETURN` is a legit criticism of BIP-47. This would eliminate that problem while still being completely deterministic (as long as the recipient still has the private key that's associated with their BIP-47 payment code, they'd always be able to regain access to their Nostr DMs sent to their payment code's `npub` pubkey).

### DIFF
--- a/47.md
+++ b/47.md
@@ -9,3 +9,71 @@ BIP-47 payment code notification via Nostr DM
 For new or existing BIP-47 payment codes (and/or PayNyms), any user can derive the payment code's associated Nostr `npub1` pubkey. That opens up the possibility to DM the payer's notification to the recipient that is normally done as an onchain `OP_RETURN` in BIP-47.
 
 The onchain "pollution" of using `OP_RETURN` is a legit criticism of BIP-47. This would eliminate that problem while still being completely deterministic (as long as the recipient still has the private key that's associated with their BIP-47 payment code, they'd always be able to regain access to their Nostr DMs sent to their payment code's `npub` pubkey).
+
+
+## Overview
+### Alice notifies Bob
+Per usual BIP-47 usage, Alice and Bob have generated their respective payment codes. Bob has published his payment code for anyone to use to send him BIP-47 payments.
+
+Alice wants to notify Bob that she will begin sending him BIP-47 payments.
+
+Alice derives the corresponding NIP-47 `npub` from Bob's BIP-47 public payment code.
+
+Alice can now DM her BIP-47 payment code to Bob's NIP-47 pubkey. This replaces the BIP-47 onchain notification tx. Alice does not need to blind her payment code since the content of her DM to Bob won't be publicly exposed.
+
+However, in order to preserve her privacy and not reveal that she is interacting with Bob's NIP-47 pubkey, Alice can use a "burner" Nostr private key to DM the one-time notification to Bob.
+
+### Bob checks his NIP-47 DM notifications
+Similar to how Alice can derive Bob's NIP-47 pubkey from his BIP-47 payment code, Bob can derive the private key associated with his NIP-47 pubkey.
+
+With the associated NIP-47 private key, he can see his NIP-47 notification DMs. For convenience, Bob's NIP-47 private key can [delegate (NIP-26)](26.md) DM access to Bob's normal Nostr key so he can access his NIP-47 notification DMs without having to juggle keys.
+
+
+## Example
+Test vectors from [BIP-47](https://github.com/bitcoin/bips/blob/master/bip-0047.mediawiki):
+```
+ALICE_MNEMONIC = "response seminar brave tip suit recall often sound stick owner lottery motion"
+ALICE_PAYMENT_CODE = "PM8TJTLJbPRGxSbc8EJi42Wrr6QbNSaSSVJ5Y3E4pbCYiTHUskHg13935Ubb7q8tx9GVbh2UuRnBc3WSyJHhUrw8KhprKnn9eDznYGieTzFcwQRya4GA"
+
+BOB_MNEMONIC = "reward upper indicate eight swift arch injury crystal super wrestle already dentist"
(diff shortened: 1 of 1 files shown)


louneskmt:
> One big advantage of NIP-47 is that as long as the recipient still has the private key that controls their BIP-47 wallet, they'd always be able to regain access to their NIP-47 notification DMs. Whereas if Alice were to directly DM her payment code to Bob, that DM could be orphaned/lost if he loses his Nostr private key or migrates away to a different identity.

It looks to me that this whole NIP is based on the assumption that DMs data cannot be lost. However, the storage of Nostr messages is directly dependable on relays goodwill. If Alice and Bob have a unique relay in common and that relay were to disappear or to delete the message, the notification DM would be lost, undependable of Bob losing his private key or migrating to a different account. Same thing for several relays. 

That means that both participants would need to keep a backup of the payment code of the other party, as the backup of the payment code private key wouldn't be sufficient. Whereas a notification transaction on Bitcoin will live forever. The purpose of the notification transaction, apart from transmitting the payer code, is to backup it. 

Also, you're never sure the notification DM will reach the recipient, as they might not be connected to the relays the sender broadcasted the message to. The relays aren't interconnected and relaying messages to each other, there is no such concept of "relay network" in Nostr.

rot13maxi:
To @louneskmt's comment, bip47 doesn't require interactivity or timely action from the recipient because the notification is durably stored on the blockchain. In the scheme you laid out above, the recipient must either ensure that they retrieve the notification from their relays in a timely manner (before it is pruned/purged from relays, either because of policy or just because the relay loses it) and then durably store the notification for future use, or they have to fetch the notification and then immediately sweep the funds to an address derived from their xpub. If you want to avoid compulsory transactions, I think you'd lean towards the first.

Another issue is that there's no guarantee that the notification DM gets to the recipient. The sender could publish to a different set of relays than what the recipient subscribes or the relays could fail to deliver the DM (either deliberately or just through operational issues). The sender wouldn't want to burn coins by sending them to an unspendable address, so they'd probably want some `ack` from the recipient that the recipient received the payment code. That leads us to an interactive protocol where the sender sends a payment code and the recipient acks it and then the sender sends the payment. 

If we're going to build an interactive protocol with one round of back-and-forth before payment, I suggest we simplify it down to:

```
Sender -> Recipient: "give me an address please"
Recipient -> Sender: "Here's an address"
sender sends to the address onchain
```

That will result in less data for the recipient to store (they don't have to back up received payment codes, only have to back up wallet descriptors or seeds), and more sender compatibility. Happy to help spec out a protocol for this :)

kdmukai:
> the recipient must either ensure that they retrieve the notification from their relays in a timely manner (before it is pruned/purged from relays, either because of policy or just because the relay loses it) and then durably store the notification for future use,

I definitely still have more to learn about Nostr to really understand the real-world possibilities and hurdles. Will the Nostr culture tend toward running your own limited archival relay? A NIP-47 bitcoin wallet would need a list of relays to connect to anyway, should be an option to point to your own and have it add your NIP-47 pubkey to its archive.

Is the Nostr culture in general trending toward "all events will ultimately fade away over time"? Or possibly pay-to-archive relays will be a thing? I'm sure there are no guarantees in any direction at the moment.

Separate from Nostr persistence (or lack thereof), I think wallets have an increasing need to ensure their label data (plus user accounts in Specter's case) is backed up on a regular basis. As an occasional wallet dev, it feels like that should be more my responsibility (to make backups robust and frequent and easy in the wallet software).

But I get the viewpoint that demands guaranteed recoverability. I don't think I share that hardliner view, but am unsure. Either way I definitely need to amend the text here to reflect the lack of guarantees.


> or they have to fetch the notification and then immediately sweep the funds to an address derived from their xpub.

In a low-fee environment this is easily done automatically in the wallet software since BIP-47 implementations are based on hot wallets. Sparrow has already set the excellent precedent of mixing straight to cold storage. This really wouldn't be any different, aside from the tx fee burden.

Wallets could also have a kind of "at risk" indicator for utxos. Hot utxos are obv at risk. Hot utxos that depend on potentially ephemeral NIP-47 payment codes are at even higher risk.


> Another issue is that there's no guarantee that the notification DM gets to the recipient. The sender could publish to a different set of relays than what the recipient subscribes or the relays could fail to deliver the DM (either deliberately or just through operational issues).

The bitcoin wallet could publish its relay metadata for its NIP-47 pubkey. Could the sender's wallet reliably look up the recipient's NIP-47 pubkey and retrieve that metadata? Could then send the notification DM directly to those relays.


> If we're going to build an interactive protocol with one round of back-and-forth before payment, I suggest we simplify it down to:
> 
> ```
> Sender -> Recipient: "give me an address please"
> Recipient -> Sender: "Here's an address"
> sender sends to the address onchain
> ```
> 
> That will result in less data for the recipient to store (they don't have to back up received payment codes, only have to back up wallet descriptors or seeds), and more sender compatibility. Happy to help spec out a protocol for this :)

There's definitely an option here to use Nostr to replace the Soroban/Tor layer of real-time direct interaction for things like negotiating a live PayJoin.

And there's some utility for automating what you describe for wallet-to-wallet one-off payments; basically btcpay server without the server--provided that the recipient's wallet is online.

You still have btcpay's challenges with addr DDOS attacks (just have to live with giving out tons of addrs that never get used and the resulting giant index gaps), but that doesn't stop btcpay from being an effective solution.

kdmukai:
BIP text updated. Phrasing about the BIP-47 onchain tx made more neutral.

More realistic tradeoffs along with some modest mitigations added.

Optional Nostr `nip47` metadata field proposed (which would actually justify this being a NIP and not just a messaging utility sitting on top of Nostr) that would point to the user's NIP-47 pubkey.

staab:
Closing due to inactivity

PR #171, opened 2023-01-15
Minor typos and a correction to NIP-09 reference to pubkey

vasil9v:
In NIP-09, my reading of the spec leads me to believe the kind-5 events publish a list of `e` tags each of which are to refer to a event `id` which should be considered to be deleted. This corrects language that refers to the event `pubkey` instead of the `id`.


fiatjaf:
Thank you very much!

PR #174, opened 2023-01-16
removed kind 6 since NIP-18 has been removed from the spec

RooSoft:
Kind 6 was still in the README.md, pointing to the removed NIP-18...


fiatjaf:
Oops, thank you!

PR #175, opened 2023-01-18
NIP-50: Keywords Filter

brugeman:
Would be great if clients allowed searching for nostr posts/people by keyword.  Currently it's only possible if client starts using a non-standard third-party API. Adding this capability at nostr level, even if not many relays choose to support it, allows to a) avoid adding yet-another-way to fetch events, b) make API a standard, so that clients could switch.

Try wss://relay.nostr.band with `["REQ","1",{"keywords":["bitcoin"], "kinds":[1], "limit": 10}]`

### DIFF
--- a/50.md
+++ b/50.md
@@ -1,0 +1,52 @@
+NIP-50
+======
+
+Keywords Filter
+---------------
+
+`draft` `optional` `author:brugeman`
+
+## Abstract
+
+Many Nostr use cases require some form of search by `content` field. The meaning and format of `content` differs 
+with event kind and use case. Thus, this NIP only establishes a general framework for search over `content`,
+and specifies the details for kind `0` and kind `1` events.
+
+## `keywords` filter field 
+
+A new `keywords` field is introduced for `REQ` messages from clients:
+```json
+{
+  "kinds": <an array of kind numbers>
+  "keywords": <an array of strings>
+}
+```
+A filter with `keywords` field MUST also include `kinds` field. This NIP specifies the behavior for kinds `0` and `1`, 
+if none of these kinds are provided along with `keywords`, relay SHOULD ignore the filter, and MAY send
+a `NOTICE` to inform the client. Future NIPs will specify behavior for other event kinds.
+
+A filter matches if one of the provided `keywords` matches. Each keyword string may contain several conditions - `words`, 
+all the words must match for a keyword to match. The meaning of `words` depends on event kind.
+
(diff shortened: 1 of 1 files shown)


ryzizub:
Agree with @joedavison, this is a great idea 🔥

gourcetools:
This is a nice idea!

mikedilger:
Add something like `Clients SHOULD use the supported_nips field to learn if a relay supports keyword filter queries. Clients MAY send keyword filter queries to any relay, if they are prepared to filter out extraneous responses from relays that do not support this NIP.`

gourcetools:
WOW! So cool!

RE-ACK


image
npm install ws
save this as node.js
```
const WebSocket = require('ws');
const ws = new WebSocket('wss://relay.nostr.band');

ws.on('open', function open() {
  ws.send(JSON.stringify(["REQ","1",{"keywords":["bitcoin"], "kinds":[1], "limit": 1}]));
});

ws.on('message', function incoming(data) {
  console.log(decodeBuffer(data));
});

function decodeBuffer(data) {
  return JSON.parse(Buffer.from(data).toString());
}

```

node node.js

brugeman:
> Add something like `Clients SHOULD use the supported_nips field to learn if a relay supports keyword filter queries. Clients MAY send keyword filter queries to any relay, if they are prepared to filter out extraneous responses from relays that do not support this NIP.`

I added the mention of supported_nips, thank you!

But I think I wouldn't add the second part - encouraging clients to query relays that don't support keyword search may overload those relays, and some relays may return an error or even drop the conn if they receive a request w/ unknown fields. It sure is a possibility to do client-side filtering, but then clients should probably send requests w/o keywords to relays not supporting this nip to at least not get their conn dropped.

mikedilger:
I just copied the clause from a previous NIP.

brugeman:
> I just copied the clause from a previous NIP.

I reconsidered my opinion on this after working on client side implementation, it makes sense to allow clients to fall-back to client-side filtering, otherwise 'search' feature won't really work for majority of users until popular relays support it. Added your recommendation.

fiatjaf:
More discussion at [note16pgzgz5lzfycm9rzxnquadthdtf9c5jnnyln98dxlwejn5d4646qmk0qsn](nostr:note16pgzgz5lzfycm9rzxnquadthdtf9c5jnnyln98dxlwejn5d4646qmk0qsn).

gourcetools:
Just found out that https://api.nostr.watch/v1/nip/50 will give you a list of relays supporting this nipple

PR #178, opened 2023-01-19
Adaptive Proof of Work

etemiz:
Spammers are having a great time. Adaptive proof of work can be a way to slow down if anything is speeding up.


fiatjaf:
Yes, but I would say just return the `pow: ` prefix instead, should be enough specially since we have no POW infrastructure in any client or relay and we don't know how that is going to evolve. Once POW gets more used these messages can get more granular.

PR #179, opened 2023-01-19
NIP-17 Git Updates and Discovery Over Nostr - first draft

melvincarvalho:


17.md
```diff
@@ -0,0 +1,52 @@
...
+
+Git events over nostr have the following attributes:
+
+**`content`** SHOULD be equal to the hex-encoded git commit hash.  Typically this will be a 40 character sha-1 hash, but git will soon be allowing sha-2 hashes too
+
+**`created_at`** SHOULD be equal to git *author date*.  This is the same format as nostr.
+
+**`tags`** MAY contain an entry identifying the previous commit, if one exists, in the form `["e", "<event_id>"]`.  Further tags can be developed to capture informative git information such as major/minor versioning.  TBD: it will be possible to hint at a recommended git URL to download the repos (e.g. from github), but it is not yet decided whether this would be a tag itself or an additional entry in an existing tag.
+
+**`pubkey`** SHOULD be associated with a given repo.  This could be communicated out of band, or, for example MAY be included in a file in the root directory of the repo, `nostr.json`
```
thesimplekid, line 41: I think it would be better for the pubkey to be that of the commit author, and the repo identifier to be a tag.  This would allow anyone to publish commits associated with a repo, however clients could filter based on author.

melvincarvalho: Thank you for the feedback.  I will try and implement a prototype and see if I can incorporate this pattern.  Then come back and refine the doc

kaichengmung: How about creating a repo like a NIP28 channel, by creating a kind-x0 event. Then you can update the repo metadata with kind x1 as well.

17.md
```diff
@@ -0,0 +1,111 @@
...
+Do not merge until implemented, work in progress
+------------------------------------------------
+
+Status: This NIP is an early draft and work in progress.  Implementations are being developed which will incorporate what is learnt.
+
+
+Introduction
+------------
+
+Git Commits over nostr enable **uses cases** similar to:
```
Giszmo, line 20:
```suggestion
Git Commits over nostr enable **use cases** similar to:
```

17.md
```diff
@@ -0,0 +1,111 @@
...
+
+- [github actions](https://github.com/features/actions)
+- [continuous integration](https://en.wikipedia.org/wiki/Continuous_integration)
+- [continuous deplyment](https://en.wikipedia.org/wiki/Continuous_deployment)
+- [travis](https://travis-ci.org/)
+
+but without central points of failure.  The aim is to create continuous distributed software development processes, without having to rely on complex or centralized services.
+
+**You need only know the nostr ID or bech32 form of a repo you will be able to discover all versions, its origins and its history.**
```
Giszmo, line 29: Wording. No shouting.

```suggestion
You only need to know the nostr ID or bech32 form of a repo and you will be able to discover all versions, its origins and its history.
```

17.md
```diff
@@ -0,0 +1,111 @@
...
+------------
+
+Git Commits over nostr enable **uses cases** similar to:
+
+- [github actions](https://github.com/features/actions)
+- [continuous integration](https://en.wikipedia.org/wiki/Continuous_integration)
+- [continuous deplyment](https://en.wikipedia.org/wiki/Continuous_deployment)
+- [travis](https://travis-ci.org/)
+
+but without central points of failure.  The aim is to create continuous distributed software development processes, without having to rely on complex or centralized services.
```
Giszmo, line 27: The nip only describes how commit messages could be replicated on nostr but doesn't go into any of these other aspects. Is it really worth thousands of events with all the tiniest commit messages to convey the existence of a project?

17.md
```diff
@@ -0,0 +1,111 @@
...
+New Tag
+=======
+
+A new tag "c" will be used to store commit hashes from git a new version of the repo 
+
+
+Bech32
+======
+
+The prefix `nrepo` can be used in front ends that support it, to allow easier copy and paste
```
Giszmo, line 49: An example would be great here. What comes after nrepo?

17.md
```diff
@@ -0,0 +1,111 @@
...
+Repo Discovery
+===============
+
+It will also be possible signal a recommended repo, just like it is possible to recommended a relay. In this way, a codebase could start on github, move to gitlab, then to an own hosted instance, and finally back to github.  All, without any breaks in service.  TODO: flesh out this section, perhaps include relay discovery, and bech32 encoding
+
+
+nostr.json
+==========
+
+The design of this NIP is such that each repo will have a public key used to send updates to a relay.  That key SHOULD appear in the top level directory of the repo in a file called `nostr.json`
```
Giszmo, line 91: So if I fork a project, I should change that nostr.json to use my pubkey ... but not really my pubkey as I would have to create a pubkey for the project? And then I would replicate all the events of the upstream project using my key?

Giszmo (commented):
I don't see the value in this nip yet but here are some more specific comments and suggestions.

17.md
melvincarvalho: This is a very good question.  Beyond the scope for now.  But it's entirely possibly, and I expect, that forked projects will have a new pubkey.  We will see!

17.md
melvincarvalho: Quite a few changes now in current version.  Scope is narrowed.

17.md
melvincarvalho: thanks

17.md
melvincarvalho: agree to differ

17.md
melvincarvalho: May expand on that, but I Think implementation and app will be more instructive.  Thanks for food for thought.

17.md
melvincarvalho: Is the same as npub really

melvincarvalho:
Update to this NIP, and narrowed scope, this will be called "Git Commits over Nostr".  I'm also working on a website which will have more extensive documentation, implementations and demos.  It will be much easier to understand when it's working.  Basically paste an pubkey or nrepo into a box, to see the repo evolution.

http://nip17.com/

This NIP is not going to do the whole of git, or github over nostr, just aims to do one small thing well, and demo the utility.  Still WIP stage.

akhavr:
Was there a discussion of higher level roadmap to integrate git and nostr?  I can't see big value in tracking git commit objects on nostr, would be glad proven wrong.

melvincarvalho:
Nostr doesnt have a roadmap, it's not that kind of project.  Just people building things in a permissionless way, without stepping on toes.  I'm going to redo this PR as the scope has changed.  When people see git, they think github.  What is needed is a system where (open source) software can itself interact with nostr.

PR #182, opened 2023-01-22
nip12-add 'l' language generic tag suggestion

monlovesmango:
add suggestion of using `l` generic tag for language.

formatted updates: https://github.com/monlovesmango/nips/blob/nip12-add-'l'-language-tag-suggestion/12.md


Giszmo:
Maybe the standardization of a language tag would deserve its own nip? While clients searching for only a certain language would currently yield virtually nothing, it could help to skip showing or to suggest translation if the language was known.

monlovesmango:
i dunno, doesn't seem complex enough for its own nip? and qualifies as generic tag query. hashtags didn't have a new nip.

i dont think we are doing any language standardization, just leveraging existing browser standards. this just to suggest what 'l' tag could be used for.

but happy to create net new nip if others agree it should be a new nip.

jb55:
does this need to be indexed?

dtonon:
> add suggestion of using l generic tag for language.
> I think it would be good for clients to be able to query by language.

+1 for the proposal by @monlovesmango

gkbrk:
> does this need to be indexed?

"Give me all tweets in German" seems like a useful thing to query. And the cardinality and size of this tag's values will be a lot smaller than other less useful generic tags. A large improvement for a tiny cost.

bu5hm4nn:
While the NIP originaly says "This NIP does not promote or standardize the use of any specific tag for any purpose." I would highly suggest that we do standardize the use of these tags (and reserving some for application specific use).

monlovesmango:
closing this PR in favor of [PR 670 - Update NIP32 with language example](https://github.com/nostr-protocol/nips/pull/670)

PR #183, opened 2023-01-23
NIP-51 Lists

monlovesmango:
recreating previous lists nip that I inadvertently closed.

previous conversation here: https://github.com/nostr-protocol/nips/pull/114

formatted: https://github.com/monlovesmango/nips/blob/lists/51.md


Giszmo (approved):
:heart:

51.md
```diff
@@ -0,0 +1,56 @@
+NIP-51
+======
+
+Lists
+-------------------------
+
+`draft` `optional` `author:fiatjaf` `author:arcbtc` `author:monlovesmango` `author:eskima`
```
eskema, line 7: `draft` `optional` `author:fiatjaf` `author:arcbtc` `author:monlovesmango` `author:eskema`

51.md
```diff
@@ -0,0 +1,56 @@
...
+-------------------------
+
+`draft` `optional` `author:fiatjaf` `author:arcbtc` `author:monlovesmango` `author:eskima` 
+
+A special event with kind `30000`, meaning "list" is defined as having a list of public and/or private tags. Public tags will be listed in the event `tags`. Private tags will be encrypted in the event `content` field. Encryption for private tags will use [NIP-04 - Encrypted Direct Message](04.md) encryption, using the list author's private and public key for the shared secret. This event kind would follow [NIP-33 - Parameterized Replaceable Events](33.md). Lists SHOULD NOT mix different tags (ie. 'p' tags can't be included in the same list as 'e' tags).
+
+The parameter for the list event will comprised of `<type>:<name>`:
+- `<type>` identifies the type of tags included in the list
+- `<name>` identifies the list by name
+
```
eskema, line 14: I'm not sure this part is needed. do we want to implicitly restrict list types here? I think that part should be handled in separate NIPs, and would allow to follow this model with lists that mix multiple kinds of tags by standardizing `d` tags.

Giszmo: In the same spirit we could allow other structured types in the encrypted body - and we should. This leads me to question the purpose of this nip. Maybe it should be more prescriptive, not less? Maybe it should be exclusively about `follow`, `mute` and `block` as that might be something the OP could readily implement and experiment with.

The event pinning and bookmarks might belong into the profile and more structured encrypted data respectively.

Giszmo (changes requested):
Sorry, I changed my mind. This nip is not ready for being merged. We probably should not pack everything containing lists of pubkeys or events in the same kind no matter what.

I would appreciate if this nip was changed to simply covering mute/follow.

51.md
monlovesmango: even if this NIP was only limited to follow, I would still want the type. I would want users to be able to create lists of hashtags, people, and events to follow. however putting those all into one event will be a mess for client implementation. there is no situation where hashtags and follows will be processed together, the client will always have to parse them out and handle them separately.

happy to drop pinned and bookmarks entirely. just threw them in bc seemed like people are itching for these.

51.md
```diff
@@ -0,0 +1,56 @@
...
+Commonly established list names for this kind are:
+- `mute` - a list of tags that the user wants to mute
+- `follow` - a list of tags that the user wants to follow
+- `pin` - a list of tags that the user wants to pin (for display on profile)
+- `bookmark` - a list of tags that the user wants to bookmark
+
```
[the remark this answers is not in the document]
monlovesmango: but how would the behavior change between the different lists?

eskema:
you always need to check each tag, you can't ever simply "trust" that it's any kind of tags in there, so specifying what each spcific word means is I think something that belongs in it's own NIP. this NIP would simply introduce the concept of using the tags and encrypted content as the same type of tags inside, but one is private, the other is public. be that exclusively p tags or exclusively e tags or t tags or any combination that anyone seems fit to their purpose, which would then be a NIP and we can expect the same behaviour

monlovesmango:
I don't think it makes sense to introduce a new concept in one NIP and then list the standard/taken names in another NIP. its confusing. especially since the behavior is not changing at all, its just about organizing the documentation.

how would it ever lead to `e:p:t:g:likes:subject:mute`? this can only happen if we don't specify what format it should be in. if I explicitly say there are only two parts, tag type and name, it leaves very little room for people to make up their own.

I think mute should cover event, pubkeys and hashtag muting. I think follow should cover event, pubkeys and hashtag follows. why create confusion and potentially more variations by defining them all individually?

monlovesmango:
>we can in fact choose that path of using p:follow, but what I'm saying is it doesn't need to be limiting in this NIP

yes but the use cases I want to enable by submitting this nip directly depend on this convention. this is central to the nip mechanics.

eskema:
> 

then you can use it. you don't need to force it on everyone else. that's why I'm saying the option is needed, otherwise I'm not going to be using this NIP, I'd have to use NIP52 or something

eskema:
it's just an unnecessary friction point. NIPs should be like functions, and some build upon others to form more complex stuff. the simple concept of public / private with tags / encrypted content is what's actually needed. the wording on the d tag is a specific use case that levarages the concept in this NIP to do something specific. we should not be limiting the concept to that particular use case alone.

monlovesmango:
I'm curious what you have against using this option?

I decided to use this format for the following reasons:
- to allow cross compatibility of lists between clients
- to allow segmentation of common list names (like `mute` or `follows`) by tag type

I think both of these are needed in a lists nip. do you disagree with either of these objectives? do you think there is a better way of accomplishing these? I'm happy to change the design

I personally don't think its efficient to introduce a lists nip that doesn't allow for identifying the tag type for private lists (without requesting and decrypting first).

eskema:
> I'm curious what you have against using this option?
> 
> I decided to use this format for the following reasons:
> 
> * to allow cross compatibility of lists between clients
> * to allow segmentation of common list names (like `mute` or `follows`) by tag type
> 
> I think both of these are needed in a lists nip. do you disagree with either of these objectives? do you think there is a better way of accomplishing these? I'm happy to change the design
> 
> I personally don't think its efficient to introduce a lists nip that doesn't allow for identifying the tag type for private lists (without requesting and decrypting first).

I have nothing against your logic, I think it might be a good thing. and IF we all go along, then it may end up working like that in the end, I don't think it belongs in the same nip, it discusses another specific thing.

eskema:
> I personally don't think its efficient to introduce a lists nip that doesn't allow for identifying the tag type for private lists (without requesting and decrypting first).

you would know what to expect by reading that specific NIP. So maybe what you want here is changing the title to be more specific of what you want to achieve with your method and introduce to novel concept of using this logic. and so we dn't refer to this NIP as the one for public / private lists but to the mute / follow / ignore NIP

mikedilger:
If these lists are "to yourself", why not just encrypt them with your public key and decrypt with your private key?  ECDH "to yourself" seems pointless to me.

monlovesmango:
thats what I originally had in mind but on nostr @mmalmi had suggested just using the existing nip04 encrypt/decrypt functions for browser signing. otherwise we would have to implement new functions. I agree that reusing nip04 is much simpler for implementation. I can be convinced either way on this one.

mikedilger:
I guess it doesn't matter. You are not exposing the key you are encrypting/decrypting with so you aren't leaking information about the keypair. And NIP-04 has an IV which does a lot.

Carry on.

jb55:
thanks for getting this going!

jb55:
> Sorry, I changed my mind. This nip is not ready for being merged. We probably should not pack everything containing lists of pubkeys or events in the same kind no matter what.
> 
> I would appreciate if this nip was changed to simply covering mute/follow.

I agree, there should be a different kind per list. this would allow clients to only pull the lists they need.

jb55:
ah nm I see you would just use nip33 for this

jb55:
I'm pretty happy with this general structure, but I don't think we need the `type:` thing. clients will probably have to know each `d` tag and know what to do with them, so perhaps those could be clarified in future nips if necessary. I would be happy with `mute` for the blocklist in damus, I don't see the point of having both `mute` and `block`. I would also be happy with `bookmarks` and `pinned`.

jb55:
damus currently implements mute lists using this + the "mute" nip33 param

51.md
```diff
@@ -0,0 +1,56 @@
...
+```json
+{
+  "kind": 30000,
+  "tags": [
+    ["d", "p:follow"],
+    ["p", "3bf0c63fcb93463407af97a5e5ee64fa883d107ef9e558472c4eb9aaaefa459d"],
+    ["p", "32e1827635450ebb3c5a7d12c1f8e7b2b514439ac10a67eef3d9fd9c5c68e245"],
+  ],
+  "content": "kHU7tKasd/HD5HXt7DtZcZvcK8Fa+BJMDduG5pfJkvh0jFktbtCG7jll6/b2diG2DRnVO+HJUhY0hfPXD/6TiSMLNh/IcWrqpdXacSIGEeI9VuYI0eoUw77VAhADWvpIPEeyN6qbm4QcEibM2hAVIYg90m1hpiG6/of+8sv5urD4W3QdhUWZqFim281THw3uWHf8rqFgzmHs4+nXX2G4xw==?iv=s74CzOnLHTonCkkjN7lcAg==",
```
jb55, line 51: can you provide a decrypted version of the content so people reading will know what to encrypt? it's not clear from this doc

monlovesmango: the decrypted version is actually listed above, so I added some verbiage to tell the reader that.

monlovesmango:
thanks all for the comments! let me rework this nip based on everyone's feedback.

arthurfranca:
I have some questions:

Regarding follow list, how can someone have many follow lists, one per topic? like if i wanted to read updates/feed of a group of pubkeys that usually post about `politics`, and then switch to a feed based on a group of pubkeys that posts about `pokemon`. Maybe enhance the p tag with a topic like this
`["p", "91cf9..4e5ca", "wss://alicerelay.com/", "alice", "pokemon"]`?

Also, how can person A use person B's public follow list, keeping it in sync instead of creating a copy that will get stale? Is it possible with this NIP?

fiatjaf:
Can we merge this today?

51.md
```diff
@@ -0,0 +1,56 @@
...
+]
+```
+
+then they would create a list event like below:
+
+```json
+{
+  "kind": 30000,
+  "tags": [
+    ["d", "follow"],
```
Giszmo, line 47: mute or follow?

51a.md
```diff
@@ -0,0 +1,18 @@
...
+
+`draft` `optional` `author:fiatjaf` `author:arcbtc` `author:monlovesmango` `author:eskema` 
+
+An event with kind `30000` is defined as a list event (see [NIP-51 - Lists](51.md)) for user public keys. The list of tags should follow the format of kind 3 events as defined in [NIP-02 - Contact List and Petnames](02.md).
+
+Commonly established list parameter names for this kind are:
+- `mute` - a list of tags that the user wants to mute
+- `follow` - a list of tags that the user wants to follow
+- `pin` - a list of tags that the user wants to pin (for display on profile)
+- `bookmark` - a list of tags that the user wants to bookmark
```
Giszmo, line 15: I'm confused. 51a is about "Public Keys" yet here you talk about tags? Isn't "tags" what we use for public list while for private list we use "content"?

```suggestion
- `mute` - a list of public keys that the user wants to mute
- `follow` - a list of public keys that the user wants to follow
- `pin` - a list of public keys that the user wants to pin (for display on profile)
- `bookmark` - a list of public keys that the user wants to bookmark
```

And what are bookmarked or pinned public keys for?

51a.md
```diff
@@ -0,0 +1,18 @@
...
+
+Outside of the commonly established list parameter names above, users are allowed to create lists with custom names.
```
Giszmo, line 17: This sentence is not needed. Of course they are allowed and they will do so at which point this nip will get amended to help interoperability.

```suggestion
```

51b.md
```diff
@@ -0,0 +1,18 @@
...
+
+An event with kind `30001` is defined as a list event (see [NIP-51 - Lists](51.md)) for note events (or kind 1 events). The list of tags should follow the format for "e" tags as defined in [NIP-10 - Conventions for clients' use of e and p tags in text events](10.md).
+
+Commonly established list parameter names for this kind are:
+- `mute` - a list of tags that the user wants to mute
+- `follow` - a list of tags that the user wants to follow
+- `pin` - a list of tags that the user wants to pin (for display on profile)
+- `bookmark` - a list of tags that the user wants to bookmark
+
+Outside of the commonly established list parameter names above, users are allowed to create lists with custom names.
```
Giszmo, line 17:
```suggestion
```

51b.md
```diff
@@ -0,0 +1,18 @@
...
+
+`draft` `optional` `author:fiatjaf` `author:arcbtc` `author:monlovesmango` `author:eskema` 
+
+An event with kind `30001` is defined as a list event (see [NIP-51 - Lists](51.md)) for note events (or kind 1 events). The list of tags should follow the format for "e" tags as defined in [NIP-10 - Conventions for clients' use of e and p tags in text events](10.md).
+
+Commonly established list parameter names for this kind are:
+- `mute` - a list of tags that the user wants to mute
+- `follow` - a list of tags that the user wants to follow
+- `pin` - a list of tags that the user wants to pin (for display on profile)
+- `bookmark` - a list of tags that the user wants to bookmark
```
Giszmo, line 15:
```suggestion
- `mute` - a list of Note Events that the user wants to mute
- `follow` - a list of Note Events that the user wants to follow
- `pin` - a list of Note Events that the user wants to pin (for display on profile)
- `bookmark` - a list of Note Events that the user wants to bookmark
```

51c.md
```diff
@@ -0,0 +1,16 @@
...
+`draft` `optional` `author:fiatjaf` `author:arcbtc` `author:monlovesmango` `author:eskema` 
+
+An event with kind `30002` is defined as a list event (see [NIP-51 - Lists](51.md)) for hashtags. The list of tags should follow the format for "t" tags as suggested in [NIP-12 - Generic Tag Queries](12.md#suggested-use-cases).
+
+Commonly established list parameter names for this kind are:
+- `mute` - a list of tags that the user wants to mute
+- `follow` - a list of tags that the user wants to follow
+- `bookmark` - a list of tags that the user wants to bookmark
+
+Outside of the commonly established list parameter names above, users are allowed to create lists with custom names.
```
Giszmo, line 16:
```suggestion
```

51c.md
```diff
@@ -0,0 +1,16 @@
...
+-------------------------
+
+`draft` `optional` `author:fiatjaf` `author:arcbtc` `author:monlovesmango` `author:eskema` 
+
+An event with kind `30002` is defined as a list event (see [NIP-51 - Lists](51.md)) for hashtags. The list of tags should follow the format for "t" tags as suggested in [NIP-12 - Generic Tag Queries](12.md#suggested-use-cases).
+
+Commonly established list parameter names for this kind are:
+- `mute` - a list of tags that the user wants to mute
+- `follow` - a list of tags that the user wants to follow
+- `bookmark` - a list of tags that the user wants to bookmark
```
Giszmo, line 14:
```suggestion
- `mute` - a list of Hashtags that the user wants to mute
- `follow` - a list of Hashtags that the user wants to follow
- `bookmark` - a list of Hashtags that the user wants to bookmark
```

Giszmo (changes requested):

fiatjaf:
Sorry for arriving late with this question, but why does this NIP specify both separate kinds and `d` tags? I think it could work with just one or the other, having both seems unnecessarily confusing. Not a deal-breaker, but weird, maybe there is a reason.

51.md
monlovesmango: good catch, thank you. fixed now.

51a.md
monlovesmango: all lists are comprised of tags. for 51a the list of tags should follow the format of kind 3 events as defined in [NIP-02 - Contact List and Petnames](https://github.com/monlovesmango/nips/blob/lists/02.md)

'pin' could be used to indicate "these are users I think are worth checking out"
'bookmark' could be used to save profiles you want to check out later

51a.md
monlovesmango: I think it is good to let people know it is ok to use names outside of those defined standardly. whats the harm?

51b.md
monlovesmango: again, it isn't a list of note events though. it a list of tags. the tag format is defined explicitly in the nip. 

"The list of tags should follow the format for "e" tags as defined in [NIP-10 - Conventions for clients' use of e and p tags in text events](https://github.com/monlovesmango/nips/blob/lists/10.md)."

51c.md
monlovesmango: again this is a list of tags. tag has been explicitly defined for 51c. 

"The list of tags should follow the format for "t" tags as suggested in [NIP-12 - Generic Tag Queries](https://github.com/monlovesmango/nips/blob/lists/12.md#suggested-use-cases)."

monlovesmango:
> Sorry for arriving late with this question, but why does this NIP specify both separate kinds and `d` tags? I think it could work with just one or the other, having both seems unnecessarily confusing. Not a deal-breaker, but weird, maybe there is a reason.

it was requested that this nip be a more generic standard for lists in general but also wanted to get started defining the frequently requested list types. now that it is in this format I actually really like that all the different kinds that are used for different list types can be referenced from one nip.

I do expect the number of list types to grow as the number of use cases grow. for example, I expect that there will be lists for long form content as well that list tags for parameterized replaceable events. also lists for relays, or lists of public lists that other users have created.

still open to changing back to a more standard format but personally like this structure more (for lists specifically, not generally).

fiatjaf:
What do you mean by a "more standard format"?

What I was saying is that we should either define one kind of each list (10010 for "mutes", 10011 for "favorites" etc) or one `d` tag for each of these, all in the same kind (30001, for example). Both ways would work.

The way it is now is slightly confusing, because I specify `d: mute`, but then I don't know what kind I must use.

fiatjaf:
Oh, about having a bunch of kinds and/or `d` tags specified in the same NIP, I think that is totally fine and good. I'm for keeping this NIP as the directory of all future lists that people may come up with, and update it as time passes. If that's what you meant.

verbiricha:
> Sorry for arriving late with this question, but why does this NIP specify both separate kinds and d tags? I think it could work with just one or the other, having both seems unnecessarily confusing. Not a deal-breaker, but weird, maybe there is a reason.

Each list kind has tags of different type, the goal is to have a different kind for lists of `e`, `p`, `t` and so on.

> still open to changing back to a more standard format but personally like this structure more (for lists specifically, not generally).

+1

51.md
```diff
@@ -0,0 +1,56 @@
...
+A "list" event is defined as having a list of public and/or private tags. Public tags will be listed in the event `tags`. Private tags will be encrypted in the event `content`. Encryption for private tags will use [NIP-04 - Encrypted Direct Message](04.md) encryption, using the list author's private and public key for the shared secret. A list event will follow [NIP-33 - Parameterized Replaceable Events](33.md) and the list name will be used as the parameter. A distinct event kind should be used for each new type of list created.
+
+## List Event Kinds
+
+| kind | description                 | NIP           |
+|------|-----------------------------|---------------|
+| 30000| User Public Key List        | [51a](51a.md) |
+| 30001| Note Event List             | [51b](51b.md) |
+| 30002| Hashtag list                | [51c](51c.md) |
+
```
Cameri, line 18:
```suggestion
| kind | description                 | NIP           |
|------|-----------------------------|---------------|
| 30000| User Public Key List        | [51a](51a.md) |
| 30001| Note Event List             | [51b](51b.md) |
| 30002| Hashtag list                | [51c](51c.md) |
| 30002| Badge list                   | [51d](51d.md) |
```

51.md
```diff
@@ -0,0 +1,56 @@
+NIP-51
+======
+
+Lists
+-------------------------
+
+`draft` `optional` `author:fiatjaf` `author:arcbtc` `author:monlovesmango` `author:eskema` `depnds:33`
```
Cameri, line 7:
```suggestion
`draft` `optional` `author:fiatjaf` `author:arcbtc` `author:monlovesmango` `author:eskema` `depends:33`
```

staab:
Late to the party, but I want to add my 2 sats. I don't like that this is separated out into multiple kinds, because I think it makes sense to have different types in the same list. Background is that I am implementing "custom views", which aren't as simple as the mute/pin/whatever use cases described above. If I have to create a list for following either pubkeys or topics by publishing two events and merging them together, that seems unnecessarily complicated.

Here's an example of what I'm hoping to do:

```
{
  "tags": [
     ["d", "All about bagels"],
     ["t", "bagels"],
     ["r", "wss://bagel.zone"],
     ["p", "3bf0c63fcb93463407af97a5e5ee64fa883d107ef9e558472c4eb9aaaefa459d"],
  ],
}
```

This nip doesn't support this use case, but I don't see why it shouldn't.

monlovesmango:
hey @staab actually this use case is exactly why I wanted explicit kinds for each "type" of list. all you need is to propose a new mixed use list type of 'custom views' that will be its own kind. then anyone who uses that kind is aware that it will have a mixed list of tag types. does that make sense? not sure I'm explaining it well.

this way people who only want to use profile lists don't need to worry about handling the multiple types of tags in your 'custom view' list, but you can easily still create 'custom view' lists for your users.

staab:
That said, if my criticism doesn't resonate with anyone else, ACK

Cameri:
> Late to the party, but I want to add my 2 sats. I don't like that this is separated out into multiple kinds, because I think it makes sense to have different types in the same list. Background is that I am implementing "custom views", which aren't as simple as the mute/pin/whatever use cases described above. If I have to create a list for following either pubkeys or topics by publishing two events and merging them together, that seems unnecessarily complicated.
> 
> Here's an example of what I'm hoping to do:
> 
> ```
> {
>   "tags": [
>      ["d", "All about bagels"],
>      ["t", "bagels"],
>      ["r", "wss://bagel.zone"],
>      ["p", "3bf0c63fcb93463407af97a5e5ee64fa883d107ef9e558472c4eb9aaaefa459d"],
>   ],
> }
> ```
> 
> This nip doesn't support this use case, but I don't see why it shouldn't.

I can see us having a kind for `Mixed` lists

fiatjaf:
> I can see us having a kind for Mixed lists

I think you have the wrong idea of what these kinds were supposed to mean.

fiatjaf:
@monlovesmango can we make these lists all have the same kind 30000 and be done with it?

Cameri:
> > I can see us having a kind for Mixed lists
> 
> I think you have the wrong idea of what these kinds were supposed to mean.

Naming things is hard. @staab wants a way to mix in multiple tags within the same list. We can call it `custom views` or `user-defined filters` or `mixed`.

> @monlovesmango can we make these lists all have the same kind 30000 and be done with it?

Using event kind 30000 for any kind of list also works IMO. We would only need to agree on the `d` tag value for interoperability. It's not always easy to see the issue with an approach until you need it to scale.

I'm PRO ditching using different event kinds for types of lists and sticking with 30000.

fiatjaf:
> Naming things is hard. @staab wants a way to mix in multiple tags within the same list. We can call it custom views or user-defined filters or mixed.

I think it is simpler for everybody if we can just fetch a "mute" list and that includes everything I have muted, including profiles and events, and other stuff that can show up later. Having to think if the contents of the list are events or people or addresses or anything else is just an extra mental and implementation burden. Also if a list is on the kind reserved for pubkeys but it contains an "e" tag? Should clients ignore that? I think they would not, it would just complicate the logic.

fiatjaf:
> I'm PRO ditching using different event kinds for types of lists and sticking with 30000.

My preferred approach is for each kind of list have a different kind number, but since no one agrees with me I'm ok with this.

> We would only need to agree on the d tag value for interoperability

This is the same requirement we would have for kind numbers, by the way, that's why I like kind numbers more.

staab:
I'm ok with using kinds to specify specific types of lists, but the default 30000 should be "whatever" or "user defined". Also, additional kinds should probably be specified by use case rather than tag type, e.g. "mute" not "all tags are pubkeys".

melvincarvalho:
> My preferred approach is for each kind of list have a different kind number, but since no one agrees with me I'm ok with this.

Chiming in a bit late, but in other systems I've used, Lists do indeed tend to have their own type/kind.  So, I agree.  But, just a minor observation, can live with any consensus.

monlovesmango:
>My preferred approach is for each kind of list have a different kind number, but since no one agrees with me I'm ok with this.

I think this is also a valid approach and happy to adapt this nip to this if others agree. If I am understanding correctly, this would mean 'mute' would be one kind, and 'bookmark' would be another kind, and each could contain any tag type right? If so it might make more sense to use replaceable events for these lists rather than parameterized replaceable events.

even if we went with this approach I would still want to have dedicated list kinds that only hold certain, predefined tag types.
this would enable letting users create custom named lists for various uses that applications could efficiently query and parse. for this, applications would need to be able to query for only lists that contain certain types of tags (depending on the use case) without knowing the 'd' parameter ahead of time (since the user can use a custom name). this is why I lean towards having list kinds be delineated by which tag types are included. @fiatjaf for these, yes I would expect clients to ignore non compliant tag types. what is everyone's thoughts on this use case?

right now this nip is delineating the list kind based on the tag type (in order to enable the custom list use case mentioned above). fiatjaf is suggesting delineating the list kind based on the list purpose ('mute', 'bookmark', 'pin'). I see 3 main options (the tables below are just examples of how list kinds could be broken out):

1. use both purpose and tag type:

| kind | description                 | NIP           |
|------|-----------------------------|---------------|
| 10000| Mute        | [51a](51a.md) |
| 10001| Bookmark       | [51b](51b.md) |
| 10002| Pin        | [51c](51c.md) |
| 30000| User Public Key List        | [51d](51d.md) |
| 30001| Note Event List             | [51e](51e.md) |
| 30002| Hashtag list                | [51f](51f.md) |
| 30008| Badge list                   | [51g](51g.md) |
2. use purpose only (doesn't leave flexibility for common custom lists): 

| kind | description                 | NIP           |
|------|-----------------------------|---------------|
| 10000| Mute        | [51a](51a.md) |
| 10001| Bookmark       | [51b](51b.md) |
| 10002| Pin        | [51c](51c.md) |
3. use tag type only ('mute', 'bookmark', 'pin' lists would be spread out across different list kinds based on tag type):

| kind | description                 | NIP           |
|------|-----------------------------|---------------|
| 30000| User Public Key List        | [51d](51d.md) |
| 30001| Note Event List             | [51e](51e.md) |
| 30002| Hashtag list                | [51f](51f.md) |
| 30008| Badge list                   | [51g](51g.md) |

anyone care to give an opinion on which option they think is best? or have alternative options to suggest? Personally I think I am leaning towards option #1.

do people agree that it make sense to use the 10000-19999 range for the purpose built lists?

staab:
Option #3 doesn't seem useful to me, it's not hard to parse tags and figure out what type they are. I'd like to hear your use case if you have one in mind. #1 seems infeasible from a speccing point of view, I don't think anyone wants to be merging new permutations of old use cases into this repo all the time. #2 all the way for me.

verbiricha:
> Personally I think I am leaning towards option 1

I think this approach addresses the main criticisms of this PR by having separate kinds for each mute/follow/bookmark list and allowing heterogenous tags in these. It still allows users to define named (and typed) lists using paramemeterized replaceable events which option 2 doesn't give us. I lean towards 1 too.

monlovesmango:
@staab my usecase for #3 would mainly be to create custom feeds. for pubkey lists, a user could create lists for 'podcasters', 'devs', 'travel', 'artists' etc based on which list feed they want to look through. same for hashtags, events, long form posts, and relay lists. 

for your concern with #1, can you expand on what you mean by 'new permutations of old use cases'? this nip is meant to be extended as new use cases arise, which is why I put it in the structure it is in now (where nip51 defines the list architecture and the sub nips define each kind of list). however once a list kind is defined (in a sub nip) it shouldn't need to updated regularly.

staab:
My permutations criticism I don't think is relevant, as we're not speccing the cross product of use cases and tag types. But I still don't see the use of homogenous lists absent of use case. Maybe we don't have the same working definition of "use case"? "User Public Key List" isn't a use case, "list of people I've muted" is. But then, "mutes" could include posts, topics, keywords, etc. as well.

monlovesmango:
use cases could be, but are not limited to:
- list of people who talk about bitcoin
- list of events about politics
- list of articles about philosophy
- list of hashtags about celebrities

the use case would be defined by the user, based on what type of event they are engaging with and whatever category they want to define. since the use case would be based on the type of event they are engaging with, different kinds are needed so that an hypothetical article app could easily query for all lists of articles a user has created.

I guess the use case would also be defined by the client as well, since the client implements the use of the lists. for pubkey lists specifically, a client could:
- make a standard feed
- create a list of trending events
- suggest new users to follow

perhaps this is still a somewhat wishy washy answer for use case. I just think this would be really useful for clients to enable their users to personalize their content.

staab:
Here's another way to ask the question, what assumptions do homogenous tags enable, and how does that relate to concrete use cases? IOW, if I find a list called "cats" and it's a mix of pubkeys, events, and topics, it doesn't seem any less straightforward to load everything up than if it was a list of pubkeys. Editing such a list might be more complex, but that would have to be supported anyway since heterogenous lists will exist.

People are probably starting to mute this thread at this point, so let's leave it at that unless anyone else wants to chime in.

monlovesmango:
an article app should be able to query for all the user's lists that contain tags of articles and only receive lists that have tags of articles. it shouldn't have to query, receive, decrypt, and parse all the user's lists to find which lists tag articles. I guess this would be what I want to enable with homogenous tags.

staab:
Again, that's use case specific, isn't it? So in that case you're not looking for lists that have all event tags, but specifically event tags of a particular _kind_. So even with this NIP as it currently stands you'd have to grab all homogenous event-tag lists, and then filter through to find the ones that are articles only.

fiatjaf:
Can we add use cases as they happen?

So far we have "mute" and "pin", which cannot be of multiple types, i.e. there is no "mute of type x" and "mute of type z". There is just "mute". We want interoperability between clients with regards to mute lists, so give those a 1000x kind number and everybody uses that. Same is valid for "pin". There is only one place to pin stuff, so that's another 1000x kind.

Then there are the more flexible "lists of things that I classify in some way", they can certainly be of many different types, so that's a 3000x kind.

That gives us:

| kind | description                 | NIP           |
|------|-----------------------------|---------------|
| 10001| Mute        | 51a |
| 10002|Pin        | 51b |
| 30001|Categorized Bookmarks | 51c |
| 30002|Categorized People | 51d |

kind 30001 and 30002 lists, for example, may have `d` tags like "football", "bitcoin", "music", each with set of pubkeys in them. Some clients may display 30002 as a custom feed, others may just display them as a list of the profiles, others may not even display them, but this is fine.

I can also see a 30003 list of URLs, for example, if someone wants to make a pinboard.in clone on Nostr -- that would comport the 3 kinds of lists above and perhaps others yet to be added to this NIP -- and would allow people to easily browse other people's lists and edit their own.

---

(I made up the numbers above without checking anything.)

fiatjaf (approved):

staab:
Oh, another thing is that 10002 is the relays list event from NIP65, so we might want to start counting from somewhere other than 10000.

staab:
Just implemented thread and person mutes https://github.com/staab/coracle/commit/c4c72abddc0e10d0311620d9044fa8cb856eb104

alexgleason:
How come Mutes (`10000`) and Pins (`10001`) are outside the NIP-33 range? That doesn't make sense to me.

Giszmo:
@alexgleason these are nip-16 replaceable events without "parametrized".

PR #186, opened 2023-01-24
add NIP-21, `nostr:` url scheme

fiatjaf:
https://github.com/nostr-protocol/nips/blob/urlscheme/21.md

### DIFF
--- a/21.md
+++ b/21.md
@@ -1,0 +1,20 @@
+NIP-21
+======
+
+`nostr:` URL scheme
+-------------------
+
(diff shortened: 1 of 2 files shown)


gkbrk:
Is there a way to allow browser extensions or web apps to handle this scheme without requiring all the links to be `nostr+web://`? Maybe it's not a problem with extensions, only web apps.

fiatjaf:
This: https://twitter.com/fiatjaf/status/1611360436162633729

fiatjaf:
Yes.

PR #191, opened 2023-01-25
NIP-53 - Nostr-based name registration system

leandro21b:
My name suggestion is NostrNameSystem (NNS)
My kind suggestion is 50


fiatjaf:
This is interesting, but I think it can be achieved in a more straightforward way on top of petnames instead: https://github.com/nostr-protocol/nips/blob/master/02.md#petname-scheme

Take a look, it is somewhat similar.

leandro21b:
interesting, but what about the idea of a decentralized DNS for domains can also be applied in NIP-02?

fiatjaf:
Well, it would be a relative DNS.

I give you a name, say `leandro`, then you can give names to other people or things, say you name your dog `cachorro`, then I can access it by visiting `cachorro.leandro`. If you dog decides to assign the name `osso` to his food then I can also access that through `osso.cachorro.leandro`.

Now if someone else calls you `nip53guy`, then they can access your dog's address at `cachorro.nip53guy`.

leandro21b:
It's an interesting concept

fiatjaf:
Now if you don't give a name to the person, apps using this scheme could just default to the name they give to themselves (as long as it is not `¨%*@$¨@$¨`) and if there are conflicts there the app can prompt the user to confirm -- or better, to assign a petname manually.

leandro21b:
Exactly, and this idea can also be implemented for domains, technically the owner of a website could define the domain and if users trusted it, they could save this domain related to the website's ip on their devices, as well as users could define a name-nickname to the website

leandro21b:
in fact, the ideal would be for people and specialized companies to organize domains and ip addresses, but this should not be centralized in a company, nostr seems very promising in this regard as well, since anyone could provide this service without having to have a server already that the data will be on multiple relays.

53.md
```diff
@@ -0,0 +1,40 @@
...
+The owner of a site himself can post a NNS event with a domain and the related IP and if people trust the owner of a site they will be able to accept this NNS event posted by him and so the client and/or the browser will automatically search for that address without the need for a third party to assign names to an IP or public key.
+
+## Example:
+
+```json
+{
+"id": ".i.d.",
+"pubkey": ".p.k.",
+"created_at": 12345,
+"kind": 50,
```
Giszmo, line 33: if it's replaceable, please pick a replaceable event kind according to nip-16 or nip-33.

53.md
```diff
@@ -0,0 +1,40 @@
...
+
+The owner of a site himself can post a NNS event with a domain and the related IP and if people trust the owner of a site they will be able to accept this NNS event posted by him and so the client and/or the browser will automatically search for that address without the need for a third party to assign names to an IP or public key.
+
+## Example:
+
+```json
+{
+"id": ".i.d.",
+"pubkey": ".p.k.",
+"created_at": 12345,
```
Giszmo, line 32: Please either put valid or no data in your example.

```suggestion
"id": <id>,
"pubkey": <pubkey>,
"created_at": <created_at>,
```

```suggestion
...,
```

53.md
```diff
@@ -0,0 +1,40 @@
...
+"kind": 50,
+"tags": [],
+"content": "{\"Domain.Example\":{\"6\":[\"[xxxx:xxxx:xxxx:xxxx:xxxx:xxxx:xxxx:xxxx]:port\"],\"4\":[\"111.111.111.111:port\",\"222.222.222.222:port\"],\"onion\":[\"xxxxxx.onion\"]},{\"@userexample\":\"npubxxxx\"}",
+"sig": ".s.g."
+}
+```
+
+Another suggestion is that in case a site has a dedicated nostr profile it can put this domain and IP data directly in the profile-medatada kind 0
```
Giszmo, line 40: that sounds like another nip? Maybe you wanted to open an issue to discuss your idea?

53.md
```diff
@@ -0,0 +1,40 @@
...
+
+Nostr-based name registration system
+------------------------------------
+
+`draft` `optional` `author:leandro21b`
+
+## Name:
+
+NostrNameSystem (NNS)
+Kind 50
```
Giszmo, line 12: Markdown eats line breaks. If you don't want these two lines on one, add a line break.

```suggestion
NostrNameSystem (NNS)

Kind 50
```

53.md
```diff
@@ -0,0 +1,40 @@
...
+
+Another alternative that is already used by some nostr clients is when you follow someone it saves the username automatically and thus displays it when you search for it without having to put the public key of that user every time, this concept can also be applied to domains so that you are not dependent on a centralized DNS server.
+
+Most Internet users do not have a fixed public IP, so the list referring to a website domain may need to be updated from time to time, for this purpose these NNS events will work similarly to kind 3, in case there is an update of an IP related to a domain, the user who posted the NNS event will be able to update the corresponding IP and the most recent event will always be seen by the relays as the correct one and will replace the previous one, as in kind 3.
+
+The owner of a site himself can post a NNS event with a domain and the related IP and if people trust the owner of a site they will be able to accept this NNS event posted by him and so the client and/or the browser will automatically search for that address without the need for a third party to assign names to an IP or public key.
```
Giszmo, line 24: This all sounds like extending nostr's [pet names](/02.md) to arbitrary data. I would appreciate if the Description  could be boiled down to much fewer lines.

Giszmo (changes requested):

53.md
leandro21b: More or less, my idea would be a decentralized DNS for urls, I put the possibility of this concept also being applied with usernames and their respective keys as a complement to this, is that one of the objectives with this idea of decentralized DNS based on nostr is to facilitate for whoever runs a relay has a url without depending on a centralized server, the same concept can also be applied to websites, if that is possible with this new NIP-53 I can give a summary and focus more on the idea of decentralized DNS.

53.md
leandro21b: Ok from what i analyzed i think it would fit best in kind 10000 (Replaceable Events)

53.md
leandro21b: This last one would be if a site or relay had a profile the ip data could be put in the profile metadata itself, but I'm not totally sure if that would be feasible so it's a suggestion

53.md
Giszmo: Replaceable events are unique by author and kind.  If each author only wants to maintain one name registration event for all registered names, kind 10000 to 19999 is for you. If you want one event per registered name, you need parametrized replaceable events kind 30000 to 39999 while using the `"d"`-tag.

leandro21b: For NIP-53 the best kind is 10000 (NIP-16 Replaceable Events)

leandro21b:
I updated NIP-53 and now as the idea of PETnames already exists in NIP-02 I decided to focus the idea on a Decentralized DNS system based on nostr.

leandro21b:
I created a new file for a new Pull Request but I don't really understand how it works, I'm still new to github, do I need to close this one? By the way, do you have any suggestions or criticisms regarding the `NIP-53`? Can it still be merged?

leandro21b:
I created a new NIP, I'm closing this one now

Giszmo:
Can you please link the follow-up PR for people interested to follow?

That said, I want petnames to flourish but that might be on the clients to actually use what's already defined.

leandro21b:
Yes I can reopen, as Github doesn't allow 2 PR from the same branch I will have to open another PR for the other `NIP` from another branch in my fork

In addition, in fact, until today, I have not seen any client adopting the `Petnames` part of `NIP-2`, at most I have seen them allowing a user to be found by username but that only works if you are following his profile on Nostr.

leandro21b:
It seems that Github still gives an error when trying to reopen the PR NIP-53 `The master branch was force-pushed or recreated.` maybe I have to reopen the NIP-53 in another PR, sorry for the confusion, I'm still learning how to use github

leandro21b:
Since Github wouldn't let me reopen this one, I created another NIP-53 https://github.com/nostr-protocol/nips/pull/386

PR #194, opened 2023-01-27
NIP-26: adding section documenting valid fields and operators for conditions string

zackwy:
This MR adds a section clarifying fields and operators supported when specifying condition strings for NIP-26, following the discussion [here](https://github.com/nostr-protocol/nips/pull/157).

We could potentially allow other conditions as well, however `kind` and `created_at` currently are the only I've specified here.

### DIFF
--- a/26.md
+++ b/26.md
@@ -35,12 +35,12 @@ nostr:delegation:<pubkey of publisher (delegatee)>:<conditions query string>
 
 The following fields and operators are supported in the above query string:
 
-###### Fields:
+*Fields*:
 1. `kind`
-   -  ###### Operators:
+   -  *Operators*:
       -  `=${KIND_NUMBER}` - delegator may only sign events of this kind
 2. `created_at`
-   -  ###### Operators:
+   -  *Operators*:
       -  `<${TIMESTAMP}` - delegator may only sign events created ***before*** the specified timestamp
       -  `>${TIMESTAMP}` - delegator may only sign events created ***after*** the specified timestamp
 
--- a/26.md
+++ b/26.md
@@ -50,7 +50,7 @@ For example, the following condition strings are valid:
 
 - `kind=1&created_at<1675721813`
 - `kind=0&kind=1&created_at>1675721813`
-- `kind=1&created_at>1675721813&created_at<1674777689`
+- `kind=1&created_at>1674777689&created_at<1675721813`
 
 #### Example
 
--- a/26.md
+++ b/26.md
@@ -31,6 +31,26 @@ The **delegation token** should be a 64-byte Schnorr signature of the sha256 has
 nostr:delegation:<pubkey of publisher (delegatee)>:<conditions query string>
 ```
 
+##### Supported Conditions
+
+The following fields and operators are supported in the above query string:
+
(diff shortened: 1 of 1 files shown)


zackwy:
Cc @markharding

Aspie96:
In the last row, did you intend to use the ">" operator twice, or should it have been a timestamp range?

zackwy:
> In the last row, did you intend to use the ">" operator twice, or should it have been a timestamp range?

Do you mean on line 53?

`kind=1&created_at>1675721813&created_at<1674777689`

Here I'm using both `>` and `<` to supply a range, however the duplication of the `created_at` field was intentional. Are you suggesting it would be better to have a single field?

I also noticed that the timestamps I used don't make sense here, I'll update them.

Aspie96:
My bad, I genuinely misread the line.

For what is worth, I support the updated specification.

bhayward93:
At a minimum, I think that it would be prudent to mention that setting a `before timestamp` with no `after timestamp` would allow clients to post historically on your behalf - just so that it's clear, because that poses a potential security risk for end users.

I would consider whether an adjustment to this spec may make sense also - if a `before timestamp` is provided, an `after timestamp` SHOULD, or even MUST be provided. It's hard to envision a project where someone needs to delegate signing responsibility to post historic notes, but perhaps there could be a use case somewhere that I'm missing?

I'm leaning towards thinking it should be something relays _should_ enforce, but wanted to hear your opinions on it.

Aspie96:
@bhayward93 I fail to see how that would add any security.

If it's the delegate party itself which generates the certificate (asking the user to sign it), it can simply set a timestamp which predates any date it could possibly want to sign events for, which would effectively be equivalent to not setting a minimum timestamp at all, so the minimum timestamp would be effectively optional even if it was a "MUST".

bhayward93:
> @bhayward93 I fail to see how that would add any security.
> 
> If it's the delegate party itself which generates the certificate (asking the user to sign it), it can simply set a timestamp which predates any date it could possibly want to sign events for, which would effectively be equivalent to not setting a minimum timestamp at all, so the minimum timestamp would be effectively optional even if it was a "MUST".

Yeah I gave it some thought and think it isn't something you can really enforce, more of a good practice for generation of delegation tokens. (See https://github.com/nostr-protocol/nips/pull/199)

Aspie96:
One idea is to advise clients (but it really shouldn't be a "MUST", and maybe not even a "SHOULD", just a "should" lowercase, without any specific technical meaning) to refuse signing certificates without a minimum created_at which also can't predate the current timestamp (with a little bit of tolerance because clocks aren't synchronized).

bhayward93:
> One idea is to advise clients (but it really shouldn't be a "MUST", and maybe not even a "SHOULD", just a "should" lowercase, without any specific technical meaning) to refuse signing certificates without a minimum created_at which also can't predate the current timestamp (with a little bit of tolerance because clocks aren't synchronized).

I see where you're coming from with this but there are edge-casey potential usecases for historic notes to be created by a delegatee on behalf of a delegator. For example, a service could offer the historic syncing of Twitter posts to Nostr.

PR #196, opened 2023-01-27
Bech32 encoded relay entities

Semisol:
the main purpose for this to be included as a `nostr:` URI.

some design choices:
- type `0` is reserved for relay since a relay type already exists (`1`)
- only one relay can be included in the encoded entity so that if metadata is added in the future there is no confusion on which relay it applies to.


Cameri (approved):

fiatjaf:
I don't know, it feels to me that using type `0` for the relay URL is better in this case.

Do you have a use case for this? Why not pass just relay URLs directly?

barkyq:
`ndelegationtoken`?

PR #199, opened 2023-01-27
NIP-26: Advice on using after operators in conditions query string

bhayward93:
The primary purpose of this PR is to advise delegators to precisely scope the time windows in their delegation token query strings, such that delegatees are not permitted to sign historic posts.

Additionally:

- Fixed a minor typo in the capitalization of Schnorr
- Extended the existing example to use the advice provided about scoping time windows.
- Fixed the example line (74) to indicate that you should be signing a ***SHA256 hash*** of the delegation token string.

### DIFF
--- a/26.md
+++ b/26.md
@@ -31,6 +31,26 @@ The **delegation token** should be a 64-byte Schnorr signature of the sha256 has
 nostr:delegation:<pubkey of publisher (delegatee)>:<conditions query string>
 ```
 
+##### Conditions Query String
+
+The following fields and operators are supported in the above query string:
+
+*Fields*:
+1. `kind`
+   -  *Operators*:
+      -  `=${KIND_NUMBER}` - delegator may only sign events of this kind
+2. `created_at`
+   -  *Operators*:
+      -  `<${TIMESTAMP}` - delegator may only sign events created ***before*** the specified timestamp
+      -  `>${TIMESTAMP}` - delegator may only sign events created ***after*** the specified timestamp
+
+In order to create a single condition, you must use a supported field and operator. Multiple conditions can be used in a single query string, including on the same field. Conditions must be combined with `&`.
+
+For example, the following condition strings are valid:
+
+- `kind=1&created_at<1675721813`
+- `kind=0&kind=1&created_at>1675721813`
+- `kind=1&created_at>1674777689&created_at<1675721813`
 
 #### Example
 
--- a/26.md
+++ b/26.md
@@ -52,7 +52,7 @@ For example, the following condition strings are valid:
 - `kind=0&kind=1&created_at>1675721813`
 - `kind=1&created_at>1674777689&created_at<1675721813`
 
-For the vast majority of use-cases, it is advisable that query strings should include a `created_at` ***after*** condition reflecting the current time to prevent delegatee's from publishing historic notes on delegator's behalves.
+For the vast majority of use-cases, it is advisable that query strings should include a `created_at` ***after*** condition reflecting the current time, to prevent delegatee's from publishing historic notes on delegator's behalves.
 
 #### Example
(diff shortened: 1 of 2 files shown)


fiatjaf:
Looks good to me.

bhayward93 (commented):
Typo on Ln 99 to adjust

26.md
```diff
@@ -52,6 +52,8 @@ For example, the following condition strings are valid:
 - `kind=0&kind=1&created_at>1675721813`
 - `kind=1&created_at>1674777689&created_at<1675721813`
 
+It is advisable when constructing a delegation token query string with a `created_at` ***before*** operator, that a `created_at` ***after*** operator is also provided to precisely scope the time window that the delegation token is valid for. This prevents the delegatee from publishing historic notes on the delegator's behalf.
```
Aspie96, line 55: I think that if adding a `created_at` ***after*** operator is advisable, then it is so regardless of whether a ***before*** operator is present or not.

They achieve different purposes, and the presence of the ***before*** operator has no bearing on whether the ***after*** operator is advisable.

bhayward93: This is true, thanks for your feedback - I will update it when I have a moment.

I may change the wording a little also, such that it indicates that for _the vast majority of use-cases_, it is advisable, rather than implying for all usecases.

bhayward93: Updated! I'm going to resolve this conversation but feel free to unresolve it or create a new one if there is anything to add.

26.md
```diff
@@ -52,6 +52,8 @@ For example, the following condition strings are valid:
 - `kind=0&kind=1&created_at>1675721813`
 - `kind=1&created_at>1674777689&created_at<1675721813`
 
+For the vast majority of use-cases, it is advisable that query strings should include a `created_at` ***after*** condition reflecting the current time, to prevent delegatee's from publishing historic notes on delegator's behalves.
```
Aspie96, line 55: Why "delegatee's"? Is this a typo, or am I parsing it wrong?

```suggestion
For the vast majority of use-cases, it is advisable that query strings should include a `created_at` ***after*** condition reflecting the current time, to prevent the delegatee from publishing historic notes on delegator's behalves.
```

bhayward93: It's not a typo, it's advice for the plural 'query strings' (of different tokens), which would have different delegators and delegatees. Your suggestion is less ambiguous so I'm happy to go with that, however I think that there has to be a change in the last three words also:

> For the vast majority of use-cases, it is advisable that query strings should include a `created_at` ***after*** condition reflecting the current time, to prevent the delegatee from publishing historic notes on the delegator's behalf.

Aspie96: I still struggle to parse "to prevent delegatee's from publishing"… Why the apostrophe? "delegatees" (plural) makes sense here. It's delegatees that are prevented from publishing historic notes. I don't get the apostrophe in this sentence.

bhayward93: Ah, understood, that is indeed a typo then - the reworded suggestion threw me off. Would you like to suggest your now preferred change and I can add it?

Aspie96: Sorry for the confusion, I wasn't sure whether the error was in the pluralization or something else.

Either version is ok for me, use your own personal preference.

Aspie96: I can't resolve the conversation because I didn't open the PR and I don't have write access. Feel free to mark as resolved, however.

bhayward93: Ah, I've been using Gitlab for the last few years so I'm no longer as familiar with Githubs PR process. Thank you very much, I'll resolve now.

bhayward93:
@fiatjaf I think this is ready - some small changes to language have been added since your last review, and also I made the recommendation to provide an after condition regardless of whether a before one is present.

PR #200, opened 2023-01-27
Fix NIP-50 typo

lpicanco:
Fixing a typo on NIP-50.

cc @brugeman


Egge7:
Filters should be an array. Does not look like a typo to me.

Egge7:
Aren't filters supposed to be an array?

fiatjaf:
> Aren't filters supposed to be an array?

No. This fix is correct.

PR #201, opened 2023-01-28
nip-39 digital identites in metadata

pseudozach:
@Semisol please check and let me know if this looks good or if you prefer to be removed from author list.
background discussion link: https://t.me/nostr_protocol/77441

Currently this NIP is supported on 1 client nostr.directory for github identity claims.

Seeking comments and review.

### DIFF
--- a/39.md
+++ b/39.md
@@ -1,0 +1,36 @@
+NIP-39
+======
+
+Digital Identities in Profile
+-----------------
+
+`draft` `optional` `author:pseudozach` `author:semisol`
+
+## Abstract
+
+Nostr protocol users may have other online identities such as usernames, profile pages, keypairs etc. they control and they may want to include this data in their profile metadata so clients can parse, validate and display this information.
+
+## `identities` field in metadata content
+
+A new optional `identities` field is introduced for `kind 0` metadata event contents in addition to name, about, picture fields as included in [NIP-01](https://github.com/nostr-protocol/nips/blob/master/01.md):
+```json
+{
+    "id": "123",
+    "pubkey": "xxx",
+    ...
+    "content": {
+        "name": "asd",
+        ...
+        "identites": [
+            {"type": "twitter", "claim": "Semisol_Public", "proof": "https://twitter.com/Semisol_Public/status/xxx"},
+            {"type": "github", "claim": "semisol", "proof": "https://github.com/Semisol"},
+        ]
+    }
+}
+```
+`identities` field is an array of objects that include:
+* `type` may be the name of the service, website, a protocol or other identity system like `bitcoin` or `paynym` depending on what's being claimed.
+* `claim` should be the name of the digital estate like the profile name or bitcoin address.
+* `proof` should point to a URL that includes a way to prove that the digital identity being claimed is indeed controlled by the pubkey that signed this event.
+
+Clients should validate the information locally when possible or crawlers can remotely check the claim being made before displaying the claims as verified to other users e.g. the linked twitter proof indeed includes a statement such as "I control pubkey xxx" where xxx is the pubkey that published this kind 0 event.


fiatjaf:
I do think these fit better in tags.

gkbrk:
Will this NIP provide a namespace for the identity types? For websites those can be fully-qualified domains perhaps.

There should be a way to go from "type": "github" to "github.com", or from "bitcoin" to the commonly used "bitcoin" as opposed to "bitcoin cash" or another fork. Who will decide which network maps to the "bitcoin" type for NIP-39? Whoever owns the .com domain, the github account, something else? Does an account at "twitter.net" or "twitter.eth" qualify to be an identity type of "twitter"?

pseudozach:
> I've said this on Telegram, but let me say again that I think these fit better in tags.

is the benefit of tags just that it makes them easily queryable from relays?

gkbrk:
> is the benefit of tags just that it makes them easily queryable from relays?

Tags are part of the event, also signed and authenticated, and they are already parsed without having to parse JSON from the content string. Even without being queryable, it is more structured than a string that needs to be parsed as JSON first.

gkbrk:
> https://gist.github.com/Semisol/3bfe26b41d670abfb958e4b841de1413

This is much better IMO, especially the limited scope of claim types and clearly showing exactly where to look for the proofs.

fiatjaf:
I like it, we can query by alternative identity? Very good. Relays may even want to verify these things by themselves before accepting an event if they are strict like @atdixon's https://github.com/atdixon/me.untethr.nostr-relay.

39.md
```diff
@@ -0,0 +1,36 @@
...
+A new optional `identities` field is introduced for `kind 0` metadata event contents in addition to name, about, picture fields as included in [NIP-01](https://github.com/nostr-protocol/nips/blob/master/01.md):
+```json
+{
+    "id": "123",
+    "pubkey": "xxx",
+    ...
+    "content": {
+        "name": "asd",
+        ...
+        "identites": [
```
eskema, line 24: "identities": [

eskema:
looks good

39.md
```diff
@@ -0,0 +1,59 @@
...
+
+### `github`
+
+Identity: A GitHub username.  
+Proof: The ID of the GitHub Gist located at `https://gist.github.com/<identity>/<proof>` with a single file with the text `Verifying that I control the following Nostr public key: <npub encoded public key>`
+
+### `twitter`
+
+Identity: A Twitter username.  
+Proof: The ID of the Twitter tweet located at `https://twitter.com/<identity>/status/<proof>` with the text `Verifying my account on nostr My Public Key: "<npub encoded public key>"`
```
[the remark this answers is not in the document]
pseudozach: I specifically changed the text to the existing template that's used on nostr.directory because there are 15k+ such tweets so that can be re-used. I'll update the explanation but strongly suggest we keep the existing tweet text to allow users with existing tweets to migrate to this schema.

39.md
```diff
@@ -41,19 +41,23 @@ If a matching metadata event is found when trying to look up an identity, a clie
...
-Proof: A Mastodon post ID. This post should be published by `<username>@<instance>` and have the text `Verifying that I control the following Nostr public key: "<npub encoded public key>"`.
+Proof: A Mastodon post ID. This post should be published by `<username>@<instance>` and have the text `Verifying that I control the following Nostr public key: "<npub encoded public key>"`.  
+This can be located at `https://<identity>/<proof>`.
 
 ### `telegram`
 
 Identity: A Telegram user ID.
-Proof: A string in the format `<ref>/<id>` which points to a message published in the public channel or group with name `<ref>` and message ID `<id>`. This message should be sent by user ID `<identity>` and have the text `Verifying My Public Key: "<npub encoded public key>"`.
\ No newline at end of file
+Proof: A string in the format `<ref>/<id>` which points to a message published in the public channel or group with name `<ref>` and message ID `<id>`. This message should be sent by user ID `<identity>` and have the text `Verifying My Public Key: "<npub encoded public key>"`.
```
eskema, line 62: do we really need the string `Verifying that I control the following Nostr public key:` everytime? isn't nostr already implied by just using `<npub encoded public key>`? 
if we do need it, then it would be better if it's always the same and here it's different in this telegram block.

pseudozach: it would be confusing for people on other social sites to see just a pubkey with no context. Having a grammatically correct statement is a good idea in my view. it's sort of a written contract/claim.

39.md
```diff
@@ -0,0 +1,63 @@
...
+Nostr protocol users may have other online identities such as usernames, profile pages, keypairs etc. they control and they may want to include this data in their profile metadata so clients can parse, validate and display this information.
+
+## `i` tag on a metadata event
+
+A new optional `i` tag is introduced for `kind 0` metadata event contents in addition to name, about, picture fields as included in [NIP-01](https://github.com/nostr-protocol/nips/blob/master/01.md):
+```json
+{
+    "id": "123",
+    "pubkey": "xxx",
+    ...
```
Giszmo, line 20: There are three options for handling stuff not touched on in the nip: omit it, show placeholders or valid data.

```suggestion
    ...
```

```suggestion
    "id": <id>,
    "pubkey": <pubKey>,
    ...
```

39.md
```diff
@@ -0,0 +1,63 @@
...
+```
+
+An `i` tag will have two parameters, which are defined as the following:
+1. `identity`: This is the identity provider name (for example `github`) and the identity  (for example `semisol`) joined together.
+2. `proof`: String or object that points to the proof of owning this identity.
+
+Clients SHOULD process any `i` tags with more than 2 values for future extensibility.  
+Identity provider names SHOULD only include `a-z`, `0-9` and the characters `._-/` and MUST NOT include `:`.  
+Identity names SHOULD be normalized by replacing uppercase letters with lowercase letters and if there are multiple aliases for an entity using the primary one.  
+If a matching metadata event is found when trying to look up an identity, a client SHOULD check if it's the latest version, and if not if it has this claim. This is to prevent showing or checking public keys that disowned a claim.
```
Giszmo, line 37: Obsolete line as replaceable events like this are already defined as "the last one counts".

```suggestion
```

39.md
Giszmo: Then still it makes sense to use the same sentence. I assume you accidentally had variations there.

39.md
```diff
@@ -0,0 +1,63 @@
...
+        ["i", "github:semisol", "9721ce4ee4fceb91c9711ca2a6c9a5ab"],
+        ["i", "twitter:semisol_public", "1619358434134196225"],
+        ["i", "mastodon:bitcoinhackers.org/@semisol", "109775066355589974"]
+        ["i", "telegram:1087295469", "nostrdirectory/770"]
+    ]
+}
+```
+
+An `i` tag will have two parameters, which are defined as the following:
+1. `identity`: This is the identity provider name (for example `github`) and the identity  (for example `semisol`) joined together.
```
Giszmo, line 31: You later refer to `ìdentity` but mean `semisol` not `github:semisol` and `identity` is specified twice in this line.

```suggestion
1. `platform:identity`: This is the platform name (for example `github`) and the identity on that platform (for example `semisol`) joined together with `:`.
```

Giszmo (changes requested):

39.md
pseudozach: @Giszmo no, tweet and telegram text is different on purpose because of 16k+ tweets+tg posts that already have the text from nostr.directory. I think we should keep it to avoid having people do the same thing again. Or if we must keep the text the same then maybe we can put a note so clients can also accept variations on the wording of the statement.

39.md
pseudozach: @Semisol can you chime in please? ok to remove this line?

39.md
pseudozach: I agree. @Semisol any objections to this?

majestrate:
this feels a lot like the w3c's `rel=me` used in activity pub. 
i think there may be a way to also have an interop with that here too. it would be a way to accelerate the network effect in adoption of this feature in the future.

aleixisp:
Anywhere to see an example of the format for the proof? 

What are the steps to add other platform besides GitHub, Mastodon, Twitter and Telegram?

majestrate:
> Anywhere to see an example of the format for the proof?
> 
> What are the steps to add other platform besides GitHub, Mastodon, Twitter and Telegram?

https://microformats.org/wiki/rel-me

my understanding is that it's done in the html document itself:

```
<a href="https://github/whatever" rel="me">github</a>
<a href="scheme://host/whatever" rel="me">other</a>
```

additionally `rel=me` seems to be more flexible wrt being able to reference external resources despite this format existing only in html documents.

majestrate:
as for proofs, `rel=me` is not "valid" unless all your references full mesh with each other.

39.md
wires: just and idea, but for github and similar sites a SSH pubkey could be used to more directly proof ownership. You could use openssl or smth to sign a message with the SSH identity that is visible on github (the pubkey can be found at a specific url. Similar for sr.ht etc)

This would then use the encoded signature as the third element of the tag array. Pubkeys are usually Ed25519 or RSA signatures, so clients would have to implement cryptography instead of using https+gist to verify validity.

This could be 'github+ssh:account', say.

wires: Just came back to say this comment above ^ is nonsensical; you would still need to get the file containing the pubkey and the. *additionally* do crypto. I'll leave this note here in case anyone else has this same brilliant idea haha.

gist + well known string = KISS

fiatjaf (approved):

Cameri (approved):

jawsh207:
"Mastodon" should be changed to "Fediverse" and the icon replaced with the proposed icon here: https://commons.m.wikimedia.org/wiki/File:Fediverse_logo_proposal.svg
Calling it the Mastodon network is inaccurate and seems to happen far to often. I expect to hear an arguement along the lines of "nobody knows what the Fediverse is but they know what Mastodon is". I dislike this arguement as it keeps people ignorant vs educating that the Fediverse consists of far more than just Mastodon instances. I'm not sure if this is the place to suggest this change, apologies if not.

jawsh207:
> > "Mastodon" should be changed to "Fediverse" and the icon replaced with the proposed icon here: [commons.m.wikimedia.org/wiki/File:Fediverse_logo_proposal.svg](https://commons.m.wikimedia.org/wiki/File:Fediverse_logo_proposal.svg) Calling it the Mastodon network is inaccurate and seems to happen far to often. I expect to hear an arguement along the lines of "nobody knows what the Fediverse is but they know what Mastodon is". I dislike this arguement as it keeps people ignorant vs educating that the Fediverse consists of far more than just Mastodon instances. I'm not sure if this is the place to suggest this change, apologies if not. EDIT: maybe ignore the part about the icon change as that may be something that needs to addressed in Amethyst if it is changed and corrected.
> 
> Please make a PR if you could

Attempted to but not sure what I'm doing wrong.

PR #205, opened 2023-01-28
NIP-56: Reporting

jb55:
This NIP provides a way for clients to report profiles and notes to relays.

This is currently implemented by Damus as an apple app store requirement, but may be useful in other moderation contexts.

https://github.com/nostr-protocol/nips/blob/reporting/56.md

### DIFF
--- a/56.md
+++ b/56.md
@@ -21,9 +21,11 @@ are reporting.
 
 If reporting a note, an `e` tag MUST also be included referencing the note id.
 
-A `report` tag MUST be included which consists of the following report types:
+A `report type` string MUST be included as the 3rd entry to the `e` or `p` tag
+being reported, which consists of the following report types:
 
-- `explicit` - explicit content which may not be appropriate for minors: nudity, profanity, etc
+- `nudity` - depictions of nudity, porn, etc.
+- `profanity` - profanity, hateful speech, etc.
 - `illegal` - something which may be illegal in some jurisdiction
 - `spam` - spam
 - `impersonation` - someone pretending to be someone else
@@ -37,7 +39,7 @@ Example events
 {
   "kind": 1984,
   "tags": [
-    [ "p", <pubkey>, "explicit"]
+    [ "p", <pubkey>, "nudity"]
   ],
   "content": "",
   ...
--- a/56.md
+++ b/56.md
@@ -66,5 +66,5 @@ Relay behavior
 --------------
 
 It is not recommended that relays perform automatic moderation using reports,
-as they can be easily gamed. Admins could use reports from trused moderators to
-takedown illegal or explicit content if the relays does not allow such things.
+as they can be easily gamed. Admins could use reports from trusted moderators to
+takedown illegal or explicit content if the relay does not allow such things.
--- a/56.md
+++ b/56.md
@@ -30,27 +30,37 @@ A `report` tag MUST be included which consists of the following report types:
 
 Some report tags only make sense for profile reports, such as `impersonation`
 
-Example code
+Example events
+--------------
 
-```swift
-enum ReportType: String {
-    case explicit
-    case illegal
-    case spam
-    case impersonation
+```json
+{
+  "kind": 1984,
+  "tags": [
+    [ "p", <pubkey>, "explicit"]
+  ],
+  "content": "",
+  ...
 }
 
-func create_report_tags(target: ReportTarget, type: ReportType) -> [[String]] {
-    var tags: [[String]]
-    switch target {
-    case .user(let pubkey):
-        tags = [["p", pubkey]]
-    case .note(let notet):
-        tags = [["e", notet.note_id], ["p", notet.pubkey]]
-    }
+{
+  "kind": 1984,
+  "tags": [
+    [ "e", <eventId>, "illegal"],
+    [ "p", <pubkey>]
+  ],
+  "content": "He's insulting the king!",
+  ...
+}
 
-    tags.append(["report", type.rawValue])
-    return tags
+{
+  "kind": 1984,
+  "tags": [
+    [ "p", <impersonator pubkey>, "impersonation"],
+    [ "p", <victim pubkey>]
+  ],
+  "content": "Profile is imitating #[1]",
+  ...
 }
 ```
 
--- a/56.md
+++ b/56.md
@@ -1,0 +1,70 @@
+
+NIP-56
+======
+
+Reporting
+---------
+
+`draft` `optional` `author:jb55`
+
+A report is a `kind 1984` note that is used to report other notes for spam,
+illegal and explicit content.
+
+The content MAY contain additional information submitted by the entity
+reporting the content.
+
+Tags
+----
+
+The report event MUST include a `p` tag referencing the pubkey of the user you
+are reporting.
+
+If reporting a note, an `e` tag MUST also be included referencing the note id.
+
+A `report` tag MUST be included which consists of the following report types:
+
+- `explicit` - explicit content which may not be appropriate for minors: nudity, profanity, etc
+- `illegal` - something which may be illegal in some jurisdiction
+- `spam` - spam
+- `impersonation` - someone pretending to be someone else
+
+Some report tags only make sense for profile reports, such as `impersonation`
+
+Example code
+
+```swift
+enum ReportType: String {
+    case explicit
+    case illegal
+    case spam
+    case impersonation
+}
+
+func create_report_tags(target: ReportTarget, type: ReportType) -> [[String]] {
+    var tags: [[String]]
+    switch target {
+    case .user(let pubkey):
+        tags = [["p", pubkey]]
+    case .note(let notet):
+        tags = [["e", notet.note_id], ["p", notet.pubkey]]
+    }
+
+    tags.append(["report", type.rawValue])
+    return tags
+}
+```
+
+Client behavior
+---------------
+
+Clients can use reports from friends to make moderation decisions if they
+choose to. For instance, if 3+ of your friends report a profile as explicit,
+clients can have an option to automatically blur photos from said account.
+
+
+Relay behavior
+--------------
+
+It is not recommended that relays perform automatic moderation using reports,
+as they can be easily gamed. Admins could use reports from trused moderators to
+takedown illegal or explicit content if the relays does not allow such things.
--- a/README.md
+++ b/README.md
@@ -30,6 +30,7 @@ NIPs stand for **Nostr Implementation Possibilities**. They exist to document wh
 - [NIP-40: Expiration Timestamp](40.md)
 - [NIP-42: Authentication of clients to relays](42.md)
 - [NIP-50: Keywords filter](50.md)
+- [NIP-56: Reporting](56.md)
 
 ## Event Kinds
 
@@ -48,6 +49,7 @@ NIPs stand for **Nostr Implementation Possibilities**. They exist to document wh
 | 43          | Channel Hide Message        | [28](28.md)            |
 | 44          | Channel Mute User           | [28](28.md)            |
 | 45-49       | Public Chat Reserved        | [28](28.md)            |
+| 1984        | Reporting                   | [56](56.md)            |
 | 22242       | Client Authentication       | [42](42.md)            |
 | 10000-19999 | Replaceable Events Reserved | [16](16.md)            |
 | 20000-29999 | Ephemeral Events Reserved   | [16](16.md)            |


monlovesmango:
thanks for putting this together! looks good to me. pretty straightforward. 

do you think it would be useful to make the report type queryable? maybe just use 'r' tag instead of 'report'?

56.md
```diff
@@ -0,0 +1,70 @@
...
+Reporting
+---------
+
+`draft` `optional` `author:jb55`
+
+A report is a `kind 1984` note that is used to report other notes for spam,
+illegal and explicit content.
+
+The content MAY contain additional information submitted by the entity
+reporting the content.
```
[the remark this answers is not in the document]
jb55: I was thinking this would just be a human comment

56.md
```diff
@@ -0,0 +1,70 @@
...
+choose to. For instance, if 3+ of your friends report a profile as explicit,
+clients can have an option to automatically blur photos from said account.
+
+
+Relay behavior
+--------------
+
+It is not recommended that relays perform automatic moderation using reports,
+as they can be easily gamed. Admins could use reports from trused moderators to
+takedown illegal or explicit content if the relays does not allow such things.
```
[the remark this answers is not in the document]
jb55: thanks

punterwantsawhalepass:
Screenshot 2023-01-28 5 07 12 PM

mikedilger (approved):
Very much how I imagined it should be. I don't have any edits to offer.

There may be other types of reports over time; we can add them as we think of them.

56.md
```diff
@@ -0,0 +1,70 @@
...
+    case .user(let pubkey):
+        tags = [["p", pubkey]]
+    case .note(let notet):
+        tags = [["e", notet.note_id], ["p", notet.pubkey]]
+    }
+
+    tags.append(["report", type.rawValue])
+    return tags
+}
+```
```
Giszmo, line 55: How about actual example events instead of the example code? The code is a

````suggestion
Example events
--------------

```json
{
  "kind": 1984,
  "tags": [
    [ "report", "explicit"],
    [ "p", <pubkey>]
  ],
  "content": "",
  ...
}

{
  "kind": 1984,
  "tags": [
    [ "report", "illegal"],
    [ "e", <eventId>],
    [ "p", <pubkey>]
  ],
  "content": "He's insulting the king!",
  ...
}

{
  "kind": 1984,
  "tags": [
    [ "report", "impersonation"],
    [ "p", <pubkey>]
  ],
  "content": "Profile is imitating npub12345...",
  ...
}
```
````

**Review by Giszmo — CHANGES_REQUESTED** [review]

File: 56.md (hunk 3 above)
Comment (Giszmo) [reply 56.md:55]: Writing these examples I noticed it would be helpful to better reference impersonation victims with their own p-tag, so maybe the target of the report could carry the report-tag value?

```json
{
  "kind": 1984,
  "tags": [
    [ "p", <pubkey>, "explicit"]
  ],
  "content": "",
  ...
}

{
  "kind": 1984,
  "tags": [
    [ "e", <eventId>, "illegal"],
    [ "p", <pubkey>]
  ],
  "content": "He's insulting the king!",
  ...
}

{
  "kind": 1984,
  "tags": [
    [ "p", <impersonator pubkey>, "impersonation"],
    [ "p", <victim pubkey>]
  ],
  "content": "Profile is imitating #[2]",
  ...
}
```

mikedilger: I like that better

56.md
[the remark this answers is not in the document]
fiatjaf: Human comment, please!

vitorpamplona:
If a report is many types in one (e.g. illegal and explicit), is it recommended to send separate events or just more than one "report" tag?

56.md
mikedilger: I mean the report kinds in the tags themselves.  But I like showing the events too.

gkbrk:
> `explicit` - explicit content which may not be appropriate for minors: nudity, profanity, etc

Do we want someone posting hardcore porn and someone saying "fuck" on a tweet to generate the same report signal? I think it would be more useful for other clients if those two cases could be separate report types.

A use case I can think of is hiding NSFW profiles in "work mode" but not hiding users and post that say "shit, the weather sucks today".

dskvr:
Just here to lol @ `kind:1984`

ursuscamp:
What would we think about adding a hint for relays to keep a report private? It could be `private` (for relay eyes only) or `trusted` (for friends' eyes only) or `public`? Reporting bad behavior can be tricky, and some people may avoid reporting a genuine issue if they feel that public reports would open them to retaliation or harassment.

jb55:
> If a report is many types in one (e.g. illegal and explicit), is it recommended to send separate events or just more than one "report" tag?

probably makes sense to just send more report tags.

56.md
jb55: yeah this is fine I suppose, report kinds in the tags without a separate report tag

jb55:
Going with @Giszmo suggestion of having the report tag types tied to the p and e tags of the thing that is being reported. it's more clear.

jb55:
I split explicit into profanity and nudity.

jb55:
> What would we think about adding a hint for relays to keep a report private? It could be `private` (for relay eyes only) or `trusted` (for friends' eyes only) or `public`? Reporting bad behavior can be tricky, and some people may avoid reporting a genuine issue if they feel that public reports would open them to retaliation or harassment.

nostr is not very private in general, private reporting would have to be a separate nip.

Giszmo (approved):

Giszmo:
@Semisol Can you live with the complaint being human readable and switch to ACK?

majestrate:
the jurisdiction applicable for the report, specifically wrt illegal content feels like information that is required that is currently omitted. 
e.g. nudity in french jurisdiction is also illegal in UEA jurisdiction, but also likely offensive in others.

fiatjaf:
If no one objects I'll merge this today.

Redpug111 (approved):

Aspie96:
To be honest, I think this draft is problematic.

Specifically, my issue is with:

> A `report type` string MUST be included as the 3rd entry to the `e` or `p` tag
> being reported, which consists of the following report types:
> 
> - `nudity` - depictions of nudity, porn, etc.
> - `profanity` - profanity, hateful speech, etc.
> - `illegal` - something which may be illegal in some jurisdiction
> - `spam` - spam
> - `impersonation` - someone pretending to be someone else

Why? What's so special about those categories of content?

Wanting to ban illegal content and spam is rather universal. The first is necessary if you don't want the police to knock at your door, the second if you want your service to remain useful. Impersonation is also often illegal.

But nudity and profanity seem completely arbitrary to me. This is not to say banning either of them is a bad idea. And it's not to say many wouldn't want to ban to them. It is to say, however, that it's an entirely subjective and circumstantial multi-factor decision, and some that other kinds of content may be just as ban-worthy to other people.

And while I understand Damus specifically needed this for compliance with Apple's terms of use, I don't believe the TOS of a specific company should affect what goes in the specification of the protocol.

There should, instead, be some system to report arbitrary kinds of content to those relays that ban that specific kind of content. The protocol itself should be neutral in regards to what content is or is not "ban worthy".

And as for "illegal", I should mention that it means almost exactly nothing if a specific jurisdiction isn't specified. Laws are territorial.

Before seeing this proposal, I have thought of a report system which wouldn't require the NIP itself mentioning some specific kinds of content which offend the whims of some specific company.

I most surely don't claim, or even remotely suggest, that I could do a better job than jb55, but I do feel very uncomfortable with the report types.

majestrate:
On Mon, 27 Feb 2023 21:32:33 -0800
Valentino Giudice ***@***.***> wrote:

> To be honest, I think this draft is problematic.
> 
> Specifically, my issue is with:
> 
> > A `report type` string MUST be included as the 3rd entry to the `e`
> > or `p` tag being reported, which consists of the following report
> > types:
> > 
> > - `nudity` - depictions of nudity, porn, etc.
> > - `profanity` - profanity, hateful speech, etc.
> > - `illegal` - something which may be illegal in some jurisdiction
> > - `spam` - spam
> > - `impersonation` - someone pretending to be someone else
> 
> Why? What's so special about those categories of content?
> 
> Wanting to ban illegal content and spam is rather universal. The
> first is necessary if you don't want the police to knock at your
> door, the second if you want your service to remain useful.
> Impersonation is also often illegal.
> 
> But nudity and profanity seem completely arbitrary to me. This is not
> to say banning either of them is a bad idea. And it's not to say many
> wouldn't want to ban to them. It is to say, however, that it's an
> entirely subjective and circumstantial multi-factor decision, and
> some that other kinds of content may be just as ban-worthy to other
> people.
> 
> And while I understand Damus specifically needed this for compliance
> with Apple's terms of use, I don't believe the TOS of a specific
> company should affect what goes in the specification of the protocol.
> 
> There should, instead, be some system to report arbitrary kinds of
> content to those relays that ban that specific kind of content. The
> protocol itself should be neutral in regards to what content is or is
> not "ban worthy".
> 
> And as for "illegal", I should mention that it means almost exactly
> nothing if a specific jurisdiction isn't specified. Laws are
> territorial.

I would go further and say illegal requires citing explicitly the law
in question. that would greatly benifit from a having an
explicit strict structured form, so you dont just reference "the law of
it hurts my sensibilities section 9001, in the criminal code of i dont
like this post". having the jurisdiction AND laws infringed on is
really not optional when making assertions that have legal
ramifications.

> 
> Before seeing this proposal, I have thought of a report system which
> wouldn't require the NIP itself mentioning some specific kinds of
> content which offend the whims of some specific company.
> 
> I most surely don't claim, or even remotely suggest, that I could do
> a better job than jb55, but I do feel very uncomfortable with the
> report types.
> 



-- 
~jeff

vitorpamplona:
> I would go further and say illegal requires citing explicitly the law in question. 

Folks, this NIP is just a flagging event. Relays and other providers/clients should aggregate these reports, review them, and take appropriate action. Users don't have the ability to provide a legal option about a matter. They can just provide feedback on the content they see and don't want to see anymore and/or don't want their followers to see it. The rest is for other entities in the field and law enforcement to navigate.

fiatjaf:
@Aspie96 I agree with you. We can reword the NIP to say the report-type reason could be anything, and list some examples. We can mention the fact that relays might have arbitrary policies and that these arbitrary policies could be used in reports and as basis for banning pubkey from these relays.

If you can submit a PR making these modifications that will be great.

fiatjaf:
@majestrate I don't agree with you. Nostr is not a legal court, it's a neutral system. The laws of the United States have pretty much the same validity as the laws of it hurt my feelings here. Everything is subjective to the parties involved.

Aspie96:
> I would go further and say illegal requires citing explicitly the law in question.

Not sure if I agree.

I think many people know that something is illegal without knowing the specific article of the specific law.

In fact, to be able to operate in one own's jurisdiction, one ought to have some understanding of the laws they have to follow, and therefore be able to identify illegal behavior, but it doesn't mean they know the reference to any specific law.

My point stands, however, that "illegal" doesn't mean anything if a specific jurisdiction isn't mentioned.

> @Aspie96 I agree with you. We can reword the NIP to say the report-type reason could be anything, and list some examples

My issue is with the report type being a word, as it currently is.

This necessarily requires the meaning of those words to be specified either in NIP-56 or in some other NIP.

There are potentially infinite rules one could apply. And there are many kinds of content one could very reasonably want to forbid which do not follow in any of the above categories.

I regret not seeing this NIP before and not developing my idea previously. I will try to do my best to write something. I will ask for feedback before sending a pull request, however.

majestrate:
On Tue, 28 Feb 2023 07:23:28 -0800
fiatjaf_ ***@***.***> wrote:

> @majestrate I don't agree with you. Nostr is not a legal court, it's
> a neutral system. The laws of the United States have pretty much the
> same validity as the laws of it hurt my feelings here. Everything is
> subjective to the parties involved.
> 



i think there is a lot of room for improvement here. adding structured 
metadata to reports when content is illegal would be a boon to those
making tooling. it would be able to make legal complaince in every
jurisdiction SO much easier. given the direction of the winds of
change, this is something we will need in the future regardless of how
we feel about it.

i'll just state that i think this "every law is just as valid as my
feelings" stance is a bad default stance. the biggest gripe with
decentralized moderation as implemented in every other system i have
seen is this exact sentament, ambiguity of report severity. adding
explicit and well defined structure to reports will only do good for
the future tooling because someone posting a swear word is absolutely
not the same severity as CSAM.

-- 
~jeff

Aspie96:
> a swear word is absolutely not the same severity as CSAM

CSAM is illegal pretty much everywhere, but that is the exception, not the rule. Most laws differ by jurisdiction.

And even in the case of child pornography, what even counts as child pornography can change quite wildly, with the exact same image being a criminal liability in one jurisdiction and perfectly in the clear in another.

And a "swear word" could totally be a legal liability depending on context and jurisdiction.

Do I feel that CSAM are worse morally than swear words? Yes. Do I want that feeling of mine to affect the protocol? No.

majestrate:
> > a swear word is absolutely not the same severity as CSAM
> 
> CSAM is illegal pretty much everywhere, but that is the exception, not the rule. Most laws differ by jurisdiction.
> 
> And even in the case of child pornography, what even counts as child pornography can change quite wildly, with the exact same image being a criminal liability in one jurisdiction and perfectly in the clear in another.
> 
> And a "swear word" could totally be a legal liability depending on context and jurisdiction.
> 
> Do I feel that CSAM are worse morally than swear words? Yes. Do I want that feeling of mine to affect the protocol? No.

that is why it is important to clarify what constitutes the claimed illegality and where it applies, because insulting the prophet Mohamed is in fact almost just as illegal in some parts of the world.

majestrate:
something to note, not all jurisdictions are compatible with each other, defaulting to assume everyone is in the sphere of influence of the USA blocks you off from very sizable regions online, when being able to interoperate in those regions can be possible using proper classification on reports. adding such opens up the protocol to being accessible in so many more places as infra can have low enough risk to host in more prudish countries.

fiatjaf:
> This necessarily requires the meaning of those words to be specified either in > NIP-56 or in some other NIP.

What is the problem with it being a word? Various words could be used by clients and relays according to whatever they think it makes sense. Once some word becomes a de facto standard we can modify this PR to add it, otherwise it can live as a de facto standard outside of the NIPs too.

Aspie96:
My issue is that without at least a short definition it would probably be way too vague.

As for including future de-facto standards in the repository, it would be worse, since it may be seen as endorsement of those rules, but some may be rather terrible.

For example, many relays, in many countries, could forbid anything related to homosexuality. I don't know if it will happen, but it's at least very possible.

Those who think such rules are completely and utterly unreasonable, perhaps even immoral, wouldn't want them in the repo, effectively requiring a vetting process for new future rule identifiers.

Regardless, the current NIP expressly says that the report type "consists of the following [...] types", so using another type name would actually be a violation of the NIP-56 specification.

Aspie96:
I opened a discussion about this: https://github.com/nostr-protocol/nips/discussions/309

(I used a discussion instead of commenting here because I am no longer addressing the text of this PR directly).

fiatjaf:
> Regardless, the current NIP expressly says that the report type "consists of the following [...] types", so using another type name would actually be a violation of the NIP-56 specification.

That's why I asked you to change the NIP to soften that rule.

pgerstbach:
I just checked my relay for reports and I only found reports on notes that are not even stored on my relay. Wouldn't it make sense to recommend in this NIP to send the report only to relays that actually have stored the note in question? Of course those reports can easily be filtered out on the relay itself as well. So another possibility is to recommnd that a relay does not need to store a report, if it does not have stored the event in question.

fiatjaf:
Yes, both these things make sense. If you want to submit a patch adding these two recommendations that will be great.

vitorpamplona:
I also think relays should take an action based on the report. If the relay doesn't agree with the report, it should delete the report event.

That will help with spammers reporting good users.

Giszmo:
We should add kyc to nostr to decide who's claim is valid.

On a more serious note, reports about non-existing events might be worth storing as those events could float around and reach you later.

In how far a relay can and should do anything about reports is another topic. Maybe it should enrich reported events with `reported: true` so the client can query the reports.

vitorpamplona:
> On a more serious note, reports about non-existing events might be worth storing as those events could float around and reach you later.

Especially if the post is coming from the author's own relay, which is incentivized to delete reports against the author.

PR #206, opened 2023-01-28
Patch-1

punterwantsawhalepass:


fiatjaf:
This is not the way to do it. You can open comments and edit suggestions on the other PR.

PR #211, opened 2023-02-01
NIP-19: add `nchannel` prefix

yukibtc:
Add NIP-19 bech32 prefix for channels.

### DIFF
--- a/19.md
+++ b/19.md
@@ -36,6 +36,7 @@ These are the possible bech32 prefixes with `TLV`:
   - `nevent`: a nostr event
   - `nrelay`: a nostr relay
   - `naddr`: a nostr parameterized replaceable event coordinate (NIP-33)
+  - `nchannel`: channel ids
 
 These possible standardized `TLV` types are indicated here:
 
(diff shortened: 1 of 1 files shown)


fiatjaf:
Nice. Is it implemented somewhere?

yukibtc:
> Nice. Is it implemented somewhere?

I'll implement here: https://github.com/rust-nostr/nostr

yukibtc:
> Nice. Is it implemented somewhere?



> Nice. Is it implemented somewhere?

Do you think that is better to add this prefix in `Shareable identifiers with extra metadata`?
Or leave as I have done?

0xtrr:
I think it would be good to add an example under the [Examples](https://github.com/nostr-protocol/nips/blob/master/19.md#examples) section in this document as well, just like we have with npub and nsec.

yukibtc:
> I think it would be good to add an example under the [Examples](https://github.com/nostr-protocol/nips/blob/master/19.md#examples) section in this document as well, just like we have with npub and nsec.

Yeah, I'll add an example in the next days

vitorpamplona:
Do we really need semantically different `eventid+relay` schemes? 

Otherwise, we are going to have ndm, nlongformtext, nproduct (NIP-45), etc..

fiatjaf:
Yeah, I think it is better to not have more different prefixes.

PR #212, opened 2023-02-01
NIP-64: Chess (WIP)

theborakompanioni:
Work-in-progress.

Reserving `kind := 30` for messages representing Chess games/moves or general mechanics to represent text-based games.

This is currently used by Jester app:
- Code: https://github.com/jesterui/jesterui
- Demo: https://jesterui.github.io/

Not yet clear on what format to settle to be future-proof and/or to support all kinds of different game mechanics.

e.g. from the [SGF format Wiki page](https://en.wikipedia.org/wiki/Smart_Game_Format):
> The main purposes of SGF are to store records of played games and to provide features for storing annotated and analyzed games (e.g. board markup, [variations](https://en.wikipedia.org/wiki/Variation_(game_tree))). It is a text-only, tree-based format. The tree structure makes the addition of variations simple. It is also text-based instead of [binary](https://en.wikipedia.org/wiki/Binary_file) for the sake of portability.

### DIFF
--- a/64.md
+++ b/64.md
@@ -35,3 +35,7 @@ tbd.
 ## Client Behavior
 
 tbd.
+
+## Additional info
+- [PGN Specification](https://github.com/mliebelt/pgn-spec-commented/blob/main/pgn-specification.md) PGN (Portable Game Notation) specification
+- [PGN Specification Supplement](https://github.com/mliebelt/pgn-spec-commented/blob/main/pgn-spec-supplement.md) Addition for adding graphical elements, clock values, eval, ...
--- a/64.md
+++ b/64.md
@@ -1,0 +1,37 @@
+NIP-64
+======
+
+Chess
+-----
+
+`draft` `optional` `author:theborakompanioni`
+
+This NIP defines a new event kind for messages representing Chess games.
+
+It reserves event kind `30` for immediate use.
+
+- `30 - chess game event`
+
+## Abstract
+
+tbd.
+
+## Motivation
+
+tbd.
+
+## Implementation
+
+tbd.
+
+#### Example
+
+tbd.
+
+## Relay Behavior
+
+tbd.
+
+## Client Behavior
+
+tbd.


kmkhami:
Why is it preferable to use a `kind` over a `tag`? I would imagine that a `tag` could add more value as it could include metadata like the `guid` for the game for which the `event` is for.

fiatjaf:
I'd settle for a very quick description of the messages, no need to write an article about motivation, abstract, nothing. You can just say: "This is a NIP that standardizes chess games over Nostr", but not even this is needed. If the title says "Chess" everybody already knows what to expect.

fiatjaf:
No need to think very much about future-proof. The simplest thing is better. Isn't the SGF format good? I think it is good, but I'm not a chess player.

itslenny:
imo this is too specific. Maybe we could reserve a kind for "game event" or even "app-to-app data" where the content is arbitrary app-specific data, and it requires a unique tag like "chess-v1" or something.

theborakompanioni:
@kmkhami 
> Why is it preferable to use a `kind` over a `tag`? I would imagine that a `tag` could add more value as it could include metadata like the `guid` for the game for which the `event` is for.

Don't want to spam the global feed of `kind := 1` messages.

@fiatjaf 
> I'd settle for a very quick description of the events, no need to write an article about motivation, abstract, nothing. You can just say: "This is a NIP that standardizes chess games over Nostr", but not even this is needed. If the title says "Chess" everybody already knows what to expect.
> 
> A link at the end to your implementation settles any remaining doubt.

:+1:

@fiatjaf 
> Nevermind, SGF is for Go. Is this NIP about Go? I think Go would be a different kind, even if very similar to this. The same app will _not_ implement both Chess and Go, so no need to be future-proof.

Did not investigate too much by now. It is for all kind of board games:
> The Smart Game Format (SGF) is a computer [file format](https://en.wikipedia.org/wiki/File_format) used for storing records of [board games](https://en.wikipedia.org/wiki/Board_game). [Go](https://en.wikipedia.org/wiki/Go_(game)) is the game that is most commonly represented in this format and is the [default](https://en.wikipedia.org/wiki/Default_(computer_science)).

Currently, Jester uses [FEN](https://en.wikipedia.org/wiki/Forsyth%E2%80%93Edwards_Notation), but this is too limited. At the moment I am testing to switch over to [PGN](https://en.wikipedia.org/wiki/Portable_Game_Notation), which would be chess-only and cover everything needed.

@itslenny 
> imo this is too specific. Maybe we could reserve a kind for "game event" or even "app-to-app data" where the content is arbitrary app-specific data, and it requires a unique tag like "chess-v1" or something.

Yes, I was thinking alike. Maybe it _can_ be flexible. But as @fiatjaf said, "This NIP should be about chess, simple and quick.". Let's see what comes out of it. Any suggestion is highly appreciated :pray: 
I'll focus on supporting everything needed for chess games for now.

1440000bytes:
Not sure why this needs a NIP and there is nothing to review

fiatjaf:
It definitely needs a NIP and it is marked as a draft.

ekzyis:
> Not sure why this needs a NIP and there is nothing to review

I think which format is used should be standardized since clients need to support the same format to be compatible.

Or it should be mentioned how the format that a chess event uses is specified; for example by using a JSON string for `content` with appropriate keys inside: "fen" for FEN format, "pgn" for PGN format etc.

fiatjaf:
@ekzyis
> Or it should be mentioned how the format that a chess event uses is specified; for example by using a JSON string for content with appropriate keys inside: "fen" for FEN format, "pgn" for PGN format etc.

I disagree. Just use a different kind for each. Also if you're doing chess, just use always the same format, not two different formats, that just doubles the work for nothing. If you're using Go do a different kind and pick a format for Go and then everybody uses that. And so on.

@theborakompanioni 
> Currently, Jester uses FEN, but this is too limited. At the moment I am testing to switch over to PGN, which would be chess-only and covers everything needed.

I don't know what I am talking about, but I think you should pick the one that has more libraries available for it in many languages. Unless it is so easy to parse that anyone can do it by hand, in that case the easiest one is better.

staab:
Closing due to inactivity. This would be a good candidate for inclusion via an external link like nostrocket if you have a nip @theborakompanioni

theborakompanioni:
> Closing due to inactivity. This would be a good candidate for inclusion via an external link like nostrocket if you have a nip @theborakompanioni

Hey @staab, I understand. Thanks for leaving it open such a long period of time. :orange_heart: 

I recently picked up looking into it once again. Maybe something will come out of it, but I would not bet on it. So closing is probably for the best. :pray:

staab:
I'll leave it closed for now but feel free to reopen whenever

ekzyis:
> Maybe something will come out of it, but I would not bet on it.

I would bet on it :)

PR #218, opened 2023-02-03
NIP-65 Relay List Metadata (was Feed Advertisements)

mikedilger:
Gossip shouldn't have to be so complicated.  Nostr should scale better.  People should be able to find what they want.

This is related to nip-23 kind10001, except that that NIP is really more about people controlling what their clients do rather than to advertise to others, and has complex hard to code runes language stuff in it and has been sitting around for a long time now.

This is related to kind-3 contact list content, except that Contact Lists are far too long for this purpose and their content was also intended for users to control their own clients rather than to advertise to others.

So I thought a new kind was called for.

### DIFF
--- a/65.md
+++ b/65.md
@@ -1,0 +1,59 @@
+NIP-65
+======
+
+Feed Advertisements
+-------------------
+
+`draft` `optional` `author:mikedilger`
+
+A special replaceable event meaning "Feed Advertisement" is defined as an event with kind `10002` having a list of `r` tags, one for each relay the author posts their outgoing "feed" to.
+
+The content is not used.
+
+A feed consists of events that the author posts and wishes for their followers to "follow".  Normally these would be kind-1 Text Note events but is not limited as such.
+
+Clients SHOULD, as with all replaceable events, use only the most recent kind-10002 event they can find.
+
+Clients SHOULD presume that if their user has a pubkey in their ContactList (kind 3) that it is because they wish to see this author's "feed" events. But clients MAY presume otherwise.
+
+Clients SHOULD interpret these Feed Advertisements as being the best relays to source the "feed" events from the author of the feed advertisement.
+
+### Motivation
+
+There is a common nostr use case where users wish to follow the content produced by other users. This is evidenced by the implicit meaning of the Contact List in [NIP-02](02.md)
+
+Because users don't often share the same sets of relays, ad-hoc solutions have arisen to get that content, but these solutions negatively impact scalability and decentralization:
+
+  - Most people are sending their posts to the same most popular relays in order to be more widely seen
+  - Many people are pulling from a large number of relays (including many duplicate events) in order to get more data
+  - Events are being copied between relays, oftentimes to many different relays
+
+### Purposes
+
+The purpose of this NIP is to help clients find the events of the people they follow, and to help nostr scale better.
+
+### Suggestions
+
+It is suggested that people spread their kind-10002 events to many relays, but to post their normal feed notes to a much smaller number of relays.  It is suggested that clients offer a way for users to spread their Feed Advertisements to many more relays than they normally post to.
+
+Authors may post events outside of the feed that they wish their followers to follow by posting them to relays outside of those listed in their "Feed Advertisement".  For example, an author may want to reply to someone without all of their followers watching, but in many cases they want all their replies to show up in their "feed".
+
+It is suggsted that relays allow any user to write their own kind-10002 event (optionally with AUTH to verify it is their own) even if they are not otherwise subscribed to the relay because
+
+  - finding where someone posts is rather important
+  - these events do not have content that needs management
+  - relays only need to store one replaceable event per pubkey to offer this service
+
+### Example
+
+```json
+{
+  "kind": 10002,
+  "tags": [
+    ["r", "wss://alicerelay.example.com"],
+    ["r", "wss://brando-relay.com"],
+    ["r", "wss://expensive-relay.example2.com"],
+  ],
+  "content": "",
+  ...other fields
+```
--- a/65.md
+++ b/65.md
@@ -22,7 +22,7 @@ If an author advertises a write relay in a kind `10002` event, that means that f
 
 Clients SHOULD presume that if their user has a pubkey in their ContactList (kind 3) that it is because they wish to see that author's feed-related events. But clients MAY presume otherwise.
 
-If an author advertises a read relay in a kind `10002` event, that means that the author may be subscribed to events that tag them on such relays. Clients SHOULD publish events that tag someone on at least the read relays of the person being tagged.
+If an author advertises a read relay in a kind `10002` event, that means that the author may be subscribed to events that tag them on such relays. Clients SHOULD publish events that tag someone on at least some of the read relays of the person being tagged.
 
 ### Motivation
 
@@ -44,7 +44,7 @@ It is suggested that people spread their kind `10002` events to many relays, but
 
 Authors may post events outside of the feed that they wish their followers to follow by posting them to relays outside of those listed in their "Relay List Metadata".  For example, an author may want to reply to someone without all of their followers watching.
 
-It is suggsted that relays allow any user to write their own kind `10002` event (optionally with AUTH to verify it is their own) even if they are not otherwise subscribed to the relay because
+It is suggested that relays allow any user to write their own kind `10002` event (optionally with AUTH to verify it is their own) even if they are not otherwise subscribed to the relay because
 
   - finding where someone posts is rather important
   - these events do not have content that needs management
--- a/65.md
+++ b/65.md
@@ -1,22 +1,28 @@
 NIP-65
 ======
 
-Feed Advertisements
+Relay List Metadata
 -------------------
 
 `draft` `optional` `author:mikedilger`
 
-A special replaceable event meaning "Feed Advertisement" is defined as an event with kind `10002` having a list of `r` tags, one for each relay the author posts their outgoing "feed" to.
+A special replaceable event meaning "Relay List Metadata" is defined as an event with kind `10002` having a list of `r` tags, one for each relay the author uses to either read or write to.
 
-The content is not used.
+The primary purpose of this relay list is to advertise to others, not for configuring one's client.
 
-A feed consists of events that the author posts and wishes for their followers to "follow".  Normally these would be kind-1 Text Note events but is not limited as such.
+The content is not used and SHOULD be blank.
+
+The `r` tags can have a second parameter as either `read` or `write`. If it is omitted, it means the author both reads from and writes to that relay.
 
 Clients SHOULD, as with all replaceable events, use only the most recent kind-10002 event they can find.
 
-Clients SHOULD presume that if their user has a pubkey in their ContactList (kind 3) that it is because they wish to see this author's "feed" events. But clients MAY presume otherwise.
+### The meaning of read and write
 
-Clients SHOULD interpret these Feed Advertisements as being the best relays to source the "feed" events from the author of the feed advertisement.
+If an author advertises a write relay in a kind `10002` event, that means that feed-related events created by the author, which the author wants their followers to see, will be posted there. Normally these would be kind-1 Text Note events, but are not limited as such.
+
+Clients SHOULD presume that if their user has a pubkey in their ContactList (kind 3) that it is because they wish to see that author's feed-related events. But clients MAY presume otherwise.
+
+If an author advertises a read relay in a kind `10002` event, that means that the author may be subscribed to events that tag them on such relays. Clients SHOULD publish events that tag someone on at least the read relays of the person being tagged.
 
 ### Motivation
 
@@ -30,19 +36,23 @@ Because users don't often share the same sets of relays, ad-hoc solutions have a
 
 ### Purposes
 
-The purpose of this NIP is to help clients find the events of the people they follow, and to help nostr scale better.
+The purpose of this NIP is to help clients find the events of the people they follow, to help tagged events get to the people tagged, and to help nostr scale better.
 
 ### Suggestions
 
-It is suggested that people spread their kind-10002 events to many relays, but to post their normal feed notes to a much smaller number of relays.  It is suggested that clients offer a way for users to spread their Feed Advertisements to many more relays than they normally post to.
+It is suggested that people spread their kind `10002` events to many relays, but write their normal feed-related events to a much smaller number of relays (between 2 to 6 relays). It is suggested that clients offer a way for users to spread their kind `10002` events to many more relays than they normally post to.
 
-Authors may post events outside of the feed that they wish their followers to follow by posting them to relays outside of those listed in their "Feed Advertisement".  For example, an author may want to reply to someone without all of their followers watching, but in many cases they want all their replies to show up in their "feed".
+Authors may post events outside of the feed that they wish their followers to follow by posting them to relays outside of those listed in their "Relay List Metadata".  For example, an author may want to reply to someone without all of their followers watching.
 
-It is suggsted that relays allow any user to write their own kind-10002 event (optionally with AUTH to verify it is their own) even if they are not otherwise subscribed to the relay because
+It is suggsted that relays allow any user to write their own kind `10002` event (optionally with AUTH to verify it is their own) even if they are not otherwise subscribed to the relay because
 
   - finding where someone posts is rather important
   - these events do not have content that needs management
   - relays only need to store one replaceable event per pubkey to offer this service
+
+### Why not in kind `0` Metadata
+
+Even though this is user related metadata, it is a separate event from kind `0` in order to keep it small (as it should be widely spread) and to not have content that may require moderation by relay operators so that it is more acceptable to relays.
 
 ### Example
 
@@ -52,7 +62,8 @@ It is suggsted that relays allow any user to write their own kind-10002 event (o
   "tags": [
     ["r", "wss://alicerelay.example.com"],
     ["r", "wss://brando-relay.com"],
-    ["r", "wss://expensive-relay.example2.com"],
+    ["r", "wss://expensive-relay.example2.com", "write"],
+    ["r", "wss://nostr-relay.example.com", "read"],
   ],
   "content": "",
   ...other fields
--- a/README.md
+++ b/README.md
@@ -48,6 +48,7 @@ NIPs stand for **Nostr Implementation Possibilities**. They exist to document wh
 | 43          | Channel Hide Message        | [28](28.md)            |
 | 44          | Channel Mute User           | [28](28.md)            |
 | 45-49       | Public Chat Reserved        | [28](28.md)            |
+| 10002       | Feed Advertisement          | [65](65.md)            |
 | 22242       | Client Authentication       | [42](42.md)            |
 | 10000-19999 | Replaceable Events Reserved | [16](16.md)            |
 | 20000-29999 | Ephemeral Events Reserved   | [16](16.md)            |
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ NIPs stand for **Nostr Implementation Possibilities**. They exist to document wh
 | 43          | Channel Hide Message        | [28](28.md)            |
 | 44          | Channel Mute User           | [28](28.md)            |
 | 45-49       | Public Chat Reserved        | [28](28.md)            |
-| 10002       | Feed Advertisement          | [65](65.md)            |
+| 10002       | Relay List Metadata         | [65](65.md)            |
 | 22242       | Client Authentication       | [42](42.md)            |
 | 10000-19999 | Replaceable Events Reserved | [16](16.md)            |
 | 20000-29999 | Ephemeral Events Reserved   | [16](16.md)            |


mikedilger:
Implemented in gossip on branch 'feed-advertisements' with some testing, but needs more testing, however:

I don't want to push out many of these events in case I am colliding with somebody else's kind number (though I searched for 10002 and didn't see it in this repo).

eskema:
I don't see the need for this because we can already say what relays we read and write from NIP23. this would not solve it, it would introduce a new thing to manage. I agree that the runes thing is complicated to implement if there's nothing that handles runes easily available. NostrTools should have support for that IMO for it to start to be used, I think that instead of this NIP, we could simply use NIP23 in a simplified mode, with only true false values that can later be replaced by runes. This would let most relays implementations switch more easily as it follows the same model of a list of relays with read and write values and would all be valid once runes are implemented

fiatjaf:
We should cancel NIP-23 and use this instead. It is good since we can query for `r` tags and learn what relays are more popular (is that a good thing?).

What do you think @dskvr @sondreb @monlovesmango?

eskema:
if we're going this route, then IMO this array should maybe be better in kind-0 as `content.relays:[]`. having what relays you post to is pretty integral part of identity, and it would reduce a lot of guessing and events back and fort since kind-0 is pretty much always needed... we already have a bunch of frivolous fields there, it's easily updateable and probably the most up to date everywhere. a simple list of relays there makes more sense to me in terms of practicality. 
if that doesn't make sense, and everyone prefers this model, then let's use this kind, but I think having the relay information readily available in kind:0 solves multiple problems at once.

KJsRepo:
I really like this NIP, it looks like it would solve the problem of not being able to find where authors are posting.

That said, I think it's solving a much different problem than NIP-23, which is about storing relays and rules for clients to connect to.  In fact, I think NIP-23 would support this one well.  If the user doesn't have access to the NIP-23 runes rules to clients, then it relies on the client to decide which relays to post 10002 events to (or implement an ad-hoc solution to allow the user to decide).  If we accept both NIPs, then it allows fine-grained interoperability between clients.

Regarding runes rules being complicated and no use-case:  We are obviously very early doors here, but I believe this is a non-issue.  Currently, 99% of everything is kind 1, so it's not too complicated to uncheck "read" on relays you don't want to see posts from (e.g. for spam prevention).  Going forward, as more kinds come about, users will absolutely want to block certain event kinds from certain relays.  The runes rules do seem a little daunting, but it's only as complicated as a user wants to make it.  Plus, it's only a matter of time before some dev figures out a UI paradigm that is intuitive for users.

fiatjaf:
> if we're going this route, then IMO this array should maybe be better in kind-0 as content.relays:[]. having what relays you post to is pretty integral part of identity, and it would reduce a lot of guessing and events back and fort since kind-0 is pretty much always needed...

Good point.

fiatjaf:
> Regarding runes rules being complicated and no use-case: We are obviously very early doors here, but I believe this is a non-issue. Currently, 99% of everything is kind 1, so it's not too complicated to uncheck "read" on relays you don't want to see posts from (e.g. for spam prevention). Going forward, as more kinds come about, users will absolutely want to block certain event kinds from certain relays. The runes rules do seem a little daunting, but it's only as complicated as a user wants to make it. Plus, it's only a matter of time before some dev figures out a UI paradigm that is intuitive for users.

I agree, but we must standardize these things once the need arises and solutions start to pop, not try to prevent what we will use and need while completely blind.

staab:
I think this is much better than NIP 23, especially since it recommends clients broadcast these events more aggressively than a user's regular notes. I also like @eskema's kind-0 idea, and it would make sense for clients to aggressively broadcast kind 0's for the same reason.

It does seem like the use case of saving which relays a user prefers to read from is still not yet solved. I wonder if a new kind for "account settings" would work well. It would work the same way as kind 0, which has become sort of an "account metadata" junk drawer, and which is mostly used by the clients of _other_ users. In contrast, kind `x` would be the same kind of junk drawer but encrypted and meant for the use of the user's own clients, including things like relay read/write settings, private follows, blocks, etc.

eskema:
> It does seem like the use case of saving which relays a user prefers to read from is still not yet solved.

the relays a user choose to read from will then get fetched from the pubkeys he follows. we can then look at all our contacts on kind-3 and assign one relay from this relays list to our kind-3 contact relay field. we can make the decision which one to use with a system like @mikedilger relay model, so the client should know where to fetch data just by looking at it's own kind-3 if nothing else available.

KJsRepo:
> It does seem like the use case of saving which relays a user prefers to read from is still not yet solved. I wonder if a new kind for "account settings" would work well. It would work the same way as kind 0, which has become sort of an "account metadata" junk drawer, and which is mostly used by the clients of _other_ users. In contrast, kind `x` would be the same kind of junk drawer but encrypted and meant for the use of the user's own clients, including things like relay read/write settings, private follows, blocks, etc.

I'm 100% for this.  The reason I picked up and ran with NIP-23 was to solve for the use-case of a relay that a user subscribes to for a very specific data stream, i.e. subscribing with your zip to get local weather on your feed (not an amazing use-case, but presents the problems to be solved).  This would need a secure way of holding your zip where it's not visible to others but moves with the user from client to client.  I was just going to provide the ability to encrypt everything in the kind 10001, but I think this would be much better.

Perhaps we can standardize the content of this schema in a new NIP.  (Frankly, NIP-23 should have been two separate NIPs anyway, which is why I won't fight too hard for it).

staab:
> we can then look at all our contacts on kind-3 and assign one relay from this relays list to our kind-3 contact relay field

That works, but I'm not completely satisfied with it. As special-purpose relays emerge, a user may want to follow a particular relay independent of any individual person who posts there.

eskema:
> As special-purpose relays emerge

I agree there, but I'd say that a special relay would also need special handling, so it would be handled however it needs to be handled anyway.. it's not like we're simply subscribing to everything in a relay, there's more into it than that so the relays could simply have a pubkey and you follow them like any other identity.. it's a total different case that we don't even know yet what it will be.. so it's not much possible to anticipate what that special handling might be..

ismyhc:
> if we're going this route, then IMO this array should maybe be better in kind-0 as `content.relays:[]`. having what relays you post to is pretty integral part of identity, and it would reduce a lot of guessing and events back and fort since kind-0 is pretty much always needed... we already have a bunch of frivolous fields there, it's easily updateable and probably the most up to date everywhere. a simple list of relays there makes more sense to me in terms of practicality. if that doesn't make sense, and everyone prefers this model, then let's use this kind, but I think having the relay information readily available in kind:0 solves multiple problems at once.
> 
> or add the `r` tags to the kind-0 event tags directly

I totally agree with this point. As I was working out thinking about this, my first thought is why do we not have a relays array on kind-0. This would simplify client side handling so much. As of right now I think @jb55 is stuffing relays in the content of kind-0. 

Im am 100% for adding a relays array to kind-0. This makes the most sense and Im surprised it wasn't there from the beginning. I should be able to get all I need about a pub key holder via kind-0.

barkyq:
Just my 2 cents.

Could use one parameterized replaceable event for each relay (the "d" tag would be the relay URL). Say kind 30001

Seems better for concurrent use by multiple clients. Each time any of my clients successfully establishes a websocket connection to a relay, the client publishes a parameterized replaceable event. This won't overwrite the other events (only the ones corresponding to the same relay). So for instance, I could request kind 30001 authored by "Mike Dilger" and filter by limit:5 to receive the 5 most recent relays successfully used by any of Mike Dilger's clients.

monlovesmango:
@mikedilger I love this idea, thanks for putting this nip together! and agree with everything @eskema has said here so far. and agree with @fiatjaf suggestion to use this rather than nip23.

I've been contemplating nostr relay strategy based on some of fiatjaf's recent comments. In my mind it seems like there are 2 basic strategies regarding users defining their relays. either users are
a) encouraged to define which relays they want read from
b) encouraged to define which relays they want to write to

option a is what has mostly been used so far, but over time it feels like it will slowly require more of option b to continue operating smoothly which can get messy. having a mix of the 2 options would also be messy. whereas option b would never need to mix in some of option a to work, which is why I like this model more (as long as relay advertisement is widely broadcasted). option b is also MUCH more scalable for relays and prevents unnecessary duplication of data across relays.

if we are going to lean towards option b, it seems unnecessary to require users to publicly define which relays they want to read from at all. optionally, if users did want the ability to do this, it would be really easy to add that, either as nip23 relay list (though nip23 supports/encourages option a as well which I think its unnecessary now) or as nip51 generic list (which could be public or private).

I would also really prefer to have this info on the kind 0 record as @eskema had suggested. and then have your kind 0 record spread widely across the network. I also like the idea of adding relays in the kind 0 tags rather than in the content.

should there be some limit to the number of relays that can be listed should be suggested? and/or some verbiage to say relays are ordered by priority? this way if a client only wants to connect to one relay they know which is the most important and also discourages people listing their whole relay list here.

fiatjaf:
I like having relays as tags. I think it unlocks some nice features like being able to query who is in each relay, but I don't know if that is worth the trouble.

I also think there should be some recommended limits on the number of relays, and honestly relays should be blocking events with too many tags already, so there is that.

Would be nice to know how Damus is doing this, so we can reuse that standard.

ismyhc:
> I guess we can all agree on using kind 0 then.
> 
> I like having relays as tags. I think it unlocks some nice features like being able to query who is in each relay, but I don't know if that is worth the trouble.
> 
> I also think there should be some recommended limits on the number of relays, and honestly relays should be blocking events with too many tags already, so there is that.
> 
> However, if Damus is not using the tags, but just using an array of strings inside the metadata body then we probably should do that. Would be nice to know what Damus is really doing.

Im fine with it being in tags as long as it lives on kind-0. :)

monlovesmango:
I only found one profile with "relay" in metadata, however this kind 0 is really nonstandard, has a content key within the content field. note1q7w36puvtxukgflu2kvugyral4tvk938lj9mmq7djvxugftazdlq2sf65g 

@ismyhc are you sure Damus is adding relay in kind 0 content? I would have expected to find many of these events if so

ismyhc:
> I only found one profile with "relay" in metadata, however this kind 0 is really nonstandard, has a content key within the content field. note1q7w36puvtxukgflu2kvugyral4tvk938lj9mmq7djvxugftazdlq2sf65g
> 
> @ismyhc are you sure Damus is adding relay in kind 0 content? I would have expected to find many of these events if so

Oh shoot. I just found my conversation with @jb55 and he puts them in kind3. In the content I believe. My bad!

I still believe that relays should be on kind-0, somewhere.

fiatjaf:
> I just found my conversation with @jb55 and he puts them in kind3

That's what most clients are doing, it's a crappy legacy from branle.

In this case I think we should go with `"r"` tags on kind 0. What are the arguments against that? Tags are also good because we can add more stuff to them after the first two elements if necessary.

staab:
Main downside would be from @barkyq 


> Could use one parameterized replaceable event for each relay (the "d" tag would be the relay URL). Say kind 30001
> 
> Seems better for concurrent use by multiple clients.

I agree with this, but I'm not sure how important concurrent use is. It has been demonstrated to be a problem thus far however, and the more we pack into a single kind the more severe it will be.

ismyhc:
> > I just found my conversation with @jb55 and he puts them in kind3
> 
> That's what most clients are doing, it's a crappy legacy from branle.
> 
> In this case I think we should go with `"r"` tags on kind 0. What are the arguments against that? Tags are also good because we can add more stuff to them after the first two elements if necessary.

%100 in agreement

mikedilger:
Just woke up and read all your comments.  I'm glad the need for this is understood.

I think nip23 still has a useful purpose for controlling your own clients as you move between them.

I'm fine with using kind-0.  I prefer using the "r" tags for the reason that metadata currently is all strings, not lists of strings, so this won't add any new complexity for anyone. It doesn't hurt to be queryable but I'm not sure why they would be queried.  If anyone has a good argument for a different non-queryable tag I'm listening.

What do people think about changing the name to "Relay Advertisement" or something other than "Feed Advertisement" as suggested by @monlovesmango ?  Since it's not a new type, I could retitle the NIP to "Relays in Metadata" or something.

Should we add a priority to the relay tags or other markers?

Can someone suggest a recommended limit?

eskema:
> I think nip23 still has a useful purpose for controlling your own clients as you move between them.
> 
agreed, but the need for a simple system is urgent, so let's do this :)

> I'm fine with using kind-0. I prefer using the "r" tags for the reason that metadata currently is all strings, not lists of strings, so this won't add any new complexity for anyone. It doesn't hurt to be queryable but I'm not sure why they would be queried. If anyone has a good argument for a different non-queryable tag I'm listening.
> 
queryable is good I think, the only doubt I had was about the letter, since we already have it in some fringe use as a url reference (I guess a relay would still qualify as a url), even though I think maybe a dedicated relay tag that is not some other type of url may be handy for all types of things, so we either reserve `r` for relays or another one, but `r` kind of makes sense for relays, so maybe we can come up with another for the `url` ones, it's not much used anyway..  

> What do people think about changing the name to "Relay Advertisement" or something other than "Feed Advertisement" as suggested by @monlovesmango ? Since it's not a new type, I could retitle the NIP to "Relays in Metadata" or something.
> 
Relays in metadata makes sense, but if we go for tags, is it really in the metadata since the metadata is supposedly only in content? this may be an irrelevant point that won't matter much and doesn't warran't loosing time discussing it, choose whatever you find appropriate.

> Should we add a priority to the relay tags or other markers?
> 
I think we should be treating those as all valid, it's a list because it gives some choice for the client to optimize, we could use the order but we already saw the need for proper markers with `root` and `reply`, maybe be can leave this open and let clients experiment a bit and maybe settle later on a specific approach that is tested

> Can someone suggest a recommended limit?
this will be a tough consensus, let's say less than 10? my sweet number would be between 3 and 6

mikedilger:
> queryable is good I think, the only doubt I had was about the letter, since we already have it in some fringe use as a url reference (I guess a relay would still qualify as a url), even though I think maybe a dedicated relay tag that is not some other type of url may be handy for all types of things, so we either reserve `r` for relays or another one, but `r` kind of makes sense for relays, so maybe we can come up with another for the `url` ones, it's not much used anyway..

I went looking for "what is a tag that means Url?" and stumbled across 'r'.

> >  Can someone suggest a recommended limit?
>
> this will be a tough consensus, let's say less than 10? my sweet number would be between 3 and 6

Does the recommendation mean users probably shouldn't write to more than that many? Or that relays should reject kind-0 with more than that many?  Might need two numbers.

eskema:
> Does the recommendation mean users probably shouldn't write to more than that many? Or that relays should reject kind-0 with more than that many? Might need two numbers.

I'm not sure it makes sense to explicitly limit, it's just a list to where the user at least writes to, we can still have other means to write to other relays as well that are not listed here, but at least it is expected that the user posts to those on the list, so we can pick any one from it (or any amount the client needs)

mikedilger:
I hesitate to push this argument because everybody came down on the other side of it and I don't want to derail this. But I think this is better as a new event kind outside of metadata for the following reasons: (1) the event is much smaller. If you spread it widely it has less impact as such, and (2) there is no content that would need moderation, so relays that require subscriptions might allow these without a subscription (metadata may need moderation).   AFAIUnderstand it, the benefits of being in metadata are that you encourage people to spread their metadata widely (not sure if that's important) to avoid a new event kind (not sure I follow) to group together similar data (I don't see the practical benefit, just a conceptual one).

fiatjaf:
> I hesitate to push this argument because everybody came down on the other side of it and I don't want to derail this. But I think this is better as a new event kind outside of metadata for the following reasons: (1) the event is much smaller. If you spread it widely it has less impact as such, and (2) there is no content that would need moderation, so relays that require subscriptions might allow these without a subscription (metadata may need moderation). AFAIUnderstand it, the benefits of being in metadata are that you encourage people to spread their metadata widely (not sure if that's important) to avoid a new event kind (not sure I follow) to group together similar data (I don't see the practical benefit, just a conceptual one).

Good points. I'm switching my allegiance to this.

mikedilger:
I'm open to suggestions for a better name. This term "feed" isn't great. "Relay Advertisements" is too easily confused with a relay operator seeking customers. If it isn't a new type I'd just title it "Becoming Findable by Advertising where you Post" or something, But a new event kind needs a shorter concise meaning full name which is eluding me.

mikedilger:
If I want to create a new post and tag someone who doesn't follow me, where can I post it that such that the tagged person will see it?  If we get rid of 'read' relays, I have no way to contact them. While I still think this PR is the way to go to find people, we shouldn't dispense with read relays so cavalierly, we need to think it through a bit more. Any thoughts on how to handle this?

BrightonBTC:
Sorry, coming to this a bit late, just reading through all the comments now. I was falling on the side of kind 0 with a list of r tags but those last points form @mikedilger are pretty strong, especially:

> (2)  there is no content that would need moderation, so relays that require subscriptions might allow these without a subscription (metadata may need moderation)

WRT to naming, maybe "relay prompt"?

mikedilger:
> Could use one parameterized replaceable event for each relay (the "d" tag would be the relay URL). Say kind 30001
> 
> Seems better for concurrent use by multiple clients. Each time any of my clients successfully establishes a websocket connection to a relay, the client publishes a parameterized replaceable event. This won't overwrite the other events (only the ones corresponding to the same relay). So for instance, I could request kind 30001 authored by "Mike Dilger" and filter by limit:5 to receive the 5 most recent relays successfully used by any of Mike Dilger's clients.

I think this presumes such parameterized replaceable events are easily findable.  But to find them, clients would need to publish them far and wide AND often.  So I'm leaning against this one.

mikedilger:
I would like to hear more arguments from people who prefer this to be in kind-0 metadata, if you still do. Many people asked for that, I don't want to just do it 'my way'. If there are good reasons that I'm unaware of I want to hear them.

ismyhc:
My reasoning for it being kind-0 was simply because it feels like the logical place for it to be. Im already getting kind-0 data and I would expect the users relay choices to be present there.

Thats really the best reasoning I have. Im not totally married to it. Ultimately as long as there is a sane way to know what relays a user posting it is enough for me.

monlovesmango:
yeah same. I agree with the technical reasoning on why new event kind is better. putting it in kind 0 just seemed better from a practical standpoint. the specific point that @BrightonBTC highlighted convinced me that the practicality isn't worth it.

I am good with using a new kind

monlovesmango:
> If I want to create a new post and tag someone who doesn't follow me, where can I post it that such that the tagged person will see it? If we get rid of 'read' relays, I have no way to contact them. While I still think this PR is the way to go to find people, we shouldn't dispense with read relays so cavalierly, we need to think it through a bit more. Any thoughts on how to handle this?

one idea I had was adding one or more "r" tags to the note for responses to be sent too. so when you write a note and add a "r" tag your saying, 'hey I am checking for responses at this relay'. if you don't want randos to be able to reach you, you don't have to add an "r" tag.

fiatjaf:
> one idea I had was adding one or more "r" tags to the note for responses to be sent too. so when you write a note and add a "r" tag your saying, 'hey I am checking for responses at this relay'. if you don't want randos to be able to reach you, you don't have to add an "r" tag.

This is a great idea.

fiatjaf:
I think we should all be ok to merge this now? Any objections?

staab:
I think the key tension here is that while kind 0 is the logical place to put relays, it invites race conditions. But putting relays in a new kind doesn't solve the core problem, which is that kind 0 is replaceable. I'm starting to think that only individual values should be used with replaceable types, and everything else should be merged on demand.

So for example, if I update my profile image on my mobile client and my relays list on desktop without first waiting for them to sync, my profile image will be reverted. Instead, if kind 0 was published as a sparse dictionary (PATCH rather than PUT semantics), clients could safely merge kind 0 in created order to get the full picture.

The downside would be all events would need to be kept around, and clients might get a partial picture. A possible solution to this would be an event-sourcey sort of thing where relays merge kind 0 PATCHes into a full result.

Now I'm just rambling, but maybe kind 0 should be split into as many different kinds as there currently are keys. This would solve the race condition issue by allowing for granular updates.

With all that said, I do think this NIP is an improvement.

fiatjaf:
I feel like we all would like to do the PATCH solution in many places, but to me it is an anti-pattern on Nostr. Events should not crucially depend on others, they must strive to be self-contained, because there is never a guarantee that the other events will be around -- as you said.

> Now I'm just rambling, but maybe kind 0 should be split into as many different kinds as there currently are keys. This would solve the race condition issue by allowing for granular updates.

Good point, yes, that would be an interesting solution to explore, although I don't think we should try to do that as of now. The takeaway is that we shouldn't make things worse by putting more stuff on kind 0 for now.

mikedilger:
> So for example, if I update my profile image on my mobile client and my relays list on desktop without first waiting for them to sync, my profile image will be reverted.

This might be solvable by putting a date stamp on each field.  You create a new clobbering metadata event, but only update the date stamp of the field you change.  Then clients can detect the clobber and keep the newest per-field.

We should talk about that on a new issue. This PR reminds us of the issue, but isn't part of kind-0 at this point.

eskema:
If everyone prefers a new kind, I won't fight it, but as a last point I'll make: a simple dumb relay list makes perfect sense in kind:0, and if we want people to use a small set of relays here, I feel kind:0 is the only event where people will care to have less information, so the list has a higher probability of beeing higher signal than relegating that to a dedicated list that will probably be obscured by most clients. I feel this list is akin to a `website:` field, as in, the relay you write is almost like your blog domain, and that usually is a concise information that we put obviously in kind:0 metadata...

ekimber:
What if a user changes their advertised relay list? Since it is a replaceable event, the client of a follower could be given an entirely new list and would then fail to find older events posted to the old relay list. Typically they would probably change say 1 out of 4 relays and it wouldn't matter but in some cases users could have an inconsistent experience. Incidentally this is why I don't see this data as a good fit with the metadata in kind 0, I feel it is qualitatively somewhat different.

@barkyq Could you explain your one event per relay idea a bit more? I think we are basically talking about the same issue, and maybe your suggestion would resolve it

fiatjaf:
When moving between relays there is always the possibility of losing history. If a user is worried about that they should somehow move all past events to the new relay too. If this is not possible so be it. Nothing is ever guaranteed.

Other things that may happen that could cause loss of events:
- a relay just dies and never returns
- the user is banned from a relay

eskema:
if users keep changing their relay list, of course it's going to be hard to follow them... that's why this list should be more static, so we know where to find stuff instead of always guessing.. having this on kind:0 or kind:30001 won't make a difference to that problem..

ekimber:
If the events are lost then so be it, nothing can be done, yet it seems wasteful to copy all events from relay X to relay Y instead of merely retaining a pointer to relay X. I don't really follow how it is static. A user's previous write relay is potentially useful information whereas a previous profile pic is irrelevant

eskema:
> If the events are lost then so be it, nothing can be done, yet it seems wasteful to copy all events from relay X to relay Y instead of merely retaining a pointer to relay X. I don't really follow how it is static. A user's previous write relay is potentially useful information whereas a previous profile pic is irrelevant

if a user decides to not refer to a relay anymore, we would only care to know past relay info for past events, not new ones. that's a different aspect than knowing where a user is writing at any given moment, than it is where as user as ever written to.

ekimber:
Aren't both past and future aspects relevant to this NIP? Users have expectations, such as being able to view the complete past timeline of posts by another user, such as not losing "likes" on their notes. Those expectations cannot always be met because of how Nostr works but arguably they are entitled to a best effort to meet them by relays and clients

eskema:
> Aren't both past and future aspects relevant to this NIP? Users have expectations, such as being able to view the complete past timeline of posts by another user, such as not losing "likes" on their notes. Those expectations cannot always be met because of how Nostr works but arguably they are entitled to a best effort to meet them by relays and clients

well I'd say that expectation should be managed by the user, so if that user wants it's content accessible, it should make it accesible and not make clients jump through hoops to find it. If users make it difficult for clients, then clients will ignore users and make their own rules

fiatjaf:
> A user's previous write relay is potentially useful information

Then the user can just not remove the relay from their relaylist. They can have the relay in there and just not publish to it anymore. It's just a decision on the part of the client.

Also some relays may decide to not replace the relays list from users, the "replaceable" bit is just a suggestion, relays can keep copies of past relaylists forever. Again, just decisions that should happen outside of the protocol.

Cameri (approved):
LGTM

mikedilger:
> What if a user changes their advertised relay list? Since it is a replaceable event, the client of a follower could be given an entirely new list and would then fail to find older events posted to the old relay list.

This is an interesting point, and fiatjaf's comment I think is a good solution:

> Then the user can just not remove the relay from their relaylist.

While that doesn't change this NIP, I think I'll have to change my client somehow to allow you to still advertise relays you no longer write to.

sondreb:
So is the idea that this will replace the non-standard usage of `content` on kind 3?

This basically means that at minimum, when opening a users profile, at least 3 events must be downloaded:

kind 0 (latest profile), kind 3 (latest following) and kind 10002 (relays).

It's likely preferable to have relays in separate event than following lists though, so they are separate updated lists and less chance of loosing content from either when using multiple different clients.

mikedilger:
> So is the idea that this will replace the non-standard usage of content on kind 3?

Yes. But it also clarifies what 'read' and 'write' mean, specifically that read does not mean you read all events from that relay, but rather it is the place you pick up events addressed to you.... and you read events you follow wherever they are written (the 'write' relays of whomever you follow).

> This basically means that at minimum, when opening a users profile, at least 3 events must be downloaded:

Yes I suppose.

But I don't see kind-10002 data's primary purpose being to show it on somebody's profile page. You could show it there, yes, but you probably will already have a cached copy if you are using it as intended. At least gossip caches kind-10002 data and only re-downloads periodically. That is because gossip refers to this data quite often, for subscribing to the feed, for adjusting those subscriptions as relays drop, and anytime you write a post mentioning someone (such posts are sent to your write relays, and also to the read relays of the people you mention).

barkyq:
@ekimber my idea was to use parameterized replaceable events with "d" tag equal to the relay url. This seems like a good way for multiple clients used by X to concurrently update which relays were most recently used, thereby making it easy for other users to "track" user X. By the nature of parameterized replaceable events, one could REQ the chosen kind, filtering by author: X and limit: N to get the N most recently used/updated relays (which will be stored in the `d` tags of the returned events). 

It does not seem to be so good though for publishing a list of relays where your "feed" would be, as the user probably wants a bit more conscious control over that.

fiatjaf:
If no one objects I'll merge this today.

monlovesmango:
no objections, please merge.

I've been thinking more about @mikedilger 's question about how to let non-follows reach you, and while I still think my suggestion [above](https://github.com/nostr-protocol/nips/pull/218#issuecomment-1416811415) is a good start, I do think it would be nice to have something a little more robust. so I wanted to get thoughts on an event kind that would be similar but complementary and secondary to this one that mike is proposing.

the new event kind would be a notifications/reach me relay list (still struggling to come up with an appropriate name). clients should NOT treat it as a general 'read' relay list that a feed or other curated content should be pulled from. this relay list should only be used to pull events that you have been tagged in, like note mentions or dms but could be extended to other kinds that can tag users as well.

I am actually really excited about this idea for the following reasons, but would love feedback as I may be overlooking something:

1. between this nip's 'Relay List Metadata' to store your write relays, the new notification relay list event I am proposing, general relay lists (which I plan to propose/add to [nip51 pr](https://github.com/nostr-protocol/nips/pull/183), and relay hints embedded in tags there should be no scenario that I can think of where a client doesn't know where to find a particular event.
2. data duplication across relays would be limited to only what the user has intentionally configured.
3. notification relays would add a level of spam filtering that client doesn't need to build and doesn't place additional burden on relays
    -   you could configure your 'Relay List Metadata' (write) relay to be a personal relay that no one else can write to and your notification relay to a public relay
    -   you could configure your 'Relay List Metadata' (write) relay to be a paid relay and your notification relay to a POW relay
    -   you could configure your 'Relay List Metadata' (write) relay to be a personal relay that no one else can write to and your notification relay to a different personal relay that only a select set of users can write to
4. ensures that people who don't want to be reached by randos never will be while also ensuring that those who do want to interact with randos can easily do so (and dictate the level of effort a rando needs to go through to reach you)

anyway apologies for the long rant. if I don't hear much back I will go ahead and open a pr anyway, but wanted vet this here first incase there are glaring errors in my logic. thanks for reading!

mikedilger:
@monlovesmango read the meaning of 'read' in this NIP again carefully.  It's not the old "read my feed from these", it means "pick up my mentions and DMs from people I don't follow from this inbox relay".  I think it is already precisely what you are suggesting.

Because the old "read" should really be read from other people's "write".

I am considering extending nostr-rs-relay to allow anybody to post, but only if I'm tagged, and nobody to be able to read those posts back again so I don't have to worry about moderation.. and then use that as my inbox. Actually maybe @scsibug would do it (please?).  And ideally if they could still read posts authored by me, then it could be my outbox too, my only relay.

mikedilger:
@monlovesmango also it's good to know it's been thought through independently and the same conclusions reached.

monlovesmango:
@mikedilger I am so sorry, I neglected to review your most recent updates! yes, this is exactly what I was trying to convey. and in fact, this is exactly the implementation I wanted (where both read and write relays are on the same record). you seem to consistently be one step ahead of me :). anyway I like this nip even more now and am very excited to try implementing it. thanks for putting this together and I'll be quite now.

scsibug:
> I am considering extending nostr-rs-relay to allow anybody to post, but only if I'm tagged, and nobody to be able to read those posts back again so I don't have to worry about moderation.. and then use that as my inbox. Actually maybe @scsibug would do it (please?). And ideally if they could still read posts authored by me, then it could be my outbox too, my only relay.

This will be possible with a gRPC interface I am building right now!

65.md
```diff
@@ -0,0 +1,70 @@
...
+Relay List Metadata
+-------------------
+
+`draft` `optional` `author:mikedilger`
+
+A special replaceable event meaning "Relay List Metadata" is defined as an event with kind `10002` having a list of `r` tags, one for each relay the author uses to either read or write to.
+
+The primary purpose of this relay list is to advertise to others, not for configuring one's client.
+
+The content is not used and SHOULD be blank.
```
mplorentz, line 13: I think it's more clear to say "an empty string" here rather than "blank".

mikedilger: Sure.

This is merged so I think you should do that change in another PR for quick merging (unless there is a way to do it from here I'm unaware of).

arthurfranca:
@mikedilger sorry for ressurecting this. But at the "The meaning of read and write" topic I may be missing directions regarding threads.

I mean, if note C is a reply to note B which is a reply to note A. Considering that C doesn't mention A's author nor anyone else besides B's author, if I got it right, the idea is to just publish C to C's author write relays and to one of B's author read relays. This is great for C's author followers when browsing feed (or at his profile) and to B's author to be notified.

But do you think, though, that A, B, C and all other replies at any level of the thread tree should additionally be posted to a common set of relays to help loading the thread?

Cause following this NIP, A replies and replies to its replies and so on are scattered across many relays, the more replies exist. This contrasts with the legacy way of choosing a fixed set of popular relays, which has its problems with centralization and scalability, but would probably be able to fetch the majority of the thread notes easier.

The NIP is great, I was just thinking if there is more to it that was left out of the text.

mikedilger:
> Considering that C doesn't mention A's author nor anyone else besides B's author...

I thought replies were supposed to include all the 'p' tags of the parent, meaning they get all the 'p' tags all the way up the chain. Could someone check (I have to dash).

But otherwise, yes thread tracking could be a problem. It might be that you should repost the entire thread (including notes you didn't author) to the read relays in question.  Just an idea off the cuff.

PR #219, opened 2023-02-03
NIP-11 pubkey prefix allowed

mikedilger:
I notice in my gossip logs a lot of invalid NIP-11s.  Digging in, it turns out people are putting significant prefixes of their pubkey, not the entire pubkey (possibly for plausable deniability?).

This PR changes NIP-11 to provide for that.

### DIFF
--- a/11.md
+++ b/11.md
@@ -14,7 +14,7 @@ When a relay receives an HTTP(s) request with an `Accept` header of `application
 {
   "name": <string identifying relay>,
   "description": <string with detailed information>,
-  "pubkey": <administrative contact pubkey>,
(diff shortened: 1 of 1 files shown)


mikedilger:
Actually I would much prefer if people didn't cut bytes in half, listing an odd number of hex digits because I can't hex decode them into bytes.  But it looks like that is what everybody is doing (many relays).  *sigh*

fiatjaf:
I don't think this should be supported at all. If someone is doing this it is better to ask them to fix it.

mikedilger:
ok I had better make sure it's not my bug first.

Issue #216, opened 2023-02-02
**DRAFT** KaTeX support


wires:
This is an request for comments on supporting TeX markup to Nostr.

**Motivation**

Mathematical notation is generally written using inlined single or double dollar delimited TeX expressions. For example:

> Relativistic energy is given by
> $$E_\mathrm{rel} = \sqrt{(m_0 c^2)^2 + (pc)^2}$$
> However using the *Lorentz factor* $\gamma$ this can be written as > $E={\gamma}mc^2$.

```
Relativistic energy is given by
$$E_\mathrm{rel} = \sqrt{(m_0 c^2)^2 + (pc)^2}$$
However using the *Lorentz factor* $\gamma$
this can be written as $E={\gamma}mc^2$.
```

KaTeX is a relatively safe and limited subset of LaTeX, see https://katex.org/docs/supported.html and https://katex.org/docs/security.html.

Below follow two proposals, one KISS and one modeled after NIP-08. Shoot!

---

### Simple KaTeX support

An event with a tag `"katex"` MUST contain KaTeX code within its `.content`.

The KaTeX code MUST be delimited by `$..$` for inline mode or `$$..$$` for display mode.

Supporting clients SHOULD replace KaTeX code (including the delimiters) with an inline or display block containing one of the following.

- fallback unicode text when appropriate (for example in terminals).
- live KaTeX DOM node
- vector/raster image resulting from a (sandboxed) KaTeX render

The tag SHOULD include the KaTeX which was used to author the message, for example `["katex","version","0.16.4"]`.

#### Persistent Macros

One or more `["katex","macro", KaTeX_macro]` tags MAY be provided,
whose `KaTeX_macro` code MUST be executed before all other TeX expressions
and it's `\gdef` commands MUST be persisted.
See https://katex.org/docs/api.html#persistent-macros

#### Metadata Prelude

We can also store a common collections of macros ("prelude") in our metadata.

If the metadata event contains a `.katex` property of the following form:

```
 "katex": {
    "version": "0.16.4",
    "prelude": {
      "v0": "\newcommand{\bla}{#1}"
    }
 }
```

Then one or more `["katex","prelude", name]` tags MAY be provided.
The client should then handle `.katex.prelude.name` as
a *per event persistent macro*.

Authors should take caution not to break old events.

---

### Alternative for KaTeX support (**discuss?**)

[...]

PR #220, opened 2023-02-03
NIP-23: long-form content

fiatjaf:
This one is much larger than I thought it would be. I lost my patience in the middle of it so the quality might have degraded.

This PR also modifies NIP-33 and NIP-19 to introduce a new canonical way of referring to parameterized replaceable events.

### DIFF
--- a/19.md
+++ b/19.md
@@ -35,6 +35,7 @@ These are the possible bech32 prefixes with `TLV`:
   - `nprofile`: a nostr profile
   - `nevent`: a nostr event
   - `nrelay`: a nostr relay
+  - `nref`: a nostr parameterized replaceable event coordinate (NIP-33)
 
 These possible standardized `TLV` types are indicated here:
 
@@ -42,11 +43,15 @@ These possible standardized `TLV` types are indicated here:
   - depends on the bech32 prefix:
     - for `nprofile` it will be the 32 bytes of the profile public key
     - for `nevent` it will be the 32 bytes of the event id
-    - for `nrelay`, this is the relay URL.
+    - for `nrelay`, this is the relay URL
+    - for `nref`, it is the identifier (the `"d"` tag) of the event being referenced
   - for `nprofile`, `nevent` and `nrelay` this may be included only once.
 - `1`: `relay`
-  - A relay in which the entity (profile or event) is more likely to be found, encoded as UTF-8. This may be included multiple times.
-  - not applicable to `nrelay`.
+  - for `nprofile`, `nevent` and `nref`, a relay in which the entity (profile or event) is more likely to be found, encoded as UTF-8. This may be included multiple times
+- `2`: `author`
+  - for `nref`, the 32 bytes of the pubkey of the event
+
+
 ## Examples
 
 - `npub180cvv07tjdrrgpa0j7j7tmnyl2yr6yr7l8j4s3evf6u64th6gkwsyjh6w6` should decode into the public key hex `3bf0c63fcb93463407af97a5e5ee64fa883d107ef9e558472c4eb9aaaefa459d` and vice-versa
--- a/19.md
+++ b/19.md
@@ -45,11 +45,13 @@ These possible standardized `TLV` types are indicated here:
     - for `nevent` it will be the 32 bytes of the event id
     - for `nrelay`, this is the relay URL
     - for `nitem`, it is the identifier (the `"d"` tag) of the event being referenced
-  - for `nprofile`, `nevent` and `nrelay` this may be included only once.
 - `1`: `relay`
-  - for `nprofile`, `nevent` and `nitem`, a relay in which the entity (profile or event) is more likely to be found, encoded as UTF-8. This may be included multiple times
+  - for `nprofile`, `nevent` and `nitem`, a relay in which the entity (profile or event) is more likely to be found, encoded as ascii
+  - this may be included multiple times
 - `2`: `author`
   - for `nitem`, the 32 bytes of the pubkey of the event
+- `3`: `kind`
+  - for `nitem`, the 32-bit unsigned integer of the kind, big-endian
 
 
 ## Examples
--- a/19.md
+++ b/19.md
@@ -35,7 +35,7 @@ These are the possible bech32 prefixes with `TLV`:
   - `nprofile`: a nostr profile
   - `nevent`: a nostr event
   - `nrelay`: a nostr relay
-  - `nitem`: a nostr parameterized replaceable event coordinate (NIP-33)
+  - `naddr`: a nostr parameterized replaceable event coordinate (NIP-33)
 
 These possible standardized `TLV` types are indicated here:
 
@@ -44,14 +44,14 @@ These possible standardized `TLV` types are indicated here:
     - for `nprofile` it will be the 32 bytes of the profile public key
     - for `nevent` it will be the 32 bytes of the event id
     - for `nrelay`, this is the relay URL
-    - for `nitem`, it is the identifier (the `"d"` tag) of the event being referenced
+    - for `naddr`, it is the identifier (the `"d"` tag) of the event being referenced
 - `1`: `relay`
-  - for `nprofile`, `nevent` and `nitem`, a relay in which the entity (profile or event) is more likely to be found, encoded as ascii
+  - for `nprofile`, `nevent` and `naddr`, a relay in which the entity (profile or event) is more likely to be found, encoded as ascii
   - this may be included multiple times
 - `2`: `author`
-  - for `nitem`, the 32 bytes of the pubkey of the event
+  - for `naddr`, the 32 bytes of the pubkey of the event
 - `3`: `kind`
-  - for `nitem`, the 32-bit unsigned integer of the kind, big-endian
+  - for `naddr`, the 32-bit unsigned integer of the kind, big-endian
 
 
 ## Examples
--- a/19.md
+++ b/19.md
@@ -35,7 +35,7 @@ These are the possible bech32 prefixes with `TLV`:
   - `nprofile`: a nostr profile
   - `nevent`: a nostr event
   - `nrelay`: a nostr relay
-  - `nref`: a nostr parameterized replaceable event coordinate (NIP-33)
+  - `nitem`: a nostr parameterized replaceable event coordinate (NIP-33)
 
 These possible standardized `TLV` types are indicated here:
 
@@ -44,12 +44,12 @@ These possible standardized `TLV` types are indicated here:
     - for `nprofile` it will be the 32 bytes of the profile public key
     - for `nevent` it will be the 32 bytes of the event id
     - for `nrelay`, this is the relay URL
-    - for `nref`, it is the identifier (the `"d"` tag) of the event being referenced
+    - for `nitem`, it is the identifier (the `"d"` tag) of the event being referenced
   - for `nprofile`, `nevent` and `nrelay` this may be included only once.
 - `1`: `relay`
-  - for `nprofile`, `nevent` and `nref`, a relay in which the entity (profile or event) is more likely to be found, encoded as UTF-8. This may be included multiple times
+  - for `nprofile`, `nevent` and `nitem`, a relay in which the entity (profile or event) is more likely to be found, encoded as UTF-8. This may be included multiple times
 - `2`: `author`
-  - for `nref`, the 32 bytes of the pubkey of the event
+  - for `nitem`, the 32 bytes of the pubkey of the event
 
 
 ## Examples
--- a/23.md
+++ b/23.md
@@ -35,7 +35,7 @@ The article may be linked to using the NIP-19 `nitem` code along with the `"i"` 
 
 ### References
 
-Writing clients should implement support for parsing pasted NIP-19 `nitem` identifiers and adding them automatically to the list of `.tags` of the event, replacing the actual content with a string like `#[tag_index]` in the same way as NIP-08 -- or, if the reference is in the form of a URL (for example, `[click here](nitem1...)`) then they should be replaced with just the tag number directly as if link with that name existed at the bottom of the Markdown (for example, `[click here][0]`).
+Clients that support publishing NIP-23 events should implement support for parsing pasted NIP-19 `nitem` identifiers and adding them automatically to the list of `.tags` of the event, replacing the actual content with a string like `#[tag_index]` in the same way as NIP-08 -- or, if the reference is in the form of a URL (for example, `[click here](nitem1...)`) then they should be replaced with just the tag number directly as if link with that name existed at the bottom of the Markdown (for example, `[click here][0]`).
 
 Reader clients should parse the Markdown and replace these references with either internal links so the referenced events can be accessed directly, with NIP-21 `nostr:nitem1...` links or direct links to web clients that will handle these references.
 
--- a/23.md
+++ b/23.md
@@ -1,0 +1,75 @@
+NIP-23
+======
+
+Long-form Content
+-----------------
+
+`draft` `optional` `author:fiatjaf`
+
+This NIP defines `kind:30023` (a parameterized replaceable event according to NIP-33) for long-form text content, generally referred to as "articles" or "blog posts".
+
+"Social" clients that deal primarily with `kind:1` notes should not be expected to implement this NIP.
+
+### Format
+
+The `.content` of these events should be a string text in Markdown syntax, including YAML front-matter for other metadata that may be necessary.
+
+### Metadata
+
+This NIP defines only `title` as the metadata. For the date of publication the event `.created_at` field should be used, and for "tags"/"hashtags" (i.e. topics about which the event might be of relevance) the `"t"` event tag should be used, as in NIP-12.
+
+### Editability
+
+These articles are meant to be editable, so they should make use of the replaceability feature of NIP-33 and include a `"d"` tag with an identifier for the article. Clients should take care to only publish and read these events from relays that implement that. If they don't do that they should also take care to hide old versions of the same article they may receive.
+
+### Linking
+
+The article may be linked to using the NIP-19 `nref` code, which should contain both the public key of the author and the article identifier (the `"d"` tag) and some relays. With that information it is possible to locate the article.
+
+### References
+
+Writing clients should implement support for parsing pasted NIP-19 `nref` identifiers and adding them automatically to the list of `.tags` of the event, replacing the actual content with a string like `#[tag_index]` in the same way as NIP-08 -- or, if the reference is in the form of a URL (for example, `[click here](nref...)`) then they should be replaced with just the tag number directly (for example, `[click here][0]`).
+
+Reader clients should parse the Markdown and replace these references with either internal links so the referenced events can be accessed directly, with NIP-21 `nostr:` links or direct links to web clients that will handle these references.
+
+The idea here is that having these tags is that reader clients can display a list of backreferences at the bottom when one article mentions another.
+
+### `kind:1` interface
+
+In case there is the desire for users or clients to share the written article on their "social" clients, a `kind:1` event must be used.
+
+Since "social" clients aren't expected to understand this NIP, these notes should contain a URL like `nostr:nref1...` specifying the `nref1` code for the article, such that it can be rendered clickable and other users can use a different client than their "social" to read the article -- and also if their client handles both kinds of posts, their client may recognize the URL and handle it internally.
+
+For clients that want to implement the ability for users to comment on these articles. `kind:1` notes should be used as the comments. They must be made in reply to another `kind:1` as above, such that other people on "social" clients can see the comments and get the context from the note above and load the article being commented on if they so desire.
+
+## Example of an event
+
+Article text:
+
+```markdown
+title: Lorem Ipsum
+---
+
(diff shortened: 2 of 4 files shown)


cmdruid:
Looks great. Is this the place to discuss both nips, or are there dedicated discussions for each NIP?

fiatjaf:
I've created this `nref` thing on NIP-19, but I dislike the name and I don't know if it is a good idea. Someone please fix that.

Also I am not sure of how one of these articles can reference another using the `nref` stuff. That part is broken in the NIP because I got confused and distracted. Must be fixed.

cmdruid:
> These articles are meant to be editable, so they should make use of the replace-ability feature of NIP-33 and include a "d" tag with an identifier for the article.

Should this be optional? Assuming relays honor the replaceable part, I don't see a simple way to track an edit history, which is a nice feature to have for a blog, especially with news articles.

> This NIP defines only title as the metadata.

Have you thought about using the "subject" field instead since it already exists in the spec?

I think the idea of standardizing a markdown-aware note for long-form content is fantastic, and supporting YAML front-matter is even more fantastic.

One obvious addition for meta-data I can see is to have a payment address for the article. This could default to an lud16 if one exists for the author in their profile, but having custom payment links per article helps with tracking engagement.

cmdruid:
> Also I am not sure of how one of these articles can reference another using the `nref` stuff. That part is broken in the NIP because I got confused and distracted. Must be fixed.

The NIP seems to make sense to me. Each nref is parsed out of the content field, and replaced with an tag reference plus a positional reference in the document. 

I suppose you can't use 'e' since the current document can change, and 'd' is used for pinning events to a static identifier. It seems like there needs to be a tag similar to 'e', but for referencing events identified by a 'd' tag correct?

23.md
```diff
@@ -0,0 +1,75 @@
...
+
+The idea here is that having these tags is that reader clients can display a list of backreferences at the bottom when one article mentions another.
+
+### `kind:1` interface
+
+In case there is the desire for users or clients to share the written article on their "social" clients, a `kind:1` event must be used.
+
+Since "social" clients aren't expected to understand this NIP, these notes should contain a URL like `nostr:nref1...` specifying the `nref1` code for the article, such that it can be rendered clickable and other users can use a different client than their "social" to read the article -- and also if their client handles both kinds of posts, their client may recognize the URL and handle it internally.
+
+For clients that want to implement the ability for users to comment on these articles. `kind:1` notes should be used as the comments. They must be made in reply to another `kind:1` as above, such that other people on "social" clients can see the comments and get the context from the note above and load the article being commented on if they so desire.
```
mplorentz, line 43: > They must be made in reply to another `kind:1` as above

I'm curious why replies wouldn't just reference the original `kind:30023`. Could you explain that further? It seems a little clunky to have to publish two events of different kinds to allow comments on your blog post.

Personally I think it's kind of powerful to let a `kind:1` note link to _any other event regardless of kind_ in its "e" tag. We have seen some interesting features emerge from this in scuttlebutt clients. One example is that some clients started rendering `block` messages from other users in the main feed, which gave people a chance to have a discussion about why the block was issued, which resulted in more people blocking spammers and other bad actors more quickly. Because the replies to a block took the same form as replies to a normal post it was pretty easy to write code to display them.

fiatjaf: I think I was just extrapolating from my experience with https://github.com/fiatjaf/nocomment, in which initially website comments would just reference a URL, but that was broken because the social clients didn't know they were supposed to look for a URL in the tags and render it, so the solution was creating a "bridge" event.

In this case however this is not necessary at all, so this was just dumb and I will just rewrite this part and make it reference the event directly.

fiatjaf: A problem remains, though: what event? If the content of the referenced article is changing we must come up with a way to reference that changeable thing and not just a single event (although a single event can also be referenced).

How about `["f", "pubkey:identifier"]`? I don't know, I'm not thinking straight.

eskema:
I still need some time to process this NIP, my brain doesn't get it yet... i find all this nref things a bit confusing and anti pattern for nostr. I think the document should be able to referenced by `note1d` and it's `hex protocol id`, and we should be able to see the edits, so a list of versions.. so I think we need at least two files for this to work. having only one creates a lot of undesirable trade-offs that ends up beeing worse. (or I simply need to process this NIP better..)
anyway.. here's what I'm thinking:

n1 =  kind 11
n2 = kind 30011 with `[d, n1.id]`, we can use the contents here for the event metadata, title, description, whatever and every time we want to edit the text, what happens is we create a new copy of the n1 and add this new id to the tags array of n2.

when sharing the event, you share the original id of n1, the client when it fetches it, sees it's a kind 11 so it searches for it's corresponding 30011 file. there we can grab the last [e] tag as the last version. if the array is empty then we have the only and last version (i.e. no edits). this lets us see what was actually written before edits (so we have accountability) instead of "permanently" deleting it. Also, we can keep updating the "metadata" of the kind 30011 content without altering the event itself

wires:
Very nice, this complements [my KaTeX proposal](https://github.com/nostr-protocol/nips/issues/216) nicely in that I've been thinking about this in a long form "plain text" format, ie. markdown or this NIP.

I wrote some comments (in my own markdown editor :-P), maybe these are helpful.

> [Section Format] ... Markdown syntax

Which specific version of Markdown? There are various flavors.

I suggest to learn from past experience and decide "once and for all" on a good Markdown-variant. To decide which one is subtle, I could write to write down an argumentation and suggestion... (based on my experiences from working heavily with md, in my daily job I we build a markdown based literate programming language and editor).

> ... YAML front-matter

We used YAML initially but its complex and has unwanted conveniences such as `"yes" === true`. Would suggest TOML for the "event format" instead and if clients insist on having YAML as input language they could convert TOML but present YAML to the user.

Would also propose to extract the TOML frontmatter into a separate property.
`.frontmatter` or `.metadata` for instance.

This avoids having an extra parsing step to seperate these two languages (YAML/MD or TOML/MD) and in turn avoids complex parsers and highlighters that must support "multi modal" editing, because now you can use a separate editor for each string.

Again if a client insists in having frontmatter in the same editor, they could concatenate both into one string and split again when building the event.

> [Section References] ... (for example `[click here][0]`

Why not use existing markdown / NIP-08 behaviour for this, as in `[click here](#[0])`. This would require no special support for detecting when the user is entering a markdown link in the client, I can just use existing autocomplete for `n*` refs/mentions and replace with `#[..]`. As opposed to having a special case for article links.

Or is the goal to have *article hyperlinking* explicitly distinguishable from regular markdown hyperlinks that *happen* to refer to `nref`s?

In that case I would suggest considering support for deeper linking, such as
- referencing sub headings (`####` = `eid§1.1.1.3`),
- or even specific formulas (see [katex proposal#216](https://github.com/nostr-protocol/nips/issues/216), section [Alternative for KaTeX support (discuss?)] aludes at this, using a NIP-08 style referencing, but would work equally well with other proposal, simply enumerating each `$..$` expression)
- or images/figures,
- and maybe even (character) ranges. (think: medium highlights, random quotes)

> Last section, example
> `"title":`

shouldn't this be `"title"` and `"content"` ?

wires:
As for the front matter, I would suggesting putting "special" / "semantically meaningful" fields under a uncommon subheading with a name such as `"$$meta"` or `NOSTR`.

In TOML:

afbeelding

This way the user is free to pick their own use for this front matter and this is very powerful (your markdown editor is now a flexible database, each document being a row).

Going further there are really two use cases for this "front matter" business:

- Allow the user to enter stuff like title, layout, enable extension, ISBN, etc.. ; the *semantically meaningful*.
- Allow the user to attach their own data to their notes (notion, airtable style)

We never got around to migrating away from multi modal editing, but we were planning to use something like this:

```typescript
type Metadata = { title: string, theme: string, pubData: string, /* URLs, DOI, Arxiv, ... */ }
type LongForm = {
   userdata: any,
   metadata: Metadata,
   markdown: string
}
```

This has the advantage of simplifying the parsing, typing, etc. and being clearing about what is each type of data for. The "front matter" business is after all just a convenient way of entering that data and I rather have the client do the parsing that this being essential part of the protocol in order to just show the title, date.

23.md
wires: Can't you force the client to make sure `"d"` is unique for a pubkey and then reference `["r",:pk,:d]`, finding all d-tagged events from pk, then link to latest seen? or am I missing something?

wires:
Additional comment on the front-matter semantically meaning/not distinction and separating these out into their own properties.

Imagine tagging a document, this can be done in YAML:

```yaml
tags:
- flep
- koek
```

or TOML

```toml
tags = ["flep","koek"
```
or even

```toml
[tags]
flep=123
lang="nl"
```

(and assuming these now mean something to nostr clients that understand tags)

however, this isn't the nicest UX and so you end up wanting to build a specialized tagging element with autocomplete, but now you need to synchronized the front matter text with the data coming from that UX element state. This kinda sucks because in order to preserve white-space while en/decoding to/from text you need to use a special parser and these are rare.

So having a "complicated" language like "YAMLFrontMatterMarkdown" is just not as nice and much more complex than just having an object sub-property that you can pass to an UI element.

wires:
> I've created this `nref` thing on NIP-19, but I dislike the name and I don't know if it is a good idea. Someone please fix that.

Another long comment on this, sorry for the spam, hope this is useful somehow. I tried to implement persistent/strongly typed referencing/querying of arbitrary (sub) content on our (nostr/matrix like) system and I came to the following insights.

Initially linking is easy because there are only 1 or two types of events in the system. But since we allow extensions it quickly becomes opaque to 'contents' of unknown kinds of events. In our case we deal with different nested/complex structures but we can provide a 'content tree structure' for most of them and map nodes in that tree to character ranges or other "lenses" into their "source of truth data" structure.

This way the client can still browse references and links even if it doesn't understand the content. I will illustrate below at `npath` section what I mean.

Because I suspect something similar is going on here:

`nref` itself is Not A Bad Idea, altho a bit "specific", as in
- they "only" are *coordinates* to query for *Parameterized Replaceable Events* of any kind
- (but not generic "(deeper) references"),
 
so maybe its better to pick a more specific/technical name such as

- `npre`
- `nprec` (PRE coodinate), or
- `npreq` (PRE query)

?

On NIP-19 for `note` it doesn't state clearly if this requires the event to be of `kind:1`, I assume not? Initially I was thinking `note` is for `kind:1` and `nevent` is the "any kind" event coordinate and in that line of thinking:

The PRE coordinate is `(pubkey, d-value)` and the relays are "just" hints. 

Maybe instead of `nref`/`nprec`, its nice to have `nlong` or `nlog` instead, which would refer *only* to long/blog type notes (and maybe their content/tags). So *more* specific to long form content, not more general like `npath`, more like what I thought `note` meant. It would have the same spec as `nref`, with `d-value`+`pubkey`+ relay hints, but with added requirement of `kind:30023` events only?

However, I can imagine more and more needs for specialized references arising.

So Instead maybe an `npath` approach would be interesting, borrowing ideas from XPath/CSS references?

These paths are 'drill down expression' into content on the nostr network and the different dimensions available at each 'drill down step' depend on the event kind at the current "location". So you start at some coordinate (`nevent`, `nprofile`,...) and then like a CSS selector step to a next coordinate. Examples of stepping dimensions could be
- following tag-indices `#[0]`,
- diving into metadata properties `.name` or other event properties,
- filtering by typestamp `:npub > notes > 2022`
- filtering by tag (value), etc.

For instance for long form we require the client implementing the `npath` nip to parse the markdown and produce a table of contents. This would be similar to outputting a parser AST and with character ranges on each tree node. This is similar to how errors work in an IDE, you click on the error/reference and you find the exact lines/char-range etc that it refers to.

To keep this simple the 'table of contents' should be a tree of (nested/non-overlapping/valid..?) character ranges into the `.content` string, and an optional name (for each branch). Then a `npath` reference could be a "note coordinate" (`nprec/note/nlong`) followed by a list of numbers, indicating each branch of the content tree. That would be a `kind:30023` table of contents dimension. But a similar construction could be used for chapter annotations in lectures or indices into a mixtape, source maps, stuff like that (altho this may changes the type of the *character range* label on the tree nodes).

If you think of the "space" of *Nostr state*, the discrete "points" in this space are the event id's. Then there is a partitioning on `(pubkey, kind)` and with NIP-33 sub partition `d-values` is added. Referencing only a `d-value` would should not be defined as a valid coordinate (as you could have `d-value` collisions from different pubkeys/kind) and `d-value`s should always be seen as a sub partitioning of `pubkey/kind` space. So `(pubkey,kind,d-value)` is a much better coordinate.

It would require some puzzling but if we define the right dimensions this is would be both simple, powerful and generic way to reference any (custom) (sub) content on the nostr network.

> Also I am not sure of how one of these articles can reference another using the `nref` stuff. That part is broken in the NIP because I got confused and distracted. Must be fixed.

Would it not suffice to replace in the `.content` references to `npre`/`note`/`nlong`/`npath` by `#[..]`, and place a tag `["r", :refType, :hex]` where :refType is `nref`,`note`, ... and :hex the decoded bech32 content?

wires:
NB: I explained this `npath` not to say this NIP should include that, rather that it makes sense to think seriously about linking/referencing and take inspiration from semantic web (failures) in a different NIP maybe and for now focus on something long form specialized and maybe using the fact that markdown has ToC structure.

fiatjaf:
This GitHub thing is terrible for discussing. I think Nostr is mature enough for discussions to be done inside nested threads there.

dskvr:
> This GitHub thing is terrible for discussing. I think Nostr is mature enough for discussions to be done inside nested threads there.
Is there a canonical event ID for this discussion?

wires:
> Is there a canonical event ID for this discussion?

I scratched an itch and made nips.be/pr/220 to redirect to PRs (see nips.be for redirect URLs).

Which client do you recommend for such threaded discussion?

Right now NIPs in various states are spread across PRs, issues and the main branch.

Would be nice if there one canonical place where NIPs are kept and a CONTRIBUTING.md file that outlines the NIPs lifecycle. Maybe a table `(sequenceNr, eventId/nref, title, short slug, title)` (on github/nostr). I could parse that and generate redirects `/:slug` and `/:nr`. Right now `nips.be` redirects to github, but I wouldn't mind serving a static page that tries to pull the longform NIP from some relays and renders it. Could even try to do this with a CDN edge function to have SSR and good indexing by search engines. I can then add `nostr:nref` deep link so you can use your preferred client to comment/edit. Or link to various suggested clients.

23.md
fiatjaf: The reason is why only the second element in the tag is indexed by relays, so we need the second element to be `":pk:d"` there for a client to display "backlinks" from other events.

fiatjaf:
Let's try to have this discussion over at [note1t7ekflxtvg7hsfh7f9zm6e6eykv3e0qyw690spw4rfzm0dw5e29qw54p32](nostr:note1t7ekflxtvg7hsfh7f9zm6e6eykv3e0qyw690spw4rfzm0dw5e29qw54p32).

fiatjaf:
> Which client do you recommend for such threaded discussion?

I recommend https://github.com/mikedilger/gossip

fiatjaf:
Answered @eskema and @wires (partially) at note1t7ekflxtvg7hsfh7f9zm6e6eykv3e0qyw690spw4rfzm0dw5e29qw54p32.

19.md
```diff
@@ -35,18 +35,23 @@ These are the possible bech32 prefixes with `TLV`:
...
 
 These possible standardized `TLV` types are indicated here:
 
 - `0`: `special`
   - depends on the bech32 prefix:
     - for `nprofile` it will be the 32 bytes of the profile public key
     - for `nevent` it will be the 32 bytes of the event id
-    - for `nrelay`, this is the relay URL.
+    - for `nrelay`, this is the relay URL
+    - for `nitem`, it is the identifier (the `"d"` tag) of the event being referenced
```
monlovesmango, line 47: I think kind is needed here no? or am I missing something? 

why not make 'nitem' applicable to general replaceable events as well? or is that bad form? first '0' type would be kind and second '0' type would be d-identifier.

fiatjaf: Yes, I think it is applicable, we just don't have a use case for that because generally the kind is implied by the purpose already so we only need the public key.

fiatjaf: Added this to the `nitem` encoding.

fiatjaf:
A prototype implementation here by @talvasconcelos: https://write.nostr.com/

19.md
```diff
@@ -35,18 +35,25 @@ These are the possible bech32 prefixes with `TLV`:
   - `nprofile`: a nostr profile
   - `nevent`: a nostr event
   - `nrelay`: a nostr relay
+  - `nitem`: a nostr parameterized replaceable event coordinate (NIP-33)
```
mplorentz, line 38: `item` is such a generic word. I've been trying to come up with a better one to suggest, but I'm not sure there is one haha. I'm guessing you had the same thought.

If an event is a representation of some data at particular point in time, then the "d" tag in the replaceable event is the address of a place where we can find the latest data. `naddress` seems almost as generic as `nitem`. `nregister` (like a CPU register) means a similar thing but is a little less generic. 

Is the "d" from NIP-33 short for something? Maybe there is some inspiration there.

23.md
```diff
@@ -12,62 +12,51 @@ This NIP defines `kind:30023` (a parameterized replaceable event according to NI
...
 
 ### Metadata
 
-This NIP defines only `title` as the metadata. For the date of publication the event `.created_at` field should be used, and for "tags"/"hashtags" (i.e. topics about which the event might be of relevance) the `"t"` event tag should be used, as in NIP-12.
+For the date of the last update the `.created_at` field should be used, for "tags"/"hashtags" (i.e. topics about which the event might be of relevance) the `"t"` event tag should be used, as per NIP-12.
+
+Other metadata fields can be added as tags to the event as necessary. Here we standardize 4 that may be useful, although they remain strictly optional:
+
+- `"title"`, for the article title
+- `"image"`, for an image to be shown along with the title
```
mplorentz, line 24:
```suggestion
- `"image"`, for a URL pointing to an image to be shown along with the title
```

23.md
```diff
@@ -12,62 +12,51 @@ This NIP defines `kind:30023` (a parameterized replaceable event according to NI
...
+The article may be linked to using the NIP-19 `nitem` code along with the `"i"` tag (see NIP-33 and NIP-19).
 
 ### References
 
-Writing clients should implement support for parsing pasted NIP-19 `nref` identifiers and adding them automatically to the list of `.tags` of the event, replacing the actual content with a string like `#[tag_index]` in the same way as NIP-08 -- or, if the reference is in the form of a URL (for example, `[click here](nref...)`) then they should be replaced with just the tag number directly (for example, `[click here][0]`).
+Writing clients should implement support for parsing pasted NIP-19 `nitem` identifiers and adding them automatically to the list of `.tags` of the event, replacing the actual content with a string like `#[tag_index]` in the same way as NIP-08 -- or, if the reference is in the form of a URL (for example, `[click here](nitem1...)`) then they should be replaced with just the tag number directly as if link with that name existed at the bottom of the Markdown (for example, `[click here][0]`).
```
mplorentz, line 38:
```suggestion 
Supporting clients should parse pasted NIP-19 `nitem` identifiers and adding them automatically to the list of `.tags` of the event, replacing the actual content with a string like `#[tag_index]` in the same way as NIP-08 -- or, if the reference is in the form of a URL (for example, `[click here](nitem1...)`) then they should be replaced with just the tag number directly as if link with that name existed at the bottom of the Markdown (for example, `[click here][0]`).
```

19.md
fiatjaf: I don't know. @Cameri please help. Also please offer your opinion here.

23.md
mplorentz, line 38:
```suggestion
Clients that support publishing NIP-23 events should implement support for parsing pasted NIP-19 `nitem` identifiers and adding them automatically to the list of `.tags` of the event, replacing the actual content with a string like `#[tag_index]` in the same way as NIP-08 -- or, if the reference is in the form of a URL (for example, `[click here](nitem1...)`) then they should be replaced with just the tag number directly as if link with that name existed at the bottom of the Markdown (for example, `[click here][0]`).
```

mplorentz:
Ok so I made a bunch of comments on the source before reading the comments about moving conversation to Nostr 😳 will post future comments there.

19.md
Cameri: the `d` stands for deduplication identifier.
I think it should be `nreg` with the kind (30000 to 39999), tag index, and pubkey.

The big issue is that the link can be broken easily by someone changing the replaceable event so I'm almost inclined to include a checksum or partial hash of the event instead of index.

eskema: `nid`?

Cameri: > I agree "register" makes sense, but `nreg` is such an ugly word, probably `naddr` is better.

register sounds good to me!

23.md
```diff
@@ -0,0 +1,64 @@
...
+
+The same principles can be applied to `nevent1...`, `note1...`, `nprofile1...` or `npub1...`.
+
+## Example Event
+
+```json
+{
+  "kind": 30023,
+  "created_at": 1675642635,
+  "title": "Lorem [ipsum][4] dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.\n\nRead more at #[3].",
```
eskema, line 52: should be the `content` field, not `title`

19.md
fernandolguevara: `nref` ?

juestr (approved):

23.md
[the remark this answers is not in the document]
fiatjaf: Yes.

fiatjaf:
A prototype implementation of a reader client here by @talvasconcelos: read.nostr.com

19.md
dskvr: > nref

full circle :laughing:

PR #222, opened 2023-02-04
Add `content` restriction to NIP-26

pablof7z:
This PR adds a `content` condition, so delegates are limited to signing a specific `content` payload instead of being able to sign anything.

### DIFF
--- a/26.md
+++ b/26.md
@@ -43,6 +43,9 @@ The following fields and operators are supported in the above query string:
    -  *Operators*:
       -  `<${TIMESTAMP}` - delegatee may only sign events created ***before*** the specified timestamp
       -  `>${TIMESTAMP}` - delegatee may only sign events created ***after*** the specified timestamp
+3. `content`
+   -  *Operators*:
+      -  `=${CONTENT}` - delegatee may only sign events with this content
 
 In order to create a single condition, you must use a supported field and operator. Multiple conditions can be used in a single query string, including on the same field. Conditions must be combined with `&`.
 
--- a/26.md
+++ b/26.md
@@ -45,7 +45,7 @@ The following fields and operators are supported in the above query string:
       -  `>${TIMESTAMP}` - delegatee may only sign events created ***after*** the specified timestamp
 3. `content`
    -  *Operators*:
-      -  `=${CONTENT}` - delegatee may only sign events with this content in RFC 3986 URI encoding
+      -  `=${SHA256(CONTENT)}` - delegatee may only sign events with content that sha256s to the specified hash
 
 In order to create a single condition, you must use a supported field and operator. Multiple conditions can be used in a single query string, including on the same field. Conditions must be combined with `&`.
 
@@ -54,7 +54,7 @@ For example, the following condition strings are valid:
 - `kind=1&created_at<1675721813`
 - `kind=0&kind=1&created_at>1675721813`
 - `kind=1&created_at>1674777689&created_at<1675721813`
-- `kind=1&created_at>1674777689&content=Hello%20world`
+- `kind=1&content=315f5bdb76d078c43b8ac0064e4a0164612b1fce77c869345bfc94c75894edd3
 
 For the vast majority of use-cases, it is advisable that query strings should include a `created_at` ***after*** condition reflecting the current time, to prevent the delegatee from publishing historic notes on the delegator's behalf.
 
--- a/26.md
+++ b/26.md
@@ -45,7 +45,7 @@ The following fields and operators are supported in the above query string:
       -  `>${TIMESTAMP}` - delegatee may only sign events created ***after*** the specified timestamp
 3. `content`
    -  *Operators*:
-      -  `=${CONTENT}` - delegatee may only sign events with this content
+      -  `=${CONTENT}` - delegatee may only sign events with this content in RFC 3986 URI encoding
 
 In order to create a single condition, you must use a supported field and operator. Multiple conditions can be used in a single query string, including on the same field. Conditions must be combined with `&`.
 
@@ -54,6 +54,7 @@ For example, the following condition strings are valid:
 - `kind=1&created_at<1675721813`
 - `kind=0&kind=1&created_at>1675721813`
 - `kind=1&created_at>1674777689&created_at<1675721813`
+- `kind=1&created_at>1674777689&content=Hello%20world`
 
 For the vast majority of use-cases, it is advisable that query strings should include a `created_at` ***after*** condition reflecting the current time, to prevent the delegatee from publishing historic notes on the delegator's behalf.
 


fiatjaf:
Looks good to me.

heelhook:
> Looks good to me.

Thoughts about a regexp support too?

```
content=/${CONTENT}/
```

fiatjaf:
No, that would be crazy.

heelhook:
> No, that would be crazy.

lol, thought so

fiatjaf:
@scsibug @jb55 @markharding @zackwy

bhayward93:
I'm not opposed at all,  but would it make more sense to use a hash of the content, or is there a reason not to? Just feels like it could become a pretty long string in some cases.

scsibug:
I am less enamored by this NIP everytime I think about it, and generally prefer the idea of some other higher-level way to link accounts (example, metadata in a user profile).  I already disabled this nip in my relay (for selfish performance reasons).

I do not understand a delegation condition for content description though, when that delegatee could just claim SFW and put whatever in the actual content.  Restricting kinds & created_at really limits the subaccount in meaningful ways, I do not think this field does so.

pablof7z:
> I'm not opposed at all, but would it make more sense to use a hash of the content, or is there a reason not to? Just feels like it could become a pretty long string in some cases.

A hash would make it impractical for users who might want to authorize the delegation. See this example on this [Alby PR](https://github.com/getAlby/lightning-browser-extension/pull/2063).

Screenshot 2023-02-06 at 1 18 32 AM

pablof7z:
> I do not understand a delegation condition for content description though, when that delegatee could just claim SFW and put whatever in the actual content. Restricting kinds & created_at really limits the subaccount in meaningful ways, I do not think this field does so.

Not sure I understand your "delegatee could just claim SFW and put whatever in the actual content", do you mean that the delegatee could just _promise_ not to abuse the delegated key? Given that the delegator cannot revoke it, I think the ability of scoping precisely what the delegatee can sign is very useful.

scsibug:
I am saying that, unlike the delegation restrictions that actually prevent a delegatee from posting kind X when only authorized for kind Y; there is no way for a relay to verify what the content tag says.  So, you have to trust the delegatee anyways, at which point, what is the point of the restriction?

pablof7z:
> I am saying that, unlike the delegation restrictions that actually prevent a delegatee from posting kind X when only authorized for kind Y; there is no way for a relay to verify what the content tag says. So, you have to trust the delegatee anyways, at which point, what is the point of the restriction?

why would there be no way for the relay to know what the content tag says? it'd be on the same event that is publishing  the delegation string. Are you thinking of kind 4 perhaps? I'm not thinking of encrypted content if that helps.

I.e. using the examples from this NIP it'd be:

```json
{
  "id": "e93c6095c3db1c31d15ac771f8fc5fb672f6e52cd25505099f62cd055523224f",
  "pubkey": "477318cfb5427b9cfc66a9fa376150c1ddbc62115ae27cef72417eb959691396",
  "created_at": 1677426298,
  "kind": 1,
  "tags": [
    [
      "delegation",
      <pubkey of the delegator>,
      "kind=1&content=315f5bdb76d078c43b8ac0064e4a0164612b1fce77c869345bfc94c75894edd3", // sha256(Hello, world!)
      <64-byte Schnorr signature of the sha256 hash of the delegation token>
    ]
  ],
  "content": "Hello, world!",
  "sig": "633db60e2e7082c13a47a6b19d663d45b2a2ebdeaf0b4c35ef83be2738030c54fc7fd56d139652937cdca875ee61b51904a1d0d0588a6acd6168d7be2909d693"
}
```

The relay or clients can see that the content is `Hello, world!`, and that it's sha256 matches the value in the delegation token.

gkbrk:
Looks nice. What's the intended use case for this? I can think of some funny ones, like giving permission to send kind=7 reactions that are only dislikes or likes etc.

Do we want to support multiple `content` hashes, similar to supporting multiple event kinds?

Cameri:
This is already supported on Nostream.

pablof7z:
> Looks nice. What's the intended use case for this? I can think of some funny ones, like giving permission to send kind=7 reactions that are only dislikes or likes etc.
Yeah, another use case I was talking with @scsibug about is, for example, going to see a world cup final, and you want to have a "Congrats Argentina!" or "Congrats France!" type notes; you could give both delegation tokens to a third party, and they can publish the note when it's time.

> Do we want to support multiple `content` hashes, similar to supporting multiple event kinds?

I think that could get complex real quick; the way I had thought about it is the delegator creates as many delegation tokens as needed, and the delegatee just uses the correct one

gkbrk:
> Yeah, another use case I was talking with @scsibug about is, for example, going to see a world cup final, and you want to have a "Congrats Argentina!" or "Congrats France!" type notes; you could give both delegation tokens to a third party, and they can publish the note when it's time.

For this kind of narrow filtering, people might think they're signing for two events to be published max, but in fact the person or system that gets this delegation token can publish an unlimited number of "Congrats France!" and "Congratz Argentina!" notes, tag different people and reply to lots of different events. This can quickly get you added to spam lists, or banned from relays.

Telling people they can give a delegation token and the third party can only post the content they've specified can give users a false sense of security. No one would think "Allow FootballApp to post 'Congratz Argentina!' on your behalf" would allow FootballApp to post this under every thread ever, or posted a million times.

A better workflow for delegation in very narrow cases might be like the following.

1. You are given one or multiple JSON arrays in the form that is used to calculate the SHA-256 event ID, with all the fields filled except the pubkey.
2. You pick the ones you want to sign, put your pubkey in the array and sign the event.
3. When the time comes, the third party can only publish the events you signed once, in exactly the same way you signed it. Because each event you signed has a single ID, they cannot sign multiple events and they cannot modify the event to add mentions. You probably won't get banned for a couple events that you vetted the content for as well.

> I think that could get complex real quick; the way I had thought about it is the delegator creates as many delegation tokens as needed, and the delegatee just uses the correct one.

Yeah, makes sense to keep everyone doing the verification simpler as well. Perhaps something similar should be done for the event kinds too.

Right now there are three conditions you can put on delegation tokens, and they all have totally different support for multiple values.

- `created_at` can specify a range with `>` and `<`. You can also specify multiple `>` and multiple `<` and the max of `>` and the min of `<` applies due to AND'ing the conditions.
- `kind` can be specified multiple times, but it will only result in a delegation that cannot sign anything valid. Multiple `kind` values can be given, but not as a range with `>` and `<` even though both `created_at` and `kind` are integers. Instead they need to be comma-separated.
- A single content hash can condition can be given. Giving multiple conditions for content will result in a delegation token that cannot sign anything.

gkbrk:
Specifically about the `content` hash conditions, I think having a `<` operation also makes sense. And comparing both hex strings and their un-hexed buffers is a well-defined operation, especially when they are both the same size.

This could be used to sign a delegated token for an entity to post kind=1 events on your behalf, but with a proof-of-work requirement to prevent them from spamming a million events.

pablof7z:
> > Yeah, another use case I was talking with @scsibug about is, for example, going to see a world cup final, and you want to have a "Congrats Argentina!" or "Congrats France!" type notes; you could give both delegation tokens to a third party, and they can publish the note when it's time.
> 
> For this kind of narrow filtering, people might think they're signing for two events to be published max, but in fact the person or system that gets this delegation token can publish an unlimited number of "Congrats France!" and "Congratz Argentina!" notes, tag different people and reply to lots of different events. This can quickly get you added to spam lists, or banned from relays.
> 
> Telling people they can give a delegation token and the third party can only post the content they've specified can give users a false sense of security. No one would think "Allow FootballApp to post 'Congratz Argentina!' on your behalf" would allow FootballApp to post this under every thread ever, or posted a million times.
> 
> A better workflow for delegation in very narrow cases might be like the following.
> 
>     1. You are given one or multiple JSON arrays in the form that is used to calculate the SHA-256 event ID, with all the fields filled except the pubkey.
> 
>     2. You pick the ones you want to sign, put your pubkey in the array and sign the event.
> 
>     3. When the time comes, the third party can only publish the events you signed once, in exactly the same way you signed it. Because each event you signed has a single ID, they cannot sign multiple events and they cannot modify the event to add mentions. You probably won't get banned for a couple events that you vetted the content for as well.

That would work when the time window for publishing the event is relatively small, but if the time window is, say, a year, you'd have to sign millions of events with roughly every possible `created_at` value.

> Specifically about the `content` hash conditions, I think having a `<` operation also makes sense. And comparing both hex strings and their un-hexed buffers is a well-defined operation, especially when they are both the same size.
> 
> This could be used to sign a delegated token for an entity to post kind=1 events on your behalf, but with a proof-of-work requirement to prevent them from spamming a million events.

I like the idea of some amount of POW that the delegator can impose on the delegatee. Can you go into a bit more detail on how this would work? If the content is fixed (e.g. `content=sha256("hello, world!")`), where would the nonce for the POW go?

barkyq:
My gut says there must be a better way than giving an untrusted third party a delegation token. But perhaps for apps which cannot connect to a browser extension, etc, and where you want to have a finite set of "reactions" you can make, this makes sense. Does seem a bit ripe for abuse though, if there is no restrictions on tags as well. If the malicious actor adds a bunch of `p` tags could get you blocked by a lot of people... as other people have mentioned above.

As for sending emojis? Which platform would only have emojis but not text messages as well?

Another aspect, how is the user importing the delegation token to these untrusted apps? If there is already some low-friction way to transfer relatively complicated information between a trusted environment and an untrusted one, can't it just be used in reverse to sign events the app wants to send? Some sort of "scan this QR code" in your nostr-app and it will open a dialog where you can publish the event. 

At least the delegation conditions `...&content=hash&...` are very simple. Perhaps annoying for relays and clients to verify.

Working on implementing some of NIP-26 now BTW that is why I am prowling the NIP-26 PRs.

pablof7z:
> My gut says there must be a better way than giving an untrusted third party a delegation token. But perhaps for apps which cannot connect to a browser extension, etc, and where you want to have a finite set of "reactions" you can make, this makes sense. Does seem a bit ripe for abuse though, if there is no restrictions on tags as well. If the malicious actor adds a bunch of `p` tags could get you blocked by a lot of people... as other people have mentioned above.

Yeah, perhaps extending NIP-26 to include tags limitations would also be a good thing. I think that’s be beyond the scope of this PR but definitely q good follow up for a new one.

> 
> As for sending emojis, which platform would only have emojis but not text messages as well? Perhaps there is some cool stuff which can be done, what do you think?

Not sure what you’re referring to wrt emojis, is that comment about this PR?

> 
> Another aspect, how is the user importing the delegation token to these untrusted apps? If there is already some low-friction way to transfer relatively complicated information between a trusted environment and an untrusted one, can't it just be used in reverse to sign events the app wants to send? Some sort of "scan this QR code" in your nostr-app and it will open a dialog where you can publish the event.
> 
I think NIP-26 is mainly useful when the original keys are not available for signing, otherwise you’d just do that instead of using a delegation token?

barkyq:
What else is this PR good for beyond allowing people to send emojis? What other fixed content strings do you anticipate?

pablof7z:
> What else is this PR good for beyond allowing people to send emojis? What other fixed content strings do you anticipate?

Any kind of content; this has a simple example that might illustrate further: https://github.com/nostr-protocol/nips/pull/222#issuecomment-1420220999

> 
> As to the third point, my argument was: if there is way to import a delegation token, there will also be a way to export unsigned events. Like, the app generates a pubkey for you, and you copy paste that to your other app to make a delegation token, and copy paste that back into the app? Or you generate a keypair and paste the secret-key + delegation token into the app? Either way it's pretty high friction. And if there was a low friction way, maybe it could be run in reverse to just export unsigned events.

Well, the way I see it is just use a browser extension to create the delegation token and the app requesting it keeps it along with the keys. This pr in getalby implements it https://github.com/getAlby/lightning-browser-extension/pull/2063 Using it this way is no different than signing an event for the user.

barkyq:
I guess another use case is little phrases like "Go France!" "Go Germany!" In a sports setting. Maybe some political culture-war orgs could ask you for permission to say things like "I Stand With Michelle Obama", "I Stand With Donald Trump", "Boycott Tesla" etc...

pablof7z:
@fiatjaf can we get this merged? Some additional operators were proposed for the `content` tag, but we should add those on subsequent PRs

barkyq:
This would be one of the few NIPs which requires relays to parse the content field (in order to hash it). Is it really worth it for this? What are the use cases? Letting some untrusted app say some predefined thing, arbitrarily many times, with arbitrary `p` and `e` tags, which you can never revoke? I suppose the `created_at<` field can help mitigate the "arbitrarily many times" complaint. 

Why can't the app just say "do you want to sign this event we have written for you" and you give permission on a per-event basis? 

The best use-case I can think of is allowing emoji reactions in some client which has no text message commenting. Maybe some sort of nostr games with limited moves.

Off the top of my head, might be better to just have some sort of per-app mapping of `kind` to `emoji` (say kind 100003020 should be `hug`, etc), and then make a delegation token which only allows those kinds.

Anyways sorry for putting up a challenge, just think it should be thought through a bit more before it is merged.

pablof7z:
> This would be one of the few NIPs which requires relays to parse the content field (in order to hash it). Is it really worth it for this? What are the use cases? Letting some untrusted app say some predefined thing, arbitrarily many times, with arbitrary `p` and `e` tags, which you can never revoke? I suppose the `created_at<` field can help mitigate the "arbitrarily many times" complaint.

Relays don't have to hash it, the delegation token is not created by a relay, it's created by something like a browser extension.

The complains about arbitrary tags are not related to reducing the permitted scope of a delegation token, it's about NIP-26 entirely; and that can be mitigated by reducing the scope further of the delegation token in a subsequent PR.

> 
> Why can't the app just say "do you want to sign this event we have written for you" and you give permission on a per-event basis?

Because the premise of the NIP-26 is that delegator keys are not available to sign at the time of signing the event. If that weren't the case then why have NIP-26 at all? just have the delegator keys always sign the event.

> 
> The best use-case I can think of is allowing emoji reactions in some client which has no text message commenting. Maybe also some sort of nostr games with limited moves.
> 
> Off the top of my head, might be better to just have some sort of per-app mapping of `kind` to `emoji` (say kind 100003020 should be `hug`, etc), and then make a delegation-token which only allows those kinds.
> 
> Anyways sorry for putting up a challenge, just think it should be thought through a bit more before it is merged.
I keep getting the impression that you are misunderstanding the intent of this PR since you bring up emojis which is something that I find utterly uninteresting. Perhaps it would be easier to hop on a call or discuss over DMs (npub1l2vyh47mk2p0qlsku7hg0vn29faehy9hy34ygaclpn66ukqp3afqutajft)?

Cameri:
I might just disable NIP-26 on Nostream at this point. It just gets more complicated and complex for very niches use-cases.

Maybe one day I'll get it.

arthurfranca:
There are many use cases similar to this one:

- Bob sends DM to Alice. Inside the DM (or inside an auxiliary event Alice's client subscribes to) it adds a tag sharing a delegation token to be used by Alice.
- Such delegation token would have the following conditions: `kind=4&content=hash_of_original_DM_content&#published_at=original_DM_created_at`
- This allows Alice's client to create a copy of the DM she can later delete. Her inbox will be the list of copies instead of list of originals.

What we earn:
1) Receiver can clear its inbox
2) Sender can still clear its sent-inbox but without affecting Receiver inbox
3) (Maybe) Sender can't delete the copy in Receiver inbox (considering delegator can't delete the delegated copy - NIP-26 isn't clear about it)
4) Sender can learn if receiver did receive the message (check if copy is there, or better yet, if auxiliary event was deleted by Receiver's client - whould need a delegation in aux event allowing Receiver to delete events with the aux id only)

Above example would need tag condition too (the published_at tag). NIP-26 still doesn't have it.

pablof7z:
No longer a NIP-26 apologists

This PR is a good indicator of where NIP-26 was going to lead of broadly adopted.

PR #223, opened 2023-02-05
NIP-?? - WIP - Source Control Context 

garyray-k:
I laid down a rough draft of how we can decentralize source control context using nostr. I am opening this Work In Progress PR so I can share it with others and gather feedback, as I'm certain that I missed something in my initial thoughts.

### DIFF
--- a/69.md
+++ b/69.md
@@ -1,4 +1,4 @@
-NIP-??
+NIP-69
 ======
 
 Source Control Context
@@ -6,18 +6,18 @@ Source Control Context
 
 `draft` `optional` `author:garyKrause_`
 
-This NIP defines ways source control (git, etc.) context can be stored using nostr. Git is already decentralized, but the contexts around a git repository are largely centralized (PR comments, issue tracking, releases, etc). With the correct approach a nostr relay could store this context using many of the existing NIPs. This NIP defines a new kind, `kind: ??`, to represent a source control repository. This NIP purposely does not indicate git as a requirement but many of the concepts will be based around git methods of source control. The author hopes that the NIP is flexible enough to accommodate other source control methodologies.
+This NIP defines ways source control (git, etc.) context can be stored using nostr. Git is already decentralized, but the contexts around a git repository are largely centralized (PR comments, issue tracking, releases, etc). With the correct approach a nostr relay could store this context using many of the existing NIPs. This NIP defines a new kind, `kind: 69`, to represent a source control repository. This NIP purposely does not indicate git as a requirement but many of the concepts will be based around git methods of source control. The author hopes that the NIP is flexible enough to accommodate other source control methodologies.
 
 ## Definition of a Source Control Event
 
-`kind: ??` uses tag the `b` tag to identify the base repository. The value of this tag SHALL be a list of remote source control repository URLs. These URLs MAY be a nostr relay URL or a traditional repository URL. The first URL in this list SHALL be considered the "default" repository URL for the context.
+`kind: 69` uses tag the `b` tag to identify the base repository. The value of this tag SHALL be a list of remote source control repository URLs. These URLs MAY be a nostr relay URL or a traditional repository URL. The first URL in this list SHALL be considered the "default" repository URL for the context.
 
 The `content` of a Source Control Event MAY contain a text description of the event or other information about the repository.
     
 ```json
 {
     ...
-  "kind": ??,
+  "kind": 69,
   "tags": [
     ["b", <repository url>, <repository url>, ...],
   ],
@@ -25,7 +25,7 @@ The `content` of a Source Control Event MAY contain a text description of the ev
   ...
 ```
 
-A relay MAY host the repository along side it's nostr database which can be discovered using a standard `REQ` message filtering for `kind: ??`.
+A relay MAY host the repository along side it's nostr database which can be discovered using a standard `REQ` message filtering for `kind: 69`.
 
 ### Merge Requests
 
@@ -34,7 +34,7 @@ A source control event MAY contain an `m` tag indicating the event is specific t
 ```json
 {
     ...
-  "kind": ??,
+  "kind": 69,
   "tags": [
     ["b", <repository url>, <repository url>, ...],
     ["m": <branch to be merged into>, <branch to be merged in>, <URL of merging in repository>],
--- a/69.md
+++ b/69.md
@@ -12,7 +12,7 @@ This NIP defines ways source control (git, etc.) context can be stored using nos
 
 `kind: 69` uses tag the `b` tag to identify the base repository. The value of this tag SHALL be a list of remote source control repository URLs. These URLs MAY be a nostr relay URL or a traditional repository URL. The first URL in this list SHALL be considered the "default" repository URL for the context.
 
-The `content` of a Source Control Event MAY contain a text description of the event or other information about the repository.
+The `content` of a Source Control Event MUST contain a text description or summary of the base repository. This enables clients that don't specifically handle this kind to display meaningful text in their UI.
     
 ```json
 {
@@ -21,7 +21,7 @@ The `content` of a Source Control Event MAY contain a text description of the ev
   "tags": [
     ["b", <repository url>, <repository url>, ...],
   ],
-  "content": <arbitrary string>,
+  "content": <summary or description>,
   ...
 ```
 
--- a/69.md
+++ b/69.md
@@ -4,7 +4,7 @@ NIP-69
 Source Control Context
 ----------------------
 
-`draft` `optional` `author:garyKrause_`
+`draft` `optional` `author:garyKrause_` `author: OpenSourceOptimist` 
 
 This NIP defines ways source control (git, etc.) context can be stored using nostr. Git is already decentralized, but the contexts around a git repository are largely centralized (PR comments, issue tracking, releases, etc). With the correct approach a nostr relay could store this context using many of the existing NIPs. This NIP defines a new kind, `kind: 69`, to represent a source control repository. This NIP purposely does not indicate git as a requirement but many of the concepts will be based around git methods of source control. The author hopes that the NIP is flexible enough to accommodate other source control methodologies.
 
--- a/69.md
+++ b/69.md
@@ -19,7 +19,7 @@ The `content` of a Source Control Event MUST contain a text description or summa
     ...
   "kind": 69,
   "tags": [
-    ["b", <repository url>, <repository url>, ...],
+    ["b", <repository remote identifier (url, etc)>, <repository remote identifier (url, etc)>, ...],
   ],
   "content": <summary or description>,
   ...
--- a/69.md
+++ b/69.md
@@ -36,8 +36,8 @@ A source control event MAY contain an `m` tag indicating the event is specific t
     ...
   "kind": 69,
   "tags": [
-    ["b", <repository url>, <repository url>, ...],
-    ["m": <branch to be merged into>, <branch to be merged in>, <URL of merging in repository>],
+    ["b", <repository remote identifier (url, etc)>, <repository remote identifier (url, etc)>, ...],
+    ["m": <commit hash (branch head) to be merged into>, <commit hash (branch head) to be merged in>, <URL of merging in repository>],
     ["l": <file name/path>, <line number>]
   ],
   "content": <arbitrary string>,
--- a/69.md
+++ b/69.md
@@ -25,8 +25,6 @@ The `content` of a Source Control Event MUST contain a text description or summa
   ...
 ```
 
-A relay MAY host the repository along side it's nostr database which can be discovered using a standard `REQ` message filtering for `kind: 69`.
-
 ### Merge Requests
 
 A source control event MAY contain an `m` tag indicating the event is specific to a merge request. If present, the `m` tag SHALL define the branch being merged into and the branch being merged in. An event MAY include the URL of the repository being merged from. This could be included when the URL being merged from is not in the `b` tag of the event or for specificity. When creating a merge request, the `content` field MAY contain a justification/explanation for the merge. The `l` tag MAY be included with an `m` tag. The `l` tag specifies the relative file path and line number that the event is referencing.
--- a/draft.md
+++ b/draft.md
@@ -6,7 +6,7 @@ Source Control Context
 
 `draft` `optional` `author:garyKrause_`
 
-This NIP defines ways source control (git, etc.) context can be stored using nostr. Git is already decentralized, but the contexts around a git repository are largely centralized (PR comments, issue tracking, releases, etc). With the correct approach a nostr relay could store this context using many of the existing NIPs. This NIP defines a new kind, `kind: ??`, to represent a source control repository. This NIP purposely does not indicate git as a requirement but many of the concepts will be based around git methods of source control. The author hopes that the NIP is flexible enough to accomodate other source control methodologies.
+This NIP defines ways source control (git, etc.) context can be stored using nostr. Git is already decentralized, but the contexts around a git repository are largely centralized (PR comments, issue tracking, releases, etc). With the correct approach a nostr relay could store this context using many of the existing NIPs. This NIP defines a new kind, `kind: ??`, to represent a source control repository. This NIP purposely does not indicate git as a requirement but many of the concepts will be based around git methods of source control. The author hopes that the NIP is flexible enough to accommodate other source control methodologies.
 
 ## Definition of a Source Control Event
 
@@ -29,7 +29,7 @@ A relay MAY host the repository along side it's nostr database which can be disc
 
 ### Merge Requests
 
-A source control event MAY contain an `m` tag indicating the event is specific to a merge request. If present, the `m` tag SHALL define the branch being merged into and the branch being merged in. An event MAY include the URL of the repository being merged from. This could be included when the URL being merged from is not in the `b` tag of the event or for specificity. When creating a merge request, the `content` field MAY contain a justification/explanation for the merge
+A source control event MAY contain an `m` tag indicating the event is specific to a merge request. If present, the `m` tag SHALL define the branch being merged into and the branch being merged in. An event MAY include the URL of the repository being merged from. This could be included when the URL being merged from is not in the `b` tag of the event or for specificity. When creating a merge request, the `content` field MAY contain a justification/explanation for the merge. The `l` tag MAY be included with an `m` tag. The `l` tag specifies the relative file path and line number that the event is referencing.
 
 ```json
 {
@@ -38,6 +38,7 @@ A source control event MAY contain an `m` tag indicating the event is specific t
   "tags": [
     ["b", <repository url>, <repository url>, ...],
     ["m": <branch to be merged into>, <branch to be merged in>, <URL of merging in repository>],
+    ["l": <file name/path>, <line number>]
   ],
   "content": <arbitrary string>,
   ...
--- a/draft.md
+++ b/draft.md
@@ -1,0 +1,48 @@
+NIP-??
+======
+
+Source Control Context
+----------------------
+
+`draft` `optional` `author:garyKrause_`
+
+This NIP defines ways source control (git, etc.) context can be stored using nostr. Git is already decentralized, but the contexts around a git repository are largely centralized (PR comments, issue tracking, releases, etc). With the correct approach a nostr relay could store this context using many of the existing NIPs. This NIP defines a new kind, `kind: ??`, to represent a source control repository. This NIP purposely does not indicate git as a requirement but many of the concepts will be based around git methods of source control. The author hopes that the NIP is flexible enough to accomodate other source control methodologies.
+
+## Definition of a Source Control Event
+
+`kind: ??` uses tag the `b` tag to identify the base repository. The value of this tag SHALL be a list of remote source control repository URLs. These URLs MAY be a nostr relay URL or a traditional repository URL. The first URL in this list SHALL be considered the "default" repository URL for the context.
+
+The `content` of a Source Control Event MAY contain a text description of the event or other information about the repository.
+    
+```json
+{
+    ...
+  "kind": ??,
+  "tags": [
+    ["b", <repository url>, <repository url>, ...],
+  ],
+  "content": <arbitrary string>,
+  ...
+```
+
+A relay MAY host the repository along side it's nostr database which can be discovered using a standard `REQ` message filtering for `kind: ??`.
+
+### Merge Requests
+
+A source control event MAY contain an `m` tag indicating the event is specific to a merge request. If present, the `m` tag SHALL define the branch being merged into and the branch being merged in. An event MAY include the URL of the repository being merged from. This could be included when the URL being merged from is not in the `b` tag of the event or for specificity. When creating a merge request, the `content` field MAY contain a justification/explanation for the merge
+
+```json
+{
+    ...
+  "kind": ??,
+  "tags": [
+    ["b", <repository url>, <repository url>, ...],
+    ["m": <branch to be merged into>, <branch to be merged in>, <URL of merging in repository>],
+  ],
+  "content": <arbitrary string>,
+  ...
+```
+
+### Issue/Bug Tracking
+
+????


fiatjaf:
Are you working on this?

I was working on a [git plugin](https://github.com/fiatjaf/git-remote-nostr) myself to use git repos stored on Nostr, but eventually decided it was a bad idea. And it is better to use normal git servers for git -- and have them be self-hosted or hosted on a variety of service providers, then use Nostr to send open issues, send patches, comment on patches and for repository discovery.

I feel like these four functions above would make Nostr into an actually decentralized alternative to GitHub. Sending and applying patches was already done by @jb55 at http://git.jb55.com/git-nostr-tools, although it isn't being used much and I don't know if and how it could be improved.

The reason why hosting git data directly on Nostr is a bad idea is that it would be extremely inefficient and storage requirements would be massive -- so relays would start blocking this kind of thing and we would have to use special relays designed just for git-nostr, and at this point it is better to just use normal git servers that already exist. Each with its own specialty.

mikedilger:
> hosting git data directly on Nostr is a bad idea

Agreed. Also the git protocol is a very efficient optimized for git, and git itself is already fully distributed.

garyray-k:
It's something I've dedicated some time to considering and working on a rough draft MVP implementation, but I wanted to open up the topic for discussion so I can hopefully learn from others that have given this some thought as well.

I 100% agree on not hosting git repositories on relays by default. But the discovery of git repos on a relay that decided to host the data could be part of the NIP, as I've outlined in the draft.

The NIP I began writing is outlining a way to store the _context_ around source control. The draft here proposes using a tag in a specific kind to track the remote URL(s) for a given repository and creating an event from that. 
From there, clients can NIP-10 most things to create context like comments/issues/etc.

There will be metadata needed for tracking line-specific comments and that metadata could be included in the tags part of an Event of this kind.

QQ: How do NIPs get numbered?

arkin0x:
@garyray-k I really like the idea of using nostr for context and keeping git hosting on git servers. I'm imagining a web-based nostr client that is essentially a github-inspired interface for working on PRs, issues, bugs, etc.

Having a conversation around a PR on nostr is easy. But I think we would need some kind of nostr-enabled git hook to publish events that represent code changes on the remote repo, like this:

image

The other thing is that making edits to a repository via this nostr-git-client seems like a potentially bad idea since putting private keys into nostr apps is still not ideal security-wise... NIP26 is out there now so maybe this is going to get better. But you could work on your repo from CLI or Git client as normal and see your pushed changes show up in nostr-git-client because they would be published to the PR by a hook on the remote repo.

Not having the ability to directly interact with a repo via the client would make things like forking more difficult and clunky though... No forking, no committing, no merging, no creating new repos...

> QQ: How do NIPs get numbered?

I was also wondering this 😅

garyray-k:
> Having a conversation around a PR on nostr is easy. But I think we would need some kind of nostr-enabled git hook to publish events that represent code changes on the remote repo, like this:

I think the client would be responsible for this aspect finding new changes and updating the display. For an MVP, I envision the need for a client with two libraries, nostr and git (oversimplifying here). The git library is largely reading repositories and the nostr library publishes events with metadata to tie the event to a specific part of the git repo.

hendore:
I've been experimenting with this for a couple of days, and made decent progress, hit a few brick walls here and there but I thought I would give my input/conclusion.

I started with a script that I could use to publish git refs (heads/tags) and objects (commits/trees/blobs) to a unique repository, the plan here was to get the event structures/schemas sorted before porting this script to a git-remote-helper so that we could simple push/fetch using git with a custom remote handler

`git remote add origin nostr://some.relay.io/814add12473702fcf229d4786bc46e5ddbabbb88027037e371ec4c7947de284d`

After getting this far, I felt quite confident this could work, so I set about creating a client to interact with these events, that turned out pretty straight forward, I was able to create new repositories, update their details/settings etc... once a repository was created I could push branches up (with my dev script / not a git-remote-helper yet). At this point, I could browse all of my repositories and see all of their refs along with view the tree and latest commit of each ref.

What I found was, all refs/objects are quite light minus blobs, moving over to packed git objects (deltas) rather than loose objects (no deltas) would certainly help here but then I thought about maybe storing blobs on the filesystem outside of the database (but this would rely on relay changes and I was trying to do this without introducing any new relay features than already proposed). Turned out it wasn't actually the end of the world with the files I tested on since my UI was only ever subscribing to blob objects when viewing a file.

Few screenshots of the progress 

Screenshot 2023-02-06 at 18 36 56
Screenshot 2023-02-06 at 18 38 17
Screenshot 2023-02-06 at 18 44 32

So things where looking good, however this was all just a big experiment, in the end I 100% came to the same conclusions that have been mentioned. Although I found there wasn't any performance issues with the test repos/branches I had published to a relay I'm sure with larger repositories it could be a concern.

The funny thing is, before I started any development I wrote about a few options, having a git-server was the in between option

Screenshot 2023-02-06 at 19 35 32

All that said, it was a fun experience, I learnt a tonne of git internals along the way so it wasn't a complete waste of time but I think it might be time I hang up the boots on this one. If anyone want's to discuss how a git-server model may look, I would be up for discussing / working on something.

mikedilger:
> QQ: How do NIPs get numbered?

What I've done is look through all the PRs, find the highest number, and go one higher.

fiatjaf:
I like this draft and the initiative. Thank you @garyray-k.

I agree with the idea of having a dedicated kind for "merge requests" (and possibly one for "issues"?), but the responses to these seem to me to fit better in `kind:1` events.

It would be nice to have these notes -- i.e. discussions about a merge request -- to show up in the normal "social" feed of users, right? And then social feed users could click on the context to see the event they were replying to and then what would they see? An event (that they would not understand) specifying a merge request. Do you think this is good or bad? Did you think about that?

ofer-elrom:
It seems we are all here in agreement that relays should be a transport layer rather than a storage medium. This points to git servers holding the data and communicating over Nostr with git clients. 
In my opinion, censorship resistance is very important for open-source code moving forward, and Git on Nostr can supply it. The trick is having publicly known code repository entities to which public communication is connected while having their data "hidden," private but verifiable.
I think the solution, in short, is that Nostr-authenticated developers can use a Nostr-compatible git client to push to and pull from Nostr-authenticated git suppliers using Nostr relays as a communication layer.

It can connect with @gayray-k suggestion above.
I am not sure I have the time to create this on my own. I would appreciate any remarks on this and help get more people on board with the idea.
For a more elaborate explanation, please see this google document: [docs.google.com/document/d/1z96XpTtrUvrFZYPjMzlT5JuE41HF6X5vI99lvImVCvU](https://docs.google.com/document/d/1z96XpTtrUvrFZYPjMzlT5JuE41HF6X5vI99lvImVCvU/)

ofer-elrom:
It seems we are all here in agreement that relays should be a transport layer rather than a storage medium. This points to git servers holding the data and communicating over Nostr with git clients. 
In my opinion, censorship resistance is very important for open-source code moving forward, and Git on Nostr can supply it. The trick is having publicly known code repository entities to which public communication is connected while having their data "hidden," private but verifiable.
I think the solution, in short, is that Nostr-authenticated developers can use a Nostr-compatible git client to push to and pull from Nostr-authenticated git suppliers using Nostr relays as a communication layer.

It can connect with @gayray-k suggestion above.
I am not sure I have the time to create this on my own. I would appreciate any remarks on this and help get more people on board with the idea.
For a more elaborate explanation, please see this google document: [docs.google.com/document/d/1z96XpTtrUvrFZYPjMzlT5JuE41HF6X5vI99lvImVCvU](https://docs.google.com/document/d/1z96XpTtrUvrFZYPjMzlT5JuE41HF6X5vI99lvImVCvU/)

garyray-k:
@hendore Nice research! I'm glad you concluded that relays _mustn't_ hold repos, but there will likely be relay servers that do both for hosting simplicity.

--------------------------------------

> It would be nice to have these notes -- i.e. discussions about a merge request -- to show up in the normal "social" feed of users, right? 

I think this will happen with `kind: 1` events as replies. To me, it makes a relay sort of a town square. You might not care to hear about merge request discussions, but by being on the "global" feed of a relay you will see it nonetheless.

> And then social feed users could click on the context to see the event they were replying to and then what would they see? An event (that they would not understand) specifying a merge request. Do you think this is good or bad? Did you think about that?

I will have to do some research on how clients are handling the "content" field. My initial thought would be to standardize the `content` field of the merge request event so a client that does not explicitly handle "merge request events" can still display something meaningful. Along with this research, I wonder if clients filter based on root events. That seems possible but it might not scale to a certain point.

- [ ] research how some clients handle unknown `kind`s

-----------

> I think the solution, in short, is that Nostr-authenticated developers can use a Nostr-compatible git client to push to and pull from Nostr-authenticated git suppliers using Nostr relays as a communication layer.

Git already handles authentication and even has mechanisms for separate authors and committers. The goal of this NIP would be to anchor series of events to a single git "thing". The simplest I reasoned about was a merge request. 
How would Nostr auth enhance existing git authentication?

fiatjaf:
> My initial thought would be to standardize the content field of the merge request event so a client that does not explicitly handle "merge request events" can still display something meaningful.

That's kind of what I was thinking.

ofer-elrom:
> Git already handles authentication and even has mechanisms for separate authors and committers. The goal of this NIP would be to anchor series of events to a single git "thing". The simplest I reasoned about was a merge request.
> How would Nostr auth enhance existing git authentication?

My purpose with "Nostr Auth" is to separate the constant public point of contact from a dynamic, nonconstant uncensorable IP generated for a git session. At the first simple stage, this IP may be the IP of a git server that is generated by a VPN per connection request. At later stages, the entire client-server git communication can be done by both the client and the server polling on private messages sent to them and transmitting data over relays with no server IP involved.
This does not help in git authentication. It helps enable a *censorship-resistant public git service* that can synchronize open-source development.

garyray-k:
@ofer-elrom 
While I appreciate the consideration, I think what you're describing would be outside the scope of a NIP. From my understanding, you're trying to create a new way for git to be communicated/networked. That would not be the goal of this NIP. This NIP is aimed at merely defining a way to reference source control primitives (ie. merge request) as a nostr event so further context can be built from the root reference point.

kouloumos:
I like were this is going. 
Feature-parity with GitHub reviews (e.g. resolving inline review comments) is one of the evaluation schemes used when thinking about GitHub alternatives for Bitcoin Core ([src](https://github.com/bitcoin-core/bitcoin-devwiki/wiki/GitHub-alternatives-for-Bitcoin-Core)) (and I assume for any other large-scale open source project). 

Maybe https://github.com/zw/bitcoin-gh-meta can give some inspiration on the required context schemes.

garyray-k:
Someone brought up a good point of consideration: how do disparate notes maintain their tie together?

Example:
Dev A creates PR and the root event for the source control context.
Dev B creates a reply note using NIP-10
Dev A and B are publishing to different relays.
Dev B's relay goes dark and nukes all data.

How do we maintain this context?

mikedilger:
Nostr provides redundancy if you utilize it.  Dev B should post to more than one relay.

Now the trick here is that people following the repo need to see Dev B's note, so instead of what we do for a microblog (everyone writing wherever they prefer), repos should probably be hosted on a (per-repo) well defined set of relays (but more than just one) chosen by the owner of the repo, and these relays configured in a way that they allow access appropriately to this repo.

69.md
```diff
@@ -0,0 +1,49 @@
...
+  "kind": 69,
+  "tags": [
+    ["b", <repository url>, <repository url>, ...],
+  ],
+  "content": <summary or description>,
+  ...
+```
+
+A relay MAY host the repository along side it's nostr database which can be discovered using a standard `REQ` message filtering for `kind: 69`.
+
```
OpenSourceOptimist, line 29: I feel like I am missing something here. Would the repository be a standard ssh/http git repo? If so then why is this needed? The relay owner that want to run a git repo can just send a kind 69 note to the relay and it will show up.

```suggestion
```

kirkawolff:
The one assumption github makes about the git repository itself is there is only one official location to push and pull from. This makes sense in the case of something that's intended to be a frontend to a git server, but it doesen't leverage the decentralized intent behind git.

Github stores PRs on the central repo on a branch, which requires write access to the central repo. It makes more sense in decentraland to point to a branch on a git remote rather than requiring the project owner to manage a ratsnest of ACLs for everyone that wants to push. 

Github solves this by allowing references to forks inside the platform, but this is clunky and clutters your own account and if you dont intend to actually fork the project then you will never make use of all of the other features in your own account.

It would be nice to have a "pure" github URI place to store your public copy of the repo without needing to access it through a bloated CMS like github. This could be as simple as your own ssh server and then anonymous read-only access is provided to the "Project context" as (multiple?) URI and sshkey fields. Alternatively, there might be future integration with git-torrent to host the central repo.

69.md
garyray-k: Good point. This was my way to describe a variation of the model but isn't needed in the NIP. Thank you.

melvincarvalho:
FYI: NIP-17 might be related and of some interest

https://nip17.com/

I've been thinking about this for a couple of years since before nostr, and I realized that nostr pub/sub was the missing piece I needed to have a source control system without lock in

I've decided that full source control or full github is too much to take an and achieve, so I will focus first on just commits, which can lead to continuous deployment/integration without lock in, and also ability to move and discover repositories.  After that is working hopefully extension.  Will follow this work with interest!

garyray-k:
> Github stores PRs on the central repo on a branch, which requires write access to the central repo. It makes more sense in decentraland to point to a branch on a git remote rather than requiring the project owner to manage a ratsnest of ACLs for everyone that wants to push.

The owner of the root repository Event would be listing the remote locations that are canonical. How they manage ACL for those is outside the scope of this NIP. 

> It would be nice to have a "pure" github URI place to store your public copy of the repo without needing to access it through a bloated CMS like github. This could be as simple as your own ssh server and then anonymous read-only access is provided to the "Project context" as (multiple?) URI and sshkey fields. Alternatively, there might be future integration with git-torrent to host the central repo.

This already exists. You can host remote repos outside of github that people can clone from. This is a decent article describing the process (https://www.inmotionhosting.com/support/website/git/git-server/). And if you look at a GitHub URL, it's a remote URL + file path.

69.md
```diff
@@ -0,0 +1,47 @@
...
+A source control event MAY contain an `m` tag indicating the event is specific to a merge request. If present, the `m` tag SHALL define the branch being merged into and the branch being merged in. An event MAY include the URL of the repository being merged from. This could be included when the URL being merged from is not in the `b` tag of the event or for specificity. When creating a merge request, the `content` field MAY contain a justification/explanation for the merge. The `l` tag MAY be included with an `m` tag. The `l` tag specifies the relative file path and line number that the event is referencing.
+
+```json
+{
+    ...
+  "kind": 69,
+  "tags": [
+    ["b", <repository url>, <repository url>, ...],
+    ["m": <branch to be merged into>, <branch to be merged in>, <URL of merging in repository>],
```
OpenSourceOptimist, line 38: I think there is a devil in the detail of how we specify the branches.

Using just the human readable names like `main`, `my-feature-branch`, etc..
there will be ambiguity about what change the request actually want to do. We
could quite easily sort out the URL ambibuity, but the more serious ambiguity is
that what a branch name points to changes over time.

Imagine you get a merge request to bitcoin core to merge the branch
`really-cool-change` and diff looks good, you send the approve event (if that is
how it will work). Then new malisious commits are pushed to the
`really-cool-change` before the branch is merged.

There is also a question around reciliency. If the commit hash is included in
the branch definition you could push your changes to multiple git servers and as
long as one is still up and not cencoring you the merge request is still there.
Just like nostr.

Using the commit hash in the event we need a new kind 69 events every time you
make a change to the request. In that scenario I imagine it would be very useful
link them. Not sure what other complication might come out of that though...

melvincarvalho:
> Someone brought up a good point of consideration: how do disparate notes maintain their tie together?
> 
> Example: Dev A creates PR and the root event for the source control context. Dev B creates a reply note using NIP-10 Dev A and B are publishing to different relays. Dev B's relay goes dark and nukes all data.
> 
> How do we maintain this context?

This is IMHO quite a hard problem so solve fully.  There may be some partial solutions, but git trees and commits can get quite complex.  There are also race conditions, which ultimately may need to be byzantine fault tolerant.  

Hence, I went for a simpler approach, and that is that the repo will be able to publish the definitive state to a relay.  Future extensions would deal with the different race conditions.  This is similar to continuous deployment type models.  Once that is working, more could be attempted.

69.md
garyray-k: Good point. branch names are just human-readable pointers to where the HEAD commit should be.

OpenSourceOptimist:
How do we feel about adding the POSSIBILITY to include the patch in a merge request event somehow?

It would enable users to create merge requests without having write access to any git server. Could be a confusing first time experience if you need to find a git server to host your changes. Clients would probably be forced to have default git servers for their users just like with images now.

The obvious downside is that including the patch would create large events. Completely agree with the general sentiment that the relays should not store the repository. Sending patch files around is the OG way to use git so it could make sense to support them in this brave new word. Maybe another kind and another NIP?

For example this is the patch for this PR https://patch-diff.githubusercontent.com/raw/nostr-protocol/nips/pull/223.patch

69.md
kirkawolff: Agreed, PRs should only be uri with a hash, but a label on a branch on a uri shouldn't be allowed to be specific enough

garyray-k:
Posting note1qj84ua9mklntsqvk2khytfpwj0jtq297x84u7k4de6nq9rz060kq9zt84j here for context:

> #[3]
>
>I think the way this bounty is stated is put is not the ideal. I think most people will read this and think we need a big website that is just like GitHub but using Nostr somehow. I think that is not what we should see (and hopefully that's not what #[1] wants either).
>
>What I would want to see are multiple apps that can interoperate and are able to perform separate functions:
>- browse code
>- comment on code (referenced by a commit)
>- create issues and comment on issues
>- send patches
>- comment on patches
>
>And how these should be done? I am not sure, but here's what I have in mind:
>- most of the comment things should probably be kind:1 events, I don't know, with some extra tags (so they could be interacted with from the normal "social" Nostr clients? or not?)
>- code should probably be hosted by standalone dedicated git servers -- and there could be centralized providers offering these services but they should interoperate seamlessly between themselves and with standalone personal servers
>- sending patches should probably be done using something like this approach by #[0]: http://git.jb55.com/git-nostr-tools/file/README.txt.html
>
>#[2] has opened a discussion on this topic on the NIPs repository that could possibly be used to coordinate the efforts: https://github.com/nostr-protocol/nips/pull/223
>
>I think we could have multiple different smallish webapps, native apps and specially command-line tools that implement one or multiple of the separate functions described above, and with that we can achieve a much better result both in terms of quality and of decentralization than if someone or some big team decides to tackle the entire cake and come up with some centralizing architecture on their own.

garyray-k:
@melvincarvalho 
> This is IMHO quite a hard problem so solve fully. There may be some partial solutions, but git trees and commits can get quite complex. There are also race conditions, which ultimately may need to be byzantine fault tolerant.
>
>Hence, I went for a simpler approach, and that is that the repo will be able to publish the definitive state to a relay. Future extensions would deal with the different race conditions. This is similar to continuous deployment type models. Once that is working, more could be attempted.

By publishing the definitive state of the repo to a relay, you're recreating git. That is **not** the goal of this NIP. Tracking which git remote repository _is_ the definitive state is encompassed here using the `b` tag mentioned in `kind: 69` events.

MazziaRick:
What about using IPFS at least for public repositories ? 
For example automatically commit a ".bundle" file to IPFS to make it accessible even if the origin git server has been shutdown in some way....

garyray-k:
Commenting note1tgy2myzq7g5dyckjm47rm7qsmu48dktgna7m5smcp4qng4kgha8qa5n56g for context.

> Linus wrote git to be distributed (decentralized) and his workflow with email patches was a core design feature.
>
>Github projected a centralized workflow onto git (branch permissions and all that jazz are not a part of git). This has its place especially in the enterprise world. So why do we want branch permissions for open source projects?
>
>I think it comes down to trusted reviewers. There are those rare combinations of talent, common sense and OCD that make excellent master branch guardians.
>
>So all I think we need is
>1. the feature of broadcasting merge requests on nostr
>2. the feature of trusted reviewers signing their approval to this request
>3. the feature of cryptographically coming up with a deterministic merge outcome and commit hash so every distributed client would have an identical merge
> 4. a git client/client extension to automatically merge based on preconfigured sign events
>
> Every developer could still hold the code local, plus there could be X public repository servers.

garyray-k:
>How do we feel about adding the POSSIBILITY to include the patch in a merge request event somehow?
>
>It would enable users to create merge requests without having write access to any git server. Could be a confusing first time experience if you need to find a git server to host your changes. Clients would probably be forced to have default git servers for their users just like with images now.
>
> The obvious downside is that including the patch would create large events. Completely agree with the general sentiment that the relays should not store the repository. Sending patch files around is the OG way to use git so it could make sense to support them in this brave new word. Maybe another kind and another NIP?

🤔
I don't see a reason not to aside from size of the event, but perhaps larger patches could be serialized into a thread of events. I would also imagine this could promote small patch sizes.

The critical thing touched on here is the barrier to entry for a new contributor. Spinning up a remote repository is not trivial.

melvincarvalho:
FYI: we decided previously on telegram to use the tag "commit" for a git commit

https://t.me/nostr_protocol/76815

This is possibly what you are calling a "merge"

garyray-k:
@MazziaRick
> What about using IPFS at least for public repositories ?
> For example automatically commit a ".bundle" file to IPFS to make it accessible even if the origin git server has been shutdown in some way....

This is a git implementation detail and outside the scope of this NIP. IPFS or otherwise, the remote accessibility of a repo would be in the `b` tag if adhering to this NIP.

melvincarvalho:
> By publishing the definitive state of the repo to a relay, you're recreating git. 

The git state is encapsulated in a git hash.  Originally the idea was to put that in the "content" field.  But that's evolved now to using a tag ie "commit" or "c" (if indexed).  Unsure that's recreating git, more publishing in a relay what the current state is, for clients to follow

> That is not the goal of this NIP. Tracking which git remote repository is the definitive state is encompassed here using the b tag mentioned in kind: 69 events.

I think the difference between this NIP and NPI17 is that I'm giving repos a nostrID, and you are giving developers a NostrID.  Seems to be different numbers of moving parts and complexities.  Looking forward to seeing what you produce!

garyray-k:
@melvincarvalho 
> This is possibly what you are calling a "merge"

Negative. A commit is a [git object](https://git-scm.com/book/en/v2/Git-Internals-Git-Objects). A [merge](https://git-scm.com/docs/git-merge) is joining two or more development histories (which is commonly tracked via the HEAD commit aka branch).

garyray-k:
> How do we feel about adding the POSSIBILITY to include the patch in a merge request event somehow?
> 
> It would enable users to create merge requests without having write access to any git server. Could be a confusing first time experience if you need to find a git server to host your changes. Clients would probably be forced to have default git servers for their users just like with images now.
> 
> The obvious downside is that including the patch would create large events. Completely agree with the general sentiment that the relays should not store the repository. Sending patch files around is the OG way to use git so it could make sense to support them in this brave new word. Maybe another kind and another NIP?
> 
> For example this is the patch for this PR https://patch-diff.githubusercontent.com/raw/nostr-protocol/nips/pull/223.patch

@OpenSourceOptimist 👀
I think this is already solved for.
"""
Once the bundle file has been generated, it can be sent over any transport to the remote host for reconstitution. This might involve burning to a CD, via a USB stick or some other network protocol. 
"""
https://alblue.bandlem.com/2011/12/git-tip-of-week-patches-by-email.html

garyray-k:
https://t.me/+oeRH7x1_OuNlYmMx

Telegram chat as a secondary location to discuss this NIP and git context on nostr in general.

nixuno:
Would the intention of this be for source control server implementations (GitLab, Gitea, etc) to be Nostr-aware and publish these events as a "user" with their repo URL as the "default" repository URL? Would it make sense for any additional metadata about these repos such as the owner (which could be tied to a npub) to be included? Otherwise, any client implementing this NIP would be reliant on the default source's implementation to provide some standardized API to pull Nostr-specific metadata.

garyray-k:
> Would the intention of this be for source control server implementations (GitLab, Gitea, etc) to be Nostr-aware and publish these events as a "user" with their repo URL as the "default" repository URL? Would it make sense for any additional metadata about these repos such as the owner (which could be tied to a npub) to be included? Otherwise, any client implementing this NIP would be reliant on the default source's implementation to provide some standardized API to pull Nostr-specific metadata.

@nixuno 
That is not the intention of this NIP but those server implementations could become Nostr-aware. That project would be outside the scope of this NIP.

ofer-elrom:
> Someone brought up a good point of consideration: how do disparate notes maintain their tie together?
> 
> Example: Dev A creates PR and the root event for the source control context. Dev B creates a reply note using NIP-10 Dev A and B are publishing to different relays. Dev B's relay goes dark and nukes all data.
> 
> How do we maintain this context?

I think the problem reflects a more general issue: participants' interactions (discussion) around the source code are sometimes as meaningful as the source code. 
The existing handling of Nostr events does not allow for preserving them at the same level a reliable git server maintains source code files.
Since this NIP is about defining an addition to the protocol and not about implementing a specific thing on top of the protocol, we need to think about what we can add to the protocol to allow builders to solve this issue.
Perhaps adding a list of relays to the Source Control event that is the root of the  Merge Request can be helpful. These relays may be more reliable for the expected duration of the merge process.
Supporting clients can then write the following connected event to all these relays.
When the merge completes, the git "service supplier" may choose to read all these events and commit them as file "next" to the repository to be scanned or propagated back to some relays on demand.

OpenSourceOptimist:
@garyray-k 
> but perhaps larger patches could be serialized into a thread of events. 


> I would also imagine this could promote small patch sizes.

> The critical thing touched on here is the barrier to entry for a new contributor. Spinning up a remote repository is not trivial.

> Edit: Thinking more on this, and it may make sense for this to be its own kind of event. That way the content of the event is just the patch text. Then clients can easily filter out those kinds where applicable.

> That way the content of the event is just the patch text. 

Then clients can easily filter out those kinds where applicable.

akhavr:
> I think the problem reflects a more general issue: participants' interactions (discussion) around the source code are sometimes as meaningful as the source code. The existing handling of Nostr events does not allow for preserving them at the same level a reliable git server maintains source code files. 

Shouldn't then development metadata (issues, discussions, PR's, etc) be stored together with the git repo itself?

OpenSourceOptimist:
@garyray-k 
> perhaps larger patches could be serialized into a thread of events.

Worried about reliability issues if you need multiple events to know the proposed change? Would also not be less data to store for relay operators.  Sound like a question for someone how have though about this stuff a lot like @mikedilger :sweat_smile: 

> [...] be its own kind of event. That way the content of the event is just the patch text.

That sounds simple, I like simple :+1: 

> Then clients can easily filter out those kinds where applicable.

That is good point, always good to have a minimal opt out for something that might annoy folks.

> https://alblue.bandlem.com/2011/12/git-tip-of-week-patches-by-email.html

Learned a lot about patches from that post, thanks! There is good tooling around sending patches over email. My problem is all my experience is working with remote git servers, but we should probably have a flow as similar as possible as the existing email ones, at least under the hood. :thinking:

ofer-elrom:
> Shouldn't then development metadata (issues, discussions, PR's, etc) be stored together with the git repo itself?

It could be, but this would be the responsibility of the solution built on top of the protocol. It can not be implemented by the protocol, itself, and this NIP deals with the protocol.

akhavr:
> > Shouldn't then development metadata (issues, discussions, PR's, etc) be stored together with the git repo itself?
> 
> It could be, but this would be the responsibility of the solution built on top of the protocol. It can not be implemented by the protocol, itself, and this NIP deals with the protocol.

My point is that I don't see the necessity to amend protocol to implement source code control system with nostr.  Will be playing with my idea over couple weekends before showing the code, though.

ofer-elrom:
> My point is that I don't see the necessity to amend protocol to implement source code control system with nostr. Will be playing with my idea over couple weekends before showing the code, though.

I think the main difference between adjusting the protocol or not would be how much various implementations of git on Nostr would be compatible with each other. 
If you are doing your implementation, I would be happy if you look at this small google doc describing how I think Nost can allow for censorship-resistant git-based open-source development:
https://docs.google.com/document/d/1z96XpTtrUvrFZYPjMzlT5JuE41HF6X5vI99lvImVCvU/edit?usp=share_link

PeerRich:
btw y'all

we kicked off: https://github.com/NostrGit/NostrGit 

demo: nostrgit.com which is in very early development but contributions coming in quickly! 

__

we'll try to do things without changing much of the current nostr protocol

fiatjaf:
My comments here: https://gateway.nostr.com/e/048f5e74bbb7e6b8019655ae45a42e93e4b028be31ebcf5aadcea6028c4fd3ec

> What I would want to see are multiple apps that can interoperate and are able to perform separate functions:
> - browse code
> - comment on code (referenced by a commit)
> - create issues and comment on issues
> - send patches
> - comment on patches
> 
> And how these should be done? I am not sure, but here's what I have in mind:
> - most of the comment things should probably be kind:1 events, I don't know, with some extra tags (so they could be interacted with from the normal "social" Nostr clients? or not?)
> - code should probably be hosted by standalone dedicated git servers -- and there could be centralized providers offering these services but they should interoperate seamlessly between themselves and with standalone personal servers
> - sending patches should probably be done using something like this approach by #[0]: http://git.jb55.com/git-nostr-tools/file/README.txt.html
> 
> #[2] has opened a discussion on this topic on the NIPs repository that could possibly be used to coordinate the efforts: https://github.com/nostr-protocol/nips/pull/223
> 
> I think we could have multiple different smallish webapps, native apps and specially command-line tools that implement one or multiple of the separate functions described above, and with that we can achieve a much better result both in terms of quality and of decentralization than if someone or some big team decides to tackle the entire cake and come up with some centralizing architecture on their own.

KaffinPX:
Hey everyone, im creating a new NIP like this one, we CANT implement an exact git over Nostr because of commit chains because it needs order which cant be possible on an async network.

spearson78:
I created a proof of concept git integration for nostr.

You can use it to create a git repo server that takes its config from nostr.

By issuing commands to a relay you can trigger the creation of git repositories and manage the permissions of these repositories. Authentication is handled by associating your nostr profile with an SSH key.

A quick demo video is here https://www.youtube.com/watch?v=G-WzlC8XfW4

The source code and documentation is here. https://github.com/spearson78/gitnostr

ofer-elrom:
> I created a proof of concept git integration for nostr.
> 
> You can use it to create a git repo server that takes its config from nostr.
> 
> By issuing commands to a relay you can trigger the creation of git repositories and manage the permissions of these repositories. Authentication is handled by associating your nostr profile with an SSH key.

The project mentioned in the link seem to be moving forward with similar concepts in mind. I think it may be good if you joined:
https://github.com/nostr-protocol/nips/pull/223#issuecomment-1456164803

PeerRich:
> I created a proof of concept git integration for nostr.
> 
> You can use it to create a git repo server that takes its config from nostr.
> 
> By issuing commands to a relay you can trigger the creation of git repositories and manage the permissions of these repositories. Authentication is handled by associating your nostr profile with an SSH key.
> 
> A quick demo video is here [youtube.com/watch?v=G-WzlC8XfW4](https://www.youtube.com/watch?v=G-WzlC8XfW4)
> 
> The source code and documentation is here. [spearson78/gitnostr](https://github.com/spearson78/gitnostr)

this is really cool @spearson78, should we team up and join forces with nostrgit.com? we're currently discussing backend and CLI

KaffinPX:
Everyone check #358

npub1zenn0:
Doesn't a patch based workflow work better over nostr? Please correct me if I'm wrong.

Sending patches is just sending events. So Gerrit over Github? Essentially the email lists based flow except not email. Just a matter of directing Gerrit backend to look at Nostr for new content?

fiatjaf:
Yes, @npub1zenn0, I think you are right.

jb55:
PRs are still useful though. The PRs over nostr demo I was tinkering with looked similar to this one:

```
./target/debug/git-nostr-pr --base master --upstream jb55@jb55.com/git-nostr-tools --origin git://jb55.com/git-nostr-tools --title "my pr" | jq .
event sent to ws://127.0.0.1:8080
{
  "content": "",
  "created_at": 1681609363,
  "id": "7a3f6d933a7e9c394b37831f1c14c0c9c5ce6add8ada79c45e94b26e8cf5e988",
  "kind": 1969,
  "pubkey": "589b887d77f39d0d24f65fb800097b970adbf75b6315d2fee8822cba9b4cf5d8",
  "sig": "6d18a2327f0c0aabedc86d733a72d94e7383b7765350cca491045266279bdbbad26d4397e35c99fcfdb5d3055627c5e75b6c766749217908cd53e383d4a7344e",
  "tags": [
    [
      "base",
      "7fcc945b352591f0a793c10af52a32ef65392960"
    ],
    [
      "origin",
      "git://jb55.com/git-nostr-tools"
    ],
    [
      "branch",
      "03471cd1237a7e8ae76fc4050e46161fc102709f",
      "rust"
    ],
    [
      "subject",
      "my pr"
    ],
    [
      "t",
      "git-nostr-tools"
    ],
    [
      "p",
      "32e1827635450ebb3c5a7d12c1f8e7b2b514439ac10a67eef3d9fd9c5c68e245"
    ]
  ]
}
```

OpenSourceOptimist:
@npub1zenn0 

> Doesn't a patch based workflow work better over nostr?

Patches fit very well with nostr and I think we should support it for several reasons (see https://github.com/nostr-protocol/nips/pull/223#issuecomment-1454794294).

On the other hand PRs is the dominant way people interact with GIT these days. PRs are just like @jb55 says still useful. But most of all they are familiar. 

We might be best of supporting both references to commits on git servers and patches in the protocol but having in mind that it should be easy for app developers to present everything as PRs because that is the most familiar.

npub1zenn0:
Right, pull requests are indeed the de facto way at this time. However, making Gerrit work over nostr should be the low hanging fruit. I would like to guess it possible to have an MVP Gerrit instance open with a separate micro service that looks at nostr and posts into the Gerrit instance as admin, in reasonably short time.

---

And on this NIP: Git LFS was brought up https://github.com/npub1zenn0/go-git-nostr/issues/1 . There should be a `tag` for including lfs pointers. Maybe something like `["lfs", "filename", <pointer_to_NIP_81/94_event>]`.

gedw99:
Thanks @npub1zenn0 

yes a pointer is exactly what’s needed

fiatjaf:
What is the difference between a PR and a patch? To send a PR you must have a git server to point to, but to send a patch you don't need that?

In terms of UX both should be equivalent, no? You send the patch and people can comment on the patch and ultimately opt to "merge" the patch or not.

npub1zenn0:
I make the distinction in that a patch is something you overwrite (mimicking force push) if you change it, while a PR is something with a branch that you push new commits onto.

jb55:
> What is the difference between a PR and a patch? To send a PR you must have a git server to point to, but to send a patch you don't need that?

patches are just a textual diff to a file, and applying patches has no link to the previous history. In certain contexts merging a branch is the only allowed workflow to preserve history and auditability. bitcoin-core works this way. Also sometimes you need to merge large chunks of code at once. Linux/linus himself uses PRs to merge large subtrees of changes, whereas maintainers collect patches into their branches for Linus to merge. They serve two different purposes, I don't think there should only be patches on nostr, since it doesn't cover the uses cases of merging large subtrees and preserving commit history.

gedw99:
Great conversation 

https://github.com/icyphox/legit/blob/master/go.mod#L7

this might help !! 

The way it does git diffs and merge is clean, and it just happens to be a git server and gui in 100% golang .

It’s highly stable. I have been using this for a few months.

DanConwayDev:
Excited to see such an active discussion. This is slightly alternative architecture and how a NIP like this could fit:

### blob storage and retrieval 

There are different viable models to store, serve and retrieve data in a way that present like a standard git server/client interaction. This may be using a standard git server on a niche relay with a nostr permissions wrapper, or using a client to mimic this delivery whilst abstracting an alternative model under the hood. There needs to be experimentation because getting the incentives right for actors to serve the underlying data is challenging given that storage and bandwidth is expensive.

Interactions with git servers and relays should be trust minimised. Nostr events following a NIP standard should prove every change made is authorised by an maintainer. Other clients and relays that want to sync with the git server which was pushed to must be able to pull the update from the git server, verify it using the nostr event before writing it. Clients cloning a repo should be able to use available events to validate they have the complete and unadulterated repository as of the date of the last Nostr event they received.

Some models for delivering git data might use Nostr events to store git metadata and store blobs using torrents. Some might use a standard git server. Both would support the above NIP and would only accept write changes that have been signed. Bridges could enable repositories to be stored using multiple models. This would reduce entry barriers repository maintainers wouldn't have to choose one model.

### Collaborative Software Development Experience

A vibrant array of clients, serving different aspects of the software development collaboration experience will emerge. micro-apps might address a single feature and other clients might support many: Kamban boards, issue tracking, comments on lines of code, PR review, etc.

Clients would utilise anchor points with a standardised referencing schema outlined in a NIP for elements both inside a git repository (lines of code, commits, etc) and outside the git repository (PRs, issues, kamban boards, etc).

These NIP(s) would facilitate innovation at the storage and retrieval layer and the collaboration experience layer to proceed independently.

Any thoughts?

Note: I just published a NIP-23 article entitled [The Bull Case for a GitHub Alternative on Nostr](https://blogstack.io/naddr1qqyr2wt9xfsk2wp3qy28wumn8ghj7un9d3shjtnyv9kh2uewd9hsyg9qpr00z4uklw56p4h6kp8gl4ts3y59m874qhd94ql732k40g6kf5psgqqqw4rsd8v7ng), which might add some context.

gedw99:
Start with a Unified Diif capability that is the core of what git does  , so that just like git the client can diff and send in patches and the relay can apply those diffs ?

gedw99:
> @gedw99, if we work purely off a patch model:
> 
> 1. We wouldn't be able to carry out maintenance involving [rewriting history](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History).
> 2. It would be efficient from a storage and bandwidth point of view. Particularly for the data relays, for which we might have an incentive challenge.
> 3. Users would have to change their normal workflow which creates additional switching barriers.
> 
> I wrote a follow up article entitled [Key Challenges for a GitHub Alternative on Nostr](https://blogstack.io/naddr1qqyxxdmyvsmrqdn9qywhwumn8ghj7mn0wd68yttsw43zuam9d3kx7unyv4ezumn9wspzpgqgmmc409hm4xsdd74sf68a2uyf9pwel4g9mfdg8l5244t6x4jdqvzqqqr4guzzjrgj). In it I suggest standardising the 'anchors' between the social experience and elements in a repository (eg. commits or line of code) as the first standard to agree in this usecase. This could be a soft-standard or a NIP. Doing this would enable developers working on the social experience and those working on the git repository side, something to work off.

thanks @DanConwayDev    just read the blog and i see the history rewrite aspects and time ordering. very well written. Ordering of mutations in an offline first collaborative system is an age old problem  ( CRDT, etc etc ).  If we can get this right, then rewrite history with UDIff presents a archetype for all collaborative systems with a git baed file system ?

DanConwayDev:
@gedw99 thanks for you feedback.

> Ordering of mutations in an offline first collaborative system is an age old problem ( CRDT, etc etc ). 

We could follow the model I outlined in 'The Conflicting Change Problem' section of the article I referenced above, with each 'Authorisation Event' referencing the previous mutation it was building upon. Do you think this model is workable? It is dependent on solving the 'Group Authorisation Problem' which I hope to be publish an outline solution to soon.

> If we can get this right, then rewrite history with UDIff presents a archetype for all collaborative systems with a git based file system ?

We could establish a standard for a new type of patch that supports all possible mutations of the repository. This could be based on the existing git patching model and UDiff. a client wrapper could simulate pushing and pulling from a remote git server whilst operating entirely through patches.

I still feel patching is an inefficient model and the state reversion problem is better solved in a more elegant way. For example, Authorisation Events compactly including diffs in git/refs. The beauty of a standard for Authorisation Events is that the patch model, as well as other models could be built out and maintain compatibility as I discussed in the 'The Interoprablilty Problem' section of the article.

gedw99:
> @gedw99 thanks for you feedback.
> 
> > Ordering of mutations in an offline first collaborative system is an age old problem ( CRDT, etc etc ).
> 
> We could follow the model I outlined in 'The Conflicting Change Problem' section of the article I referenced above, with each 'Authorisation Event' referencing the previous mutation it was building upon. Do you think this model is workable? It is dependent on solving the 'Group Authorisation Problem' which I hope to be publish an outline solution to soon.
> 
> > If we can get this right, then rewrite history with UDIff presents a archetype for all collaborative systems with a git based file system ?
> 
> We could establish a standard for a new type of patch that supports all possible mutations of the repository. This could be based on the existing git patching model and UDiff. a client wrapper could simulate pushing and pulling from a remote git server whilst operating entirely through patches.
> 
> I still feel patching is an inefficient model and the state reversion problem is better solved in a more elegant way. For example, Authorisation Events compactly including diffs in git/refs. The beauty of a standard for Authorisation Events is that the patch model, as well as other models could be built out and maintain compatibility as I discussed in the 'The Interoprablilty Problem' section of the article.

Hey @DanConwayDev 

Can i answer in general for now, as i read your blog article and the other things and it makes sense but the jargon is a little vague. 

If we don't allow for offline mutations, which we don't AFAIK. Then it is a matter of ensuring all mutations are globally ordered, just like any async message broker architecture  The tricks i know to do that are:
- time smeer clock to avoid leap year hell. AWS and google run these in there DC, and there is golang implementation.  
- time ordered rewrite buffer. Because there is a delay between a mutation send and arriving you have an in memory buff on the server that collects them and reorders them before letting them through to the Server proper. 

Then your all set.. You can also then run subscription queries, where a query for data causes the actor to be added to the presence map and so then when any data changes on the server / relay all actors get a CDC like CUD event, which they use to update their local copy of the data. Very standard stuff for this area. Why do this ? because then the clients never have stall data, so when they later send a mutation, it's a mutation based on up to date data, and so you massively reduce the number of validation through backs from the server.   

I guess you can think of this as a best practice way of ensuring linearizability, but not inside a DB but outside one. The old "Turning the database inside out" by Martin Kleppmann ... 

Web Clients - You can do this inside anything. These days a Service worker PWA for example. So this model support web apps too.  Indexed DB limits you to 50 MB non deterministically, so some cache eviction strategies are needed. I only say this because i see most Nostre use cases with their only being a Relay, and then Clients. Some Use cases however might be a proper durable server being fed off the relay too and then clients talking t the Server ? don't know....

I hate being so general about he patterns, but again its seems like Nostr is really imitating this classic pattern above.  Hope this is helpful in some way ??

DanConwayDev:
@gedw99 Thanks for taking the time; that was helpful. I thought it best to flesh out my proposition a bit more to allow for a more specific critique.  My proposal for authorisation groups is outlined in this [NIP draft](https://github.com/nostr-protocol/nips/pull/483).

I have traded-off simplicity of design and implementation against speed of state change confirmation. The needs here are medium-low frequency of changes so I don't think change conflicts are going to be a regular occurrence. Most users interact asynchronously at low-frequencies with git services. CI/ID and webhooks less so...

gedw99:
hey @DanConwayDev I

 see what you mean about speed of changes being low and so mutations very likely to be ordered and life is good then. Its a good solution and i agree it's the least worst path forward. Its always "what is the least worst solution" in software and engineering :)  

CRDT etc is YAGNI if this works.

thanks for taking what i said into consideration.

DanConwayDev:
https://ngit.netlify.app/
https://github.com/DanConwayDev/ngit-cli

I wrote a prototype under the name ngit that builds on the patch model but brings it closer to PR workflow.

```pull```, ```push``` and even ```clone``` and ```merge``` are achieved by applying patches issued as nostr events.

There are events for repos, branches, patches, PRs, merges and groups.

A patch is always applied to one 'branch'. A 'branch' isn't reissued to refer to the latest commit like in git. Instead it is just an id linked to a repo that patches can be associated with.

Merges indicate that a patch from one branch is applied to another branch.

PRs indicate a desire to merge one branch into another. Comments and further patches can applied to a branch as the conversation around the an associated PR continues. 

Groups manage the permissions on branches so that maintainers and authors can add patches to PR branches and only maintainers can merge or commit directly to the main branch. This partially implements the [Nostr Groups draft NIP]( https://github.com/nostr-protocol/nips/pull/483).

The patch application process is adjusted so that commit ids are preserved which addresses the history and audibility needs of projects like bitcoin core.

lonerodriguez:
> All that said, it was a fun experience, I learnt a tonne of git internals along the way so it wasn't a complete waste of time but I think it might be time I hang up the boots on this one. If anyone want's to discuss how a git-server model may look, I would be up for discussing / working on something.

hello hendore, is there a possibility to try out/download your client? I think the presentation is pretty good and I would like to work on it further.

PR #224, opened 2023-02-05
NIP-57: Lightning Zaps

jb55:
This NIP defines a new note type called a lightning zap of kind `9735`. These represent paid lightning invoice receipts sent by a lightning node called the `zapper`. We also define a another note type of kind `9734` which are `zap request` notes, which will be described in this document.

Having lightning receipts on nostr allows clients to display lightning payments from entities on the network. These can be used for fun or for spam deterrence.

https://github.com/nostr-protocol/nips/blob/zaps/57.md

### DIFF
--- a/57.md
+++ b/57.md
@@ -30,7 +30,7 @@ Having lightning receipts on nostr allows clients to display lightning payments 
 
 3. Clients may choose to display a lightning zap button on each post or on the users profile, if the user's lnurl pay request endpoint supports nostr, the client SHOULD generate a `zap invoice` instead of a normal lnurl invoice.
 
-4. To generate a `zap invoice`, call the `callback` url with `amount` set to the milli-satoshi amount value. A `nostr` querystring value MUST be set as well. It a uri-encoded `zap request` note signed by the user's key. The `zap request` note contains an `e` tag of the note it is zapping, and a `p` tag of the target user's pubkey. The `e` tag is optional which allows profile tipping. The `zap request` note must also have a `relays` tag, which is gathered from the user's configured relays.
+4. To generate a `zap invoice`, call the `callback` url with `amount` set to the milli-satoshi amount value. A `nostr` querystring value MUST be set as well. It is a uri-encoded `zap request` note signed by the user's key. The `zap request` note contains an `e` tag of the note it is zapping, and a `p` tag of the target user's pubkey. The `e` tag is optional which allows profile tipping. The `zap request` note must also have a `relays` tag, which is gathered from the user's configured relays.
 
 5. Pay this invoice or pass it to an app that can pay the invoice. Once it's paid, a `zap note` will be created by the `zapper`.
 
--- a/57.md
+++ b/57.md
@@ -133,10 +133,11 @@ A reference implementation for the zapper is here: [zapper][zapper]
 
 Clients MAY fetch zap notes on posts and profiles:
 
-`{"kinds": [9735], "#p": [...]}`
+`{"kinds": [9735], "#e": [...]}`
 
 To authorize these notes, clients MUST fetch the `nostrPubkey` from the users configured lightning address or lnurl and ensure that the zaps to their posts were created by this pubkey. If clients don't do this, anyone could forge unauthorized zaps.
 
+Once authorized, clients MAY tally zaps on posts, and list them on profiles. If the zap request note contains a non-empty `content`, it may display a zap comment. Generally clients should show users the `zap request` note, and use the `zap note` to show "zap authorized by ..." but this is optional.
 
 ## Future Work
 
--- a/57.md
+++ b/57.md
@@ -124,6 +124,8 @@ When receiving a payment, the following steps are executed:
 
 6. Create a nostr note of kind `9735` that includes the `p` tag AND optional `e` tag. The content SHOULD be empty. The created_at date SHOULD be set to the invoice paid_at date for idempotency.
 
+7. Send the note to the `relays` declared in the `zap request` note from the invoice description.
+
 A reference implementation for the zapper is here: [zapper][zapper]
 
 [zapper]: https://github.com/jb55/cln-nostr-zapper
--- a/57.md
+++ b/57.md
@@ -30,7 +30,7 @@ Having lightning receipts on nostr allows clients to display lightning payments 
 
 3. Clients may choose to display a lightning zap button on each post or on the users profile, if the user's lnurl pay request endpoint supports nostr, the client SHOULD generate a `zap invoice` instead of a normal lnurl invoice.
 
-4. To generate a `zap invoice`, call the `callback` url with `amount` set to the milli-satoshi amount value. A `nostr` querystring value MUST be set as well. It is a uri-encoded `zap request` note signed by the user's key. The `zap request` note contains an `e` tag of the note it is zapping, and a `p` tag of the target user's pubkey. The `e` tag is optional which allows profile tipping. The `zap request` note must also have a `relays` tag, which is gathered from the user's configured relays.
+4. To generate a `zap invoice`, call the `callback` url with `amount` set to the milli-satoshi amount value. A `nostr` querystring value MUST be set as well. It is a uri-encoded `zap request` note signed by the user's key. The `zap request` note contains an `e` tag of the note it is zapping, and a `p` tag of the target user's pubkey. The `e` tag is optional which allows profile tipping. The `zap request` note must also have a `relays` tag, which is gathered from the user's configured relays. The `content` MAY be an additional comment from the user which can be displayed when listing zaps on posts and profiles.
 
 5. Pay this invoice or pass it to an app that can pay the invoice. Once it's paid, a `zap note` will be created by the `zapper`.
 
--- a/57.md
+++ b/57.md
@@ -56,7 +56,7 @@ The lnurl server will need some additional pieces of information so that clients
 
 	e. There should be a `relays` tag with the relays to send the `zap` note to.
 
-5. If valid, fetch an invoice where the description is this note and this note only. No additional lnurl metadata is included in the description.
+5. If valid, fetch a description hash invoice where the description is this note and this note only. No additional lnurl metadata is included in the description.
 
 At this point, the lightning node is ready to send the zap note once payment is received.
 
--- a/57.md
+++ b/57.md
@@ -26,7 +26,7 @@ Having lightning receipts on nostr allows clients to display lightning payments 
 
 1. Calculate the lnurl pay request url for a user from the lud06 or lud16 field on their profile
 
-2. Fetch the lnurl pay request static endpoint (`https://host.com/.well-known/lnurlp/user`) and gather the `allowsNostr` and `nostrPubkey` fields. If `allowsNostr` exists and it is `true`, and if `nostrPubkey` exists and is true, associate this information with the user. The `nostrPubkey` is the `zapper`'s pubkey, and it is used to authorize zaps sent to that user.
+2. Fetch the lnurl pay request static endpoint (`https://host.com/.well-known/lnurlp/user`) and gather the `allowsNostr` and `nostrPubkey` fields. If `allowsNostr` exists and it is `true`, and if `nostrPubkey` exists and is a valid BIP 340 public key, associate this information with the user. The `nostrPubkey` is the `zapper`'s pubkey, and it is used to authorize zaps sent to that user.
 
 3. Clients may choose to display a lightning zap button on each post or on the users profile, if the user's lnurl pay request endpoint supports nostr, the client SHOULD generate a `zap invoice` instead of a normal lnurl invoice.
 
--- a/57.md
+++ b/57.md
@@ -6,7 +6,7 @@ Lightning Zaps
 
 `draft` `optional` `author:jb55` `author:kieran`
 
-This NIP defines a new note type called a lightning zap of kind `9735`. These represent paid lightning invoice receipts sent by a lightning node called the `zapper`. We also define a another note type of kind `9734` which are `zap request` notes, which will be described in this document.
+This NIP defines a new note type called a lightning zap of kind `9735`. These represent paid lightning invoice receipts sent by a lightning node called the `zapper`. We also define another note type of kind `9734` which are `zap request` notes, which will be described in this document.
 
 Having lightning receipts on nostr allows clients to display lightning payments from entities on the network. These can be used for fun or for spam deterrence.
 
--- a/57.md
+++ b/57.md
@@ -1,0 +1,143 @@
+NIP-57
+======
+
+Lightning Zaps
+--------------
+
+`draft` `optional` `author:jb55` `author:kieran`
+
+This NIP defines a new note type called a lightning zap of kind `9735`. These represent paid lightning invoice receipts sent by a lightning node called the `zapper`. We also define a another note type of kind `9734` which are `zap request` notes, which will be described in this document.
+
+Having lightning receipts on nostr allows clients to display lightning payments from entities on the network. These can be used for fun or for spam deterrence.
+
+
+## Definitions
+
+`zapper` - the lightning node or service that sends zap notes (kind `9735`)
+
+`zap request` - a note of kind `9734` created by the person zapping
+
+`zap invoice` - the bolt11 invoice fetched from a custom lnurl endpoint which contains a `zap request` note
+
+
+## Protocol flow
+
+### Client side
+
+1. Calculate the lnurl pay request url for a user from the lud06 or lud16 field on their profile
+
+2. Fetch the lnurl pay request static endpoint (`https://host.com/.well-known/lnurlp/user`) and gather the `allowsNostr` and `nostrPubkey` fields. If `allowsNostr` exists and it is `true`, and if `nostrPubkey` exists and is true, associate this information with the user. The `nostrPubkey` is the `zapper`'s pubkey, and it is used to authorize zaps sent to that user.
+
+3. Clients may choose to display a lightning zap button on each post or on the users profile, if the user's lnurl pay request endpoint supports nostr, the client SHOULD generate a `zap invoice` instead of a normal lnurl invoice.
+
+4. To generate a `zap invoice`, call the `callback` url with `amount` set to the milli-satoshi amount value. A `nostr` querystring value MUST be set as well. It a uri-encoded `zap request` note signed by the user's key. The `zap request` note contains an `e` tag of the note it is zapping, and a `p` tag of the target user's pubkey. The `e` tag is optional which allows profile tipping. The `zap request` note must also have a `relays` tag, which is gathered from the user's configured relays.
+
+5. Pay this invoice or pass it to an app that can pay the invoice. Once it's paid, a `zap note` will be created by the `zapper`.
+
+### LNURL Server side
+
+The lnurl server will need some additional pieces of information so that clients can know that zap invoices are supported:
+
+1. Add a `nostrPubkey` to the lnurl-pay static endpoint `/.well-known/lnurlp/user`, where `nostrPubkey` is the nostr pubkey of the `zapper`, the entity that creates zap notes. Clients will use this to authorize zaps.
+
+2. Add an `allowsNostr` field and set it to true.
+
+3. In the lnurl-pay callback URL, watch for a `nostr` querystring, where the contents of the note is a uri-encoded `zap request` JSON.
+
+4. If present, the zap request note must be validated:
+
+	a. It MUST have a valid nostr signature
+
+	b. It MUST have tags
+
+	c. It MUST have at least one p-tag
+
+	d. It MUST have either 0 or 1 e-tag
+
+	e. There should be a `relays` tag with the relays to send the `zap` note to.
+
+5. If valid, fetch an invoice where the description is this note and this note only. No additional lnurl metadata is included in the description.
+
+At this point, the lightning node is ready to send the zap note once payment is received.
+
+## The zap note
+
+Zap notes are created by a lightning node reacting to paid invoices. Zap notes are only created when the invoice description (committed to the description hash) contains a `zap request` note.
+
+Example zap note:
+
+```json
+{
+    "id": "67b48a14fb66c60c8f9070bdeb37afdfcc3d08ad01989460448e4081eddda446",
+    "pubkey": "9630f464cca6a5147aa8a35f0bcdd3ce485324e732fd39e09233b1d848238f31",
+    "created_at": 1674164545,
+    "kind": 9735,
+    "tags": [
+      [
+        "p",
+        "32e1827635450ebb3c5a7d12c1f8e7b2b514439ac10a67eef3d9fd9c5c68e245"
+      ],
+      [
+        "e",
+        "3624762a1274dd9636e0c552b53086d70bc88c165bc4dc0f9e836a1eaf86c3b8"
+      ],
+      [
+        "bolt11",
+        "lnbc10u1p3unwfusp5t9r3yymhpfqculx78u027lxspgxcr2n2987mx2j55nnfs95nxnzqpp5jmrh92pfld78spqs78v9euf2385t83uvpwk9ldrlvf6ch7tpascqhp5zvkrmemgth3tufcvflmzjzfvjt023nazlhljz2n9hattj4f8jq8qxqyjw5qcqpjrzjqtc4fc44feggv7065fqe5m4ytjarg3repr5j9el35xhmtfexc42yczarjuqqfzqqqqqqqqlgqqqqqqgq9q9qxpqysgq079nkq507a5tw7xgttmj4u990j7wfggtrasah5gd4ywfr2pjcn29383tphp4t48gquelz9z78p4cq7ml3nrrphw5w6eckhjwmhezhnqpy6gyf0"
+      ],
+      [
+        "description",
+        "{\"pubkey\":\"32e1827635450ebb3c5a7d12c1f8e7b2b514439ac10a67eef3d9fd9c5c68e245\",\"content\":\"\",\"id\":\"d9cc14d50fcb8c27539aacf776882942c1a11ea4472f8cdec1dea82fab66279d\",\"created_at\":1674164539,\"sig\":\"77127f636577e9029276be060332ea565deaf89ff215a494ccff16ae3f757065e2bc59b2e8c113dd407917a010b3abd36c8d7ad84c0e3ab7dab3a0b0caa9835d\",\"kind\":9734,\"tags\":[[\"e\",\"3624762a1274dd9636e0c552b53086d70bc88c165bc4dc0f9e836a1eaf86c3b8\"],[\"p\",\"32e1827635450ebb3c5a7d12c1f8e7b2b514439ac10a67eef3d9fd9c5c68e245\"],[\"relays\",\"wss://relay.damus.io\",\"wss://nostr-relay.wlvs.space\",\"wss://nostr.fmt.wiz.biz\",\"wss://relay.nostr.bg\",\"wss://nostr.oxtr.dev\",\"wss://nostr.v0l.io\",\"wss://brb.io\",\"wss://nostr.bitcoiner.social\",\"ws://monad.jb55.com:8080\",\"wss://relay.snort.social\"]]}"
+      ],
+      [
+        "preimage",
+        "5d006d2cf1e73c7148e7519a4c68adc81642ce0e25a432b2434c99f97344c15f"
+      ]
+    ],
+    "content": "",
+    "sig": "b0a3c5c984ceb777ac455b2f659505df51585d5fd97a0ec1fdb5f3347d392080d4b420240434a3afd909207195dac1e2f7e3df26ba862a45afd8bfe101c2b1cc"
+  }
+```
+
+* The zap note MUST have a `bolt11` tag containing the description hash bolt11 invoice.
+
+* The zap note MUST contain a `description` tag which is the invoice description.
+
+* `SHA256(description)` MUST match the description hash in the bolt11 invoice.
+
+* The zap note MAY contain a `preimage` to match against the payment hash of the bolt11 invoice. This isn't really a payment proof, there is no real way to prove that the invoice is real or has been paid. You are trusting the author of the zap note for the legitimacy of the payment.
+
+The zap note is not a proof of payment, all it proves is that some nostr user fetched an invoice. The existence of the zap note implies the invoice as paid, but it could be a lie given a rogue implementation.
+
+
+### Creating a zap note
+
+When receiving a payment, the following steps are executed:
+
+1. Get the description for the invoice. This needs to be saved somewhere during the generation of the description hash invoice. It is saved automatically for you with CLN, which is the reference implementation used here.
+
+2. Parse the bolt11 description as a JSON nostr note. You SHOULD check the signature of the parsed note to ensure that it is valid. This is the `zap request` note created by the entity who is zapping.
+
+4. The note MUST have only one `p` tag
+
+5. The note MUST have 0 or 1 `e` tag
+
+6. Create a nostr note of kind `9735` that includes the `p` tag AND optional `e` tag. The content SHOULD be empty. The created_at date SHOULD be set to the invoice paid_at date for idempotency.
+
+A reference implementation for the zapper is here: [zapper][zapper]
+
+[zapper]: https://github.com/jb55/cln-nostr-zapper
+
+
+## Client Behavior
+
+Clients MAY fetch zap notes on posts and profiles:
+
+`{"kinds": [9735], "#p": [...]}`
+
+To authorize these notes, clients MUST fetch the `nostrPubkey` from the users configured lightning address or lnurl and ensure that the zaps to their posts were created by this pubkey. If clients don't do this, anyone could forge unauthorized zaps.
+
+
+## Future Work
+
+Zaps can be extended to be more private by encrypting zap request notes to the target user, but for simplicity it has been left out of this initial draft.
--- a/README.md
+++ b/README.md
@@ -30,6 +30,7 @@ NIPs stand for **Nostr Implementation Possibilities**. They exist to document wh
 - [NIP-40: Expiration Timestamp](40.md)
 - [NIP-42: Authentication of clients to relays](42.md)
 - [NIP-50: Keywords filter](50.md)
+- [NIP-57: Lightning Zaps](57.md)
 
 ## Event Kinds
 


57.md
```diff
@@ -0,0 +1,143 @@
+NIP-57
+======
+
+Lightning Zaps
+--------------
+
+`draft` `optional` `author:jb55` `author:kieran`
+
+This NIP defines a new note type called a lightning zap of kind `9735`. These represent paid lightning invoice receipts sent by a lightning node called the `zapper`. We also define a another note type of kind `9734` which are `zap request` notes, which will be described in this document.
```
0xtrr, line 9:
```suggestion
This NIP defines a new note type called a lightning zap of kind `9735`. These represent paid lightning invoice receipts sent by a lightning node called the `zapper`. We also define another note type of kind `9734` which are `zap request` notes, which will be described in this document.
```

57.md
```diff
@@ -0,0 +1,144 @@
...
+
+1. Calculate the lnurl pay request url for a user from the lud06 or lud16 field on their profile
+
+2. Fetch the lnurl pay request static endpoint (`https://host.com/.well-known/lnurlp/user`) and gather the `allowsNostr` and `nostrPubkey` fields. If `allowsNostr` exists and it is `true`, and if `nostrPubkey` exists and is true, associate this information with the user. The `nostrPubkey` is the `zapper`'s pubkey, and it is used to authorize zaps sent to that user.
+
+3. Clients may choose to display a lightning zap button on each post or on the users profile, if the user's lnurl pay request endpoint supports nostr, the client SHOULD generate a `zap invoice` instead of a normal lnurl invoice.
+
+4. To generate a `zap invoice`, call the `callback` url with `amount` set to the milli-satoshi amount value. A `nostr` querystring value MUST be set as well. It a uri-encoded `zap request` note signed by the user's key. The `zap request` note contains an `e` tag of the note it is zapping, and a `p` tag of the target user's pubkey. The `e` tag is optional which allows profile tipping. The `zap request` note must also have a `relays` tag, which is gathered from the user's configured relays.
```
benthecarman, line 33: > It a uri-encoded `zap request`

It `is` a uri-encoded zap request

lylepratt:
nice! awesome spec! Will work on implementing this.

fiatjaf:
Can we remove the `allowsNostr` field entirely and just rely on the presence of `nostrPubkey`?

jb55:
probably, I can't remember why I had them separate.

jb55:
I kinda rushed this and might be underspecified, let me know if anything seems unclear.

jb55:
> probably, I can't remember why I had them separate.

maybe for temporarily disabling it while still allowing clients to validate older zaps? I think I had some kind of reasoning along those lines.

fiatjaf:
> maybe for temporarily disabling it while still allowing clients to validate older zaps? I think I had some kind of reasoning along those lines.

Makes sense.

fiatjaf:
I don't understand the need for the encrypted zap request, since HTTPS calls are already encrypted. Did I misunderstood something?

jb55:
This would allow you to hide who created the invoice. you could still tally zaps and only the person receiving the zap could see who it was from. The description would be an encrypted `zap request` note.

jb55:
Well it would be slightly more complicated, it would have to be a kind 9733 note with relays encrypted to the zapper (to reduce fingerprinting), and the content is an encrypted 9734 note encrypted to the user getting zapped.

I have this all thought out but didn't want to complicate this spec with it just yet.

57.md
```diff
@@ -0,0 +1,146 @@
...
+
+
+## Protocol flow
+
+### Client side
+
+1. Calculate the lnurl pay request url for a user from the lud06 or lud16 field on their profile
+
+2. Fetch the lnurl pay request static endpoint (`https://host.com/.well-known/lnurlp/user`) and gather the `allowsNostr` and `nostrPubkey` fields. If `allowsNostr` exists and it is `true`, and if `nostrPubkey` exists and is true, associate this information with the user. The `nostrPubkey` is the `zapper`'s pubkey, and it is used to authorize zaps sent to that user.
+
```
benthecarman, line 30: What is the point of `allowsNostr`, what difference does it server vs just having `nostrPubkey` defined or not

jb55: https://github.com/nostr-protocol/nips/pull/224#issuecomment-1418256137

benthecarman: also:

> `nostrPubkey` exists and is true

should probably be ``nostrPubkey` exists and is a valid BIP 340 public key

jb55: thanks!

57.md
```diff
@@ -0,0 +1,146 @@
...
+`zap invoice` - the bolt11 invoice fetched from a custom lnurl endpoint which contains a `zap request` note
+
+
+## Protocol flow
+
+### Client side
+
+1. Calculate the lnurl pay request url for a user from the lud06 or lud16 field on their profile
+
+2. Fetch the lnurl pay request static endpoint (`https://host.com/.well-known/lnurlp/user`) and gather the `allowsNostr` and `nostrPubkey` fields. If `allowsNostr` exists and it is `true`, and if `nostrPubkey` exists and is a valid BIP 340 public key, associate this information with the user. The `nostrPubkey` is the `zapper`'s pubkey, and it is used to authorize zaps sent to that user.
```
fiatjaf, line 29:
```suggestion
2. Fetch the lnurl pay request static endpoint (`https://host.com/.well-known/lnurlp/user`) and gather the `allowsNostr` and `nostrPubkey` fields. If `allowsNostr` exists and it is `true`, and if `nostrPubkey` exists and is a valid BIP 340 public key, 32 bytes, hex-encoded, associate this information with the user. The `nostrPubkey` is the `zapper`'s pubkey, and it is used to authorize zaps sent to that user.
```

bumi:
I am excited to see the interest of deeply integrating lightning payments into Nostr. The name zaps is great! 

Here are a few thoughts/questions: 

Do I understand it correctly that the zap request should be stored in the invoice description? Or only the hash of the zap request in the invoice description_hash? "Parse the bolt11 description as a JSON nostr note" indicates that the whole nostr node should be saved in the invoice description. 
This would be different to what is currently described in the lnurl-pay spec? 

Would it be possible to reuse LUD18 payerdata for this? afaik LUD18 is already generic and allows the storage generic payer data. It is also committed in the description_hash.


Would it be possible to allow the client to publish the zap? This would remove the requirement that LNURL implementations/service providers need to implement Nostr connections. The smaller the usage specific requirement is the more tools/service providers will be able to be used. I am worried that this is a too big limitation. IMO ideally the LNURL spec should not have any social network specific requirements and is independent. 

The main data that would be needed is the preimage, correct? With the preimage and the payment request/hash the validation can also be done. 
I could imagine that clients will deeply integrate lightning wallets and thus also have access to the preimage or alternatively with a proposal like the [lnnurl-verify](https://github.com/lnurl/luds/issues/182) the preimage could be retrieved without any Nostr specific requirement on side of the the LNURL-spec.

Thanks for pushing this forward! This can bring the [value4value](https://value4value.info/) idea to so many more usecases!

fiatjaf:
I was pretty happy with this NIP, but after thinking about it for many minutes. This question has occurred to me: is there a reason for reusing LNURL endpoints even though this is not an LNURL payment at all?

Wouldn't it be more elegant and much simpler to decouple this entirely from LNURL and use an independent spec (and a different `/.well-known` endpoint) for requesting the invoice? It wouldn't need all the other LNURL things, and it would be automatically a nostrified endpoint (so no need for the `nostrEnabled` flag).

Moreover the communication could be reduced to a single step. Could be something like:

A user sets a field in their Nostr metadata like: `"nip57": "https://zapper.com/"`

Then a client can call

```
GET https://zapper.com/.well-known/zap
```

and get back

```json
{"pubkey": "<hex pubkey of the zapper>"}
```

And when sending a zap the client would call:

```
POST /.well-known/zap
{
  ...zap request event, as it is today, but also including the amount as a tag
}
```

and immediately get an invoice back, check that the description hash matches against the event id of the `zap request`, and pay it. The rest is the same.

The only reason I can see for keeping the lud06/lud16 things in place and reusing them are that people don't have to change their metadata and can start getting zaps immediately as soon as their zapper provider (currently LNURL provider) implements this NIP. But I think this reason is not good enough and that a dedicated protocol that doesn't ever say anything about LNURL such as above would be less confusing and easier to implement for everybody.

fiatjaf:
Another idea is to embed the zapper pubkey in the metadata of the user who is receiving the zaps directly, so the field could be something like `"nip57": "["<zapperpubkeyhex>", https://zapper.com"]` instead. Then other clients don't have to actually call the zapper endpoint to verify the zaps, they can just use the zapper key taken from the metadata directly.

Now I am thinking: why restrict the number of zappers a user can have to only one? Maybe I am crazy, but users may want to have more than one, as fallbacks. So the field could be `"nip57": [["key", "url"], ["key", "url"]]`, but this is ugly, better put it on the event tags: `"tags": [["zapper", "key", "url"], ["zapper", "key", "url"]]`. I don't know.

fiatjaf:
The problem with my spec suggestion above is that it assumes the `zapper` service will know who is receiving the funds from the `zap request` event, which is definitely a possibility, but shouldn't be mandatory as it may complicate things for some potential providers.

It is probably better that the specified zapper URL be just an arbitrary URL and clients just POST to that URL directly, without the `.well-known` shenanigans. Using arbitrary URLs solve things since `zapper` services can give URLs with internal IDs for their users and use these to discriminate incoming zap payments.

bumi:
I think being able to build on all the adoption of LNURL/LUD16 out there is a great advantage. Imo ideally people can use their current setup as much as possible and developers also don't need to start from scratch again building nostr LN support. So I am +1 for LNURL

What exactly do we miss from LNURL?

starbackr-dev:
> Would it be possible to allow the client to publish the zap? This would remove the requirement that LNURL implementations/service providers need to implement Nostr connections. The smaller the usage specific requirement is the more tools/service providers will be able to be used. I am worried that this is a too big limitation. IMO ideally the LNURL spec should not have any social network specific requirements and is independent.
> 
> The main data that would be needed is the preimage, correct? With the preimage and the payment request/hash the validation can also be done. I could imagine that clients will deeply integrate lightning wallets and thus also have access to the preimage or alternatively with a proposal like the [lnnurl-verify](https://github.com/lnurl/luds/issues/182) the preimage could be retrieved without any Nostr specific requirement on side of the the LNURL-spec.
> 
> Thanks for pushing this forward! This can bring the [value4value](https://value4value.info/) idea to so many more usecases!

Would it be possible to allow the client to publish the zap?  - I like Bhumi's point on having this option. There are many wallets such as Strike don't have LNURL implemented. Zaps break if wallet does not support it. 

There should be a manual option for zapper user to input the preimage after the payment and the client creates kind 9735 event on behalf of the zapper user. This eliminates dependency on wallets implementing LNURL and Zapper specs.

lylepratt:
> Zaps break if wallet does not support it.

I don't think this is true. The Nostr client requests a Bolt11 invoice from the server which all wallets should be able to pay.

starbackr-dev:
> > Zaps break if wallet does not support it.
> 
> I don't think this is true. The Nostr client requests a Bolt11 invoice from the server which all wallets should be able to pay.

Sorry for not being clear. I was referring to Zapper node/wallet which needs to create note kind 9735.

fiatjaf:
> developers also don't need to start from scratch again building nostr LN support

This is not correct. What is "from scratch"? To implement zap support everybody has to very much start from scratch since there is nothing in common with LNURL here. This is not a bad thing since it is so simple and the functionality and purpose is so different anyway.

LNURL libraries cannot be used at all without heavy patching. LNURL providers will also require heavy patching (relatively heavy, of course, since everything is still simple) that is equivalent to implementing a handler for this from scratch -- probably worse.

Most likely any server patching will introduce spaghetti code for switching between normal lnurl and zaps, so it would be better to just have an HTTP handler dedicated to zaps instead. Much cleaner.

Other advantages of dropping LNURL:
1. doesn't confuse people into thinking this is LNURL in any sense
2. Nostr clients don't need to parse `lnurl...` bech32 codes
3. Nostr clients don't need to parse lud16 addresses
4. with a new single-step communication protocol I think bolt12 support can be introduced very easily if that becomes a thing, we will not need a new zap-bolt12 protocol or anything like that

For reference: https://github.com/nostr-protocol/nips/pull/224#issuecomment-1418302436

ok300:
Some of the issues raised here made me wonder if there is a better way to do zaps.

So I came up #228. The interesting part is:

> Benefits vs current approach:
> - no LNURL changes needed
> - minimal nostr client changes needed (send 1 new ephemeral msg kind; validate and tally zaps)
> - nostr clients don't need to parse BOLT11 invoices
> - nostr clients don't need to parse LUD06 or LUD16 to zap

It's still theoretical and early, but from what I can tell it solves or even avoids some of the main issues raised here.

ok300:
For this PR, I agree with @fiatjaf that a separate non-LNURL approach is better.

@bumi is right that LNURL is better because of the network effects. But this doesn't look like LNURL anymore. It needs an extra arg for LNURL-pay, extra validation, etc. It's not even a `successAction` extension like AES [^1].

One thing I disagree with is https://github.com/nostr-protocol/nips/pull/224/files#diff-503ae34c31b21445fb1624ece3f7924b5a0b7ab1ab31b19ea065fb7975b3f83bR31 :
> 3. Clients may choose to display a lightning zap button on each post or on the users profile, **if the user's lnurl pay request endpoint supports nostr, the client SHOULD generate a `zap invoice` instead of a normal lnurl invoice.**

Seems to indicate zaps "override" LNURL if a recipient supports both. I hope its a vestige of the original idea, where LNURL was planned for both. But if zaps use a non-LNURL way, the clients should still support standard LNURL. Could be offered as checkbox in zap popup, or as client setting.


[^1]: https://github.com/lnurl/luds/blob/luds/10.md

jb55:
> Another idea is to embed the zapper pubkey in the metadata of the user who is receiving the zaps directly, so the field could be something like `"nip57": "["<zapperpubkeyhex>", https://zapper.com"]` instead. Then other clients don't have to actually call the zapper endpoint to verify the zaps, they can just use the zapper key taken from the metadata directly.

How would clients set this key? The zapper pubkey is determined by the zapper's private key which is implemented by the wallet service.

jb55:
I still think it's worth making this an lnurl extension, users don't have to do anything and wallet providers can just implement zaps if their users demand it. seemed pretty natural to me 🤷‍♂️

jb55:
> I was pretty happy with this NIP, but after thinking about it for many minutes, this question has occurred to me: is there a reason for reusing LNURL endpoints even though this is not an LNURL payment at all?
> 
> Wouldn't it be more elegant and much simpler to decouple this entirely from LNURL and use an independent spec (and a different `/.well-known` endpoint) for requesting the invoice? It wouldn't need all the other LNURL things, and it would be automatically a nostrified endpoint (so no need for the `nostrEnabled` flag).
> 
> Moreover the communication could be reduced to a single step. Could be something like:
> 
> A user sets a field in their Nostr metadata like: `"nip57": "https://zapper.com/"`
> 
> Then a client can call
> 
> ```
> GET https://zapper.com/.well-known/zap
> ```
> 
> and get back
> 
> ```json
> {"pubkey": "<hex pubkey of the zapper>"}
> ```
> 
> And when sending a zap the client would call:
> 
> ```
> POST /.well-known/zap
> {
>   ...zap request event, as it is today, but also including the amount as a tag
> }
> ```
> 
> and immediately get an invoice back, check that the description hash matches against the event id of the `zap request`, and pay it. The rest is the same.
> 
> The only reason I can see for keeping the lud06/lud16 things in place and reusing them are that people don't have to change their metadata and can start getting zaps immediately as soon as their zapper provider (currently LNURL provider) implements this NIP. But I think this reason is not good enough and that a dedicated protocol that doesn't ever say anything about LNURL such as above would be less confusing and easier to implement for everybody.

I'm fine with this even though I think its busywork... what do you think @v0l ?

starbackr-dev:
Since all the proposed solutions involves some heavy lifting work on the wallet side as well, We should also hear from wallet providers such as WOS, breez, Zeus on supporting this. They need to be connected and publish the event to relays where the zapper sender and receiver are connected.

rolznz:
Please correct me if I'm wrong, but currently zaps are simply a way to show receipts of "paid upvotes" a post/profile in a way that adds more signal than empty likes/reposts (unless you are on a relay which requires payment per action).

I see there is a "content" field on the zap:

> The content MAY be an additional comment from the user which can be displayed when listing zaps on posts and profiles.

Will this create a weird scenario where there are two separate threads stored in two different locations (zaps with comments and standard replies) which could break conversation flow. Would another option for the person zapping to create a standard text_note, and then attach the zap to that rather than including text content inside the zap (I feel like this will be difficult to expand outside of simple comments)?

Then zaps would simply verify a payment of X amount on any type of note, And the note the zap is attached to can contain any type of metadata.

This could allow zaps to be used on all other types of content, such as replies to polls.

fiatjaf:
> How would clients set this key? The zapper pubkey is determined by the zapper's private key which is implemented by the wallet service.

They would get the data from the zapper's website beforehand. They could get it in a standard format (like `nzap1pubkeyurl...`) and the Nostr client would accept that and parse it, but this may be an overcomplication.

I think it would be good so that all other clients don't have to query the zapper HTTP endpoint all the time to get their public key.

I think the user could input just their zap address like `https://zapper.com/myuser` and the Nostr client queries that URL to get the public key immediately, then stores the public key in the user's metadata.

fiatjaf:
> I don't see what all the fuss is about, it's very simple already (i added Zaps support to strike.army in a day)

There is no fuss, it's just that zaps and lnurl are completely separate things and keeping them together gain us nothing, it's just strictly worse. See my comments above.

Think about all the non-lnurl providers that could become zap providers.

Implementing standalone zap support anywhere would take less than a day. And it would be much easier than implementing lnurl and then heavily modifying it to add zaps.

fiatjaf:
Without the LNURL garbage it becomes much simpler to implement a zaps provider.

For example, one could do a zaps provider that doesn't require any registration. The provider just has a static URL https://zapper.com and infer from the `p` tag of the zap request who should receive that money.

Then pay it out to the lud06 or lud16 addresses on that profile metadata -- or allow them to withdraw later, or even send it out to the Bitcoin address implied by the Nostr key (probably not a good idea, but I'm sure people will do it).

v0l:
> Without the LNURL garbage it becomes much simpler to implement a zaps provider.
> 
> For example, one could do a zaps provider that doesn't require any registration. The provider just has a static URL https://zapper.com and infer from the `p` tag of the zap request who should receive that money.
> 
> Then pay it out to the lud06 or lud16 addresses on that profile metadata -- or allow them to withdraw later, or even send it out to the Bitcoin address implied by the Nostr key (probably not a good idea, but I'm sure people will do it).

This all sounds great, but seems more like additional to the LNURL flow.

If you decouple it from LNURL flow then you somehow have to determine who the recipient is, at least with LNURL support you already know who the recipient is.

fiatjaf:
> If you decouple it from LNURL flow then you somehow have to determine who the recipient is

For all current providers who have accounts registered with usernames or ids and already give you an LNURL like `https://strike.army/.well-known/lnurlp/username` they can now just give a new one that is similar, like `https://strike.army/.well-known/zap/username` (or use any other path).

It would work great with platforms like lnbits too that don't have usernames, `https://legend.lnbits.com/zap/3745729`.

v0l:
> > If you decouple it from LNURL flow then you somehow have to determine who the recipient is
> 
> For all current providers who have accounts registered with usernames or ids and already give you an LNURL like `https://strike.army/.well-known/lnurlp/username` they can now just give a new one that is similar, like `https://strike.army/.well-known/zap/username` (or use any other path).
> 
> It would work great with platforms like lnbits too that don't have usernames, `https://legend.lnbits.com/zap/3745729`.

Seems reasonable, but would you still consider dropping the LNURL flow?

fiatjaf:
> Seems reasonable, but would you still consider dropping the LNURL flow?

Why not? There is nothing of LNURL anywhere in this spec, except reusing LNURL endpoints.

v0l:
I guess i'm ok with it then.

It might also be better to let the nostr client broadcast the receipt too, all the zap service needs to do is sign the event and return it to the client so it can broadcast it.

Maybe the `POST` to `/.well-known/zap/username` will return something like:
```json
{
  "pr": "lnbc1....",
  "callback": "https://example.com/path-to-get-zap-receipt"
}
```

This makes it a bit harder for clients, but easier for zap services to implement.

This also removes the need to tag relays in the zap note.

If the invoice is not paid yet the callback should return `204` status, clients can poll this endpoint until `200` status response with plain json zap receipt.

This doubles as a way to detect completed payments when it's made outside the client.

How does this sound?

jb55:
I think switching to polling would greatly complicate client implementations. I don't even know how I would implement that in damus since there's no way to know if a user gave up paying an invoice.

Most wallet providers can send the zap after they receive a payment for the invoice, I don't see the point in having the client send it.

jb55:
(via DM) @v0l makes a good point: "With the pace of change in Nostr, it might be hard for them to manage all the features, AUTH etc could break their zaps"

Not to mention zappers couldn't zap to paid relays. Perhaps it wouldn't hurt to have a way for clients to poll for the zap so that it could broadcast it itself to paid and AUTHd relays.

fiatjaf:
I think zappers publishing events is good. To solve the AUTH stuff and paid relays shenanigans I think zappers could optionally advertise a set of relays that they themselves always push stuff to [^1].

That way the users that are sending the zaps -- or the users that are receiving zaps -- can fetch the zap events later, if they think that is necessary, and post to their own restricted relays.

[^1]: which could be a fake relay at their own host if they want, and it just returns their own zap events

fiatjaf:
Now an outcome that will be both good and likely to happen is that paid relays will start accepting zap notes that target their own paying users automatically, so this might end up not being a problem.

Unless it is an anti-Bitcoin relay.

bitkarrot:
> 4\. bolt12

Please do consider Bolt12 support in zaps, especially for the potential applications that are dreaming about Bolt12 style recurring offers and selective invoice info reveal

ok300:
I may be missing something, but I think this can be simplified:

Zaps are essentially nostr notes of paid invoices. That comes down to LNURL-pay with a public note sent by the receiving LN node, which has a link to the sender pubkey.

Then why not just use LNURL addresses with the sender pubkey embedded in them, like gmail does for labels:

- normal LN address: alice@ln.com
- if Bob wants to zap to Alice, the nostr client can show the LN address alice+bob_pubkey@ln.com
- if Bob wants to zap to one of Alice's notes, the nostr client can show the LN address alice+bob_pubkey+alice_note_id@ln.com
- the zapper is a LNURL-pay endpoint that will also:
    - extract the pubkey from the LN address (and includes it as arg in the callback link for getting the invoice)
    - perform normal LNURL-pay
    - monitor the invoice[^1], post zap note when paid

The zap note can be a simple JSON with the amount paid and comment (e.g. no BOLT11).

The clients can tell if a LN Address supports zaps by calling a path like `/.well-known/zap`.

Benefits:
- any LNURL wallet can be used to zap
- very simple client logic:
  - to zap, just append own pubkey in the recipient LN Address
  - to read zaps, query zap notes and sum up amounts (no BOLT11 parsing)

Would that work? Am I missing something?


[^1]: LND: https://lightning.engineering/api-docs/api/lnd/invoices/lookup-invoice-v2

ok300:
The only downside I can think of is that it would allow Carol to zap Alice at alice+bob_pubkey@ln.com.

This would appear as a zap from Bob to Alice, but it's like Carol donated to Alice in Bob's name. Who would mind that?

bumi:
I love the zaps - as everybody :) seeing the excitement and interest on this is a great sign for monetization options.

As mentioned I am a bit concerned about the Nostr specific requirements for LNURL providers. This adds implementation complexity and we also create a Nostr island that is no longer interoperable. 
LNURL has brought some standardization and allows people to use their preferred tools and service providers and wallets.  The support for it among the lightning ecosystem is big and growing. For example the podcasting 2.0 community also discusses the use of LNURL (starting with LUD16) as an alternative to the currently used keysend payments.

To me reusabiltity and interoperability is an important goal. If people can use their existing lightning setup to receive payments on Nostr then this is favorable and simplifies and especially speeds up adoption. It took years for many providers and tools to implement LNURL-pay and we are still far from seeing its full potential.

I also think the lightning related spec should be universal. Being able to use it also for other areas like for example blogging makes it more powerful.

If we couple this closely to Nostr and if it is only useful within Nostr then imo it's better to just have a Nostr specific setup (as @fiatjaf says). 


From my understanding the requirements are:

* having a new nostr note type that references the payment (kind 9735 - the zap) to tally the payment
* being able to associate the payment to the note or the zap on the lightning node
* having the possibility to somewhat validate that the payment had happend - but ultimately we're trusting the recipient.

is this correct?


With [LUD18](https://github.com/lnurl/luds/blob/luds/18.md) we have the protocol to pass a payerData record to the LNURL server

## A potential protocol flow:

### Client side

1. Calculate the lnurl pay request URL for a user from the lud06 or lud16 field on their profile

2. Fetch the lnurl pay request static endpoint (`https://host.com/.well-known/lnurlp/user`) and gather the payerData record.

3. To get a `invoice`, call the `callback` url with `amount` set to the milli-satoshi amount value and the payerData as described in [LUD18](https://github.com/lnurl/luds/blob/luds/18.md).
The `pubkey` is the nostr pubkey.
a `payable` could contain a URI of the note (nostr:note...) or the full note itself. This allows the recipient to reference from the LN side the receiving note.
To align with the current proposal the zap request could be passed in here.
If a LNURL provider does not implement this it would still work the recipient would mainly miss the reference to the note which was paid.

4. Pay this invoice or pass it to an app that can pay the invoice. For improved usability I expect more clients to integrate lightning wallets through APIs like the ZBD API, Breez SDK, Alby API, WebLN, etc. (web clients already do this with WebLN) Those APIs return the preimage to the initiating client.

If the invoice is passed to an app the preimage can not be passed back. In this case the [LNURL-verify](https://github.com/lnurl/luds/issues/182) can be used to check the payment status and retrieve details like the preimage from the LNURL-server. (this is what @v0l proposes above) 
If none is available and the preimage can not be gathered the validation would not work.


5. Create and publish the zap note including the payment request and the preimage for some validation
Note: the description_hash included the payerdata. The payment hash from the invoice can be checked against the preimage. The note maybe also should include a lnurl-verify link for other clients to validate the payment (basically clients can ask the recipient if that zap was received). Like LNURL itself this would simpley use HTTP. 


### LNURL Server side

+ the LNURL server will create the invoice as described in LUD18 - it needs support for LUD18. If not then just no Nostr data would be saved. (like currently with email/name etc.)

+ if publishing the zap from the receiving node is really preferred (instead of 5. on the client side) the zap request could be loaded from the LUD18 field and published. - This would be the only Nostr specific implementation then on the LNURL-Sever side.

+ the LNURL Server should not need to implement any Nostr specific requirements and no new protocols (stick to HTTP and not add a WS dependency) 



As mentioned I generally think that the LNURL-pay spec should not contain any application/usage specific requirements. From a LNURL point of view Nostr is "just" one usage. I think with some abstraction this can be done and does not require high-level Nostr coupling.


P.S. 
the speed here in Nostr is amazing. I started typing this a few days ago but was too busy finishing it and so much had happend. :) 
Go Nostr!

starbackr-dev:
> 5. Create and publish the zap note including the payment request and the preimage for some validation
>    Note: the description_hash included the payerdata. The payment hash from the invoice can be checked against the preimage. The note maybe also should include a lnurl-verify link for other clients to validate the payment (basically clients can ask the recipient if that zap was received). Like LNURL itself this would simpley use HTTP.
> 

Thanks Bumi for the approach. I really like multiple options in who is creating the note kind - 9735. Nostr clients should determine this and should not force lightning providers to create this. It is not important at all on who is creating this note kind 9735 only the content in this note such as preimage, bolt11 and LNURL-verify is important. 

This also opens up another scenario/use case for anonymous zaps where the zap sender pay for sats but don't want his pub key associated with it.  Think of Canadian truck drivers protest. To keep anonymity, after the payment is successful, Nostr client can create this note with a random private key and destroy the key after it is posted. Zap receiver or the public don't know who zapped it but they receive sats which is powerful. 

Am also up for implementing multiple approaches to see which one is liked and adopted by the users. Thanks to @jb55 and @v0l for pushing a working version into the wild and seeing the reaction.

fiatjaf:
> It is not important at all on who is creating this note kind 9735

In fact it is. When displaying the zaps, clients are trusting that they come from a reputable provider -- or from the direct zap receiver themselves. If anyone can create an event then I can create a trillion zap events myself and fake that I have tipped the entire universe.

starbackr-dev:
> clients are trusting that they come from a reputable provider

Don't trust but verify...! if you are trusting that they came from a reputable provider then there is no difference between today's world (Apple trusting VisaMastercard). We are simply replacing the current overlords with new ones. You can argue that the new overlords are better but that is not the point. The lightning/wallet providers should provide a verifiable link for the payment [LN-URL](https://github.com/lnurl/luds/issues/182) which can be embedded on the note.

fiatjaf:
About @bumi's proposal above, I want to say that it would be nice to have it realized, but I am afraid the added complexity wouldn't be worth the trouble. @jb55 has actually implemented a zap flow using LUD-18 as far as I know, but later decided to drop it, which I think was sensible (I just don't understand why he didn't go all the way and dropped everything, which is what I want to happen).

The complexity I mention above is mostly the reliance on the `lnurl-verify` bit which is not implemented anywhere, and then the need for clients to poll that endpoint.

There is also the problem that the payer is the one publishing the zap, which opens the door for anyone to forge zaps @starbackr-dev (unless we introduce another requiment, that the invoice is signed by the receiver nostr public key or sometihng like that, i.e. more complexity). I think the receiver should be the one signing the `zap` events. My solution for a pure-lnurl flow at https://github.com/nostr-protocol/nips/pull/243#issuecomment-1427030831 addresses that (actually it is @ok300's proposal, I just made a suggestion on top of it), but it introduces too many requirements and complexity, so I don't think it is worth the trouble either.

Considering all this I believe the most sensible path forward is to get rid of lnurl entirely and make a simpler protocol that is specific to Nostr.

bumi:
> In fact it is. When displaying the zaps, clients are trusting that they come from a reputable provider

the client only needs to be able to verify it. In my proposal this is done by asking the receiver (reputable provider) - not through a nostr note signing but through the same method we actually receive the invoice from the receiver. (with a HTTP call)

both LUD-18 and LNURL-verify would be optional. LUD-18 is only needed to tell the recipient for which note that payment was. and LNURL-verify would only be for clients to check if the claimed payment actually happened. 

@fiatjaf can you describe the complexity that you see? 
because to me it is much less. LUD18 is already often implemented and the verification is one HTTP endpoint. It only requires data that the provider already has and the provider already speaks HTTP to create the invoice. So it is not more than another route. 
Requiring LNURL providers now to implement event validation, signing and websockets is much more complexity to me and requires a whole new flow. LNURL providers should not be required to implement social networking publishing features - also not to Nostr.

If we would do that, I am with @fiatjaf that this should no longer be LNURL. That spec should be universal and not have such a tight coupling to something else like some social networking.

jb55:
what makes something lnurl vs not lnurl? I agree it's not lnurl, all I'm doing is piggy-backing on the lnurl endpoint since these servers are already setup to fetch invoices from backends. the only thing that changes is the metadata gets swapped out with a nostr note.

This is the entirety of the change required on my lnurl server:

```diff
From e47262ff9826169360b6f2015271af8938b19af9 Mon Sep 17 00:00:00 2001
From: William Casarin <jb55@jb55.com>
Date: Sun, 8 Jan 2023 17:11:26 -0800
Subject: [PATCH] accept nostr notes

---
 index.js          | 94 ++++++++++++++++++++++++++++++++++++++++++-----
 package.json      |  2 +-
 public/index.html |  2 +-
 3 files changed, 87 insertions(+), 11 deletions(-)

diff --git a/index.js b/index.js
index d836358..89a793c 100644
--- a/index.js
+++ b/index.js
@@ -10,6 +10,7 @@ const path = require('path')
 
 const SUPPORT = 'jb55@jb55.com'
 const PRICE = '39999000'
+
 const INVOICE_AUTH = {
 	id: "03f3c108ccd536b8526841f0a5c58212bb9e6584a1eb493080e7c1cc34f82dad71",
 	address: "monad.endpoint.jb55.com",
@@ -45,12 +46,12 @@ async function make_request(auth, method, params) {
 }
 
 
-async function make_invoice(opts, amount)
+async function make_invoice(opts, amount, extra)
 {
 	try {
 		const res = await make_request(opts, "invoice", {
 			amount_msat: amount || "any",
-			description: make_metadata_str(opts),
+			description: make_metadata_str(opts, extra),
 			deschashonly: true,
 			label: crypto.randomUUID().toString(),
 		})
@@ -67,7 +68,7 @@ async function make_invoice(opts, amount)
 	return null
 }
 
-function make_metadata(opts) {
+function make_metadata(opts, extra={}) {
 	let metadata = []
 	if (opts.description) {
 		metadata.push(["text/plain", opts.description])
@@ -75,18 +76,24 @@ function make_metadata(opts) {
 	if (opts.longDescription) {
 		metadata.push([`text/long-desc`, opts.longDescription])
 	}
-
 	if (opts.user)
 		metadata.push(["text/identifier", opts.user + "@sendsats.lol"])
 
 	if (opts.thumbnail) {
 		metadata.push([`image/${opts.thumbnailType};base64`, opts.thumbnail])
 	}
+
+	if (extra.nostr)
+		metadata.push(["application/nostr", extra.nostr])
+
 	return metadata
 }
 
-function make_metadata_str(opts) {
-	return JSON.stringify(make_metadata(opts))
+function make_metadata_str(opts, extra) {
+	if (extra && extra.nostr)
+		return JSON.stringify(extra.nostr)
+
+	return JSON.stringify(make_metadata(opts, extra))
 }
 
 async function find_metadata(user)
@@ -125,6 +132,23 @@ function invalid_username(res) {
 	return json_err(res, "Bad username", 400)
 }
 
+function get_nostr_note(parsed) {
+	let nostr = parsed.searchParams.get('nostr')
+	if (nostr == null)
+		return
+
+	try {
+		nostr = JSON.parse(nostr)
+	} catch (e) {
+		throw new Error(`nostr parameter must be a nostr note: ${e}`)
+	}
+
+	const validated = validate_note(nostr)
+	if (validated !== true)
+		throw new Error(validated)
+	return nostr
+}
+
 async function handle_static_payreq(parsed, req, res)
 {
 	let user = user_from_wellknown(parsed.pathname)
@@ -137,13 +161,26 @@ async function handle_static_payreq(parsed, req, res)
 	if (metadata === null)
 		return user_not_found(res)
 
-	const resp = {
+	let nostr
+	try {
+		nostr = get_nostr_note(parsed)
+	} catch (e) {
+		return json_err(res, e.toString(), 400)
+	}
+	const extra = { nostr }
+
+	let resp = {
 		status: "OK",
 		tag: "payRequest",
 		minSendable: 1,
 		maxSendable: 10000000000,
 		callback: `https://sendsats.lol/@${user}`,
-		metadata: make_metadata_str(metadata),
+		metadata: make_metadata_str(metadata, extra),
+	}
+
+	if (user === "jb55") {
+		resp.allowsNostr = true
+		resp.nostrPubkey = "9630f464cca6a5147aa8a35f0bcdd3ce485324e732fd39e09233b1d848238f31"
 	}
 
 	return json_ok(res, resp)
@@ -161,12 +198,21 @@ async function handle_payreq(parsed, req, res)
 	if (!amount)
 		return json_err(res, "Missing amount", 400)
 
+	
+	let nostr
+	try {
+		nostr = get_nostr_note(parsed)
+	} catch (e) {
+		return json_err(res, e.toString(), 400)
+	}
+	const extra = { nostr }
+
 	const metadata = await find_metadata(user)
 	if (metadata === null)
 		return user_not_found(res)
 
 	try {
-		const pr = await make_invoice(metadata, amount)
+		const pr = await make_invoice(metadata, amount, extra)
 		const routes = []
 		json_ok(res, {pr, routes})
 	} catch (e) {
@@ -316,6 +362,36 @@ async function serve_page(pathname, req, res)
 
 }
 
+function validate_note(note)
+{
+	if (note.kind !== 9734)
+		return "Expected tip request note (kind 9734)"
+
+	if (!note.tags)
+		return "No tags found?"
+
+	// Make sure we only have one p tag
+        const ptags = note.tags.filter(t =>
+		t && t.length && t.length >= 2 && t[0] === "p")
+
+        if (ptags.length !== 1)
+                return "None or multiple p tags found"
+
+	// Make sure we have 0 or 1 etag (for note tipping)
+	const etags = note.tags.filter(t =>
+		t && t.length && t.length >= 2 && t[0] === "e")
+	if (!(etags.length === 0 || etags.length === 1))
+		return "Expected 0 or 1 e tags"
+
+	// Look for the relays tag, the wallet will broadcast to these relays
+	const relays_tag = note.tags.find(t =>
+		t && t.length && t.length >= 2 && t[0] === "relays")
+	if (!relays_tag)
+		return "No relays tag found"
+
+	return true
+}
+
 function handle_request(opts, req, res)
 {
 	console.log("%s - %s", req.method, req.url)
diff --git a/package.json b/package.json
index eba4435..d68a4c0 100644
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "sendsats.lol",
   "description": "sendsats.lol",
-  "version": "0.1.0",
+  "version": "1.2.2",
   "repository": {
     "url": "https://github.com/jb55/sendsats.lol"
   },
diff --git a/public/index.html b/public/index.html
index 3f7cbc0..0c26a25 100644
--- a/public/index.html
+++ b/public/index.html
@@ -102,7 +102,7 @@
 
 	  <script src="js/qrcode.js?v=3" > </script>
 	  <script src="js/lnsocket.js?v=3" > </script>
-	  <script src="js/index.js?v=10" > </script>
+	  <script src="js/index.js?v=11" > </script>
 
     </body>
 </html>
-- 
2.39.1

```
The proposed alternative seems like I would have to rejig everything for reasons which I don't understand.

jb55:
afaik snort, damus and amethyst have already implemented the original spec

fiatjaf:
The change is small, but it introduces spaghetti in your lnurl server.

And it is only small because you already have an lnurl server.

Now for people that want to make zap-only servers, they will have to make a full lnurl server first, then introduce your changes -- while if the protocol didn't rely on anything related to lnurl their complete job of implementing zaps would be much smaller.

fiatjaf:
Another problem with the way this NIP was made is that now users can't have separate zap and lnurl services on their profile.

They could choose to have an lnurl for donations -- and that same service doesn't support zaps, but they can't.

Or they could decide to use a service that just supports zaps, not lnurl, and now people will see on clients that they accept lnurl and just be frustrated when paying.

The proposal will either make user experience worse everywhere or make it so only providers implementing both protocols will be chosen by users, which will have a centralizing effect.

starbackr-dev:
According to @fiatjaf  comments, both approaches appear to have equal pros and cons. However, @jb55  stated that the specifications have already been implemented in three popular Nostr clients. So I went ahead and started putting it into action. It only took a couple of hours for me to change the backend so that LNURL will continue to work for non-Nostr use cases and posting type 9735 will just be implemented for invoices created by Nostr clients. It is preferable to test this in public to obtain user feedback rather than iterating inside github.


Here’s the context of where and why we did it. 

We founded starbackr in 2021 to empower the creator economy with the power of bitcoin lightning payments. We iterated on many features over the last year and moved toward value for value. Tipping was popular, with sats flowing from wallet to wallet based on good content. So, in November 2022, we removed the like button and replaced it with Zaps. To pay sats on posts, all you had to do was click one button. @Egge7 , my co-founder, wrote a [blog post](https://starbackr.com/liberating-social-media/) about it. 

It was an instant success, and many people began using the zaps to tip posts with the best content. We began cross-posting on Nostr in December, and many posts received zaps from Nostr users. As a result, we decided to build our Native app using the Nostr protocol. 

We spent the last few weeks developing the app for both iOS and Android, and we just received TestFlight approval from Apple. We will begin rolling out to users in the coming days, and they will be able to use zap immediately by sending/receiving sats.

The app name “Current” is built on react native, and we open sourced it so that we can move forward quickly as a community. Here’s the [project github](https://github.com/starbackr-com/current) link which details the features, our approach and roadmap. 

We received a lot of feedback from users about how long it will take them to start using the Nostr protocol and what the minimum viable features will be. Ability to find users based on Twitter followers, create a verified ID, and use an integrated wallet to send/receive zaps were the main ones. All of these features are available today, but as separate web apps. So we combined them all into a single app.

Cross-client support will be critical to Nostr's success. Users will be able to use their preferred app/client while still communicating with everyone in their network. We will do everything possible to support any functionality added by other clients and keep moving forward. 

I'm grateful to be a part of this community and can't wait to meet many of you in Nostrica...!

fiatjaf:
Turns out this was settled long before the PR was opened, there were no answers to any the feedback presented here at all. Committing directly to master would have been better and would have wasted less time from others.

ok300:
Agree. Very unfortunate.

pgerstbach:
Reading this PR makes me think why the rush? Nostr is developing fast, that's great, but more "PRs" like this and Nostr might end up in a complete mess. It suppose it is good practice to propose something and implement it in parallel to do a reality-check, but if the discussion shows that some NIP has flaws early developers need to re-do the work and not stick to it. I wonder how this will turn out...

To my understanding (as a interested user, not as a developer) I find it confusing to have a working lnurl service, but zaps do not work. And if I want to test zaps with a supported provider, I can't use my working lnurl service.

ok300:
There was a proposal to address exactly that: #244

ok300:
There were also other proposals above that made zaps not interfere with LNURL, or at least be backwards compatible with existing LNURL services.

But hey, this was easy to implement bro, don't understand what the fuss is all about.

PR #226, opened 2023-02-06
Update README.md kind table

lionello:
Add `kind` 1000-9999 per NIP-16 ~~and 6 "boosts" (no NIP?)~~

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,7 @@ NIPs stand for **Nostr Implementation Possibilities**. They exist to document wh
 | 44          | Channel Mute User           | [28](28.md)            |
 | 45-49       | Public Chat Reserved        | [28](28.md)            |
 | 22242       | Client Authentication       | [42](42.md)            |
+| 1000-9999   | Regular Events Reserved     | [16](16.md)            |
 | 10000-19999 | Replaceable Events Reserved | [16](16.md)            |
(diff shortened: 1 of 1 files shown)


fiatjaf:
For kind 6 I think it is better to wait until we have a NIP.

lionello:
or I can write one based off of Damus' code?

lionello:
Sorry, just noticed https://github.com/nostr-protocol/nips/pull/86

README.md
```diff
@@ -49,6 +49,7 @@ NIPs stand for **Nostr Implementation Possibilities**. They exist to document wh
 | 44          | Channel Mute User           | [28](28.md)            |
 | 45-49       | Public Chat Reserved        | [28](28.md)            |
 | 22242       | Client Authentication       | [42](42.md)            |
+| 1000-9999   | Regular Events Reserved     | [16](16.md)            |
```
Cameri, line 52: What's a Regular Event?

lionello: According to https://github.com/nostr-protocol/nips/blob/master/16.md#regular-events
> A regular event is defined as an event with a kind 1000 <= n < 10000. Upon a regular event being received, the relay SHOULD send it to all clients with a matching filter, and SHOULD store it. New events of the same kind do not affect previous events in any way.

PR #227, opened 2023-02-07
Nip 05 security proposal

erikwestra:
Proposing a couple of changes to the NIP-05 protocol to reduce the chance of fraudulent use of "verified" public keys.  At present, I could create an account on a well-known verifying server under a random name, and then send DMs pretending to be someone else, and there's no easy way for users to tell who the verifying account actually belongs to.

As well as displaying the name of the account on the verifying server, this PR suggests an enhancement to the JSON data being returned so that clients can redirect the user to the user's profile page on the server.  This will make it much easier for users to check that someone who claims to have verified their Nostr account is who they claim to be.

### DIFF
--- a/05.md
+++ b/05.md
@@ -35,7 +35,7 @@ It will make a GET request to `https://example.com/.well-known/nostr.json?name=b
 }
 ````
 
-or with the **optional** `"relays"` attribute:
+or with the **optional** `"relays"` and/or `"account_uris"` attributes:
 
 ```json
 {
@@ -44,6 +44,9 @@ or with the **optional** `"relays"` attribute:
   },
   "relays": {
     "b0635d6a9851d3aed0cd6c495b282167acf761729078d975fc341b22650b07b9": [ "wss://relay.example.com", "wss://relay2.example.com" ]
+  },
+  "account_uris": {
+    "b0635d6a9851d3aed0cd6c495b282167acf761729078d975fc341b22650b07b9": "https://bob.com/profile/bob"
   }
 }
 ````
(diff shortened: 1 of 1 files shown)


fiatjaf:
Looks good to me.

mplorentz (approved):
This looks good to me. The `account_uris` are backwards compatible but provide much-needed space for account verifiers to give context to the verification.

This has been open for a few days, it has the thumbs up from fiatjaf, and no dissenting opinions, so I'm clicking the merge button!

fiatjaf:
@erikwestra @mplorentz since the account URIs must only be at the same domain name as the NIP-05 address I figured out we should just strip out that part, which makes the job of webservers easier (it is hard for a webserver to know its own hostname): https://github.com/nostr-protocol/nips/commit/d87763781dc1213d7c1b53ab0a4172f8237cbdf3

fiatjaf:
But to be fair I don't think this should have been merged so quickly. I still think questions would arrive. For example, I just realized that this is only valid for platforms that already exist and already have accounts and usernames in them, is that right, with profile pages and all that.

Do we even have any of these platforms hosting NIP-05 addresses for users today? I guess there is https://stacker.news/, but I can't think of any other. Is that a thing we expect to see in the future? Maybe we should wait until that future to arrive before merging this then.

mplorentz:
@fiatjaf ah sorry for jumping the gun. I went ahead and reverted the merge. Still figuring out the process here.

@erikwestra I don't think I can reopen this PR. Could you create a new one with your branch? And consider cherry-picking `d877637` from @fiatjaf.

erikwestra:
@fiatjaf the reason I included the domain name in the URI was to allow the URI to reference a subdomain on the verifying domain.  For example, if the user enters a NIP-05 verification code of `john@example.com`, the server could use an `account_uri` of `https://accounts.example.com/john` and it would still work.

Not sure if that's too much of an edge case -- your suggested change is simpler.  Let me know what you think -- we can either go with your simplification or allow for subdomains as I suggested in the original.  I'll then resubmit a PR with whatever you folks decide (though it might not be until next week as I'm tied up for the next 24-36 hours).

mplorentz:
@erikwestra that makes sense to me. Many people like to use subdomains for their aliases on scuttlebutt servers. I'm sure people will want to do the same on Nostr. My NIP-05 might be matt@nos.social, but if I have a profile page at matt.nos.social I want that to open when people tap on matt@nos.social in clients.

I could see another use case being NIP-05 hosts that do some type of verification. Maybe I own nostr-real-names.com where people have to show a government ID to get verified. Tapping matt@nostr-real-names.com could take me to nostr-real-names.com/about-verification which explains this.

PR #229, opened 2023-02-07
docs: add nip-58 badge event and profile badges

Cameri:
Adds event kind 8 meaning Badge and event kind 30008 parameterized replaceable event meaning Profile Badge.

### DIFF
--- a/58.md
+++ b/58.md
@@ -82,6 +82,7 @@ Clients SHOULD render the most appropriate badge thumbnail according to the numb
   "pubkey": "<alice's pubkey in hex>",
   "content": "",
   "tags": [
+    ["d", "profile_badges"],
     ["e", "<charlie's issued badge event id>"],
   ]
   ...other fields...
--- a/58.md
+++ b/58.md
@@ -51,6 +51,7 @@ Clients SHOULD render the most appropriate badge thumbnail according to the numb
 
 ### Example of Badge Event
 
+Event:
 ```json
 {
   "id": "<charlie's issued badge event id>",
@@ -59,29 +60,34 @@ Clients SHOULD render the most appropriate badge thumbnail according to the numb
   "tags": [
     ["p", "<alice's (recipient) pubkey in hex>"]
   ],
-  "content": "{
-    \"long_desc\": \"This user joined our relay 2 years ago.\",
-    \"short_desc\": \"Two-Year Club\",
-    \"image\": \"https://relay.nostr/badges/two-year-club.png\",
-    \"thumbs\": {
-      \"512x512\": \"https://relay.nostr/badges/two-year-club-512x512.png\",
-      \"256x256\": \"https://relay.nostr/badges/two-year-club-256x256.png\",
-      \"128x128\": \"https://relay.nostr/badges/two-year-club-128x128.png\",
-      \"64x64\": \"https://relay.nostr/badges/two-year-club-64x64.png\",
-      \"32x32\": \"https://relay.nostr/badges/two-year-club-32x32.png\",
-      \"16x16\": \"https://relay.nostr/badges/two-year-club-16x16.png\"
-    },
-    \"sha256sums\": {
-      \"image\": \"8d43b3081c8091147e03e41878604606c8aa53d1337d49a4d5e0e8c29c223798\",
-      \"512x512\": \"af94c2afee5321b6acbca7ba00fcfa4ae2e103dda3a3a038e8b0aec4207ef164\",
-      \"256x256\": \"8c1eaf36d0353e20e30511c473411f1df0b9e58f68daadad972a8a6955824a0e\",
-      \"128x128\": \"86b0ebe1cf924f39a0eb4cfde5cdfe8a271b882284fc4f1c353bf6589eef1870\",
-      \"64x64\": \"a92ec141189c8e0228d8c388b9c0e822ecf1347367b834c43c7e82427569e06d\",
-      \"32x32\": \"6afa09c719f0729253ce8034618743d3a3aadd7aa4215554b7e5aa72b46c41a9\",
-      \"16x16\": \"abacf2675c05b7710cf853800a95e4191c10b2005540a39adb4b7cdda617249b\"
-    }
-  }",
+  "content": "<json stringified object shown below>",
   ...other fields...
+}
+```
+
+Content:
+```json
+{
+  "long_desc": "This user joined our relay 2 years ago.",
+  "short_desc": "Two-Year Club",
+  "image": "https://relay.nostr/badges/two-year-club.png",
+  "thumbs": {
+    "512x512": "https://relay.nostr/badges/two-year-club-512x512.png",
+    "256x256": "https://relay.nostr/badges/two-year-club-256x256.png",
+    "128x128": "https://relay.nostr/badges/two-year-club-128x128.png",
+    "64x64": "https://relay.nostr/badges/two-year-club-64x64.png",
+    "32x32": "https://relay.nostr/badges/two-year-club-32x32.png",
+    "16x16": "https://relay.nostr/badges/two-year-club-16x16.png"
+  },
+  "sha256sums": {
+    "image": "8d43b3081c8091147e03e41878604606c8aa53d1337d49a4d5e0e8c29c223798",
+    "512x512": "af94c2afee5321b6acbca7ba00fcfa4ae2e103dda3a3a038e8b0aec4207ef164",
+    "256x256": "8c1eaf36d0353e20e30511c473411f1df0b9e58f68daadad972a8a6955824a0e",
+    "128x128": "86b0ebe1cf924f39a0eb4cfde5cdfe8a271b882284fc4f1c353bf6589eef1870",
+    "64x64": "a92ec141189c8e0228d8c388b9c0e822ecf1347367b834c43c7e82427569e06d",
+    "32x32": "6afa09c719f0729253ce8034618743d3a3aadd7aa4215554b7e5aa72b46c41a9",
+    "16x16": "abacf2675c05b7710cf853800a95e4191c10b2005540a39adb4b7cdda617249b"
+  }
 }
 ```
 
@@ -94,8 +100,8 @@ Clients SHOULD render the most appropriate badge thumbnail according to the numb
   "content": "",
   "tags": [
     ["d", "profile_badges"],
-    ["e", "<charlie's issued badge event id>"],
-  ]
+    ["e", "<charlie's issued badge event id>"]
+  ],
   ...other fields...
 }
 ```
--- a/58.md
+++ b/58.md
@@ -20,6 +20,8 @@ The `image` field is the URL of a high-resolution image representing the badge.
 
 The `thumbs` field is an object containing thumbnail versions of the image referenced in the `image` field. The format of the keys are the dimensions `width`x`height` in pixels and the value is an URL pointing to respective thumbnail image.
 
+The `sha256sums` field is an object containing the fields `image` and one for every thumbnail in the `thumbs` object with the same key. The values are the SHA256 digests of each one of image files.
+
 ### Profile Badges Event
 
 The number of badges a pubkey can receive is unbounded. The Profile Badge event allows individual users to choose
@@ -67,7 +69,16 @@ Clients SHOULD render the most appropriate badge thumbnail according to the numb
       \"128x128\": \"https://relay.nostr/badges/two-year-club-128x128.png\",
       \"64x64\": \"https://relay.nostr/badges/two-year-club-64x64.png\",
       \"32x32\": \"https://relay.nostr/badges/two-year-club-32x32.png\",
-      \"16x16\": \"https://relay.nostr/badges/two-year-club-16x16.png\",
+      \"16x16\": \"https://relay.nostr/badges/two-year-club-16x16.png\"
+    },
+    \"sha256sums\": {
+      \"image\": \"8d43b3081c8091147e03e41878604606c8aa53d1337d49a4d5e0e8c29c223798\",
+      \"512x512\": \"af94c2afee5321b6acbca7ba00fcfa4ae2e103dda3a3a038e8b0aec4207ef164\",
+      \"256x256\": \"8c1eaf36d0353e20e30511c473411f1df0b9e58f68daadad972a8a6955824a0e\",
+      \"128x128\": \"86b0ebe1cf924f39a0eb4cfde5cdfe8a271b882284fc4f1c353bf6589eef1870\",
+      \"64x64\": \"a92ec141189c8e0228d8c388b9c0e822ecf1347367b834c43c7e82427569e06d\",
+      \"32x32\": \"6afa09c719f0729253ce8034618743d3a3aadd7aa4215554b7e5aa72b46c41a9\",
+      \"16x16\": \"abacf2675c05b7710cf853800a95e4191c10b2005540a39adb4b7cdda617249b\"
     }
   }",
   ...other fields...
--- a/58.md
+++ b/58.md
@@ -1,0 +1,89 @@
+NIP-58
+======
+
+Badges
+------
+
+`draft` `optional` `author:cameri`
+
+A special event meaning "Badge" is defined as an event with kind `8` having a list of `p` tags,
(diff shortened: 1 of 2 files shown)


58.md
```diff
@@ -0,0 +1,89 @@
...
+      \"256x256\": \"https://relay.nostr/badges/two-year-club-256x256.png\",
+      \"128x128\": \"https://relay.nostr/badges/two-year-club-128x128.png\",
+      \"64x64\": \"https://relay.nostr/badges/two-year-club-64x64.png\",
+      \"32x32\": \"https://relay.nostr/badges/two-year-club-32x32.png\",
+      \"16x16\": \"https://relay.nostr/badges/two-year-club-16x16.png\",
+    }
+  }",
+  ...other fields...
+}
+```
```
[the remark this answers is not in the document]
Cameri: Kind 5?

58.md
```diff
@@ -0,0 +1,90 @@
...
+
+### Motivation
+
+Users MAY be awarded badges in recognition, in gratitude, for participation, or in appreciation of a certain goal, task or cause.
+
+Users MAY choose to decorate their profiles with badges for fame, notoriety, recognition, support, etc., from badge issuers they deem reputable.
+
+### Recommendations
+
+Badge issuers MAY include some Proof of Work as per [NIP-13](03.md) when minting badges to embed them with an energy cost, arguably making them more special and valuable for users that wish to collect them.
```
antonleviathan, line 40: Should be `[NIP-13](13.md)`, right?

fiatjaf:
Looks good.

gkbrk:
For `image` and `thumbs` that are used to display images from a URL, should the minted badge include the SHA-256 hash of the image as well? I'm guessing we don't want the image host or badge issuer to be able to change the badge images based on time-of-day, or client IP without the client being able to notice something has changed.

gkbrk:
Oh, another thing related to the URLs in this NIP. Are the image and thumbnail URLs supposed to be HTTP or HTTPS only? Or are other schemes valid as well? Like data URLs, `ipfs://` URLs, Bittorrent magnet URLs etc.

Cameri:
> Oh, another thing related to the URLs in this NIP. Are the image and thumbnail URLs supposed to be HTTP or HTTPS only? Or are other schemes valid as well? Like data URLs, `ipfs://` URLs, Bittorrent magnet URLs etc.

URL so it can be located anywhere.

sondreb:
I like this idea, just one question: What is the purpose of having ["d", "profile_badges"] in the tags? Is this not already known through the kind 30008?

gkbrk:
> I like this idea, just one question: What is the purpose of having ["d", "profile_badges"] in the tags? Is this not already known through the kind 30008?

This prevents another kind 30008 event from replacing it unless that event is also for the key `profile_badges`.

These are parameterized replaceable events, so the kind isn't super important. You could use kind 30008 for another purpose, and if you used a different `d` tag things wouldn't overwrite each other.

58.md
```diff
@@ -0,0 +1,107 @@
...
+  "kind": 30008,
+  "pubkey": "<alice's pubkey in hex>",
+  "content": "",
+  "tags": [
+    ["d", "profile_badges"],
+    ["e", "<charlie's issued badge event id>"]
+  ],
+  ...other fields...
+}
+```
```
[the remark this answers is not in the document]
Cameri: hoping to use kind 30008 for Badge lists, I'd like to keep it a parameterized replaceable event

fiatjaf:
> This prevents another kind 30008 event from replacing it unless that event is also for the key profile_badges.
> These are parameterized replaceable events, so the kind isn't super important. You could use kind 30008 for another purpose, and if you used a different d tag things wouldn't overwrite each other.

I am not happy with this answer. This feels like introducing two ways of doing the same thing.

Why not just define kind `10003`  as meaning "profile badges" and put all badges there?

The answer above says "the kind isn't super important". To me if it isn't important it shouldn't be defined or used.

Do you also think kind `0` events should be a parameterized replaceable event and have a `d` tag that says "metadata"?

gkbrk:
> I am not happy with this answer. This feels like introducing two ways of doing the same thing.

To clarify, I didn't write this NIP so my answer might not reflect that of the author.

> Why not just define kind `10003` as meaning "profile badges" and put all badges there?

Makes sense to me.

> The answer above says "the kind isn't super important". To me if it isn't important it shouldn't be defined or used.

Kind isn't an optional field, so it needs to be used.

You can't query filter by `d` tags unless the relay implements the generic tag query NIP, but you can fall back to a kind query to find this event.

> Do you also think kind `0` events should be a parameterized replaceable event and have a `d` tag that says "metadata"?

Yes.

I also think setting aside arbitrary ranges for replaceable/ephemeral/parameterized replaceable was dumb, should have been tags or another field that determines that. Why should a relay hardcode that kind=0 is a replaceable event, instead of the event specifying that in can be replaced.

The person sending the event already knows if it should be replaced or not, or ephemeral or not, or if it should be replaced with `pubkey + kind`, or `pubkey + kind + replacement key`, there is no need to use ranges for this.

fiatjaf:
> you can fall back to a kind

This is not good. If the kind is enough we should use just the kind.

> Yes.

Has that been working bad so far such that it would be improved by having a `d` tag? The best you'll achieve with that is to just replace the entire concept of kinds with that of `d` tags, which essentially become named kinds.

> Why should a relay hardcode that kind=0 is a replaceable event

The event specifies that by being in a kind range that is replaceable. No need to transfer an additional useless property called "replaceable", no need to parse that every single time. Ideally we wouldn't have these ranges and everything would be hardcoded, but that's not feasible, so the ranges are a nice solution.

I know programmers like to parse things at runtime and have the program dynamically handle everything, but that's not a good idea for open protocols. Some things must always be hardcoded. And the more things you can manage to hardcode the better.

Or maybe you wanted, for example, that each event had a JSON schema attached to itself specifying what fields it contains. It would also have to specify a hashing algorithm in itself, hashing function, curve and signature algorithm, because it would be dumb to have these things hardcoded! Not to mention the hardcoding of the JSON spec, why not allow events to have any encoding they want, they just specify that in their leading byte, we don't want to be dumb and hardcode the JSON format! And then you also need relays to advertise what protocols they accept for connections, because it would be dumb to hardcode that they all use websockets!

58.md
```diff
@@ -0,0 +1,107 @@
+NIP-58
+======
+
+Badges
+------
+
+`draft` `optional` `author:cameri`
+
+A special event meaning "Badge" is defined as an event with kind `8` having a list of `p` tags,
+one for each pubkey the author (issuer) wishes to award. A parameterized replaceable event meaning "Profile Badges" is defined as an event with kind `30008` with a `d` tag identifier equal to `profile_badges` and having an ordered list of `e` tags, one for each badge the user wishes to show on their profile. Badges as immutable events are non-transferrable.
```
fiatjaf, line 10:
```suggestion
one for each pubkey the author (issuer) wishes to award. A parameterized replaceable event meaning "Profile Badges" is defined as an event with kind `10003` and having an ordered list of `e` tags, one for each badge the user wishes to show on their profile. Badges as immutable events are non-transferrable.
```

58.md
```diff
@@ -0,0 +1,107 @@
...
+  },
+  "sha256sums": {
+    "image": "8d43b3081c8091147e03e41878604606c8aa53d1337d49a4d5e0e8c29c223798",
+    "512x512": "af94c2afee5321b6acbca7ba00fcfa4ae2e103dda3a3a038e8b0aec4207ef164",
+    "256x256": "8c1eaf36d0353e20e30511c473411f1df0b9e58f68daadad972a8a6955824a0e",
+    "128x128": "86b0ebe1cf924f39a0eb4cfde5cdfe8a271b882284fc4f1c353bf6589eef1870",
+    "64x64": "a92ec141189c8e0228d8c388b9c0e822ecf1347367b834c43c7e82427569e06d",
+    "32x32": "6afa09c719f0729253ce8034618743d3a3aadd7aa4215554b7e5aa72b46c41a9",
+    "16x16": "abacf2675c05b7710cf853800a95e4191c10b2005540a39adb4b7cdda617249b"
+  }
```
fiatjaf, line 90:
```suggestion
  "image": "https://relay.nostr/badges/two-year-club.png"
```

58.md
```diff
@@ -0,0 +1,107 @@
...
+
+### Example of a Profile Badges Event
+
+```json
+{
+  "kind": 30008,
+  "pubkey": "<alice's pubkey in hex>",
+  "content": "",
+  "tags": [
+    ["d", "profile_badges"],
```
fiatjaf, line 102:
```suggestion
  "kind": 10003,
  "pubkey": "<alice's pubkey in hex>",
  "content": "",
  "tags": [
```

58.md
fiatjaf: Aren't these a little excessive? Who is going to check these checksums? It's annoying to do, probably can't be done in a browser. And also: if someone has actually included checksums, that means they are not intending to change their images -- but if they are intending to they can just _not include_ the checksums and get away with that, so I don't see how they're doing anything.

I think we shouldn't checksums now, but leave room for them to be specified later when the need arises.

About the thumbnails, I think they also should be optional.

fiatjaf:
I think all these properties should be moved to tags and content should be an optional human-readable string describing the context of that specific badge award.

```js
"tags": [
  ["p", "deadbeef"],
  ["r", "bravery"], // this is so one can query how many of any given tags a public key has issued
  ["name", "Bravery"], // optional
  ["description", "Given to people who shows signs of bravery"], // optional
  ["image", "https://..."], // optional
  ["thumb", "https://..."], // optional
  ["image_sha256sum", "..."] // optional
],
"content": "bob won this badge because his actions on field"
```

Cameri:
> > This prevents another kind 30008 event from replacing it unless that event is also for the key profile_badges.
> 
> > These are parameterized replaceable events, so the kind isn't super important. You could use kind 30008 for another purpose, and if you used a different d tag things wouldn't overwrite each other.
> 
> 
> 
> I am not happy with this answer. This feels like introducing two ways of doing the same thing.
> 
> 
> 
> Why not just define kind `10003`  as meaning "profile badges" and put all badges there?
> 
> 
> 
> The answer above says "the kind isn't super important". To me if it isn't important it shouldn't be defined or used.
> 
> 
> 
> Do you also think kind `0` events should be a parameterized replaceable event and have a `d` tag that says "metadata"?

I was hoping to introduce more badge lists like on NIP-51 Lists: archived, awards, etc. If we make it a kind 10008 we won't be able to do this.

Cameri:
> > I am not happy with this answer. This feels like introducing two ways of doing the same thing.
> 
> 
> 
> To clarify, I didn't write this NIP so my answer might not reflect that of the author.
> 
> 
> 
> > Why not just define kind `10003` as meaning "profile badges" and put all badges there?
> 
> 
> 
> Makes sense to me.
> 
> 
> 
> > The answer above says "the kind isn't super important". To me if it isn't important it shouldn't be defined or used.
> 
> 
> 
> Kind isn't an optional field, so it needs to be used.
> 
> 
> 
> You can't query filter by `d` tags unless the relay implements the generic tag query NIP, but you can fall back to a kind query to find this event.
> 
> 
> 
> > Do you also think kind `0` events should be a parameterized replaceable event and have a `d` tag that says "metadata"?
> 
> 
> 
> Yes.
> 
> 
> 
> I also think setting aside arbitrary ranges for replaceable/ephemeral/parameterized replaceable was dumb, should have been tags or another field that determines that. Why should a relay hardcode that kind=0 is a replaceable event, instead of the event specifying that in can be replaced.
> 
> 
> 
> The person sending the event already knows if it should be replaced or not, or ephemeral or not, or if it should be replaced with `pubkey + kind`, or `pubkey + kind + replacement key`, there is no need to use ranges for this.

It wasn't dumb, Nostream recognizes these ranges and uses indexes to maintain the uniqueness of replaceable events. But yeah maybe we could have done the same without the ranges? The good thing about ranges is that the relays already support many kinds as special events without the need to make changes to the relay.

58.md
Cameri: The checksums are _optional_ and there so someone can't change the image to a dick pic and deface your beautiful profile page.
If we can't verify this in web clients perhaps we should reconsider, but I thought it was possible just didn't look into it.

Cameri: > Aren't these a little excessive? Who is going to check these checksums? It's annoying to do, probably can't be done in a browser. And also: if someone has actually included checksums, that means they are not intending to change their images -- but if they are intending to they can just _not include_ the checksums and get away with that, so I don't see how they're doing anything.
> 
> 
> 
> I think we shouldn't checksums now, but leave room for them to be specified later when the need arises.
> 
> 
> 
> About the thumbnails, I think they also should be optional, and less sizes specified, there are so many here, not even Google requires so many different sizes for publishing apps to the Android store.

I can specify that the sizes there are only recommendations and are all optional. Would that suffice? The high resolution version of a badge may be expensive to download specially if clients allow you to set many on your profile, so we'll need thumbs.

verbiricha:
> I was hoping to introduce more badge lists like on NIP-51 Lists: archived, awards, etc. If we make it a kind 10008 we won't be able to do this.

With this in mind I think is wort allowing for future extensibility with kind `30008`.

Also think is better to put badge data in tags intead of a JSON string in the content as @fiatjaf suggests, and image + thumbnail should be enough to display these in clients, no need to overcomplicate with lots of resolutions.

brenzi:
May argue for the need of a time-to-live (TTL) attribute for badge events?

Here's my use case: Personhood-based sybil resitance.
The [Encointer](https://encointer.org) protocol (as does IDENA and others) requires persons to attend regular concurrent key signing gatherings. In order to get a unique personhood proof, such a proof of attendance has to be ephemeral and [reputation](https://book.encointer.org/protocol-reputation.html) is only valid for 10days after which badges need renewal.

Motivation for nostr: Persons can only maintain one nostr profile with high reputation (because they can not be in more than one place at the same time of the gatherings), which can be useful for a variety of use cases

fiatjaf:
@brenzi you could just say "this person attended a meeting on day x" and then people browsing that knew what Encointer was would realize that the last meeting they attended was 5 years ago so that doesn't mean anything.

fiatjaf:
I think this NIP works better if badges don't have metadata in them.

If I am issuing the same badge to a bunch of people it doesn't make sense for me to write the description and image in all these badges. And worse: as time passes I may want to change the description and images of badges I've given in the past, which is desirable.

I think this works better if I do

- `kind:30009`: "define badge", an event with description, images, can be replaced to be updated
- `kind:8`: "badge award", an event in which I give a badge to people, contains an `a` tag (see nips.be/33) referencing the badge
- `kind:30008`: "profile badges", event users can use to decide what badges they "accept" and want to display, contains multiple `a` and `e` tags referencing the pairs of `kind:8` and `kind:30009` above

brenzi:
> @brenzi you could just say "this person attended a meeting on day x" and then people browsing that knew what Encointer was would realize that the last meeting they attended was 5 years ago so that doesn't mean anything.

Agreed, this can be solved client side. My preference would still be that the badge somehow can be marked as perishable with a TTL because the cycle period could change and clients would have to look up the lifetime of those badges dynamically, which I think is unnecessary overhead. No doubt we can find a workaround for our use case though.

fiatjaf:
@brenzi there is a NIP for generalized event expiration already. That can be used for badges too.

fiatjaf (approved):

ok300:
I assume there's no way to revoke an awarded badge, like with a replaceable event, since the awarding is `kind:8`?

The use-case I'm thinking of is badges for membership levels, like Premium Supporter, Platinum Member, etc. These can be contingent on regular payment (badge re-issued every month), or on good behavior in a community (badge issued once, but may be revoked at any point in the future if necessary).

> there is a NIP for generalized event expiration already. That can be used for badges too.

General event TTL expiration wouldn't work for that, you'd have to decide in advance how long this badge will live. Even so, it's impractical because each time it's renewed (re-awarded), the recipient has to update or re-issue the `kind:3008` Profile Badge event to point to the new award.

Or am I missing something? Is there a simpler, more robust way to revoke badges?

fiatjaf:
Do not bother about expiration time, just make badges with dates in them and write in them that they are valid only for a month. Then issue a new badge every month.

weex:
@ok300 If you feel this is still a problem, perhaps it's best to create a new issue to discuss since this PR is already merged.

PR #230, opened 2023-02-07
NIP-66 Relay Discovery and Liveness Monitoring (Draft 7)

dskvr:
`draft7` 

[Rendered NIP](https://github.com/nostr-protocol/nips/blob/cc3d7c63f59631810032c14d1f56ff034084583e/66.md)

**tl;dr**
A flexible ~parameterized replaceable~ addressable event featuring implicit, subjective liveness detection.

**Implementations**
- `NIP-66` optimized relays [1](https://history.nostr.watch) [2](https://relaypag.es) [3](https://relay.nostr.watch) [4](https://monitorlizard.nostr1.com)
- [nostrpul.se](https://nostrpul.se) _first-principles `NIP-66` demo, draft7_
- [legacy.nostr.watch](https://legacy.nostr.watch) _relay directory, draft1_
- [@nostrwatch/nocapd](https://github.com/sandwichfarm/nostr-watch/tree/next/apps/nocapd) _monitor daemon, draft7_  active monitors: [NE](https://njump.me/nprofile1qqsfhwacghjmdjp3c2tcnyq8dxzr4dpmk5z840nf0pcvk59kljdljgcpzamhxue69uhhyetvv9ujumn0wd68ytnhv96xx6qunp0y6) [BR](https://njump.me/nprofile1qqsfhgwh3ykdq4l44jjav2d95cqlvj7rurclcmkeeyucgh396hs7y4qpzamhxue69uhhyetvv9ujumn0wd68ytnhv96xx6qtcr78h) [USA-W](https://njump.me/nprofile1qqsfhgxw8hxz3snd5rgdslayvrrckcp2rq9kr6mskc4t5pzst33nraqpzamhxue69uhhyetvv9ujumn0wd68ytnhv96xx6qv9cq3h) [USA-E](https://njump.me/nprofile1qqsfhfjggqp736yxqruha0la39ajl6p82vyzarsvmr4pn2kqlu4hz2cpzamhxue69uhhyetvv9ujumn0wd68ytnhv96xx6qrxaqmd) [AU](https://njump.me/nprofile1qqsfhgzxmdtt3eng93y2lrtyyhl7spps50xss4xe2wq67f796f72pacpzamhxue69uhhyetvv9ujumn0wd68ytnhv96xx6q22c87m) [IN](https://njump.me/nprofile1qqsfhd7djntmdz9yqupqtk0mt6wv567hs8l8e2l8srselhfr5qmwpggpzamhxue69uhhyetvv9ujumn0wd68ytnhv96xx6qc8q2qp) [KR](https://njump.me/nprofile1qqsfhtpatrh45dx8cj5mqkc8ex8y4lzkv425ywrmf5mvn5nslzv9jtspzamhxue69uhhyetvv9ujumn0wd68ytnhv96xx6qn9lthj) [ZA](https://njump.me/nprofile1qqsfhpw4fnztezrdvpuzlqxkwmjph33ha5lvcu7jhdd2htwyn8t2xsqpzamhxue69uhhyetvv9ujumn0wd68ytnhv96xx6qzney00)
- ~[@nostrwatch/kit](https://github.com/sandwichfarm/nostr-watch/tree/next/apps/nocapd)~ _fetch helper, draft6, deprecated_
- ~[@nostrwatch/idb](https://github.com/sandwichfarm/nostr-watch/tree/next/apps/nocapd)~ _NIP-66 indexedDB, had adapter for kit, draft6, deprecated_
- [monitorlizard](https://github.com/relaytools/monitorlizard/) _monitoring agent, draft7_ [monitor](https://njump.me/nprofile1qqst9j2fcrahn64zsd7n3cl0flna2lldum8u8sq09nm4erxtmtfv3ggpremhxue69uhk6mmwd96x7unvd9axzuny9ehx7um5wgcjucm0d5qqtlv9)
- [nostr-event-deleter](https://github.com/fiatjaf/nostr-event-deleter) ([delete.nostr.com](https://delete.nostr.com)) _draft7_
- [snort relays page](https://snort.social/settings/relays) _draft7_
- [nostrudel](https://nostrudel.ninja/#/discovery/relays) _draft7_
- [next.nostr.watch](https://next.nostr.watch) _draft 7_ (alpha)
- [jumble](https://jumble.social/) _draft7_

**Kinds**
`30166` - Relay Discovery
`10166` - Monitor Discovery

**Use Cases**
1. **Gossip/Outbox/Inbox**: Sort and prioritize coalesced and grouped relay lists on various dimensions of data, such as proximity and round-trip time, for a more optimal outbox experience. 

1. **Geographic Relay Discovery**: Identify relays situated near a specific geographic location or within a particular country, facilitating localized network interactions.

2. **NIP Support Filtering**: Search for relays based on their support for specific Nostr Implementation Possibilities (NIPs), ensuring compatibility with desired protocol features.

3. **Accessibility Search**: Locate relays that are free to use, that have spam protection via payment and/or require NIP-42 authentication. 

4. **Real-Time Status Monitoring**: Utilize a status client to display up-to-date statuses of various relays, providing insights into their current operational state. For example, within a social or dedicated client, or within a client's relay list. 

5. **Relay Network Analysis**: Analyze connections and patterns between relays using various attributes, aiding in understanding network topology and security assessments.

6. **Error Detection in Relay Lists/Set**: Clients can detect and rectify problematic entries in relay lists; for example, notify a user that relay on their list has been offline for `n` months. 

7. **Performance Benchmarking**: Sort relays based on performance metrics like round-trip times and uptime, aiding in the selection of the most efficient relays for specific needs.

8. **Language and Content Filtering**: Identify relays catering to specific languages or content types, enabling users to engage in a more targeted and relevant social networking experience.

**History**
- _**draft1** implemented in `January 2023` and proposed `February 2023`. "Too many one-letter indexable tags"_
- _**draft2** experimented `March 2023` "Everything should be expirable"_
- _**draft3** experimented, never proposed_ 
- _**draft4** experimented, never proposed_ 
- _**draft5** proposed `December 2023`, "Needs to be indexed", heavy reliance on NIP-32_
- _**draft6** proposed `February 2024`, "Just use one-letter indexable tags" (lol) and split Discovery and parse cases._
- _**draft7** proposed `July 2024`, remove Relay Meta, offload case onto existing nips with voluntary publishing of rich metadata in monitor-defined shapes._

**Usage**

Use the filters below to inspect events from the nostr.watch Amsterdam monitor.

  **relay:**
  
  ```
  relay.nostr.watch
  ```

**filter:**
  ```
  filter: {
    "authors": ["9bbbb845e5b6c831c29789900769843ab43bb5047abe697870cb50b6fc9bf923"],
    "kinds": [30166],
    "since": Math.round(Date.now()/1000)-(60*60*2)
  }
  ```
  
  Filter against search relays (NIP-50)
  ```
  filter: {
    "authors": ["9bbbb845e5b6c831c29789900769843ab43bb5047abe697870cb50b6fc9bf923"],
    "kinds": [30166],
    "#N": ["50"],
    "since": Math.round(Date.now()/1000)-(60*60*2)
  }
  ```

  Filter against community relays (NIP-29)
  ```
  filter: {
    "authors": ["9bbbb845e5b6c831c29789900769843ab43bb5047abe697870cb50b6fc9bf923"],
    "kinds": [30166],
    "#N": ["29"],
    "since": Math.round(Date.now()/1000)-(60*60*2)
  }
  ```

Filter against free relays
  ```
  filter: {
    "authors": ["9bbbb845e5b6c831c29789900769843ab43bb5047abe697870cb50b6fc9bf923"],
    "kinds": [30166],
    "#R": ["!payment"],
    "since": Math.round(Date.now()/1000)-(60*60*2)
  }
  ```     
  
  Filter against paid relays
  ```
  filter: {
    "authors": ["9bbbb845e5b6c831c29789900769843ab43bb5047abe697870cb50b6fc9bf923"],
    "kinds": [30166],
    "#R": ["payment"],
    "since": Math.round(Date.now()/1000)-(60*60*2)
  }
  ```  

Filter against relays without auth
  ```
  filter: {
    "authors": ["9bbbb845e5b6c831c29789900769843ab43bb5047abe697870cb50b6fc9bf923"],
    "kinds": [30166],
    "#R": ["!auth"],
    "since": Math.round(Date.now()/1000)-(60*60*2)
  }
  ```     
     
These examples hard-code an established NIP-66 monitor in the `authors` filter for demonstration purposes. A more robust implementation may dynamically find monitors by finding `10166` events and selecting them based on various criteria. 

Another implementation may only care about self-published `30166` events from a relay's operator (`satellite` relays self-publish `30166` events for example)
     
An even more robust implementation may encourage their users to submit reports, and utilize a user's web of trust to find relays.
     
There is a wide range of implementation possiblities, from simple to complex.


mikedilger:
I get the idea and I think it's a good one.

First, shouldn't tag names should be globally unique across all event kinds?  People may search for events with certain tags without sub-specifying an event kind.  So using 'r' for read collides with using 'r' for reference.

Second, not sure these tags all need to be searchable.

Third, should normalized URLs have a trailing slash? Have we settled on this elsewhere?  My code for example normalizes without a trailing slash, but I don't know if that is good or not.

I may have more comments later, I didn't try to be complete here.

dskvr:
> First, shouldn't tag names should be globally unique across all event kinds
I'm not aware of any such limitation, that would mean only 26 one-letter tags for all of nostr indefinitely. 

> Second, not sure these tags all need to be searchable.
Agreed

> Third, should normalized URLs have a trailing slash? 
Normalized URLs have a trailing slash, yes. It's a **should** not a **must**

mikedilger:
Ok but what if I search for all event with a "p" tag that has my pubkey, without specifying an event kind?  I might get events where "p" was defined to mean "pedestrian" or "positive" or "playground" or anything else.  That would kinda suck.

barkyq:
My attempts at critiques follow below. Hopefully you can successfully defend your NIP idea against me :) 

Seems like this can be generated client side. E.g., a client which somehow gathers the "top" relays (perhaps by fetching relay list events from NIP-65), and then opens websockets to those relays and generates response time and the NIP-11 documents etc. Then the client can display these results back to the user.

Most of this info (SSL, IP4, IP6, topics, ability to read/write, is down, location) seems like it could be determined by a purpose built web client and/or served by the relay (i.e., the relay owner set the topics/location in some NIP-11 extension?)

As I am typing this, I am realizing that a client-side approach will perhaps be computationally expensive, and a lot of this info does not need to be constantly re-computed (like the server "location"). And it would theoretically be nice to do a query by server location and receive a list of relay URLs...

But still... I do not think that the information here will actually be that useful for people deciding which relays to use. Rather people will be drawn to relays which "perform" well and are related to their "social graph."

Also, agree with @mikedilger that the 26 queryable tags should not be used with such abandon. Will people really query for events based on SSL status, IP4 and IP6? I guess I lean a bit on the conservative side when adding query-able tags.

dskvr:
Seems like this can be generated client side. E.g., a client which somehow gathers the "top" relays (perhaps by fetching relay list events from NIP-65), and then opens websockets to those relays and generates response time and the NIP-11 documents etc. Then the client can display these results back to the user.

@barkyq Yes, I've been doing this with nostr.watch since around November.  The problem is, that there are large amounts of logic required to aggregate this data, logic that not every client needs to write. As there are more relays, more computing power is required to process the relays, and it becomes a memory hog client-side. 

> Most of this info (SSL, IP4, IP6, topics, ability to read/write, is down, location) seems like it could be determined by a purpose built web client and/or served by the relay (i.e., the relay owner set the topics/location in some NIP-11 extension?)

Yes

> As I am typing this, I am realizing that a client-side approach will perhaps be computationally expensive, and a lot of this info does not need to be constantly re-computed (like the server "location"). And it would theoretically be nice to do a query by server location and receive a list of relay URLs...

Correct.

> But still... I do not think that the information here will actually be that useful for people deciding which relays to use. Rather people will be drawn to relays which "perform" well and are related to their "social graph."

It is not for that purpose. It is for clients to utilize and to aid in the discovery of relays with useful datapoints. 

> Also, agree with @mikedilger that the 26 queryable tags should not be used with such abandon. Will people really query for events based on SSL status, IP4 and IP6? I guess I lean a bit on the conservative side when adding query-able tags.

I believe there is consensus on this already. I added all tags as queryable as to have the community identify which ones are valuable as being queryable. 

> Also, agree with @mikedilger that the 26 queryable tags

I have yet to find literature to confirm this is true, please correct me and provide citations otherwise. I'm also have trouble finding a use-case where it makes sense to subscribe without a kind, unless you were subscribing to a purpose-build relay.

fiatjaf:
```
"tags": [  
    ["d","wss://some.relay/"],
    ["t","nostrica"],
    ["t","bitcoin"],
    ["g","ww8p1r4t8", "Amsterdam", "NL", "EU", "Earth"],
    ["ip", "1.1.1.1"],
    ["ip", "2001:db8:3333:4444:5555:6666:7777:8888"]
    ["open","true"],
    ["read","true"],
    ["write","false"],
    ["ssl","true"]
  ]
```

What about this for the tags?

dskvr:
I added two new tags to the NIP. 
- `users` - Total number of users on the relay
- `events` - Total number of events on the relay 

These values are presently available through several services, but their data is not open. It really should be.

Additionally, the following statement is no longer relevant:

> Seems like this can be generated client side.

As generating these values client-side is not feasible (and as previously demonstrated, generating many of these values at scale is not feasible client-side, either) 

Updated example event:
```
{
  "id": "<eventid>",
  "pubkey": "<pubkey>",
  "created_at": "<created_at>",
  "signature": "<signature>",
  "content": "{}",
  "tags": [  
    ["d","wss://some.relay/"],
    ["t","nostrica"],
    ["t","bitcoin"],
    ["g","ww8p1r4t8", "Amsterdam", "NL", "EU", "Earth"],
    ["ip", "1.1.1.1"],
    ["ip", "2001:db8:3333:4444:5555:6666:7777:8888"]
    ["open","true"],
    ["read","true"],
    ["write","false"],
    ["ssl","true"],
    ["events", "502341"],
    ["users","37482"],
  ]
}
```

dskvr:
Assigned `NIP-66`, which as far as I can tell, is not taken. Will update the filename/squash if/when this NIP gets any traction.

barkyq:
Yes I guess the main argument against queryable tags is that the relay needs to process them much more than non queryable tags. I do agree that each `kind` is allowed to have their own interpretations of the queryable tags. Although it would be aesthetically-pleasing if the interpretations were consistent across events kinds.

Agreed that it is useful to query on the `online` status. One option is to say something like "Publishers SHOULD only publish updates for relays which are online" (so all events have an implicit `["online", "true"]` tag) and then the `created_at` filter becomes a sort of proxy for which relays which were recently online. This seems like it would be useful in practice, for people trying to discover relays. Although I can see why some would not like it, because people like to get notified when something goes offline. In any case, it is very easy for clients to decide if a given relay is online or not... simply try to establish a websocket connection.

If a relay `wss://foo.bar` goes offline, the client can make a subscription to the publishers `kind 30303 `feed filtering by `d` tag `wss://foo.bar`. When an event comes back with a new `created_at` timestamp, the client knows the relay `wss://foo.bar` is online again, at least according to the publisher.

Another possibility is to also use the NIP-40 `expiration` tag, making it slightly longer than the republish frequency. Maybe a bit too convoluted. This would reduce storage size though, and also signal to users when they can expect an update etc. Dedicated endpoints could store the events for long term and display the "expired" events as the "offline" relays.

majestrate:
> I get the idea and I think it's a good one.

in general, it does first sound great. however protocol extensions like these are "abused" and always eventually become a point of protocol dialect bifurcation.

dskvr:
> however protocol extensions like these that allow arbitrary free form data are "abused" and always eventually become a point of protocol dialect bifurcation.

Could you elaborate? Are you referring to the `eventt.content` field?

majestrate:
Could you elaborate? Are you referring to the event.content field? I feel as though this statement is extremely broad. 


yes, as provided in the changeset:

>JSON **should** be extended by NIPs.

with this `SHOULD` here, it has a risk of becoming a dumping ground for arbitrary data.

barkyq:
> Yes it is. It's very easy to check if a relay is online if you already know the relay, but not if you want to find all relays that are online very quickly, and without adding additional business logic to your client.

Yes, agreed. But that is why I think publishers should only publish events for relays which are online. E.g., I get a list of relays by querying kind 30303, and I implicitly assume the recently published events are for online relays. If I try to connect to one of the relays in a `d` tag, and it is offline, I just drop it and continue to the next kind 30303. Not a big deal. I only connect to the relays that I would actually want to connect to, i.e., I don't need to bother "checking" every relay returned by kind 30303.

> No need, this is a parameterized replaceable event and is replaced by the next event. The footprint of 1000 30303's is somewhere around 1mb...

Yes, understood. But still, `expiration` tag has a value beyond reducing storage size. If I see `created_at = 1600000000` and `expiration = 1600003600`, then I know that the publisher anticipates to update this relay within 3600 seconds. I think this is useful information (and it also does reduce storage requirements for relays receiving these `kind` 30303). As mentioned above, dedicated backends for storing this kind 30303 could store expired events to perhaps show users "historical" relays which are offline. Returning expired events to a `REQ` would be slightly breaking the rules of the `expiration` tag. But a dedicated web backend would be allowed to show them to users of the website.

dskvr:
> as a rule of thumb, when designing protocols, places that people can tack on arbitrary junk data will always have people > tacking on arbitrary junk data. this tends to be very hairy if not kept in check with constraints. i think that is something to > keep in mind here.

Yes, this is a good point. The purpose of this was left vague because I didn't want to bloat this spec. There are data sets where `tags` are not appropriate because there's data hierarchy. I'll think about a solution to this. 

> with this SHOULD here, it has a risk of becoming a dumping ground for arbitrary data.

`event.content` fields are not validated by relays, so **should** cannot be **must**.

dskvr:
> If I try to connect to one of the relays in a d tag, and it is offline, I just drop it and continue to the next kind 30303. Not a big deal. I only connect to the relays that I would actually want to connect to, i.e., I don't need to bother "checking" every relay returned by kind 30303

Yes, this makes sense for the *client persona*. I'm looking at this from several different perspectives though. 

>  As mentioned above, dedicated backends for storing this kind 30303 could store expired events to perhaps show users "historical" relays which are offline. 

Well, you've touched on my next NIP proposal, but that's for another day. 

> This is solved by using the expiration tag, because it is a proxy for the publisher's update frequency. Do not filter on since, unless you specifically want that.

Didn't think of this, good point. 

> Agreed, but quite niche, and should not be something that normal "relay searching clients" are concerned with. This could be handled by a dedicated backend storing expired events, as detailed above

I need to stress, what we consider "normal" today due to the present network effect of nostr may just be one case of many  in a year. For instance, this is not an edge case for me. 

Lets look at the pros and cons:
Pros (expiration) 
1. More elegant
2. Less data on relays
3. Less parsing by clients 
4. Ability to filter without additional indices. 

Cons: 
1. Inability to query all relays, regardless of expiration. There is exorbitant logic and maintenance required to get these results. 
2. Loss of potentially valuable data
3. Less linear control-flow for the processing of events. 

(existing schema) 
Pros
1. Explicit values 
2. Persistence of potentially valuable data

Cons
1. Cannot be filtered without repurposing topics.
2. Data bloat

So it does seem that expirations have more benefits than the existing specs. I think I would be fine with offloading the historical data into another NIP I've been working on. Also, it would clean up the separation of concerns a bit more. `30303` would be for timely data only. 

Let me sleep on this.

majestrate:
On Sun, 12 Feb 2023 15:18:14 -0800
Sandwich ***@***.***> wrote:

> > as a rule of thumb, when designing protocols, places that people
> > can tack on arbitrary junk data will always have people > tacking
> > on arbitrary junk data. this tends to be very hairy if not kept in
> > check with constraints. i think that is something to > keep in mind
> > here.
> 
> Yes, this is a good point. The purpose of this was left vague because
> I didn't want to bloat this spec. There are data sets where `tags`
> are not appropriate because there's data hierarchy. I'll think about
> a solution to this. 

i feel like multipart mime messages should have tought everyone the
lessons of why allowing free form, vauge but sort of structured data can
be an issue.

> 
> > with this SHOULD here, it has a risk of becoming a dumping ground
> > for arbitrary data.
> 
> `event.content` fields are not validated by relays, so **should**
> cannot be **must**. 
> 

fair.


-- 
~jeff

dskvr:
[majestrate](https://github.com/majestrate)

>  `event.content` fields are not validated by relays, so **should** cannot be **must**.

After sleeping on this, the most I could do is "`event.content` fields **should** be empty. The nostr protocol, by design, allows for flexibility of the `event.content` field. Therefor the following statement

> places that people can tack on arbitrary junk data

Is out of the scope of this NIP entirely. 

So the question is, is it better to provide a guidance to encourage standardization, or attempt to discourage use of the field entirely. 

I tend to lean towards the former because then at least there is stated incentivisations, which humans tend to respond to better than disincentivisations.

Thoughts?

majestrate:
On Mon, 13 Feb 2023 02:58:46 -0800
Sandwich ***@***.***> wrote:

> [majestrate](https://github.com/majestrate)
> 
> >  `event.content` fields are not validated by relays, so **should**
> > cannot be **must**.
> 
> After sleeping on this, the most I could do is "`event.content`
> fields **should** be empty. The nostr protocol, by design, allows for
> flexibility of the `event.content` field. Therefor the following
> statement
> 
> > places that people can tack on arbitrary junk data
> 
> Is out of the scope of this NIP entirely. 
> 
> So the question is, is it better to provide a guidance to encourage
> standardization, or attempt to discourage use of the field entirely. 
> 
> I tend to lean towards the former because then at least there is
> stated incentivisations, which humans tend to respond to better than
> disincentivisations.
> 
> Thoughts?

I agree, do we have a way to nudge people towards standardizing
extensions? perhaps having a suggested maximum number of non standard
extensions advertised at a time, or some other kind of less rigid nudge.

>  
>  
> 



-- 
~jeff

dskvr:
> I agree, do we have a way to nudge people towards standardizing
extensions? perhaps having a suggested maximum number of non standard
extensions advertised at a time, or some other kind of less rigid nudge. 

No, can only be encouraged. An example extension is `NIP-11` info. Might seem redundant, but for a client it means they don't need to do a fetch; it will be more compatible with their existing business logic. Each nip-11 field could be tags, but this gets tricky with key `tag[1]` values as they could become long. Additionally, it would be annoying to update the spec everytime NIP-11 was updated. But we shouldn't discuss the justification of the above, just an example.

dskvr:
@barkyq I'm going to be updating the NIP based on what we have discussed. 
@majestrate I've gotten input and there's nothing blocking a NIP from validating `event.content` is empty, just hasn't been done yet.

majestrate:
i reviewed this NIP a bit more with the benefits and drawbacks presented here. having`event.content` as optional free form data does seem like the lesser of evils here. please disregard my previous feedback, i was mistaken.

dskvr:
The following changes have been pushed: 
1. Use of NIP-40 expiration
2. Updated language to be in line with aforementioned change
3. Removal of "online" tag
4. Updated examples.

fiatjaf:
Looking good.

dskvr:
I do question the usefulness of `g` being an index. It would not be possible to filter the events based on proximity, unless the client generated tens of thousands of geo hashes within a radius of a target hash, and then filtered that way.

igmaat:
> So "continent" or "country" would probably be a better value to index than a geohash

There are only so many reasons why consumers of these proposed `30303` events may be interested in the geographic location of relays. From the perspective of a nostr user making a decision about relay selection or adjustment, the most obvious are:

- network distance, which is latency
- considerations about language (or likely language)
- considerations about legal jurisdictions

Specifically excluded here is "where's the people I want to follow", since that is a problem dealt with by NIPs 1, 2, 5, 65.

Continent designation doesn't mean much for any of those. Having only country-level metadata covers all three to a semi-reasonable extent, even though imperfectly.

So, some suggestions:

1) add optional tag for country
2) add optional tag for region/state
3) add optional tag for city
3) add optional tag for language

Language could be a useful addition to consider. It's meaning would be an indication of the expected/intended predominant language on the relay.

Arguably, a simpler scheme for this metadata is preferable to a more complicated one. In practice, the presence and accuracy of this metadata will depend on the discretion of the publishers of these `30303` events.

Furthermore, this scheme (except language) is also friendly for GeoIP databases, which commonly contains at least country/region/city resolution. If some publisher of these `30303` events cared particularly about the accuracy of the geographical metadata, it is rather trivial to derive the country/region/city details from GeoIP databases.

> there is no existing "Region" standard I have found...

Exhaustive standards do exist. ISO3166 in general, [ISO 3166-2](https://en.wikipedia.org/wiki/ISO_3166-2) in particular. Whether the latter is actually practical for this context, I'm doubtful.

Aside from the geographic metadata, there's another detail that could benefit from further clarification & discussion.

Provision is made for IP addresses. The NIP does not state what the technical purpose of those IP addresses are, nor what potential consumers of these events should (and should not) do with them. The event IP address may not always agree with the relay's DNS-resolved address, e.g. because A-record changes or other reasons.

dskvr:
@igmaat 

> ISO3166

Yes. Hmmm, my explanation was wrong. I use country codes exclusively in my client, I'm referring to standardization for continental regions, not countries.

> if some publisher of these 30303 events cared particularly about the accuracy of the geographical metadata, it is rather trivial to derive the country/region/city details from GeoIP databases.

Logically trivial, but not logistically. To implement isn't terribly difficult, but it does require additional services, some of which may be free but rate limited. Additionally, the varying transport methods and formats APIs can bloat the client and introduce undesirable patterns. Alternatively, when stored on nostr the client can leverage familiar patterns. A more advanced client would reduce the `30303` events to a subset (example: sort latency ascending, shift(5)) and then run the computationally expensive checks. Reduces load time and ingress. For example, many of these checks are "trivial" and "fast,"  but not for slower relays. There are many relays. It's a lot of data to transmit. It's a lot of clock-time to consume. Some clients are using capped data plans, one event is both faster and less data over the wire [history/archival relay] than hitting both a DOH service and a geo-ip API or dataset. 

> The event IP address may not always agree with the relay's DNS-resolved address, e.g. because A-record changes or other reasons.

This is also true, which is why multiple publishers can be helpful. Event data can be self-descriptive. `event.content` should be extended to check conditions.

dskvr:
Had to tthink about some of your other suggestionos

>> add optional tag for country
add optional tag for region/state
add optional tag for city

What about something like `["g", "DE", "country"]` or **`["g", "ww8p1r4t8", "geohash"]`**

>> add optional tag for language

I suppose publishers with crawlers could do language detection and append several languages that were detected as tags.

igmaat:
> [geoIP lookups] Logically trivial, but not logistically. To implement isn't terribly difficult, but it does require additional services, some of which may be free but rate limited

I mentioned GeoIP lookups relevant to a specific context: some "publisher" of these `30303`'s who cares about the accuracy of the geographical metadata. Secondly, because even local-lookup free datasets like MaxMind's GeoLite dataset furnishes the data for the simplified scheme I suggested: country, region, city. See [this](https://www.npmjs.com/package/geoip-lite), for example.

> A key could be added to tagIP[2], when parsed signifies the order of IP during DNS check.

Sure. Maybe I'm missing something obvious, but I'm still not clear on the technical utility/purpose of having IP's included in these events, given that the (mandatory) relay address is already the definitive, canonical network location. Would you mind sharing your thinking around including the IP's in these events?

> What about something like ["g", "DE", "country"] or ["g", "ww8p1r4t8", "geohash"]

I think this scheme is a definite improvement. Particularly because it allows geo items to be populated on a best-effort basis, depending on what is available to a given publisher of these `30303` events. And without leaving potential blanks in a fixed-format, single tag scheme.

> I suppose publishers with crawlers could do language detection...

That's a possibility. I figured a language tag would be useful, because I anticipate that as the nostr network continues to grow, many more nostr relays will come online which are regionally-focused.

@dskvr do you imagine that nostr.watch and other such services related to nostr network/relay health would come to act as _both_ publishes and consumers of these events?

Lastly, it's worth considering an amendment/extension to NIP11, so that some of the useful metadata proposed in this NIP may optionally be provided as part of the relay information document. The reasoning here is to give relay operators an authoritative means to advertise these details to the network, which could then in turn be read by parties interested in publishing these `30303` events.

dskvr:
> nostr.watch and other such services related to nostr network/relay health would come to act as both publishes and consumers of these events?

Yes. But they aren't the only, nor the most important persona that would be interested in this data. The ability to quickly aggregate details about relays is becoming important in client development. nostr.watch already uses `30303` events and has for roughly a month (though, am right now technically using `30304` due to a relay bug I'm waiting to be resolved) 

> Would you mind sharing your thinking around including the IP's in these events?

They are data-points that can bring context to the results. Publishers will have different results, the delta between these datasets can be useful. 

What are your reservations about IP data in these events?

igmaat:
I agree that the comparison of deltas in the IP metadata of relays can be useful. Like, for instance, detecting "anomalous" DNS resolution behaviour.

> What are your reservations about IP data in these events?

Not reservations as such. I was thinking through potential scenarios of adversarial/malicious abuse (fabrication) of the metadata in these events - not just the IP's, but all metadata except the actual relay address. Two cases in particular, with the hypothetical assumption that client consumption of these events becomes commonplace in the future:

**Scenario A**  
An untrusted/new/random publisher (pubkey) floods the nostr relay network with `30303`'s

**Scenario B:**  
A bad actor compromises the private key of a publisher which is considered "reliable" by some number of clients that leverage `30303`'s for efficient relay aggregation/selection.

And my take away here is that any potential weirdness from such scenarios would be highly dependent on **what** a particular client may choose to do with these events.

dskvr:
@igmaat Yes, understood. I believe the NIP clearly addressed these scenarios at it's genesis. 

> ## Consumption
The data in `30303` **may** be erroneous, intentionally or otherwise. Where accuracy is required, the data in `30303` events **should** be subscribed to by populating the `authors` filter array with the pubkeys of trusted **publishers**

gzuuus:
Hi, sorry to join the conversation, I'm just here to talk about the kind you want to use, the 30303. I'm using the same kind for my client nostree.me . I wanted to know what is the status of implementation of this proposal to decide whether to change the kind I use in my client or discuss about it. Thanks in advance

dskvr:
@gzuuus I have been using `30303` in production for roughly 7 months.

fiatjaf:
@v0l what do you think of this for the Snort relay recommendation thing?

v0l:
Too much for me, all im looking for is relay metadata but on nostr, if relays could somehow bootstrap themselves onto the network by publishing their own relay information then relay discovery is very easy.

For example:
```json
{
  "kind": 10011,
  "content": "",
  "pubkey": "<relay-pubkey>",
  "tags": [
    ["r", "wss://relay.snort.social"],
    ["desc", "Snort subscribers relay"],
    ["g", "u09tv"],
    ...other tags relating to NIP-11 relay document
  ]
}
```
In terms of liveness and general health, i dont know if that information is really useful for everybody to have available on nostr.

v0l:
I just deployed a page also to view this data: https://snort-relays.pages.dev/

dskvr:
@v0l Thanks for sharign! I'm well versed iin using events for this purpose, as I've been doing it for 10 months. Self-reporting isn't different from this NIP. While it was not included in the language originally, it is not it was discussed. 

What you're describing is the difference between Yahoo Directories/DMOZ and a web crawlers, the reason web crawlers prevailed over self-reported directory structures, is that relying on self-reporting generally results in unreliable datasets.

staab:
I'd stick with a single `30066` event for a first version. This can serve for both self-reporting and crawler reporting. I don't think the reporter metadata event adds very much useful information, and 1066 can be covered by simply retaining 30066 events. I'd go a little lighter on indexable tags, for example `ipv6` probably doesn't need to be searchable. Concept ACK though.

dskvr:
Thanks for the feedback!

A few notes:
- `1066` is something that I've found through usage is required, because different data points have different aggregation patterns and publishing intervals, and therefor are most simple when the processes that produce these data product different events; ie, are from different monitors (pubkeys). While it's certainly possible to do it with the same keys, it greatly complicates implementation on both sides of the equation (consumption and production of events)
- `10066` is actually a simplification of `30066` as it just allows a relay to self-publish and not worry about implementation details it should have to, such as normalizing their own url and setting their pubkey as the `d` tag. 
- `1066` is significantly more space efficient than `30066` because it only stores deltas. For the past year I've been doing more or less what you've described by storing `30066` events, and other than consuming conservatively 100x more space, it's also significantly more fragile and less accessible.

hzrd149:
This looks good for my use-case. using the rtt tag should allow me to find the "best" relay for a user and adding better relay hints into events.
I also like the addition of the "count" and geohash tags to allow clients to show a lit of the most popular or closes relays

vitorpamplona:
On the ipv4 tags, Someone and I were playing around with a DNS-like event type for Nostr. An `nevent` that points to a kind `30053` would redirect to the relay IP in the same way DNS works. Similar to NIP-65, the idea was to send these DNS events to as many relays as possible. 

I believe nos.lol still updates their records: 

```js
{
  "content": "",
  "created_at": 1690383291,
  "id": "a5c234eda701518192656667eb710e39e745a7a14c7420f5fdb410738bd52d1a",
  "kind": 30053,
  "pubkey": "1ec6810f1ae144be134b9f34ab09336ab2cd2027294aaf26ea069303687eb097",
  "sig": "54f1291dfcb37528e586dac9475c8059c1addbc8150f1fb24fbad31721dec24746ae312777b707524242bf14560ec94db086ea047befb2fdcef704a0f3d404d5",
  "tags": [
    [
      "d",
      "e.nos.lol.nns."
    ],
    [
      "ip4",
      "5.161.158.164"
    ],
    [
      "version",
      "1"
    ]
  ]
}
```

dskvr:
Don't mind using another kind, but seems to make sense to store this data
in NIP-66, which has been actively propagated for over a year.

Are these DNS events defined  in a NIP? Can't find it...

vitorpamplona:
> Are these DNS events defined  in a NIP? Can't find it...

No, we were just playing with a way to fully replace all types of DNS records with a single Nostr event. It would work for any server, not only relays.

dskvr:
@vitorpamplona got it, would be awesome if these DNS events were defined somewhere.

vitorpamplona:
I don't know what is the goal of this ipv4 tag, but if it is to replace the DNS resolving, we can try to use a complete DNS record.

dskvr:
@vitorpamplona Described in detail [here](https://github.com/nostr-protocol/nips/pull/230#issuecomment-1873403475)

> ipv6 and ipv4 can help identify parent-child relationships between "relays" that share the same ip. A good example here is nostr.band feeds (common domain, often same IP) vs nostr1 relays (common domain, usually different IP). There are only a few objective datapoints that enable automatic relationship detection, and IPs are one of them. Yes, these relationships can be identified locally, and in some cases are self-reported, but as the number of relays and the softwares that runs them grows, an upper limit will eventually be reached, requiring an ability to filter on these values for accessibility. No client in their right mind would have an ETL for these events.

dskvr:
Added NIP-11 integer support via non-indexable tag.

dskvr:
Added `draft` and filters on PR OP.

dskvr:
`draft6` implementation: https://sandwichfarm.github.io/nip66-dataviz-demo/

verbiricha:
I'm using Relay Discovery (kind 30166) in [Nostrrr](https://nostrrr.com) instead of the nostr.watch API to find a list of online relays from nostr.

My usage so far is very basic but given this spec it is trivial to:
 - show more information about online relays and allow more granular filtering than by URL
 - allow users to choose a different Relay Monitor (kind 10066) as data provider
 - use Relay Metadata (kind 30066) events on relay details page to add more information about a relay than just the NIP-11

Looks great for my use case.

dskvr:
Need some ideas for monitor `o` tags, as proposed, they don't really work as spoofing is trivial. I have some ideas, but don't think any of them are good. 

Some notes: 
- For `o` descriptor using the term `operator` over `owner` as this terminology is more intuitive, accurate and public domain friendly.
- While theoretically, most "operators" would only run 1 monitor, larger scale monitors would run many. In the case of `nostr.watch`, I'll be running 7 geographically distributed monitors that only check round-trip times, and then 6+ more that publish varying resolution of meta data (one monitor for each meta-type, and one that publishes all meta in a single event). The 6+ meta monitors would exist to optimize bandwidth for clients, if you only want NIP-11 details, you don't need to stream excess data. In terms of efficiency, those 6 monitors would run off one instance, and have expiration tags. 

1. Proofs - The `o` tag could contain a cryptographic proof at position 2 that is the pubkey of the monitor signed by the operator. Not good, because arbitrary signatures don't have widespread support.
2. Bi-directional relationship: The relay meta/discovery events could have an `a` tag (in addition to the `o` tag) that references a generic list published by the operator, that list would the pubkeys of the monitors it operates. Not great, because it would be difficult to implement and has too many hops, so nobody would actually utilize it.
3. Uni-directional reference. The operator publishes a list of monitors it operates. Like the `draft6` solution, this is also easily spoofed. However, because the `operator` would likely be well-known and in the web-of-trust, the operator would likely be the canonical reference to relay monitor events, and thus relay discovery and relay meta events. Therefor, attempts to spoof these generic lists would be a fools-errand, as the list would be discovered via the operator's pubkey. 

The purpose of the `o` tag was to reduce hops, dependencies and logic for subscribing to `30166` and `30066` for operators, but ultimately, it's implementation is convoluted and in the end, may be entirely unneeded. 

I personally opt for `#3` here, as it is the most aligned with protocol, externalizes complexities to the web-of-trust and is dead-simple. 

**draft7**
This draft will be mostly additive, less any changes to `o` tag mentioned above. 
- The NIP language will be streamlined, propaganda will be placed into a NIP-33 article and a reference to the event left in the NIP itself or in this thread.

jeremyd:
Hey all, I know I'm late to the party on this NIP but this is looking pretty cool.

I am brainstorming right now, what would a user of nostr ideally want from the "discover" events based on user feedback and just general nostr usage.

They'd want to know what relay has the best latency from their location to the relay's location.  This already seems possible by doing some geo calculations or simply by selecting country.

They'd want to categorize relays by features or keywords.

Examples off the top of my head:
DVM relays
Private Inbox Relays
Search relays
purplepag.es (metadata relays)
relays with specific content or community name as a keyword (ie: christian relays)

So I was wondering if we could expand the 30166 kind to address these categories/keywords by adding category and keyword tags to the event.  Additional category brainstorming is happening here w/ @vitorpamplona https://github.com/nostr-protocol/nips/issues/1282 and I think we may be reaching a stage where it makes sense to come up with additional fields for both NIP11 and NIP66 to facilitate relay discovery in clients.  I envision these discoverability events will exist on more than just a few specialized relays, they can be broadcast by the relay operation team (or anyone) to a wider range of public relays and clients will be able to crawl these to show the user a directory of relays they might be interested in without having a centralized source.

dskvr:
> They'd want to categorize relays by features or keywords.

Yes. 

> They'd want to know what relay has the best latency from their location to the relay's location. This already seems possible by doing some geo calculations or simply by selecting country.

Yes. 

> So I was wondering if we could expand the 30166 kind to address these categories/keywords by adding category and keyword tags to the event. Additional category brainstorming is happening here w/ @vitorpamplona https://github.com/nostr-protocol/nips/issues/1282 and I think we may be reaching a stage where it makes sense to come up with additional fields for both NIP11 and NIP66 to facilitate relay discovery in clients.

Absolutely, yes. II even [commented on this issue](https://github.com/nostr-protocol/nips/issues/1282#issuecomment-2154527024) a while ago

jeremyd:
Very cool.  I've got my first cut of a monitoring agent live, running on a raspberry pi, that I will be using to publish these events and a relay that accepts only these events and related @ wss://monitorlizard.nostr1.com

https://github.com/relaytools/monitorlizard

I will be adding the relay discovery events and better geotags etc.

@dskvr I assume you probably don't want other monitors publishing to history.nostr.watch?  It didn't seem to want to let me :)

dskvr:
That's awesome! You clearly understand the spirit of this NIP. You're the
first community monitor I am aware of. Another relay and monitor
implementation is clutch for testing and client-side development. thank you.

I locked down history.nostr.watch after an attack about a year ago and
havent changed much because there werent any other montors. I am planning
to migrate to another relay for performance reasons soon, and it would
accept monitor events from anyone.

I'm calling it a night but will reach out tomorrow.

jeremyd:
I'm doing a side-by-side comparision with NIP11 and keeping in mind what clients might want to see in kind 30166 so they can display maximum info to a user, without overloading the event tags too much.  Any recommendation on what the names of the tags should be or etc?  Indexed/vs non indexed.  Use the content field?  I could submit a pull req to this branch after a little feedback on these.

I think the bare minimum additions so far, in order for any client to display a relay's info without additional queries would be,

- [ ] icon
- [ ] payments_url
- [ ] posting_policy
- [ ] description

dskvr:
`30166` is for discovery, none of those items are particularly useful for filtering against.

Those fields could be added to `30066` meta events. Note that `30066` events can also have any of the tabs from `30166`

Reasoning:
1. Keep bandwidth low for discovery case.
2. Allow metadata to be discoverable.

jeremyd:
> `30166` is for discovery, none of those items are particularly useful for filtering against.
> 
> Those fields could be added to `30066` meta events. Note that `30066` events can also contain any of the tags from `30166`
> 
> Reasoning:
> 
> 1. Keep bandwidth low for discovery case.
> 2. Allow metadata to be discoverable.

Ahh, makes total sense.  My thought was that to keep bandwidth lower I wasn't doing the stringified NIP11 in the content.  But this is just as easy and won't clog up the tags, so I will add support for this in my tools.

dskvr:
draft 7 updates: 
- added `T` for relay types in `30166`
- added `s` for relay software in `30166` 
- moved `rtt` with different key signature to `30166`
- Removed `30066` to simplify NIP and subsequently its implementation. Relay Meta data that is not specified in NIP may be published by utilizing other existing NIPs and autonomously communicating the data's shape

vir2alexport:
Can you extend 30166 kind event, so users of https://relays.xport.top for example be able to share their results with the network?

`q` tag stands for quality and inverse-proportional
```
[
	...
	"q": <number-of-connections>,
	...
]
```

`content`
```
{
	<relay-1-addr>: { "rtt-open": <rtt-open>, "rtt-read": <rtt-read>, "rtt-write": <rtt-write> },
	...
	<relay-n-addr>: { "rtt-open": <rtt-open>, "rtt-read": <rtt-read>, "rtt-write": <rtt-write> }
}
```

dskvr:
@vir2alexport completely breaks the `30166` kind format and how NIP-66 functions in a wider context. Each relay gets its own event, which is how relays remain discoverable. If multiple relays are included in each event, not only would the events be too large, but it would not be possible to discover relays.

You can instead add counts using NIP-32 `l` tags

For example

`['l', '12345', 'count.events-per-hour']`

vir2alexport:
@dskvr I see, thank you. So this is good for liveness monitoring rather than discovery in my case.

dskvr:
This is good for both liveness and discovery.

66.md
```diff
@@ -0,0 +1,250 @@
...
+
+#### `created_at`
+The `created_at` field in a NIP-66 event should reflect the time when the relay liveness (and potentially other data points) was checked.
+
+#### `tags`
+
+##### Meta Tags (unindexed)
+- `rtt-open` The relay's open **round-trip time** in milliseconds.
+- `rtt-read` The relay's read **round-trip time** in milliseconds.
+- `rtt-write` The relay's write **round-trip time** in milliseconds.
```
v0l, line 61: How are these measured?

66.md
```diff
@@ -0,0 +1,250 @@
...
+### Summary
+`30166` is a `NIP-33` Parameterized-Replaceable Event [PRE], referred to as a "Relay Discovery" event. These events are optimized with a small footprint for protocol-level relay Discovery.
+
+### Purpose 
+Discovery of relays over nostr. 
+
+### Schema 
+
+#### Content 
+`30166` content fields **MAY** include the stringified JSON of the relay's NIP-11 informational document. This data **MAY** be provided for informational purposes only.
```
v0l, line 51: Would rather this be a **SHOULD**

66.md
dskvr: read: `time.start` subscribe, subscription fulfilled or EOSE recieved (whichever first) `time.end`
write: `time.start` send event, ok received `time.end` 

Notes:
1. Both of these values are more useful when represented as an SMA; some relay implementations are slower than others. 
2. Both of these values can confirm relay behaviors when cross-referencing other relay attributes. 
3. These values cannot be universally applied since relays treat all users differently. 
4. Write tests are particularly questionable. I'm still using the testing methodology that I inherited from previous relay checking attempts.

66.md
dskvr: OK.

66.md
dskvr: After some thought, `read` and `write` should probably be removed from this NIP. For these kinds of tests to be meaningful in the present-day nostr climate requires more context. 

I wrote an [auditor](https://github.com/sandwichfarm/auditor), similar to to mike's relaytester, and it seems clear that read checks will be useful, but with very specific parameters to highlight strengths and weaknesses between relays. 

How to propagate or aggregate this data is still unclear, but it is very clear (and has been for a while tbh) that the simple read and write checks of yester-year provide little to no value anymore. I even completely disregard them in the new nostr.watch NIP-66 client.

kehiy (commented):
can a relay publish an event with its contact pubkey and announce its liveness on other relays? then we can query most recent relays who announced themselves and new relays will be recognized as well on the network.

dskvr:
@biasesyes, relays can self-publish and this case was always considered valid. However, depending on the intended use case, this may be less useful. 

If the intention is to find relays, you'll come up with a woefully incomplete dataset and imolementation complexiry rises. If the intende purpose is to disovery relay simply by operatoe pubkey, then that is the only case that really makes sense.

When there are several crawlers, their incentive is to provide the best data. Relay monitor datasets will always be more complete than self-published data, and their incentives promote honesty. Some real life examples would be relays that advertise NIP-50 via NIP-11, but either don't adhere NIP-50 or almost entirely reject NIP-01. Most of them are case sensitive and somewhat useless because of this. Most of them are inadequate not by design but because there is no feedback loop. 

The comparison I like to use is early internet directories, such as DMOZ and Yahoo Directories. This approach of self-publishing incentivizes self-promotion, which is inherently antithetical to honesty and completedness. It also was incresibly incomplete, and difficult to maintain as websites went offline. This is also a core reason why crawlers won over self-published directories, because crawlers had incentives around distilling data to provide users with the best possible data and they inherently accumulated liveness data. While this eventually trended against users in the 2010's, the open-index nature of NIP-66 should mitigate algorothmic biases. This dataset ispotentially a foundation of other, more data rich datasets. You would be surprised how important liveness is to discovery. 

Monitors are incentivized to test a relay against their stated capabilities, whereas a relay operator is not incentivized to publish their own inadequacies. 

Arguments against relay monitors include not wanting to trust a monitor to truthfuoly broadcast data about a relay. However, there are several clear arguments against this. 

1. When considering relay monitors are just pubkeys, then it's obvious they can be in other pubkeys' web of trust. Therefor an implementer doesn't need to trust anything, they defer that decision to pubkeys. 
2. Nostr already relies on various points of trust. We trust relays will store our data. We trust relays will fulfill filter requests honestly. We trust clients will fairly serve data and verify events. The trust is counter-balanced by the inherent fluidity of nostr and ability to change rrlays and clients on a whim. That ethos is carried over in this nip. 
3. In the case of "bad monitors" this is exactly why NIP-66 was written. 2 years ago the userbase relied on nostr.watch almost exclusively for relay data, and I never liked this and thought it was a huge weakness. Over 100 repos implement the centralized nostr.watch API. Hence, as written, any number of monitors can compete with nostr.watch, provide better data or higher resolution data. Or Clients could run their own monitor and rely on their own data. This essentially creates an open index where incentives promote honesty, and censorship is inherently difficult if not impossible. 
4. Bonus: NIP-66 also alleviates probing spam because if every user from every client probes relays, the cost of running a relay goes up substantially. 

The new nostr.watch is a NIP-66 nostr client, and while defaults will serve data produced by nostr.watch so that I can ensure a level of consistency from my domain, users will have full control over who serves their data. This is no different from any other nostr client. Other monitors could just as easily run the client with different defaults or run their own if it were ever compromised in any way or nostr.watch became "evil." This is by design.

kehiy:
> @kehiy relays can self-publish and this case was always considered valid. However, depending on the intended use case, this may be less useful.
> 
> If the intention is to find relays, you'll come up with a woefully incomplete dataset and implementation complexity rises. If the intended purpose is to discover a relay's meta by an operator pubkey your already know, then that is a good case for self-published events.
> 
> nostr1.com already monitors their customers' relays via their `monitorlizard` relay monitors using their own software written in go. It has statically set relays, and would be ideal for monitoring your own relays or your own relay list.


so, considering this, isn't it logical to add it to the standard?

1l0:
NIP-66 saves the world.

dskvr:
> so, considering this, isn't it logical to add it to the standard?

I thought it was still in there, but upon review it was removed during the last wave of simplification. It is still there as it falls under the umbrella as ad-hoc monitoring, but it should be explicitly stated in the NIP. Thanks for pointing it out.

staab:
Hilarious that this hasn't been merged yet. I just checked and there are active implementations in at least jumble and nostrudel. @dskvr shall I merge this or would you like to continue to pound out drafts?

v0l:
Snort also uses this

pablof7z:
Screenshot 2025-03-03 at 9 27 41 PM

End of an era.

staab:
Nostr 2.0 is live

dskvr:
lol

PR #237, opened 2023-02-10
NIP-04: clarify how shared secret is computed

jeffthibault:
Adds note on the differences between default behavior in libsecp256k1 and how it's done in nostr.


Sgiath:
OMG, I have spent so much time trying to figure out why I cannot decode the Encrypted messages with my implementation. Finally, I switched ECDH library, and it started working. Thanks for this, even though I probably won't use it at this point.

PR #238, opened 2023-02-11
Delegation support in NIP-07

ursuscamp:
While delegation support is currently limited in clients, it will probably build as demand for more key security grows.

One of the UX pain points now is that NIP-07 compatible extensions have no ability to manage delegated identities, which means it likely must be done at the client level. This may involve copying delegatee information across multiple web clients, entering delegator private keys into web clients to sign delegation tokens, etc.

This could be done more universally at the NIP-07 extension level. Extensions could allow a user to create a delegated identity key pair, and additionally store only the delegator's public key and a signed delegation token. By returning the delegator's pub key from the `getPublicKey` method, the user could sign in with their normal feed. When asked to sign an event, the extension would just need to modify the event prior to returning it, so that the event is sent and signed by the delegatee, with the `delegation` tag.

To support this, clients would only need to: A) accept the event returned from `signEvent` wholesale, which I don't believe any major client does, or B) check for the presence of a `delegation` tag and accept the minimum fields from the signed event.

I have implemented an example workflow from the perspective of both extension and client here: https://github.com/ursuscamp/nip07-delegation-example. The code comments document step-by-step the workflow, as I imagine it.

Additionally, this is currently implemented as an experimental feature in Nostore. It is in a branch, but should make it into a build in a few days: https://github.com/ursuscamp/nostore/tree/nip26

I'm not sure if you need me to code up an example web client that supports this, but I should be able to handle it if that is necessary.

### DIFF
--- a/07.md
+++ b/07.md
@@ -4,7 +4,7 @@ NIP-07
 `window.nostr` capability for web browsers
 ------------------------------------------
 
-`draft` `optional` `author:fiatjaf`
+`draft` `optional` `author:fiatjaf` `author:ursuscamp`
 
 The `window.nostr` object may be made available by web browsers or extensions and websites or web-apps may make use of it after checking its availability.
 
@@ -12,7 +12,7 @@ That object must define the following methods:
 
 ```
 async window.nostr.getPublicKey(): string // returns a public key as hex
-async window.nostr.signEvent(event: Event): Event // takes an event object, adds `id`, `pubkey` and `sig` and returns it
+async window.nostr.signEvent(event: Event): Event // takes an event object, adds `id`, `pubkey`, `sig`, and optional `delegation` tag and returns it
 ```
 
 Aside from these two basic above, the following functions can also be implemented optionally:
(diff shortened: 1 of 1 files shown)


ursuscamp:
Formatted version: https://github.com/ursuscamp/nips/blob/nip07-delegation-support/07.md

fiatjaf:
I don't understand the need for this NIP. As you said, this doesn't require any changes in existing extensions -- and it shouldn't require any changes in clients as long as they are already using the full event as returned from `signEvent()`. This just requires an extension to do the delegation stuff.

ursuscamp:
The spec reads like the extension is only “allowed” to modify the id, pubkey and sig in the event. In practice, all of the web clients only seem to use the sig. 

So the goal was to allow adding delegation tags to the list of “allowed” things an extension can do and remain in spec.

ursuscamp:
I created a simple web client that obeys this workflow and publishes a note event:

Github: https://github.com/ursuscamp/delegation-client
In the wild: http://ursus.camp/experiments/delegation/

Of course, the only extension that supports this workflow right now (that I know of) is nostore in an experimental branch.

Here is a delegated note published with the client: https://snort.social/e/note1547fhelqtu0h4ekltjk55t3kxfpa3g45ss6q6xnz4rsm3lcsyrxqc0hl32

staab:

PR #244, opened 2023-02-12
make zaps independent from lnurl and other minor tweaks

fiatjaf:
Modifying according to suggestions here: https://github.com/nostr-protocol/nips/pull/224#issuecomment-1418302436


ok300:
Damn, was looking forward to this.

PR #248, opened 2023-02-13
[NIP-26] Minor change to make delegation token/string naming consistent

catenocrypt:
Typo fix: the delegation token spec mentioned a hash of delegation _token_, correctly it is of the delegation _string_.


barkyq:
eagle eyes

PR #252, opened 2023-02-14
NIP-57: Add amount tag to zap request

Semisol:
This prevents an attack where a zap request note may be used for claiming a larger zap was done by the zapper than the zapper intended.


fiatjaf:
I'm afraid you'll have to submit a PR to Damus first.

benthecarman:
Is the zap request even supposed to be broadcast? The NIP doesn't say

jb55:
yeah this makes sense

jb55:
ACK

fabianfabian:
> Is the zap request even supposed to be broadcast? The NIP doesn't say

I'm a bit confused by this also, 9734 (with or without amount) is already broadcast within the description tag of 9735. 9735 already contains the invoice and amount in the bolt11 tag, so why do clients need to look for another amount field? Are we supposed to broadcast the 9734 separately and for what?

PR #254, opened 2023-02-14
Create ETH address support

kernel1983:
I have implement with one client and one relay:

Client https://github.com/kernel1983/nostrweb/tree/eth_address
Relay https://github.com/kernel1983/pyrelay/tree/eth_address

Please kindly review the NIP and code.

### DIFF
--- a/N.md
+++ b/N.md
@@ -1,0 +1,48 @@
+NIP-N
+======
+
+ETH address support
+------------
+
+`draft` `optional` `author:kernel1983` 
+
+This NIP defines the alternative address system that the user may choose to use. New NIPs may  choose to support this NIP to work with the ETH compatible NFT, DID and token ecosystem. 
+
+
+## Events and signatures
+
+Each user has a secret/private key. Signatures, address, and encodings are done according to the Ethereum way.
+
+In the `event`, will change the `pubkey` to `eth_addr` :
+
+```json
+{
+  "id": <32-bytes lowercase hex-encoded sha256 of the the serialized event data, no "0x">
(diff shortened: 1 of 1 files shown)


PierreRochard:
20,000+ "crypto" NIPs would make navigating this repo challenging

fiatjaf:
You can already use ETH addresses, by the way. If an ETH address is the hash of a secp256k1 pubkey you can just take the actual pubkey and use that as your Nostr key.

ETH-aware clients can then hash the Nostr key and check Infura for the existence of that address on Ethereum before showing a "send tip NFT" button or something like that.

kernel1983:
> You can also add your ETH address as a tag or property in your `kind:0` metadata and just go from there. This will work for any "cryptos" or even Paypal addresses, if people want. There could be one single NIP accommodating this if there is interest from multiple shitcoiners.

Yes, you can do it in a tag way, but your signature can not show you have the ownership of certain NFT or Token.
I believe more people will need this.

fiatjaf:
In that case we can agree on a canonical string to sign that includes the Nostr pubkey and attach a signature of that to the tag: `tags: [["ethereum_address", "0x...", "signature"]]`

kernel1983:
1 People can choose to show their 0x addr when they wish, but two popup to sign: nos2x + metamask
2 Maybe the double signed event are too early that the relay have already deleted

It would be easy to keep the 'pubkey' as it is fixed, we can do pubkey + eth_addr. However, it would be nice to also accept eth signature. It opens to a big world (NFT, Token, Metamask users).

fiatjaf:
We can't, because that would be a breaking change.

PR #259, opened 2023-02-15
Update NIP 11 to support relay recommendations

staab:
The intention of this NIP is to create space in the protocol for specifying NIPs for more advanced use cases without forcing relays to implement them or be considered a partial implementation.

Some use cases that would be prohibitively expensive for hobby relays:

- Content recommendations and search
- Social graph analysis
- Network analytics

Without discoverability, clients will hard-code special-purpose relays, creating a centralizing pressure on the network. In addition, some relays might not be picked up by the known/hard-coded add-on services, leaving their data out of calculations.

Recommendations would allow relays to advertise complementary relays to clients. Complementary might mean that it's run by the same people, contains the same data, or is a preferred source for the given functionality. Master/slave configurations could be set up using this scheme too — master relays would not support requests, while slaves would not support publishing events.

The downside of this is that it relies on clients being smart and following these references. Someone told me about how dApps used to follow this model, but it slowed things down and developers ended up getting lazy and hardcoding services anyway.

### DIFF
--- a/11.md
+++ b/11.md
@@ -17,6 +17,7 @@ When a relay receives an HTTP(s) request with an `Accept` header of `application
   pubkey: <administrative contact pubkey>,
   contact: <administrative alternate contact>,
   supported_nips: <a list of NIP numbers supported by the relay>,
+  extensions: <a list of nip number, url tuples>
   software: <string identifying relay software URL>,
   version: <string version identifier>
 }
@@ -48,6 +49,10 @@ An alternative contact may be listed under the `contact` field as well, with the
 ### Supported NIPs ###
 
 As the Nostr protocol evolves, some functionality may only be available by relays that implement a specific `NIP`.  This field is an array of the integer identifiers of `NIP`s that are implemented in the relay.  Examples would include `1`, for `"NIP-01"` and `9`, for `"NIP-09"`.  Client-side `NIPs` SHOULD NOT be advertised, and can be ignored by clients.
+
+### Extensions ###
+
+Some NIPs may be more burdensome than others for relays to implement. Relays may advertise support for these NIPs as `extensions` by advertising other relays that support those NIPs for some or all of the given relay's events. For example, if a relay that functioned as a read replica wanted to advertise support for NIP 42 on the master relay it follows, it would set its `extensions` key to `[["NIP-42", "wss://my-master.example.com"]]`.
 
 ### Software ###
 


staab:
Here's another example:

Screen Shot 2023-02-15 at 11 45 36 AM

fiatjaf:
> The intention of this NIP is to create space in the protocol for specifying NIPs for more advanced use cases without forcing relays to implement them or be considered a partial implementation.

But this is already the intent of the normal NIPs!

staab:
Technically, yes, the `supported_nips` field exists. But just because you can find a relay that supports the NIP you're looking for doesn't mean you can find that support for a set of events you're interested in. So implementation is partial for both the relay, and the data on it.

In other words, this nip is focused not on just finding functionality, but on finding state. Picking a random relay that supports a given nip doesn't imply it has the data you want. A simple example is if you have a relay cluster with one master and multiple followers. The master supports publishing, the followers support requests, and they advertise each other.

huumn:
Rather than extensions there might be reason to begin specifying a second layer over relays. I've been talking with @brugeman about this.

Keep relays dumb - IP routers are an apt metaphor. Let aggregation queries be done in another layer (TCP-like) with its own protocol and clients can pick from "aggregators" like they do relays ... Clients can fallback to "raw" relay and offer a limited feature set if aggregators aren't specified.

As more functionality/queries get stuffed into relays, fewer people will implement and run them.

staab:
That's exactly the thought process I went through leading up to this NIP.

There are basically two options for supporting advanced functionality:

1. Add NIPs that relays can optionally implement
2. Create a second layer protocol

Option #1 violates the "keep relays dumb" rule. The more functionality that gets packed into relays, the less small relays will be able to keep up with increasing system resource requirements, and the more complex relay implementations will discourage smaller relay implementations (ala Browser Wars).

Option #2 breaks decentralization because 1. it weakens the guarantee of interoperability by creating a new protocol (or many protocols, or ad-hoc specifications, which I think is more likely), and 2. second-layer services don't fit into a homogenous relay topology. Relay hints, lists, and other places where relays are a first-class concept are incredibly valuable for allowing users to navigate the network and choose where they want to be. A second layer would need to piggyback on the relay recommendation problem somehow, in order to prevent some other mechanism from selecting gatekeepers for the network.

It seems to me the problems associated with option #2 are actually much worse than option #1. If we start with the topology of clients/relays as it already exists in the protocol and in practice, we have the hard problem of decentralization solved. Relays can then reference additional functionality, and clients can follow those references to fulfill their requirements. This has the slightly hidden benefit of improving the relevance of service discovery, since it relies on relays to tell clients which service provider will yield the best results for its users. Finding a content search service doesn't help me if it's indexing the wrong content.

Whether these second-layer services are implemented by relays or not is not actually that important. I think they should be, since a consistent API would reduce the complexity of implementation, and would keep the protocol coherent. The idea of adding new API to the protocol is already built in to NIPs, for example [`COUNT`](https://github.com/nostr-protocol/nips/pull/144).

So, the key takeaways of the approach I'm suggesting:

- Relays are first-class concepts for the nostr protocol, and for users. They solve discoverability, and user customization. They should form the backbone of service discovery.
- Service discovery is more complex than finding a service that implements an interface. Services are generally going to be stateful, which means you need to fine the one that has what you're looking for.

Here are a few other ways I've considered for solving service discovery:

- Build it into the protocol, repeating relay hints/lists with a new type of thing.
- Add-on services could advertise which relays they cover, but how would a client find the right service for its users?
- Add-on services could take relays as a parameter to requests, and fetch the data if it's not in its cache already. I think this will result in either massively scaled up service providers, or a poor cache hit rate.
- Clients could hardcode urls for services. This encourages vertical scaling, centralization of service providers, and doesn't attempt to find data that isn't indexed by the service provider.

fiatjaf:
No functionality should be stuffed, only suggested.

Relays can't really be dumb, the protocol doesn't work with relays being dumb. Relays must be smart enough to allow others to abuse them, and they also should be encouraged to be smart enough to offer a better experience to their target audience (i.e. by filtering only content that their audience will fill interesting -- or, if their audience is very general, by keeping everything and being fast). IP routers are also not dumb at all if they.

I'm yet to see a use case for indexers. The introduction of indexers seems to me to have giant centralizing effects. Clients should not be made to expect the existence of indexers, but they can very easily be made to expect just barebones relays and have UX enhancements when encountering relays that implement extra NIPs.

fiatjaf:
I think we should talk in more concrete terms here. What do you think of [NIP-50](../blob/master/50.md)? Is that stuffing things into relays? I hope no one is expecting all relays to implement that. Why can't it just fit the normal list of NIPs?

staab:
I agree indexers will have centralizing effects, the intent of this NIP is to get ahead of that, and articulate add-ons as discoverable through relays, rather than sitting in front of them.

All functionality for relays is optional, so all NIPs can be considered add-on services. That's why NIPs specifying sophisticated add-ons should keep the same relay protocol, since there's no fundamental difference between simple functionality like being able to publish to a relay and more advanced stuff like SUGGEST or whatever.

So for NIP 50, by all means add it to your relay implementation and advertise support via `supported_nips`. But if I have a user who wants to search relay `a` specifically for some content, but relay `a` doesn't support NIP 50, the client currently can't use the search functionality. That's sort of fine, but as a client author, I'd like to be able to search anything my user wants to search. Recommendations allow relay `a` to say "I don't support search, but I pay relay `b`, which is a relay instance optimized for search and nothing else, to index my database".

This allows for specially tuned relays to exist and build a business model through specialization, without as much of a risk of centralization.

staab:
I'll also say that my plan for Coracle when I reach the stage where I need some additional piece of functionality that doesn't exist on relays and can't be performed client side is to create a coracle-specific indexer (actually my link preview stuff already sort of is that). I don't want to do that in a way that damages the integrity of the network though, adding my service to a relay implementation and letting my client discover it organically would be much preferred.

huumn:
> Is that stuffing things into relays?

Yes, I think relays adding anything more computationally intensive than what they already do is stuffing. AFAIK they already have enough problems to solve in terms of storage and bandwidth.

We already have a problem with relay discovery, right? It seems like it'd complicate things further by having relays implement bespoke NIPs. Are they relays or everything-lays?

The internet's design works because each layer (approximately at least) does the minimal amount of work necessary for its layer to function. It lets that layer (in this case relays) optimize its functionality to death like ISPs do to the IP layer. Specialization scales well.

AFAICT relays already have enough functionality to optimize: event storage and making that storage highly available through basic queries while also protecting itself from abuse. That's A LOT to do well even at the network's current size.

Counting and search and complicated queries are stuffing IMO. Yet, they are absolutely necessary for meeting user expectations. Some relays could provide multiple layers at one address if they wanted, but architecturally, it makes more sense to add this kind of additional functionality in a layer - a layer relays do not concern themselves with - rather than make a single heterogeneous relay layer.

staab:
@pablof7z you talked about this problem a bunch in the relays panel yesterday, I'd be interested in your thoughts on this as a way to reduce centralization, encourage data locality, and ask clients to be smarter.

huumn:
As I said in #522 I think, naive as I am, that all relay discovery efforts are probably worth exploring.

As such I don't really have much to add. I think this design pattern is consistent with the ways clients discover relay functionality.

Another pattern that might be worth considering is one that's a little less stateful and therefore flexible like redirection. ie when client tries to use a nip that isn't supported, the relay responds with a NOTICE that tells them to try it on some other relay. The main advantage this would have over the proposal is it could be dynamic (the relay sees some sub-relay is down, etc) and the client would have a little less global state to manage.

PR #260, opened 2023-02-15
First Draft of Proposed NIP-76: Trace Resistant Private Posts

d-krause:
Dear Reviewer,

I am not expecting this to get merged right away.  I am only trying to establish the concept, get some feedback, etc. 

I have a little more explaining to do with regard to this NIP, some of which could require 1 or 2 more NIPs.  If needed, I was hoping I could group all of them together sequentially (ie 76,77,78), whatever the numbers may be.

The other two NIPS would focus on expanding NIP-06, to make it more secure and support multiple users (think like other facets of a user, not actual multiple users) from one wallet, and another to describe private profile metadata that can be used to control access to the user posts at the relay level.

I have been working on https://animiq.com/ since 2019, and as soon as I discovered NOSTR, I was like, **WOW**!!! It is **exactly** the protocol I have been waiting for to scale the animiq code up.   All my source is on Keybase right now,  I will be bringing it out in the open to github soon.   Then I will work on a NOSTR client port and relay support.

Let me know any thoughts or concerns you have.

Thanks,
Dave
animiq@idk.email

### DIFF
--- a/76.md
+++ b/76.md
@@ -1,0 +1,85 @@
+NIP-76
+======
+
+Trace Resistant Private Posts
+-------------------------------
+
+`draft` `optional` `author:d-krause`
+
+Given the privacy concerns detailed in [A Cypherpunk's Manifesto](https://nakamotoinstitute.org/cypherpunk-manifesto/), additional measures can be taken in the NOSTR protocol to further maintain the privacy of NOSTR users when they desire it. This NIP defines a protocol that can be implemented by conforming relays and clients to maximize the privacy of a user's posts.
+
+## Goal
+The goal of NIP-76 is to provide a way for users to share private data while also making it very difficult for governments or other third parties to scan content to identify posters and subscribers of the content.  The third party "needing" this information must start with identifying the targeted entity **first**, *and only then* can they begin to work on obtaining access to their NOSTR private contents.
+
+## Two BIP-32 Extended Private Keys Required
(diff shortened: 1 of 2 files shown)


mikedilger:
I presume the next nips (77 and 78) will detail at least how to exchange these new keys?

d-krause:
> I presume the next nips (77 and 78) will detail at least how to exchange these new keys?
Yes indeed.  And a few other details and clean up work.  Just need time to focus on it.

jeffthibault:
If "Pay-to-Relay" (whitelistling pubkeys that have paid) becomes the preeminent way for relays to handle spam, then it will become costly for the sender seeking privacy to use this as they will have to pay the relay cost every time they post.

d-krause:
> If "Pay-to-Relay" (whitelistling pubkeys that have paid) becomes the preeminent way that relays handle spam, then it will become costly for the sender seeking privacy to use this as they will have to pay the relay cost every time they post.

I agree.  But also perhaps freedom isn't free?  Then again, I have not looked at the "Pay-To-Relay", is there a NIP for it, or a pull request to a NIP?  (I did look before asking).

If I could see the spec on that it would help. One could for example pay to whitelist hundreds of pubkeys, a bulk rate perhaps?  Again, though I need to see the spec before I could hypothesize.

jeffthibault:
There is not a NIP. I was just mentioning this as a possible concern.

d-krause:
@mikedilger @jeffthibault - I created a new PR with more information here:
 [NIP-76: Private Channels - Draft 2 - Includes Client Implementation and Tool kit library](https://github.com/nostr-protocol/nips/pull/413)

Issue #236, opened 2023-02-09
Generalise event lifetime preferences


hoytech:
## Background

Implementing NIP-33 and NIP-40 for my relay (strfry) caused me to think a bit about user-specified event lifetime preferences. We can roughly classify these into the following categories:

* Replaceable events, including parameterised replaceable (NIP-16, NIP-33): These indicate to a relay that an event can be removed when another more recent qualifying event is received
* Ephemeral events (NIP-16): These events should be broadcast to all connected users with a matching filter, but then deleted (or not stored in the first place)
* Expiring events (NIP-40): These events should be deleted when a specifed timestamp is reached

Note that if a relay supports NIP-09 deletions then all of the above could in principle be implemented with delete events:

* Replaceable events could be deleted immediately before posting a replacing event
* Ephemeral events could be deleted immediately after posting
* Expiring events could be deleted at their expiration time

By using deletions, any of these categories of event lifetimes could be implemented for events of any kind.

## Problems

Replaceable, parameterised replaceable, and ephemeral events are specified to only apply to events within a ranges of kinds. To me this seems sub-optimal:

* Protocols can't group their kinds together if they need different event lifetimes for different kinds. For instance, NIP-28 (public chat) reserves the range of kinds 40-49 but specified kind 41 to be replaceable, before NIP-16 was specified, I assume. With NIP-16, a non-contiguous kind in the 10000-19999 range would need to be reserved.
* Kinds can't compose with event lifetimes. A new protocol might like to have some messages of the same kind have different event lifetimes. For example, a protocol might like to support editable (replaceable) postings as well as "permanent" postings using the same kind.
* Users can't specify custom lifetimes for events of existing protocols. For example, you could imagine sending an ephemeral kind 1 message, as a "disappearing" temporary message that would pretty much just work on any existing client.
* The 10k blocks of ranges are arbitrarily sized, and they could be either too big (new ranges would have to be reserved in the future) or too small (prime kind real-estate will forever go unused).

## Importance of parameterised replaceable events

I believe NIP-33 is a very important specification and will be heavily used by future protocols.

The NIP itself doesn't contain any use-cases, but as a basic example, I could imagine using these types of event lifetimes for sharding and segregating your kind-3 contact lists. You could imagine having multiple different contact lists, like `family`, `friends`, `co-workers`, `randos1`, and `randos2`. This would future-proof contact lists growing too large, improve the efficiency of updates, and provide a natural organisation schema for users.

[...]

PR #264, opened 2023-02-16
Make kind 7 act as parameterized repleaceable event

arthurfranca:
Considering NIP-25 is marked as a draft, thought of a way of fixing some problems with the reactions while not breaking current clients. Although the possible values of the `d` tag may look unusual, it is just specific to kind 7.

This should make it possible to:
- react to all kinds of events (but ephemeral) or to a pubkey (profile)
- react once to save relay disk space
- allow edits to content
- re-use relays existing parameterized repleaceable event handling logic to do edits
- request just the reactions to an event (using `{ "kinds": [7], "#d": ["note id"] }` filter),  instead of including all thread reactions

I think it would be better to also suggest to include just 2 tags, the `d` tag (required) and only one `p` so that the owner of the reacted to post/profile can be notified, instead of everybody in the thread, while also saving space. But i didn't add it to not make current clients mad.

### DIFF
--- a/25.md
+++ b/25.md
@@ -24,26 +24,26 @@ or the client MAY display this emoji reaction on the post.
 Tags
 ----
 
-The reaction event SHOULD include `e` and `p` tags from the note the user is
+The reaction event MUST include a `d` tag with the `id` of the note that is being reacted to.
+Including it enables clients to pull all the reactions associated with individual posts.
+Relays that support NIP-33 SHOULD treat a reaction as a Parameterized Replaceable Event to avoid duplicates.
+
+The reaction event SHOULD include `p` tags from the note the user is
 reacting to. This allows users to be notified of reactions to posts they were
-mentioned in. Including the `e` tags enables clients to pull all the reactions
-associated with individual posts or all the posts in a thread.
-
-The last `e` tag MUST be the `id` of the note that is being reacted to. 
-
-The last `p` tag MUST be the `pubkey` of the event being reacted to.
+mentioned in. The last `p` tag MUST be the `pubkey` of the event being reacted to.
 
 Example code
 
 ```swift
 func make_like_event(pubkey: String, privkey: String, liked: NostrEvent) -> NostrEvent {
-    var tags: [[String]] = liked.tags.filter { 
-    	tag in tag.count >= 2 && (tag[0] == "e" || tag[0] == "p") 
+    var tags: [[String]] = liked.tags.filter {
+    	tag in tag.count >= 2 && tag[0] == "p"
     }
-    tags.append(["e", liked.id])
+    tags.append(["d", liked.id])
     tags.append(["p", liked.pubkey])
     let ev = NostrEvent(content: "+", pubkey: pubkey, kind: 7, tags: tags)
     ev.calculate_id()
     ev.sign(privkey: privkey)
     return ev
 }
+```


barkyq:
Smart. Starts to get a bit heavy though if every reaction needs a `d`, `p`, `e`, `a` tag, with at least one repetition (since the value of the `d` will match the value of one of the `e`, `p` or `a`). What do you think about this?

arthurfranca:
> Smart. Starts to get a bit heavy though if every reaction needs `d`, `p`, `e` tags, with probably one repetition (since the value of the `d` will probably match the value of one of the `e`, `p`). What do you think about this?
> 
> Also the reaction needs to include all the `p`, `e`, `a` tags from the event being reacted to? Seems a bit overkill, but I guess you agree with me on this point but don't want to make current clients mad.

@barkyq I've pushed changes in this direction. I do think adding all those `p`, `e`, `a` tags are a waste of disk space (bigger event size) and also of bandwidth because one asks for direct reactions but receives "all" (the relay most likely will limit it anyway) thread reactions. It is also strange to put many `p` tags which would mess with notifications (why would someone want to be notified of a reaction to the reply to a reply to a reply to his post? =o)

barkyq:
:+1: Nice that it blocks users from reacting multiple times on the same pubkey without requiring new relay logic. Also consistent with #236

shafemtol:
I think I would want to keep the `e` tag for reactions to replaceable events, or at least some reference to the id of the event originally reacted to. Without such a reference, the event may be replaced in a way that changes the context and meaning of the reaction, without any indication that the reaction was to an earlier event.

arthurfranca:
@shafemtol it totally makes sense. However, if the created_at of the reaction is older than the created_at of the current version of a replaceable event, it is enough to know you liked a regular note before it changed into a pr0n ad. Per [NIP-16](https://github.com/nostr-protocol/nips/blob/master/16.md) an event can only be replaced if it has newer timestamp. Is it enough or would you still require the original event id for your use case?

shafemtol:
I thought of the `created_at` timestamps, but in my mind relying on those would not be enough. The changed note could have a timestamp one second later than the original note and thus most likely have an earlier timestamp than any reactions.

barkyq:
yes in order for the timestamp thing to work, relays need to not allow posting of events with old `created_at`

shafemtol:
> yes in order for the timestamp thing to work, relays need to not allow posting of events with old `created_at`

That would just not be viable. That would mean that if you went out of sync with one of your relays for too long and then came back, some of your events on that relay would remain permanently unavailable or stale. At least until NIP-03 is widely deployed and actively used, relays have to take old timestamps at face value.

arthurfranca:
> The changed note could have a timestamp one second later than the original note

There is [NIP-22](https://github.com/nostr-protocol/nips/blob/master/22.md) (Event created_at Limits) that could prevent it. But as it is optional, you are right, or else we would need to add it as a dependency to this NIP, which doesn't seem right. Also, the extra `e` tag allows a client to just fetch reactions to the current version of the replaceable event if it wished to hide stale reactions, instead of filtering client side. This would encourage just typo edits near the publish time instead of full edits.

fiatjaf:
Isn't this a relay policy choice?

arthurfranca:
> Isn't this a relay policy choice?

Yes NIP-22 range is custom relay config indeed. One more reason to need the event id of the replaceable event.

fiatjaf:
Yes. Change the NIP to say that relays must deduplicate reactions to the same event.
Do not change anything else in the event format. Instead relays can be smart enough to do

```
if event.kind == 7:
  if current = get(kind = 7, #e = event.tags.e):
    replace(current, event)
  else:
    add(event)
```

arthurfranca:
@fiatjaf What about [this](https://github.com/nostr-protocol/nips/blob/a741f5ad7bb852260c92b72ef309383e82801640/25.md#deduplication)?
It is very similar to contact lists that can leverage relays replaceable events logic by just doing:
```
function isReplaceableEvent (event) {
  return event.kind === eventKinds.CONTACT_LIST ||
    (event.kind >= 10000 && event.kind < 20000)
}
```

Relays would need to just update the `isParameterizedReplaceableEvent` function to also check for `event.kind === eventKinds.REACTION`. So no special logic just for a kind as in your example.

Relays that have implemented parameterized replaceable events will automatically treat today's reaction as having d = '' as per NIP-33 specs. The relays which haven't implemented it will work just like today, storing duplicates. Clients rely on external relays, so relays are kings. In the end of the day, relays dictate nostr protocol, not the clients. Even Damus would have to change eventually.

fiatjaf:
I just don't see why add this extra `d` tag when we already have the `e` tag that can be used for replacement. And then require all clients to change when it's much easier for relays to change their behavior. Relays should probably do this for other kinds too.

arthurfranca:
> I just don't see why add this extra d tag when we already have the e tag that can be used for replacement

The previous version was a Frankestein because I tried to keep old text intact by not removing this: `Including the e tags enables clients to pull all the reactions
associated with individual posts or all the posts in a thread.` Very unlikely to need all reactions to all posts in a thread. It would be massive amount of data. Please see [updated version](https://github.com/nostr-protocol/nips/blob/15fdf8b898e5f6ee7b9c079976c35c525201c00c/25.md#tags).

At first, there wasn't an official way to categorize an event as replaceable. Replaceable Events came as NIP-16 and Parameterized Replaceable Events followed at NIP-33.

"Set Metadata", "Channel Metadata" and "Contact List" are in essence Replaceable Events, but as they were created before NIP-16, they just aren't in NIP-16 kind range.
But the logic to update them is the same as any replaceable event: `find by kind and pubkey > check if new created_at is greater > replace it`

My point is that reactions are in essence a parameterized replaceable events, though it is missing the `d` tag, cause it was created before NIP-33 even existed.
With the `d` tag we can do: `find by kind, pubkey and d tag > check if new created_at is greater > replace it`.

Just like future event kinds that wish to be replaceable shouldn't expect special logic from relays. Just use the right kind range and, if needed, the `d` tag.

fiatjaf:
My only problem with all your suggestions is that even though they are good you completely disregard the problematic effects to breaking things. Open https://nostr.net/ and see how many implementations of relays and clients exist today. It is really not worth making breaking changes to spec when all these codebases exist and are doing things in one way. It will just create massive confusion.

arthurfranca:
I did pondered over the breaking change aspect.
In this specific case of reactions, that aren't main functionality kind of events, on a NIP marked as draft, by applying changes that wouldn't break feed nor thread loading and would eventually be noticed by client authors when they see 
unexpected reaction counts, I thought it was a good idea.
Yet I get you and others don't think the same way. Thanks for replying.

fiatjaf:
Sorry, I feel bad rejecting all your suggestions given that you really seem to care.

staab:
I think this ship has sailed, adding a MUST here is a little too strong for an event type of which we have millions. I think @fiatjaf is right that this is a relay policy issue, and should probably be added to NIP 07 as an additional requirement for relays. Closing for that reason and for inactivity.

Issue #154, opened 2023-01-07
Login with Nostr


lukechilds:
I think Nostr makes a good candidate to be used as a very simple DID layer. Having "Login with Nostr" auth on websites solves a lot of problems in a very elegant way, and Nostr's main use case as a social network protocol makes it highly suited to be used as your main identity proving key.

Compare "Login with Nostr" to "Login with Lightning" to see some easy and obvious advantages:

### Remote signer vs local signer

Login with Lightning requires access to remote keys, login with Nostr requires access to local keys ideally stored in a browser extension. Due to the way Lightning works you can only really have one instance. You need all your client devices linked to a single Lightning node, this means most clients will be connecting to the signer remotely. Now if your Lightning node goes down or you lose your connection you also can't auth with any service. This could cause circular dependencies where you lose the connection to your Lightning node so you can't auth with the services you need access to to debug the issue with the connection like your hosting provider or VPN account. You could technically solve this by replicating your LN keys to other client devices only to be used for local auth signing but that introduces other risks.

### Unique identifier vs identity

A Lightning node is not really an identity but a unique identifier. It just tells you the person that auths is the same random person that authed last time, it doesn't tell you who they are. A nostr pubkey is an identity. It tells you who they are, what their name is, what they look like, who they know, how you can pay them, how you can message them.

This is much more useful as an identity layer for an application. The application can show their profile picture, username, send secure cross platform push notifications via NIP-04 encrypted Nostr DMs, etc.

### Consistent identity across services

Lightning pubkeys are sensitive private information and can leak confidential financial information, Nostr pubkeys are safe to share with anyone. LNURL-auth adds extra steps to solve this by creating derived subkeys for identities that are unique to each service you auth with. This does not seem ideal, it seems the default case is that an identity is something that you do want to follow you across all your accounts. Nostr based auth behaves more appropriate in this regard. In the rare case you need to achieve privacy and separation between certain services you can still do that by using use a throwaway Nostr key for those services.

### User relationships across services

Since authing with Nostr shares a real social identity with the service, they can also see your Nostr social graph. This could be useful for connecting you to people you already know on the new service.

## Implementation ideas

This could be done today without any spec changes by just using existing NIP-07 functionality:

```js
const pubkey = async window.nostr.getPublicKey()
const message = {service: 'auth.example.com', challenge}
const signature = async window.nostr.signEvent(message)

if (!isValidSignature(pubkey, signature)) throw new AuthenticationError()

[...]

PR #268, opened 2023-02-18
NIP-XX: Deterministic Private Key Generation from Ethereum Wallet Signature

0xc0de4c0ffee:
> Note : this is a draft pull request for feedback.

[NIP-XX: Deterministic Private Key Generation from Ethereum Wallet Signature](https://github.com/dostr-eth/nips/blob/ethkeygen/xx.md)

# Example code : 
```js
const secp256k1 = require('@noble/secp256k1');
const {hexToBytes, bytesToHex} = require('@noble/hashes/utils');
const {hkdf} = require('@noble/hashes/hkdf');
const {sha256} = require('@noble/hashes/sha256');

// const wallet = // connected ethereum wallet with ethers.js
let username = "me@domain.eth.limo"
let chainId = wallet.getChainId(); // get chainid from connected wallet
let address = wallet.getAddress(); // get address from wallet
let info = `eip155:${chainId}:${username}:${address}`;

let message = `Login to Nostr as ${username}\n\nImportant: Please verify the integrity and authenticity of your Nostr client before signing this message.\n${info}`
let signature = wallet.signMessage(message); // request signature from wallet

let password = "horse staple battery"
let inputKey = await sha256(hexToBytes(signature.slice(2))); //skip "0x"
let salt = await sha256(`eip155:${chainId}:${username}:${password?password:""}:${signature.slice(68)}`);
let dkLen = 42; 

let hashKey = await hkdf(sha256, inputKey, salt, info, dkLen);
let privKey = secp256k1.utils.hashToPrivateKey(hashKey);
let pubKey = secp256k1.schnorr.getPublicKey(privKey);
```

### DIFF
--- a/xx.md
+++ b/xx.md
@@ -1,0 +1,61 @@
+NIP-XX
+======
+
+Nostr specific keys from Ethereum wallet signatures
+----------------------------------------------------
+
+`draft` `optional` `author:0xc0de4c0ffee` `author:sshmatrix`
+
+This NIP ?is optional specification for Nostr clients to generate private key from deterministic Ethereum wallet signatures.
+
+
+A) Username : 
+```
+let username = "mypetname" || "mynip05@domain.eth.limo" || "domain.eth.limo" || "sub.domain.eth.limo"
+```
+
+   1) `mypetname` is NIP02 compatible name
+   2) `mynip05@domain.eth.limo` is NIP05 compatible ID
+   3) `domain.eth.limo` is NIP05 equivalent to `_@domain.eth.limo`
+   4) `sub.domain.eth.limo` is NIP05 equivalent to `_@sub.domain.eth.limo`
+
+~~Implementing clients should verify if generated public keys match NIP05 and NIP02 records.~~ 
+
+B) Message : 
+```
+let message = `Login to Nostr as ${username}\n\nWARNING : DO NOT SIGN THIS REQUEST FROM UNTRUSTED NOSTR CLIENTS.\n${checksum(wallet.address)}`
+```
+
+C) Private Key : 
+
+```js
+import { hkdf } from '@noble/hashes/hkdf';
+import { sha256 } from '@noble/hashes/sha256';
+import * as secp256k1 from '@noble/secp256k1';
+
+let username = 'name@domain.eth.limo';
+let optPassword = "horse staple battery"; 
+// optional pin/password
+
+let address = wallet.getAddress(); 
+// get checksum'd address from eth wallet
+let signature = wallet.signMessage(message); 
+// request signature from eth wallet
+// 0x + bytes1 + bytes32 + bytes32 
+//  2 + (2     +   64    +     64) = 132 length  
+
+let inputKey = sha256(signature);
+let salt = sha256(`nostr:${username}:${optPassword}:${signature.slice(68)}`); //68~132
+// nostr:${username}:${signature.slice(68)}
+
+let info = `nostr:${username}:${address}`;
+
+let dkLen = 42; // 32 + 8 + 2
+//FIPS 186 B.4.1 limit must be >=keylen+8
+
+let hashKey = hkdf(sha256, inputKey, salt, info, dkLen);
+let privKey = nobleSecp256k1.hashToPrivateKey(hashKey); 
+//FIPS 186 B.4.1 @noble/secp256k1
+
+let pubKey = nobleSecp256k1.getPublicKey(privKey)
+```
--- a/xx.md
+++ b/xx.md
@@ -1,87 +1,118 @@
 NIP-XX
-======
+--
 
-Nostr specific keys from Ethereum wallet signatures
-----------------------------------------------------
-
+Nostr-Specific Deterministic Private Key Generation from Ethereum Wallet Signature
+--
 `draft` `optional` `author:0xc0de4c0ffee` `author:sshmatrix`
 
-This NIP ?is optional specification for Nostr clients to generate private key from deterministic Ethereum wallet signatures.
+## Abstract
+
+This specification provides an optional method for Nostr clients to generate deterministic private keys from Ethereum wallet signatures.
+
+## Terminology
+### a) `username`
+Username can be either of the following:
+
+`username` or `user@domain.eth.limo` or `domain.eth.limo` or `sub.domain.eth.limo`, where,
+
+- `username` is a [NIP-02](https://github.com/nostr-protocol/nips/blob/master/05.md) compatible name,
+- `name@domain.eth.limo` is a [NIP-05](https://github.com/nostr-protocol/nips/blob/master/05.md) compatible name,
+- `domain.eth.limo` is [NIP-05](https://github.com/nostr-protocol/nips/blob/master/05.md) equivalent of
+`_@domain.eth.limo`,
+- `sub.domain.eth.limo` is [NIP-05](https://github.com/nostr-protocol/nips/blob/master/05.md) equivalent of
+`_@sub.domain.eth.limo`.
+> Note : `sub@domain.eth.limo` and `sub.domain.eth.limo` are NOT same ID as their signatures will be different.  
 
 
-## A) Username : 
-```
-let username = "mypetname" || "mynip05@domain.eth.limo" || "domain.eth.limo" || "sub.domain.eth.limo"
-```
-
-   1) `mypetname` is NIP02 compatible name
-   2) `mynip05@domain.eth.limo` is NIP05 compatible ID
-   3) `domain.eth.limo` is NIP05 equivalent to `_@domain.eth.limo`
-   4) `sub.domain.eth.limo` is NIP05 equivalent to `_@sub.domain.eth.limo`
-
-~~Implementing clients should verify if generated public keys match NIP05 and NIP02 records.~~ 
-
-## B) Password : 
-Password is optional value used in HKDF salt:
+### b) `password`
+Password is optional string value used in HKDF salt,
 ```js
-let username = "name@domain.eth.limo"
 let password = "horse staple battery"
-let signature = wallet.signMessage(message)
-let salt = sha256(`nostr:${username}:${password}:${signature.slice(68)}`);
-if(!password || password == ""){
-   salt = sha256(`nostr:${username}:${signature.slice(68)}`)
-}
+let salt = await sha256(`eip155:${chainId}:${username}:${password?password:""}:${signature.slice(68)}`);
 ```
 
-
-## C) Message : 
-
+### c) `message`
 ```js
-let message = `Login to Nostr as ${username}\n\nWARNING : DO NOT SIGN THIS REQUEST FROM UNTRUSTED NOSTR CLIENTS.\n${checksum(wallet.address)}`
+let message = `Login to Nostr as ${username}\n\nImportant: Please verify the integrity and authenticity of your Nostr client before signing this message.\n${info}`
 ```
-## D) Signature :
-Signature is 
-
-## C) HKDF : 
-We use HKDF with SHA256
+### d) `signature`
+Deterministic signature from connected wallet. Signatures are 65 bytes long, `bytes1(v)+bytes32(r)+bytes32(s)`.
 ```js
-import { hkdf } from '@noble/hashes/hkdf';
-import { sha256 } from '@noble/hashes/sha256';
-import * as secp256k1 from '@noble/secp256k1';
+let signature = wallet.signMessage(message); 
+```
+### e) `HKDF`
+HKDF-SHA-256 is used to derive 42 bytes long hash key.
+`hkdf(sha256, inputKey, salt, info, dkLen = 42)` 
+- `Input key` is SHA-256 hash of signature bytes.
+   ```js
+   let inputKey = await sha256(hexToBytes(signature.slice(2)));
+   ```
+- `Salt` is SHA-256 hash of following identifier string. `signature.slice(68)` is hex `s` value of signature, last 32 bytes.
+   ```js
+   let salt = await sha256(`eip155:${chainId}:${username}:${password?password:""}:${signature.slice(68)}`);
+   ```
+- `Info` is string with following format.
+   ```js
+   let info = `eip155:${chainId}:${username}:${address}`;
+   ```
+- `Derived Key Length` is set to 42. FIPS 186/4 B.4.1 require hash length to be >=n+8 where, n is length of final private key. (42 >= 32 + 8)
+   ```js
+   let dkLen = 42; 
+   ```
+- `hashToPrivateKey` function is FIPS 186-4 B.4.1 implementation to convert derived hash keys from `HKDF`to valid `secp256k1` private keys. This function is implemented in js `@noble/secp256k1` as `hashToPrivateKey`.
+   ```js
+   let hashKey = hkdf(sha256, inputKey, salt, info, dkLen=42);
+   let privKey = secp256k1.utils.hashToPrivateKey(hashKey);
+   let pubKey = secp256k1.schnorr.getPublicKey(privKey);
+   ```
 
-//let username = 'name@domain.eth.limo';
-//let optPassword = "horse staple battery"; 
-// optional pin/password
+## Implementation Requirements
 
-//let address = wallet.getAddress(); 
-// get checksum'd address from eth wallet
-//let signature = wallet.signMessage(message); 
-// request signature from eth wallet (v,r,s)
-// 0x + bytes1(v) + bytes32(r) + bytes32(s)
-//  2 + (2 + 64 + 64) = 132 String length  
+- Connected Ethereum Wallet signer MUST be EIP191 and RFC6979 compatible.
+- The message MUST be string formatted as `Login to Nostr as ${username}\n\nImportant: Please verify the integrity and authenticity of your Nostr client before signing this message.\n${info}`.
+- HKDF input key MUST be generated as the SHA-256 hash of 65 bytes signature.
+- HKDF salt MUST be generated as SHA-256 hash of string `eip155:${chainID}:${username}:${password?password:""}:${signature.slice(68)}`.
+- HKDF derived key length MUST be 42.
+- HKDF info MUST be string formatted as `eip155:${chainId}:${username}:${address}`
 
-let inputKey = sha256(signature);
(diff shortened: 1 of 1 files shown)


barkyq:
Seems a bit risky that the web app would have the private key, unless this got implemented in, e.g., metamask. But if we are going down that road, perhaps metamask could just implement NIP-07, and use this method under the hood.

In any case, do we really want to open the door to putting various key generation methods into their own NIPs? What about a NIP to turn a BTC private key into a nostr private key, and a NIP to turn a Monero private key into a nostr private key, and a NIP to turn a NANO private key into a nostr private key, etc...

On a further thought, each app can just implement their own method of: `<insert private key format>` to <nostr private key format>`

barkyq:
See also #258 for the inverse problem.

fiatjaf:
I agree the only way for this to work would be if this was done on Metamask directly, and Metamask implemented NIP-07.

0xc0de4c0ffee:
Thanks for feedback @barkyq @fiatjaf 🙏 

> Seems a bit risky that the web app would have the private key, unless this got implemented in, e.g., metamask. But if we are going down that road, perhaps metamask could just implement NIP-07, and use this method under the hood.

It's less $ risker compared to reusing same ETH keys in Nostr clients. This is optional specs for web/mobile Nostr clients and `window.nostr`/`window.ethereum` providers to generate multiple ID/keys with deterministic signature, username and optional password.

> In any case, do we really want to open the door to putting various key generation methods into their own NIPs? What about a NIP to turn a BTC private key into a nostr private key.....

We can update this as [Chain Agnostic Improvement Proposals (CAIP)](https://chainagnostic.org/) approach to support  *ALL* coins/chains compatible with RFC6979 like deterministic signature.

Example using  BIP44 (SLIP44) coin type and EIP155 chain ID. 
```js
let info1 = `bip44:${coin_type}:${username}:${address}`;
let info2 = `eip155:${chain_id}:${username}:${address}`;
let info3 = `cosmos:${hub_id}:${username}:${address}`; // not sure about this, have to lookup
//... 
let salt1 = await sha256(`bip44:${coin_type}:${username}:${password?password:""}:${last_32bytes_ofSignatureHex}`);
let salt2 = await sha256(`eip155:${chain_id}:${username}:${password?password:""}:${last_32bytes_ofSignatureHex}`);
//...
```

0xc0de4c0ffee:
`Login with Nostr` #154 is our inverse-brah 😆 

We're working on different BIP47 "like" stealth address generation too, more details at https://github.com/nostr-protocol/nips/pull/258#issuecomment-1437199109


>  I agree the only way for this to work would be if this was done on Metamask directly, and Metamask implemented NIP-07.
 
There's "Metamask Flask" plugin for Nostr, [Schnorr Snap](https://github.com/neeboo/schnorr_snap). MM-Flask is experimental wallet not ready for average users, it will take more time for major coin-wallets to implement Nostr+schnorr feature. Can't wait for major wallet provider to implement that, & hope they don't reuse their coin-keys for Nostr again. 

This specification is for `Nostr + X` compatible app/clients to sign Nostr stuffs with `window.nostr` and X-chain stuffs with `window.ethereum` like wallet providers. [CAIP-122: Sign in with X](https://github.com/ChainAgnostic/CAIPs/pull/122)  feature is used to generate app-specific deterministic keys for Nostr app/clients and compatible NIP07 providers.

Updating chain and account identifiers to as defined in [CAIP-02:Blockchain ID Specification](https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-2.md) and [CAIP-10:Account ID Specification](https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-10.md). 

> Note : We were exploring SLIP/BIP44 `coin_type` as chain identifier but CAIPs are using that for assets type inside BIP122, so we're using that BIP122:.. leaving BIP44/:coin_type as assets type for future.  🙏 

Example: CAIP-10, CAIP-02 with NIP02/NIP05 user identifiers.
```
# Bitcoin mainnet (BIP122)
bip122:<first 16 bytes of genesis/fork hash>:address:username
bip122:000000000019d6689c085ae165831e93:address:username

# EVM chains  (EIP155)
eip155:<chain_id>:address:username
eip155:1:address:username

# Cosmos Hub (Tendermint + Cosmos SDK/Binance)
cosmos:<hub/ChainName>:address:username
cosmos:cosmoshub-2:address:username
cosmos:Binance-Chain-Tigris:address:username
cosmos:iov-mainnet:address:username

...

# Dummy max length (8+1+32 = 41 chars/bytes) +":address:username"
chainstd:8c3444cf8970a9e41a706fab93e7a6c4
```

Still WIP, Example
```js
let nostr = widow.nostr || nostrTools ;
let coinWallet = await connectEthers(window.ethereum);

let username = 'me@example.com';
let password = 'opt#password';
let pubkey = await nostr.nipxx.newIdentity(username, password="", chainType="eip155", coinWallet="");
// wrapped as nipxx.loginWithX(...) function. 

// registration: set new nip02 and nip05 records for this pubkey.
// login: verify if there's matching nip02 and nip05 records.
```

sshmatrix:
> Is this method of generating keys a standard among altcoins?

No, I wouldn't say so. But it is standard and well-known in general Cryptography. I think @0xc0de4c0ffee might be more aware of previous implementations of it, if any, in cryptosphere.

0xc0de4c0ffee:
> Is this method of generating keys a standard among altcoins?

It's fully chain agnostic `sign-in-win-x` CAIP122 used to generate deterministic keys.

There are few Ethereum dapps using straight `sha256(half_signature)` to derive app-specific private keys from signatures. We're adding chain and address identifiers (CAIP02+CAIP10) format with NIP05/NIP02 user info (+optional password) and with HKDF to generate deterministic keys.

Example from umbra.js stealth payment codes. 
> https://github.com/ScopeLift/umbra-protocol/blob/master/umbra-js/src/classes/Umbra.ts#L549
Screenshot 2023-03-21 at 21 50 22

We're signing stuffs from both `window.nostr` & `window.ethereum`.. for basic e2ee msgs and stealth payments, so we don't want to reuse or mix ETH keys with Nostr keys.

sshmatrix:
Hi @fiatjaf,

Hope you are doing good. Dostr client is nearing its official launch (Mar 31). In reality, the client is already online at:

- on web3 at [dostr.eth.limo](https://dostr.eth.limo), and
- on web2 at [dostr.xyz](https://dostr.xyz),

for earlybirds. There won't be any drastic changes to the client from now on other than cosmetic fine-tuning of the UI/UX.

With respect to the NIP:

- We have changed the numbering of the NIP in PR to NIP-111 as its our "favourite" number. 
- More importantly, we have finalised the format of the message to eth-Sign to generate Nostr keys. 

This NIP is in its final boss form from our side 😊

digi-monkey:
are you still refactoring the signing message? I like this idea and I am implementing at https://github.com/digi-monkey/flycat-web/pull/82, I will hope the message can be finalized and not changed again.

111.md
```diff
@@ -0,0 +1,242 @@
...
+   let hashKey = hkdf(sha256, inputKey, salt, info, dkLen = 42);
+   let privKey = secp256k1.utils.hashToPrivateKey(hashKey);
+   let pubKey = secp256k1.schnorr.getPublicKey(privKey);
+   ```
+
+## Architecture
+
+The resulting architecture of NIP-111 can be visually interpreted as follows:
+
+![](https://raw.githubusercontent.com/dostr-eth/resources/main/graphics/nip-xx.png)
```
bumblefudge, line 111:
```suggestion
![](https://raw.githubusercontent.com/dostr-eth/resources/main/graphics/nip-111.png)
```

111.md
```diff
@@ -0,0 +1,242 @@
...
+
+## Implementation Requirements
+
+- Connected Ethereum wallet Signer **MUST** be EIP-191 and RFC-6979 compatible.
+- The `message` **MUST** be string formatted as
+```
+`Log into Nostr client as '${username}'\n\nIMPORTANT: Please verify the integrity and authenticity of connected Nostr client before signing this message\n\nSIGNED BY: ${caip10}`
```
bumblefudge, line 118: I'm a little confused, is this "inspired by" Sign-In With Ethereum/CAIP-122 or is it meant to be conformant to one or both of those two specs? If the latter was intended, the message should probably conform the CAIP-122/EIP-4361 ABNF, which could include the entire `message` (with `\n`s removed) in the `statement` field, use the `salt` value for `nonce`, and `info` as first entry in the `resources` array (not really sure what to do with `inputKey`, as I'm unclear on its exact function on a cursory read-through).  If it's not too late to go in that direction, I think it might have some benefits, such as being displayed to metamask users in the familiar, locked-down "Sign-In With Ethereum" modal rather than as a generic "personal_sign" modal... which only displays when presented with a personal_sign `message` matching the ABNF :D 
image

sshmatrix:
> @sshmatrix are you still refactoring the signing message? I like this idea and I am implementing at [digi-monkey/flycat-web#82](https://github.com/digi-monkey/flycat-web/pull/82), I will hope the message can be finalized and not changed again since I am going to merge the PR and rollout this feature

Hi @digi-monkey, the message contained in the present draft is final. It won't be changed. We have adopted the same on [Dostr](https://dostr.eth.limo) client and frozen it. There was a typo in the draft however and we have fixed it. It didn't affect the implementation apart from a mislabeling of `caip10` variable in the `statement`. 

@bumblefudge Implementing CAIP-122 as is won't work in this case, since we need the final key output to be deterministic. Adding the `nonce` and `timestamp` will add variable `salt` to the key derivation function resulting in variable keys. Deterministic key generation requires deterministic message and this led to the current choice of `message` to sign.

111.md
sshmatrix: Thanks @bumblefudge! We intended to conform to CAIP-122 as much as possible but we had to strip all the variable quantities from the `message` format to retain deterministic nature of key derivation. In some sense, this implementation is not really "Sign-In" but more of "Ephemeral KeyGen and then Sign-In", and hence the necessary deviation from the standard.

bumblefudge: Well, the problem with an ABNF-based syntax is that you can't deviate and still get interop with other SIWX libraries or take advantage of SIWX support built into Metamask! Can I suggest hard-coding into the spec and template rather than removing the values you don't need, so that the message can still conform to the ABNF and get displayed to the users as a SIWX message?  I would note that ephemeral keygen is already baked into the SIWX standard, and is being used for that exact usecase by most implementers (the generated ephemeral public key is usually included as a value in the `Resources` array, although this isn't really mentioned explicitly in the specification itself!)

bumblefudge:
`timestamp` can be zero (i.e. midnight 1/1/1970) and nonce can be any string.  in fact, CASA member @skgbafa demo'd a usage of SIWE for deterministic key generation at a CASA event, so I see no reason a more conformant version couldn't be used for that use-case.  

apologies that I missed this notification and wasn't able to make these suggestions before final status! if SIWX conformance is of use, maybe it's worth a superceding spec or an extension spec that puts the same inputs into a conformant `message` value, so that individual implementations can choose to implement that variant instead if taking advantage of wallet special-handling of SIWX messages is worth the extra work to them?

bumblefudge:
Oh!  Actually I meant an extension NIP, but hey, maybe an extension-CAIP* might make just as much sense.  I would also add that if the NIPs elaborate enough complexity and cohesion that there are nostr-enabled wallets, signing methods, etc, it might some day make sense to create a Nostr [namespace](https://namespaces.chainagnostic.org/), where you can define a profile for CAIP-10 (syntax of a URN scheme for CAIP-10 encodings of nostr accounts/pubkeys), CAIP-122 (if a Nostr app which held a private key wanted to "sign in with nostr" to another service), etc etc.

* = note, it wouldn't be CAIP-111, unfortunately, because CAIPs are numbered to the PR# that creates them, so if you opened a PR today it would be CAIP-222 --- which is a funny coincidence, hurry up before someone else nabs that number!

111.md
sshmatrix: Fixed!

111.md
0xc0de4c0ffee: > I'm a little confused, is this "inspired by" Sign-In With Ethereum/CAIP-122 or is it meant to be conformant to one or both of those two specs?  If the latter was intended, the message should probably conform the CAIP-122/EIP-4361 ABNF, which could include the entire message (with `\n` removed) in the statement field, use the salt value for nonce, and info as first entry in the resources array (not really sure what to do with inputKey, as I'm unclear on its exact function on a cursory read-through).
 
@bumblefudge 🙏  
Should we remove CAIP122 ref as we can't pass full strict ABNF? or req all as new CAIP?

We actually started as basic signature request before upgrading to full CAIP122/ERC4361 format but had to remove all extradata for deterministic keys as they are too strict to fit in all.. So it's now back to generic "personal_sign" modal for chain agnostic "sign-in-with-x" in Nostr context, internally it's using deterministic signature from wallet to generate "app specific deterministic keys" across all Nostr clients.

We could fill in all ABNF *required* formats with deterministic/fixed values but we can't pass `URI` validation in wallets, and "`${service} wants you to sign in with <chain> account: \n<addr>\n...must:have`" text message for ALL web2+3 D/Apps is too strict for our deterministic keygen, web3+dapps are not supposed to have single URLs/apps as entry point.. 

"Important: .....\n\n" is supposed to be "Warning! ..." statement, it's not key: value. All other extradata is wrapped in `SIGNED BY : ${CAIP10}`, & our msg is simple 3 blocks `<title>\n\n<statement>\n\n<key:value>...` 


> I would note that ephemeral keygen is already baked into the SIWX standard, and is being used for that exact usecase by most implementers (the generated ephemeral public key is usually included as a value in the Resources array, although this isn't really mentioned explicitly in the specification itself!) 

Can you add some ref codes/links for that?

sshmatrix:
> Oh! Actually I meant an extension NIP, but hey, maybe an extension-CAIP* might make just as much sense. I would also add that if the NIPs elaborate enough complexity and cohesion that there are nostr-enabled wallets, signing methods, etc, it might some day make sense to create a Nostr [namespace](https://namespaces.chainagnostic.org/), where you can define a profile for CAIP-10 (syntax of a URN scheme for CAIP-10 encodings of nostr accounts/pubkeys), CAIP-122 (if a Nostr app which held a private key wanted to "sign in with nostr" to another service), etc etc.
> 
> `*` = note, it wouldn't be CAIP-111, unfortunately, because CAIPs are numbered to the PR# that creates them, so if you opened a PR today it would be CAIP-222 --- which is a funny coincidence, hurry up before someone else nabs that number!

Hi @bumblefudge, linking the answer to your questions here: https://github.com/nostr-protocol/nips/pull/268#discussion_r1161227145. GitHub notifications are unreliable on multiple threads

111.md
sshmatrix: > > I'm a little confused, is this "inspired by" Sign-In With Ethereum/CAIP-122 or is it meant to be conformant to one or both of those two specs?  If the latter was intended, the message should probably conform the CAIP-122/EIP-4361 ABNF, which could include the entire message (with `\n` removed) in the statement field, use the salt value for nonce, and info as first entry in the resources array (not really sure what to do with inputKey, as I'm unclear on its exact function on a cursory read-through).
> 
> @bumblefudge 🙏 Should we remove CAIP122 ref as we can't pass full strict ABNF? or req all as new CAIP?
> 
> We actually started as basic signature request before upgrading to full CAIP122/ERC4361 format but had to remove all extradata for deterministic keys as they are too strict to fit in all.. So it's now back to generic "personal_sign" modal for chain agnostic "sign-in-with-x" in Nostr context, internally it's using deterministic signature from wallet to generate "app specific deterministic keys" across all Nostr clients.
> 
> We could fill in all ABNF _required_ formats with deterministic/fixed values but we can't pass `URI` validation in wallets, and "`${service} wants you to sign in with <chain> account: \n<addr>\n...must:have`" text message for ALL web2+3 D/Apps is too strict for our deterministic keygen, web3+dapps are not supposed to have single URLs/apps as entry point..
> 
> "Important: .....\n\n" is supposed to be "Warning! ..." statement, it's not key: value. All other extradata is wrapped in `SIGNED BY : ${CAIP10}`, & our msg is simple 3 blocks `<title>\n\n<statement>\n\n<key:value>...`
> 
> > I would note that ephemeral keygen is already baked into the SIWX standard, and is being used for that exact usecase by most implementers (the generated ephemeral public key is usually included as a value in the Resources array, although this isn't really mentioned explicitly in the specification itself!)
> 
> Can you add some ref codes/links for that?

`URI` is definitely a bigger issue and it hadn't come to my mind before. While `timestamp` and `nonce` can be replaced with placeholders, `URI` is used by Metamask for its community-audited safe dApp list. This is problematic for dApps or services without a unique entry point. CAIP122 is too strict for all use-cases and pretty much a death sentence for the deterministic use-case.

bumblefudge: Couldn't the uri be:
https://github.com/nostr-protocol/nips/blob/master/111.md
? It doesn't need to be on the `domain`

sshmatrix: > Couldn't the uri be: https://github.com/nostr-protocol/nips/blob/master/111.md ? It doesn't need to be on the `domain`

That's not a bad idea honestly but it will make users highly suspicious and wary of a service with 1/1/1970 in timestamp and a GitHub link in URI. I personally won't sign and subscribe to such a service at first glance. I believe there is room for a new CAIP detailing a separate signature format standard for deterministic use-cases.

0xc0de4c0ffee: > Couldn't the uri be: https://github.com/nostr-protocol/nips/blob/master/111.md ? It doesn't need to be on the `domain`

When we connect MM at `https://app.dostr.eth.limo` it'll check and add "WARNING" if connected domain isn't matching `domain` URL. It's good feature for web2 & web3 apps with single point of entry but for our use case we're adding extra user info & pw with static signature request as basic security.

It all work as plaintext "sign in with xyz on Nostr" signature request so we're missing all SWIx interface/features. 

+ it's better if we explain why we're doing this, our keygen design is inspired by Umbra cash & old Whisper(ssh) newKeypair/ newIdentity function. We use `window.ethereum` to sign eth tx/permits then use deterministic keys in `window.nostr` to send that over Nostr relays for off-chain features like stealth payments, DEFI, NST markets, alt-mempool for AA, xyz services, bundlers/bots, games...

digi-monkey: personally I think if we don't need timestamp, adding a placeholder for it is not a good idea.

Issue #247, opened 2023-02-13
NIP-26: Delegation coordination flow, new events (general discussion)


catenocrypt:
Specifying delegation coordination between two apps is not in the scope of [NIP-26](https://github.com/nostr-protocol/nips/blob/master/26.md), but some hints could be included; and some extra specification would be helpful for improved UX experience.

The goal of delegation is that the client using the delegated identity does not possess the secret of the main identity. As a consequence, the delegation needs to be created (signed) in another client (a client with the role of identity manager). But how does the delegation token gets to the delegated client?

Some options:
- Copy & paste
- QR code
- Remote signing (NIP-46 Nostr Connect #153 )
- Nostr note

Copy-paste of delegation token is not optimal UX, as it's quite long. Copying of anything more complex than a pubkey is not optimal.

QR code has the same concerns, and in addition, not all clients have access to camera.

Signing of the delegation token through remote signing. The condition for this is that the identity manager client also acts as a remote signer. Also, remote signing protocol may need extension. Delegation parameters are set in the delegatee, so signer should be able to meaningfully display them for the user.
In general, Delegation and Remote signer are two different approaches for the same problem (not having main secrets in a client), and mixing/tying them is not a good idea.

Sharing through Nostr notes is convenient, but what kind of note to use?
- Metadata
- Custom
- New kind

Metadata could be extended with list of (valid) delegations. However, extending metadata usage increases the risk of race conditions between clients.

Flow:
- in delegated client, user creates a new identity, copy the new pubkey
- in identity manager client, user chooses Delegation, pastes the delegatee pubkey, sets parameters, sign, create delegation token
- Post new event that includes the delegation token
- Delegated client notices the event, and extract the delegation token. If it does not find it (some events may get lost), as a fallback the main identity pubkey can be entered, and it can retrieve.

Alternatively, a custom or a new event kind could be used for this purpose. The concern here is event kind bloat.

Request delegation. The UX flow initiated by the delegated client should also be considered:
- User installs new secondary client, creates new identity
- It searches for its main identity (e.g. by search on username or other metadata; fallback can be copy-paste)
- Selects 'Request delegation' action

As a result, a relevant event is posted. Special event kind or custom kind, or metadata with unsigned delegation.
User starts its primary (identity manager) client, it noitces the delegation request (complete with delegatee pubkey and proposed parameters), and can perform the delegation.

[...]

PR #269, opened 2023-02-19
NIP-26: Add hint notes on typical delegation UX flow (by copy&paste of delegation tag)

optout21:
Add some notes on how is delegation performed in a typical case, how delegation tag can be transported by simple copy-and-paste to the delegated client.  This is to offer a hint to client developers.
Related to issue #247.


staab:

PR #280, opened 2023-02-21
NIP-DID: Combining Nostr Keys and Decentralized Names

zgayjjf:
Hello, frens!
Many nostr users are from Web3, and most use Decentralized Names like .bit, ENS, Unstoppable Domain, etc. It would be great if we could support them.

Hereby I propose NIP-DID,  a protocol combining Nostr Keys and Decentralized Names.

With this proposal, we can simplify the process of sharing and following. Users will use a human-readable name instead of a long, irregular [NIP-19](https://github.com/nostr-protocol/nips/blob/master/19.md) pubkey to follow people.

**Before:**
Drop a public key somewhere else and say please follow this pubkey: `npub1p0ew5ln6hq4c20l7zftkwyrnn03php05zl8aydntugg74cu8m28sfk25c6`.

**After:**
Tell your friends to follow your DID name like `nostr.bit`. And that's all!.

### Key points
1. Set up a map between DID name and Nostr pubkey.
2. Use DID name to get rid of long, irregular npub pubkey(NIP-19)

### DIFF
--- a/NIP-DID-NAME.md
+++ b/NIP-DID-NAME.md
@@ -83,11 +83,19 @@ It can be a single purpose service resolving only one suffixes. It can also be a
 
 Here are 2 example of possible Data Providers:
 
-**SDK for JavaScript Client**
+**SDK for JavaScript Clients**
+
+The JavaScript SDK is the most convenient option for JavaScript Client to use.
+
 - [AllDID](https://github.com/dotbitHQ/AllDID)
 
 **HTTP API**
+
+HTTP API service can provide a wider range of applicability.
+
 - [AllDID-API](https://github.com/dotbitHQ/AllDID-api)
+
+> These Data Providers are **not** mandatory. Clients can choose whichever Data Provider the like, as long as it can resolve the DID names they want to support.
 
 
 ### SubDID
--- a/nip-did-name.md
+++ b/nip-did-name.md
@@ -19,7 +19,7 @@ Consider the following event:
 {
     "pubkey": "npub1p0ew5ln6hq4c20l7zftkwyrnn03php05zl8aydntugg74cu8m28sfk25c6",
     "kind": 0,
-    "content": "{\"nipdid\": \"nostr.bit\"}"
+    "content": "{\"nipdid\": [\"nostr.bit\"]}"
 }
 ```
 
@@ -50,6 +50,19 @@ For example, if a user clicks a button to follow the profile of `nostr.bit`, whi
 ### User set public keys can either be in `NIP-19` npub format or hex format
 Most user can only get their public key in `NIP-19` npub format, so it is difficult to force them to use hex format public key. So we should allow users to set their public key innpub format. As a result, client should consider both hex-format and `NIP-19` format when comparing the value.
 
+### The value of `nipdid` key is array instead of string
+In an ideal scenario, we should use a single DID name to handle all situations. However, from a practical perspective, many of us will have multiple DID names. Also, considering forward compatibility, the value of 'nipdid' should be an array of string instead of a single string. Currently, the client-side should always take the first string as the primary DID name.
+
+### The key of 'nipdid' has a higher priority than 'nip05'.
+This is because the authenticity and uniqueness of the DID name are guaranteed by a decentralized network, while the name in NIP-05 is issued by a centralized server, making it relatively less trustworthy than the DID name. 
+Therefore, when the user sets both NIP-DID and NIP-05, the client should prioritize the valid NIP-DID field and then use the NIP-05 field.
+
+### Keys for Nostr in DID Names
+By default, the key of Nostr public key stored in the DID name records should be `nostr`. However, since different DID names have different data storage methods, here are the relevant keys specified:
+- ENS: `nostr`
+- .bit: `profile.nostr`
+- Unstoppable Domain: `nostr`
+- Solana Name Service: `nostr`
 
 ## Terminology
 
@@ -58,7 +71,7 @@ When different people talk about DID, they are actually referring to different t
 
 The DID and DID name that we mention here are decentralized, human-readable, and formally similar to domain names. 
 
-Popular DIDs include `.bit`, `ENS`, `Unstoppable Domain`, `SpaceID`, `Bonfida`, etc. 
+Popular DIDs include `ENS`, `.bit`, `Unstoppable Domain`, `Bonfida`, etc. 
 
 Each DID has a different suffix. For example, .bit's suffix is directly `.bit`, ENS's suffix is `.eth`, and Bonfida's suffix is `.sol`. Clients can support one or more suffixes (DID) based on their own scenarios.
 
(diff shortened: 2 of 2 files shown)


eskema:
this nip does nothing in reality, it tries to create a new field for all possible did scenarios which is not optimal, instead, each of those did methods should be its own nip, just like nip05, so you'd have a nip for .bit and would maybe assign the field `bit` or `did_bit` or `nipnumber`, so those wanting to support that format know what to expect when they encounter it. no need to include any other possible scenario that the client won't handle, keeps the logic cleaner.

zgayjjf:
> this nip does nothing in reality, it tries to create a new field for all possible did scenarios which is not optimal, instead, each of those did methods should be its own nip, just like nip05, so you'd have a nip for .bit and would maybe assign the field `bit` or `did_bit` or `nipnumber`, so those wanting to support that format know what to expect when they encounter it. no need to include any other possible scenario that the client won't handle, keeps the logic cleaner.

Hello, @eskema,
Thank you for your feedback. Based on what you've described, it sounds like you're suggesting that instead of having one NIP that tries to handle all possible DID names, each DID name should have its own NIP. This would allow for a more modular approach, where clients can expect specific fields for each DID name they encounter.

However, I think the clients can choose which DID name(suffix) they support. They can handle those suffixes they would like to support and ignore those suffixes they don't want to support or don't know how to support.

The DID name suffix is another type of format of the field you mentioned.

I think this didn't make the logic more complex and can prevent the NIP boom(That means we have endless NIP for endless possible suffixes). 😄 

What do you think?

eskema:
> > this nip does nothing in reality, it tries to create a new field for all possible did scenarios which is not optimal, instead, each of those did methods should be its own nip, just like nip05, so you'd have a nip for .bit and would maybe assign the field `bit` or `did_bit` or `nipnumber`, so those wanting to support that format know what to expect when they encounter it. no need to include any other possible scenario that the client won't handle, keeps the logic cleaner.
> 
> Hello, @eskema, Thank you for your feedback. Based on what you've described, it sounds like you're suggesting that instead of having one NIP that tries to handle all possible DID names, each DID name should have its own NIP. This would allow for a more modular approach, where clients can expect specific fields for each DID name they encounter.
> 
> However, I think the clients can choose which DID name(suffix) they support. They can handle those suffixes they would like to support and ignore those suffixes they don't want to support or don't know how to support.
> 
> The DID name suffix is another type of format of the field you mentioned.
> 
> I think this didn't make the logic more complex and can prevent the NIP boom(That means we have endless NIP for endless possible suffixes). 😄
> 
> What do you think?

I already said what I think.. what you want clients to handle with those suffixes is worst than a direct field, because it forces you to make an additional check for no good reason

zgayjjf:
Sorry, the first paragraph is generated by ChatGPT. I've been overusing it a bit lately. 😉 

But different fields would force the client to check all the fields they support, right?

Say I'm a client supporting 10 DID names, then I need to check all the 10 fields to see which field the user used.

Further, if the user set multiple fields, the client has to know which one should it use.

eskema:
> But different fields would force the client to check all the fields they support, right?
> 
> Say I'm a client supporting 10 DID names, then I need to check all the 10 fields to see which field the user used.
> 

A client that supports 10 suffixes need to support that right? what's the difference?

zgayjjf:
@eskema, I think the main difference is one single field vs multiple fields, right?
Multiple NIPs are unnecessary, as we can define all the DID name-related fields in one NIP.

tuotuoTongtong:
just imagine when I have my own did in my moments not just like something that outsider can't understand, 
I would definitely use it
 Before:
Drop a public key somewhere else and say please follow this pubkey: npub1p0ew5ln6hq4c20l7zftkwyrnn03php05zl8aydntugg74cu8m28sfk25c6.

After:
Tell your friends to follow your DID name like nostr.bit. And that's all!.

barkyq:
1. I hope `nipdid` is just a temporary placeholder name, and it will eventually be replaced with something like `nip83`.
2. Potential issues if a client has set both `nip05` and `nipdid` field. What does the client display? Both ids?
3. Related to @eskema's point, what if a user has multiple `did`s, are they forced to just pick one?
4. What are the benefits of your proposal compared to, say, simply putting the did names in the "desciption"? Do clients really need to verify this?

zgayjjf:
Hi, @barkyq, Thanks for your feedback!

1. That's a temporary placeholder. A `nipxx` will replace it before we approve this proposal.

2. I think `nipdid` has a higher priority than `nip05`, as `nip05`'s data is hosted by a centralized server, which is easy to manipulate, and thus reduces its credibility. Here is an [example.](https://nost.vip/.well-known/nostr.json?name=npub1p0ew5ln6hq4c20l7zftkwyrnn03php05zl8aydntugg74cu8m28sfk25c6)

3. That's a good question. The value could be an array with multiple DID names. That would be more flexible. The first one of the array is mandatory, and the other ones are optional and can be displayed when needed. Can this solve the problem?

4. For forward resolution, which means finding a pubkey based on a DID name, we indeed do not need this NIP. The user only needs to set up the public key in their DID name.
However, for reverse resolution, which means finding a name based on a pubkey, we need this protocol to display the appropriate name to the user instead of the pubkey.

5. Actually, I have a project called [AllDID-API](https://github.com/dotbitHQ/AllDID-api) that can provide the HTTP API. However, I didn't want to mention it as I didn't want to limit the possibility of the Data Provider. But you've inspired me: I should include it in the examples to make the proposal clearer. I will improve the proposal tomorrow as it is late here.

melvincarvalho:
-1 web3 is an affinity scam

https://dev.to/melvincarvalho/the-web-a-brief-history-38df

melvincarvalho:
Regarding making a nostr did, better to wait for the DID folks to add schnorr signatures

https://w3c-ccg.github.io/security-vocab/#SchnorrSecp256k1Signature2019

Been waiting on this 4 years

Else we'd have to make our own vocab and context

zgayjjf:
@melvincarvalho  It's not a whole new Nostr DID; it's just leveraging the existing DID names to provide a better user experience. :P

zgayjjf:
@barkyq I have already converted the data type to an array that accepts multiple names. While the first one is what we use mostly.
And I have added 2 examples of Data Providers. 1 JavaScript SDK and 1 HTTP API. As you said, a basic HTTP Get request is the best.

melvincarvalho:
> It's not a whole new Nostr DID

Thanks for the clarification

So you want to use an arbitrary library to resolve the TLDs squatted by unstoppable domains:

https://resolve.unstoppabledomains.com/supported_tlds

TLDs like `.bitcoin` dont belong to unstoppable domains.  Any anyone could make such a registry.  Then the while idea of NIP5 verifications then become dubious.

The principle behind Linked Data which powers DID, is that "anyone can say anything about anything".  So there can be multiple different people claiming to be a .bitcoin or .x domain.  It's the equivalent of one of those websites that lets you name a star after yourself.

Clients are going to struggle to resolve this ordinarily in the same way they resolve relays or websites, via DNS.  It's also a lot more work for clients, a new attack surface, new things to debug, more support.  Also certainly many clients WONT support this leading to an inconsistent user experience.

I do see a possible exception to `.nostr` names which the community could come to a consensus on how to resolve, since nostr is so unique a name that only we use it.  But `.bitcoin`, `.crypto`, `.x`, `.dao` anyone can make up.  It doesnt make sense to me for verification.

jacany:
Sounds like larrys nip-69

vitorpamplona:
Closing this due to inactivity and @melvincarvalho's recent work at the W3C for did:nostr.

PR #283, opened 2023-02-22
Add NIP66 draft - Decimate

hendore:
This NIP paves the way for more control over a users data with the possibility of event management clients appearing to migrate / cleanup old events when switching to new preferred relays.

There are some considerations to take into account which is open for ideas. This may also be too dangerous to roll out until more users understand how to use delegated event signing. My biggest concern is a rouge client deleting all of a users events from all of their preferred relays. This concern is a current vulnerability with the current delete event kind, a bad client could receive all common events (0, 1, 4, 10002, etc...) and send a delete event with those ids tagged to many relays.

If anything comes out of this, maybe it's an educational push to get users creating client specific keys without the ability to delete anything 🤔

### DIFF
--- a/66.md
+++ b/66.md
@@ -1,0 +1,79 @@
+NIP-66
+======
+
+Decimate
+-------------------
+
+`draft` `optional` `author:hendore`
+
+A destructive event with kind `10`, the decimate event (from Latin decem, meaning "ten") will delete matching events stored on the receiving relay.
+
+The content is not used and SHOULD be an empty string.
+
+Relays SHOULD accept new events in the future from the same pubkey, this event only effects what is currently stored on the relay at the time it was published.
+
+Relays MAY delete the events from their database or soft delete but they SHOULD NOT emit those events in the future if soft deleted.
+
+## Matching
+
+Matches are based on event pubkey, relay tags and include/exclude tags.
+
+A list of `r` tags will be used to prevent accidental deletion from a preffered relay. Any relays receiving this event that are not defined in the list of `r` tags would ignore the event.
+
+Either an `include` or `exclude` tag must be provided but not both. The value of these tags contain a list of event kinds that should be taken into account when matching events to be removed.
+
+## Purpose
+
+The purpose of this NIP is to provide a way for users to retain control over the relays that store and share their events. Over time, it's likely users would find a few small sets of relays they prefer to publish to and would like to remove any/all events previously stored on relays used in the past.
+
+This NIP paves the way for more control over a users data with the possibility of event management clients appearing to migrate / cleanup old events when switching to new preferred relays.
+
+## Examples
+
+Given we publish the events below to the following relays
+
+- wss://example.relay.io
+- wss://another-example.relay.io
+- wss://some-new.relay.io
+- wss://another-new.relay.io
+- wss://one-more.relay.io
+
+***Deleting everything except profile and preferred relay metadata***
+
+Any events stored on the two relays listed in the `r` tags would be deleted except for events with the kind 0 or 10002. The remaining 3 relays that received the event can safely ignore it.
+
+```json
+{
+  "pubkey": "...........",
+  "kind": 10,
+  "tags": [
+    ["r", "wss://example.relay.io"],
+    ["r", "wss://another-example.relay.io"],
+    ["exclude", [0, 10002]]
+  ],
+  "content": "",
+  ...other fields
+}
+```
+
+
+***Deleting only notes from a single relay***
+
+All of the users kind 1 events would be deleted on `wss://one-more.relay.io` the remaining 4 relays that received the event can safely ignore it.
+
+```json
+{
+  "pubkey": "...........",
+  "kind": 10,
+  "tags": [
+    ["r", "wss://one-more.relay.io"],
+    ["include", [1]]
+  ],
+  "content": "",
+  ...other fields
+}
+```
+
+### Considerations
+
+Some thought may need to go into rouge clients, in it's current state a bad acting client may sign and publish a decimate event to many relays. [NIP26](26.md) if widely adopted could prevent this in the future if users only give clients enough power for what the client needs rather than their root keypair.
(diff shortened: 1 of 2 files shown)


dskvr:
NIP-66 is already taken #230

hendore:
@dskvr No problem, I can change this one, it wasn't to clear what nips have been reserved so I just went with the next value after the last merged NIP. Is there a method for finding the next nip id that hasn't been reserved yet?

barkyq:
Some concerns/questions:

1. When would someone want to use this? One can just send a `kind:5` with a bunch of `e` tags. Seems to cover most use cases. Kind of goes against the NIP-design principle that there should not be multiple ways to achieve the same outcome. This is I think the main problem with this proposal... One difference though is that it is on a per-relay basis (that cannot be achieved with `kind:5`). 
2. Agreed with @Semisol that it should only delete events before `created_at`. Otherwise there is a replay attack if some malicious user somehow gets a hold of the event (i.e., if they make a `REQ` for all `kind:10` events). Would not even require a "rogue client."
3. If this is supposed to be on a per-relay basis, maybe makes more sense as a command, e.g., `["DELETE", <filter>]`, used with NIP-42 (similar to `REQ`, but forcing the filters to have a single author equal to the NIP-42 authorized user).
4. Relays often do not know what their URL is, especially at the database level (i.e., the SQL database probably doesn't know the URL). Complicates the logic of checking whether or not to delete the events. To implement NIP-42, the relay URL needs to be passed to the software somehow. This offers support for why I think this could fit nicely with NIP-42 as outlined in 3.
5. What should a client do if it receives one of these events? Is it supposed to track which relay each event came from?
6. Also think this (and indeed `kind:5`) should be ephemeral, but that is just my own personal bias.

hendore:
> 1. When would someone want to use this? One can just send a `kind:5` with a bunch of `e` tags. Seems to cover most use cases. Kind of goes against the NIP-design principle that there should not be multiple ways to achieve the same outcome. This is I think the main problem with this proposal... One difference though is that it is on a per-relay basis (that cannot be achieved with `kind:5`).

I did wonder whether this was close to the existing Delete event however however there a few differences, targeting a single relay is one but also most relay software have limits on things such as max tags / request size so if attempting to remove a lot of data (after a long period of usage) from an old relay you no longer want holding your data then it may hit those limits.

> 2. Agreed with @Semisol that it should only delete events before `created_at`. Otherwise there is a replay attack if some malicious user somehow gets a hold of the event (i.e., if they make a `REQ` for all `kind:10` events). Would not even require a "rogue client."

Totally agree, 100% (though I think providing a from and until filter could also be useful, so long as until is not greater than the events created_at timestamp)

> 3. If this is supposed to be on a per-relay basis, maybe makes more sense as a command, e.g., `["DELETE", <filter>]`, used with NIP-42 (similar to `REQ`, but forcing the filters to have a single author equal to the NIP-42 authorized user).

That's a very good suggestion, I think it makes much more sense to be a command now you have suggested it.

> 4. Relays often do not know what their URL is, especially at the database level (i.e., the SQL database probably doesn't know the URL). Complicates the logic of checking whether or not to delete the events. To implement NIP-42, the relay URL needs to be passed to the software somehow. This offers support for why I think this could fit nicely with NIP-42 as outlined in 3.

👍 Brilliant, couldn't agree more.

> 5. What should a client do if it receives one of these events? Is it supposed to track which relay each event came from?

If we introduce a delete command rather than event then I'm assuming relays wouldn't broadcast the deletion to clients. Clients that have already fetched and cached events that have since been deleted don't need to do anything.

> 6. Also think this (and indeed `kind:5`) should be ephemeral, but that is just my own personal bias.

Makes sense.

To be honest, after the feedback so far I'm actually doubting if this is a good idea myself now, the idea was more around migrating my own data from one relay to another, we can re-publish events to a new relay and then it was a tidy up operation to delete those events from the old relay you no longer write to, the events still exist elsewhere but you no longer want a certain relay to provide them but now that feels like it shouldn't matter if another relay still has those events.

hendore:
I think I'll leave it open for another 24/48 hours, see if any other thoughts on this come in but I'm leaning more toward just closing this proposal now that I've thought about it more and heard feedback.

barkyq:
I think it is a reasonable proposal nonetheless. It is my self-imposed volunteer duty to come up with whatever criticisms I can, as you can determine by looking at some of my other comments.

hendore:
@barkyq I wouldn't want it any other way 🤘🏼

dskvr:
@hendore 
>  Is there a method for finding the next nip id that hasn't been reserved yet?
It's basically wild west, I had to do the same thing as you, just trying to not have to do it over and over again :)

staab:
My two cent is NACK, this seems like a good nich for another client app to fill. A new special-purpose client can handle the complexity and edge cases better than a new NIP that will have spotty support across implementations.

barkyq:
> My two cents is NACK, this seems like a good niche for another client app to fill. A new special-purpose client can handle the complexity and edge cases better than a new NIP that will have spotty support across implementations.

The one thing I think this proposal brings is a per-relay deletion method (right now, deletion events apply to any relay and client which receives them). Perhaps this use case is too niche though.

staab:
> The one thing I think this proposal brings is a per-relay deletion method (right now, deletion events apply to any relay and client which receives them). Perhaps this use case is too niche though.

That's a good point, now that I re-read the NIP for deletions. Maybe a simple solution would be to allow kind 7 to specify policy. Specifying whether to `propagate` the deletion, and whether the deletion should be `ephemeral` would probably be enough.

At any rate, deletes are never more than a polite request, so it seems like the use cases above rely too heavily on relay cooperation. Use case #1 above could be accomplished simply by instructing the private relay to ignore deletions. #2 is impossible assuming we're talking about valuable licensed content, piracy on nostr is just going to be a thing.

barkyq:
@staab yes very good points

staab:
Covered partly by NIP 62, partly by regular deletions and synchronization mechanisms.

PR #293, opened 2023-02-23
NIP-10: Add marked p tags

Semisol:


fiatjaf:
Is this implemented somewhere? Was it discussed?

eskema:
I was for marked p tags before, but I'm against now. they achieve nothing and promote sharing misinformation. if someone is to not be tagged for notifications, then the tag should be removed. we gain nothing and will overload old clients anyway. clients up to date should have a simple way to show the user who is beeing tagged and given the option to remove unwanted ones. 
this will simply introduce complexity with questionable benefits, unless I'm not seeing something...

mikedilger:
> clients up to date should have a simple way to show the user who is beeing tagged and given the option to remove unwanted ones.

Gossip had that once. I had to remove it for some temporary reason I don't even remember.

arthurfranca:
> Copy any `p` tags from the event being replied to, if any...

I know this is somewhat already written before this PR, but do we really want to notify all `p` chain when many levels deep in the tree? (Also it makes event heavier) For instance, does the root user want to be notified of a reply not directed to him, just because it is part of the thread he started?

> `silent`: This public key is being referenced in this note but should not be notified (such as public key lists).

Don't the mentioned pubkey deserve to be notified when not a private message? Sorry I didn't understand the "public key lists" example.

10.md
```diff
@@ -54,9 +54,33 @@ A direct reply to the root of a thread should have a single marked "e" tag of ty
...
+- `<pubkey>` is the public key being referenced.
+- `<relay-url>` is the URL of a recommended relay associated with this public key. This is optional and may be an empty string.
+- `<type>` is required. It can be:
+    - `reply`: This public key is involved in the reply thread.
+    - `mention`: This public key is mentioned in this note or was in the reply thread.
+    - `silent`: This public key is being referenced in this note but should not be notified (such as public key lists).
+    - A blank string.
+
+A client should follow the following steps while creating a note.
+A `p` tag should not be added if one with the same type and pubkey exists in the tag list.
```
barkyq, line 79: suggest removing "with the same type"?

v0l:
I think it would be better to just exclude the `p` tag and to simply put the pubkey in the content, then it would be up to clients to extract this pubkey as a mention which is clickable.

Now you have "mentioned" without notifying them.

fiatjaf:
You could also use a different tag, like `["pp", "pubkey"]` instead of `p`, so it won't show up as a notification. It has the benefit of not introducing yet another thing for clients to parse (although some clients will require a micro update).

PR #296, opened 2023-02-24
Add NIP-XX: Proof of Freshness

shafemtol:
Proof of freshness based on Bitcoin block hashes.

I came up with the use for this while working on another concept which I've currently named _secret broadcasts_, The idea of which is that clients subscribe to high-PoW events of a certain message kind (reception is supposed to be anonymous, therefore everyone subscribes to all events).

The potential issue I see is that an attacker, even with limited computing resources, can build up an attack over days or weeks by generating events with PoW offline in order to later publish these in a short time-span, flooding subscribers with a sudden wave of spam. Proof of freshness solves this by requiring that the proof of work is recent.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,7 @@ NIPs stand for **Nostr Implementation Possibilities**. They exist to document wh
 - [NIP-56: Reporting](56.md)
 - [NIP-57: Lightning Zaps](57.md)
 - [NIP-65: Relay List Metadata](65.md)
+- [NIP-XX: Proof of Freshness](XX.md)
 
 ## Event Kinds
 | kind          | description                      | NIP                     |
@@ -104,6 +105,7 @@ When experimenting with kinds, keep in mind the classification introduced by [NI
 | subject    | subject                 |                   | [14](14.md)              |
 | d          | identifier              |                   | [33](33.md)              |
 | expiration | unix timestamp (string) |                   | [40](40.md)              |
+| f          | block hash (hex)        |                   | [XX](XX.md)              |
 
 ## Criteria for acceptance of NIPs
 
--- a/XX.md
+++ b/XX.md
@@ -1,0 +1,78 @@
+NIP-XX
+======
+
+Proof of Freshness
+------------------
+
+`draft` `optional` `author:shafemtol`
+
+Proof of freshness proves that the id of an event, and thus its signature and
+associated proof of work, was generated after some specific point in time.
+
+The proof of freshness is provided in an `f` tag, with the tag value being the
+hexadecimal representation of the hash of a recent Bitcoin block header.
+
+
+Rationale
+---------
+
+Proof of work, described in `NIP-13`, can be used to deter spam by adding
+computational cost to the creation of events. However, it does not prevent a
+spammer from generating large amounts of events over time offline in order to
+later flood the network by publishing all these events at once.
+
+If required to also provide proof of freshness, a spammer is forced to perform
+all their proof of work in near real time, making them unable to perform such a
+flooding attack.
+
+Another use case of proof of freshness is to prove that a message was not signed
+ahead of time.
+
+A single-letter tag `f` is used in order to allow filtering on tag values in
+subscriptions, as described in `NIP-12`, thus protecting subscriptions from spam
+attacks.
+
+
+Choice of Block Hashes
+----------------------
+
+In the interest of reducing the subscription filter size for time ranges, block
+hashes should be chosen from a limited set.
+
+The first `f` tag provided in an event SHOULD contain the hash of the latest
+safe block whose height is divisible by 144. A block is considered safe if it's
+in the best chain with at least 5 blocks on top of it.
+
+144 blocks is roughly the number of blocks mined in 24 hours. An event MAY
+provide additional `f` tags with different divisibility choices.
+
+
+Relay Behavior
+--------------
+
+A relay MAY reject an event with an `f` tag referencing a block with a timestamp
+more than 2 hours ahead of the event's `created_at` timestamp.
+
+A relay MAY reject an event if it does not recognize the hash given in an `f`
+tag, as long as the relay has an up-to-date view of the best Bitcoin chain as of
+the event's `created_at` timestamp.
+
+
+Querying Example
+-------------------
+
+The following filter contains the hashes for blocks 777744 and 777888.
+Assuming events follow the rules given above for choice of block hashes, the
+filter subscribes to events with a certain proof of work made between
+approximately 2023-02-22 03:46 and 2023-02-24 00:18:
+
+```json
+["REQ", "subid",
+ {"ids": ["000000000"],
+  "#f": ["00000000000000000004744ecc695175c314a076bd888ca990140605ce02cf62",
+         "00000000000000000000470a063ad158dfb60983e6b0b81363aafc35dd5fcf77"]}]
+```
+
+By filtering this way, the client is guaranteed to only receive events whose
+proof of work has been generated after the time of the earliest block hash
+provided.


Egge7:
While I like the concept, the attack you describe here isnt really worrying me. The idea of POW in notes is to add a cost to spam. It still does, even when mined in the shadows. Spam usually relies on cheap distribution, otherwise the cost-benefit-ratio is negative.

Now here are my concerns: If this NIP would be implemented by Relays, scheduling posts would become a lot harder. Even if a note did not include any POW, signing would need to happen right before publishing, which might be easy for native clients with access to the keys, but impossible for web-clients, no matter how they implement scheduling. And scheduling is a very important tool for most creators.

shafemtol:
> If this NIP would be implemented by Relays, scheduling posts would become a lot harder.

The idea here is to keep `f` tags completely optional. The relay should not require an `f` tag under any circumstance as far as this NIP is concerned. Relays only have the option to do some sanity checks on `f` tags if provided. Otherwise this is completely client-based, applicable in more application-specific contexts or where the author actively chooses to prove that signing happened after some given point in time.

shafemtol:
Also, when I use the term _spam_, I mean it in a general sense, as applicable to someone who wants to disrupt a service, not necessarily profit directly economically from their activity.

Egge7:
> Otherwise this is completely client-based, applicable in more application-specific contexts or where the author actively chooses to prove that signing happened after some given point in time.

Thank you for clarifying. Then I am aligned.

MaxHillebrand:
I think this does not solve the problem for you.

Including a recent block hash is a proof of absence, meaning that the note cannot have been created before that block was mined. It proofs the lower bound, that the message is older than that block. The attacker can include a very old blockhash years after it was mined.

What you are looking for instead is a proof of existence, meaning that the note was created before a certain time. You can do this with NIP3 open timestamps, when including a reference to a note ID into a bitcoin transaction and block, you can verify the upper bound, that this note was written before that block was mined. If the attacker now years later proposes a new note with enough proof of work, then he still cannot include this note into the old block, so the attack is deterred.

Of course you can easily combine the two methods, include the recent bitcoin blockhash into your nostr message, then put the note ID into open timestamps. This way you proof both absence and existence, meaning you have a cryptographic proof of the time window when a note was created.

Egge7:
> Including a recent block hash is a proof of absence, meaning that the note cannot have been created before that block was mined. It proofs the lower bound, that the message is older than that block. The attacker can include a very old blockhash years after it was mined.

Thats a good point. In order for this to be a proof of freshness a Client would need to be synced the chain in order to lookup a blocks median timestamp. That is quite the overhead.

MaxHillebrand:
> Thats a good point. In order for this to be a proof of freshness a Client would need to be synced the chain in order to lookup a blocks median timestamp. 

Including a block cannot be a prior of freshness.

Look, here is the hash of block 1 00000000839a8e6886ab5951d76f411475428afc90947ee320161bbf18eb6048, but this doesn't mean that I wrote this message at the time of / before block 1. It proofs that it happened AFTER that block. 

For example, I cannot tell you the hash of block 1 million, because it has not yet been mined. 

On the other hand, if there would be a nostr note in block 1, then we would have proof that fiatjaf is in fact Satoshi and had the idea of nostr already back then. 


>That is quite the overhead. 

It can be done through a public block explorer rather efficiently.

shafemtol:
> Including a recent block hash is a proof of absence, meaning that the note cannot have been created before that block was mined. It proofs the lower bound, that the message is older than that block. The attacker can include a very old blockhash years after it was mined.

I think you're missing the idea here. The point is exactly to prove that a note's id was created after a certain block was mined. But the block is not chosen by the attacker. It must match a set of recent blocks by the filtering client. The client is only interested in recent notes, not historical ones. By using a filter as provided in the example, the client is guaranteed (by the promised behavior of any relay that follows NIP-01 + NIP-12) that any notes returned were created after the list of block hashes chosen by the client.

Proof of freshness does not solve the case where an attacker concentrates PoW on creating notes targeting a very old block hash. For that case, you can combine this NIP with an opentimestamps proof as you mentioned.

> Client would need to be synced the chain in order to lookup a blocks median timestamp

A client only needs to fetch the most recent block headers from somewhere. They are themselves proven by very large amounts of PoW, so including all headers all the way down to the genesis block is likely overkill in most instances. Still, headers are only 80 bytes per block.

eskema:
isn't this trying to achieve the same as [NIP-03](https://github.com/nostr-protocol/nips/blob/master/03.md)?

shafemtol:
> isn't this trying to achieve the same as [NIP-03](https://github.com/nostr-protocol/nips/blob/master/03.md)?

No, NIP-03 lets you prove that the information within the event existed _before_ a given point in time, whereas this NIP lets you prove that the event _id_ was generated _after_ a given point in time. The two are complementary and can be combined in order to prove that an event id was generated between two given points in time.

Another difference is that, AIUI, NIP-03 requires specific support from relays, whereas this NIP only requires NIP-12 support in order to be useful in querying.

barkyq:
Pretty niche IMO. Feels like trying to solve a currently non-existent problem. Has anyone noticed spam with POW yet? Adding yet more baggage to events just to prevent this fairly esoteric attack vector. Also requires fairly widespread adoption, otherwise by filtering on the `f` tag one will miss any events/users which do not use this.

Also I think should be a way to achieve something like this without requiring dependence on bitcoin (or any blockchain for that matter...)

Final thought, I think anti-spam is best handled at the relay level. That way, as a user, if I successfully post my message to a relay, I can be fairly confident my "subscribers" are getting it (rather than their client adopting some anti-spam filters). Just a personal opinion.

shafemtol:
Maybe I should not use the term spam, but DoS attack. Let me try to more clearly state the DoS scenario I have in mind.

Imagine a protocol where there are 1000 participants all connected to one relay and with the same subscription filter. Anyone who posts an event matching that filter gets their event out to all 1000 participants. Posting to the group should be anonymous, so posting may happen via some sort of proxy. The filter is made somewhat narrow with a PoW requirement, so that the events don't overload the relay. But an attacker can come along with a whole stack of events that they have generated the required PoW for over several days, and post them all at once. The relay gets overloaded as its outgoing messages start to queue up, resulting in a DoS situation that lasts until the attacker is out of PoW events.

If we want to put the responsibility on the relay for not getting into this situation, what could it do? It could try limiting the rate of subscription matches that any one entity is allowed to cause. But the problem, if we want to retain anonymity, is to apply this limit to each entity individually. You can't just put the limit on a per-IP-address basis, as the attacker might share IP address with other legitimate users, and the attacker may also perform a sybil attack, appearing through multiple IP addresses.

I see one alternative to proof of freshness in this particular scenario: The relay itself is PoW aware and, if detecting high load, might itself temporarily require some PoW for relaying events. It may weight this requirement per event based on the number of subscription matches at the time of posting. Two alternative ways of doing this:
- PoW in the event itself, as defined in NIP-13. A legitimate user will have to either wait or perform another PoW calculation at a higher difficulty and then perform another signing operation, while an attacker will simply delay or discard their event. A potential complication with this approach is if the event needs to go to multiple relays and was already posted on a different relay.
- Extra PoW sent through some extension of the client/relay messaging protocol.

The latter would perhaps be the most viable option. A downside of such an extension is that it requires both clients and relays to implement it, whereas this NIP for proof of freshness only requires client-side implementation. And one thing I quite like about nostr, as described in the [nostr README](https://github.com/nostr-protocol/nostr#readme):
> A relay is very simple and dumb. It does nothing besides accepting posts from some people and forwarding to others.

shafemtol:
Now, there are also other use cases for proof of freshness. One example is in a warrant canary, where you want to make sure the message was not signed ahead of time. See for example a [Qubes canary](https://github.com/QubesOS/qubes-secpack/blob/master/canaries/canary-033-2022.txt#L60-L87), where they use a recent Bitcoin block hash as a definitive proof of freshness.

> Also I think should be a way to achieve something like this without requiring dependence on bitcoin (or any blockchain for that matter...)

The alternative, if you want proof of freshness, is reports of news events (see e.g. Qubes canary above). You'll likely need multiple headlines from multiple sources in order to be somewhat reliable for a human to verify. More data than a simple block hash. Making a machine verify it would be difficult. Sure, you could check news RSS feeds and hope that they didn't change their headlines between the time of publishing and time of verification, that the headlines in the proof could not be predicted ahead of time, that the sites are up and not compromised, etc.

Or, with a block hash, you can check against _any_ third party to get the header, check that the header matches the hash, check that the hash starts with enough zero bits and extract the timestamp from the header. That is really all you need to be 100% certain that the proof of freshness is valid. You don't even have to know what a bitcoin or a blockchain is.

tx350z:
I'm just starting to look at Nostr because it is very similar in concept to something I've been working on for a couple years. My project started as a foundational framework for a distributed user-hosted MMO/RPG that somehow morphed into a purpose-agnostic ecosystem that is resistant to censorship and deplatforming. I don't believe the design of my framework is easily translatable to Nostr, but many of the issues being discussed here are addressed in the design. My hope is to find opportunities to contribute to Nostr.

My solution to address DDoS is a WAR ("whack-a-rodent") defense. It relies upon multiple servers (analog to relays in Nostr) to function as a loosely coupled cooperative cluster. Each server has it's own identity and an immutable affiliation identity that is shared by all of the servers in the cluster. In the event of a DDoS on one node, that node closes it's end-point to new connections, force closes the DDoS connection, notifies the cluster nodes of the attack, then redirects all remaining connected clients to other nodes in the cluster. The DDoS targeted node keeps it's end-point closed to new connections for a random period of time. I'm not sure if this approach can be useful to Nostr since (as I understand it) every relay is independent and functions only as a message distributor service.

If anyone is interested, I'd enjoy discussing this and other features that could be beneficial to Nostr.

shafemtol:
If 64 nibbles (32 bytes, 256 bits) feels too long, it can be cut down to the last 32 nibbles (16 bytes, 128 bits) and remain secure.

If storage is a concern: A relay implementation can optimize storage of `f` tags (along with `a`, `e` and particularly `p` tags!) by internally them with a reference to an entry of all tag values.

If bandwidth is a concern: WebSocket supports per-message compression with context takeover, meaning information in previous messages is used when compressing subsequent messages. Events in the same time period will have the same `f` tag, so if sent in succession, compression will remove the redundancy.

Issue #173, opened 2023-01-16
NIP-18 may need some improvement


RooSoft:
According to note1wlpjwra6xgeqzvf88l0rk6zkfhgz6en3vm2jmgzsp8nfw06qpaast77ut6, Damus is not implementing NIP-18 as is and doesn't intend do to so.

<img width="511" alt="Screenshot 2023-01-16 at 1 48 32 PM" src="https://user-images.githubusercontent.com/665949/212747986-6ea417bf-05d7-4f75-88fe-2ddf4ff04b44.png">

Its boost feature includes most of the reposted note's data in the contents field and this is apparently needed for many reasons.

Seems like snort is going in the same direction.

As a developer trying to follow the specs in these early days of the network, I already need to differ from what's documented if I want to remain compatible with the most prevalent clients.

fiatjaf:
You're right, I'm going to remove this NIP from here.

But if you're going to implement it like Damus is doing that will be very sad, since it is a hack that is probably not healthy for the network in the long term. I believe that's why William didn't propose it as a NIP himself. We'll see.

RooSoft:
Will do whatever is best for network concensus, you guys tell me what to do...

🫡

RooSoft:
Why is this issue closed? Has it been resolved already?

fiatjaf:
Yes, I've deleted NIP-18.

Giszmo:
Why? Why so impulsive? Where are the arguments? One second to the other we scrapped it just like that? Or is this your way of pushing us to something more decentralized than github?

fiatjaf:
Give me arguments for keeping a NIP that no one implements which I wrote and committed myself trying to describe what Damus was doing only to learn that Damus wasn't doing that.

Meanwhile Astral and probably Hamstr are doing the same thing in what seems to be a cleaner way, which is described in NIP-10.

Giszmo:
It's a very dynamic space right now, with new clients popping up every day. It's ok to nuke stuff but I would appreciate to get a chance to comment before nips are removed.

I agree with your decision to remove this nip but when nips disappear without a chance to ask for a timeout it feels like shaky grounds to implement less popular nips on. It just looks better to have two or three ACK between PR and merge.

On the subject of this nip:

* kind-6 with empty body adds little to kind-1 mentioning events via e-tags. It saves a few bytes over kind-6 with serialized event in content.
[...]

Issue #267, opened 2023-02-17
We need a way to request direct replies to a note


arthurfranca:
## EDIT: Hello client authors, please skip to [this comment](https://github.com/nostr-protocol/nips/issues/267#issuecomment-1455080477) to read the tl;dr of the last version of this proposal.

Currently, if we use the filter `{ kinds: [1], #e: ['note id'] }`, it won't bring only the direct replies to the note, making it hard (impossible?) to lazily (e.g. infinite scroll) build a thread, specially the ones with lots of engagement.

This is because of:
- [NIP-08](https://github.com/nostr-protocol/nips/blob/master/08.md) (Handling Mentions) supporting clients can add extra `e` tags mentioned in content
- [NIP-10](https://github.com/nostr-protocol/nips/blob/master/10.md) supporting clients can add many `e` tags (1 root, maybe 1 reply and maybe one or more mentions)

There is no way to fetch just the direct replies (that would be the NIP-10 with the root marker OR the NIP-10 with the reply marker; it depends).

Currently, nostr expects clients to fetch all thread at once (which will probably get limited by relays when the thread is big).
It gets really difficult when you use `limit` on the request filter as it will sort by created_at desc, so it may retrieve latest non-direct replies or mentions.
E.g.:  `{ kinds: [1], #e: ['root note id'], limit: 10 }` may bring 2 mentions and 8 replies not directed to the root note

**So i think we need to add a tag denoting *direct reply*. Maybe an `R` tag** (just one max per event).

**What do you think of adding it so that a client can use the following filter?**
`{ kinds: [1], #R: ['note id or long form content address'], limit: 10 }`

 And this should supersede NIP-10 reply marker.

staab:
I like this solution a lot. Replies are the one universal concept in social media applications, event tags are far too broad to support them efficiently.

barkyq:
I like this too. In general, would be nice to have efficient ways to build reply trees.

Use-case 1: Start at a leaf note and move to the root note. Use the `IDs` field in the query filter, taking the ID from the `R` tag.

Use-case 2: Start at root note and fetch the child notes (building the reply tree from top down). With this proposal, can use `R` query tag. Anticipated improvement in bandwidth + specificity of the query. Still will probably return a fair amount spam if spammers include the `R` tag linking to popular messages. Limiting only one `R` tag per note could mitigate that a bit though.

fiatjaf:
[...]

PR #310, opened 2023-03-01
Update NIP-10 to clarify usage of reposts and quote posts

alexgleason:
Per our discussion in https://github.com/nostr-protocol/nips/issues/173#issuecomment-1449105521, I'm proposing some changes to NIP-10. Note that NIP-10 is still in a draft state.

1. Renamed NIP-10 from "Conventions for clients' use of `e` and `p` tags in text events" to "Replies, Reposts, and Quote Posts in text events". I think this is a more realistic title, and will help implementers find what they're looking for.
2. Removed the "DEPRECATED" section of the document. The document seems like it was originally written to solve a problem that is no longer prevalent.
3. Clarified the meaning of markers, and described how they're used specifically to create reposts and quote posts.
4. Removed the "root" marker. Someone please tell me if there's history I'm missing behind this, but I can't comprehend why a "root" marker would be needed, especially because it's often very difficult or _impossible_ to add it.
5. Clarified that reposts and quote posts MUST reference the event like `#[0]` in the event content.

### DIFF
--- a/08.md
+++ b/08.md
@@ -12,6 +12,6 @@ Clients that want to allow tagged mentions they MUST show an autocomplete compon
 
 Once a mention is identified, for example, the pubkey `27866e9d854c78ae625b867eefdfa9580434bc3e675be08d2acb526610d96fbe`, the client MUST add that pubkey to the `.tags` with the tag `p`, then replace its textual reference (inside `.content`) with the notation `#[index]` in which "index" is equal to the 0-based index of the related tag in the tags array.
 
-The same process applies for mentioning event IDs.
+The same process is used to indicate related events. See NIP-10 for more information.
 
 A client that receives a `text_note` event with such `#[index]` mentions in its `.content` CAN do a search-and-replace using the actual contents from the `.tags` array with the actual pubkey or event ID that is mentioned, doing any desired context augmentation (for example, linking to the pubkey or showing a preview of the mentioned event contents) it wants in the process.
--- a/10.md
+++ b/10.md
@@ -28,7 +28,7 @@ Where:
 
 When `"mention"` is used, the event content MUST include an index marker with the notation `#[index]` as specified in NIP-08, where the index matches the position of the `"e"` tag in the tags array. Note that a repost is a quote post without comment. For example, a quote post's content may look like `hello world\n#[0]`, while a repost would contain only `#[0]`.
 
-Clients MUST NOT include multiple `"e"` tags with the same marker in an event. The order of "e" tags is not relevant.
+Clients MUST NOT include multiple `"e"` tags with `"reply"` or `"root"` markers. The order of "e" tags is not relevant.
 
 ## The "p" tag
 Used in a text event contains a list of pubkeys used to record who is involved in a reply thread.
--- a/10.md
+++ b/10.md
@@ -1,39 +1,13 @@
 NIP-10
 ======
 
-Replies, Reposts, and Quote Posts in Text Events (kind 1).
---------------------------------------------
+Replies, Reposts, and Quote Posts in Text Events (kind 1)
+---------------------------------------------------------
 
 `draft` `optional` `author:unclebobmartin` `author:alexgleason`
 
 ## Abstract
 This NIP describes how to use "e" and "p" tags in text events to denote replies, reposts, and quote posts.
-
-## Positional "e" tags (DEPRECATED)
->This scheme is in common use; but should be considered deprecated.
-`["e", <event-id>, <relay-url>]`  as per NIP-01.
-
-Where:
-
- * `<event-id>` is the id of the event being referenced.
- * `<relay-url>` is the URL of a recommended relay associated with the reference.  Many clients treat this field as optional.
-
-**The positions of the "e" tags within the event denote specific meanings as follows**:
-
- * No "e" tag: <br>
- This event is not a reply to, nor does it refer to, any other event.
-
- * One "e" tag: <br>
- `["e", <id>]`: The id of the event to which this event is a reply.
-
- * Two "e" tags:  `["e", <root-id>]`, `["e", <reply-id>]` <br>
- `<root-id>` is the id of the event at the root of the reply chain.  `<reply-id>` is the id of the article to which this event is a reply.  
-
- * Many "e" tags: `["e", <root-id>]` `["e", <mention-id>]`, ..., `["e", <reply-id>]`<br>
-There may be any number of `<mention-ids>`.  These are the ids of events which may, or may not be in the reply chain.  
-They are citings from this event.  `root-id` and `reply-id` are as above.
-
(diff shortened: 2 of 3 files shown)


mikedilger:
I think you've gone a bit overboard.

The deprecated section helped me to understand how to interpret events that are created in clients that use the deprecated method. Many events in nostr use that deprecated way of doing things, even new events on old clients or clients that haven't bothered to change, and it is good to have documented how clients should interpret such events, even if clients should not be generating events in that way anymore.

The "root" marker helps clients find the entire thread. If you have that marker, you can get all the events that reference it in a single relay query.  Otherwise you have to go back and forth to the relay getting the parent, then the grandparent, then the great grandparent, etc.   If you find it impossible to add a root marker, then just add a reply marker and clients will have to deal with the lack of information by climbing.  Among a pride of well behaved nostr clients (pride is the collective noun of a group of ostriches) when you reply to an event, that event has the root tag, you just copy it.

I didn't actually read your changes yet, I'm just replying to your comment.

arthurfranca:
I've created a [PR](https://github.com/alexgleason/nips/pull/1) to @alexgleason 's branch to fix [#293](https://github.com/nostr-protocol/nips/pull/293) and [#267](https://github.com/nostr-protocol/nips/issues/267). Also added footnote linking to previous versions of NIP-10 because of what @mikedilger pointed out.

fiatjaf:
Aside from removing the "root" marker and deleting the deprecated part of the NIP I like the clarifications introduced here.

As @mikedilger said, it is not impossible to add the "root".  If the event you're replying to has no `e` tags it is the root, if it has a "root" tag you copy it, otherwise there is nothing you can do, but so far all clients have been properly adding "root" tags so it's been working.

alexgleason:
Thank you both for the feedback, here are the latest changes:

- Added back the "deprecated" section describing positional "e" tags.
- Added back the "root" marker.

Appreciate the explanation about the root marker @mikedilger, makes perfect sense!

monlovesmango:
I like these changes, much clearer. 

>Clients MUST NOT include multiple "e" tags with the same marker in an event.

Shouldn't you be able to have multiple 'mention' events tagged?

alexgleason:
I updated the document with these changes:

1. Moved "deprecated" section to the bottom, and added additional text about Kind 6.
2. Updated the text to support multiple quotes per post, as pointed out by @monlovesmango

alexgleason:
My first instinct on NIP-10 was that you shouldn't be able to mention multiple events, and that it shouldn't be part of the text. When I created this MR I sought to document the system as it currently exists, but after seeing this insanity on my timeline I think we should consider limiting to ONE mentioned event:

image

Egge7:
Don't mind me just adding my 2 sats here... I really don't like the "subscribe to the root ID to get the thread approach". This will work as long as nostr is early, but with rising adoption this is an incredibly inefficient way to construct a thread. You can even see this today with a "Jack-post" that has hundreds of replies. Suddenly building a thread means fetching multiple megabytes of data... 

In a perfect world a reply would include all its thread event's ids, maybe even with an additional Marker 'ancestor'. A reply would always include it's parents id as a reply e Tag and additionally include all other r tags as ancestors. Then clients could fetch all events, loop through them and build a chain by looking at each events marked 'reply' tag

arthurfranca:
@Egge7 you described issue 267. Your opinion (or should I call it display of reality) is important for the repository maintainers to notice that replies are currently flawed.
[Take a look at this](https://github.com/nostr-protocol/nips/issues/267#issuecomment-1455080477) and see how what you say is similar to what I propose. The difference is I came to the conclusion that an extra level/depth tag is a way to keep `e` tags for all ancestors while being able to access any level down the reply tree, not just the direct descendants.

fiatjaf:
Then a conversation with 100 back-and-forth responses would get you events with 100 tags, which is horrible for scalability and also a ton of data.

If you're hitting threads with over 100 replies, or even a thousand replies, then you probably need relay-level filtering. No one wants to read 100 replies to any comment, at least not all at once. And 99% of these are worthless "I agree" or "you suck" replies anyway (not to mention the spam). A good client will allow users to filter by relay so they can opt to read only valuable (as deemed by the relay choices) replies or read everything if they really want.

arthurfranca:
> Then a conversation with 100 back-and-forth responses would get you events with 100 tags, which is horrible for scalability and also a ton of data

Not true. A leaf (in the reply tree) will only store `e` tags related to the branch it is part of. It won't know about other branches.

For instance, 100 direct replies to the root will start 100 new branches. So such events only have one ancestor, the root one.

If another event replies to one of above 100 replies, it will have only 2 ancestors.

arthurfranca:
"What if a branch grows 100 levels deep?" – one may ask

I've already antecipated it in one of my closed PRs. See below
> A relay MAY limit events to a max level. As a recommendation, clients SHOULD keep 4 as a maximum depth (`l` tag as "4") by not appending an `e` tag if already at level "4"

Considering a branch A-B-C-D-E, following above recommendation will simply consider F a reply to D, instead of to E. Could even add a `p` tag with E pubkey making it possible to say "replying to E pubkey" on UI.

monlovesmango:
@alexgleason from the looks of your example the problem is nested mentions are pulling in, not that there are multiple mentions

This PR initially was created to just clarify wording and I'd say let's keep to that scope and merge this PR as is to avoid having multiple in flight PR's to the same NIP.

I'm not opposed to changing the methodology of tagging events but I think thats a separate discussion that is being had in PR [We need a way to request direct replies to a note #267](https://github.com/nostr-protocol/nips/issues/267). Lets keep the discussion around tagging events methodology going there so there is a central place for our rationale.

alexgleason:
The goal of this PR was not just to improve wording, but to clarify what a repost _is_ and how to create one.

I'm not sure we should merge this until we address this comment: https://github.com/nostr-protocol/nips/issues/173#issuecomment-1478465856

The idea that a repost is defined as a kind 1 note with `"#[0]"` is something I was skeptical of when I first discovered it. After seeing it in action and hearing a little bit of pushback from other people, I think we should reconsider using kind 6 events which are still in common use.

monlovesmango:
not opposed to bringing back kind 6, however kind 6 is only for reposts to my understanding. quote posts would still need to be able to use the mention 'e' tag, right?

Egge7:
> not opposed to bringing back kind 6, however kind 6 is only for reposts to my understanding. quote posts would still need to be able to use the mention 'e' tag, right?

Quote-Posts could be kind 6 with a non-empty content field

monlovesmango:
not opposed to that. but we would still need `root`, `reply`, and `mention` markers then for kind 6 then.

alexgleason:
With NIP-18 being restored, and NIP-27 being added, most of the concerns this MR was trying to address are now resolved.

PR #318, opened 2023-03-02
Update NIP-10

### LINKED ISSUE #267: We need a way to request direct replies to a note

arthurfranca:
fix #267, #293, #173

The goal is to bring together all current fixes in one place because some open conflicting PRs were touching the same text.

### DIFF
--- a/10.md
+++ b/10.md
@@ -1,62 +1,81 @@
 NIP-10
 ======
 
-
-On "e" and "p" tags in Text Events (kind 1).
+Replies, Reposts, and Quote Posts in Text Events (kind 1).
 --------------------------------------------
 
-`draft` `optional` `author:unclebobmartin`
+`draft` `optional` `author:unclebobmartin` `author:alexgleason` `author:Semisol` `author:arthurfranca`
 
 ## Abstract
-This NIP describes how to use "e" and "p" tags in text events, especially those that are replies to other text events.  It helps clients thread the replies into a tree rooted at the original event.
+This NIP describes how to use "n" and "l" tags in text events to denote replies. Also "e" tags for reposts, and quote posts and "p" tags for notifications.
 
-## Positional "e" tags (DEPRECATED)
->This scheme is in common use; but should be considered deprecated.
+## The "n" tag
+It means reply a[n]cestor. The ordered "n" tag set lists all ancestors of the reply event until reaching the root. Used in conjunction with one "l" tag to associate an event with another event as a **reply**.
 
-`["e", <event-id>, <relay-url>]`  as per NIP-01.
+Format: `["n", <event-id or event-address>, <relay-url>]`
+- `<event-id or event-address>` address of the replied to event if replying to a [long-form content](https://github.com/nostr-protocol/nips/blob/master/23.md) or else regular event id is used.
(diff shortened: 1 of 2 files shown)


arthurfranca:
@fiatjaf could you review this PR please por favorzin?

The #n (before it was #b) and #l tags are needed for requesting just slices of a thread, while still allowing for a full thread request (as explained [in this comment](https://github.com/nostr-protocol/nips/issues/267#issuecomment-1450591924) with example filters – I know it is a big comment but NIP-10 is so important and I think it will allow for better clients, that show the most interesting replies at top, using a fraction of the mobile internet quota)

In the PR text footnote I point to deprecated info but I could add old text explicitly if it is better. Maybe even recommending mixing both ways for maximum compatibility.

@Semisol @mikedilger @barkyq @staab please join too

fiatjaf:
What we needed to do was clarify this NIP so everybody is implementing the same thing, that's what Alex was doing, but here you are introducing breaking changes, so that's not good. As I said on the other PR, we can add the new way of referencing events _as a third section on this NIP_ if all the biggest clients are excited about the idea of changing their logic, otherwise it makes no sense.

arthurfranca:
> ...if all the biggest clients are excited about the idea of changing their logic...

Hard requirement. Thanks for reviewing. I had the wrong assumption that a rewrite was ok because the load-all-thread-at-once way would still be supported, although with different tags, thus making it a better standard cause it would allow more query options/styles without sacrificing the current load-everything style. Oh well I guess this is like betamax defeat 😅

As the PR is in the wrong direction I will close it and add that as a third section if and when more people like the idea at [#267](https://github.com/nostr-protocol/nips/issues/267).

PR #320, opened 2023-03-03
Polls RFC 69

toadlyBroodle:
Request for Comment on new Poll event (kind #6969)
-Poll notes are paid poll events on nostr used to conduct quantitative opinion polls that users can vote on with zaps

### DIFF
--- a/51.md
+++ b/51.md
@@ -104,9 +104,9 @@ Some clients have used these lists in the past, but they should work on transiti
   "kind": 30004,
   "tags": [
     ["d", "jvdy9i4"],
-    ["name", "Yaks"],
-    ["picture", "https://cdn.britannica.com/40/188540-050-9AC748DE/Yak-Himalayas-Nepal.jpg"],
-    ["about", "The domestic yak, also known as the Tartary ox, grunting ox, or hairy cattle, is a species of long-haired domesticated cattle found throughout the Himalayan region of the Indian subcontinent, the Tibetan Plateau, Gilgit-Baltistan, Tajikistan and as far north as Mongolia and Siberia."],
+    ["title", "Yaks"],
+    ["image", "https://cdn.britannica.com/40/188540-050-9AC748DE/Yak-Himalayas-Nepal.jpg"],
+    ["description", "The domestic yak, also known as the Tartary ox, grunting ox, or hairy cattle, is a species of long-haired domesticated cattle found throughout the Himalayan region of the Indian subcontinent, the Tibetan Plateau, Gilgit-Baltistan, Tajikistan and as far north as Mongolia and Siberia."],
     ["a", "30023:26dc95542e18b8b7aec2f14610f55c335abebec76f3db9e58c254661d0593a0c:95ODQzw3ajNoZ8SyMDOzQ"],
     ["a", "30023:54af95542e18b8b7aec2f14610f55c335abebec76f3db9e58c254661d0593a0c:1-MYP8dAhramH9J5gJWKx"],
     ["a", "30023:f8fe95542e18b8b7aec2f14610f55c335abebec76f3db9e58c254661d0593a0c:D2Tbd38bGrFvU0bIbvSMt"],


01.md
```diff
@@ -16,7 +16,7 @@ The only object type that exists is the `event`, which has the following format
 
 ```json
 {
-  "id": <32-bytes lowercase hex-encoded sha256 of the the serialized event data>
```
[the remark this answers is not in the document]
toadlyBroodle: just fixing "the the" typo

69.md
```diff
@@ -0,0 +1,99 @@
...
+  "created_at": <unix timestamp in seconds>,
+  "kind": 6969,
+  "tags": [
+    ["e", <32-bytes hex of the id of another event>, <recommended relay URL>],
+    ["p", <32-bytes hex of the key>, <recommended relay URL>],
+    ["poll_options": [
+      ["0", <poll option 0 description string>],
+      ["1", <poll option 1 description string>],
+      ["n", <poll option n description string>],
+    ],
```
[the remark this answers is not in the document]
toadlyBroodle: Ok how is this syntax?

69.md
```diff
@@ -0,0 +1,99 @@
...
+    ["tally_method", "value"||"count"],
+    ["consensus_threshold", <integer (0..100)>],
+    ["closed_at", <unix timestamp in seconds>],
```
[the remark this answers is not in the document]
toadlyBroodle: Done

README.md
```diff
@@ -95,18 +97,23 @@ When experimenting with kinds, keep in mind the classification introduced by [NI
...
+| g                   | geohash                 |                   | [12](12.md)              |
+| nonce               | random                  |                   | [13](13.md)              |
+| subject             | subject                 |                   | [14](14.md)              |
+| d                   | identifier              |                   | [33](33.md)              |
+| expiration          | unix timestamp (string) |                   | [40](40.md)              |
+| poll_options        | options (integer array) | description (str) | [69](69.md)              |
+| poll_option         | poll option (integer)   |                   | [69](69.md)              |
+| tally_method        | "value" OR "count"      |                   | [69](69.md)              |
+| consensus_threshold | consensus % (integer)   |                   | [69](69.md)              |
+| closed_at           | unix timestamp (string) |                   | [69](69.md)              |
```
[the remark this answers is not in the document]
toadlyBroodle: Ok, removed.

vitorpamplona:
I would break poll by value and by amount into separate events. In that way:
- new nips can create new talking methods
- clients can declare full compliance to a specific kind and avoid open-ended specs.

toadlyBroodle:
There have been some concerns raised during private conversations that I think are important to be part of the public review process here. They may be largely summarized as two different attack vectors for potential poll manipulations: 1) single whale votes discrediting many shrimp participants' votes 2) fake/bot accounts swaying outcomes via vote flooding. To address both attack vectors, in the purpose section I've elaborated on the intent of 'value' vs 'count' tally methods and the strengths and weaknesses of both. Additionally, I've added a optional 'value_minimum' tag for specifying a minimum satoshi value for votes to be included in tallies.

toadlyBroodle:
There have been some concerns raised during private conversations that I think are important to be part of the public review process here. They may be largely summarized as pointing to two different attack vectors for potential poll manipulations: 1) single whale votes discrediting many shrimp participants' votes 2) fake/bot accounts swaying outcomes via vote flooding. 

To address both attack vectors, I have eliminated the tally_method tag altogether and replaced its 'value' and 'count' attributes with optional 'value_minimum' and 'value_maximum' tags for specifying satoshi value limits on zaps to be included in tallies. Imposing upper limits on satoshi values of tallied zaps should decrease the likelihood of whales swamping shrimps, while imposing lower limits should make fake vote flooding attacks prohibitively expensive.

Additionally this resolves @vitorpamplona's suggestion, while significantly simplifying the standard, and allowing for the maximum freedom in poll design. 

I've included further details in the revised specification.

toadlyBroodle:
> Should there be a max number of choices (in [zap polls for Damus story](https://github.com/damus-io/damus/issues/402) I suggested four (4) max; this is the expected pattern from cagedbird)? I can imagine e.g. 100 choices being unwieldy for a mobile client. Is ten (10) too many? Is five (5)?

@alltheseas I've attempted to specify the least restrictive rule set here to ensure polling functions correctly and fairly, while allowing devs the greatest freedom to implement other non-critical details as they see fit. A large number of poll_options does not break anything inherent to the polling process, so I think the responsibility is on clients to impose such limits individually, if they choose.
 
> Is this variable `consensus_threshold` grouped with a minimum number of votes? If one (1) user votes, is consensus reached - i.e. does voting end prematurely with `consensus_threshold` surpassee, or is this evaluated after the poll closes?

consensus_threshold is the minimum percentage for any single poll_option's percentage of the total tally to attain poll consensus status. It's intended as a simple 'measuring bar' and has no effect on prematurely closing polls. If only one person votes in a poll, then consensus is sure to be attained, when specified: how meaningful it is, depends on other statistical measures. It should probably be publicly visible before closing. Maybe this should be elaborated on in the spec?

> Is there a minimum hard-coded for `value_minimum` at 1 sat? Are free polls possible with this NIP? Are combined free + zap polls possible?

Lightning wallets impose their own limits on minimum payment amounts. Don't think it's necessary to re-iterate here that values need be a ['positive decimal integer'](https://github.com/lightning/bolts/blob/master/11-payment-encoding.md). By definition, zaps are not free. Certainly free polls are possible, but this NIP defines paid polls.
 
> If poll votes take place with zaps, is there a list of voters displayed?

Clients can display any additional event information they choose.

> Is there an option for a non-zap poll vote?

No.

toadlyBroodle:
> this looks great, but needs some changes. also, in the future, we may need polls that count people that fall under specific criteria, having a NIP-05, 2 degrees of separation from the user that does the poll, ... but that can be left for later

Thanks @Semisol. Do the changes since your last review adequately address your concerns? 

I think specifying `value_minimum` and `value_maximum` amounts for inclusion in tallies will allow for flexible poll designs that can successfully deter certain undesired participants. But definitely it would be nice to include options for more specific voter filters in the future.

fernandolguevara:
@toadlyBroodle 

here is a draft for surveys/polls #111 that defines a survey/poll by using tags instead of a new kind

Why don't we decorate kind *1* with tags indicating that we want to start a survey/poll?

toadlyBroodle:
> Why don't we decorate kind _1_ with tags indicating that we want to start a survey/poll?

@fernandolguevara zap poll events are a significant departure from kind `1`, particularly in how clients must handle them. Pretty sure they warrant their own 'kind' designation, if only to allow clients to query relays specifically for poll events.

staab:
My two cents is that I'd like to see a way to vote "other" with a custom value. It's already commonly done, and when it's not, polls almost always shoehorn you into a false dichotomy. 

I like zaps for polls, and given the choice between zap-only and non-zap-only polls I'd prefer the former. But a solution that accepts both based on a required level of participation set by the poll author might be a good thing.

toadlyBroodle:
> My two cents is that I'd like to see a way to vote "other" with a custom value. It's already commonly done, and when it's not, polls almost always shoehorn you into a false dichotomy.

@staab Poll makers are free to include an 'other' poll_option, users may include comments in zap 'content' field, and clients can choose how to render these comments.

toadlyBroodle:
@vitorpamplona excellent feedback on some key omissions!

>     1. Poll should specify a relay to collect&tally votes. Client applications should send votes there. In that way, there will be no result inconsistency between users that are using different relay sets

Great point, added requirement for primary relay inclusion.

>     2. Shouldn't it require tallying clients only to count if zap votes were sent to the author? Otherwise fake Zaps can easily take over. How do we protect participants against the author faking their own zaps?

Absolutely! This got me thinking of how to solve both problems at once: allowing for distribution of funds across predetermined public-keys. This way pollers can design more authentic/legitimate polls by allowing voters a choice for where they send their funds. This also addresses @alltheseas's last question, and enables distribution of funds functionality that was of importance to @StackSats.IO.

>     3. "Received" time (`MUST NOT tally late votes received after closed_at time`) is irrelevant. Most clients will download these events after pools close. The receiving time should never be used in Nostr. 
>     4. Clients can create events in the past, adding votes after the `closed_at` date by creation time. Zappers can also send zaps in the past, especially if the zapper is controlled by the author or a party interested in certain results. Maybe this is a good use for NIP-03

Great catch of a glaring omission on my part and excellent suggestion for a fix. Added requirement for `ots` when `closed_at` time is specified.

69.md
```diff
@@ -0,0 +1,117 @@
...
+```json
+{
+  "id": <32-bytes lowercase hex-encoded sha256 of the serialized event data>
+  "pubkey": <32-bytes lowercase hex-encoded public key of the event creator>,
+  "created_at": <unix timestamp in seconds>,
+  "kind": 6969,
+  "tags": [
+    ["e", <32-bytes hex of the id of the poll event>, <primary poll host relay URL>],
+    ["p", <32-bytes hex of the key>, <primary poll host relay URL>],
+    ["poll_options", "{
```
Cameri, line 45: I would change this to be one tag entry per poll option:

```
["poll_option","0","description"],
["poll_option","1","description"]
```

With that we avoid having to parse JSON again.

toadlyBroodle: @Cameri changed recursive json serialized `poll_options` tag to multiple poll_option tags for each vote option.
I also added a rule for limiting each participant to a single vote when min/max values are equal, while allowing participants to change their original vote choice.
Additionally, I added a Zap Voting section to clarify zap event submission rules for voting clients.

toadlyBroodle (commented):
Do these changes resolve your review?

braydonf:
Considering that a zap receipt is _not a proof of payment_, the poll results can be manipulated to achieve any preferred poll result.

From NIP-57:
> "The zap receipt is not a proof of payment, all it proves is that some nostr user fetched an invoice. The existence of the zap receipt implies the invoice as paid, but it could be a lie given a rogue implementation."
https://github.com/nostr-protocol/nips/blob/master/57.md#appendix-e-zap-receipt-event

57.md
```diff
@@ -96,7 +96,11 @@ Example zap note:
       [
         "preimage",
         "5d006d2cf1e73c7148e7519a4c68adc81642ce0e25a432b2434c99f97344c15f"
-      ]
+      ],
```
v0l, line 99: This tag should be inside the 9734 not the 9735

toadlyBroodle:
> Considering that a zap receipt is _not a proof of payment_, the poll results can be manipulated to achieve any preferred poll result.
> 
> From NIP-57:
> 
> > "The zap receipt is not a proof of payment, all it proves is that some nostr user fetched an invoice. The existence of the zap receipt implies the invoice as paid, but it could be a lie given a rogue implementation."
> > https://github.com/nostr-protocol/nips/blob/master/57.md#appendix-e-zap-receipt-event

Yes, this is a known issue that applies to all zaps, not just polls. A solution will hopefully emerge soon.

57.md
toadlyBroodle: Fixed

toadlyBroodle (commented):
Moved poll_options tag from 9735->9734

fiatjaf:
Is this implemented somewhere? I saw some poll things on Amethyst. Is it ready to merge?

Can we rename it to "Zap Polls" or something like that? Just to differentiate it from non-paid polls that could be done in some other NIP later?

toadlyBroodle:
> Is this implemented somewhere? I saw some poll things on Amethyst. Is it ready to merge?

Yes, it's already released in Amethyst v0.32+


> Can we rename it to "Zap Polls" or something like that? Just to differentiate it from non-paid polls that could be done in some other NIP later?

Yeah absolutely.

fiatjaf:
Can you fix the conflicts?

v0l:
I don't think we event implement this spec as described though, there is some reference to e/p tags and that was just ignored, idk if you want to update the NIP to reflect this or we should discuss with @vitorpamplona why he implemented it like this

toadlyBroodle:
> I don't think we implement this spec as described though, there is some reference to e/p tags and that was just ignored, idk if you want to update the NIP to reflect this or we should discuss with @vitorpamplona why he implemented it like this

Most of the spec is implemented as described (except additional recipients and ots), however @vitorpamplona wanted to disable all optional features for first release. Once it's proven stable, I'm sure we'll start rolling out the extra features: min/max values, consensus, closing time, etc.

`e`/`p` tags are same as kind `1`, additional `p` tags will probably be implemented as additional recipients, but I'm waiting on possible wallet splits to add this feature. `e` tag in kind `9734` reference the original poll note, while `p` is the zap recipient.

braydonf:
I don't think Zaps are solving the Sybil-attack issue, even if it was a proof-of-payment; though it certainly doesn't help.

I think a solution to that is to only consider poll results from those that you follow, everything else is ignored. This could be done with another Poll specification or with this one. If kept with zaps, I agree with @fiatjaf to rename to "Zap Polls".

fiatjaf:
I agree with @braydonf that zaps don't provide sibyl-resistance, my preference is to only trust poll votes from relays that offer sibyl-resistance, but I didn't want to disrupt this NIP.

toadlyBroodle:
> IMHO min/max values, consensus, and closing time are too complicated for an average user that just want to do a weak poll. We will probably have a default setting for them and some semi-hidden way to customize: a fixed zap amount for the communists out there (lol), closing time of 24 hours (similar to Twitter to simplify), and a default consensus number based on the number of options.

I coded up most of the features for Amethyst for an initial proof of concept. If you don't want to include them all (or just use defaults) that's fine, as extras are optional to give clients maximum flexibility in how they implement, while users can choose which implementation they prefer. There do seem to be alot of communists on nostr, who knew!

toadlyBroodle:
> I agree with @braydonf that zaps don't provide sibyl-resistance, my preference is to only trust poll votes from relays that offer sibyl-resistance, but I didn't want to disrupt this NIP.

Didn't realize there were such relays, how does that work? At any rate, clients are free to show, or not, any notes they choose. If certain clients want to only show polls from certain relays, they're free to do so, as every vote must specify the same primary relay as the original poll event.

> I think a solution to that is to only consider poll results from those that you follow, everything else is ignored. This could be done with another Poll specification or with this one.

Basically the main point of this NIP is to standardize how votes are counted across clients to avoid ambiguity of tallies. We can certainly add further optional tally limitations, at the cost of complexity.

fiatjaf:
> Basically the main point of this NIP is to standardize how votes are counted across clients to avoid ambiguity of tallies.

This is necessarily impossible on Nostr. The entire point of Nostr is that each client will have a different view of the network depending on what relays they connect to.

fiatjaf:
I'm reading the spec now and I see you have a "main relay" of the poll. This is weird, because it makes it so only people allowed by that specific relay can vote, when in fact it would be much nicer to allow anyone to vote, but then each viewer can see a different result depending on the relay.

For example, if you ask "is inflation good?", you could see an overwhelming majority of "no" on a relay used mainly by Bitcoin enthusiasts and an overwhelming majority of "yes" on a relay used by economists -- and all sorts of interesting experiments could take place.

I also see a bunch of idiosyncratic requirements like "min zap amount", "max zap amount", "closed at" time that do not make sense on Nostr. The way I see this is: you allow anyone participate in any way they want -- specially because you _cannot_ prevent them and you count the results later according to your own criteria.

For example, I could have a poll for "who is going to win the election" and people can keep voting as time passes, then I can filter the results in different timeframes and get completely different results and see the evolution of people's predictions.

I see your requirements might be related to the idea of achieving "consensus", but I don't get the purpose of that. How can you be in consensus with a bunch of sibyls on the internet? Maybe there is a hidden use case there that I am missing.

toadlyBroodle:
@fiatjaf The whole point of polling is to determine public consensus, yes? This nip defines the protocol rules that make this possible across compliant clients. I don't see the point in having every client render different outcomes for every user, as this defeats the entire purpose of polling in the first place. A primary relay is necessary if all applicable votes are to be counted. Yes rogue clients may submit non-compliant votes, however compliant clients will not include non-compliant votes in the tally, according to the specifications.

Regarding sibyl attacks: the best defense against sibyl attacks is attaching a cost to participation. As far as I know, zaps are the best tool we have for doing so on nostr. Yes it is still vulnerable to sibyl: perhaps we could add a PoW component to kind `9735` events to help mitigate this?

toadlyBroodle:
> This is necessarily impossible on Nostr. The entire point of Nostr is that each client will have a different view of the network depending on what relays they connect to.

Requiring a primary hosting relay makes it possible.

jimbojw:
I suspect that even perfect proof of payment would not be Sibyl resistant. An attacker could spin up any number of voter npubs and have them “pay” invoices that the attacker receives. It may bleed a few sats due to routing fees, but the bulk of funds would be preserved, and the bleed could be mitigated by channel management.

I suspect the only way to truly prevent fake voting is to make the payment irrecoverable via proof-of-work, as described for example in NIP13.

69.md
```diff
@@ -0,0 +1,126 @@
...
+  * MUST include the same primary hosting relay URL as the `p` tags
+* MUST contain a primary description string, specified in the `content` field
+* MUST contain at least 2 `poll_option` tags, each specifying an index and a unique description string, formatted as below
+* SHOULD specify a `closed_at` time:
+  * when specified, MUST include a valid [`ots` tag](03.md), proving original poll publishing time
+  * a `closed_at` value of null or less than or equal to the `created_at` field indicates a poll SHOULD NOT be closed.
+* MAY specify a `value_maximum` satoshi value for zapped votes to be included in the tally
+* MAY specify a `value_minimum` satoshi value for zapped votes to be included in the tally
+  * `value_minimum` MUST be less than or equal to `value_maximum`, when both are specified
+* MAY include a `consensus_threshold` (0-100), representing a percentage threshold for any single vote option to achieve poll consensus
```
jimbojw, line 35: Perhaps clarify that `consensus_threshold` applies only if `closed_at` was specified and only after that time has passed. Otherwise, would not the first vote cross any specified threshold immediately? Or can the threshold be crossed and then uncrossed? Not sure what this means.

69.md
```diff
@@ -0,0 +1,126 @@
...
+A careful balancing of all poll attributes should enable pollers to conduct tailored polls that deliver meaningful and valuable outcomes.
+
+## Zap poll format
+
+A poll event:
+* MUST specify 1 or more `p` tags, each including the same primary hosting relay
+* if an `e` tag is specified, it:
+  * MUST include the same primary hosting relay URL as the `p` tags
+* MUST contain a primary description string, specified in the `content` field
+* MUST contain at least 2 `poll_option` tags, each specifying an index and a unique description string, formatted as below
```
jimbojw, line 28: Must the indices be numeric? Must they be in order? If so, could they be omitted and inferred from the order they appear in the tag list?

What if an index is repeated? Would this invalidate the poll? Is it undefined behavior?

69.md
```diff
@@ -0,0 +1,126 @@
...
+
+## Zap vote format
+
+The zap request event (kind `9734`): 
+* MUST specify an `e` tag that references the original poll event
+  * MUST include a primary hosting relay URL in the `e` tag
+* MUST include at least 1 `p` tag specifying the recipient's key, chosen from the original poll event's list of keys 
+  * MUST include the same primary hosting relay URL in all `p` tags as is specified in the `e` tag
+* MUST include exactly 1 `poll_option` tag which references the chosen vote option by its corresponding index `n`
+* MUST include a valid `ots` tag, if the original poll event specifies a `closed_at` time
```
jimbojw, line 80: Where is the `ots` tag described? I’m unable to deduce its meaning from context.

69.md
toadlyBroodle: When no `closed_at` specified, `consensus_threshold` will fluctuate with every new vote

69.md
toadlyBroodle: Indices must be numeric, must be present, but don't need to be in order, as they are strings. Clients can decide how to handle non-compliant, undefined, event structures.
Defined structure:
>["poll_option", "0", "poll option 0 description string"],
>["poll_option", "1", "poll option 1 description string"],
>["poll_option", "n", "poll option <n> description string"]

69.md
toadlyBroodle: [NIP03](03.md)

toadlyBroodle:
> I suspect that even perfect proof of payment would not be Sibyl resistant. An attacker could spin up any number of voter npubs and have them “pay” invoices that the attacker receives. 

Recent addition of [zap forwarding](https://iris.to/note1txtlardxrcyhqdaq4wtwc4tp6ez67ap7uvf8ygu6adqqyama6alq5z03d7) should help mitigate poll authors' ability to self-zap.

> I suspect the only way to truly prevent fake voting is to make the payment irrecoverable via proof-of-work, as described for example in NIP13.
Yes this may eventually be needed.

toadlyBroodle:
@Semisol @v0l @Cameri @jimbojw just merged Zap Polls with recent changes and remove last TODOs. 
As both Snort and Amethyst have now implemented required aspects of NIP69, can we get this merged into protocol?

fiatjaf:
Yes, but why is this PR editing 22 files?

toadlyBroodle:
ah good ol' CRLF, sorry, didn't catch that.

toadlyBroodle (commented):
Reverted unnecessarily changed files with CRLFs, added `poll_option` references to 57.md examples

69.md
jimbojw: Thanks. Recommend linking this in the text for future readers.

```
* MUST include a valid ([NIP03](./03.md)) `ots` tag, ...
```

toadlyBroodle: Added

vitorpamplona:
Can we merge this? Amethyst and Snort implement this NIP and has been working well.

staab:
I realize this ship has already sailed, but NACK on making zaps a hard dependency of other nostr features. Payment proof should be open ended and optional.

alexgleason:
Agreed @staab. We will need a different type of poll to bridge polls from ActivityPub.

vitorpamplona:
@staab @alexgleason I don't see this as the only poll implementation in Nostr and I generally agree that the dependency on zaps is not good. However, I do think for those using zaps, this is already good enough.

fernandolguevara:
> @staab @alexgleason I don't see this as the only poll implementation in Nostr and I generally agree that the dependency on zaps is not good. However, I do think for those using zaps, this is already good enough. 

I think we can make this better, Dependency on zaps sucks...
We need a better way to make this use case over the network ... 
#111

toadlyBroodle:
> I haven't merged it yet because it is breaking the README and I haven't had time for or remembered fixing it myself.

Fixed

vitorpamplona:
I am coming here again asking to merge this. Let's move on.

staab:
Consensus seems to be that dependency on zaps is bad. I vote against this PR.

staab:
There have been 3-4 other poll implementations. We could vote on those or someone could make a new NIP. I'd rather not merge for no reason and make a bad idea a standard. Do we have a tally of how many implementations each one has? We've had polls in various places for over a year, I remember some in branle.

toadlyBroodle:
> There have been 3-4 other poll implementations.

I've yet to see a single other NIP that comes anywhere close to as comprehensive as NIP69.

57.md
```diff
@@ -37,6 +37,7 @@ In addition, the event MAY include the following tags:
 
 - `e` is an optional hex-encoded event id. Clients MUST include this if zapping an event rather than a person.
 - `a` is an optional NIP-33 event coordinate that allows tipping parameterized replaceable events such as NIP-23 long-form notes.
+- `poll_option` is a tag used for voting by [zap poll events](69.md).
```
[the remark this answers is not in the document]
toadlyBroodle: This is under optional MAY conditional list, for documentation purposes. It is a required field for zap-poll event zap requests: which is why it's included here. Still want it removed?

README.md
```diff
@@ -1,201 +1,197 @@
...
-1. They should be implemented in at least two clients and one relay -- when applicable.
-2. They should make sense.
-3. They should be optional and backwards-compatible: care must be taken such that clients and relays that choose to not implement them do not stop working when interacting with the ones that choose to.
-4. There should be no more than one way of doing the same thing.
-5. Other rules will be made up when necessary.
-
-## License
-
-All NIPs are public domain.
+- [List](#list)
```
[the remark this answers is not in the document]
toadlyBroodle: ?

69.md
```diff
@@ -0,0 +1,120 @@
...
+
+## Zap vote format
+
+The zap request event (kind `9734`): 
+* MUST specify an `e` tag that references the original poll event
+  * MUST include a primary hosting relay URL in the `e` tag
+* MUST include at least 1 `p` tag specifying the recipient's key, chosen from the original poll event's list of keys 
+  * MUST include the same primary hosting relay URL in all `p` tags as is specified in the `e` tag
+* MUST include exactly 1 `poll_option` tag which references the chosen vote option by its corresponding index `n`
+* MUST include a valid [NIP03](./03.md) `ots` tag, if the original poll event specifies a `closed_at` time
```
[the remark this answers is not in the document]
toadlyBroodle: removed

PR #336, opened 2023-03-09
NIP-33: Fixed example with multiple values

lapulpeta:
Fixed example so that the statement is true:
> Events from the same author with any of the following `tags`
replace each other:


fiatjaf:
I don't understand this. Why is the current example wrong?

lapulpeta:
Shouldn't all the `d` tags be the same so that they can replace each other? The last one has a `d` tag set to `test`, which is different than the rest (empty string).

fiatjaf:
I see, you're right.

It's super confusing, though, why use an empty string as the example?

lapulpeta:
For me it's not confusing as long as the examples are right. I had to re-read the NIP multiple times until I was fairly sure it was a mistake 😅

Issue #300, opened 2023-02-25
Idea: Save hash of external media in the event


jonas-lundqvist:
I posted this on [reddit](https://www.reddit.com/r/nostr/comments/11an7xr/nip_idea_save_hashes_of_external_media_within_the/), but I don't know if it reached the relevant people there so here is a copy-paste:

Since external media (primarily images) are stored externally and Nostr notes just contain the URL users are relying on the host of said media to always be online and not modify the content. One mitigation for users to verify that whatever image someone posted in a note is immutable would be to save the hash of it within the note.

I propose to handle this as such:
An URL to an image inside .content should have the form `[<url> | <hash>]` where `<hash>` is a SHA256 hexencoded string of the raw image data.

Example:

```
{
  "content": "Nostr in sheer decline: [ https://asdf.com/89asdf.gif | 396c4a297752a82a080f61fb8a4614c5ff3bef0f666830e70276a4f3465d2597 ]\n\nIt was all a dream.",
  "created_at": 1677184528,
  "id": "<id>",
  "kind": 1,
  "pubkey": "3bf0c63fcb93463407af97a5e5ee64fa883d107ef9e558472c4eb9aaaefa459d",
  "sig": "<sig>",
  "tags": [],
  "relays": [
    "wss://nostr.mom"
  ]
}
```
Old clients can still just post the URL and clients aware of this NIP could display it with an "unverified" notification or such. Once clients starts to post images in the new format old clients will display some "garbage" data. I see this being a rather small change and would like some feedback from clients.

If anything similar has been proposed (and rejected) please let me know.

shafemtol:
Not sure about the specifics of how and where to include such a hash, but strongly support the concept.

In fact, I would rather like to see all media links directly addressed by their hash, then perhaps another replaceable event could resolve that hash to a URL. That way, if some host goes down, only the URL event needs to be replaced to update the URL, the original event stays intact.

shafemtol:
Another thing to consider: To account for media that can be streamed or otherwise viewed before fully downloaded, one may want to use a Merkle root rather than a plain hash. That way, if given either the full list of leaf hashes or a Merkle proof on parts of the leaf hashes, a client can verify the downloaded parts before viewing them.

Content type should perhaps also be explicitly stated.

fiatjaf:
The best way to do it would be to add tags to the event containing the URL hashes.

mikedilger:
It's a good idea as a tag.

v0l:
[...]

PR #337, opened 2023-03-09
Nip 94 - File Header

frbitten:
A first approach to organize the dissemination and management of files in the NOSTR protocol

@jonas-lundqvist I used the Hash suggestion so that whoever downloads it can confirm that the file is correct and is still the same as at the time of the event

@kinakuta-co I used your idea to allow encrypted files to avoid server censorship.

@fiatjaf It may generate a certain conflict of interest with the NIP-78, but I preferred to make a separate nip than suggest a change in the NIP-78.


staab:
I'm a yes on both. This spec is well-written and very simple. It's nice that 94 and 95 are separate so relays can advertise support for one and not the other. As an aside, https://github.com/nostr-protocol/nips/pull/259 would allow non-supporting relays to recommend a server they trust to host NIP 95 files.

The one reservation I have is making these replaceable. That is how links work on the internet, but it leads to a lot of possible rug-pull type scenarios. Maybe `name` could be a separate tag, and `d` could be an optional "slot" (uuid or semantic pointer of some kind).

fiatjaf:
Looks good to me, but I missed the link between the two.

frbitten:
> I'm a yes on both. This spec is well-written and very simple. It's nice that 94 and 95 are separate so relays can advertise support for one and not the other. As an aside, #259 would allow non-supporting relays to recommend a server they trust to host NIP 95 files.
> 
> The one reservation I have is making these replaceable. That is how links work on the internet, but it leads to a lot of possible rug-pull type scenarios. Maybe `name` could be a separate tag, and `d` could be an optional "slot" (uuid or semantic pointer of some kind).

The d tag refers to the NIP-33 that defines replaceable events. I suggested the id be the file name to be more readable than any ID. But you can put an ID and use the "title" tag to put the name.

I imagine that the fact that they are replaceable generates some implementation complications. But since there is already this behavior defined for events, I thought it fit the scenario. But I'll only be sure when I start trying to implement the proposal.

frbitten:
> Looks good to me, but I missed the link between the two.

I first thought of NIP-95 but the problem of how to propagate it without overloading the network by sending very large events worried me. When reading the other nips and issues about files, I came up with the idea of NIP-94 being a file header, which would solve the problem of disclosing the NIP-95 event and other files for external sharing.

As I put it in the NIP-95, it should not be returned in searches on the relay. So the only way to know that the NIP-95 event exists is via a NIP-94 event. With this, the NIP-95 is only transmitted when someone really wants to access the complete file.

fiatjaf:
OK, that sounds good.

The only question I have is: is anyone really going to want to store very large files like that? Does that really make sense in practice? It's ok to consider it as a possibility, but it is a bad idea to have a NIP that no one will ever implement.

mikedilger:
minor: please put endquotes after the JSON, I can't read all that red text.

94.md
```diff
@@ -0,0 +1,44 @@
...
+The purpose of this NIP is to allow an organization and classification of shared files. So that relays can filter and organize in any way that is of interest.
+Also the goal is to create a base on the protocol for this bountsr "Filesharing App" (https://bountsr.org/p2p-filesharing/) to be implemented.
+
+Nostr event
+------------------
+This NIP specifies the use of the `30063` event type (parameterized replaceable event), having in `content` a description of the file content, and a list of tags described below:
+* `d` containing the name of the shared file to allow it to be overwritten in the future.
+* `url` the url to download the file
+* `type` a string indicating the data type of the file. The MIME types format must be used (https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types)
+* `decrypt` (optional) If the file is encrypted, you must indicate in the first value the algorithm used to encrypt the file and in the second value the information (the key) that allow the decryption of the file
```
v0l, line 14: `Key` is probably too specific here, Maybe more generically `Decryption Params` because not every encryption scheme will use the same params

v0l: Ref: https://github.com/bitwiseshiftleft/sjcl/blob/master/core/convenience.js#L14

v0l:
> I'm a yes on both. This spec is well-written and very simple. It's nice that 94 and 95 are separate so relays can advertise support for one and not the other. As an aside, #259 would allow non-supporting relays to recommend a server they trust to host NIP 95 files.
> 
> The one reservation I have is making these replaceable. That is how links work on the internet, but it leads to a lot of possible rug-pull type scenarios. Maybe `name` could be a separate tag, and `d` could be an optional "slot" (uuid or semantic pointer of some kind).

Is it possible to use the event id when referencing the file? Or will relays just delete the old event when its replaced? 

Would be kinda nice to have a history actually.

Egge7:
I proposed something very similar a couple of months ago (https://github.com/nostr-protocol/nips/pull/112). Back then the general consensus was, that we should not be storing base64 encoded files on relay, due to the inefficiency of that approach. Therefore I came up with https://github.com/nostr-ing/nostr-ing-protocol which is very similar to this approach, but moves NIP-95 to a subprotocol that allows the transmission and storage of raw binary data: https://github.com/nostr-ing/nostr-ing-protocol.

frbitten:
> I proposed something very similar a couple of months ago (#112). Back then the general consensus was, that we should not be storing base64 encoded files on relay, due to the inefficiency of that approach. Therefore I came up with [nostr-ing](https://github.com/nostr-ing/nostr-ing-protocol) which is very similar to this approach, but moves NIP-95 to a subprotocol that allows the transmission and storage of raw binary data.

I will read your protocol suggestion later and comment on it.
On the question of performance it is not true at all. There are ways to store that lose a little in space but can gain in availability.
The point is that relays that use a relational database are impractical to implement NIP-95. You will need a No-SQL database such as mongodb, which even already has tests and uses with storing large files. The main advantage of using mongodb is the ease of replicating data across multiple servers.
You can read more details here:
https://www.mongodb.com/developer/products/mongodb/storing-large-objects-and-files/
https://www.mongodb.com/docs/manual/core/gridfs/

An alternative to base64 is BSON which can also be used. I opted for Base64 because it is something better known, but I can include the BSON option in NIP-95 as well.

But my idea of the NIP-95 is sharing small files and data. Icons, images. And just be the communication protocol and not necessarily the storage. As I described, the relay can store it on disk and give http access to the file.

frbitten:
@fiatjaf  I think I should split this PullRequest in two. One for NIP-94 and one for NIP-95. I see that the NIP-95 will demand much more discussion and the approval of one does not depend on the approval of the other.

Egge7:
> I will read your protocol suggestion later and comment on it. On the question of performance it is not true at all. There are ways to store that lose a little in space but can gain in availability. The point is that relays that use a relational database are impractical to implement NIP-95. You will need a No-SQL database such as mongodb, which even already has tests and uses with storing large files. The main advantage of using mongodb is the ease of replicating data across multiple servers. You can read more details here: https://www.mongodb.com/developer/products/mongodb/storing-large-objects-and-files/ https://www.mongodb.com/docs/manual/core/gridfs/

I dont get your point. Storing a file in Base64 will increase required storage space compared to storing the same file in binary. Of course there is compression that will reduce the impact.

The reason why mongoDB has no issues storing large files is pricesely that they do not store base64 encoded data, but BSON (as you mentioned yourself), which supports raw binary. Adding BSON to this nip wouldn't make sense IMHO because the nostr protocol does not support the transmission of BSON anyways. Therefore relays would have to do all the encoding and decoding BSON -> Base64 and then clients would again have to decode base64 -> to binary to use the file.

frbitten:
I moved the NIP-95 to a Pull Request of its own. We can continue the discussion about it in #345

frbitten:
> I dont get your point. Storing a file in Base64 will increase required storage space compared to storing the same file in binary. Of course there is compression that will reduce the impact.

Base64 encoding adds 1 byte to the data for every 3 original binary bytes. Which means that storing in base64 increases the final space used.

vitorpamplona:
Isn't a replaceable event a bad design here? 

Think about a file that is highly shared as a tag inside a TextNote (TextNote references the file). Clients can easily preview the file as part of the experience. Now the file author's keys leak and the attacker replaces the file (and hash) with porn. Everyone will have a post citing content they didn't intend in the first place. 

My suggestion is to make a regular event. Once created, you cannot replace it.

v0l:
I would implement this on Snort if it's not a replaceable event and has some notes on how you might use these events as attached files:

```
Clients MAY reference this file header with a marked e tag using the marker "attachment"
```

Using NIP-51 lists it would still be possible to reference the non-replaceable file header event and prevent getting rugged.

arthurfranca (changes requested):

frbitten:
My knowledge of replaceable events was very superficial at the time of the proposal. After better understanding how it works, I agree with the suggestion that it should be a regular event. For this I changed the kind to 1063 to be a regular event and it will not be allowed to change the header of the file after it is published.
If the owner needs to change it, he must remove the event (NIP-09) and create a new one.

frbitten:
> I would implement this on Snort if it's not a replaceable event and has some notes on how you might use these events as attached files:
> 
> ```
> Clients MAY reference this file header with a marked e tag using the marker "attachment"
> ```
> 
> Using NIP-51 lists it would still be possible to reference the non-replaceable file header event and prevent getting rugged.
I would love to see this suggestion implemented in Snort.

But I didn't understand your comment or suggestion. Could you explain in more detail to try to understand.

arthurfranca:
> Clients MAY reference this file header with a marked e tag using the marker "attachment"

I see this NIP is useful to something like the [example it lists](https://user-images.githubusercontent.com/21368082/224114562-90a9ce66-0079-4c5a-8813-7e954de54992.png). But maybe it is not very useful for kind 1 mentions?

Considering just the image use case, I wonder if it will make a difference for reusing same image on kind 1 notes.
Cause when a client converts the mentioned NIP-94 event id to a regular url (from url tag) or to a base64 data url (from NIP-95 e tag[s]), it will place it in an e.g. img tag src (if a web app).

So users when reposting the image in an new kind 1 note will do:
1) copy image url (https://xxx.webp or data uri) then paste it
2) or downloag it (like to have a folder full of memes one likes) then re-upload it

For 1. the original NIP-94 event won't be reused and it will embbed the same url 👍or the base64 data uri (from NIP-95) 👎
On 2. the original NIP-94 event won't be reused and the reuploaded image embedded in the new kind 1 note would be a new NIP-94 + new NIP-95 e tag 👎or a new NIP-94 + new url tag 👎or just a new url 👎

Even [this suggestion that uses webtorrents](https://github.com/nostr-protocol/nips/issues/300#issuecomment-1450123862) will become an img tag in the end of the day, with the same 2 problems.

____
Solutions? Don't know but maybe the following helps:
NIP-95(#345) should be served as an url, like https://relaydomain/nostr_media/sha256(or blur?)_hash_of_file_content?alternative_relay[]=xxx&alternative_relay[]=xxx, instead of as an event with base64 content, so:
1) if a user copy and paste the url, it won't be a base64 one 👍
2) if a user downloads and then tries to reupload the image, first the client can calculate the hash and see if there is a NIP-94 event with the same hash, reusing it 👍
*<- 2. although there is a problem if client does image compression, changing the hash, so it should maybe search by both hashes (before and after compression)?*
*<- 2. also another problem: at what relay will the client search for the NIP-94 event?*

Also, maybe videos and other media should be a separate NIP (instead of NIP-95), cause it would probably be a problem hashing big files client side for then searching for the corresponding NIP-94 by hash?

frbitten:
The idea of this NIP-94 is just a simple way to index and organize file sharing via the various ways currently possible. It would work as a URL spreader only.

Users could comment on the quality/usefulness of the file. They could disclose with boosted. But I don't see much use in using your reference in a kind 1 post. In the case of kind 1, the img tag is a better solution.

In my opinion, it would be a use of NOSTR separately from conversation notes. For example, today we have Folows, Global and may include another organization called Files. Or have applications specialized in file sharing. Instead of looking for torrent files on obscure internet sites I can use a NOSTR client.

For the case of using the NIP-95, it might make some sense for me to be able to reference the event with the base64 data for them to appear in my kind 1 note.

94.md
```diff
@@ -0,0 +1,49 @@
...
+The purpose of this NIP is to allow an organization and classification of shared files. So that relays can filter and organize in any way that is of interest.
+Also the goal is to create a base on the protocol for this bountsr "Filesharing App" (https://bountsr.org/p2p-filesharing/) to be implemented.
+
+Nostr event
+------------------
+This NIP specifies the use of the `1063` event type (parameterized replaceable event), having in `content` a description of the file content, and a list of tags described below:
+* `url` the url to download the file
+* `type` a string indicating the data type of the file. The MIME types format must be used (https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types)
+* `decrypt` (optional) If the file is encrypted, you must indicate in the first value the algorithm used to encrypt the file and in the second value the parameters that allow the decryption of the file
+* `hash` containing the SHA-256 hexencoded string of the file or the BlurHash string (for images). The third array value is the lowercase chosen algorithm name: "sha256" or "blurhash".
```
arthurfranca, line 14: Just ocurred to me that `h` instead of `hash` will make it possible to find by the tag, so to know if a file header already exists before effectively uploading a file.

```suggestion
* `h` containing the SHA-256 hexencoded string of the file or the BlurHash string (for images). The third array value is the lowercase chosen algorithm name: "sha256" or "blurhash".
```

94.md
```diff
@@ -0,0 +1,49 @@
...
+{
+  "id": <32-bytes lowercase hex-encoded sha256 of the the serialized event data>,
+  "pubkey": <32-bytes lowercase hex-encoded public key of the event creator>,
+  "created_at": <unix timestamp in seconds>,
+  "kind": 1063,
+  "tags": [
+    ["url",<string with URL of file>],
+    ["decrypt",<algorithm>,<Decryption Params>],
+    ["type", <MIME type>],
+    ["hash", "LKN]Rv%2Tw=w]~RBVZRi};RPxuwH", <"blurhash" or "sha256">]
```
arthurfranca, line 26: `h` instead of `hash` to make it searchable
```suggestion
    ["h", "LKN]Rv%2Tw=w]~RBVZRi};RPxuwH", <"blurhash" or "sha256">]
```

94.md
```diff
@@ -0,0 +1,49 @@
...
+======
+`draft` `optional` `author:frbitten` 
+
+The purpose of this NIP is to allow an organization and classification of shared files. So that relays can filter and organize in any way that is of interest.
+Also the goal is to create a base on the protocol for this bountsr "Filesharing App" (https://bountsr.org/p2p-filesharing/) to be implemented.
+
+Nostr event
+------------------
+This NIP specifies the use of the `1063` event type (parameterized replaceable event), having in `content` a description of the file content, and a list of tags described below:
+* `url` the url to download the file
```
arthurfranca, line 11: `r` is already used elsewhere as "reference" with url value.

The reason to make it searchable is that I imagined a client may want to have an user media panel that the user would fill by using a context menu pop-up over images on notes (like "add to memes collection" for instance).

For that, it would detect used image url on kind 1 note content and create a corresponding kind 1063 event. Making 1063 url tag searchable will allow the client to know if an event with that url from the same pubkey already exists.

It would be an alternative to searching by hash.

```suggestion
* `r` the url to download the file, without trailing slash
```

94.md
```diff
@@ -0,0 +1,49 @@
...
+* `hash` containing the SHA-256 hexencoded string of the file or the BlurHash string (for images). The third array value is the lowercase chosen algorithm name: "sha256" or "blurhash".
+
+```json
+{
+  "id": <32-bytes lowercase hex-encoded sha256 of the the serialized event data>,
+  "pubkey": <32-bytes lowercase hex-encoded public key of the event creator>,
+  "created_at": <unix timestamp in seconds>,
+  "kind": 1063,
+  "tags": [
+    ["url",<string with URL of file>],
```
arthurfranca, line 23: `r` instead of `url` to make it searchable

```suggestion
    ["r", <string with URL of file>],
```

arthurfranca (changes requested):

94.md
nryo-o:
```suggestion
    ["r",<string with URI of file>],
```
@arthurfranca note I'm using `URI` instead of a `url` since this reference should be an identifier, not necessarily a location.
I suggest using a general uri to accommondate different ways of receiving data.

94.md
```diff
@@ -0,0 +1,49 @@
...
+
+The purpose of this NIP is to allow an organization and classification of shared files. So that relays can filter and organize in any way that is of interest.
+Also the goal is to create a base on the protocol for this bountsr "Filesharing App" (https://bountsr.org/p2p-filesharing/) to be implemented.
+
+Nostr event
+------------------
+This NIP specifies the use of the `1063` event type (parameterized replaceable event), having in `content` a description of the file content, and a list of tags described below:
+* `url` the url to download the file
+* `type` a string indicating the data type of the file. The MIME types format must be used (https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types)
+* `decrypt` (optional) If the file is encrypted, you must indicate in the first value the algorithm used to encrypt the file and in the second value the parameters that allow the decryption of the file
```
nryo-o, line 13:
```suggestion

```
Why would we want public files to be encrypted?
I propose this, although optional, can be discarded completely in favor of simplicity.

94.md
frbitten: I didn't put it as a letter because I don't see why the Hash tag should be indexed. I see no reason to do a database hash lookup.

94.md
frbitten: The idea of encryption in public files is that the server hosting the file does not know its contents. Thus avoiding censorship on the hosting server. It is a suggestion from NIP-68 (#250 ) that I have incorporated into the proposal.

v0l:
> > I would implement this on Snort if it's not a replaceable event and has some notes on how you might use these events as attached files:
> > ```
> > Clients MAY reference this file header with a marked e tag using the marker "attachment"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Using NIP-51 lists it would still be possible to reference the non-replaceable file header event and prevent getting rugged.
> 
> I would love to see this suggestion implemented in Snort.
> 
> But I didn't understand your comment or suggestion. Could you explain in more detail to try to understand.

Instead of clients posting `http://` links to images in text notes, we can use [NIP-27](https://nips.be/27) style `nostr:nevent` links which point to file headers.
We could also include the `kind` entry in the `nevent` to give clients a hint that its a file and not some other note (which might be important for clients who don't want to have to load every mentioned event, [NIP-19](https://nips.be/19))

```json
{
  "content": "Hey check out this meme!\nnostr:nevent1qqqqqqqqqqq",
  ...
}
```

Where `nevent1qqqqqqqqqqq` points to a NIP-94 file header

v0l:
Here is an example event:
```json
{
  "content": "Testing NIP-94 \nnostr:nevent1qqstxx3lk7zqfyn8cyyptvujfxq9w6mad4205x54772tdkmyqaay9scrqsqqqpp8x4vwhf",
  "created_at": 1680687580,
  "id": "2cab915f26f3d440e8c34330c822f3cb0ac946c04401fb26d2cf6cfb79814bba",
  "kind": 1,
  "pubkey": "63fe6318dc58583cfe16810f86dd09e18bfd76aabc24a0081ce2856f330504ed",
  "sig": "0f928bf7e3abe1d4a12301153d5ae8e16f901b918d87226a0ddbfda1ff43f79d6f105d49af2bca74961ce04a874563720b3e735aec0da2a8b0b1fd902f7b642e",
  "tags": []
}
```
```json
{
    "pubkey": "63fe6318dc58583cfe16810f86dd09e18bfd76aabc24a0081ce2856f330504ed",
    "kind": 1063,
    "created_at": 1680687574,
    "content": "Screenshot 2023-04-04 at 12.37.40.png",
    "tags": [
        [
            "r",
            "http://localhost:8001/d/TUicS3SHUDufVNdp5pMSoK.png"
        ],
        [
            "hash",
            "66ea4d5d508100db89844cc50b90b3b27f85e1f3cc928cfd492b51b197f1f1fb",
            "sha256"
        ],
        [
            "type",
            "image/png"
        ]
    ],
    "id": "b31a3fb784049267c10815b3924980576b7d6d54fa1a95f794b6db64077a42c3",
    "sig": "8fd342c48734f517f371d723ca19155eab0eb082ece77edc235aee768a9f63362e0d08f5d4e5ef9687f8550032a7bab37973bfe3fd0fc1ceeb59e0b77fe7adb6"
}
```

melvincarvalho:
Just FYI: in [nosdav](https://github.com/nosdav/spec) I've settled on the "u" tag for a URI.  If you wanted to be consistent with that, you may want to consider "u" instead of "r", but not a huge deal.

Incidentally, sharing of file (ie N or more users/groups access to a file) is planned for nosdav, but not necessarily "file sharing".

Looking fwd to seeing how this develops!

v0l:
With magnet link on a real file:
```json
{
  "content": "Screenshot 2023-04-04 at 12.37.40.png",
  "created_at": 1680689710,
  "id": "a9a7a14bc00aa27e8490119f0579af3de810abc88c4a4627e844aba0baad3711",
  "kind": 1063,
  "pubkey": "63fe6318dc58583cfe16810f86dd09e18bfd76aabc24a0081ce2856f330504ed",
  "sig": "6910d866e1d9fa39e5d328c2ac757fe7e48d39489c8c98f226c806094f553fc1bafea857616a08dc4ff4830353f2d6d9dcb3394c9f8944e7784b867fdc38640f",
  "tags": [
    [
      "u",
      "https://void.cat/d/BZtHvLUyHQNMTiYaaiGHvR.webp"
    ],
    [
      "hash",
      "472a56a29b9dabbff5aa6d33e10b09ea819ff0a14acc672c0b60bfc263d6f059",
      "sha256"
    ],
    [
      "type",
      "image/webp"
    ],
    [
      "u",
      "magnet:?xt=urn:btih:8725e3dee476e3b16180f21d99d9f19d4ec36fbb&dn=Screenshot 2023-04-04 at 12.37.40.png&tr=wss%3A%2F%2Ftracker.btorrent.xyz&tr=wss%3A%2F%2Ftracker.openwebtorrent.com&tr=udp%3A%2F%2Ftracker.opentrackr.org%3A1337%2Fannounce&tr=udp%3A%2F%2Ftracker.openbittorrent.com%3A6969%2Fannounce&tr=http%3A%2F%2Ftracker.openbittorrent.com%3A80%2Fannounce&xs=https%3A%2F%2Fvoid.cat%3A443%2Fd%2FBZtHvLUyHQNMTiYaaiGHvR.torrent"
    ]
  ]
}
```

melvincarvalho:
Small nit FYI: doing a hash like this is fine, but there's a couple of ways of combining both the hash and the hash aglorithm together"

- [https://www.rfc-editor.org/rfc/rfc6920.html](RFC 6920: Naming Things with Hashes](https://www.rfc-editor.org/rfc/rfc6920.html)
- multihash: [example](https://cid.ipfs.tech/#QmXv7TpGgdhyjp7L1QkSFNjdq9zwoG1YHi4nykYt16Ve1r) which is used in ipfs CIDs

Normally there is a registry of hash algorithms kept somewhere for these things

I'll be working on a multihash uri scheme later in the year, consensus is that it will look like mh:

94.md
```diff
@@ -0,0 +1,49 @@
+NIP-94 - File Header
+======
+`draft` `optional` `author:frbitten` 
+
+The purpose of this NIP is to allow an organization and classification of shared files. So that relays can filter and organize in any way that is of interest.
```
vitorpamplona, line 5: immutable shared files?

94.md
```diff
@@ -0,0 +1,49 @@
...
+Also the goal is to create a base on the protocol for this bountsr "Filesharing App" (https://bountsr.org/p2p-filesharing/) to be implemented.
+
+Nostr event
+------------------
+This NIP specifies the use of the `1063` event type (parameterized replaceable event), having in `content` a description of the file content, and a list of tags described below:
```
vitorpamplona, line 10: It's not a "parameterized replaceable event" right?

frbitten: Yes. According to some suggestions, being a replaceable event would lead to problems and difficulties in keeping relationships up to date. And if it is a new URI it is a new file that must be used a new event.

fiatjaf:
Can we standardize the encryption mechanism to be always the same single one and use only that always until further notice?

94.md
fiatjaf: I think the best way would be to replace this with two tags: `h`, which contains the hex-encoded lowercase sha256 hash of the file, for search purposes (our own little IPFS), and `blurhash` in a separate tag, as it has just cosmetic purposes.

94.md
```diff
@@ -0,0 +1,49 @@
...
+
+The purpose of this NIP is to allow an organization and classification of shared files. So that relays can filter and organize in any way that is of interest.
+Also the goal is to create a base on the protocol for this bountsr "Filesharing App" (https://bountsr.org/p2p-filesharing/) to be implemented.
+
+Nostr event
+------------------
+This NIP specifies the use of the `1063` event type, having in `content` a description of the file content, and a list of tags described below:
+* `r` the url to download the file
+* `type` a string indicating the data type of the file. The MIME types format must be used (https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types)
+* `decrypt` (optional) If the file is encrypted, you must indicate in the first value the algorithm used to encrypt the file and in the second value the parameters that allow the decryption of the file
```
fiatjaf, line 13:
```suggestion
* `xchacha20` (optional) If the file is encrypted using XChaCha20, you must indicate in the first value the encryption key and in the second value the nonce, as hex
```

94.md
```diff
@@ -0,0 +1,49 @@
...
+
+```json
+{
+  "id": <32-bytes lowercase hex-encoded sha256 of the the serialized event data>,
+  "pubkey": <32-bytes lowercase hex-encoded public key of the event creator>,
+  "created_at": <unix timestamp in seconds>,
+  "kind": 1063,
+  "tags": [
+    ["r",<string with URI of file>],
+    ["decrypt",<algorithm>,<Decryption Params>],
```
fiatjaf, line 24:
```suggestion
    ["xchacha20", <key>, <nonce>],
```

frbitten:
> Can we standardize the encryption mechanism to be always the same single one and use only that always until further notice?

I don't see any problem.

94.md
fiatjaf: Why not make it `h` so it is searchable? I think that, since users can write anything in the `sha256` field and not necessarily the correct hash, we will need relays to volunteer support for this extra requirement, if they want, and block events that have the wrong hashes, otherwise the system will not work.

If relays are already going through this tedious and intensive work of downloading the files and validating their hashes, they may as well support queries for the filter `#sha256`.

94.md
v0l: Are there decent libs for `xchacha20` ? void.cat uses AES-GCM

fiatjaf: I'm happy with AES-GCM. I just heard XChaCha20 is "better". I just think we need to decide on one.

v0l: XChaCha20 is much newer so it's a bit harder to find decent libs, while AES ubiquitous

jb55:
> Can we standardize the encryption mechanism to be always the same single one and use only that always until further notice?

:+1: having one way of doing things (initially) is the nostr way.

fiatjaf:
Looks good to me.

94.md
```diff
@@ -0,0 +1,49 @@
...
+`draft` `optional` `author:frbitten` 
+
+The purpose of this NIP is to allow an organization and classification of shared files. So that relays can filter and organize in any way that is of interest.
+Also the goal is to create a base on the protocol for this bountsr "Filesharing App" (https://bountsr.org/p2p-filesharing/) to be implemented.
+
+Nostr event
+------------------
+This NIP specifies the use of the `1063` event type, having in `content` a description of the file content, and a list of tags described below:
+* `r` the url to download the file
+* `type` a string indicating the data type of the file. The MIME types format must be used (https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types)
```
staab, line 12: It might be nice if this was searchable. I don't want to load exe files, but I do want to load pngs.

94.md
v0l: `["aes-256-gcm", "<key>", "<iv>", "<ts>"]`

Where `key` & `iv` are hex, `ts` is the Tagsize in bits

v0l: https://github.com/bitwiseshiftleft/sjcl/blob/85caa53c281eeeb502310013312c775d35fe0867/core/gcm.js#L17-L24

v0l: Or you could specify the tagsize is 128 its and just leave it out of the tag

lovvtide:
@frbitten @v0l @fiatjaf There's just one thing - we agreed in the other thread that the `size` tag should be required and not optional, right? See https://github.com/nostr-protocol/nips/pull/417#issuecomment-1500951502

Otherwise looks good to me!

arthurfranca:
> r is already used elsewhere as "reference" with url value.
The reason to make it searchable is that I imagined a client may want to have an user media panel that the user would fill by using a context menu pop-up over images on notes (like "add to memes collection" for instance).
For that, it would detect used image url on kind 1 note content and create a corresponding kind 1063 event. Making 1063 url tag searchable will allow the client to know if an event with that url from the same pubkey already exists.
It would be an alternative to searching by hash.
Suggested change
	* `url` the url to download the file
	* `r` the url to download the file, without trailing slash

I've suggested using `r` tag and @frbitten had accepeted it but now with NIP-81 merge it got reverted. I still think it would be useful for the use case i mentioned above. For images it may be ok to hash it before searching, but for videos maybe not. What do you think?

frbitten:
> @frbitten @v0l @fiatjaf There's just one thing - we agreed in the other thread that the `size` tag should be required and not optional, right? See [#417 (comment)](https://github.com/nostr-protocol/nips/pull/417#issuecomment-1500951502)
> 
> Edit: One other thing - wouldn't it be a good idea to standardize an (optional) tag for the file name? I suggested having an optional `name` tag, and @frbitten I remember you said that people could maybe use the NIP-14 `subject` tag. I think many apps are going to need files to have a name and it would be a good idea to not leave that ambiguous.
> 
> Otherwise looks good to me!

The problem with forcing to inform a size that the tendency is to have wrong values in the tag. Or force customers to download the file to check the size.

Because in many cases the file will already be in the URL, torrent or magnet and whoever wants to publish it in NOSTR will not know the size.

I'll include the name tag to avoid confusion.

frbitten:
> > r is already used elsewhere as "reference" with url value.
> > The reason to make it searchable is that I imagined a client may want to have an user media panel that the user would fill by using a context menu pop-up over images on notes (like "add to memes collection" for instance).
> > For that, it would detect used image url on kind 1 note content and create a corresponding kind 1063 event. Making 1063 url tag searchable will allow the client to know if an event with that url from the same pubkey already exists.
> > It would be an alternative to searching by hash.
> > Suggested change
> > * `url` the url to download the file
> > * `r` the url to download the file, without trailing slash
> 
> I've suggested using `r` tag and @frbitten had accepeted it but now with NIP-81 merge it got reverted. I still think it would be useful for the use case i mentioned above. For images it may be ok to hash it before searching, but for videos maybe not. What do you think?

@fiatjaf  who suggested keeping as url. I don't know if it's a good idea to abuse indexable tags for the long term of NOSTR. But to serve everyone I would change "url" to "r" . Let's wait a few days to see if anyone objects to this.

arthurfranca:
@fiatjaf I mentioned the use case for the indexable url above. It would be useful to search by url to know if the item already exists before trying to add an image/video to a personal "gallery" of NIP-94 events. It could instead be done by searching by hash but I wonder if it would be a problem for videos (would have to download it completely to hash it).

v0l:
LGTM

cryptoquick:
Would it be okay to indicate what type of hash is being used, perhaps with a prefix like `sha256:` for the `hash` field? This feature would be really useful for us to implement Nostr support for carbonado.io, but we use blake3 because it's much faster. Technically sha256 hw acceleration is almost as fast as blake3, but that's really only for datacenter server CPUs like EPYC; sha2 acceleration extensions are not widespread like AES is.

Then, if a client doesn't support that hash alg, hash verification can be skipped. If a reasonably complete set of hash algs are supported by the client, then anything using an unsupported alg can be ignored.

Does this all make sense?

lovvtide:
@cryptoquick So this was my original idea as described in NIP-84, that there could be multiple `x` tags with different hashes and a marker of some kind to differentiate the kind of hash, but as I recall we decided instead that it would be better to use a different tag for each new kind of hash, i.e. to use tags to differentiate usage instead of markers. As is stands now we're using `x` for `sha256` and `i` for the torrent `infohash`. For your use case, the simplest thing to do would be to just use another tag for `blake3`

fiatjaf:
> Would it be okay to indicate what type of hash is being used, perhaps with a prefix like sha256: for the hash field? This feature would be really useful for us to implement Nostr support for carbonado.io, but we use blake3 because it's much faster. Technically sha256 hw acceleration is almost as fast as blake3, but that's really only for datacenter server CPUs like EPYC; sha2 acceleration extensions are not widespread like AES is.
> 
> Then, if a client doesn't support that hash alg, hash verification can be skipped. If a reasonably complete set of hash algs are supported by the client, then anything using an unsupported alg can be ignored.
> 
> Does this all make sense?

No, this defeats the entire purpose of even having a hash field. The only two possible outcomes of having multiple hashes are

1. No one implements anything and hashes are never checked
2. Everybody has to implement all possible hashing algorithms, which increases the burden on everybody

If blake3 is so much better then everybody should agree on using blake3 and not sha256, but since everybody is already doing sha256 all the time on all these events, sha256 is already available in all Nostr implementations and so on it will be hard to change this now.

fiatjaf:
I'm looking at https://github.com/diba-io/carbonado now and it's a very interesting project, but I don't see how it could be fully integrated into Nostr at all ever, so I don't see the point in trying to achieve compatibility just with the hashing algorithm.

s3x-jay:
I like the general idea of this NIP. Unlike actually storing the files on relays, it doesn't make the problems with CSAM and pirated content any worse than they are now since it's basically just relating meta data about the file. 

One suggestion - mention the [NIP-36](https://github.com/nostr-protocol/nips/blob/master/36.md) `content-warning` tag. Technically NIP-36 content warnings can be added to any event, but they're particularly relevant here. 

And the proposed "[NIP-69](https://github.com/nostr-protocol/nips/pull/457)" defines a "vocabulary" that can be used with NIP-36 (and NIP-56).

PR #339, opened 2023-03-09
NIP-04: Minor wording change

jeffthibault:
I think "read" is misleading because the content is encrypted.

### DIFF
--- a/04.md
+++ b/04.md
@@ -46,4 +46,4 @@ let event = {
 
 ## Security Warning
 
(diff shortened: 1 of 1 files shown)


1440000bytes (approved):
ACK https://github.com/nostr-protocol/nips/pull/339/commits/1bf4866d0e707f4773bb1b109e15854e2de09a28

According to [NIP 4](https://github.com/nostr-protocol/nips/blob/master/04.md) content MUST be equal to the base64-encoded, aes-256-cbc encrypted string of anything a user wants to write, encrypted using a shared cipher generated by combining the recipient's public-key with the sender's private-key.

PR #345, opened 2023-03-10
NIP-95 - Storage and Shared File

frbitten:
Initially this NIP was together with the PR of NIP-94 but I thought it better to separate it because it will require a greater discussion and it is not necessary to link the approval of one NIP with the other.

I suggest you read #337 before starting the discussion here.

@Egge7 we can continue the discussion here.

### DIFF
--- a/95.md
+++ b/95.md
@@ -12,7 +12,6 @@ This NIP specifies the use of the `30064` event type (parameterized replaceable 
 * `d` containing the name of the shared file to allow it to be overwritten in the future.
 * `type` a string indicating the data type of the file. The MIME types format must be used (https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types)
 * `decrypt` (optional) If the file is encrypted, you must indicate in the first value the algorithm used to encrypt the file and in the second value the parameters that allow the decryption of the file
-* `format` ("Base64" or "BSON") Defines how the data is stored in the content. Having the option of being in base64 format or in BSON format.
 
 This event must not be returned in generic searches. It should only be returned if specifically requested by your ID. This avoids bandwidth consumption and unnecessary overload on the relay and client.
 
@@ -29,7 +28,6 @@ The `NIP-94` can be used to broadcast the ID of that event, with the ` and ` tag
     ["decrypt",<algorithm>,<Decryption Params>],
     ["p", <32-bytes hex of a pubkey>, <recommended relay URL>],
     ["hash",< SHA256 hexencoded string of the raw data>],
-    ["format",<"Base64" or "BSON">]
   ],
   "content": <string with base64 data>,
   "sig": <64-bytes hex of the signature of the sha256 hash of the serialized event data, which is the same as the "id" field>
--- a/95.md
+++ b/95.md
@@ -8,7 +8,7 @@ The intention is to combine the ease of NOSTR, with many clients being developed
 
 Nostr event
 ------------------
-This NIP specifies the use of the `1064` event type (parameterized replaceable event), having in `content` the binary data that you want to be stored in Base64 format.
+This NIP specifies the use of the `1064` event type, having in `content` the binary data that you want to be stored in Base64 format.
 * `type` a string indicating the data type of the file. The MIME types format must be used (https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types)
 * `decrypt` (optional) If the file is encrypted, you must indicate in the first value the algorithm used to encrypt the file and in the second value the parameters that allow the decryption of the file
 
@@ -55,7 +55,7 @@ The relay can allow access to this data via URL, having its own URL pattern for 
 
 Client Behavior
 ---------------
-The client can use this event for any type of storage it wants. The functionality of uploading an image directly through the client's interface and including it in the note that is being created is a very useful facility for the end user. In this case, 3 events would be created. Kind 30064 with the file converted to base64, Kind 30063 which is the header of this file for dissemination and Kind 1 with the post and the image.
+The client can use this event for any type of storage it wants. The functionality of uploading an image directly through the client's interface and including it in the note that is being created is a very useful facility for the end user. In this case, 3 events would be created. Kind 1064 with the file converted to base64, Kind 1063 which is the header of this file for dissemination and Kind 1 with the post and the image.
 
 Also, when used in conjunction with `NIP-94`, it allows the app to create image galleries (memes, animations) that can be reused countless times in different notes. As it exists in whatsapp, tegegran, etc.
 
--- a/95.md
+++ b/95.md
@@ -1,8 +1,10 @@
-NIP-95 Base64 File
+NIP-95 - Storage and Shared File
 ======
 `draft` `optional` `author:frbitten` 
 
 This NIP allows users to send binary data to be stored in the relay. This data can be small pieces of information or even large files.
+
+The intention is to combine the ease of NOSTR, with many clients being developed and a common communication protocol between all. To enable decentralized file sharing. NOSTR communication must work together with other protocols to get the best out of each technology. For lay users to be able to publish their photo, file, audio on several servers in a simple and easy way.
 
 Nostr event
 ------------------
@@ -10,6 +12,7 @@ This NIP specifies the use of the `30064` event type (parameterized replaceable 
 * `d` containing the name of the shared file to allow it to be overwritten in the future.
 * `type` a string indicating the data type of the file. The MIME types format must be used (https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types)
 * `decrypt` (optional) If the file is encrypted, you must indicate in the first value the algorithm used to encrypt the file and in the second value the parameters that allow the decryption of the file
+* `format` ("Base64" or "BSON") Defines how the data is stored in the content. Having the option of being in base64 format or in BSON format.
 
 This event must not be returned in generic searches. It should only be returned if specifically requested by your ID. This avoids bandwidth consumption and unnecessary overload on the relay and client.
 
@@ -25,7 +28,8 @@ The `NIP-94` can be used to broadcast the ID of that event, with the ` and ` tag
     ["d", <string with name of file>],
     ["decrypt",<algorithm>,<Decryption Params>],
     ["p", <32-bytes hex of a pubkey>, <recommended relay URL>],
-    ["hash",< SHA256 hexencoded string of the raw data>]
+    ["hash",< SHA256 hexencoded string of the raw data>],
+    ["format",<"Base64" or "BSON">]
   ],
   "content": <string with base64 data>,
   "sig": <64-bytes hex of the signature of the sha256 hash of the serialized event data, which is the same as the "id" field>
@@ -54,3 +58,9 @@ Suggested Use Cases
 -------------------
 * Provide file storage service that is quickly integrable with nostr clients.
 * Create an application similar to pinterest without the need for other protocols.
+
+References
+---------------
+* https://www.mongodb.com/developer/products/mongodb/storing-large-objects-and-files/
+* https://www.mongodb.com/docs/manual/core/gridfs/
+* https://www.linkedin.com/pulse/do-store-files-your-database-mongodbs-gridfs-thinks-otherwise-gordon/
--- a/95.md
+++ b/95.md
@@ -14,8 +14,6 @@ This NIP specifies the use of the `1064` event type (parameterized replaceable e
 
 This event must not be returned in generic searches. It should only be returned if specifically requested by your ID. This avoids bandwidth consumption and unnecessary overload on the relay and client.
 
-The `NIP-94` can be used to broadcast the ID of that event, with the ` and ` tag referencing the event id, without having to send the full event which can be a lot of data.
-
 ```json
 {
   "id": <32-bytes lowercase hex-encoded sha256 of the the serialized event data>,
@@ -30,6 +28,22 @@ The `NIP-94` can be used to broadcast the ID of that event, with the ` and ` tag
   "sig": <64-bytes hex of the signature of the sha256 hash of the serialized event data, which is the same as the "id" field>
 }
 ```
+
+Another defined event is the `1065` which is used as a header for the data contained in an event 1064. This way the data can be disclosed without overloading the communications when sending a large amount of data.
+
+```json
+{
+  "id": <32-bytes lowercase hex-encoded sha256 of the the serialized event data>,
+  "pubkey": <32-bytes lowercase hex-encoded public key of the event creator>,
+  "created_at": <unix timestamp in seconds>,
+  "kind": 1065,
+  "tags": [
+    ["e",<event id 1064 you want to disclose>],
+  ],
+  "content": <description>,
+  "sig": <64-bytes hex of the signature of the sha256 hash of the serialized event data, which is the same as the "id" field>
+}
+````
 
 Relay Behavior
 ---------------
--- a/95.md
+++ b/95.md
@@ -8,8 +8,7 @@ The intention is to combine the ease of NOSTR, with many clients being developed
 
 Nostr event
 ------------------
-This NIP specifies the use of the `30064` event type (parameterized replaceable event), having in `content` the binary data that you want to be stored in Base64 format.
-* `d` containing the name of the shared file to allow it to be overwritten in the future.
+This NIP specifies the use of the `1064` event type (parameterized replaceable event), having in `content` the binary data that you want to be stored in Base64 format.
 * `type` a string indicating the data type of the file. The MIME types format must be used (https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types)
 * `decrypt` (optional) If the file is encrypted, you must indicate in the first value the algorithm used to encrypt the file and in the second value the parameters that allow the decryption of the file
 
@@ -22,12 +21,10 @@ The `NIP-94` can be used to broadcast the ID of that event, with the ` and ` tag
   "id": <32-bytes lowercase hex-encoded sha256 of the the serialized event data>,
   "pubkey": <32-bytes lowercase hex-encoded public key of the event creator>,
   "created_at": <unix timestamp in seconds>,
-  "kind": 30064,
+  "kind": 1064,
   "tags": [
-    ["d", <string with name of file>],
     ["decrypt",<algorithm>,<Decryption Params>],
     ["p", <32-bytes hex of a pubkey>, <recommended relay URL>],
-    ["hash",< SHA256 hexencoded string of the raw data>],
   ],
   "content": <string with base64 data>,
   "sig": <64-bytes hex of the signature of the sha256 hash of the serialized event data, which is the same as the "id" field>
--- a/95.md
+++ b/95.md
@@ -29,21 +29,7 @@ This event must not be returned in generic searches. It should only be returned 
 }
 ```
 
-Another defined event is the `1065` which is used as a header for the data contained in an event 1064. This way the data can be disclosed without overloading the communications when sending a large amount of data.
-
-```json
-{
-  "id": <32-bytes lowercase hex-encoded sha256 of the the serialized event data>,
-  "pubkey": <32-bytes lowercase hex-encoded public key of the event creator>,
-  "created_at": <unix timestamp in seconds>,
-  "kind": 1065,
-  "tags": [
-    ["e",<event id 1064 you want to disclose>],
-  ],
-  "content": <description>,
-  "sig": <64-bytes hex of the signature of the sha256 hash of the serialized event data, which is the same as the "id" field>
-}
-````
+The file description is defined by [NIP-94](https://github.com/nostr-protocol/nips/blob/master/94.md)
 
 Relay Behavior
 ---------------
--- a/95.md
+++ b/95.md
@@ -29,7 +29,36 @@ This event must not be returned in generic searches. It should only be returned 
 }
 ```
 
-The file description is defined by [NIP-94](https://github.com/nostr-protocol/nips/blob/master/94.md)
+Another defined event is the `1065` which is used as a header for the data contained in an event 1064. This way the data can be disclosed without overloading the communications when sending a large amount of data.
+
+```json
+{
+  "id": <32-bytes lowercase hex-encoded sha256 of the the serialized event data>,
+  "pubkey": <32-bytes lowercase hex-encoded public key of the event creator>,
+  "created_at": <unix timestamp in seconds>,
+  "kind": 1065,
+  "tags": [
+    ["e",<event id 1064 you want to disclose>],
+    ["m", <MIME type>],
+    ["x",<Hash SHA-256>],
+    ["size", <size of data in bytes>],
+    ["dim", <size of data in pixels>],
+    ["thumb", <event id that contains the data for thumb>],
+    ["summary", <excerpt>],
+    ["alt", <description>]
+  ],
+  "content": <description>,
+  "sig": <64-bytes hex of the signature of the sha256 hash of the serialized event data, which is the same as the "id" field>
+}
+````
+* `e` event id 1064 you want to disclose
+* `m` a string indicating the data type of the file. The [MIME types](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types) format must be used, and they should be lowercase.
+* `x` containing the SHA-256 hexencoded string of data.
+* `size` (optional) size of data in bytes
+* `dim` (optional) size of data in pixels in the form `<width>x<height>`
+* `thumb` (optional) event id that contains the data for thumb, which can be an event 1063 or 1064
+* `summary` (optional) text excerpt
+* `alt` (optional) description for accessibility
 
 Relay Behavior
 ---------------
@@ -37,13 +66,11 @@ Relays that use a relational database may have more difficulty implementing this
 
 Another solution is for Relays that want to implement this functionality and use a No-SQL database with mongodb that already supports large documents without harming performance.
 
-The relay can allow access to this data via URL, having its own URL pattern for these cases. And if you receive a `NIP-94` referring to a `NIP-95` you can include the URL in the proper `NIP-94` tag
-
 Client Behavior
 ---------------
-The client can use this event for any type of storage it wants. The functionality of uploading an image directly through the client's interface and including it in the note that is being created is a very useful facility for the end user. In this case, 3 events would be created. Kind 1064 with the file converted to base64, Kind 1063 which is the header of this file for dissemination and Kind 1 with the post and the image.
+The client can use this event for any type of storage it wants. The functionality of uploading an image directly through the client's interface and including it in the note that is being created is a very useful facility for the end user. In this case, 3 events would be created. Kind 1064 with the file converted to base64, Kind 1065 which is the header of this file for dissemination and Kind 1 with the post and the image.
 
-Also, when used in conjunction with `NIP-94`, it allows the app to create image galleries (memes, animations) that can be reused countless times in different notes. As it exists in whatsapp, tegegran, etc.
+Also, it allows the app to create image galleries (memes, animations) that can be reused countless times in different notes. As it exists in whatsapp, tegegran, etc.
 
 Example: <https://ibb.co/Fnj5TMg>. 
 
--- a/95.md
+++ b/95.md
@@ -1,0 +1,56 @@
+NIP-95 Base64 File
+======
+`draft` `optional` `author:frbitten` 
+
+This NIP allows users to send binary data to be stored in the relay. This data can be small pieces of information or even large files.
+
+Nostr event
+------------------
+This NIP specifies the use of the `30064` event type (parameterized replaceable event), having in `content` the binary data that you want to be stored in Base64 format.
+* `d` containing the name of the shared file to allow it to be overwritten in the future.
+* `type` a string indicating the data type of the file. The MIME types format must be used (https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types)
+* `decrypt` (optional) If the file is encrypted, you must indicate in the first value the algorithm used to encrypt the file and in the second value the parameters that allow the decryption of the file
+
+This event must not be returned in generic searches. It should only be returned if specifically requested by your ID. This avoids bandwidth consumption and unnecessary overload on the relay and client.
+
+The `NIP-94` can be used to broadcast the ID of that event, with the ` and ` tag referencing the event id, without having to send the full event which can be a lot of data.
+
+```json
+{
+  "id": <32-bytes lowercase hex-encoded sha256 of the the serialized event data>,
+  "pubkey": <32-bytes lowercase hex-encoded public key of the event creator>,
+  "created_at": <unix timestamp in seconds>,
+  "kind": 30064,
+  "tags": [
+    ["d", <string with name of file>],
+    ["decrypt",<algorithm>,<Decryption Params>],
+    ["p", <32-bytes hex of a pubkey>, <recommended relay URL>],
+    ["hash",< SHA256 hexencoded string of the raw data>]
+  ],
+  "content": <string with base64 data>,
+  "sig": <64-bytes hex of the signature of the sha256 hash of the serialized event data, which is the same as the "id" field>
+}
+```
+
+Relay Behavior
+---------------
+Relays that use a relational database may have more difficulty implementing this NIP, but a possible solution is for this NIP not to be recorded in the database, but on disk, the file name being the event id. So it can be easily found and searched. And because it is not in the database, it does not interfere with the indexing of common events.
+
+Another solution is for Relays that want to implement this functionality and use a No-SQL database with mongodb that already supports large documents without harming performance.
+
+The relay can allow access to this data via URL, having its own URL pattern for these cases. And if you receive a `NIP-94` referring to a `NIP-95` you can include the URL in the proper `NIP-94` tag
+
+Client Behavior
+---------------
+The client can use this event for any type of storage it wants. The functionality of uploading an image directly through the client's interface and including it in the note that is being created is a very useful facility for the end user. In this case, 3 events would be created. Kind 30064 with the file converted to base64, Kind 30063 which is the header of this file for dissemination and Kind 1 with the post and the image.
+
+Also, when used in conjunction with `NIP-94`, it allows the app to create image galleries (memes, animations) that can be reused countless times in different notes. As it exists in whatsapp, tegegran, etc.
+
+Example: <https://ibb.co/Fnj5TMg>. 
+
+To do this, just select the image from the gallery (events NIP-94) and include the URL of the selected image
+
+Suggested Use Cases
+-------------------
+* Provide file storage service that is quickly integrable with nostr clients.
+* Create an application similar to pinterest without the need for other protocols.


frbitten:
@Egge7 
Is BSON not supported by NOSTR just because no relay or client has implemented support for it? Or is it due to some BSON data parser issue where it is not expecting BSON to exist in the JSON string?
I've never worked with BSON and I don't know what implications its use in the event would have.

Egge7:
> @Egge7 Is BSON not supported by NOSTR just because no relay or client has implemented support for it? Or is it due to some BSON data parser issue where it is not expecting BSON to exist in the JSON string? I've never worked with BSON and I don't know what implications its use in the event would have.

BSON can not exists inside a JSON as JSON is text and BSON is binary. BSON is a binary representation of a JSON object, which enables it to hold raw data keys (this makes it great for storage as you do not lose space due to encoding). Websockets can transmit blobs as well as text, but everything on nostr is designed to be text ,which is why I proposed a sub-protocol for binary transmissions.

vitorpamplona:
Is there a need for the `hash` tag now that the contents are in the json itself?

nschairer:
I don't think so, since the id should contain the data and the signature will verify nothing changed. 

It would be pretty neat for large files to chunk them out across relays though, and then you could reduce load off one relay and query many for different chunks of the file. In that case you would probably want the hash for verification when you reassemble everything.

vitorpamplona:
Yep, breaking the file down into multiple events could be interesting to connection hiccups requiring apps to restart from scratch. 

It could be very simple: 

```json
{
  "id": <32-bytes lowercase hex-encoded sha256 of the the serialized event data>,
  "pubkey": <32-bytes lowercase hex-encoded public key of the event creator>,
  "kind": 30065,
  "tags": [
    ["d", <string with name of file>],
    ["decrypt",<algorithm>,<Decryption Params>],
    ["p", <32-bytes hex of a pubkey>, <recommended relay URL>],
    ["hash",< SHA256 hexencoded string of the complete raw data>],
    ["format",<"Base64" or "BSON">], 
    ["s", <part1>, <relay>],
    ["s", <part2>, <relay>],
    ["s", <part...>, <relay>],
    ["s", <partn>, <relay>],
  ],
  "content": "",
  "sig": <64-bytes hex of the signature of the sha256 hash of the serialized event data, which is the same as the "id" field>
}
```

vitorpamplona:
If you use gzip to compress the entire JSON automatically when connecting with relays, Base64 is not that bad: https://lemire.me/blog/2019/01/30/what-is-the-space-overhead-of-base64-encoding/

nschairer:
Would be interesting to see performance comparison between @Egge7 's protocol when performing encoding/decoding + reassembling + hash validation client / relay side

Egge7:
> Would be interesting to see performance comparison between @Egge7 's protocol and just keeping it in nostr with base64 chunks when performing encoding/decoding + reassembling + hash validation client / relay side

As I said I only started spec'ing and never actually implemented anything. You can definitely do so and then do some benchmarking. However even without benchmarking we know that base64 will take up about 33% more space vs. raw binary data because it uses 6bits per byte.

frbitten:
About the "hash" tag really makes no sense. It ended up coming in copy/paste on NIP94. I will remove it.

I also changed it to be a regular event as being replaceable could cause more trouble than ease. With that the kind changed to 1064

vitorpamplona:
The issue with that idea is that every part cites every other part. They become heavier objects to parse. Instead, my proposal was to do just one header kind that cites all parts, has the hash, decrypting info, etc. Each part only has its contents and cites the header.

frbitten:
> The issue with that idea is that every part cites every other part. They become heavier objects to parse. Instead, my proposal was to do just one header kind that cites all parts, has the hash, decrypting info, etc. Each part only has its contents and cites the header.

That's why I suggested the NIP-94 (#337 ) which is a file sharing header.

frbitten:
It can be used in the same way.

As I put in the description, the NIP-94 must be used to disclose the NIP-95 file referencing the event. Because the NIP-95 is not returned in broad searches and only if the event is effectively requested.

I can reference the event by the e tag. And use multiple tags and for multiple parts.
The URL tag can also be used in this case if the relay provides the option to download NIP-95 event data via HTTP.

You don't need to create another kind for this. It is perfectly possible to use the NIP-94

vitorpamplona:
I would separate them. It's too confusing to make them the same event and there might be reasons to separate a search for files in url from a search for files inside relays. Clients might want to support one but not the other, etc. 

There is no shortage of integers for each kind.

frbitten:
> Eu os separaria. É muito confuso torná-los o mesmo evento e pode haver motivos para separar uma pesquisa de arquivos em url de uma pesquisa de arquivos dentro de retransmissões. Os clientes podem querer oferecer suporte a um, mas não ao outro, etc.
> 
> Não faltam números inteiros para cada tipo.

I see no problem in creating another specific event. But the operation would be the same as the NIP-94 and would only remove the url tag.

I'll put the definition of this new event here in the description.

vitorpamplona:
Hummmm NIP 94 shouldnt describe file parts. This new one should describe how to reassemble and test hashes.

arthurfranca:
> [...] a possible solution is for this NIP not to be recorded in the database, but on disk, the file name being the event id [...]

I think I get @frbitten idea. The relay is expected to save the file data in one complete disk file (or blob/bson at db). Then, it will serve it as an event (also, he says it could optionally serve from an https url, but this wouldn't be nostr).

For this to work, the relay would recreate the event **dynamically** (with same created_at and all other keys), upon request, before serving it to a client. So, maybe the one who uploaded won't be the owner of the NIP-95 event (or else, relay wouldn't be able to recreate the event with same id). The private key would be one the relay owns.

**Better yet, the private key used to sign the event when recreating it dynamically should be a shared one, so that other relays could use it to sign and recreate the event themselves – I think the NIP-94 event id could be used as a private key.**

People later suggested it would be better to split file in chunks. Now it won't be only one disk file/db record. 

NIP-94 could have (when referencing a NIP-95 event instead of a simple url tag) tags pointing to the parts like @vitorpamplona said:
```js
{
  /*...*/,
  tags: [
    ["e", /* NIP-95 event id */, /* recommended relay url */], // first part
    ["e", /* NIP-95 event id */, /* recommended relay url */], // second part
    ["r", /* other relay to where the parts were also submitted to */],
    ["r", /* other relay to where the parts were also submitted to */],
    ["r", /* other relay to where the parts were also submitted to */]
  ]
}
```

While a NIP-95 event would be
```js
{
  id: 'an id',
  pubkey: 'a pubkey',
  tags: [
    ["e", /* NIP-94 event id */, /* recommended relay url */] // file header,
  ],
  content: /* string with sliced base64 data */,
  created_at: 11111111
}
```

So NIP-95 wouldn't be stored as an event. It will be a file with created_at as file metadata, compressed base64 data and filename taken from the event.id. I think the e tag can't be a filesystem metadata like created_at can, so maybe it should be inside an auxiliary event-id.json. Well so better to put all metadata inside the json file like info about how the relay stored the main file, like the file extension (e.g. .gzip), created_at, stringified tags array.

🤔 All this to save [like 28%](https://github.com/nostr-protocol/nips/pull/345#issuecomment-1471963100) of space in comparison to storing the uncompressed base64 event. Worth it!

vitorpamplona:
I am not sure why you would do all that. You can simply receive the event, take the base64 contents, convert it back to binary, save it on disk with the event ID as filename, and store the JSON without the `.content` field in the database. When a client requests it, pull the event from the database, use the Id to fetch the `.content` from the disk, recreate the base64 version, and send it signed by the original author as if nothing has happened. 

Easy.

vitorpamplona:
In fact, relays should do that for all events whose `.content` is bigger than.. say... 1MB...

frbitten:
They are all valid options. The issue is that the way to store it is up to the relay.
Another option is to use a NO-SQL database such as mongoDB which already supports large files (up to 16MBs) and could put everything in the database. And with that, it would gain the facility to replicate the database and files in several clustered instances to meet a large demand from customers.

NOSTR defines the form of communication. How the data is stored is up to the relay to define the best format. I see no problem spending a little more on the transfer for simplicity. As I mentioned above, I can't imagine anyone wanting to store gigabytes that way. The trend will be small files. The Relay can also define maximum limits.

vitorpamplona:
I see this NIP being used for profile pictures, images, and short videos inside posts. Large files, but not THAT large.

arthurfranca:
> I am not sure why you would do all that.

I was trying to figure out what @frbitten suggested when he also said "the file name being the event id. So it can be easily found and searched. And because it is not in the database, it does not interfere with the indexing of common events."

If the event isn't stored, relay wouldn't be indexing the event.id, created_at and e tag (although `e` tag could be `ee` instead just to avoid indexing if it is expected to request NIP-95 by its event id only). So it would save a bit of space.



> I am not sure why you would do all that. You can simply receive the event, take the base64 contents, convert it back to binary, save it on disk with the event ID as filename, and store the JSON without the .content field in the database. When a client requests it, pull the event from the database, use the Id to fetch the .content from the disk, recreate the base64 version, and send it signed by the original author as if nothing has happened.

Easier indeed.

frbitten:
That's exactly the idea. This event is outside of all indexing, searching, and nostr relay processing. It will only be returned if requested by its own ID. This avoids a lot of processing and sending unwanted data

frbitten:
> Events should not be used for storing data. There's the issue with base64 encoding increasing storage usage by 33%, there's the issue with being not streamable since a single WS message cannot be streamed, there's the issue of WS message size limits, ...

The increased data size is a price to pay for ease. As described at the beginning, the idea is not to store large data, but small information that is constantly used by applications or that needs to be shared easily. Currently there is nothing simple and practical to share small data. The WS limits are quite large and easily meet this need. Relays can define their maximum limits that are accepted.

frbitten:
> Taking a step back… Having Nostr be the source for files (rather than just having text that may link to files elsewhere) increases the chance of problems with illegal content. Do we really want to be the people providing censorship resistance for CSAM (child sexual abuse materials)? Yes, by potentially linking to problem content, we're already involved. But actually distributing and hosting the files in question is a whole other level of involvement.
> 
> NIP-94 seems like a great idea. But this one worries me. I get where there are use cases where this may be really beneficial, but are they sufficient to outweigh the problems created by unintended consequences? I'd rather see the actual hosting of large files be something separate from Nostr and just have Nostr focus on the types of things it's already showing it can do well.

It happens with the notes. Nothing prevents taking text from a book and putting it in a note. Or use the markdown PIN to spread illegal content. Or give instructions on how to buy illegal products. Sharing a link to a copyrighted image in a note is the same issue. If you think about it that way, we have to do away with NOSTR and anything decentralized and rely on centralized companies and platforms.

Most only think of the NIP-95 for sharing large files, but its biggest use is for small things for quick and simple access.

frbitten:
> This is a bad idea. WebSockets have a limited frame size, base64 is inefficient, and a lot more with using Nostr (such as having to put this data in the DB).

@Semisol I don't know which frame limit you are referring to. But in the websocket specification the limit is almost infinite (9,223,372,036,854,775,807 bytes ~= 9.22 exabytes), maybe it's some limitation of the language you're used to. The websocket is designed to break the large message into small TCP blocks and send all these blocks and reassemble the message at the destination ([https://www.rfc-editor.org/rfc/rfc6455#section-10.4](https://www.rfc-editor.org/rfc/rfc6455#section-10.4)). And even if a language defines a frame limit that limit applies to everything, I manage to hit it with a standard NOSTR note as well.

Yes base64 is inefficient, as transferring data in JSON is too. But NOSTR is not a protocol with an emphasis on efficiency, but on its simplicity and practicality.
Therefore, I do not see these as issues that impede the NIP. You may find it a bad idea and it's your right but everyone is free to implement the NIPs they think are most important in their relay. So I don't see a technical reason for a pull request rejection.

Each relay can create their limits that they are willing to work with and implement the NIPs they find useful.

bndw:
> uploading an image ... In this case, 3 events would be created. Kind 30064 with the file converted to base64, Kind 30063 which is the header of this file for dissemination and Kind 1 with the post and the image.

This seems like a lot of complexity for a traditionally very simple operation. I would suggest file uploads be handled by POSTing `multipart/form-data` in the same way almost every HTTP client does today. The form could include metadata like signatures, content hashes, pubkeys, etc. Maybe an event is generated as well.

For real world examples, this is how nostr.build and nostrimg.com work today: 
https://github.com/michaelhall923/nostr-media-spec

I would be curious your thoughts on the above spec.

frbitten:
> > uploading an image ... In this case, 3 events would be created. Kind 30064 with the file converted to base64, Kind 30063 which is the header of this file for dissemination and Kind 1 with the post and the image.
> 
> This seems like a lot of complexity for a traditionally very simple operation. I would suggest file uploads be handled by POSTing `multipart/form-data` in the same way almost every HTTP client does today. The form could include metadata like signatures, content hashes, pubkeys, etc. Maybe an event is generated as well.
> 
> For real world examples, this is how nostr.build and nostrimg.com work today: https://github.com/michaelhall923/nostr-media-spec
> 
> I would be curious your thoughts on the above spec.

I don't see why create a whole other parallel protocol that will need to be maintained and updated to do the same thing that NOSTR already does.
Everything described is solved with NIP-94 and NIP-95. Without needing a parallel protocol at all.

Yes, I can create event 1064, then 1065 and reference 1065 in a post. But I can simply send it to the relay as 1064 and use a URL to reference in the note. Or even references note 1064 via tag e.

The purpose of NIP-95 is to do the same as its alternative, to standardize the way in which binary data is sent to the relay for storing such data. The form of storage is up to the relay to decide which is the most efficient for your case. I can send it as note 1064 but then access it via URL using the note id in the url. So many people can share the same link that an unknown person sent. Without having several people upload the same image to the server creating duplicity. Because the hash has how to know if the image already exists or not on the server.

bndw:
> I don't see why create a whole other parallel protocol that will need to be maintained and updated to do the same thing that NOSTR already does.

I'm suggesting to consider accepting an HTTP POST request of form data at the relay. This is supplemental to the nostr protocol. There is no maintenance overhead.

> to standardize the way in which binary data is sent to the relay for storing such data

This is what the document I liked does, and is how the popular image hosting sites on nostr work today.

> The form of storage is up to the relay to decide which is the most efficient for your case.

This is what the document I liked does too.

> Without having several people upload the same image to the server creating duplicity. Because the hash has how to know if the image already exists or not on the server.

The document I liked also does content hashing to prevent duplication.

In summary, I think this proposed solution is taking the historically very simple operation of sending a file to a server and doing it in an overly complex way. A big appeal of nostr is the simplicity of the protocol. I'm still not sure why sending 3 events to upload an image is a better solution than simply POSTing form-data.

Is forcing everything over websockets a requirement for nostr?

vitorpamplona:
As long as the "binary" data is a signed Nostr-event JSON it should be fine having a post method. 

The goal here is to link files without a domain name (no urls) so that they can migrate from relay to relay without having to change the events citing and quoting that file. 

A kind 1 would cite an nevent to download an image. That nevent can be on any relay. Client apps will figure out how to find it. We will be free from centralizing file storage services.

frbitten:
@bndw 
To post a file, you only need one event, the 1064. The others are for publicizing or other extra search functionality. Note 1064 would be the same thing as your POST suggestion.

For your solution to have the same functionality as a NOSTR event, you must have:
- File search by uploader, tags, url, hash, etc
- Simple download option for sending to another server with the same data structure. All servers would need to have the same implementation
- Being able to publish on several servers at the same time
- Allow gateways to be used to unify information from various file servers

And so many other features that the NOSTR environment creates daily and we can't even imagine. But I can imagine a few:
- paid download
- Sale of digital content
- Emojis and Memes

NIP-95 is not something to be implemented in all relays. It will be an implementation for some specialized relays. For example, it is not possible to make a Pinterest or an Instagram based on NOSTR and trust external file servers via URL. This will require specialized relays and specialized clients. But everything will be NOSTR internally. The single most global multi-client application would be an emoji and meme server to be used by all. I don't need to save lists of URLs to respond to a thread with that animated gif. The application itself can suggest some options for emojis or memes to be used, as currently happens on Whattswap, Telegran, etc.
gifs-telegran

It becomes a decentralized and collaborative database of emojs and memes, doing this with URL becomes a big job for those who program the applications, because each application would have to have its list of URLs or someone implement a whole new API for this.

In the future no one will know what NOSTR is, the concept will remain for a few geeks, but everyone will use and benefit from applications that work in the NOSTR ecosystem.

One day I still want to see a NostrTube :)

vitorpamplona:
Amethyst implements the current version of this NIP-95. But feel free to keep refining it. I will be updating it as you folks develop it further. 

https://github.com/vitorpamplona/amethyst/releases/tag/v0.37.0

jb55:
> This is a bad idea. WebSockets have a limited frame size, base64 is inefficient, and a lot more with using Nostr (such as having to put this data in the DB).

seconded.

benthecarman (changes requested):
NACK, this is a terrible idea. Putting whole files on relays makes no sense and will cause many smaller relays to break. Encoding files in base64 will take up extremely large amounts of space.

mikedilger:
I haven't read this NIP but my opinion is that an event kind representing arbitrary binary data of arbitrary length should not contain the data but instead:
1) provide the means to decrypt it
2) provide a hash of the contents so it can be validated.
3) provide a URL to wherever the data actually is.

The reasons I am against embedding the data into nostr include those already mentioned (base64 size blowout, websocket frame size, heavy cost on relays) 

but also because relay operators may find they are being sent child pornography by some asshole spammer and can't filter it out fast enough, and that can be used as an excuse by authorities to shut down nostr relays. I don't want upstream providers to have a reasonable saleable (to normies) excuse to shut down nostr relays.  There was a PR somewhere about storing binary data externally and even obfuscating it, which I thought was a good idea.

vitorpamplona:
Lots of assumptions and little evidence in the replies.

ok300:
Excellent attack against Nostr.

- It sounds reasonable enough to bind dev resources in good-faith attempts at analyzing and critiqueing it. Which is great, because good-faith attempts at honest analysis can always be defeated if time is on your side. Starve the opposition of time and attention by placing the burden on proof on them (frame their points as "assumptions" with "little evidence"), when its actually you who should prove the change is worth it. Don't worry, devs are busy, they might try once or twice to make a point, but if they're ignored or side-tracked, most will give up.

- It opens the door to countless social and technical attacks for anyone who can (or wants to) think 2-3 steps ahead, as already detailed in comments above. All for the amazing benefit of being able to "store small images".

- The draft PR already acts as a "foot in the door" for quasi-interested and less-discerning implementations to go ahead and "just code it bro", as has happened before with the zap PR, completely ignoring 95% of critiques. Once 2-3 implementations support it, its a de facto standard, regardless of PR status.

Respect, well played!

fiatjaf:
I hate this personally and I think people shouldn't do it all, but if there are people implementing it then it should be a NIP -- even if it's only a NIP that we haters can point to and say: "no!"

frbitten:
> I hate this personally and I think people shouldn't do it all, but if there are people implementing it then it should be a NIP -- even if it's only a NIP that we haters can point to and say: "no!"

I think the discussion should remain open here. And in that case the criterion of two implementations must be followed.

Unfortunately I don't have time to develop at the speed that the community asks for answers. But one day I show something.

mikedilger:
> I hate this personally and I think people shouldn't do it all, but if there are people implementing it then it should be a NIP -- even if it's only a NIP that we haters can point to and say: "no!"

I guess I'm okay with that. Better that people know what is out there than not know.

I'm just going to leave this here though as what I think could be a good alternative to this NIP (hat tip @melvincarvalho): https://github.com/nosdav/spec

0xtrr:
NACK.

If this is merged, I at least hope that all relay implementations blacklist these kinds by default so that relay operators can enable it if they want to and have the resources to support it. I'd bet that most relay operators today doesn't have enough storage to support this NIP. I also don't think everyone want to be a psuedo-FTP server for larger files.

fiatjaf:
Is this ready to merge?

bleetube:
ACK
And I hate it.  I can see the case for siloing this behavior to a specific NIP, so I'm not opposed to seeing it merged. I'm [proactively blocking](https://github.com/bleetube/strfry-plugin/commit/d96ea5aabbbd4865cc7a87a3ddd74477e472184a) nip-95 on my free relays. I'm glad that while I vehemently disagree that this is a useful application of the nostr protocol, I will be able to opt-out.

The real risk is if people implement file storage ad-hoc in other ways that are harder to filter.

frbitten:
@fiatjaf I think the rule of having implementations in this case should be followed.  We still don't have a relay implementation for the NIP, I intend to do it but it will take some time because I have other commitments at the moment.  

For this NIP to be functional in the way described, the relays have to have a better report of supported NIPs and a better control of usage restrictions.  I imagine that if you send giant notes in text most of the relay will accept it and have storage problems, so the network is not mature to prevent attacks at the moment.  This is another advantage of NIP will force relays to pay better attention to what they are receiving and their policies for allowed events.

I also think it should be standard behavior for all relays to refuse unsupported events.  There are several events that require a specific behavior on the part of the relay and often the option to do nothing is harmful.  But I think this should be a new NIP to discuss use cases more comprehensively.

fiatjaf:
What if this NIP specified that the events must be signed by the private key `1` and have `created_at: 0`? This way we could safely skip the signatures and if multiple people published the same file they would all get the same event id.

vitorpamplona:
We could do that for 1065 (only the data) and get author and created at from 1064 (header info).

vitorpamplona:
If both files are the same (hash(data) is the event id), does it matter?

frbitten:
> If both files are the same (hash(data) is the event id), does it matter?

I did not understand the question.

vitorpamplona:
Ohh sorry. I misunderstood the private needs. If we go forward with the private key = 1 idea, I would assume private relays would delete all 1064 (data) events that don't have a header (1065) pointing to it in the same relay. Easy daily/hourly pruning procedures.

bleetube:
>  To enable decentralized file sharing.

This should be removed. Relying on centralized relay services to be a trusted counter-party to distribute data is at best as decentralized as using a CDN.

Bittorrent is a lot more decentralized by comparison because the tracker doesn't host the data. Even despite that, there is a history of bittorrent trackers being shut down for facilitating "decentralized file sharing."  iirc Search engines got a pass in the end for only linking to torrents.

A better description this approach is taking is written a few sentences later "For lay users to be able to publish their photo, file, audio on several servers in a simple and easy way." So I don't think the sentence needs to be replaced with anything.

It's also a sentence fragment.

frbitten:
> 

I don't understand the difficulty some have in understanding that this NIP is not for file sharing.  It's to enable social networking with other media and not just text.  I want to see make an Instagram using Bittorrent.

benthecarman:
> I want to see make an Instagram using Bittorrent.

That is not what this enables, this would just be encoding instagram images in base64 on relays. That is why people are against this.

ok300:
>I don't understand the difficulty some have in understanding that this NIP is not for file sharing

Maybe its because

1. The NIP text proposed by the PR says `To enable decentralized file sharing.`
2. The technical implications of what the PR describes is file sharing.

If you have a bigger goal in mind, that relies on file sharing as a stepping stone, doesn't invalidate the dangers of enabling file sharing which people brought up.

frbitten:
Images are files, photos are files, gif are files, videos are files.  I just did it generically so as not to limit future usability.  But to think that it will be used to share files and compete with Bittorrent is too exaggerating just to be against it.  In addition to the challenge of having a Github nostr will have to go through a similar solution.

vitorpamplona:
The goal is to identify files (or any byte array) with event ids instead of URLs, such that they can be moved from one relay to another without having to update any references to it.

ok300:
If the goal is *only* to identify files by event ID, then what you're moving from one relay to another is references. Like moving a pointer.

Seems to me the PR is more about being able to store and move the actual data. Ergo, files themselves.

frbitten:
> Se o objetivo é _apenas_ identificar arquivos por ID de evento, o que você está movendo de um retransmissor para outro são as referências. Como mover um ponteiro.
> 
> Parece-me que o PR é mais sobre ser capaz de armazenar e mover os dados reais. Portanto, os próprios arquivos.

NOSTR is about communication not storage.  So much so that it is very clear that the relay defines the best way to store it and limits its size.

vitorpamplona:
> If the goal is only to identify files by event ID, then what you're moving from one relay to another is references. Like moving a pointer.

No, you would move the data itself. If a kind 1 cites a nevent1, the nevent1 can move from relay to relay and the Client will figure out where to get it from, it doesn't matter which relay the actual file is currently parked.

frbitten:
> If the goal is _only_ to identify files by event ID, then what you're moving from one relay to another is references. Like moving a pointer.
> 
> Seems to me the PR is more about being able to store and move the actual data. Ergo, files themselves.

NOSTR is about communication not storage.  So much so that it is very clear that the relay defines the best way to store it and limits its size.

ok300:
> No, you would move the data itself. 

So it stores the data itself. The file.

> NOSTR is about communication not storage. So much so that it is very clear that the relay defines the best way to store it and limits its size.

Same can be said about a simple HDD. Why is nobody offering free file hosting and sharing online, under a domain linked to their name? I wonder...

bleetube:
> The relay can allow access to this data via URL, having its own URL pattern for these cases. And if you receive a NIP-94 referring to a NIP-95 you can include the URL in the proper NIP-94 tag

The syntax here should probably be "may" rather than "can."

Also, it is very peculiar for a nostr proposal to define non-nostr communications between relays and clients. 

This NIP would impress me a lot more if it included a monetization scheme, where relays could charge sats for bytes. I think you could make a much stronger defense of the NIP if it included nominal specification for that, even if the default was to charge zero sats per byte to accomplish whatever it is you're trying to do.

frbitten:
Obviously it's going to be a paid service, you can't keep something like that free.  But I leave it free for the relay to define its business model.  As described in NIP is for specialized relays, it will not be supported by most relays.

ok300:
Having files addressable by event ID has the same problem as having them addressable by URL: relays can go offline and the event can be gone.

The solution of having files redundantly replicated on multiple relays is the same as a torrent magnet link with multiple trackers.

"But I pay for this relay, I'm reasonably sure it will provide storage for my data events" => how is that different than the relay seeding the torrents? Providers already exist that offer multiple services (NIP-05 aliases, LN Addresses, etc). What's stopping a paid relay operator from providing torrent seeding services for paying subscribers?

A simpler version of that: what's stopping them from providing simple image hosting on a URL? (Which some already do)

What's stopping several such providers from agreements among themselves, to mirror each other's stored images?

IMO storing the files in event data brings very little to no benefit relative to addressing via URL.

The only benefit I see is it removes reliance on DNS. But otherwise same risks remain (hoster can go offline), same mitigations (mirroring / replication across hosters).

vitorpamplona:
> Having files addressable by event ID has the same problem as having them addressable by URL: relays can go offline and the event can be gone.

No. We do this today with regular events: there is always a second relay that has them. It works like a charm.

frbitten:
> Having files addressable by event ID has the same problem as having them addressable by URL: relays can go offline and the event can be gone.

This already happens today with common events, it is a normal NOSTR reality. Nothing is perfect and there are always weaknesses.

> The solution of having files redundantly replicated on multiple relays is the same as a torrent magnet link with multiple trackers.
> 
> "But I pay for this relay, I'm reasonably sure it will provide storage for my data events" => how is that different than the relay seeding the torrents? Providers already exist that offer multiple services (NIP-05 aliases, LN Addresses, etc). What's stopping a paid relay operator from providing torrent seeding services for paying subscribers?

NOSTR can be used as a communication protocol to broadcast or upload torrent files. A more user-friendly platform via NOSTR can greatly increase NOSTR usage. It can be a frontend to Torrent, like a frontend to Git or any other file system. The main strength of NOSTR is the ease of integration between many different systems. The tendency is for NOSTR to become the generic API used by the most diverse systems.

> A simpler version of that: what's stopping them from providing simple image hosting on a URL? (Which some already do)
> 
> What's stopping several such providers from agreements among themselves, to mirror each other's stored images? Images can be addressable by hash, providers could auto-rename the files to their hash, so they're easily addressable.

Exactly. Nothing prevents. As nothing prevents them from using the NOSTR protocol itself to do this, since it is a protocol that everyone knows.

The file hosting server has its NOSTR relay and with that it can receive new files, share files with other servers, charge for the service and even provide files for other applications. No need to define and implement several different communication interfaces.
 
> IMO storing the files in event data brings very little to no benefit relative to addressing via URL.
> 
> The only benefit I see is it removes reliance on DNS. But otherwise same risks remain (hoster can go offline), same mitigations (mirroring / replication across hosters).

URL addressing is not storage. It is communication and data transfer. storage is handled by the file system made operating system.
In terms of communication and in the tests carried out and released, NOSTR loses nothing in terms of performance compared to HTTP.

You yourself come to the conclusion that HTTP and NIP-95 have the same risks. So saying that HTTP is better makes no sense. If the risks and mitigations are the same, the use is the same. If neither has benefits over the other, there is no reason for 1 to be better than the other technically. It's just a matter of taste.

I don't understand why they want to separate HTTP from NIP-95 so much. Ideally, servers that implement both approaches. They can receive files via NIP-95 and make them available in a URL as well as receive via HTTP and make them available via NIP-95.

There is no perfect solution, which is why it is important to leave the testing of new approaches free. The essence of NOSTR is just that, everything is optional so everyone is free to test and create innovative versions.

bleetube:
Maybe initially this NIP should be opt-in only for relays, such that any client that wants to store data using this nip should first check if the relay explicitly supports NIP-95 by checking the nip-11 `supported_nips`. Otherwise, complacent relay operators may be shocked at the increased resource consumption, or face a denial of service due to running out of disk space.

mikestaub:
I think this NIP is dangerous to merge and is relay scope creep. The events should always be exclusively JSON strings. We don't commit binaries to our git repos, we use https://git-lfs.com or https://github.com/iterative/dvc

frbitten:
> I think this NIP is dangerous to merge and is relay scope creep. The events should always be exclusively JSON strings. We don't commit binaries to our git repos, we use https://git-lfs.com or https://github.com/iterative/dvc

But the NIP is precisely for that. To not send binaries. By using Base64 it guarantees that JSON strings will be sent.

By using https://git-lfs.com you are centralizing the solution and you will not get a really decentralized solution.

vitorpamplona:
> An event bus should not have to care or know about blob storage systems.

This sentence doesn't make much sense. An event bus doesn't have a storage component. By design, event buses only carry ephemeral objects. Nostr is not an event bus alone. On AWS, for instance, to replicate Nostr, you would have to instantiate an Event Bus and the Event Replay for storage. And that storage product takes in everything you can think of.

vitorpamplona:
Pros: 
- URL/DNS independence. Referencing event IDs without a relay reference is better for censorship resistance. 
- Composability. NIP95 events can be cited directly that every other event: they can be liked, reported, zapped, reshared, etc. You can't zap a URL. 
- Notifications: Unlike URLs, resharing an event ID opens the door to notifying the author of the event
- API Independence: clients don't need to code a different uploading API for every image server out there
- Faster than http if already connected to the relay. 

Cons: 
- Not re-using the HTTP multipart uploading and streaming protocols for download (though using WebSockets for binary streaming is becoming more common). 
- Creates Large JSONs whose `.content` should, but might not need to, be stored separately from the other JSON properties.
- Transfers data in base64 which are inherently more costly to data plans out there (however, I would argue that the use of HEX keys everywhere IDs, pubkeys, signatures, tags, etc is a worse offender). Nevertheless, a regular JSON text compression basically reverts any losses: https://lemire.me/blog/2019/01/30/what-is-the-space-overhead-of-base64-encoding/

frbitten:
> @frbitten When I said "binaries" I meant non-human readable encoded data of any form. The events must remain small and reference external files. They can be hosted on IPFS, Arweave, Sia, etc if you care about centralization.
OK. So it's against markdown, NIP-23, NIP-78 and big text in posts. Shall we limit the number of characters equal to Twiter?
Long events have always been possible, the relay having to define its limits. the NIP-95 didn't invent that now.

frbitten:
> The events being small is less about resource constraints on the relay and more about software architecture patterns. An event bus should not have to care or know about blob storage systems. They scale in different ways and thus should remain decoupled.

NOSTR is about communication and message passing. It is not exactly an event system, much less a storage system. The technology used in the storage is a choice of the relay. A relay that wants to store large data will need to choose a compatible storage technology for this functionality.

frbitten:
> This is a pretty controversial NIP I think for good reason, but the main sticking point for some people seems to be how inefficient/ugly base64 encoding images and embedding them into JSON is. What would everyone think of an alternative approach:
> 
> * Client sends an ephemeral event requesting permission to upload a file to the relay's server.
> * Relay responds with a url to POST the file to, requiring headers, content type, etc.
> * Client POSTs the file to the url
> * On successful upload, the relay creates something very similar to a NIP-94 event, except instead of a url, it specifies a file name.
> * Clients can dereference that file id by building a url for it at a known path, e.g. `https://my-relay.example.com/nostr/nip95/<eventid>/<filename>` (event id is included to ensure uniqueness)
> * If a file event is re-published to a different relay (or the relay actively replicates the event), the new relay MUST request the file and host it if it is to serve the original event.
> 
> This fixes the base64 inefficiency, and enables all the nice features that go along with using HTTP, including caching, CDNs, etc. Adding a requirement for relays to run an HTTP server is not novel because of NIP 11, and the endpoint is optional anyway since it's only required for relays that support NIP 95.



> This is a pretty controversial NIP I think for good reason, but the main sticking point for some people seems to be how inefficient/ugly base64 encoding images and embedding them into JSON is. What would everyone think of an alternative approach:
> 
> * Client sends an ephemeral event requesting permission to upload a file to the relay's server.
> * Relay responds with a url to POST the file to, requiring headers, content type, etc.
> * Client POSTs the file to the url
> * On successful upload, the relay creates something very similar to a NIP-94 event, except instead of a url, it specifies a file name.
> * Clients can dereference that file id by building a url for it at a known path, e.g. `https://my-relay.example.com/nostr/nip95/<eventid>/<filename>` (event id is included to ensure uniqueness)
> * If a file event is re-published to a different relay (or the relay actively replicates the event), the new relay MUST request the file and host it if it is to serve the original event.
> 
> This fixes the base64 inefficiency, and enables all the nice features that go along with using HTTP, including caching, CDNs, etc. Adding a requirement for relays to run an HTTP server is not novel because of NIP 11, and the endpoint is optional anyway since it's only required for relays that support NIP 95.

Finally a good suggestion and not just an empty criticism complaining about the size of the events.

It's a good suggestion and I think it might be viable. I can't say if the message and response mechanics are well defined in NOSTR. I don't know that part well yet.

A negative point of this approach and which is, in my point of view, the main advantage of NOSTR in decentralization is the ease of sending events to several relays in a simple way. Because it is exactly the same event sent to everyone.
We also lost the option of proxy relays, which group several relays on a single connection and can manage the message for the specialized relay to handle it. Even a backup option for your events is hampered with this approach.

I imagine that in the not too distant future we will have several layers of relays that provide services to client applications, but the user can still choose the final relay to store his data.

frbitten:
The data increase problem when using base64 can be solved in the future by having pure TCP connections between client app and relays communicating the same NOSTR events but in compressed format to save internet bandwidth.

Soon it will be sendable to open a nostr app globally and the avalanches of new events will consume all the cell phone's resources. Both 4G band and CPU. A more optimized connection may be the solution not only for the NIP-95 but also for searches that will return thousands of events at once.

New features like NIP-95 are optional features that in the future only relays and apps that need this resource saving will implement.

BlowaterNostr:
After reading NIP95, I found that it's almost identical to how Blowater has been sending encrypted imaged through normal kind-4. [Blowater](blowater.deno.dev) has been experimenting with this idea for 2 months and here is my take.

Basically, Blowater cuts an image to chunks, base64 them, and encrypt the base64 string. The receiver reassembles them.

Each part as a separate event and refers to the first event (header) in tags.

The problem of this approach is that, it's slow at client side because every time a new event arrives, the client has to check if it's part of an image and if has received all of its parts.

Many relays drop/forget data if a pubkey has used too much. Therefore, it's possible that some parts never arrive.

Of course, if relays choose to implement NIP95 style storage, this won't be a problem.

Still, reassemble is expensive and will cause performance issues, especially for single threaded JavaScript applications.

Another problem is that, NIP95 or current Blowater approach break the atomicity of Nostr events. Make other parts of a client harder to implement, such as Pin/Bookmark a message because it's not represented by a single event anymore.

I believe the argument here is not about should we use relay as a storage layer. But rather, should we keep Nostr events atomic and represent a whole thing.

frbitten:
> After reading NIP95, I found that it's almost identical to how Blowater has been sending encrypted imaged through normal kind-4. [Blowater](blowater.deno.dev) has been experimenting with this idea for 2 months and here is my take.
> 
> Basically, Blowater cuts an image to chunks, base64 them, and encrypt the base64 string. The receiver reassembles them.
> 
> Each part as a separate event and refers to the first event (header) in tags.
> 
> The problem of this approach is that, it's slow at client side because every time a new event arrives, the client has to check if it's part of an image and if has received all of its parts.
> 
> Many relays drop/forget data if a pubkey has used too much. Therefore, it's possible that some parts never arrive.
> 
> Of course, if relays choose to implement NIP95 style storage, this won't be a problem.
> 
> Still, reassemble is expensive and will cause performance issues, especially for single threaded JavaScript applications.
> 
> Another problem is that, NIP95 or current Blowater approach break the atomicity of Nostr events. Make other parts of a client harder to implement, such as Pin/Bookmark a message because it's not represented by a single event anymore.
> 
> I believe the argument here is not about should we use relay as a storage layer. But rather, should we keep Nostr events atomic and represent a whole thing.

Why divide the images into several pieces? Wouldn't it be simpler to send everything in a single event?

Did you encounter any limitations that forced you to split it into smaller pieces? What size images are you sending?

BlowaterNostr:
@frbitten most relays have this 100KB event size limit. Some relays uses 256KB limit. I use 32KB chunk size. After base64 encoding, some could be as large as 64KB. A high resolution screen shot can easily be larger than 1 MB, consider now many people have 2K+ phone screens and 4K+ desktop monitors.

frbitten:
@BlowaterNostr  OK. So it's not a protocol or base64 problem. The problem is that you want to circumvent the maximum relay limitations. The correct thing is to send to relays that allow events with more data.

That's what I say in the NIP. This option will be made for specialized relays, it should not be expected that any relay will support this.

And the main advantage of NIP existing is knowing which relays allow this functionality and which do not. So clients can only send events to relays that support NIP. And relays can easily block this type of data.

arthurfranca:
> Maybe we could not waste time on suboptimal solutions like this and make a Nostr for file storage

Yes it would be better for sure. But how would that look like?

I think nostr trump card (lol don't know if this expression is right - its outstanding feature) are the relays, not the clients/users. So I imagine that in a bittorrent(-like) network, the relays should be the seeds, not the clients/users with webtorrent. Cause relays are always online, clients aren't. And web/mobile client storage is usually limited (desktop isn't).

To be lower on relay storage requirements, relays should be seeds without the entire set of the data, never keeping the whole file, so to share the burden.

Clients would ideally download from a regular url at a relay that behind the scenes does the complex stuff of temporally assembling the full file. (Although I know that nostr usually pushes for client complexity and relay simplicity)

Hard to achieve but would be great to see it. Is there a genius in the room? If not, maybe we'll end up using NIP-95 anyway xd

bleetube:
> But how would that look like?

@arthurfranca egge7's [nostr-ing](https://github.com/nostr-ing/nostr-ing-protocol) is one idea towards this.

arthurfranca:
@bleetube The @Egge7 spec is clever. Just don't see much difference besides base64 vs binary. It trades reusing the current installed relay structure for saving some space. This alone doesn't seem to bring interest.

> [Receiving Data](https://github.com/nostr-ing/nostr-ing-protocol/blob/main/README.md?plain=1#L65) Receiving clients subscribe to transmissions by connecting to a nostr-ing relay and passing a transmission ID as part of the connection URL.

It seems that one is supposed to upload full file to a relay (or to each of many). But them you download all chunks from just one relay. So, the same as NIP-95.

The slight difference is that nostr-ing chunks have sequence numbers embedded to each of them (not in the header event - "metadata event") while NIP-95 uses `e` references in the header event and may have [a third event(?), the NIP-94 one](https://github.com/nostr-protocol/nips/pull/345/files#diff-319c9bcebb9aa1a6bb2187a77604e2cb409af365a9f504e6d38f01627b6810d4R54), to allow downloading from a url. <- This thing of allowing download from url is great but having 3 different kinds of events seems too much?

[This old PR](https://github.com/nostr-protocol/nips/pull/112) from @Egge7 is similar to NIP-95 and was rejected. I guess this one won't be merged too.

staab:
> Yes it would be better for sure. But how would that look like?

Most of you probably already saw this: https://github.com/nostr-protocol/nips/pull/345#issuecomment-1538640440 but I figured I'd bump it. Any feedback on the concept? It seems to solve the problem cleanly by being opt-in, HTTP-based, and piggybacking on nostr's relay infrastructure. There are two upvotes on the comment, so unless there are any specific objections to it, it could be way forward.

arthurfranca:
@Semisol @staab a merge of your ideas seem great. New PR? =)

The [#345 comment](https://github.com/nostr-protocol/nips/pull/345#issuecomment-1538640440) miss some things like the DELETE method. The fixed url format (`https://my-relay.example.com/nostr/nip95/<eventid>/<filename>`) is good cause allows clients to search for that eventid on other relays.

No doubt the DELETE method must use [NIP-98](https://github.com/nostr-protocol/nips/pull/469). @staab If we use NIP-98 for the POST too, it may be possible to skip the ephemeral event step and use a fixed url path for uploads.

I would add that if client sends file to three relays, when later posting the url on a note it can use the above url format plus query string `?r=wss://second-relay.com&r=wss://third-relay.com`. <- *Just don't know yet how the best way to end up with the same eventid (nip95/\<eventid\>/) on the url (fields like created_at must be the same no matter to how many relays i'm sending the same file). Maybe `https://my-relay.example.com/nostr/nip95/<sha256hash/>` is better and then one can search on other relays for the event by hash instead of by id.

This would instruct my-relay.example to redirect to second-relay.com (discarding just the first `r` query param) if the first relay doesn't have the file anymore. The second one can do the same redirecting to the third one if it doesn't have the file too. Probably wise to limit it to 3 `r` query params to avoid redirect hell.

Not bittorrent-like but seems very good.

arthurfranca:
> @Semisol @staab a merge of your ideas seem great. New PR? =)

I'm writing a PR for this

ondra-novak:
Stupid question. Do you know, that websocket is capable to transfer binary messages (I didn't read whole discussion]

vitorpamplona:
> Stupid question. Do you know, that websocket is capable to transfer binary messages (I didn't read whole discussion]

It's very early development in binary sockets (few years) but it's growing really fast and now that we have seen the adoption by [streaming services](https://www.twilio.com/docs/voice/tutorials/consume-real-time-media-stream-using-websockets-python-and-flask), it tends to grow faster.

ondra-novak:
> > Stupid question. Do you know, that websocket is capable to transfer binary messages (I didn't read whole discussion]
> 
> It's very early development in binary sockets (few years) but it's growing really fast and now that we have seen the adoption by [streaming services](https://www.twilio.com/docs/voice/tutorials/consume-real-time-media-stream-using-websockets-python-and-flask), it tends to grow faster

binary messages in websocket protocol are there from the begging. Support in browsers is' chrome 15, Firefox 11, literally internet stone age

https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/binaryType

vitorpamplona:
> > > Stupid question. Do you know, that websocket is capable to transfer binary messages (I didn't read whole discussion]
> > 
> > 
> > It's very early development in binary sockets (few years) but it's growing really fast and now that we have seen the adoption by [streaming services](https://www.twilio.com/docs/voice/tutorials/consume-real-time-media-stream-using-websockets-python-and-flask), it tends to grow faster
> 
> binary messages in websocket protocol are there from the begin Support in browsers is' chrome 15, Firefox 11, literally internet stone age
> 
> https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/binaryType

Yeah, but in practice libraries were all broken until 3-4 years ago (I tried using it in another project). It's way better right now.

frbitten:
We have Amethyst that implements NIP-95, but we need relays that implement the proposal and support this functionality.

If you are available to create the relay, the NIP can be approved. But remember that it's an optional implementation and most relays won't implement it.

ondra-novak:
This works for me, but it's a very poorly thought out design. For example, it doesn't address file storage within encrypted communication - which I'm trying to address in my proposal (which I'm currently testing).

The advantage is that the NIP-95 does not need to support the relay in any way. Which I consider to be a form of "hacking" rather than a reasonable protocol design.

Personally, I was wondering if I should block the NIP-95 on my relay or regulate it in some other way. It is very inefficient to store and poorly managed in the database. I also tried the option of base64 decoding the content for storage purposes. Which in turn has the base64 problem described above.

It would be optimal to use binary messages, which will also be part of my proposal. Stay tuned

frbitten:
the NIP-95 has encryption options, you can use the decrypt tag for this purpose.

The NOSTR protocol is about communication only. It does not define how the data is stored, it is free for each relay to choose the best way in their case.

All NOSTR NIPs are optional, so it's not a NIP-95 issue. NOSTR by definition is 100% optional.

Binary data is not compatible with NOSTR and therefore all implementations are via external protocols, but you are free to block NIP-95 or not. No relay is required to implement all proposed NIPS.

staab:
A little rusty on the topic, but https://github.com/nostr-protocol/nips/pull/547 seems more promising to me. Or are the two NIPs orthogonal?

frbitten:
@staab I quickly read NIP-96, it really seems like good continuity in the discussion to arrive at an optimal solution. I'll follow along and read it carefully as soon as I have time.

jb55:
On Thu, Aug 03, 2023 at 11:23:28AM -0700, Ondrej Novak wrote:
>It would be optimal to use binary messages, which will also be part of my proposal. Stay tuned

awesome, thanks for starting this.

ondra-novak:
#694

ondra-novak:
#719  version 2

capitalistdog:
nack to this whole concept of files on relays

- it puts a large burden on resources and therefore will have a centralizing effect on relays
- that's in the best case of if it gains traction at all, the most likely outcome is people publishing things that don't load on most clients
- delivery of video and large image files for example also requires optimization, are ffmpeg recipes getting a nip too?
- still need an http server for meeting users in things like opengraph
- there are already better protocols turn-key for decentralizing delivery, and nostr is perfectly equipped to simply pass the metadata for these
- a sane nip would be standardizing the client-side only for uploading / paying / retrieving the metadata and then posting and displaying that. Services can then compete for this. 

Example of how we're looking at this at Lightning Video: We wrap all hosted files in **WebTorrent** so that mirrors and p2p swarms add redundancy and censorship resistance. This WebTorrent magnet can be delivered just as a regular note, and contains traditional HTTP URI's and file hashes.

Duplication of files across relays gains nothing in terms of censorship resistance, as Nostr + Magnet links take the namespace element of censorship largely out of the equation. This simple redundancy is proven over years of torrent evolution.

Bear in mind that censorship resistance is ultimately about incentives. The tech must be simple. Ex: a paid host has no incentive to remove your vaccine content vs a free platform taking in billions of advertising dollars from Pfizer.

vitorpamplona:
> Example of how we're looking at this at Lightning Video: We wrap all hosted files in WebTorrent so that mirrors and p2p swarms add redundancy and censorship resistance. This WebTorrent magnet can be delivered just as a regular note, and contains traditional HTTP URI's and file hashes.

Do you have a specific proposal for this? None of the magnet proposals I have seen are actually decentralizing in nature. But as long as the magnet URI is not fixed to any particular domain name or provider, it should be fine. Otherwise, it's just fake decentralization: the content may be decentralized, but the URI might still require a controllable central entity to "forward" the Clients to content endpoints.

capitalistdog:
> Do you have a specific proposal for this? None of the magnet proposals I have seen are actually decentralizing in nature. But as long as the magnet URI is not fixed to any particular domain name or provider, it should be fine. Otherwise, it's just fake decentralization: the content may be decentralized, but the URI might still require a controllable central entity to "forward" the Clients to content endpoints.

Best to look at the WebTorrent docs, but we do have an example server available here: https://github.com/shocknet/seed - an SDK with a client is coming, at which point we'll draft a NIP... but I'm happy to collaborate sooner.

As you'll see we return the magnet with our Tracker and URI information, but a client could also inject a redundant provider values before posting. The WebTorrent client would then consider all the sources listed in the combined magnet, seamlessly to the user, while also benefiting from the p2p assist.

capitalistdog:
> So, once the user creates a URI (even with multiple providers) and saves it in an event, is the user stuck with those providers forever? What happens when all providers go out of business as they will inevitably do? Does the event's linked content stop loading? How can the user upload his/her content to another provider after the event has been signed and fully distributed? Or how can clients automatically search for the content if the provider isn't available anymore?

Consideration could be made to make the publish events replaceable by giving them a neutral identifier that's added to a library.  

Things like search come down to tagging like anything else, it's irrelevant to the file host as they aren't the ones doing the indexing. 

The presupposition that "if you can see the note you should absolutely be able to see the content" is counterintuitive when the note has no reach in the first place.

vitorpamplona:
> Consideration could be made to make the publish events replaceable by giving them a neutral identifier that's added to a library.

Not possible in most events/use cases. 

NIP-95 (this NIP), [NIP-96](https://github.com/nostr-protocol/nips/pull/547), and [NIP-97](https://github.com/nostr-protocol/nips/pull/719) all offer ways to "rebroadcast" content to another server/relay after the fact. Even decades after the signature. Clients can find it dynamically. So that no matter what happens to any provider, as long as the user has a backup of the original content and can re-broadcast it, the content linked in past events can always be easily found and will always be verifiably available. I suggest you assemble the proposal with the same feature in mind and then we can debate it further.

capitalistdog:
> Not possible in most events/use cases.

Why not? The content is not itself an event to be referenced.

Just as a user can update any number of properties about themselves, they too could update the contents of their library, which contains the value apps would reference. 
 
> ways to "rebroadcast" content to another server/relay after the fact. Even decades after the signature, clients can find the content dynamically. So that no matter what happens to any provider, as long as the user has a backup of the original content and can re-broadcast it, the content linked in past events can always be easily found and will always be verifiably available. I suggest you assemble the proposal with the same feature in mind and then we can debate it further.

That's called self-hosting and functionally no different than seeding your own WebTorrent.

vitorpamplona:
> Just as a user can update any number of properties about themselves, they too could update the contents of their library, 

Ok, so what's saved into the event itself? Can you give an example?

capitalistdog:
> Ok, so what's saved into the event itself? Can you give an example?

NIP-15 marketplaces look like a good example. Replace stall with content category, and product with a content identifier who's specs contain the latest magnet. 

I personally like the idea of treating content as a SKU because we've had a lot of traction with paywalls, but its also useful for other engagement types.

Replacable lists would also work, each piece of content being it's own list containing the event for the updated magnet. 

In any case, when you make a note to your feed, you'd tag the product/list in either case.

vitorpamplona:
Oh ok, so the idea is for the user to make a replaceable event for each magnet uri. If the magnet URI provider goes out of business the user has to look back on all replaceable events of that provider and update them all with a new URI from a new provider. No posts will ever link to a magnet URI directly. They will always link to a replaceable redirection first. 

It adds a lot of work for the user, but it works. I just don't know why using a magnet. That same scheme can also have direct links to the servers that have the image. If the server goes out of business, look for all replaceable events with that server and update it to a new one.

capitalistdog:
> has to look back on all replaceable events

Do you mean for monitoring for liveness? 

In just updating vendors theres no retrace, its just 2 updates, one for the resource and the other for the library index. 

> I just don't know why using a magnet. That same scheme can also have direct links to the servers that have the image.

The magnet is in the context of redundancy, since the whole point of these proposals is people are unhappy with URL's for some reason.

vitorpamplona:
Well, I look forward to the full proposal so that we can analyze how decentralizing it actually is.

fiatjaf:
So what is the state of this NIP? Is it being used by anyone? I can see a bunch of kind:1064 events in the wild.

vitorpamplona:
Amethyst still uses it. Anything under 100KB (memes and stuff) works really well and is accepted by most relays.

fiatjaf:
Why do you say "still"? Are there plans to stop using it?

vitorpamplona:
> Why do you say "still"? Are there plans to stop using it?

Because most people got angry when we first offered the support and wanted me to drop support immediately. 

I don't have plans to remove it. I think it is a useful kind for compressed web images, icons, and some other types of light files.

95.md
```diff
@@ -0,0 +1,75 @@
...
+======
+`draft` `optional` `author:frbitten` 
+
+This NIP allows users to send binary data to be stored in the relay. This data can be small pieces of information or even large files.
+
+The intention is to combine the ease of NOSTR, with many clients being developed and a common communication protocol between all. To enable decentralized file sharing. NOSTR communication must work together with other protocols to get the best out of each technology. For lay users to be able to publish their photo, file, audio on several servers in a simple and easy way.
+
+Nostr event
+------------------
+This NIP specifies the use of the `1064` event type (parameterized replaceable event), having in `content` the binary data that you want to be stored in Base64 format.
```
NfNitLoop, line 11: [NIP-01] says that:
> for kind `n` such that 1000 <= n < 10000, events are **regular** [...]

Not parameterized or replaceable.

Which is good. I wouldn't want to quote or reply to some image only to have the author troll by changing it out to something else later.

[NIP-01]: https://github.com/nostr-protocol/nips/blob/master/01.md

95.md
```diff
@@ -0,0 +1,75 @@
...
+}
+````
+
+Relay Behavior
+---------------
+Relays that use a relational database may have more difficulty implementing this NIP, but a possible solution is for this NIP not to be recorded in the database, but on disk, the file name being the event id. So it can be easily found and searched. And because it is not in the database, it does not interfere with the indexing of common events.
+
+Another solution is for Relays that want to implement this functionality and use a No-SQL database with mongodb that already supports large documents without harming performance.
+
+The relay can allow access to this data via URL, having its own URL pattern for these cases. And if you receive a `NIP-94` referring to a `NIP-95` you can include the URL in the proper `NIP-94` tag
```
NfNitLoop, line 54: Including URLs in events can lead to linkrot. (Though they're a nice fallback for clients that haven't implemented this NIP for pulling contents out of events.)

Maybe a better alternative would do document a way for relays to advertise with [NIP-11] what the URL pattern is to fetch these files from it via HTTP(S).

[NIP-11]: https://github.com/nostr-protocol/nips/blob/master/11.md

95.md
```diff
@@ -0,0 +1,75 @@
...
+
+Client Behavior
+---------------
+The client can use this event for any type of storage it wants. The functionality of uploading an image directly through the client's interface and including it in the note that is being created is a very useful facility for the end user. In this case, 3 events would be created. Kind 30064 with the file converted to base64, Kind 30063 which is the header of this file for dissemination and Kind 1 with the post and the image.
```
NfNitLoop, line 58: Are the kinds 30063 & 4, or 1064 & 5?

95.md
```diff
@@ -0,0 +1,75 @@
...
+
+```json
+{
+  "id": <32-bytes lowercase hex-encoded sha256 of the the serialized event data>,
+  "pubkey": <32-bytes lowercase hex-encoded public key of the event creator>,
+  "created_at": <unix timestamp in seconds>,
+  "kind": 1064,
+  "tags": [
+    ["decrypt",<algorithm>,<Decryption Params>],
+    ["p", <32-bytes hex of a pubkey>, <recommended relay URL>],
```
NfNitLoop, line 25: Is `p` redundant with `pubkey` here? Is this just so that you can recommend a relay URL for that pubkey?

Do you want to include an example `type` tag here?  Actually, metadata like that should probably be moved to the 1065 kind, right? Would be useful for clients to know mime type, size, dimensions, etc, before choosing whether to fetch the file.

NfNitLoop (commented):
I've been referring to this NIP as I do some brainstorming for an experimental implementation of similar functionality. Since this is still an open PR, I'll share thoughts here in case you want to consider any of them for this NIP.

Things I'm planning on changing for my experiment:

The "meta" event (kind 1065):
* should have the lower ID, because it will be both searched/fetched by users first, and sent to the server first.
* MUST include metadata for mime type and (final, deserialized, binary) file size, in bytes, so that clients can choose whether they want to fetch it.
* MUST include an ~~"o"~~ "x" tag with the sha256 of the file.
* Can support large(r) files by referring to multiple other event IDs, the binary contents of which can be concatenated together, in tag-order, to reconstruct the original file.
* MUST specify, when multiple "content" events are included, a tag `block_size`. The "block_size" is the binary size, in bytes, of the data of each block before the last block (which may be smaller, but not zero). This allows relays and clients to only fetch certain parts of a file, ex: when streaming video, or resuming downloads.

Clients:  
Because of the overhead of signing/encoding/fetching/sending each message, clients should choose larger block sizes for multi-event file uploads when possible, but they may choose smaller block sizes when it makes sense for content, or to meet relays' max_message_length limitations/policies. Clients SHOULD prefer block sizes that are multiples of 3, because those encode into base64 efficiently. (3bytes:4chars).  

Relays:
* MAY choose to reject a "meta" event if it doesn't want to store files, a user is over a quota, the file content type is not acceptable, the file is lager than allowed by that relay, or whatever reasons the relay operators wish. Rejection of the "meta" event serves as a proxy rejection for all of its associated "content" events, too, so clients shouldn't send "content" events in that case.
* Relays SHOULD reject any content events they receive if they have not yet accepted the associated "meta" event.
* Relays MAY(/SHOULD?) enforce a minimum block size for multi-part files, and/or a maximum number of parts, to avoid abuse. (encoding a large file with 1-byte blocks would have ridiculous overhead.)

Regarding the overhead of Base64 -- it's unfortunate for the wire protocol but short of changing that, the best workaround is that servers can just pull the file contents out and store them in binary (either in a file or as a BLOB). They may want to store the binary in a content-addressable store, and requiring a sha hash of the file helps with that.

I'm https://njump.me/nfnitloop.com if anyone wants to brainstorm more about this. I'm new to Nostr, so forgive me if I missed anything that should be obvious. 😅

95.md
frbitten: Yes. We had an argument at the beginning and we changed the kid and I forgot to change the description. But a file cannot be changed. The change means being a new file in a new note.

95.md
frbitten: the NIP-94 can be generated either by the client when creating the NIP-95 note (in this case you will need a definition of the URL format) or it can be generated by the relay itself to list your saved files. Because in NIP-94, the author of the note is not as important as it is the disclosure of the URL

95.md
frbitten: Another case that was not changed after the kind change.

95.md
frbitten: Yes. Just to reference other users, following the NIP-01 standard. It is not mandatory.

frbitten:
> I've been referring to this NIP as I do some brainstorming for an experimental implementation of similar functionality. Since this is still an open PR, I'll share thoughts here in case you want to consider any of them for this NIP.
> 
> Things I'm planning on changing for my experiment:
> 
> The "meta" event:
> 
> * should have the lower ID, because it will be both searched/fetched by users first, and sent to the server first.
> * MUST include metadata for mime type and (final, deserialized, binary) file size, in bytes, so that clients can choose whether they want to fetch it.
> * MUST include an "o" tag with the sha256 of the file.
> * Can support large(r) files by referring to multiple other event IDs, the binary contents of which can be concatenated together, in tag-order, to reconstruct the original file.
> * MUST specify, when multiple "content" events are included, a tag `block_size`. The "block_size" is the binary size, in bytes, of the data of each block before the last block (which may be smaller, but not zero). This allows relays and clients to only fetch certain parts of a file, ex: when streaming video, or resuming downloads.
> 
> Clients: Because of the overhead of signing/encoding/fetching/sending each message, clients should choose larger block sizes for multi-event file uploads when possible, but they may choose smaller block sizes when it makes sense for content, or to meet relays' max_message_length limitations/policies. Clients SHOULD prefer block sizes that are multiples of 3, because those encode into base64 efficiently. (3bytes:4chars).
> 
> Relays:
> 
> * MAY choose to reject a "meta" event if it doesn't want to store files, a user is over a quota, the file content type is not acceptable, the file is lager than allowed by that relay, or whatever reasons the relay operators wish. Rejection of the "meta" event serves as a proxy rejection for all of its associated "content" events, too, so clients shouldn't send "content" events in that case.
> * Relays SHOULD reject any content events they receive if they have not yet accepted the associated "meta" event.
> * Relays MAY(/SHOULD?) enforce a minimum block size for multi-part files, and/or a maximum number of parts, to avoid abuse. (encoding a large file with 1-byte blocks would have ridiculous overhead.)
> 
> Regarding the overhead of Base64 -- it's unfortunate for the wire protocol but short of changing that, the best workaround is that servers can just pull the file contents out and store them in binary (either in a file or as a BLOB). They may want to store the binary in a content-addressable store, and requiring a sha hash of the file helps with that.
> 
> I'm https://njump.me/nfnitloop.com if anyone wants to brainstorm more about this. I'm new to Nostr, so forgive me if I missed anything that should be obvious. 😅

The idea was for each NIP-95 event to have a NIP-94 to describe it. NIP-94 already supports all the description fields you suggested. To prevent clients and servers from having to transmit large volumes of data just so the user knows the basic characteristics of the file.

Therefore, the NIP-95 event is only returned by relays in a direct request for its ID, it will not be returned in broad searches.

Ex: Imagine that I have a 1MB image sent on a NIP-95. If I just want to know its size, I will need to receive the entire event of more than 1MB to have this information. With NIP-94 I have all the "meta" information without having to transfer 1MB on the network.

vitorpamplona:
FYI, Amethyst's implementation still uses kind `1065` as a header for kind `1064` blobs. Kind `1065` is just a copy of NIP-94s kind 1063 file header without the URL-based tags.

frbitten:
> FYI, Amethyst's implementation still uses kind `1065` as a header for kind `1064` blobs. Kind `1065` is just a copy of NIP-94s kind 1063 file header without the URL-based tags.

Vitor well remembered.
I no longer remember what was the reason for proposing to use a different kind and not using the 1063 kind of NIP-64. Remember?

vitorpamplona:
It was mostly to make sure NIP-94 clients don't get bogged down having to support and manage NIP-95 blobs. Separating the headers allows clients to filter what they can support (url-based files or base64 blobs)

frbitten:
I was distracted and no longer remembered all the details. I adjusted it according to @vitorpamplona  warning. Please see if @NfNitLoop meets your needs now.

NfNitLoop:
@frbitten I'll have a look a bit later today. But I didn't see you comment on my ideas for multi-part files.  (breaking a large file up into smaller chunks so that each chunk fits within message size limits.)

Too far for this NIP? It would probably break existing implementations, so it's probably best to make it its own kind, so that could be a separate NIP.  🤔   Can probably re-use the "content" event message and make a new "meta" kind, as with the NIP-94/1065 split.

NfNitLoop (commented):
(OK, I was curious so I looked now. 😊) 

One of the features from my earlier brain-dump is that clients SHOULD send the meta (1065) event to relays before the content (1064) event(s). Then relays MAY reject the 1065 event, and thus spare themselves the bandwidth of receiving 1064 events if
* They don't support them at all
* The user/pubkey isn't allowed to post such events
* The file is too big
* The file (hash) has been banned
* etc, etc.

vitorpamplona:
> I didn't see you comment on my ideas for multi-part files.

This must be defined in some way. I think we never did it because none of us had that need at the time. But breaking the file into many events and recombining them must be deterministic.

NfNitLoop:
@vitorpamplona 

> But breaking the file into many events and recombining them must be deterministic.

Yep!  See my definition of how to do it [here].

> The "meta" event:
> * [...]
> * Can support large(r) files by referring to multiple other (kind 1064) "content" event IDs, the binary contents of which can be concatenated together, in tag-order, to reconstruct the original file.
> * MUST specify, when multiple "content" events are included, a tag `block_size`. The "block_size" is the binary size, in bytes, of the data of each block before the last block (which may be smaller, but not zero). This allows relays and clients to only fetch certain parts of a file, ex: when streaming video, or resuming downloads.

[here]: https://github.com/nostr-protocol/nips/pull/345#pullrequestreview-1964486737

vitorpamplona:
Nice! Do you have a demo somewhere?

NfNitLoop:
> Do you have a demo somewhere?

@vitorpamplona  Not yet. I'm working toward implementing a demo, which is why I shared my thinking here. I'll have a lot more time to spend on it in a couple weeks. I'm taking the long/scenic route, [writing my own relay][1] (and [probably web client][2]) to test it out in. 

Wasn't sure if I should shove multiple events into a kind 1065 or grab a new kind # to experiment with.  (Is there an atomic way to grab a kind # or nip # for this kind of experimentation? Or find one that's not yet in use?)

[1]: https://github.com/NfNitLoop/yastr
[2]: https://njump.me/nevent1qqsvycxdu79x85fxxchmgcuxzzp4srrpas27jk74v9fcveqz4plu4yczyrwyxyhydv8rsgg9692zjrzpnesx5ueqqnxhyr00rysspwg45xu6cpuxvps

vitorpamplona:
> Is there an atomic way to grab a kind # or nip # for this kind of experimentation? Or find one that's not yet in use?

If you are designing your own relays, just connect your tests to your relay and then delete when you finish the test. Otherwise, yes, just pick a number you have not seen yet in the major relays.

frbitten:
To need to break into several events, I imagine you are thinking about transmitting very large files. I don't know if Nostr is the best option for this.
Relays have a size limit at the event. This is how they limit the volume of data.

One option is for event 1065 to have several `e` tags. With each one having its part id and probably a second attribute that would be the index to define the concatenation order. I don't know if relying on the order in JSON is a good option.

NfNitLoop:
> To need to break into several events, I imagine you are thinking about transmitting very large files. I don't know if Nostr is the best option for this.

It's not necessarily "very large" files. It's just more the idea of having a general-purpose solution. The current state of NIP-95 only supports files up to the (encoded) message size that relays will accept.  I think I've seen that's usually something like 128KB? Even a simple JPEG can be larger than that.  


> One option is for event 1065 to have several e tags. 

This is just what I'm proposing above! 😊 

> and probably a second attribute that would be the index to define the concatenation order. I don't know if relying on the order in JSON is a good option.

Adding an index is unnecessary. Tags are in a JSON array, and JSON arrays (unlike objects/hashes) are ordered. (Which I assume is precisely the reason that the tags are in an array. It gives them a deterministic order for the event ID hash function.)

I've seen other nips warn about relying on ordering tags for giving extra meaning to them (ex: this `e` tag is what I'm replying to, this `e` tag is the "root" of the thread.)

But IMO in kind 1065, the ONLY `e` tags that should be supported/allowed would be to the 1064 events that make up the file contents, so this isn't an issue there.

frbitten:
The size limit of the relays is precisely because we don't want data that is not conversations. If this nip is merged they will prohibit these types.  So dividing into several just to get around the limit is not a good option.
The correct option would be to use relays that support this functionality.
For example, if you use the MongoDB database, the size limit for each event in the database is 16MB.  

Your suggestion is good but you have to pay attention to what the use would be.

vitorpamplona:
> Your suggestion is good but you have to pay attention to what the use would be.

This is why we need a demo with lots of testing. 

I agree that breaking the size down just to avoid limits in the current relays is not a good solution. Breaking into multiple payloads should be about parallelizing downloads, avoiding websocket payload limits, or some other reason.

NfNitLoop:
> Breaking into multiple payloads should be about parallelizing downloads, avoiding websocket payload limits,

I thought that's what I'd said. Maybe I communicated it poorly. 😅 

Agreed, I'm not trying to work around web servers' content *policies*.  In fact I explicitly made some suggestions to allow them to enforce policies:


> One of the features from my earlier brain-dump is that clients SHOULD send the meta (1065) event to relays before the content (1064) event(s). Then relays MAY reject the 1065 event, and thus spare themselves the bandwidth of receiving 1064 events if
> * They don't support them at all
> * The user/pubkey isn't allowed to post such events
> * The file is too big
> * The file (hash) has been banned
> * etc, etc.

vitorpamplona:
I think we are all in agreement. Let's just make sure we test this file chunking thing very well.

95.md
```diff
@@ -0,0 +1,88 @@
...
+`draft` `optional` `author:frbitten` 
+
+This NIP allows users to send binary data to be stored in the relay. This data can be small pieces of information or even large files.
+
+The intention is to combine the ease of NOSTR, with many clients being developed and a common communication protocol between all. To enable decentralized file sharing. NOSTR communication must work together with other protocols to get the best out of each technology. For lay users to be able to publish their photo, file, audio on several servers in a simple and easy way.
+
+Nostr event
+------------------
+This NIP specifies the use of the `1064` event type, having in `content` the binary data that you want to be stored in Base64 format.
+* `type` a string indicating the data type of the file. The MIME types format must be used (https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types)
```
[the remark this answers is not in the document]
AsaiToshiya: If kind `1065` is optional, `type` is needed. What do you think?

NfNitLoop:
Working on my experimental implementation for multi-part files, and had some questions come up for kind 1065:

* What's the difference between the `summary` tag, and the "description" which goes into the event's `content`? Might be worth calling that out.  (👍 for explicitly documenting `alt` is for accessibility.)
* Might be handy to have a `fileName` (`file_name`? `file`?) tag, to assist when folks want to download the file.

frbitten:
> Working on my experimental implementation for multi-part files, and had some questions come up for kind 1065:
> 
> * What's the difference between the `summary` tag, and the "description" which goes into the event's `content`? Might be worth calling that out.  (👍 for explicitly documenting `alt` is for accessibility.)
> * Might be handy to have a `name` (or: `fileName`, `file_name`, `file`?) tag, to assist when folks want to download the file. (I'm going to use `name`.)

Summary is used as if it were the file name. The concept of file name is not valid in this case because we do not have a file but a data buffer. That's why summary was used instead of filename. It is a very short description to easily identify the data.

For downloading, I think it's best to use the rule in NIP-96 that was created after this text. This maintains a standard already used.

For the relay to inform you where the download path is, we have two options:
- Use the NIP-96 format where there is a file with the settings at/.well-known/nostr/nip95.json.
- In NIP-11 there is a configuration option and limits for each relay functionality. But there is no such forecast today.

I would particularly prefer there to be a specific kind to access the operating properties of a relay. To keep everything within the nostr protocol and not have to use the HTTP protocol. But I'm not familiar with this format and needs to propose how it works.

haorendashu:
@vitorpamplona Hello, Vitor. I think the if NIP-95 image be quoted by other notes. It should display like other normal image not a quoted note.

Maybe the NIP-95 is better for backup and recover.  I am planning to add support for it.

vitorpamplona:
> @vitorpamplona Hello, Vitor. I think the if NIP-95 image be quoted by other notes. It should display like other normal image not a quoted note.

This is a possibility. Amethyst only displays them as quotes because we couldn't finish the new design to move the actions you can do to the image itself out of the usual quote interface, like react, boost, report, zap etc.

95.md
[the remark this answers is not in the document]
NfNitLoop: I think it's important to separate the metadadta (kind 1065) and the real data (kind 1064). To quote another part of this doc:

> Another defined event is the `1065` which is used as a header for the data contained in an event 1064. This way the data can be disclosed without overloading the communications when sending a large amount of data.

I don't need a "type" here if the "m" tag is present on 1065 -- it'll be redundant. And I'm not sure 1065 events should be considered optional. Clients should have a smaller 1065 metadata event available so that they can decide whether they want to download the (likely) larger 1064 event.  (Or, in the case of multi-part attachments, multiple events.)

staab:
This is better offloaded to blossom, closing due to inactivity.

PR #349, opened 2023-03-10
Add kinds 10 and 11 to prevent race conditions when updating contact lists

staab:
A very common experience on Nostr is that of "losing follows" due to race conditions when sending kind `3` events. Earlier this week someone signed in to Coracle, their contact list failed to fully sync before they followed someone, and they ended up deleting all their follows. The only solution a client can implement to avoid this currently is to disallow certain functionality if a person's account isn't fully synced, but of course, you don't know what you don't have, it's always possible a user hasn't yet followed anyone.

This change is backwards compatible, and simply introduces two new event kinds to solve the above problem. There's no reason to remove kind `3`, since there are valid use cases when a user might want to definitively say, "this is my contacts list".

The same problem exists for relay lists to a less severe degree (because relay list cardinality is lower), and for https://github.com/nostr-protocol/nips/pull/183, but I figured we'd have the conversation here first.

Sources:

nostr:nevent1qyt8wumn8ghj7etyv4hzumn0wd68ytnvv9hxgtcpz4mhxue69uhhyetvv9ujuerpd46hxtnfduhszythwden5te0dehhxarj9emkjmn99uq3kamnwvaz7tmjv4kxz7fwdaexzmn8v4cxjmrv9ejx2a30qy2hwumn8ghj7un9d3shjtn4w3ux7tnrdakj7qghwaehxw309aex2mrp0yhxummnw3ezu6twvehj7qgewaehxw309aex2mrp0yhxummnw3exzarf9e3k7mf0qyt8wumn8ghj7un9d3shjtnddaehgu3wwp6kytcqyr9yfkqe2yuky5lyl6wgexn6s6wjc9erntn9zp20j5cqdlk6qywsgnfgvxn)

### DIFF
--- a/02.md
+++ b/02.md
@@ -6,7 +6,7 @@ Follow List
 
 `final` `optional`
 
-A special event with kind `3`, meaning "follow list" is defined as having a list of `p` tags, one for each of the followed/known profiles one is following.
+Three special events having kinds `3`, `10`, and `11` are defined as having a list of `p` tags. Kind `3` defines the complete set of pubkeys one is following at a given point in time; kind `10` adds one or more pubkeys to that list; and kind `11` removes one or more pubkeys from that list.
 
 Each tag entry should contain the key for the profile, a relay URL where events from that key can be found (can be set to an empty string if not needed), and a local name (or "petname") for that profile (can also be set to an empty string or not provided), i.e., `["p", <32-bytes hex key>, <main relay URL>, <petname>]`. The `content` can be anything and should be ignored.
 
@@ -28,6 +28,8 @@ For example:
 Every new following list that gets published overwrites the past ones, so it should contain all entries. Relays and clients SHOULD delete past following lists as soon as they receive a new one.
 
 Whenever new follows are added to an existing list, clients SHOULD append them to the end of the list, so they are stored in chronological order.
+
+When building a contact list, clients should query for the latest kind `3` event, and any subsequent kind `10` and `11` events that modify the original list. Contacts SHOULD use kind `10` and `11` in order to prevent race conditions when updating contact lists from multiple clients simultaneously, unless the intention is to write a canonical, complete contact list.
 
 ## Uses
 
--- a/README.md
+++ b/README.md
@@ -90,6 +90,8 @@ They exist to document what may be implemented by [Nostr](https://github.com/nos
 | `6`           | Repost                     | [18](18.md)              |
 | `7`           | Reaction                   | [25](25.md)              |
 | `8`           | Badge Award                | [58](58.md)              |
+| `10`          | Follow Contact             | [2](02.md)               |
+| `11`          | Unfollow Contact           | [2](02.md)               |
 | `16`          | Generic Repost             | [18](18.md)              |
 | `40`          | Channel Creation           | [28](28.md)              |
 | `41`          | Channel Metadata           | [28](28.md)              |


staab:
One potential implementation difficulty is that if someone follows/unfollows a single user multiple times, when streaming the events back in out of order they'll have to keep track of when _each pubkey's_ follow status was last updated.

fiatjaf:
Interesting.

mikedilger:
I like to think of the ContactList sync problem fundamentally being that you create an event that doesn't account for past events you didn't know about.  This can be solved as suggested here with "diff" type events.

But most software that deals with merging things that have gone out of sync (git, palm OS sync, AvantGo's thing) has over time been done by storing different states, and computing differences from a common ancestor, and not storing the differences themselves. The reasons are subtle I think and have to do with missing diffs just like you might miss contact lists.

In any case, all we are missing from doing that kind of diff algorithm is the link back to the previous contact list.  If every contact list had a link back to the previous contact list that it thought it was replacing, we could find a common ancestor, do the diff algorithms, and merge without needing kinds 10 and 11, and I think it would avoid future issues with diffs themselves going missing.

Using the motivating  example of this PR:  User starts coracle and adds a new person to follow.  Coracle generates a ContactList event B with just 1 person, pointing back to NULL (somehow indicating it has no prior).  Then Coracle receives a contact list A pointing back to NULL that was created earlier.  This triggers it to create a merged contact list C which it then pushes out, and that merged list points back to both A and B.

mikedilger:
Having thought a bit more on this, I'm not sure about my idea. It could be better or worse and I can't decide. You would have to keep old ContactLists to find a common ancestor so it couldn't be replaceable meaning we'd need a new kind. And I'm not actually sure what the downsides of using patches (kind 10/11 as suggested) are, I just have a belief that the industry learned not to do that over time, but my understanding is not nuanced enough to really make any determination without doing some research.

monlovesmango:
in astral I just created a warning every time someone creates their first follow to confirm that they don't actually have any follows.

my gut feeling about this pr is that its is trying to solve an edge case that is only an issue bc of our lack of targeted relay strategy. as we adopt @mikedilger 's proposal for relay metadata lists this should really cease to be an issue. generally it feels wrong and overly complex to have 3 different kinds just for your follows list. 

without fixing the relay strategy first, we are likely to have the same issues we see with kind 3 with kind 10 and 11 as well. what in this proposal makes kind 10 and 11 more reliably retrievable than kind 3? and once relay strategy is more targeted, will kind 3 really be so unreliable to find that we still need kind 10 and 11?

if we do move forward on this pr, can we have more recommendations around when kind 3, 10, and 11 should actually be created? should all three be created with every contact list update? if only 10 and 11 should be updated regularly, when should we be updating kind 3? when kind 10 is initially created should it contain all existing follows from kind 3?

vitorpamplona:
I am in favor of killing the idea of a contact list altogether. I know this breaks current implementations and makes it heavier to assemble a user's home feed, but here's the thing: the list shouldn't even exist. It's not how people think about their follows. We are forcing a consensus (order of events) in the protocol that is not required. 

This is equivalent to Nostr not needing a blockchain: the consensus of an order of events is not required for a social protocol. 

Think about contacts as likes (kind 7). If you like a user, you are following it. If you delete the like, you stop following. Same for relay lists. Follows is a simple type filter with authors=me and my followers is a similar type filter, but with p == me. 

Then we also kill the aberration of a sub-10000 replaceable event.

monlovesmango:
at first glance, I really like this idea @vitorpamplona . making an individual follow event for each person you follow seems pretty straightforward and eliminates the possibility of wiping existing contact lists.

staab:
As I work more with it, I think Nostr is basically an event-sourcing database with domain objects implicitly layered on top. In this PR I started by trying to define new "edit operations" on our implicit "contact list" domain object because the current ones don't represent what the user wants to do. But we should be thinking about what the user wants to do and modeling those operations instead. I think @mikedilger's proposal is more in line with "editing implicit domain objects directly", and @vitorpamplona's observation gets at the heart of the issue (although I think there are use cases for saying "forget everything I said before, THIS is my contact list", so I don't see a reason to deprecate kind 3's).

I looked briefly into whether CRDTs would help with this problem, and I don't think they're really relevant, except to say that a list with independent set, add, and remove operations _is_ a CRDT (CRDTs are usually more complex because they deal with order or dependency of elements). The problem we currently have is that we're trying to map add/remove onto the set operation, resulting in apparent conflicts.

Some relevant conversation also exists on the NIP 65 discussion at https://github.com/nostr-protocol/nips/pull/218#issuecomment-1416823469 and following.

vitorpamplona:
> The problem we currently have is that we're trying to map add/remove onto the set operation, resulting in apparent conflicts.

The real problem is the thinking that there is a single contact list shared among the multiple relays a user has access to. You are assuming there is only ONE list to add/remove to. Which is incorrect. The majority of active users have outdated lists in old relays, either on purpose (because in that relay, they follow more people) or not.

fabianfabian:
@vitorpamplona no need to kill contact lists, they are useful for building a web of trust, they should be shared as a courtesy to others so they can be used as input against spam or other things. 

Seperate from that It's up to the clients to decide how they implement their following feed, using whatever is coming from the relays as the source of truth is probably not a good idea.

vitorpamplona:
> no need to kill contact lists, they are useful for building a web of trust

A list is not required to build the web of trust. Individual events to demonstrate follows work just as well as an ordered list. My point is simply that the following list doesn't need to be ordered, or packed together in the same event.

fiatjaf:
I like Vitor's idea, which is more-or-less what this PR proposes. We'll have to live with both approaches, but eventually one may win against the other.

Is there a difference between deleting a "follow" event and publishing an "unfollow" event though? What is better?

staab:
To delete an event, you must know its id. For lots of reasons, you might not have that id, but especially if the follow came from a kind 3 or 10002.

fiatjaf:
Wait, kind 10002 has follows too?

vitorpamplona:
> To delete an event, you must know its id. For lots of reasons, you might not have that id, but especially if the follow came from a kind 3 or 10002.

From a UI perspective, you have to know that the user is following another to hit unfollow. You will have to have the event.

staab:
> Wait, kind 10002 has follows too?

Sorry, I get confused between follow lists and relay lists, ignore that

> From a UI perspective, you have to know that the user is following another to hit unfollow. You will have to have the event.

If the follow came from a kind 3, you don't want to delete the entire list and re-create it. More broadly, I think event deletions are a mistake, because they introduce a dependency graph between events. They also don't represent a user's intention in many cases.

staab:
It occurred to me this morning that another use case this supports is allowing clients to show notifications when you are followed/unfollowed by someone. Currently, since 3's are replaceable, there's no way to do a diff without keeping a cache of kind 3 for _all_ users.

Egge7:
@staab can you give an example of how a kind 10 event would look like? If i understand this PR right, then clients would have to gather the latest kind3, and all kind 10 and 11 events that are older than the kind 3 in order to build a list, right? If I follow 2-5 keys per day, that would add up quite quickly, no?

staab:
I'm thinking something along the lines of: `{kind: 10, content: "", tags: [["p", <pubkey1>], ["p", <pubkey2>]]}`

> all kind 10 and 11 events that are older than the kind 3 in order to build a list

All 10 and 11 events that are _newer_, but I expect that's what you meant. Pulling a few thousand events doesn't seem that bad to me, especially since the alternative is pulling a single event with thousands of tags instead.

vitorpamplona:
Practically speaking this proposal's performance and kind-3s alone are quite the same. For reference, Amethyst takes 5-10 seconds to process 4000 kind 3s of your followers (to do the follower count). 

Processing 4000 individual reactions to do the same work happens in less than 0.5 seconds.

arthurfranca:
- kind 3 CL only
⬇ [bad] slower follows counting cause it is big
⬇ [bad] may have race condition problem (maybe cause of bad client logic maybe not)
⬇ [bad] unknown max followers due to relay event size limit
⬆ [good] faster for getting all followers for loading followers feed
⬆ [good] easy to migrate event to other relay

- 3 CL + 10 follow + 11 unfollow (by @staab)
⬇ [bad] slow follows counting cause it still uses CL along with extra events
⬇ [bad] legacy clients may still wipe CL;
⬇ [bad] slower for getting all followers, having to deal with follow+unfollow+follow_again_same_person
⬇ [bad] hard to migrate events to other relay (can get rate limited)
⬇ [bad] some relays may have incomplete set of events
⬇ [bad] not clear when to merge 10 and 11 into updated CL.
⬆ [good] may ease race condition problem
⬆ [good] "someone has followed you" notification
⬆ [good] "someone has stopped following you" notification

- kind X follow (by @vitorpamplona)
⬇ [bad] slower followers list building/counting but faster than 3+10+11 kinds
⬇ [bad] clients may ignore and not migrate out from CL
⬇ [bad] hard to migrate events to other relay (can get rate limited)
⬇ [bad] some relays may have incomplete set or events that should be already deleted
⬇ [bad] some old events may get deleted by relay policy since they won't be refreshed like CL updates
⬆ [good] no race condition
⬆ [good] faster follows counting
⬆ [good] "someone has followed you" notification

Now I think kind 3 CL isn't that bad xD. Maybe @staab idea should be just a client-side logic between CL edit moments.

As an addition, we could keep kind 3 CL and just add follow events with expiration 3 days ahead with the sole purpose of notifying like "someone has followed you". Or maybe it's not worth it.

staab:
That list seems pretty unfair tbh, most of those downsides on my proposal are marginal performance differences, general criticisms of nostr itself (relays may have incomplete set of events), or restatements of how kind 3 is already broken and the proposal doesn't automatically fix kind 3. Whereas the importance of the race condition is understated, it's one of the most common and annoying problems people experience with nostr.

vitorpamplona:
Additional point, after loading up all contact lists, if your client just wants to listen to contact lists updates (to keep things up to date) the amount of events is significant and they are all super large events. If we switch to a lighter model, clients could keep listening and just receiving the new follows/unfollows as opposed to an entire list again. 

Keep in mind Battery life and networking use (most people will use NOSTR in their phones, on mobile data plans).

staab:
It occurs to me a difficulty with this is counting followers, since if these aren't replaceable events, there can be duplicates. A "d" tag would probably fix this.

mikedilger:
While there are these other reasons to not have monolithic full contact list events (and those battles need to continue to be fought), I've never been convinced that the current state of affairs actually creates race condition problems of any real significance. I think if you do the following you should be good (and I must admit gossip isn't doing this currently):

1. Clients need to remember the date of the contact list they are using, and never update from an older contact list than what they already have.  They may still go from very old to old, but that's better than going from newest to older.
2. That leaves the problem of clients editing and writing back from an old contact list and thus clobbering the newer one. But that can't happen if they first read from the relay they intend to write to (except for rapid race condition which is quite unlikely). If they find a newer contact list there, they should use that one and merge as necessary.  They may write back a newer contact list to the relay that isn't the newest of all, but only to relays that already had old lists so it shouldn't actually clobber anything important.

alexgleason:
Kind 3 events can have an `e` tag pointing to the previous kind 3 event the new one modifies. This would allow clients to detect forks and create a new merged event if so.

Egge21M:
> Kind 3 events can have an `e` tag pointing to the previous kind 3 event the new one modifies. This would allow clients to detect forks and create a new merged event if so.

This would require a change in NIP-02 and an upgrade of all clients and relays to work, as as of right now, those are not expected to keep old kind-3 events

staab:
> Kind 3 events can have an e tag pointing to the previous kind 3 event the new one modifies. This would allow clients to detect forks and create a new merged event if so.

I don't think so, because 3 is a replaceable event. So you'd need to keep all copies around, and if you were missing one you might not be able to build the tree.

alexgleason:
The client stores your events in its cache, so when a new kind 3 replaces an old one in its cache you will have access to both in that moment.

If the new one refers to the old one, you can replace it. Otherwise you should refetch from relays and wait a moment to ensure you have the most recent version.

Clients are usually 1 installation for 1 person, so you can keep a history of kind 3 events for the user on the client to improve resolution if you want.

Btw instead of an `e` tag pointing to the old event version, it should be a `prev` tag so it can work in any NIP-51 list, including bookmarks and pins (which already have `e` tags).

This shouldn't anything break except naive implementations that fail if any kind besides `p` exists in the tag array.

Egge21M:
> The client stores your events in its cache, so when a new kind 3 replaces an old one in its cache you will have access to both in that moment.
> 
> 
> 
> If the new one refers to the old one, you can replace it. Otherwise you should refetch from relays and wait a moment to ensure you have the most recent version.

What do clients do if they receive a new kind-3 event, that they don't have the previous one too, and the relays doesn't have it either? Also how would clients handle a new kind-3 event without any e/prev tag?

erskingardner:
> What do clients do if they receive a new kind-3 event, that they don't have the previous one too, and the relays doesn't have it either? Also how would clients handle a new kind-3 event without any e/prev tag?

They can do whatever they want. That's the point. Clients decide how to handle situations like this.

> Btw instead of an e tag pointing to the old event version, it should be a prev tag so it can work in any NIP-51 list, including bookmarks and pins (which already have e tags).

I like this. `prev` tag also doesn't get indexed so doesn't add much, if any, overhead to relays. Obviously, again, it's up to clients to handle this in more complex ways if they want. The naive approach of just accepting the newest event as the current is still fine in most cases.

Egge21M:
> The naive approach of just accepting the newest event as the current is still fine in most cases. 
> 
> 

This is what I do not like about this approach. It does not fix the issue at hand. If a client does not have the latest state, it will overwrite the old state, and once relays receive the new event, they will also drop and overwrite. Nothing is gained vs. regular caching of kind-3 events and chaining by timestamp.

mikedilger:
Is this is a fundamentally hard problem? Do we need some caffeinated University graduate students in the prime of their mental capacity working on it under a research grant? None of the solutions proposed actually work.

My mental capacity has suffered recently due to unknown factors (I think perhaps it is sleep apnea and am getting tested soon).

alexgleason:
This is one of those weird cases where putting data on a blockchain-like structure (really just a linked-list) would kind of make sense.

alltheseas:
@jb55 how do we fix broken contact lists via NIP

jb55:
On Mon, Jan 29, 2024 at 08:42:37AM -0800, alltheseas wrote:
>@jb55 how do we fix broken contact lists via NIP

I'm planning on resolving this by creating a delta-contact list format,
so each change can be represented by "diff" of the previous one you have
stored. This will allow us to show every change in a compact way, very
much like "git diff" for your contact list, and it's backwards
compatible with everything. Then you will be able to review and rollback
to any previous revision, perhaps damus can do this automatically if it
sees large drops of contacts at once.

jb55:
> My original proposal has the same problem, it just seems to me that establishing dependencies between events only adds complexity without solving any problems.

all I can do is show how the new list is different from the local version, it will not be a consistent history like git.

staab:
> all I can do is show how the new list is different from the local version, it will not be a consistent history like git.

So is it essentially the same as this PR? Or is there some significant difference?

jb55:
it doesn't require any changes to the protocol

staab:
I see, you're just talking an implementation detail to detect large deltas?

jb55:
yeah I was thinking it could just be an implementation detail of the contact list storage. the biggest issue is that contact lists are huge, so most people don't store each revision. if they were smaller it would be more viable: start with a base contact list and store deltas.

staab:
I can see how that would be useful, but not how it helps with the race condition problem.

mikedilger:
> On Mon, Jan 29, 2024 at 08:42:37AM -0800, alltheseas wrote: @jb55 how do we fix broken contact lists via NIP
> I'm planning on resolving this by creating a delta-contact list format, so each change can be represented by "diff" of the previous one you have stored. This will allow us to show every change in a compact way, very much like "git diff" for your contact list, and it's backwards compatible with everything. Then you will be able to review and rollback to any previous revision, perhaps damus can do this automatically if it sees large drops of contacts at once.

How about two new event kinds:  One is the full list, one is a diff.  A diff points back to one event of either kind.  A full list points back to 1 or more events of either kind.  Full lists can consolidate long sequences of diffs, or merge multiple divergent histories.  You could have a full list point at a full list and clients can compute the diff... or you can publish a diff and clients can compute the full list.  I think having both of these has benefits.

jb55:
On Mon, Jan 29, 2024 at 03:43:36PM -0800, Michael Dilger wrote:
>> On Mon, Jan 29, 2024 at 08:42:37AM -0800, alltheseas wrote: @jb55 how do we fix broken contact lists via NIP
>> I'm planning on resolving this by creating a delta-contact list
>> format, so each change can be represented by "diff" of the previous
>> one you have stored. This will allow us to show every change in a
>> compact way, very much like "git diff" for your contact list, and
>> it's backwards compatible with everything. Then you will be able to
>> review and rollback to any previous revision, perhaps damus can do
>> this automatically if it sees large drops of contacts at once.
>
>How about two new event kinds:  One is the full list, one is a diff.  A
>diff points back to one event of either kind.  A full list points back
>to 1 or more events of either kind.  Full lists can consolidate long
>sequences of diffs, or merge multiple divergent histories.  You could
>have a full list point at a full list and clients can compute the
>diff... or you can publish a diff and clients can compute the full
>list.  I think having both of these has benefits.

I wasn't thinking about publishing these, I'm not sure how useful they
would be to the wider network. I was brainstorming some "delta-notes"
that could be stored as a note itself. It might not be the most
efficient so I'm not sure I'll do it that way yet.

jb55:
On Mon, Jan 29, 2024 at 03:43:25PM -0800, hodlbod wrote:
>I can see how that would be useful, but not how it helps with the race condition problem.

Indeed, It's more of a way to revert when things break instead of fixing the underlying problem.

alexgleason:
@staab This idea is still legit I think. But now it's invalid because [NIP-29](https://github.com/nostr-protocol/nips/blob/master/29.md) reserved kinds 10 and 11.

alexgleason:
@staab I wonder if it should just be labels. You can label people you follow.

mleku:
this seems like a logical solution for follow list hell and makes a path towards sanity, ack, pls resolve conflicten

jb55 (commented):
event sourcing is the way. this is simple and is easy to pull incrementally.

I suspect it would be still be buggy in the context of other clients updating kind3 without knowledge of 10,11s ?

pablof7z:
I have a similar proposal for kind-scoped follows, which can also be applied to non-kind-scoped follows. The idea is to create a follow event of kind 967 whenever you follow someone (or multiple people) and delete that event when you unfollow them.

The reason I prefer this over the deltas approach is that deltas require reconstructing the full state to determine the correct current state (not easy in nostr). The other approach allows for operating with a partial state.

https://github.com/nostr-protocol/nips/pull/1605

mikedilger (changes requested):
This proposal seems to describe an operation-based conflict-free replicated data type representing an LWW-Element-Set (see https://en.wikipedia.org/wiki/Conflict-free_replicated_data_type).  After taking time out to actually think about this issue, I think this is the right solution, modulo a small detail.

It could be that instead of two kinds 10 and 11, there is only a single kind, and instead two kinds of tags: "add" and "remove".  I don't think they need to be searchable tags.

But that isn't terribly important.

fiatjaf:
So if I add someone and remove and add and remove a thousand times that means people will have to download a thousand events in order to figure out who am I following?

staab:
The only practical way to use this approach is probably to use it only when you're not sure you have the most recent kind 3, need to do a conflict-free follow/unfollow for some reason, or want to reduce the number of times you're writing a truly massive follow list. In other words, these events should probably be temporary and expire after a few days/weeks, and be periodically collected and merged into a kind 3 list to get a "snapshot". Once that's done, previous events can be dropped. I'm not sure the complexity is worth it, but I'd be interested to hear @mikedilger elaborate on why he thinks it's the best solution.

staab:
> They are inherently tied to the context and purpose of the client being used.

The cost of this is fragmenting the social graph, which is one of the biggest selling points of nostr. I don't disagree there are different reasons you might follow someone, but being overly monolithic with follows and fixing it with lists is a better default than being over granular with follows and having to start over with every new client.

vitorpamplona:
> The cost of this is fragmenting the social graph

The social graph is still the same. It's just more appropriately labeled. It's indeed more costly but it is also better for UX.

We are stuck in the kind 1 mindset where the whole of Nostr is just kind1s. That does not work anymore. It only worked because Nostr was just kind 1s until last year. As soon as we started adding other event kinds, follow lists immediately lost all the sense of it. 

I favor having all CONTACTS (not follows) in separate events, like on https://github.com/nostr-protocol/nips/pull/761. When the user installs a new client, the user can pick which labels he/she wants to import from. We just need to specify the context used for the `n` tags that each client will have. There won't be a single `follow` tag.

mikedilger:
> So if I add someone and remove and add and remove a thousand times that means people will have to download a thousand events just in order to figure out if I am following that person or not?

If you wanted to do that... yes.  To prevent this from happening incidentally, people would publish kind-3's that aggregate the prior change set now and then.
 
> And then a relay cannot purge old events anymore because that will destroy the state computation? So it has to keep many thousands of events on my behalf?

It could purge. Embrace the chaos!

fiatjaf:
> I favor having all CONTACTS (not follows) in separate events, like on #761. When the user installs a new client, the user can pick which labels he/she wants to import from. We just need to specify the context used for the `n` tags that each client will have. There won't be a single `follow` tag.

I agree with the goals here, and in theory I like the relationship status NIP very much, but I am afraid in practice fetching a big list will be so much better.

If we are going to standardize different "contexts" for follow lists already, why not make each of these a different NIP-51 list?

mikedilger:
> I'm not sure the complexity is worth it,

Even Jon doesn't defend his own proposal, so I'm alone in this?

> The notion of a "general follow/unfollow" no longer holds true in today's Clients.

That is true if you are talking about what you see in your feed.  But without it, gossip couldn't do it's friends-of-friends calculation.

>  I'd be interested to hear @mikedilger elaborate on why he thinks it's the best solution.

I may have misjudged this PR as an implementation of LWW-Element-Set.  I think that thing has two sets, an add set and a remove set, and each entry has a timestamp (when added or removed).  A dumb leap of logic made me think each event could be thought of as a line in the set, as each event is timestamped... but I think each event would have to communicate both entire sets else it becomes an operations-based CRDT instead of a state-based one (which needs reliable messaging that we don't have).

So I was wrong.

vitorpamplona:
> If we are going to standardize different "contexts" for follow lists already, why not make each of these a different NIP-51 list?

That's how we do it today, but: 

1. Race conditions are common
2. Relays block bigger lists due to event size limits
3. Small changes are costly: tag table update and reindexing.
4. Wastes space with duplicate pub keys in lists.
5. Slow lookups: Checking a user’s participation in a list requires comparing it against pubkeys in all NIP-51 lists.
6. Poor UX: Users think in terms of labels when visiting profiles, not lists, making management unintuitive.

fiatjaf:
> in practice fetching a big list will be so much better

For example, most relays nowadays limit stored events query results to 500 events by default, and I think that is a pretty generous limit. But for most users who follow more than that this will introduce unnecessary complications when fetching lists of follows.

Not to mention all the extra signature verifications and event JSON overhead, but these I think are ok to disregard.

mikedilger:
Forget this one. I pushed a new PR.

vitorpamplona:
> most relays nowadays limit stored events query results to 500 events by default

Working around these limits is better than the current straight-up rejection of big lists.

Most clients have to go around the 500 events already simply to get all reactions/zaps/boots/replies of popular kind1 notes.

vitorpamplona:
Relays will always store all the garbage users want. If they dont store as one list, we will break it down into multiple events which might be worse for relays. 

But the "garbage" WILL be there.

fiatjaf:
> Relays will always store all the garbage users want. If they dont store as one list, we will break it down into multiple events which might be worse for relays.

This is so incredibly dumb and short-sighted. If you try to force relays and remove their choice you will just eventually exclude those the smaller ones, the community relays, the more idealistic, lower-budget people, the cheap relays that would provide a good service and add to the decentralization -- and you'll leave the ecosystem to be dominated only by the megacorps or whoever has other secondary interests and may accept that storing garbage is a reasonable cost for them.

If relays prefer lists lists or not I don't know, this wasn't even the topic, but the way you're approaching this question is certainly wrong.

fiatjaf:
> Forget this one. I pushed a new PR.

This is @mikedilger's new PR, for the record: https://github.com/nostr-protocol/nips/pull/1630

It agains chooses to ignore the concerns I raised here.

@pablof7z approach in https://github.com/nostr-protocol/nips/pull/1605 with deletes is much better than these attempts to create a consistent chain of events that assume a globally always-accessible state in Nostr, although lists are still probably better.

vitorpamplona:
> This is so incredibly dumb and short-sighted.

Most clients I have looked into have coded ways around many different kinds of relay limits. They are doing the same thing, just via other means. Some Nostr libraries provide those things for free. My point is not about how dumb or smart this is. I am just saying that it IS happening right now and will continue to happen because users want it. 

It's the same debate about binary images on relays. Relays kept rejecting NIP-95, so users just added base64 directly on Kind 1s. Done. Now base64 images are everywhere. And if relays start filtering by content, they will change the content signature to go around it. It's a dumb cat and mice race. 

Some relay limits work. Others just make clients go around to do the same things via other means. The ones people can go around are dumb and just make Nostr more difficult to code without solving the problem for the relay operator. List size limits are one of them.

fiatjaf:
> I am just saying that it IS happening right now and will continue to happen because users want it.

What kind of rationale is that? It is happening now therefore it will always happen?

I am making a statement about how I see future developments and consequences of actions, you're just describing what is happening, predicting that it will continue to happen and defending it as the best possible thing just because you saw it happening today.

> In practice, many of these limits don't achieve what relay operators want to achieve and are just making things harder.

I agree, but relays will eventually find ways to ban everything as real attackers come, so it's pointless to keep pushing them now by acting as an attacker that fakes as a normal client, it's confusing for everybody.

vitorpamplona:
> What kind of rationale is that? 

I am a very pragmatic person. I don't care what you or I think is going to happen. What our preferred solution is. I only care about things that DO happen. Wishful thinking is not for me. I am also not defending anything. My personal opinion is irrelevant in the face of reality. 

> I am making a statement about how I see future developments and consequences of actions

My point in bringing that up is that your statement is misaligned with what I see in reality. You are asking clients to play nice and not abuse relays out there. Which is the complete opposite of what we have seen for the past 2 years. It's wishful thinking. 

Users want 10,000 people follow lists. Corporates want 1M+ people follow lists. Saying things like: "Rejection of big lists is a good measure" or assuming big lists are a "ton of garbage" or that these users are "reckless people" doesn't work in your favor. No client today is "helping users keep smaller lists". Absolutely NO ONE. In fact, it is common for those users to abandon clients that even suggest that they should reduce their follow lists. 

I know because I tried multiple times. No user is happy when the client is trying to control how many follows they can have. 

And they are right.

fiatjaf:
OK, this discussion is pointless.

But let me just say that you are living in a big contradiction because not caring about the future is not "pragmatism" and if you really only cared about things that "DO happen" you would never DO anything, because in order to DO anything you have to make a plan, even if it's a plan for the next 2 minutes, and a plan involves thinking about the future and thinking about what you will DO -- and anything you haven't DONE yet hasn't yet happened, therefore it is outside the realm of what you just said you can conceive.

But I don't expect you to understand this.

vitorpamplona:
I agree. It's pointless.

Issue #107, opened 2022-12-22
NIP-04 considered harmful


adiabat:
Hi - just nostr seems promising, and end to end encrypted communication is a very important part of it, but the NIP-04 spec as written should not be implemented.  

There is another issue here: https://github.com/nostr-protocol/nips/issues/72 pointing out the non-uniform AES key.  That issue has been closed but probably shouldn't be -- the spec still has the unhashed / truncated result of the DH as the AES key.

Another, I think more severe problem is that protocol as described uses aes-256-cbc with no message authentication.  This means that messages can be undetectably altered in transit.  Anyone relaying the message can change the message, and the receiver can't tell that it was changed.  I would remove NIP-04 before people start trying to use it.

There are other encryption protocols that are used with the secp256k1 keys used in nostr that could be adapted for direct messages, such as BIP324 (https://github.com/bitcoin/bips/pull/1378) or lightning's bolt 8 (https://github.com/lightning/bolts/blob/master/08-transport.md).  There are also ratcheting protocols which have forward secrecy like the one used by signal, but those have other trade-offs.

The BIP324 and bolt 8 protocols aren't for discrete messages; they are for encrypted & authenticated communication channels.  That means it doesn't directly replace NIP-04; the bolt 4 (https://github.com/lightning/bolts/blob/master/04-onion-routing.md) onion messages would be a closer fit.  But encrypted transport between nodes is also important.

The code from BIP324 and LN is available in different languages and open licenses so I think that's the best bet for getting some code that's been used and tested.

jonasschnelli:
I have to agree with @adiabat here. NIP04 is potentially harmful.

However, @fiatjaf's comment on the non-uniform AES key issue seems to be a reasonable approach. There is just the risk that people start to use NIP04 with sensitive information without knowing the downsides.

> By the way NIP-04 was never supposed to be the thing in the first place, it was made as a prototype to show it was possible and expecting someone to supersede it with a better protocol (I personally think we must still find some other more p2p more private way to send direct messages between Nostr contacts that doesn't involve broadcasting to relays).

IMO NIP04 should at least state its infancy and thus risks of metadata exposure (who sent to who, when and how many bytes, etc.), missing message authentication (alteration by relays) and potential vulnerable crypto (no key derivation after DH). I would even go a step further and put a paragraph into NIP04 that clients implementing is, MUST warn users of the its infancy and risks.

NIP04 is probably *not* production ready which can lead to harm,... especially with faster adoption of nostr in general.

[...]

PR #351, opened 2023-03-12
NIP-59: improved direct messages

davestgermain:
This NIP borrows the ideas from the SimpleX Messaging Protocol in order to obscure the metadata of encrypted direct messages. It adds an initial negotiation process, allowing for encrypted, private communication across multiple relays.

### DIFF
--- a/59.md
+++ b/59.md
@@ -31,7 +31,9 @@ Her client will create a request that contains pointers to a relay and event kin
 }
 ```
 
-The `content` is a random token which will be used to create a "conversation key" as follows: `sha256((token + bob_public_key)`
+The `content` is a random token which will be used to create a "conversation key" as follows: 
+
+`sha256(token + bob_public_key)`
 
 Next, the client will generate an ephemeral private key and create a standard kind=4 DM, with the content being the jsonified request above:
 ```json
@@ -58,7 +60,9 @@ Next, the client will generate an ephemeral private key and create a standard ki
 
 Only the recipient metadata (Bob's pubkey) is revealed to the public, in this first message of the negotiation.
 
-Bob's client will see a DM from an unknown contact, will decrypt and validate that the request is addressed to him (the "p" tag is his public key). The client should present Bob with a prompt, displaying the sender's information, the requested relay and kind. If the requested kind is an ephemeral event (as shown in this example), the client can indicate that the chat with Alice would (may?) not be saved on the relay.
+Bob's client will see a DM from an unknown contact, will decrypt and validate that the request is addressed to him (the "p" tag is his public key). 
+
+The client should present Bob with a prompt, displaying the Alice's information, the requested relay and kind. If the requested kind is an ephemeral event (as shown in this example), the client can indicate that the chat with Alice would (may?) not be saved on the relay.
 
 If Bob accepts the request, his client would first create another ephemeral private key and repeat the process described above, with his own conversation request to Alice, pointed to a relay of his choosing:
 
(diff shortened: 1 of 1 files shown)


fyookball:
Improving direction messaging is a good idea. I like the general direction, but I think it could be improved by only having either Bob or Alice send a private encrypted message, not both of them. You could do it by having Alice create ephemeral key A2, which informs B1 that he should create B2 to talk to A3. Also, there is low hanging fruit in terms of fixing issues with NIP-04 that are considered "harmful".  https://github.com/nostr-protocol/nips/issues/107

davestgermain:
> I think it could be improved by only having either Bob or Alice send a private encrypted message, not both of them. You could do it by having Alice create ephemeral key A2, which informs B1 that he should create B2 to talk to A3. 

Thanks for pointing to that thread. Some of the ideas seem similar to this proposal, if I'm reading correctly. (I am also not a cryptographer. I think ease of implementation should be a consideration here)

Maybe it's not clear in the description, but only one kind=4 event is ever sent. Both parties send an encrypted "request event", but Bob sends his to Alice's A2 at Alice's relay.

[Here's a program](https://gist.github.com/davestgermain/12974fef590fc1edf12e9a25b9dfa7b5) that demonstrates the negotiation.

> Also, there is low hanging fruit in terms of fixing issues with NIP-04 that are considered "harmful". #107

As far as changing the encryption method, that's beyond my pay grade. There were a lot of ideas in that ticket thread, but somebody's got to pick one. Using nostr events gives you a lot "for free" (for instance, I'm using ephemeral events and the expiration tag in this proposal). My preference is to leverage the existing tools rather than introduce totally new technologies to the stack.

One possible change could be, instead of specifying a relay to connect to, Alice could send an identifier as specified in [NIP-39](https://github.com/nostr-protocol/nips/blob/master/39.md), to recommend connecting to telegram, signal, etc to continue the conversation.

shafemtol:
See also:
- #303 
- my WIP PRs #306 and #333, which are meant as building blocks for a later NIP for DMs
- #56, which to me looks rather similar to this proposal

mikedilger:
As a general comment (reserving the ability to comment on the method later on) you cannot change the meaning of kind=4.  If you change how the process works it must be a new event kind.  Nostr must always be backwards compatible and clients that do it the old way must continue to work.

As the proposal stands, Bob at key 18d is going to receive a jsonified request and his client will render it like it was a DM and Bob is going to say WTF?

no-prob:
See also:
NIP-31 Incognito Direct Messages https://github.com/nostr-protocol/nips/pull/410/

vishalxl:
> Bob's client will see a DM from an unknown contact, will decrypt and validate that the request is addressed to him (the "p" tag is his public key).

The initial invitation, coming from an unknown contact, is a weak point here, from the angle that any well known public key could be spammed by random spammers, and then the recipient's client could ignore messages from unknown keys under attack from spammers.

If this initial invitation could be sent from a well-known pubkey ( of Alice) to Bob's, then there is more meta-data leak ( that Alice sent Bob an invite), but that's it, and after that there should not be a way to know whether Bob replied to Alice or not; that could be done to and may be considered.

ntheden:
> The initial invitation, coming from an unknown contact, is a weak point here, from the angle that any well known public key could be spammed by random spammers

Is it reasonable to spend resources decrypting potential spam? Because if it is, then the contents of the invitation can include a signature from Alice of the message metadata proving it's from her, and then meta data is not leaked.

staab:
Solved by nip 17

Issue #354, opened 2023-03-12
JSON serialization for deriving `event.id` is not adequately specified


shafemtol:
Correct derivation of `event.id` is critical for correctly referencing, signing and verifying events. NIP-01 currently specifies the derivation of `event.id` as follows:

> To obtain the `event.id`, we `sha256` the serialized event. The serialization is done over the UTF-8 JSON-serialized string (with no white space or line breaks) of the following structure:
> 
> ```json
> [
>   0,
>   <pubkey, as a (lowercase) hex string>,
>   <created_at, as a number>,
>   <kind, as a number>,
>   <tags, as an array of arrays of non-null strings>,
>   <content, as a string>
> ]
> ```

JSON itself does _not_ specify a canonical way to serialize strings and numbers. That means the above specification is ambiguous and can lead to mutually incompatible implementations.

Excluding control characters, `\` and `"`, JSON strings can directly represent any Unicode codepoint as itself. JSON provides some special escape sequences like `\n`, `\\`, `\"` and `\/`(!). In addition it provides `\u`, which can be used to represent *any* UTF-16(!) codepoint. A newline can be represented as either `\n` or, in theory, `\u000a` or `\u000A`. The letter ß can be represented as `ß`, `\u00df`, `\u00DF` or even something as silly as `\u00dF`.  Some JSON implementations will escape any non-ASCII character by default, others might not. Some [escape the forward slash](https://stackoverflow.com/questions/1580647/json-why-are-forward-slashes-escaped) by default, others do not.

The mention of "UTF-8 JSON-serialized" _can_ be understood to mean that non-ASCII characters MUST NOT be escaped, but this is not clear.

I see two different ways to deal with the incompleteness of the current specification:
- Encode strings as [RFC 3629](https://datatracker.ietf.org/doc/html/rfc3629) UTF-8, do not escape characters in strings unless absolutely needed (`\`, `"`, control characters), and use the special, short escape sequences where available (e.g. `\n` instead of `\u000a`). For `\u` escape sequences I would assume lower case simply from two implementations I've tested, but this _really_ has to be specified. Represent integers without any fraction or exponent.
- Represent the values exactly as they are represented in the event object itself. This would make implementation easier when _generating_ events and should be backwards compatible with all existing events, but it would complicate implementation everywhere else.

[...]

PR #355, opened 2023-03-12
NIP-01: Specify canonical JSON for id derivation

shafemtol:
Minimal changes for addressing #354.

I haven't included any specifics of JCS in NIP-01. I think it might be better to just refer to the spec rather than include parts of it, so that implementors actually go to the referenced specification and have a lesser chance of missing important details.

As mentioned in #354, all web clients most likely already conform to JCS as used here.

### DIFF
--- a/01.md
+++ b/01.md
@@ -30,7 +30,9 @@ The only object type that exists is the `event`, which has the following format 
 }
 ```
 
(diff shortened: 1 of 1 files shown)


vitorpamplona:
Since this, we have refined the definition of id derivation on NIP-01. Hope that helped fix all the issues.

PR #360, opened 2023-03-13
NIP-04: fix bug in code sample

fcked:
The code sample assumed that a `Uint8Array` was actually a hex string.


fiatjaf:
Can you also change the import from `noble-secp256k1` to `@noble/secp256k1`? That old version returned hex strings indeed, only the new versions return UInt8Array.

PR #363, opened 2023-03-14
NIP-100: WebRTC Signaling over nostr

jacany:
This NIP allows for nostr ephermeral kind `25050` to act as a signaling layer for Peer-to-peer WebRTC connections

**Potential Uses:**
- Video Conferencing App
- [Nostr Nests](https://nostrnests.com) over nostr
- Discord-like app
- Direct Phone Calls

What's next for this NIP is making connection data securely encrypted, this ensures that nobody can peek into who you're communicating with, or see your connection addresses.

### DIFF
--- a/100.md
+++ b/100.md
@@ -1,8 +1,10 @@
-# NIP-100
-## WebRTC
+# NIP-100 WebRTC
 `draft` `optional` `author:jacany`
 
 This NIP defines how to do WebRTC communication over nostr
+
+## Defining Rooms
+Rooms are essentially the space that you will be connected to. This must be defined by a `r` tag following the pubkey of the person you are connected to or some other identifier (maybe a lobby id for a joinable voice channel, etc.)
 
 ### Broadcasting that you are present
 Announces that you are here, and ready to connect to others.
@@ -12,7 +14,8 @@ The connection ID is inferred from the provided pubkey.
     "kind": 25050,
     "pubkey": "<sender pubkey>",
     "tags": [
-        ["type", "connect"]
+        ["type", "connect"],
+        ["r", "<room id>"]
     ]
 }
 ```
@@ -23,7 +26,8 @@ The connection ID is inferred from the provided pubkey.
     "kind": 25050,
     "pubkey": "<sender pubkey>",
     "tags": [
-        ["type", "disconnect"]
+        ["type", "disconnect"],
+        ["r", "<room id>"]
     ]
 }
 ```
@@ -36,7 +40,8 @@ Used when responding to a `type:connect`. Offering to connect to that peer.
     "pubkey": "<sender pubkey>",
     "tags": [
         ["type", "offer"],
-        ["p", "<reciever pubkey>"]
+        ["p", "<reciever pubkey>"],
+        ["r", "<room id>"]
     ],
     "content": {
         "offer": "<offer>",
@@ -52,7 +57,8 @@ Used when responding to a `type:connect`. Offering to connect to that peer.
     "pubkey": "<sender pubkey>",
     "tags": [
         ["type", "answer"],
-        ["p", "<reciever pubkey>"]
+        ["p", "<reciever pubkey>"],
+        ["r", "<room id>"]
     ],
     "content": {
         "sdp": "<sdp>",
@@ -68,7 +74,8 @@ Used when responding to a `type:connect`. Offering to connect to that peer.
     "pubkey": "<sender pubkey>",
     "tags": [
         ["type", "answer"],
-        ["p", "<reciever pubkey>"]
+        ["p", "<reciever pubkey>"],
+        ["r", "<room id>"]
     ],
     "content": {
         "candidate": "<sdp>",
--- a/100.md
+++ b/100.md
@@ -1,0 +1,80 @@
+# NIP-100
+## WebRTC
+`draft` `optional` `author:jacany`
+
+This NIP defines how to do WebRTC communication over nostr
+
+### Broadcasting that you are present
+Announces that you are here, and ready to connect to others.
+The connection ID is inferred from the provided pubkey.
+```json
+{
+    "kind": 25050,
+    "pubkey": "<sender pubkey>",
+    "tags": [
+        ["type", "connect"]
+    ]
+}
+```
+
+### Closing
+```json
+{
+    "kind": 25050,
+    "pubkey": "<sender pubkey>",
+    "tags": [
+        ["type", "disconnect"]
+    ]
+}
+```
+
+### Offering to connect
+Used when responding to a `type:connect`. Offering to connect to that peer.
+```json
+{
+    "kind": 25050,
+    "pubkey": "<sender pubkey>",
+    "tags": [
+        ["type", "offer"],
+        ["p", "<reciever pubkey>"]
+    ],
+    "content": {
+        "offer": "<offer>",
+        "type": "offer"
+    }
+}
+```
+
+### Answering an Offer
+```json
+{
+    "kind": 25050,
+    "pubkey": "<sender pubkey>",
+    "tags": [
+        ["type", "answer"],
+        ["p", "<reciever pubkey>"]
+    ],
+    "content": {
+        "sdp": "<sdp>",
+        "type": "answer"
+    }
+}
+```
+
+### Broadcasting ICE Candidate
+```json
+{
+    "kind": 25050,
+    "pubkey": "<sender pubkey>",
+    "tags": [
+        ["type", "answer"],
+        ["p", "<reciever pubkey>"]
+    ],
+    "content": {
+        "candidate": "<sdp>",
+        <misc>
+    }
+}
+```
+
+Essentially, just directly feed what comes out of the WebRTC functions straight into the `content` field. It makes things cleaner and easier.


100.md
```diff
@@ -0,0 +1,95 @@
...
+
+### Offering to connect
+Used when responding to a `type:connect`. Offering to connect to that peer.
+```json
+{
+    "kind": 25050,
+    "pubkey": "<sender pubkey>",
+    "tags": [
+        ["type", "offer"],
+        ["p", "<reciever pubkey>"],
```
fiatjaf, line 53: receiver

100.md
```diff
@@ -0,0 +1,95 @@
...
+```
+
+### Answering an Offer
+```json
+{
+    "kind": 25050,
+    "pubkey": "<sender pubkey>",
+    "tags": [
+        ["type", "answer"],
+        ["p", "<reciever pubkey>"],
```
fiatjaf, line 70: receiver

100.md
```diff
@@ -0,0 +1,95 @@
...
+```
+
+### Broadcasting ICE Candidate
+```json
+{
+    "kind": 25050,
+    "pubkey": "<sender pubkey>",
+    "tags": [
+        ["type", "candidate"],
+        ["p", "<reciever pubkey>"],
```
fiatjaf, line 87: receiver

fiatjaf:
This is fine, but I wonder how much it is tailored to a single application but it is posing as a generic thing without actually being generic.

jacany:
> This is fine, but I wonder how much it is tailored to a single application but it is posing as a generic thing without actually being generic.

Any changes that would make it more generic?

ArcadeCityMayor:
Concept ACK! We've been talking about WebRTC signaling via Nostr in the TG channel here and there for a year or so, great to see it formalized.

fiatjaf:
> Any changes that would make it more generic?

I don't have any changes to suggest, I don't know if it should be more generic. Maybe it should be more specific. A NIP for the entire concept of app calls, for example, would make sense and allow calls apps to be interoperable between them. That would make more sense than a pseudo-generic "webrtc NIP". I am just throwing ideas.

Velua:
How do clients learn of other client ids / public keys? There's a broadcast message, so does that get pushed to the other clients in the room then? And the relay tracks who is in each room?

jacany:
Broadcast message is listened to by clients subbed to that “room", once they hear it they know that’s a new client to connect to.
Discovery of these “rooms" is not a feature because it supposed to be general purpose,

I should probably rename rooms to like address

Velua:
So for connecting clients would they first push out the broadcast followed by the ICE candidate broadcast? 
Should other clients wait around for the ICE broadcast to come through before attempting a connection to the other peers?

Velua:
Probably out of scope, but messages which allow a peer to request another peer to mute their microphone or video would be helpful, would be up to the peers / environment to determine to actually honour the request / demand. 
Even if it was just a toggle on the client "Allow others to mute me".

jacany:
> So for connecting clients would they first push out the broadcast followed by the ICE candidate broadcast? Should other clients wait around for the ICE broadcast to come through before attempting a connection to the other peers?
> 
> If Alice is waiting a room for Bob and receives Bobs initial broadcast, she won't know if she should wait a moment longer for his ICE details before attempting connection or not.

https://developer.mozilla.org/en-US/docs/Web/API/WebRTC_API/Signaling_and_video_calling#signaling_transaction_flow
You can also watch this process happen in real time by opening https://nrtc.jacany.com on two different devices and opening the console on both of them before enabling microphone, it's all logged to the console.

> Probably out of scope, but messages which allow a peer to request another peer to mute their microphone or video would be helpful, would be up to the peers / environment to determine to actually honour the request / demand. Even if it was just a toggle on the client "Allow others to mute me".
> 
> Though suppose you could transmit this type of data over the peer connection rather than Nostr...
> 
> However, if you were to broadcast your mic / video state, new connecting peers could determine if a lack of a video stream is due to a connection fault, or it being turned off.

I'm confused here, clients can locally mute a peer. No need to ask the other client to stop transmitting.

Client disconnects are typically detected by the other peer ~5 seconds after it actually disconnected so it's almost negligible. Though this time only considers when the tab is forcibly closed (like closing the browser tab or force quitting an app), when a client wants to disconnect gracefully they can send a `type:disconnect` and it'll instantly be closed.

Velua:
I just meant in the event you've got 4 peers in a group call, where someone mutes the mic of another and it's reflected across all connections just like Zoom does it. 

Have you got a link to the source code of https://nrtc.jacany.com/ ?

jacany:
> I just meant in the event you've got 4 peers in a group call, where someone mutes the mic of another and it's reflected across all connections just like Zoom does it.
Oh I see. That's Out of Scope for this NIP I believe because that's the client's job.

> Have you got a link to the source code of https://nrtc.jacany.com/ ?
https://github.com/jacany/nrtc

adamritter:
I was just thinking of doing something similar, but instead of having rooms I just want a simple audio / video call application where people can call me using my pubkey.

Shouldn't room id be optional in that case? I understand that for traditional webrtc signalling room ids were needed, but I don't understand why they are needed for direct calls in nostr.

adamritter:
The reason why I believe it would be important to specify the direct phone and video call use case is that I believe we should have both separate apps for voice / video calls and probably some clients will want to integrate it.

It would be also cool if I could click call / video call in a nostr social network client on an author and that could open an external app that can handle the call.

jacany:
> Shouldn't room id be optional in that case? I understand that for traditional webrtc signalling room ids were needed, but I don't understand why they are needed for direct calls in nostr.
The spec could be modified to that room id and pubkey are interchangeable. ex: pubkey for one-on-one and `r` for group settings.

adamritter:
Sure, any solution is good, I just think it should be written in the spec so we can play around with apps that do that (your code looks easy to modify :) )

bu5hm4nn:
I love this NIP. Potentially even more could be done over WebRTC like gossipping about relays or possibly even publishing each others notes (in a large enough webrtc pool this might help obfuscate IP's?)

What do we need to do to progress on getting this merged?

fiatjaf:
This should probably be merged with https://github.com/nostr-protocol/nips/pull/498.

jacany:
Maybe this needs to be nip-04 encrypted?

jacany:
@fiatjaf could this be merged please?

tekknolagi:
It seems like this is not just for voice/video but would also support bootstrapping a text channel as well. This could make having a relay in the browser possible with no additional software:

1. Do WebRTC signaling over nostr via an existing relay
2. Start WebRTC data channel
3. Send and receive Nostr events over WebRTC in full duplex connection

I don't know the particulars around WebRTC for >2 participants, but if it's possible to do group video calls it must be possible to do a group relay type thing.

fiatjaf:
I maintain my previous opinion on this. More concretely: it is not really a standard if it doesn't specify what clients should do once they get the webRTC connection, i.e. what messages should they send and what should they expect in return.

Also 0xchat as far as I know has not implemented anything that is compatible with any other app, they're making their own custom stuff with no regard for standardization or interoperability, so it's hard to get a grasp of what is really there or not.

jacany:
> I maintain my previous opinion on this. More concretely: it is not really a standard if it doesn't specify what clients should do once they get the webRTC connection, i.e. what messages should they send and what should they expect in return.
I think that's because it's so generic. It's really an application-to-application basis.
Maybe the proper path forward would be to change this.

> Also 0xchat as far as I know has not implemented anything that is compatible with any other app, they're making their own custom stuff with no regard for standardization or interoperability, so it's hard to get a grasp of what is really there or not.
Good point.

Will give this some more thought on how it can be improved, thanks.

tekknolagi:
> I maintain my previous opinion on this. More concretely: it is not really a standard if it doesn't specify what clients should do once they get the webRTC connection, i.e. what messages should they send and what should they expect in return.

Would it be sufficient to suggest things with MAY? The initial discovery/signaling is pretty clear cut but thereafter there is a *lot* of flexibility inside WebRTC. People use it for voice, video, any live p2p syncing, chat, etc. I don't know that adding the entire WebRTC solution space to the NIP is the move.

I don't know enough about Nostr, WebRTC, or 0xchat to comment much more.

k-kozika:
I think this is excellent. Can this be improved by indicating uses such as "video call", "voice chat" or "send/receive nostr events over webrtc" in this nip?

k-kozika:
> My position is that it would be valuable to have a basic WebRTC NIP (like NIP-01) that other NIPs build upon (the audio calling NIP builds on the WebRTC NIP)

I see, so this nip is about leaving the rest to other specifications once the signalling is done. I thought it was similar to nip-59 and nip-17 in that it separates the underlying methodology from the specific solution to make it more generic.

riccardobl:
I've built my own [rtc signaling](https://github.com/riccardobl/nostr-rtc) for nostr and i'm now looking into adapting it to this nip.

I have some suggestions:
1. always require a roomId even for one-to-one connections (a room of two people), since it makes the implementation easier as it doesn't have to handle two special cases
2. do not encrypt 'r', since encryption cannot be used to broadcast the 'connect' event to a specific room before knowing its participants. Instead the spec could suggest to use random or otherwise non sensitive room ids.
3.  For authentication (password protected rooms), include an optional preshared secret in the events instead of relying on the encrypted 'r' tag.

Issue #369, opened 2023-03-16
NIP-07 Extension : The need for message / challenge signing from `window.nostr`


b35363:
The fact that `nip-07` supports only Event signing is frustrating. Why not signing also messages ?

PR #370, opened 2023-03-16
NIP-79 : `window.nostr` Offline Message Signature & Verification

b35363:
The ability to sign and verify messages must be available client-side in order to have a proper implementation in the case where Users and Remote needs to be mutually verified.

The authentication between clients and relays can already be implemented with nip-42, but this standard may lead to a much simpler authentication process where no Event has to be submitted on the network to verify authenticity.


b35363:
This proposition is based on this Issue #369

fiatjaf:
Now about the proposal here, you say it is good for other things but you don't say what other things.

b35363:
Apologies for not being clear enough in my previous explanation. The proposed `NIP-79` standard for `window.nostr` offline message signature and verification is primarily intended to provide a simpler and more direct way to authenticate clients and relays without having to send events through the network.

However, there are other potential use cases for this standard as well. For example, it could be used where both parties can sign and verify messages to ensure authenticity and integrity. It could also be used in applications that require secure and verifiable message exchanges.

It is a fact that NIP-42 standard provides a secure way for authentication through challenge signature. However, the NIP-79 standard for `window.nostr` offline message signature and verification offers additional benefits in situations where sending events on the network for authentication is not desirable or practical.

One potential benefit of `NIP-79` is that it enables direct verification of message authenticity without relying on the network for verification. This can be particularly beneficial in scenarios where network connectivity is intermittent or unreliable, or where sending events on the network can introduce unnecessary latency and delays.

Overall, the `NIP-79` standard offers a flexible and versatile solution for implementing offline message signature and verification, with potential benefits for a wide range of use cases in the field of information technology.

fiatjaf:
But you can sign an event without having to send it through the network!

b35363:
yes but you only can sign events, which is fine unless you need to sign something else. This is not a big deal, 
I'm just really confused how it has not been implemented before, this is literary 10 lines of code and allows a better flexibility for future protocols.

Example: NIP-11, relays can now identify themselves, how can we be sure that this relay is really who he claims to be ?

We could bind relay specific data (dns, ip or certificate) and sign it. How can the client verify the integrity of the said certificate ?  Yes we could create an event with the certificate in message section then sign it, but I find being forced to use events is counter-productive

Issue #319, opened 2023-03-02
NIP-08 fails to specify how to represent `#[0]` etc. as-is


shafemtol:
### Problem

Let's say I want to discuss NIP-08 through nostr itself. I might try to mention an example of what the actual plaintext looks like behind a mention. So I might try to say inside a kind-1 text note as a reply to someone else:

`Hi, @alice. In the raw json, your mention of @bob inside your note is actually represented as #[0].`

On all clients I have tested, this comes out wrong: The text `#[0]`, which should be represented as-is, collides with whatever the client put at position 0 in `tags`, likely either one of the mentioned users or the id of the note being replied to. It might come out as e.g.:

`Hi, @alice. In the raw json, your mention of @bob inside your note is actually represented as @alice.`

Which ruins the message in this case.

### Suggested Solution

- NIP-08 should specify that whenever the user tries to actually write `#[0]`, `#[1]` etc., the client MUST NOT put a tag other than some specified placeholder tag at the corresponding position in the `tags` list, putting any other tags at other positions in the list.
- NIP-08 should specify/clarify that whenever there is an apparent NIP-08 reference such as `#[0]` but where there is not an `e` or `p` tag (or other tag explicitly specified as having NIP-08 functionality), the client MUST show the text as-is.

These two together, if implemented by clients, fix the issue in a backwards-compatible way.

As for the placeholder tag, this could be something like `ph` (single-letter should be avoided due to NIP-12).

shafemtol:
Demonstration of broken notes here: [note1ztk0np5vpy58k8gytsxpmhshx7vqeyttyvse47p4aghcwhkvvadq3vvfsy](https://astral.ninja/note1ztk0np5vpy58k8gytsxpmhshx7vqeyttyvse47p4aghcwhkvvadq3vvfsy)
In the 4 notes referenced from this one, I try using four different client to write in plain text:
`Testing reply from <client>: #[0] #[1] #[9]`
All clients put the id of the note being replied to and/or the pubkey of the user being replied to at `tags` positions 0 and/or 1, resulting in an erroneous mention. NIP-08 should more clearly specify how clients can avoid this situation by specifying a placeholder tag to put at positions 0 and 1 in this case.

[...]

PR #371, opened 2023-03-16
NIP-80: Preventing False Tag Reference Matches

shafemtol:
WIP, opening PR to reserve the NIP number.

Addresses the remaining part of #319, complementing #328.


staab:
I think we can call this done, NIP 08 is not in use anymore.

PR #375, opened 2023-03-17
Supply Chain EPCIS NIP-128 & NIP-420

mabbus:
The purpose of this NIP is to replace VRS and AS2 servers. They are overly complicated systems that require third parties or extensive budget to implement. NOSTR provides an easy way to implement this communication and solves the "ownership" problem by having a public key attached to all notes.

Pura Vida!

### DIFF
--- a/128.md
+++ b/128.md
@@ -1,0 +1,45 @@
+NIP-128
+======
+
+GS1-128 Barcode
+-------------------------
+
+`draft` `optional` `author:mabbus`
+
+The goal of this NIP is to create a decentrilized ledger of events for Supply Chain distributors for interoperability that could potentially replace AS2 messaging and VRS check.
+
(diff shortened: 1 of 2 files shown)


fiatjaf:
Honestly I have no idea of what is happening here, but it looks interesting. Is this just a way for companies to share data about products and their respective barcodes?

fiatjaf:
What is the probability of any of these companies to adopt this method of sharing data?

mabbus:
Thanks for the stokeness! I haven't given up on it. Currently implementing with NIP-01 and NIP-04 instead of NIP-128 and NIP-420 since we don't need to make everything a NIP as I've read in other discussions.  Although, would be cool to have is documented as a NIP. I'm a noob at all things nostr but I can see the possibility of it revolutionizing product tracking. Any advice and guidance is appreciated. :)

PR #380, opened 2023-03-21
Drop #[index] notation

arthurfranca:
Searching for regular http urls (or `nostr:...` for clients supporting NIP-21) inside note content and replacing them with images, videos or links is already being done by nostr clients.

So the `#[index]` notation isn't bringing extra functionallity to the table.

In fact, such notation is doing more harm than good, because it requires adding `e` tags which negatively interferes with thread loading.


fiatjaf:
No matter how bad this standard is, we can't just drop it from the NIP like that.

fiatjaf:
Maybe create a separate NIP first with another approach, see if client makers like that, then that's it.

We need a way to mark some NIPs as "unrecommended" or something.

mikedilger:
See: https://github.com/nostr-protocol/nips/pull/371
(this is not a recommendation)

PR #381, opened 2023-03-21
Add alternative mention handling NIP

arthurfranca:
From [previous discussion](https://github.com/nostr-protocol/nips/pull/380):

> Searching for regular http urls (or `nostr:...` for clients supporting NIP-21) inside note content and replacing them with images, videos or links is already being done by nostr clients.
> 
> So the `#[index]` notation isn't bringing extra functionallity to the table.
> 
> In fact, such notation is doing more harm than good, because it requires adding `e` tags which negatively interferes with thread loading.


fiatjaf:
I like this. It seems to me it is a good idea to get everybody to switch from the old model to the new model and abandon the old model. Is that feasible?

Egge7:
I like it and it would be easy for us (Current) to implement this, as we are parsing content for URLs anyways.

27.md
```diff
@@ -0,0 +1,18 @@
...
+Text Note References
+--------------------
+
+`draft` `optional` `author:fiatjaf` `author:scsibug` `author:arthurfranca`
+
+This document standardizes the treatment given by clients of inline references of other events and pubkeys inside the content of `text notes` (currently kinds 1 and 30023).
+
+Clients that want to allow inline mentions of profiles MUST show an autocomplete component or something analogous to that whenever the user starts typing a special key (for example, "@") or presses some button to include a mention etc -- or these clients can come up with other ways to unambiguously differentiate between mentions and normal text.
+
+Once a mention is identified, for example, the pubkey `27866e9d854c78ae625b867eefdfa9580434bc3e675be08d2acb526610d96fbe`, the client MUST replace its textual reference (inside `.content`) with the notation `nostr:nprofile1qqsw3dy8cpu...6x2argwghx6egsqstvg` as per NIP-21. If client doesn't support NIP-21, it can instead use a regular http link to a nostr client such as `https://snort.social/p/nprofile1qqsw3dy8cpu...6x2argwghx6egsqstvg`. The client MAY add that pubkey to the `.tags` with the tag `p` if it wishes to notify mentioned user.
```
staab, line 13: Probably best not to mention a specific client in NIPs

08.md
```diff
@@ -1,3 +1,5 @@
+> __Warning__  `unrecommended`: discouraged in favor of NIP-27
```
staab, line 1: "deprecated in favor of NIP-27"

staab (approved):

monlovesmango:
I like this too

>The same process applies for referencing event IDs, except that the client SHOULD NOT add the event id to the `.tags` with the tag `e` so to not negatively impact thread loading.
Because `e` tags on `text notes` are usually considered ancestor events used to build a conversarion thread (as per NIP-10), it is advised to use inline references when mentioning events and NOT adding corresponding `e` tag to the `.tags` event key.

As it stands, NIP-10 does allow for 'e' tags for quoting/reposting of text events so I don't think we should be restricting the usage of 'e' tags in here. Once there has been a decision on how to update NIP-10 we can circle back and update this NIP as needed. Until then we shouldn't be deprecating the ability to query for quotes/reposts of a particular kind 1 event.

fiatjaf:
Wait, the NIP doesn't make it clear what is the format of the mentions. From the PR comment I thought the format of the mention was a NIP-21 `nostr:...` link. Why isn't it? Without that we can't know what is being mentioned.

v0l:
Makes sense, its also what i suggested for silent mentions: https://github.com/nostr-protocol/nips/pull/293#issuecomment-1450127192

arthurfranca:
@monlovesmango 
Counting citations of scientific articles is a thing, so ok if your long-form content is a paper add `a` tags.
I'm not sure if counting kind 1 mentions is useful, but if it is, there's the problem of subscribing to `e` tags and receiving a bunch of replies instead of mentions. So I would suggest using a different tag for mentions.
But maybe we would be better with [kind 6 (repost)](https://github.com/nostr-protocol/nips/issues/173#issuecomment-1478465856) ressurection. What do you think?

Egge7:
> Makes sense, its also what i suggested for silent mentions: https://github.com/nostr-protocol/nips/pull/293#issuecomment-1450127192

I don't really see the benefits of having silent mentions, aside from some edge cases. But in the end this new approach makes the p-tag optional and allows users (or clients) to choose wether to include a p-tag for notifications.

But I agree with @fiatjaf. If we go for this then clients MAY show autocomplete hints for pubkeys, but it's just QoL not required.

And as outlined above clients MAY include a mentioned pk as p-tag for notifications.

fiatjaf:
@arthurfranca can you clarify this? https://github.com/nostr-protocol/nips/pull/381#issuecomment-1480361524

arthurfranca:
@fiatjaf @Egge7 I've improved the text

fiatjaf (approved):

27.md
```diff
@@ -0,0 +1,18 @@
...
+Text Note References
+--------------------
+
+`draft` `optional` `author:fiatjaf` `author:scsibug` `author:arthurfranca`
+
+This document standardizes the treatment given by clients of inline references of other events and profiles inside the content of `text notes` (currently kinds 1 and 30023).
+
+Once a mention is identified by a client, for example, when an user pastes a NIP-19 nostr profile string inside the event content input field (or optionally by any other means the client may wish to support, such as selecting an user from a context menu or autocompleter), the client MUST replace it with the notation `nostr:nprofile1qqsw3dy8cpu...6x2argwghx6egsqstvg` as per NIP-21 before effectively creating the event. The client MAY add the corresponding pubkey to the `.tags` with the tag `p` if it wishes to notify mentioned user.
+
+The same process applies for referencing event IDs (using `nostr:nevent1...`), except that the client SHOULD NOT add the event id to the `.tags` with the tag `e` so to not negatively impact thread loading.
```
Egge7, line 13: should be "(using `nostr:note1...`)" instead of "(using `nostr:nevent1...`)"

Egge7: Oh, I apologize. Obviously my knowledge of NIP-19 was not up to speed!

staab:
Just happened to be working on this stuff and came up with a possible ambiguity, should `@` be included in the `content` of the note? So in other words, the content would look something like this: `hello @nostr:npub19879237`, and be decoded to `hello @bob`?

27.md
fiatjaf: Shouldn't the NIP be agnostic as to what is given there? There is also `naddr1...`.

fiatjaf:
@mplorentz @mikedilger

arthurfranca:
> should @ be included in the content of the note?

@staab I don't think so. While the client can display @anything_it_wants, behind the scenes the event .content has `hello nostr:nprofile1qqsw3...` (please avoid npubs xD).

The client could display `hello @nprofile1...` (i wouldn't use "nostr:" prefix as a nostr client is already in the context of nostr). Or if a NIP-05 is avaliable from metadata like bob@example.com i suggest showing @bob.example.com. If instead you have an username from metadata like Papa Jack i would turn it into @papa_jack

27.md
monlovesmango: yeah I guess. but kinda seems good to nudge everyone into including relay data as would make everything more reliable/targeted.

naddr1 is a good call

verbiricha (approved):

bu5hm4nn:
I object to line 13 of this proposal in respect to 'e' tags. 

We already have approved the change to [NIP-10](https://github.com/nostr-protocol/nips/blob/master/10.md) that **discourages unmarked** 'e' tags and **encourages marking** 'e' tags.

If we enforce marking 'e' tags by deprecating unmarked tags in every client (by not augmenting unmarked tags in any way) we could get the desired outcome of smarter and slimmer thread building by filtering 'e' tags. So in other words we need to make relays and filters smarter, not a new NIP.

The proposal of this PR "the client SHOULD NOT add the event id to the `.tags` with the tag `e` so to not negatively impact thread loading" ([link](https://github.com/nostr-protocol/nips/blob/45b1860b52025182d6400fd804a5d376a183ee40/27.md?plain=1#L13)) will make parsing events more difficult for clients that actually WANT to load every mention, for example because they want to display mentions in a rich way (for example a mousover preview of the mentioned post), as these clients would now need to parse the content as well.

fiatjaf:
@bu5hm4nn are you talking about notes that mention other notes or notes that _are mentioned_ by other notes?

bu5hm4nn:
My understanding is line 13 is talking about _mentioning_ other notes. I don't think there is any way mentioned notes can know they are being mentioned in the future (unless we quickly submit a NIP for time travel) ;-P

arthurfranca:
@bu5hm4nn If your client is using NIP-08, you are already parsing .content looking for `#[0]`, `#[1]` and so on.

If you were mentioning events without NIP-08 by just adding marked `e` tags, I think maybe your implementation is incomplete. Because I consider NIP-08 the go-to NIP for mentions but I may be wrong.

Inline mentions (inside .content) should be the correct way of mentioning, cause it gives you the flexibility to choose where in the text the mention will be displayed (top, middle or bottom of the text).

I think there is no reason to keep two ways of mentioning.

bu5hm4nn:
Your assumption is that the fetching and presenting are done in the same code layer, but I would argue that is not the case for the majority of the nostr network.

The first are relays: if mentions are somewhere in the tags, then relays don't have to parse the `.content` at all. They can just parse and index the `.tags`.

The other are multithreaded applications like gossip client, where I just submitted a PR for UI rendering of mentions. In gossip client there is a background worker doing all the fetching and it currently does not have to parse the `.content`. Only once we actually display the note in the UI thread do we need to parse the `.content` and replace the `#[0]` mention tags.

I think for any background processing like that having a clear separation is a huge bonus. Also think about how much more expensive computationally a regex is across the `.content` as opposed to just traversing a well formatted `.tags.` array.

monlovesmango:
i would say a bigger reason for having it in the tags is that it allows you to query relays for all quotes of an event. if you only have it in the content it is impossible to do this.

arthurfranca:
> The first are relays: [...] They can just parse and index the .tags

If you want relays to index it so that you can count mentions, please check [this comment](https://github.com/nostr-protocol/nips/pull/381#issuecomment-1480435739).

> [...] think about how much more expensive computationally a regex is across the .content as opposed to just traversing a well formatted .tags. array.

So if I follow your logic, for the sake of consistency...
if I were to remove line 13 text, it would mean in practice that you also wants this kind of event (so to make it "easier" to augment the links with cards or popups on hover as you said):
```
{
  //...,
  content: 'Visit my blog at https://myblog.com and look at this image https://myblog.com/image.jpg',
  tags: [
    ['r', 'https://myblog.com', 'page'], // tags with markers so to not rely on regexp (don't validate, just traverse tags)
    ['r', 'https://myblog.com/image.jpg', 'image'],
  ]
}
```

If instead you want to close this PR, it means you want this:
```
{
  //...,
  content: 'Visit my blog at #[0] and look at this image #[1]',
  tags: [
    ['r', 'https://myblog.com', 'page'], // tags with markers so to not rely on regex (don't validate, just traverse tags)
    ['r', 'https://myblog.com/image.jpg', 'image'],
  ]
}
```

While with this PR I want this:
```
{
  //...,
  content:  'Visit my blog at https://myblog.com and look at this image https://myblog.com/image.jpg',
  tags: []
}
```

bu5hm4nn:
@arthurfranca I don't care about 'r' tags, because they point to something on the http protocol. I am proposing to keep mentions to other events ('e' tags) in the tags array because they need to be retrieved from a relay, possibly a different relay (which is why NIP-10 proposes including the relay url for the event). 

To stay consistent you would need to modify NIP-10 with this PR as well so as not to have conflicting implementations.

We need someone who actually wrote a relay to comment on this. As a client developer, I would like to keep the 'e' tags for a mention as they have a schema and are more clear than parsing .content.

v0l:
TBH I have no idea what you guys are talking about, tagging will work the same as it always has.

v0l (approved):

bu5hm4nn:
At a minimum you need to remove "mention" marking on 'e' tags from NIP-10 as it conflicts with your "SHOULD NOT" wording. You are proposing to make all event mentions silent.

I sympathise with your intention to improve thread loading. However you are also making rich client experiences harder by making mentions invisible to relays. I believe you can also achieve the improvement by proposing filter options to have relays exclude "mention" tags.

fiatjaf:
@bu5hm4nn clients can choose to include the tags or not, @arthurfranca doesn't like the mentions included as tags because he thinks that conflicts them with replies. I personally think any mention is a form of reply so I don't see any issue there.

This NIP gives clients the option to add a tag or not, which they already have today, to be honest, it's just not standardized.

If your problem is just the wording of that paragraph that says "SHOULD NOT" then maybe @arthurfranca can reword it. Or we can merge this now and reword later. If you want we can also make a new tag for "event-mentions-that-are-not-meant-to-be-used-as-replies-but-are-still-indexed", like `m`, I don't know.

v0l:
@jb55 / @mmalmi / @vitorpamplona When support? I can't really switch until at least parsing `nostr:` links will work on other clients

vitorpamplona:
Amethyst already parses `nostr:` and `@<nip19>` words, with or without attached tags. 

FYI, this is a very verbose NIP to just say that tags are not required and that the standard inline reference for events is `nevent` and for users is `nprofile`.

KoalaSat:
Can we add at least one example? Maybe will help to avoid that during a migration a client just mix old and new implementation.

arthurfranca:
If the user/client considers mentions as replies, no problem adding the corresponding `e` tag. I may try rephrasing line 13-14, although the current text already says "should not" instead of "must not".

Problem is when people want to count mentions, when they mean "reposts", as something separate from reply count, as Twitter does, cause the subscription filter will inefficiently fetch a bunch of replies along with the mentions. So for this, maybe use the `m` tag as @fiatjaf  says.

But then we have mentions that sometimes add `e` and other times add `m` but clients will probably end up picking just one of these possibilities, and now we have two ways of doing things. So maybe for the "repost" meaning, lets [ressurect kind 6](https://github.com/nostr-protocol/nips/pull/397) and we don't add `e` nor `m` tags when mentioning?

Egge7:
> @jb55 / @mmalmi / @vitorpamplona When support? I can't really switch until at least parsing `nostr:` links will work on other clients

Current will be adding clickable nostr: URIs next week

arthurfranca:
> @vitorpamplona – this seems to be a very verbose NIP
> @KoalaSat – Can we add at least one example?

Tough xD

vitorpamplona:
> If the user/client considers mentions as replies, no problem adding the corresponding `e` tag. I may try rephrasing line 13-14, although the current text already says "should not" instead of "must not".
> 
> Problem is when people want to count mentions, when they mean "reposts", as something separate from reply count, as Twitter does, cause the subscription filter will inefficiently fetch a bunch of replies along with the mentions. So for this, maybe use the `m` tag as @fiatjaf  says.
> 
> But then we have mentions that sometimes add `e` and other times add `m` but clients will probably end up picking just one of these possibilities, and now we have two ways of doing things. So maybe for the "repost" meaning, lets [ressurect kind 6](https://github.com/nostr-protocol/nips/pull/397) and we don't add `e` nor `m` tags when mentioning? (cause if we add just `e`, some clients will keep counting reposts with the said filter instead of fetching kind 6s)

This is so confusing... 

Why not writing exactly what should be done in each case? 

Amethyst solves this by parsing the `.content` for tags. All e,a tags that were directly cited in the text become quotes, not replies and all p tags that were cited become direct citation and not simply a nudge to each reply author's notification. I recommend using this system to avoid depending on the goodwill of the client signing the message implementing the most up-to-date protocol.

dtonon:
I like the idea of a self contained contents that embed events and npubs mentions in a proper format without depend on tags.
But I agree with @bu5hm4nn and @monlovesmango: 'e' tags flagged as "mention" are useful to retrieve content from the relays and they work as a processing cache. It's a cache layer that need to be curated by clients and this is good if we want to keep relays simple and healthy dumb. Why should "mention" e tags slow down thread loading? Is clients' matter to improve the threading strategy.

PS: Every time we say that clients SHOULD / MAY instead of MUST we shift the complexity management towards relays and risk to make the standard fragile. Is that what we want?

dtonon:
@arthurfranca 

> Problem is when people want to count mentions, when they mean "reposts", as something separate from reply count, as Twitter does, cause the subscription filter will inefficiently fetch a bunch of replies along with the mentions. So for this, maybe use the m tag as @fiatjaf says.

Why not implement the 'e' tag marker in the REQ event?

mikedilger:
So is the point of this that people want some of their events to refer to other events but not be searchable as such?  Such as when you point out Joe Biden's idiocy but don't want Joe Biden to know you are doing it?  I think that is an option we should support.

I think there definitely should be an option to have an 'e' tag as well.

It seems this bloats the size of messages.  My suggestion long ago to allow 'e' tags to point to multiple relays was shot down based on bloating the size of messages.  But nobody here seems to care anymore, inserting very long bech32 encoded things into messages certainly bloats their size, especially if most of the time they refer to the same thing the 'e' tag refers to. IMHO that is quite inconsistent group think.  If we really care about message size, we should consider a shorthand like '#[0]' ... oh hey! looks familiar! It is already coded!   If we don't care about message size, 'e' tags should support multiple recommended relay URLs.

/out

fiatjaf:
> Why not implement the 'e' tag marker in the REQ event?

Because that is an annoying and unelegant breaking change.

fiatjaf:
It looks like we have a lot of agreement here. I'll merge this and if anyone wants to send patches to improve the text let's do it in another thread.

alexgleason:
> If your client is using NIP-08, you are already parsing .content looking for `#[0]`, `#[1]` and so on.

You don't parse the content to do that. You loop over the tags, construct a string for the current index `#[${i}]` and replace-all on that string. I thought it was designed that way to specifically not require parsing content.

fiatjaf:
> You don't parse the content to do that. You loop over the tags, construct a string for the current index #[${i}] and replace-all on that string. I thought it was designed that way to specifically not require parsing content.

This is true, but it was a bad idea because clients will parse the content anyway to try to fix bad implementations from others that ended up leaking raw npubs in the content, then the protocol starts to have two separate ways of doing mentions and references. Many clients are already doing it or planning on doing it, so we will be better if we standardize this here and get rid of the other way.

fiatjaf:
@arthurfranca I've mostly rewritten your NIP and used my dictator powers to commit to master directly: https://github.com/nostr-protocol/nips/blob/9b575b151467b9eba055585820a4fa56822c6d64/27.md

Please complain if you think the result was very bad.

fiatjaf:
By the way, I will also modify NIP-23 to use this new format. It currently says it should do NIP-08 style mentions but I don't think anyone has implemented or used such things at all ever.

bu5hm4nn:
The dictator saves the day and even did all the work to fix it

dtonon:
@fiatjaf 

> Because that is an annoying and unelegant breaking change.

Adding a "#em" (event + marker) filter option should not break anything and gives the power to do more specifics query. I agree that it is not elegant at all, perhaps as much as adding the marker itself instead of using different tags. Was just an idea against the "e" mention-marked tags removal, but I also think it is not really needed.

I like the final version.

Egge7:
> Adding a "#em" (event + marker) filter option should not break anything and gives the power to do more specifics query. I agree that it is not elegant at all, perhaps as much as adding the marker itself instead of using different tags.

If markers are the way to go forward relays should offer a way to filter for them. It allows much more granular filtering and safes tons of data / bandwidth, while allowing the protocol to evolve further.

I proposed adding a marker-key to filters here:
https://github.com/nostr-protocol/nips/pull/388#issuecomment-1481389533

fiatjaf:
Added support for pasing both NIP-10 and NIP-27 references in a single function on https://github.com/nbd-wtf/nostr-tools#parsing-references-mentions-from-a-content-using-nip-10-and-nip-27.

Feedback and patches welcome.

v0l:
> So is the point of this that people want some of their events to refer to other events but not be searchable as such? Such as when you point out Joe Biden's idiocy but don't want Joe Biden to know you are doing it? I think that is an option we should support.
> 
> I think there definitely should be an option to have an 'e' tag as well.
> 
> It seems this bloats the size of messages. My suggestion long ago to allow 'e' tags to point to multiple relays was shot down based on bloating the size of messages. But nobody here seems to care anymore, inserting very long bech32 encoded things into messages certainly bloats their size, especially if most of the time they refer to the same thing the 'e' tag refers to. IMHO that is quite inconsistent group think. If we really care about message size, we should consider a shorthand like '#[0]' ... oh hey! looks familiar! It is already coded! If we don't care about message size, 'e' tags should support multiple recommended relay URLs.
> 
> /out

Id prefer if we dropped the relay hints completely in favor of putting the pubkey of the event creator, then use NIP-65 relays to find the event. Posting relay URL's in immutable events doesn't make sense to me.

fiatjaf:
> Id prefer if we dropped the relay hints completely in favor of putting the pubkey of the event creator, then use NIP-65 relays to find the event. Posting relay URL's in immutable events doesn't make sense to me.

That makes sense, but I think both approaches can coexist and relay hints could still be useful in some cases. For what you want you can just put the pubkey inside the `nevent` code.

vitorpamplona:
> > Adding a "#em" (event + marker) filter option should not break anything and gives the power to do more specifics query. I agree that it is not elegant at all, perhaps as much as adding the marker itself instead of using different tags.
> 
> If markers are the way to go forward relays should offer a way to filter for them. It allows much more granular filtering and safes tons of data / bandwidth, while allowing the protocol to evolve further.
> 
> I proposed adding a marker-key to filters here: [#388 (comment)](https://github.com/nostr-protocol/nips/pull/388#issuecomment-1481389533)

+1 For the ability to search markers, but I would do it in a more generic way so relays don't need to do this only for kind 1 events: https://github.com/nostr-protocol/nips/issues/267#issuecomment-1457107761

arthurfranca:
> I've mostly rewritten your NIP and used my dictator powers to commit to master directly

No problem. I confess it is much better now @fiatjaf

bu5hm4nn:
Support for displaying and linking NIP-27 links merged to gossip/unstable
https://github.com/mikedilger/gossip/pull/329

Egge7:
NIP-27 support coming to Current with the next release: https://github.com/starbackr-com/current/commit/5eed1b23aa8a4abcb4ae000d031183c6e5b1ae6f

PR #382, opened 2023-03-21
NIP-60 - Secure file load from URL

leandro21b:
This `NIP` allows images, videos, audios, source codes, etc. to be loaded more securely by `clients` and allows these files to be stored on several servers at the same time.

### DIFF
--- a/60.md
+++ b/60.md
@@ -1,0 +1,39 @@
+NIP-60
+======
+
+Secure file load from URL
+-------------------------
+
+`draft` `optional` `author:leandro21b`
+
+## Description:
+
+A mechanism that allows the client to verify the hash of a file such as image, video, source codes, and others And ensure that the server that hosted the file has not changed any data, in addition that same file can be stored on several servers at the same time, so if one server goes down the file can still be loaded via other servers.
+
+When a user is going to publish a file, the client takes the file type ex: (img), the format ex: (png), and generates a SHA256 hash, then the client sends the file to the servers that the user chooses, then the client takes the urls that carry the file, unites all this data and declares it in the field (tags), several different files can be declared in a post as in the example below:
+
+## Example:
+
+```json
+{
+  "id": <id>,
+  "pubkey": <pubkey>,
+  "created_at": <created_at>,
+  "kind": 1,
+  "tags": [
+    ["img", "png", <ImgFileSHA256>, <ImgURL1>, <ImgURL2>],
+    ["vid", "mp4", <VidFileSHA256>, <VidURL1>, <VidURL2>],
+    ["aud", "mp3", <AudFileSHA256>, <AudURL1>, <AudURL2>],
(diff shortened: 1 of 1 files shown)


Egge7:
How exactly does this allow a file to be loaded more securely, if a client has to load the file in order to then verify the hash?

leandro21b:
The idea of hash verification is to guarantee that the file that someone sees is the same one that was sent, when someone wants to publish an image for example, before the file is sent to the servers, the client generates a hash of that file, later if the server try to change this file will give error in the hash and the client of the user who is seeing this image for example can warn him about this, this increases security and reduces the risk of data manipulation.

arthurfranca:
This is already proposed at [NIP-94](https://github.com/nostr-protocol/nips/pull/337)

leandro21b:
Interesting, also because it can be integrated with the `BitTorrent` protocol, but does it also support several `HTTP` Urls to load a file? Actually this is my main idea with `NIP-60`, because today to load an image via http you depend on a centralized server, hash verification is a way to prevent some of these servers from changing the file that was published, I think that the name `SecurefileloadfromURL` isn't even appropriate to describe my idea, as I didn't focus the name on Decentralization of storage.

arthurfranca:
> [...] but does it also support several HTTP Urls to load a file? [...]

Sure, just add many url tags instead of one.

leandro21b:
Got it, very interesting this `NIP-94`

leandro21b:
Attention: As the main idea of `NIP-60` already exists in `NIP-94` and as I already had another open PR `NIP-53` that I had to close to open this one because both are from the `master branch` of my fork and Github does not allow the 2 PR at the same time from the same branch I am closing this one, if I need to reopen this PR I will have to open another `PR NIP-60` unless there is a way to have 2 PR from the same branch, I apologize for the confusion is that I I'm still learning how to use Github

mikedilger:
@leandro21b Yes put each PR on a separate branch.

@authorfranca this PR isn't the same as NIP-94 which is only a file. This PR is for content referenced from text notes. They are related and should probably be sychronized to use very similar mechanisms, but I think @leandro21b's idea is separate, important and should not be lost.  Unless kind-1 events are supposed to point to kind-1063 events instead of to the content directly, but if so that has not been NIPped anywhere AFAIK.  If I'm talking crazy just let me know, I'm sleepy.

arthurfranca:
@mikedilger yes this PR is an alternative way of referencing media with metadata on kind 1 notes using new tags. But I think we shouldn't introduce a new form of referencing things on text notes.

If all goes well with #381 (NIP-27) and NIP-94 (file metadata, not the file data itself), one could instead reference inline (inside .content) the NIP-94 event id when on text notes, without tags. Also, one could reference inline the http url directly (instead of NIP-94 event) if it doesn't need media metadata.

PR #383, opened 2023-03-22
NIP-101 Fix for stolen secret key problem

etemiz:
NIP-101 Fix for stolen secret key problem

### DIFF
--- a/101.md
+++ b/101.md
@@ -1,0 +1,67 @@
+NIP-101
+=======
+
+Fix for stolen secret key problem
+----------------------------------
+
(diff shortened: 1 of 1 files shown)


mikedilger:
This has some merit. But there are multiple solutions to this problem in multiple PRs. It would be good if those various authors talked to each other to sort it out. I'm mostly staying out of it, but waiting for the ultimate solution because we definitely need one.

fiatjaf:
Also there is no consensus, one user may set multiple successors, either as an attack or a mistake. Also, this exact same proposal was made last week, it's probably two or three PRs before this. NIP-41 solves the ambiguity of this by ensuring there is only one successor in all cases.

PR #385, opened 2023-03-22
Nip21 patch - Change URL to URI

nryo-o:
Changes proposed as per RFC3986 URI specification and not a URL.

Since we want to _identify_ (`uri`) nostr recources, and not _locate_ (`url`) `nostr:` resrouces I propose to change this accordingly.

### DIFF
--- a/21.md
+++ b/21.md
@@ -1,12 +1,12 @@
 NIP-21
 ======
 
-`nostr:` URL scheme
+`nostr:` URI scheme
 -------------------
 
 `draft` `optional` `author:fiatjaf`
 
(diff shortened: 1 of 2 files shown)


mikedilger:
IMHO:
This distinction is pedantic. No URL "locates" anything, they all "identify" them and a protocol (e.g. http) locates/fetches them via DNS lookup, TCP transport, etc.  The RFCs that distinguished these terms long ago were because some people thought this might be important and were annoying arguers so they won. WhatWG dispensed with the term URI since most humans are more familiar with URL.

nryo-o:
Closing in favor of @mikedilger s and WhatWGs argument

PR #386, opened 2023-03-22
NIP-53 - Nostr-based domain registration system

leandro21b:
Reopening the NIP-53 https://github.com/nostr-protocol/nips/pull/191

### DIFF
--- a/53.md
+++ b/53.md
@@ -14,13 +14,13 @@ Nostr Domain System (NDS)
 
 A domain registration system based on nostr where users can create domains associated with their IP addresses and onion, then this data is signed by the user's key and sent to several relays, so that the registrations do not depend on a central server
 
-As IP addresses can change from time to time, domain registration events must be replaceable and therefore are classified as KIND-10002 according to the (NIP-16 Replaceable Events)
+As IP addresses can change from time to time, domain registration events must be replaceable and therefore are classified as KIND-10020 according to the (NIP-16 Replaceable Events)
 
 ## Example:
 
 ```json
 {
-    "kind": 10002,
+    "kind": 10020,
     "tags": [
         ["dr" "Domain1", "ip4:port", "ip6:port", "onion"],
         ["dr" "Domain2", "ip6:port", "onion"] /* dr = domain register */
@@ -37,7 +37,7 @@ There are several ways to avoid confusion due to the same domain being registere
 
 ```json
 {
-    "kind": 10003,
+    "kind": 10021,
     "tags": [
         ["dro" "domainregistryevent"],
         ["dro" "domainregistryevent"] /* dro = domain registration organizer */
@@ -46,7 +46,7 @@ There are several ways to avoid confusion due to the same domain being registere
 }
 ```
 
-This event should also be replaceable as domain events may need to be changed as stated above, so it is classified as KIND-10003
+This event should also be replaceable as domain events may need to be changed as stated above, so it is classified as KIND-10021
 
 It is very important not to depend on a single organizing user, although the domain registration itself is not done by them.
 


leandro21b:
To solve the problem of duplicate names, the idea is for the user to program the device to choose which of the users who posted an `NDS event` he will trust, technically there may be double record in which case the device will ask which one the user trusts, then the user can warn who registered with the same name but for different IPs to correct it and this warning could be sent automatically by the client when detecting a duplicate record, I believe that what will avoid many wrong registrations will be the incentives of the market and good communication, the idea of the `NDS` is that ANYONE can provide this service so that it is as decentralized as possible.

There is no integration with the existing DNS, it is a new independent domain registration mechanism that devices could be configured to use and could choose to also use this in parallel with the existing DNS, but this does not need to be included directly in the `NIP-53`

KaffinPX:
This NIP is against to Nostrs nature

leandro21b:
Why? if the idea is just an event that contains names linked to IPs?

akhavr:
Won't fly.

1. Alice trust Bob.  Carol trusts David.  
2. Bob registers `nips.directory` to `mygithubclone.onion`.  David registers `nips.directory` to `myeviledits.onion`.
3. Alice and Carol refer to `nips.directory/1` but are referring to completely different texts.

Redpug111:
What about integrating the user's public key into the domain name? Only one user has that public key so duplicate registrations wont be possible. If possible, NIP-05 addresses could be implemented as an alternative for the domain root.

fiatjaf:
Comments about the existing NIP-02 petname system on https://github.com/nostr-protocol/nips/pull/191 still apply.

leandro21b:
> What about integrating the user's public key into the domain name? Only one user has that public key so duplicate registrations wont be possible. If possible, NIP-05 addresses could be implemented as an alternative for the domain root.

Interesting, that's exactly what I was thinking, actually that's the idea, the registration will be done and signed by the domain owner, the other participants will only help this information to reach more people, but they won't do the registration itself: I should have specified this better, I'll edit the NIP later.

Redpug111:
> Comments about the existing NIP-02 petname system on https://github.com/nostr-protocol/nips/pull/191 still apply.

I see a potential issue with using relative petnames. Alice wants to have her name point to the ip address x.x.x.x. Bob has her in his contact list as "alice" so when he visits the domain "alice" he is served x.x.x.x. However, if Bob refers to the domain "alice" in one of his notes, how are other users supposed to know what domain he is talking about? Should we expect clients to compile a table of that Bob's contacts to find out what alice means to him? My pis that there is no objective way to define a domain.
(I apologize for how verbose this reply was, i couldn't think of a way to give an example while also being concise)

leandro21b:
I made several changes to the NIP and updated some important concepts mainly involving avoiding duplication, I also realized that it would be better to do this in the "tags" field and not in the "content" field, although nothing prevents it from being done in this field if necessary.

Redpug111:
Correct me if i'm wrong but this doesn't solve duplication across the whole network does it? if two people follow two different "domain organizers", if the organizers say that "example.nostr" points to different ip addresses, we have the same issue. there's no way for a user to refer to a specific domain that will be understood across the entire nostr network.

akhavr:
I have a feeling that you're trying to reinvent blockchain over the nostr.  Nostr is not a blockchain and I don't believe it should be.  If you need blockchain - use bitcoin.  If it's too expensive for the purposes - use RGB second layer.

Redpug111:
> I have a feeling that you're trying to reinvent blockchain over the nostr.  Nostr is not a blockchain and I don't believe it should be.  If you need blockchain - use bitcoin.  If it's too expensive for the purposes - use RGB second layer.

Not at all. I'm proposing that to objectively refer to a domain accross the network, we need some constant to refer to that is the same for all users, and can't be duplicated. I believe this constant could be a users pubkey, so you could register example.<pubkey> as a domain and you would be able to set the ip or onion addresses as you wish. I recognize that pubkeys are long and would be problematic to type in every time, but it's the only way i can think of to refer to a domain without question. maybe users could use their personal pet names for domains, and the full pubkey would only be used for communicating between users?

leandro21b:
The idea is that a domain is registered by its owner and signed by him, then asks several well-known `organizer users` to register this event in their public list, the idea is that there is a list of well-known `organizer users` with a good reputation who provide this service , just as the ideal would be for all people on nostr to send their public posts to several relays known and used by everyone so that the post is seen by all users on the network and the more relays everyone uses the greater the decentralization. This same logic applies to `organizer users`. Of course this will not completely prevent duplication as the idea is not to have a blockchain or consensus mechanism, but it is an idea that could work depending on market incentives.

leandro21b:
Yes DNS service providers can run their own servers, the idea was that instead this would be done by the nostr network itself so as not to have to leave a server on 24h. also because the list would need to be replaced all the time these NDS events should be replaceable but this information changing all the time could create a mess without a consensus mechanism.

After analyzing these points, I believe that it would be extremely difficult for this to work without having a consensus mechanism.

leandro21b:
I will still leave it open here so that other people can analyze these concepts that can be applied in other areas outside the DNS

PR #387, opened 2023-03-22
NIP-60 - Secure file load from URL

leandro21b:
Reopening the NIP-60 https://github.com/nostr-protocol/nips/pull/382

The main ideas already exist in `NIP-94` but I'll leave this one open so ideas can be discussed


staab:
Solved by blossom, kind 15 offers a solution for encrypted files as well.

PR #388, opened 2023-03-23
Added "ancestor" marker to NIP-10

Egge7:
This PR proposes a new "marked" e-tag "ancestor". An ancestor is an event up a reply thread thats neither the direct parent, nor the root. Ancestor tags can be used to construct a reply thread by only fetching relevant events and without the need to fetch all events that reference the root id.

In order to avoid endless tag arrays for long reply threads, clients may choose to not include every single ancestors, but only the 5 closest ones. Clients can check wether a reply thread is complete by checking the oldest ancestors tags and see if the root is its direct parent.  

This idea was discussed [here](https://github.com/nostr-protocol/nips/pull/310#issuecomment-1467469952) initially, but in order to keep conversation organized and seperated I decided to move this into a new PR.

### DIFF
--- a/57.md
+++ b/57.md
@@ -6,69 +6,139 @@ Lightning Zaps
 
 `draft` `optional` `author:jb55` `author:kieran`
 
-This NIP defines a new note type called a lightning zap of kind `9735`. These represent paid lightning invoice receipts sent by a lightning node called the `zapper`. We also define another note type of kind `9734` which are `zap request` notes, which will be described in this document.
+This NIP defines two new event types for recording lightning payments between users. `9734` is a `zap request`, representing a payer's request to a recipient's lightning wallet for an invoice. `9735` is a `zap receipt`, representing the confirmation by the recipient's lightning wallet that the invoice issued in response to a zap request has been paid.
 
 Having lightning receipts on nostr allows clients to display lightning payments from entities on the network. These can be used for fun or for spam deterrence.
 
-
-## Definitions
-
-`zapper` - the lightning node or service that sends zap notes (kind `9735`)
-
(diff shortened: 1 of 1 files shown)


monlovesmango:
this would break the ability to query for direct replies only for non-root events

Egge7:
> it doesn't sound bad to me either, but then this PR would need to wait until that functionality has been approved

I think this puts us in a weird spot, where one NIP waits on another NIP to be merged, just for the other NIP to be updated, once the first NIP gets merged...

There is SO much going on right now regarding NIP-10, Reposts and Mentions, that I start to believe it might be best to agree on the things that are being discussed right now, deprecate NIP-10 all together and introduce a completly new NIP that handles all these things.

- There is a new NIP for mentions that could make the mention marker obsolete: https://github.com/nostr-protocol/nips/pull/381
- There are ongoing discussions about brining back NIP-18 with Kind 6 Reposts: https://github.com/nostr-protocol/nips/issues/173#issuecomment-1478465856
- And then there is this discussion about the phrasing of NIP-10 in general: https://github.com/nostr-protocol/nips/pull/310#issuecomment-1481300469

If the changes above get merged at some point we wouldnt need to worry about Mentions and Reposts no more. Then we could have NIP-10 (or a new NIP) focus on Markers for Root and Replies while introducing Relay-filtering support at the same time.

arthurfranca:
This is also being discussed at issue 267. This [comment there](https://github.com/nostr-protocol/nips/issues/267#issuecomment-1455080477) shows the similarity to this PR. This PR is missing the `l` tag to be able to solve the same problems.

PR #391, opened 2023-03-24
NIP-101 Algorithm Transition for Signatures and Encryption

eznix86:
A generic method for transitioning from old cryptographic algorithms to new ones for signatures and encryption without breaking compatibility with existing implementations. 

For example, if we want to transition to a post-quantum algorithm like Falcon.

Got triggered by that: https://youtu.be/-UrdExQW0cs

https://thestack.technology/post-quantum-cryptography-nist/
https://en.wikipedia.org/wiki/Post-quantum_cryptography

### DIFF
--- a/101.md
+++ b/101.md
@@ -1,0 +1,100 @@
+NIP-101
+======
+
+Algorithm Transition for Signatures and Encryption
+--------------------------------------------------
+
+`draft` `optional` `author:eznix86`
+
+This NIP introduces a generic method to transition from an old cryptographic algorithm to a new one for signatures and encryption without breaking compatibility with existing implementations. The proposal allows for a smooth upgrade path as new cryptographic algorithms become available or necessary.
+
+Transition Event
+----------------
+
+A new event `kind` `101`is created for the transition between algorithms. The event structure is as follows:
+
+jsonCopy code
+
+```json
+{
+  "id": <id>,
+  "kind": 101,
+  "created_at": <created_at>,
+  "content": '[["old_pubkey","<old_public_key>"],["new_pubkey","<new_public_key>"],["old_alg","<old_algorithm_name>"],["new_alg","<new_algorithm_name>"]]',
+  "old_signature": "<old_signature>",
+  "new_signature": "<new_signature>"
+  "tags": [],
+}
+```
+
+
+**`content`**: A string of array of arrays containing the following information:
+
+*   `["old_pubkey", "<old_public_key>"]`: The old public key being transitioned from.
+*   `["new_pubkey", "<new_public_key>"]`: The new public key being transitioned to.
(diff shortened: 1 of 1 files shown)


Egge7:
How would changing the algorithm not be "breaking compatibility"? Clients and Relays which do not support this NIP will reject all incoming events that are signed using a new algorithm, wont they?

eznix86:
> How would changing the algorithm not be "breaking compatibility"? Clients and Relays which do not support this NIP will reject all incoming events that are signed using a new algorithm, wont they?

It is in fact a smooth transition. For example. Let's say in few years the current algorithms are not suitable anymore.

This NIP will guide new users or old account to the new implementation of the new algorithm.

Today we are taking bitcoin as a standard (secp256k1) for Nostr. Bitcoin evolves and Nostr will need to evolve too.

mikedilger:
Post-quantum crypto matters right now because people can store encrypted events now and decrypt them in 10-15 years when the technology becomes widely available.  BUT this is only a problem for encryption, not signature. Nobody can currently forge signatures.  Being post-quantum safe in terms of digital signature doesn't matter yet. This is a problem that **won't be a problem for a long time**.

If we do this, we have to be backwards compatible, therefore:
1. The "id" will always have to be the SHA256 of the contents.
2. The events will continue to have a "signature" field that uses secp256k1
 
We could theoretically add a new signature field using SPHINCS+ for example, but that would require a number of steps:

1. Modify NIP-01 to specify that events may have additional fields and unrecognized fields should be ignored. This alone will break some clients but I don't know what to do except try to get everybody to manage this.
2. Continue providing the current "signature" field, but add a new field such as "signature_sphincsplus" which signs the ID (the hash) with the post-quantum algorithm.  Keep in mind that signature verification is already a heavy operation for mobile phones, two signatures will be worse.

I don't think being "generic" (as this NIP currently is) is particularly useful because switching algorithms doesn't happen so often that we couldn't just create a new NIP for each new algorithm.

Overall I think this is **not something I'm going to code** even if the NIP is accepted because it certainly will be fraught with very hard to solve issues and it's not something we need for a very long time.

eznix86:
Thank you for your feedback on this NIP. While it's true that digital signature forgery is currently not a concern, it's important to think **about Nostr's long-term security and its ability to resist censorship**.

Regarding backward compatibility, I agree that it's critical to maintain it. Your suggestion of retaining the SHA256 ID and the secp256k1 signature field is reasonable. Adding a new signature field using SPHINCS+ could be a separate NIP, as you proposed. Additionally, Nostr could consider adding a flag to the existing transition event to indicate whether it involves a signature or encryption algorithm transition, to prevent confusion.

I know that post-quantum cryptography implementation isn't an immediate concern, but it's crucial to plan ahead for it. The proposed NIP offers a generic way to transition between cryptographic algorithms, making it simpler to adopt new standards in the future. If you have any further ideas, please share them with me.

PR #392, opened 2023-03-24
Re-write nip 57 to consolidate flow and clarify terminology

staab:
As someone with not a lot of experience coding for lightning, NIP 57 was extremely challenging to implement. I've re-written the NIP to:

1. Consolidate the protocol flow
2. Clarify the agents in the flow, introducing terms "sender", "recipient", "zap request", "zap receipt", and "recipient's lnurl server".
3. Break out heftier examples and validation requirements into their own sections, which are referenced from the main protocol flow.


57.md
```diff
@@ -6,69 +6,136 @@ Lightning Zaps
...
 
-4. To generate a `zap invoice`, call the `callback` url with `amount` set to the milli-satoshi amount value. A `nostr` querystring value MUST be set as well. It is a uri-encoded `zap request` note signed by the user's key. The `zap request` note contains an `e` tag of the note it is zapping, and a `p` tag of the target user's pubkey. The `e` tag is optional which allows profile tipping. An optional `a` tag allows tipping parameterized replaceable events such as NIP-23 long-form notes. The `zap request` note must also have a `relays` tag, which is gathered from the user's configured relays. The `zap request` note SHOULD contain an `amount` tag, which is the milli-satoshi value of the zap which clients SHOULD verify being equal to the amount of the invoice. The `content` MAY be an additional comment from the user which can be displayed when listing zaps on posts and profiles.
+This request should return a JSON response with a `pr` key, which is the invoice the sender must pay to finalize his zap. Here is an example flow:
 
-5. Pay this invoice or pass it to an app that can pay the invoice. Once it's paid, a `zap note` will be created by the `zapper`.
+```json
```
blakejakopovic, line 68: This is javascript.

fiatjaf (approved):

57.md
```diff
@@ -6,69 +6,136 @@ Lightning Zaps
...
+This NIP defines two new event types for recording lightning payments between users. `9734` is a `zap request`, representing a payer's request to a recipient's lightning wallet for an invoice. `9735` is a `zap receipt`, representing the confirmation by the recipient's lightning wallet that the invoice issued in response to a zap request has been paid.
 
 Having lightning receipts on nostr allows clients to display lightning payments from entities on the network. These can be used for fun or for spam deterrence.
 
+## Protocol flow
 
-## Definitions
+1. Client calculates a recipient's lnurl pay request url in the form `https://<domain>/.well-known/lnurlp/<name>` based on the lud06 or lud16 field on their profile. The client MUST fetch this url and parse the response. If `allowsNostr` exists and it is `true`, and if `nostrPubkey` exists and is a valid BIP 340 public key in hex, the client should associate this information with the user, along with the response's `callback`, `minSendable`, and `maxSendable` values.
```
Egge7, line 15: 1. Client calculates a recipient's lnurl pay request url by decoding their lud06 or lud16 field on their profile according to the [lnurl specifications](https://github.com/lnurl/luds). The client MUST send a GET request to this url and parse the response. If `allowsNostr` exists and it is `true`, and if `nostrPubkey` exists and is a valid BIP 340 public key in hex, the client should associate this information with the user, along with the response's `callback`, `minSendable`, and `maxSendable` values.

Egge7 (changes requested):
lud06 does not dictate an endpoints format. So the url might actually differ from the common schema <domain>/.well-known/lnurlp/<user>.

57.md
```diff
@@ -6,69 +6,136 @@ Lightning Zaps
...
 
-`zap invoice` - the bolt11 invoice fetched from a custom lnurl endpoint which contains a `zap request` note
+A `zap request` is an event of kind `9734` that is _not_ published to relays, but is instead sent to a recipient's lnurl pay `callback` url. This event's `content` MAY be an optional message to send along with the payment. The event MUST include the following tags:
 
+- `relays` is a list of relays the recipient's wallet should publish its `zap receipt` to. Note that relays should not be nested in an additional list, but should be included as shown in the example below.
+- `amount` is the amount in _millisats_ the sender intends to pay, formatted as a string. This is recommended, but optional.
+- `lnurl` is the lnurl pay url of the recipient, encoded using bech32 with the prefix `lnurl`. This is recommended, but optional.
+- `p` is the hex-encoded pubkey of the recipient.
+- `e` is an optional the hex-encoded event id. Clients MUST include this if zapping an event rather than a person.
+- `a` is an optional NIP-33 event coordinate that allows tipping parameterized replaceable events such as NIP-23 long-form notes.
```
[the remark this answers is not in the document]
staab: We do have the user tag, the lnurl was added later on, I think to help with validation.

staab:
Thanks @Semisol and @Egge7, I've incorporated your feedback.

Egge7:
> can we please not use bech32-encoded lnurl tags? using bech32 is redundant in this case since there is no binary data and it just adds extra complication

Are you proposing to drop lud06 support ?

Egge7:
> > > can we please not use bech32-encoded lnurl tags? using bech32 is redundant in this case since there is no binary data and it just adds extra complication
> 
> > 
> 
> > Are you proposing to drop lud06 support ?
> 
> 
> 
> no I am proposing we just include the URL instead of encoding it with bech32 for any future uses

This works for zaps, but most wallet applications will not handle URLs well. For maximum compatibility I'd suggest sticking with LNURL. However we could introduce a new lud that would be lud06 without encoding. @fiatjaf can you give some background on why lud06 had bech32 encoding in the first place?

staab:
This conversation should be moved to a different issue/PR, this one is just about changing the language, not introducing changes to the protocol,

fiatjaf:
> @fiatjaf can you give some background on why lud06 had bech32 encoding in the first place?

A Bitrefill employee without any foresight thought they would make smaller QR codes than raw URLs, which is not even true.

fiatjaf:
Merging this because it looks good enough. Please send other PRs if anything is wrong here.

PR #397, opened 2023-03-25
Maybe bring back NIP-18

arthurfranca:
This is just a suggestion. The PR is just a way to reopen discussions.

[Some arguments here](https://github.com/nostr-protocol/nips/issues/173#issuecomment-1478465856) along with previous discussion.

This would be a way to [address this problem](https://github.com/nostr-protocol/nips/pull/381#issuecomment-1480435739) that affects this other PR (#381).

An alternative to ressurecting kind 6 reposts would be keep using kind 1 with a different tag other than `e`, like an `m` tag for instance.


Egge7:
We have recently implemented reposts by abstracting kind 6 logic from other clients source code. I think we should at least resurrect kind 6 reposts until there is an alternative that finds more adoption

fiatjaf:
Kind 6 reposts are alive and well, the fact that the NIP doesn't exist doesn't prevent them from being used. I've only deleted it because it didn't reflect the behavior of Damus and wasn't inclined to figure out the exact behavior of Damus and other clients at the time to write the NIP. If someone wants to do that work and write the NIP we should definitely merge it.

Egge7:
At Current we build a kind 6 event without content and add the reposted notes e- and p-tag. I think this is what Damus does as well. Would be great to hear other implementers approach before writing a NIP.

arthurfranca:
Damus [puts reposted event as json string inside content](https://github.com/damus-io/damus/blob/d2cf18aeeee1cadf32980af9219de9c9c3ecba07/damus/Nostr/NostrEvent.swift#L564). Although it seems a bit redundant as it already adds the `e` tag with the event id.

I've read somewhere that instead of the enconded event as json, optionally adding extra text to .content would be good (Twitter gives the option to just repost with no extra content OR write something when reposting).

But then it would be different from Damus. So if we want both things, maybe optional tag with extra content is the way to go, although bizarre.

Or we just edit the NIP to be exactly as Damus. But I don't know yet if other clients are using kind 6 in a different way than Damus.

Egge7:
> Or we just edit the NIP to be exactly as Damus. But I don't know yet if other clients are using kind 6 in a different way than Damus. Edit: @Egge7 is using without content, which won't make Damus error out, but will be an [empty repost](https://github.com/damus-io/damus/blob/d2cf18aeeee1cadf32980af9219de9c9c3ecba07/damus/Views/EventView.swift#L75) for those viewing at Damus.

I thinks it's more appropriate to use the content field for comments by the user that is reposting. Making it essentially what Twitter calls a Quote-Repost. Of course that would mean that clients would have to fetch the referenced event to actually show the whole repost or quote...

And by the way: I am more than happy to change how Current implements reposts, as soon as there is a NIP for it.

arthurfranca:
Revisiting my NIP-26 delegation [(far-fetched) comment](https://github.com/nostr-protocol/nips/pull/222#issuecomment-1465223776), I noticed how Damus kind 6 is in fact meant to be used for transferring ownership (give delete rights) to the reposter. Meaning the event is effectively a copy that carries the proof of whom originally wrote it, while not letting the original author delete it.

I think it does unlock many use cases, but personally would better name it as "Event Copy".

Although for semantics of disposable Twitter-like "Reposts", I think fiatjaf's version is the perfect fit cause it is leaner. It should use a new event kind to not collide with the "Event Copy".

Or we just use the bulk Damus "Event Copy" thing for reposts too and call it a day.

@fiatjaf which path do you prefer?
1) "Reposts" exactly as Damus version (serialized json event inside content)?
2) OR "Event Copy" NIP as Damus version and separate "Reposts" NIP with new kind and empty content (optionally with comment as content for quote-reposts)?

Egge7:
I would prefer to have it without the duplication in content, however clients SHOULD include relay hints for the event.

arthurfranca:
@Egge7 I've ressurected `Reposts` as kind 6 with blank .content with optional comment inside .content. Main reason is that the majority of clients are supporting this version of kind 6.

Took the opportunity to turn the Damus version (with stringfied original event json inside .content) into a kind 9 event for `Event Copy`, for different use cases.

Poking @mmalmi @staab @Semisol @monlovesmango @Cameri @eskema @jb55 @KoalaSat @verbiricha

KoalaSat:
As @arthurfranca said, at this moment Nostros is only compatible with the legacy NIP-08 and the new NIP-27.

 If you think about what the app actually does is just automatically inserts the `nostr:...` key inside the content, resulting in no difference between this and a manual paste of the user.

My point is, is the intention of `kind: 6` to make a difference with NIP-27? is there really a difference? If not, it sounds to me redundant.

arthurfranca:
@fiatjaf edited it as you suggested.

It is a shame that reposts and quote reposts ended up as different kinds.

Open for text improvements.

arthurfranca:
@KoalaSat unfortunatelly clients were implementing things differently. Some were using kind 1 with NIP-08(deprecated) or NIP-27 and others were using kind 6.

I think kind 6 is better cause it enables easy counting by subscribing just to the reposts. (Quote) reposts as kind 1 will get mixed with replies when counting and displaying.

fiatjaf:
> It is a shame that reposts and quote reposts ended up as different kinds.

Maybe, I keep swinging my opinion back and forth on this. But it is what it is.

Kind 1 can have multiple quoted notes, so that's a good thing. And kind 6 being just blank reposts can be ignored, counted or displayed at will, which is also good. Once you put your own comment in the thing then it becomes an actual post of yours, so kind 1 makes sense there.

fiatjaf:
Thank you very much for your valuable work, @arthurfranca, I hope things are going well in Franca-SP, your hometown.

PR #399, opened 2023-03-26
Clarify what "replacing an event" means

Giszmo:


Giszmo:
This change is in preparation for a nip proposal that would not discard old events like in nip16 and nip33.

PR #400, opened 2023-03-26
nip "Replaceable Versioned Events"

Giszmo:


fiatjaf:
The problem of creating another special range for this is that

1. this will make it so eventually all relays are expected to support storing all this;
2. we would have to know beforehand if the stuff we're writing fits in the 30000s or on the 40000s, which is probably an impossible problem and everybody will just default to 40000s since they are so much easier to use and provide expected infinite storage;
3. we can't use kind 30023 and the infrastructure that exists around it today for wiki articles or decentralized NIP registries -- but we should!

fiatjaf:
Instead I think we can provide for this use case with some nicer tricks:

1. create a special-purpose relay dedicated to storing old versions of parameterized-replaceable events
2. reference the specific event id in comments or endorsement events made over parameterized-replaceable events 
3. store the specific event ids on that special purpose relay

Giszmo:
> provide expected infinite storage

I contend. We need relay monetization and relays will inevitably charge for storing all versions of wikipedia forever in one form or another or simply not do it. Some relays might explicitly reject kind-wikipedia content because they haven't figured out a way of gauging the cost or how to charge for it. Others will organically accept the content because they figured it out.

A new kind range allows clients to signal to the relay the author's intent regarding retention. If the relay charges the author per stored event, the author might still ask to delete certain events and cause chaos with people who relied on that content. Maybe "deleting" in this range could signal to the relay that the original author released his own endorsement, making it more likely to actually be deleted as the author stops paying for its retention. Relays that charge those who query the content might be better suited but I don't want to impose one way or the other on the relays. I just want a way to signal preference and [a way to measure costs](https://nostrbounties.com/b/naddr1qq9rzd3h8y6nqwf5xyuqygzxljlrqe027xh8sy2xtyjwfzfrxcll8afxh4hh847psjckhkxwf5psgqqqw4rsty50fx) to then figure out how to make the right users pay for stuff and prune what nobody will ever miss in a generic way.

A special purpose relay ... will we have one for every use case? nip33 doesn't hold a promise to delete stuff and people are curious about how those replaceable events evolve, so this special purpose relay will exist but it will neither store all the data for free. It will need financing for that, too and knowing if something was meant to be versioned or not helps with that.

1440000bytes:
I agree with @fiatjaf

PR #403, opened 2023-03-27
Add event kind ranges table

Semisol:


staab:
NACK, I actually like this as part of the main table

AsaiToshiya:
We can close this because the same proposal merged in 2a38a7b.

PR #404, opened 2023-03-27
NIP-70 - Collaborative Events

leandro21b:


Egge7:
I like the concept. However couldn't relays be used to share partially signed events as well? I that case a relay would replace a kind 70 event with another kind 70 event of the same id if there are more signatures of the specified authors present. That way authors do not have to send the event in a DM or of protocol.

leandro21b:
The idea is interesting, but in this case I would create another kind for partially signed posts, another 2 to allow collaborative events after they are fully signed to be replaceable and others not, without all of them having to be necessarily replaceable

fiatjaf:
Better use https://github.com/nickfarrow/frostr

leandro21b:
UPDATE: Added the possibility for events to be partially signed, each time a signature is included in the event it will be replaced, after events are fully signed they can be replaceable or not

Giszmo:
I don't see why this nip would mention kinds at all. If relays/clients started dealing with collaboratively authored events, there would really not be a good reason to limit them by kind. Why not collaboratively author a kind-1 note or any other type of event?

@fiatjaf frostr uses Schnorr signatures that by themselves would not use single public keys and not list all the keys involved explicitly.

For the nip at hands, the question would be how relays would handle this new type of events and how to query for stuff. I think it would be straight-forward if you would not introduce new queries but if you want to allow searching for events signed by A and B, not only A or B, you'd need some syntax extension.

Due to the events not being nip-1 valid events, I would expect some bumpy introduction for a questionable benefit, so I would wait to see devs that really need this to go forward on this before considering a blessing for a standard.

PR #405, opened 2023-03-29
docs: nip p2p trading

nicolasburtey:
very first draft of a p2p trading spec. more work is needed but early feedback would be appreciate to see who else can be interested by this nip.

### DIFF
--- a/31001.md
+++ b/31001.md
@@ -61,7 +61,7 @@ E: Expiration date in Unix format
 
 e: Referring to a previous order. the client SHOULD remove the previous order from the list of order currently displayed.
 
-c: Trading channels in use. Separate multiple channels with spaces, e.g., "IBAN Revolut." For specifying channels for both left and right currencies, use a "/", e.g., "Onchain Lightning/IBAN Revolut"
+c: Trading channels in use. Separate multiple channels with comma, e.g., "IBAN,Revolut,US Amazon Gift Card,UK Faster Payments" For specifying channels for both left and right currencies, use a "/", e.g., "Onchain,Lightning/IBAN,Revolut,US Amazon Gift Card,UK Faster Payments"
 
 ### replace and delete
 
@@ -73,7 +73,7 @@ tag-e can be used to replace an offer, for example, when the price changes.
 
 Although it's possible to build this on top of the draft marketplace, there are several reasons to create a dedicated NIP:
 
-- Initially, P2P tradin clients will probably dedicated to this application. By using a specific kind instead of the generic {kind: 1}, it improves discoverability.
+- Initially, P2P trading clients will probably dedicated to this application. By using a specific kind instead of the generic {kind: 1}, it improves discoverability.
 - The data structure employs tags, allowing for native filtering across currencies, amounts, and expiration dates on existing relays.
 
 


11001.md
```diff
@@ -0,0 +1,90 @@
...
+an event looks like this:
+
+```
+let event = {
+  kind: 11001,
+  created_at: Math.floor(Date.now() / 1000),
+  tags: [
+    ["D", "BUY"],
+    ["C", "BTC/NGN"],
+    ["P", "10000000.00000000"], // price, 1 BTC = 10000000 NGN
```
fiatjaf, line 30: I think this may be unpractical. It is better to have the price as a percentage difference from the price given by some exchange API, as Bisq does, for example.

11001.md
```diff
@@ -0,0 +1,90 @@
...
+m: Minimum amount
+M: Maximum amount
+
+For a fixed-amount offer, set the minimum (m) and maximum (M) amounts equal.
+
+E: Expiration date in Unix format
+
+e: Referring to a previous order. the client SHOULD remove the previous order from the list of order currently displayed.
+
+c: Trading channels in use. Separate multiple channels with spaces, e.g., "IBAN Revolut." For specifying channels for both left and right currencies, use a "/", e.g., "Onchain Lightning/IBAN Revolut"
```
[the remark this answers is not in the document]
nicolasburtey: that could be another solution. the idea behind the free form `string` field, versus some `enum`, is that it's more "permissionless", which seems to go well with how nostr works. adding new payment channels with a PR is probably not that complex. maybe that should be considered, this would improve liquidity discovery with a limited initial cost. starting with the robosats list can be a good starting point.

11001.md
Egge7: Why not support both. P could have a type flag that's either "static" or "dynamic". If it's dynamic then an endpoint is expected

nicolasburtey: > Why not support both.

I guess the main reason to converge to a single solution is that the more complex the spec is, the more cumbersome it will be for clients to build on top of it.

> I think this may be unpractical. It is better to have the price as a percentage difference from the price given by some exchange API, as Bisq does, for example.

I think the solution @fiatjaf suggest would put more burden on the order `taker`, as opposed to the `maker` (which can typically be a market market, that could automate the price update using a replace `e` tag). 

Also it would add complexity to the client because now client have to fetch price on the exchange (and manage all the use case like if the exchange doesn't respond) in addition to the nostr websocket connection. 

My current thinking is that it should be optional for the client to compare the actual price than what is present on some  exchange, but this comparison should be optional and serve as a reference. I think Paxful does that (show the premium/discount) but I need to double check.

11001.md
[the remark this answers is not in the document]
nicolasburtey: yup make sense

1440000bytes:
Does it requires a NIP? Because I have tested it for a project with basic nostr events and it works as expected.

nicolasburtey:
> Does it require a NIP? Because I have tested it for a project with basic nostr events and it works as expected.

using event `{kind: 1}` ?

I guess this is my thoughts about it: https://github.com/nostr-protocol/nips/pull/405/files#diff-1a3de61680a29732cfe482cc8f4362e494a03ab61bb2d8a0eea98c2e2b6f21deR72

Egge7:
> > Does it require a NIP? Because I have tested it for a project with basic nostr events and it works as expected.
> 
> 
> 
> using event `{kind: 1}` ?
> 
> 
> 
> I guess this is my thoughts about it: https://github.com/nostr-protocol/nips/pull/405/files#diff-1a3de61680a29732cfe482cc8f4362e494a03ab61bb2d8a0eea98c2e2b6f21deR72

There should definitely be a separate kind for this... on the other hand I don't think making a nip for every single usecase works well. Maybe we should have a list of "application specific" kinds somewhere? Similar to the list of coins for derivation paths.

bu5hm4nn:
I love the idea. I agree with @Egge7 that marketplace events should have their own kind. However I also think that we need relays to specialize on serving marketplace content. Because as adoption grows, the number of entries will become far too many for client side filtering. I see P2P trading as one sub-kind of marketplace event. Filters I would like to use to retrieve marketplace content:

- location (radius)
- sub-kind (like P2P BTC, all the craigslist categories, should be an extensible enum defined in a NIP so that entries stick to the schema)
- price range
- delivery/pickup

I there is a agreement on this general schema, I can write up a NIP that roughly sketches this so that this P2P BTC pull request can build on it and move forward.

11001.md
```diff
@@ -0,0 +1,90 @@
...
+
+While it would be likely that clients are initially dedicated to this application, an ideal scenario would be that, over time, generic clients also integrate this NIP. This would incentivize a large group of users who would own Bitcoin, from zaps for instance, to trade with each other for local fiat currencies, without having the current barrier to entry of having a dedicated app.
+
+## Data structure
+
+an event looks like this:
+
+```
+let event = {
+  kind: 11001,
```
grunch, line 25: Consider using nip33 for this, events with kind `30000 <= n < 40000` can be replaced by new ones, this way you can update the tags on the order in case you needed

11001.md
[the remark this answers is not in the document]
bu5hm4nn: An argument for standardized values would be to let users filter for payment methods. If you want to allow free text, I would suggest adding a standardized payment type field for rough searching (e.g. "Cash", "Bank Transfer", "Payment Company")

11001.md
```diff
@@ -0,0 +1,90 @@
...
+
+Although it's possible to build this on top of the draft marketplace, there are several reasons to create a dedicated NIP:
+
+- Initially, P2P tradin clients will probably dedicated to this application. By using a specific kind instead of the generic {kind: 1}, it improves discoverability.
+- The data structure employs tags, allowing for native filtering across currencies, amounts, and expiration dates on existing relays.
+
+
+### TODO:
+
+- extend this nip with a kind for settled trades, __key__ extension, as it's the one needed to build reputation.
```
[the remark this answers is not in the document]
nicolasburtey: I think the reputation should be out of the scope for this nip. It will be very similar at the end than what is coming up with nostr for the twitter use case. some sort of web of trust.

nicolasburtey:
> I love the idea. I agree with @Egge7 that marketplace events should have their own kind. However I also think that we need relays to specialize on serving marketplace content. Because as adoption grows, the number of entries will become far too many for client side filtering. I see P2P trading as one sub-kind of marketplace event. Filters I would like to use to retrieve marketplace content:
> 
> * location (radius)
> * sub-kind (like P2P BTC, all the craigslist categories, should be an extensible enum defined in a NIP so that entries stick to the schema)
> * price range
> * delivery/pickup
> 
> I there is a agreement on this general schema, I can write up a NIP that roughly sketches this so that this P2P BTC pull request can build on it and move forward.

this makes sense. will incorporate the idea in the draft shortly.

11001.md
grunch: nip16: all events you send will be replaced, if you want to publish 2 orders, you publish the first one, then publish the second one, the first one will be replaced with the second one.

nip33: you identify the events with a specific `d` tag, so when you need to replace an event you just use that same `d` tag value.

11001.md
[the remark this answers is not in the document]
HenrikJannsen: If you want to have that trade protocol be based on reputation, you should think **now** how to do it. 

The protocol problem is trivial but decentralized reputation is the hard problem. Centralized reputation is rather easy but you do not gain anything from censorship resistance as the fate of Local Bitcoin, Paxful shows us. 

Web of trust is not the solution. It needs to include some costs, otherwise what prevents one from creating a web of fake identities. WoT is also a UX problem. Why has PGP WoT never become a great success? Its too complex and requires lot of work for verification and risk assessment.

11001.md
nicolasburtey: > nip16: all events you send will be replaced, if you want to publish 2 orders, you publish the first one, then publish the second one, the first one will be replaced with the second one.

gotcha. when you said `all events you send will be replaced`, the `you` represent an `npub`? ie: this is like a "singleton" instance per npub?

grunch: yes, all events from the same npub will be replaced, so one user can't send more than one

nicolasburtey: updated: https://github.com/nostr-protocol/nips/pull/405/commits/8713cf25e75fe2a4a5ee02a8d27394826ff4a373

11001.md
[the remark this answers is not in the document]
nicolasburtey: > Web of trust is not the solution. It needs to include some costs, otherwise what prevents one from creating a web of fake identities. 

how is this problem applying to this nip in particular? I think it's a problem that apply to all nostr use case (the twitter nostr use case have the same challenges) and should be address in a separate nip.

nicolasburtey:
diving into your comment @bu5hm4nn:

- is there a nip to define location and radius? if not, I think this should be done first. maybe you can get a stab at it?
- sub kind: we don't really need that, it's already implicit in the `["C", "BTC/NGN"]` array. if you want to sell gold, just use the associated ticker.
- price range: why would you set a price range? order always have a fixed price afai see from existing platform. but feel free to provide an example if I'm missing something. 
- delivery pickup: this should be done by messaging, or details can be mentioned in the `content` part of the message. I don't think this should be part of the nip.

11001.md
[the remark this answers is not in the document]
nicolasburtey: ok for now I have gone to use comma to separate payment methods. we could defined standardized value in a separate nip: https://github.com/nostr-protocol/nips/pull/405/commits/8250274a22f4882f621510df0054fd6167c10c9e

bu5hm4nn:
Thanks for considering my comment. I was thinking in terms of general marketplace notes, of which P2P Bitcoin could be a category. Basically Craigslist/Facebook Marketplace on nostr. 

Location tags in the form of geohashes are mentioned in NIP-12 and use the tag 'g'. Not sure if any relay implements a search filter for them though.

nobu-maeda:
Thanks @Reckless-Satoshi for bringing this NIP to my attention.

I have had a similar idea (many have these days) for using Nostr as a global order-book storage through relays, and to have an open inter-operable protocol so all potential clients implementations, liquidity providers, etc can be working with the same order book. Thus creating a much bigger pool of liquidity that would help bridging Bitcoin to TradFi, and thus enhancing adoption and price stability.

I am however in the camp that is against bloating the Nostr base layer for specific applications. This includes even Bitcoin marketplace applications. The more NIPS an application depends on, the higher the likelihood it will not be implemented by some relays, and thus having less relays to bolster multicast decentralization. Ideally trade traffic is ephemeral and mixed into plain-Jane social media relays, and any specification inside trade messaging is a layer 2 specification on top of Nostr.

For n3xB. My latest thinking is just to use tags for filtering, and use `kind = 30078` (NIP-78). As such no new NIP should be needed for the implementation. See https://github.com/nobu-maeda/n3xb-nostr-derisk as an example.

I do understand that client filtering will hit its limit eventually, but I'd consider it bike shedding before this takes off. If we look at Bisq or RoboSat volumes today, this would be a very good problem to have.

PatrickGeyer:
Shouldn't there be a NIP for escrow services? Those'll be required for all sorts of nostrils apps, and it would be good to be able to broadcast an escrow service, specify which payment methods are supported (hodl invoices for e.g.), and which app types can use it as an escrow.
Then a nostr app that requires escrow (marketplace, Uber, Airbnb) can just query the network for escrow services, find one that both buyer and seller have indicated support for, and make the trade from then on.
Is there conversation about this somewhere already?

31001.md
```diff
@@ -0,0 +1,90 @@
...
+    ["c", "lightning/ACH"] // payment methods
+  ],
+  content: "0 reputation are not allowed", // comment on the order. could be anything
+  pubkey,
+};
+```
+
+Optional tags are the following:
+```
+    ["e", "formerEventId" ], // used for replace order
```
dipunm, line 43: Check NIP-33, why not use "d" to allow order replacements?

Each order is unique based on its (kind,pubkey,d-tag) values and the most recent one will be the valid one.

31001.md
```diff
@@ -0,0 +1,90 @@
...
+
+E: Expiration date in Unix format
+
+e: Referring to a previous order. the client SHOULD remove the previous order from the list of order currently displayed.
+
+c: Trading channels in use. Separate multiple channels with comma, e.g., "IBAN,Revolut,US Amazon Gift Card,UK Faster Payments" For specifying channels for both left and right currencies, use a "/", e.g., "Onchain,Lightning/IBAN,Revolut,US Amazon Gift Card,UK Faster Payments"
+
+### replace and delete
+
+NIP-09 allows for the removal of an offer, such as when a trade is completed and liquidity is no longer available.
```
dipunm, line 68: Instead of relying on deletions (or additionally), maybe a status can be used alongside the replacement mechanism? I feel like that would be more reliable than expecting relays to honour deletions.

Also maybe consider NIP-42 expirations.

31001.md
```diff
@@ -0,0 +1,90 @@
...
+Although it's possible to build this on top of the draft marketplace, there are several reasons to create a dedicated NIP:
+
+- Initially, P2P trading clients will probably dedicated to this application. By using a specific kind instead of the generic {kind: 1}, it improves discoverability.
+- The data structure employs tags, allowing for native filtering across currencies, amounts, and expiration dates on existing relays.
+
+
+### TODO:
+
+- extend this nip with a kind for settled trades, __key__ extension, as it's the one needed to build reputation.
+- nip for escrows (escrow could probably be the same than for the marketplace?)
```
dipunm, line 83: Check out NIP-32?

31001.md
```diff
@@ -0,0 +1,90 @@
...
+- extend this nip with optional bonds (a la robosats)
+- specify how trade can be deleted
+- geo for in person trading
```
dipunm, line 86: NIP-12?

31001.md
```diff
@@ -0,0 +1,90 @@
...
+
+Users could also choose to create a one-time identity and participate on the same based platform by simply creating a new key-pair for each trade. It would make it possible to merge currently separate use cases where platforms today are either relying on bonds (like RoboSats) or reputation (like Paxful, Binance P2P, HodlHodl).
+
+An ecosystem where different clients use the same specification on how to create and settle trades would create a thriving environment where clients compete in their UX and product offerings, to the benefit of the users and the global liquidity of the network.
+
+While it would be likely that clients are initially dedicated to this application, an ideal scenario would be that, over time, generic clients also integrate this NIP. This would incentivize a large group of users who would own Bitcoin, from zaps for instance, to trade with each other for local fiat currencies, without having the current barrier to entry of having a dedicated app.
```
dipunm, line 17: To support non-dedicated apps, consider describing how to format an "alt" tag in the spec. NIP-31

grunch (changes requested):
Is it possible to allow users to set a `market price` order? I mean, I want to sell bitcoin and get 10 EUR but instead of setting a bitcoin price I want to sell at the market price at the moment the order is taken with premium of 2.

for this I think we should indicate the API that is being used by the origin (Mostro, bisq, etc) and we need a tag for the premium

31001.md
```diff
@@ -0,0 +1,90 @@
...
+an event looks like this:
+
+```
+let event = {
+  kind: 31001,
+  created_at: Math.floor(Date.now() / 1000),
+  tags: [
+    ["D", "BUY"],
+    ["C", "BTC/NGN"],
+    ["P", "10000000.00000000"], // price, 1 BTC = 10000000 NGN
```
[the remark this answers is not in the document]
grunch: I thought this better, we also should allow to add a price tag, but only one of both should be mandatory, users that use price tag don't send the fiat price and the other way around

nobu-maeda:
> Is it possible to allow users to set a `market price` order? I mean, I want to sell bitcoin and get 10 EUR but instead of setting a bitcoin price I want to sell at the market price at the moment the order is taken with premium of 2.
> 
> for this I think we should indicate the API that is being used by the origin (Mostro, bisq, etc) and we need a tag for the premium

I gave this some thought when drafting https://github.com/nobu-maeda/n3xb/blob/master/specs/maker-note/maker-note.md . I am trying to envision an architecture that minimizes reliance on a central coordinator. To create a market price order, the Maker and the Taker will need to agree on which price quote to use, a sort of mechanism to negotiation and establish agreement on this is needed. See n3xB on my current proposal.

Again, I am hoping P2P trading protocol to be a Layer 2 protocol. I don't think a NIP is necessary and will bloat the Nostr protocol. And also, only just one NIP and a single event message is probably insufficient to establish and complete a trust minimized trade anyways

vitorpamplona:
Was this superseded by NIP-69?

vitorpamplona:
Closed due to lack of activity

PR #406, opened 2023-03-29
NIP-47 Nostr Wallet Connect

Semisol:
This is a rewritten version of the original Nostr Wallet Connect proposal to be more flexible and reduce metadata leaks.


Egge7:
Big fan! Especially because this will enable clients to do more complex payments (like splits) with good UX and without having a native wallet. Currently working on a POC.

fiatjaf:
I will wait for @vitorpamplona to approve it.

vitorpamplona:
The text is better. But I am not a fan of this One-event-kind-that-contains-all-message-types-via-json-in-content trend. 

I am ok in merging `success` and `error` in one event kind somehow. But I do think balance and payment should be 2 different types. Additional `cmd`s should be new event kinds. With that, payment operators can filter only the messages they can parse and we don't need to deal with a "separate layer-2 spec" for these fake RPC calls. 

Parseable JSONs with sub-message types is damaging to the simplicity of Nostr and should be avoided at all costs. 

That being said, If I am overridden, I will change Amethyst to this new style.

vitorpamplona:
PS: I know this is hypocritical, since my NIP-82 (https://github.com/nostr-protocol/nips/pull/357) draft also has a sub-message type as a json in `.content`

vitorpamplona:
Ping to @kiwiidb and @bumi for their position on the `cmd`-style messages. @kiwiidb started with that message style and I moved him back to a more "native" Nostr implementation (breaking down message types in event kinds)

05nelsonm:
Json is a huge road block here (at least for any Android apps that care about performance and mitigating jank). Would have to catch the Intent and then move off UI thread just to deserialize json?

kiwiidb:
Thanks for starting the discussion and opening the PR. 

- I think that the privacy gain from merging the 2 response kinds is marginal and if users want privacy they should get it by connecting to a relay that requires authentication.
- *If* we decide to go down the JSON-payload route, I believe that the payload structure should be streamlined with NIP-46.
-  I myself don't feel strongly about it even though I lean towards the no-JSON camp because it does make things simpler. 
- Adding the balance call to the spec gives a bit more reason to go down the JSON-path, but I don't think there is a real demand from clients to add this. The only good reason that I would think of is to inform the user that they are broke, but this can be also in an error message.
- If you are not responding to events by using the `e` tag how would a client match responses with their respective requests? I guess it can be done in the NIP-46 way in the JSON case, but I think that makes it even more complicated.
- I am ok with mandating the server-generated private key.

arthurfranca:
> The key is harder to leak since it is **not shown to the user** and backed up.

Sorry in advance as I may have misunderstood things.

But if I got it right, the private key is the `secret` query string param. So it will be shown to the user when it copies the url (if not using QR Code).
Also, as the app will have to keep it around (like in localstorage) for signing future payment requests, doesn't this have the same problem as pasting your main nsec in a new app?

I'm a bitcoin noob so maybe the lightning node/wallet service won't auto-pay and instead is expected to always display a confirmation dialog for authorization at each pay_invoice request so no problem?

vitorpamplona:
> But if I got it right, the private key is the secret query string param.

Yes

> So it will be shown to the user when it copies the url (if not using QR Code).

The idea is to only show it once when you register for a new session in the wallet and do it either via QR code or app-to-app deep link calls in a phone. In that way, the private key is not "visible". The app reads it and places it inside it's encrypted storage, together with the person's nsec. 

> 'm a bitcoin noob so maybe the lightning node/wallet service won't auto-pay and instead is expected to always display a confirmation dialog for authorization at each pay_invoice request so no problem?

That's how it's coded in Amethyst. If they have a Wallet Connect key, it tries to pay through there and if it doesn't, or if it fails, it opens a regular wallet selection dialog in the phone.

arthurfranca:
@vitorpamplona thanks for explaining

> [...] If they have a Wallet Connect key, it tries to pay through there [...]

My main doubt is at that quoted moment. Is the connected wallet expected to just check if the pay_invoice request event (e.g. sent from Amethyst) is valid and has the pubkey generated from the right secret (from the query param) and if true will pay the invoice without user having to confirm anything?

My concern isn't Amethyst but some malicious or vulnerable web app. Like if said app has the secret, it could send many pay_invoice requests at 4a.m. that will be automatically accepted and payed while the user is sleeping, something like that.

So the user has to trust the app just like it trusts when pasting a nsec (instead of using NIP-07 browser extension, for instance).

vitorpamplona:
> So the user has to trust the app just like it trusts when pasting a nsec (instead of using NIP-07 browser extension, for instance).

Yep, not only trust the app but also trust the wallet connect service on the other side. The wallet connect service can add protections to block transactions that are too big or too frequent.

The main point here is indeed that the Nostr client can control auto-pay. This allows clients to create all types of Nostr-native subscription services with frequent payments without having to confirm with the wallet at every time. For instance, I wanted to do a nightly zap to all posts the user spent any time on, with a total daily allowance split proportionally to the amount of time spent per post. The app would just automatically send these if users want.

jb55:
I like this

47.md
```diff
@@ -0,0 +1,112 @@
...
+Both the request and response events SHOULD only contain one `p` tag, containing the public key of the **wallet service** if this is a request, and the public key of the **client** if this is a response.
+
+The content is encrypted with [NIP04](https://github.com/nostr-protocol/nips/blob/master/04.md), and is a JSON object. The content depends on the kind.
+
+Request:
+```jsonc
+{
+    "cmd": "pay_invoice", // command, string
+    "data": { // data, object
+        "invoice": "lnbc50n1..." // command-related data
```
thesimplekid, line 34:
```suggestion
        "invoice": "lnbc50n1...", // command-related data
        "description": "<invoice description>"
```

47.md
```diff
@@ -0,0 +1,112 @@
...
+## Commands
+
+### `pay_invoice`
+
+Description: Requests payment of an invoice.
+
+Request:
+```jsonc
+{
+    "invoice": "lnbc50n1..." // BOLT11 invoice, string
```
thesimplekid, line 80:
```suggestion
    "invoice": "lnbc50n1...", // BOLT11 invoice, string
    "description": "<invoice description>" // Invoice description, string
```

thesimplekid:
I think it makes sense to include the invoice description in the request. Core lighting has deprecated paying invoices without being able to verify the description hash though enforcing this has been reverted, I still think its best to include it. 

https://github.com/ElementsProject/lightning/pull/6092

thesimplekid (commented):
I think it makes sense to include the invoice description in the request. Core lighting has deprecated paying invoices without being able to verify the description hash though enforcing this has been reverted, I still think its best to include it. 

https://github.com/ElementsProject/lightning/pull/6092

kiwiidb:
@thesimplekid I believe this is a whole other discussion. For my opinion, see https://github.com/ElementsProject/lightning/pull/6092#issuecomment-1492941924

47.md
```diff
@@ -0,0 +1,112 @@
...
+Nostr Wallet Connect
+--------------------
+
+`draft` `optional` `author:kiwiidb` `author:bumi` `author:semisol` `author:vitorpamplona`
+
+## Rationale
+
+Paying zaps should be possible without the user needing to open a different app to only pay a Lightning invoice.
+This NIP describes a way for users to control a remote Lightning node or a custodial Lightning wallet. When self-hosting, this setup does not require the user to run their own server, thereby bypassing certain hurdles that are commonly encountered when trying to connect to a Lightning node remotely.
+
```
vitorpamplona, line 13: This NIP is designed to forward the responsibility of executing lightning payments to specialized Nostr micro apps whose sole purpose is to connect with the user's desired lightning service, generally using the service's proprietary API, and establish security controls to approve such transactions. Regular Nostr Clients are then exempt from implementing payment interfaces with each lightning provider out there and from implementing security settings to authorize certain payment types.

47.md
```diff
@@ -0,0 +1,112 @@
...
+    }
+}
+```
+
+The data field SHOULD contain a `message` field with a human readable error message if the status is `error`.
+
+## Nostr Wallet Connect URI
+**client** discovers **wallet service** by scanning a QR code, handling a deeplink or pasting in a URI.
+
+The **wallet service** generates this connection URI with protocol `nostr+walletconnect:` and base path it's hex-encoded `pubkey` with the following query string parameters:
```
vitorpamplona, line 55: There is no + in the current implementations: `nostrwalletconnect:`

47.md
```diff
@@ -0,0 +1,112 @@
...
+    "data": { // data, object
+        "invoice": "lnbc50n1..." // command-related data
+    }
+}
+```
+
+Response:
+```jsonc
+{
+    "status": "ok", // status, "ok" | "error"
```
vitorpamplona, line 42: Should we use the `OK` of the base Nostr protocol for success? Why duplicate this here?

47.md
```diff
@@ -0,0 +1,112 @@
...
+    "event": "0123456789abcdef...", // event the command is in response to, string
+    "data": { // response data
+        "preimage": "0123456789abcdef..." // command-related data
```
vitorpamplona, line 45: Does anybody need this preimage? I don't really use this in the current implementation

47.md
```diff
@@ -0,0 +1,112 @@
...
+### `balance`
+
+Description: Requests the balance of the wallet.
+
+Request: an empty JSON object.
+
+Response:
+```jsonc
+{
+    "balance": 100000 // balance in msat, int
```
vitorpamplona, line 100: In typed languages, you have to inform the object you want to map your JSON into before calling the parser. In this proposal, there is no way to know if the JSON is going to be a balance or a result beforehand, which makes it impossible to *correctly* use in typed languages. Of course, typed languages can always navigate the JSON map manually, but it's not a great implementation.

47.md
```diff
@@ -0,0 +1,112 @@
...
+
+- `relay` Required. URL of the relay where the **wallet service** is connected and will be listening for events. May be more than one.
+- `secret` Required. 32-byte randomly generated hex encoded string. The **client** should use this to sign events when communicating with the **wallet service**.
+    - Authorization does not require passing keys back and forth.
+    - The user can have different keys for different applications. Keys can be revoked and created at will and have arbitrary constraints (eg. budgets).
+    - The key is harder to leak since it is not shown to the user and backed up.
+    - It improves privacy because the user's main key would not be linked to their payments.
+
+The **client** should then store this connection and use it when the user wants to perform actions like paying an invoice. Optionally it can display metadata about the connected **wallet service** from it's profile (name, image, url).
+
```
vitorpamplona, line 65: Since this NIP uses ephemeral events, the relay used by the wallet service cannot terminate connections due to a lack of use from either the wallet service or the client side. Once a connection is established it should stay alive until the client closes it.

47.md
```diff
@@ -0,0 +1,112 @@
...
+        "invoice": "lnbc50n1..." // command-related data
+    }
+}
+```
+
+Response:
+```jsonc
+{
+    "status": "ok", // status, "ok" | "error"
+    "event": "0123456789abcdef...", // event the command is in response to, string
```
vitorpamplona, line 43: This event should be an `e` tag, not a json property.

47.md
```diff
@@ -0,0 +1,112 @@
...
+
+The data field SHOULD contain a `message` field with a human readable error message if the status is `error`.
+
+## Nostr Wallet Connect URI
+**client** discovers **wallet service** by scanning a QR code, handling a deeplink or pasting in a URI.
+
+The **wallet service** generates this connection URI with protocol `nostr+walletconnect:` and base path it's hex-encoded `pubkey` with the following query string parameters: 
+
+- `relay` Required. URL of the relay where the **wallet service** is connected and will be listening for events. May be more than one.
+- `secret` Required. 32-byte randomly generated hex encoded string. The **client** should use this to sign events when communicating with the **wallet service**.
```
vitorpamplona, line 58: If `secret` is required, the text should read: 

The **client** MUST use this to sign events when communicating with the **wallet service**. The user's private key should NEVER be used in this NIP.

vitorpamplona:
Folks, now that I read it again, I am not inclined to code this json parsing idea for 2 reasons: 
1. The current implementation doesn't use it. We don't have any client that has coded this and is happy with it. 
2. The JSON request/responses require a structure that accounts for the polymorphic nature of the incoming and outgoing JSons such that a regular JSON parser can correctly return the appropriate object without knowing what that object will be beforehand. This is particularly required in typed languages.

jb55:
@vitorpamplona 

1. I was about to implement this in damus and I think I prefer this variant. Getting balances and things of that nature would be super useful! I like the future extensibility of this and reduce metadata leakage.
2. Surely this isn't hard to build parser for?

vitorpamplona:
> I was about to implement this in damus and I think I prefer this variant. Getting balances and things of that nature would be super useful! I like the future extensibility of this and reduce metadata leakage.

Those are different event kinds, one for each type of call. The question here is between making them a separate kind or creating a type field inside a generic json like this proposal. 

> Surely this isn't hard to build parser for?

Yes, we can always manually do it. But why complicate it? JSON parsers have a structure for polymorphism already built-in, but it wouldn't work here.

vitorpamplona:
BTW, the JSON structure of this proposal is already different than [NIP 46](https://github.com/nostr-protocol/nips/blob/master/46.md). 

This is already annoying and it's just 2 NIPs. 

In NIP 46, there is no polymorphism in the base layer-2 protocol. But this `<anything>` object damages implementations in typed languages. It has to basically be an Object that must be type cast every time one uses it. 

Request
```jsonc
{
  "id": <random_string>,
  "method": <one_of_the_methods>,
  "params": [<anything>, <else>]
}
```

Response
```jsonc
{
  "id": <request_id>,
  "result": <anything>,
  "error": <reason>
}
```

In this NIP 47 proposal, polymorphic objects are required. But while the request can be polymorphically parsed, the response cannot. The status "ok" field does not imply the type of the `data` field. 

Request:
```jsonc
{
    "cmd": "pay_invoice", // command, string
    "data": { 
        "invoice": "lnbc50n1..." // command-related data
    }
}
```

Response:
```jsonc
{
    "status": "ok", // status, "ok" | "error"
    "event": "0123456789abcdef...", // event the command is in response to, string
    "data": { // response data
        "preimage": "0123456789abcdef..." // command-related data
    }
}
```

While the original [NIP 47](https://github.com/getAlby/nips/blob/master/47.md) addresses different message types at the Nostr base layer, which is already implemented by all clients. 

- NIP_47_REQUEST_KIND: 23194. Content should be a bolt11 invoice.
- NIP_47_SUCCESS_RESPONSE_KIND: 23195. Content should be the preimage of the succesful payment.
- NIP_47_ERROR_RESPONSE_KIND: 23196. Content should be an error message.

jb55:
> But this <anything> object damages implementations in typed languages

fwiw I had no issues implementing this in swift: 

```swift
struct WalletRequest<T: Codable>: Codable {
    let cmd: String
    let data: T?
}

func make_wallet_pay_invoice_request(invoice: String) -> WalletRequest<PayInvoiceRequest> {
    let data = PayInvoiceRequest(invoice: invoice)
    return WalletRequest(cmd: "pay_invoice", data: data)
}

func make_wallet_balance_request() -> WalletRequest<EmptyRequest> {
    return WalletRequest(cmd: "balance", data: nil)
}

struct EmptyRequest: Codable {
}

struct PayInvoiceRequest: Codable {
    let invoice: String
}

func make_wallet_connect_request<T>(req: WalletRequest<T>, to_pk: String, keypair: FullKeypair) -> NostrEvent? {
    let tags = [["p", to_pk]]
    let created_at = Int64(Date().timeIntervalSince1970)
    guard let content = encode_json(req) else {
        return nil
    }
    return create_encrypted_event(content, to_pk: to_pk, tags: tags, keypair: keypair, created_at: created_at, kind: 23194)
}

```

vitorpamplona:
Yep, the request creation is fine. The response parsing is the issue.

jb55:
is the concern that there is no tag? can't you determine that from the request type?

vitorpamplona:
> is the concern that there is no tag? can't you determine that from the request type?

Yep. It feels like the app should need to parse the tag, go to the database and get the event just to know which type of answer this is.

tiero:
Concept nACK.

1. **Redundancy**: NIP-46 has been created for the same exact purpose ie. nostr apps talking with remote nostr signers, to exchange data back & forth, whatever it should be: an LN invoice, a BTC address, a PSBT for a coinjoin round etc... 
Reading the comments in this NIP, it seems to me there could have been posted as issues to amend to the original NIP-46, instead to create (yet) another competing standard for connecting apps with wallets/signers. 

2. **Privacy**: there is literally no reason to communicate to the entire world that two pubkeys are into economic activity and what is a request vs a response. Using a generic kind for ALL possible "intents" increase the anonymity set and add plausible deniability to signers/wallets to not having to support specific use cases one by one. That being said, happy to hear more, I may be mistaken and not see the utility of it.

3. Fragmentation: having both `nostrconnect://` and `nostrwalletconnect://` only creates pains for wallets/signers. Is there a good reason for this?

4. Future bloat: what happens if another wallet/signer wants to have some different use case? We will create another NIP with another kind? Ie. Push a PTLC, pay offer Bolt12, sign a PSBT for coinjoin, or sign for a DLC bet...
 It also looks like a liability for relayers and Nostr maintainers, what if coinjoin-related kinds are risky (reputation/legal/regulator?)
 
I wish all these comments and needs came out during (the very long) discussion we had in NIP-46 merging process (You have also @Semisol reviewed, so you were 100% aware of it).

I am available to have offline (or online) discussions with @jb55 and @vitorpamplona to understand better what were the issues for Damus and Amethyst to not adopt NIP-46, and what could be done as a last attempt in good faith to fix it.
I am open to making changes even fundamental ones if necessary, even if two major Bitcoin wallets are started already into integrating it as it is ( to sign PSBT, share xpubs, balances, redeem scripts, etc..)

PS: I have zero personal attachment to NIP-46, I can renounce the "authorship" and transfer it to y'all today. 
Looking forward to hearing more, cheers

vitorpamplona:
@tiero I don't think NIP-46 is designed to do what this does. I also don't think we should merge every RPC-like call into one or two event kinds in NIP-46. The fact that messages are similar, from an engineering abstraction perspective, doesn't matter. 

NIP-46 was designed to specify the negotiation between a wallet and a signing device. This one is between a wallet and a lightning node. These are different needs. The new URI scheme helps to make sure they don't mix, as it is possible to have a wallet, a signing device, AND a lightning API at the same time. 

This is not a replacement for NIP-46. Both NIPs will have their "supported methods", with extensive descriptions of what they accept and don't accept.

tiero:
> the negotiation between a wallet and a signing device. This one is between a wallet and a lightning node.
I see.  In the end, the "request and response" semantic is useful for different applications. 
I feel that my intuition is that 46 & 47 could share a common approach of doing generic RPC over Nostr, which seems to be the common trait.

> This is not a replacement for NIP-46. Both NIPs will have their "supported methods", with extensive descriptions of what they accept and don't accept.

It's indeed perfectly fine to have additional NIPs to additional supported methods.

tiero:
> The new URI scheme helps to make sure they don't mix

What do you think of using the `metadata` field to communicate intent? ie. type (either 46 or 46) and eventual `secret`, to have a single URI format?

vitorpamplona:
> @vitorpamplona What do you think of using the metadata field to communicate intent? ie. type (either 46 or 46) and eventual secret, to have a single URI format?

What's next? Are we going to merge with `nostr:` schemas as well and just add a `metadata` field? I don't think it's a good idea.

Apps register schemes with their OSs by role. Each scheme represents a semantical behavior the app is supporting and willing to execute. Separating 46 and 47 makes good sense because some apps only want to be a signing-less client, others only want to be signing devices and others only want to be ln services. And when one app can do all 3, it better let each of those roles crystal clear when users select their default app to manage each role on the OS-controlled popup. Different schemes help them do just that.

> My intuition is that 46 & 47 could share a common approach of doing generic RPC over Nostr, which seems to be the common trait.

A common **approach**, yes. I would love to see both following the same pattern of message types. But they should offer very different methods, message types, and semantical roles. So different that complexity might go up just to support said merge of NIPs. 

I don't know about you, but I don't want to see a 10-page-long NIP where no client is able to quickly support all methods and declare full compliance with the NIP. Nostr works because it's simple and anyone can incrementally declare compliance to individual NIPs as they improve their clients. Trying to merge it all in one is a disservice to that vision.

tiero:
Alright, thanks for the extra context @vitorpamplona and fine for me to split things up. I am glad you recognize "request & response over nostr" is a common need.

In the end, my only issue may be for the `wallet` and `connect` parts in the name and the prefix, as this is basically only useful for custodial lightning services and technical users that run a reachable server on top of their standalone node.

To me, "wallet" is much more than lightning nodes and is a bit misleading imho as not all wallets are lighting nodes. Do you think a different prefix/name can be considered at this point?

47.md
kiwiidb: It acts like a proof of payment and I think Will said that it would allow for an update to NIP-57 where the client updates the Zap event.

47.md
kiwiidb: Yes. Nostr allows for native responding through the `e` tag, so why not use the `e` tag?

47.md
```diff
@@ -0,0 +1,147 @@
...
+* **client**: Nostr app on any platform that wants to pay Lightning invoices
+* **wallet service**: Nostr app that typically runs on an always-on computer (eg. in the cloud or on a Raspberry Pi).
+
+## Events
+
+There are two event kinds:
+- `NIP-47 request`: 23194
+- `NIP-47 response`: 23195
+
+Both the request and response events SHOULD only contain one `p` tag, containing the public key of the **wallet service** if this is a request, and the public key of the **client** if this is a response.
```
kiwiidb, line 24: I still don't get why the response should not contain an `e` tag to refer to the event it is responding to?

47.md
```diff
@@ -0,0 +1,147 @@
...
+    "cmd": "pay_invoice", // command, string
+    "data": { // data, object
+        "invoice": "lnbc50n1..." // command-related data
+    }
+}
+```
+
+Response:
+```jsonc
+{
```
kiwiidb, line 40: The response should have the same structure as NIP-46:

```suggestion
{
  "id": <request_id>,
  "result": <anything>,
  "error": <reason>
}
```

47.md
```diff
@@ -0,0 +1,147 @@
...
+```
+
+### `create_invoice`
+
+Description: Requests creation of an invoice.
+
+Request:
+```jsonc
+{
+    "amount": 1000, // amount in msat, int
```
kiwiidb, line 107: Use sats instead of msats.

47.md
```diff
@@ -0,0 +1,147 @@
...
+```jsonc
+{
+    "invoice": "lnbc50n1..." // BOLT11 invoice, string
+}
+```
+
+Response:
+```jsonc
+{
+    "preimage": "0123456789abcdef..." // preimage after payment, string
```
kiwiidb, line 135: Should use NIP-46 payload structure.

kiwiidb (changes requested):
Overall my main feedback is that we should make the JSON structure the same as NIP-46 and not add too much bloat to the NIP.

tiero:
> the JSON structure the same as NIP-46

Actually, I agree with @vitorpamplona to account for typed data structures and polymorphic objects.
Currently, NIP46 has a JSONRPC-ish style, but there was little design work on that part TBH as we focused on many other things. 

I agree JSON/payloads share the same format between nips, I am happy to amend NIP-46 to follow anything that comes out of NIP-47, instead of the other way around.

vitorpamplona:
New learning: `Balance` should **not** be an RPC call. 

It should be a regular replaceable event. The wallet connects service only signs them when the balance changes. Nostr client simply subscribes to those events. 

It doesn't make any sense to request balance via RPC at every second just to have the most up-to-date information.

This is the danger of a generic interface. We start thinking everything fits the mold of an RPC call...

bumi:
great discussions in here! I just wanted to also add my overall thoughts even though the discussion progressed already.
Initially I was toally in favor of some generic RPC interface that can be easily extended for various cases. But thanks to @vitorpamplona I came to appreciate the simplicity and the idea in Nostr of doing one thing and have it defined and easily implementable for any app. Nobody needs to learn a big spec and it is immediatelly clear how/what works. 

Having this as a generic RPC call requires now the app and the wallet provider to not only agree on the Nostr events as communication but also on actualy RPC methods and details. 
As already mentiond here we currently do not have any RPC spec in the LN world that defines the methods and responses. And this also should not be in a NIP as it is independent of the protocol here. 
(even a balance call is different in any node implementation - what's even THE balance?) 

If we go down that road then where do we stop? Do we specify calls for paying an invoice, requesting an invoice, fetching the blance - but what is with doing a keysend payments or getting channel information or other LN details?
Some of the calls might not work with all wallet providers. How do apps know what calls are supported? 

So I think we should have specific NIPs for the most basic features (e.g. here sending a payment) and then have an universal RPC one for more generic things that can and should be extendible indpendently of a NIP. 

So in total: I think we should keep it simple and specific the complexity to make a universal LN interface is too big and will also hinder adoption. And stopping half way with only a 3-4 calls is not worth it. 
But I have no strong opinion, all that matters to me is adoption and making it simple for developers.

vitorpamplona:
@Semisol, @bumi and I were talking and we think we should remove the balance spec from this NIP and make it purely about a payment action. Keeping it as simple as possible. 

New NIPs can add more active action calls to be executed by the wallet service (which would use the RPC spec) as well as more passive behaviors that a wallet service can push into the relay when it wishes to do so (which could use a non-ephemeral event type). 

The conclusion is that we don't have a good use case for wallet information as of now and we should not build for something we don't have good clarity for.

47.md
benthecarman: definitely should be in msats, no reason to lose granularity

kiwiidb:
To clarify why I think the scope of the NIP should be as limited as possible: Nostr is first and foremost a social media protocol. A rather tight integration with Lightning has developed, which is nice, but I think everyone agrees we should be careful to not add too much Lightning specific things in the NIPs. 

The reasoning for the NIP, as described in our [original draft](https://github.com/getAlby/nips/blob/master/47.md#rationale) was to make it easier to **pay zaps**. We identified a *clear need* for this across clients. I feel there is a much less clearer need across clients to do things like show balances (my bitrefill app also doesn't show me my LN wallet balance), create invoices (we already have LN Addresses for that, like them or not), do channel management or whatever. If you want to do these things over a Nostr Relay, that's awesome and a great idea, but I do not think it should be in a NIP. It should have it's own seperate spec for people that want to build wallet or node UI's.

vitorpamplona:
> I'm down with that except get_info. I want to keep that since it specifies what extensions are supported -- for now we can have that be barebones and only send the supported extensions.

But then I think `get_info` should not be a command. It should be a simple replaceable event the service can sign upon start and leave it on the relay so that clients can filter for. There is no need to create a request/response API for this.

bumi (commented):
if we go with the RPC flow: maybe we can adopt wordings from the other NIP and/or follow JSON-PRC wordings. 

and I actually also see this then on two levels:
the base level defines the events for the RPC calls and the second layer defines the methods and request/response data of different calls.
The base level is fixed, but the methods can be extended.

Right now it seems to me we mix this. Should commands really be defined here? How will this be expanded then? 
isn't a NIP for the concept and then NIPS for individual methods better? (the RPC calls potentially don't even need to be Nost specific but are rather lightning specific and could be defined elsewhere)

jb55:
> @bumi: I also still think that NIP-57 should events published by the clients

nip47 makes this much more doable, and the zapper key could just be the user's pubkey.

arcbtc:
I like the idea so much, I made same proposal here after clearly not searching nips properly https://github.com/nostr-protocol/nips/pull/433.

I would suggest term `preimage` being a generic `payment_hash`, which could be a preimage if using bitcoin, as it makes the nip more useful for using outside of bitcoin, ie a traditional Stripe invoice could be paid.

I would also suggest adding `check_payment`

In my proposal:

```
### Checking a payment in/out

`service client` sends `check_payment` event:

{
   "check_payment":{
      "payment_hash":"<string>"
   }
}

`funding source client` checks the pubkey has permission, checks the payment status and sends `payment_status` event:

{
   "payment_status":{
      "paid":"<bool>"
   }
}
```

kiwiidb:
I have made a PR to this PR: https://github.com/nostr-protocol/nips/pull/437

I think I have incorporated all feedback that was expressed here. I hope we can reach consensus and get this merged soon.

kiwiidb (approved):

thesimplekid (approved):

ArcadeCityMayor:
ACK

Implemented NIP-47 demo [here](https://twitter.com/TheArcadeApp/status/1653469210624835606), all worked as expected. Amazing UX!

PR #408, opened 2023-04-01
Add „offset“ filter attribute.

Egge7:
This NIP introduces an optional filter attribute that will offset requested stored events by X. This will allow clients to retrieve paginated results (which will reduce load and data usage) without handling multiple timestamps and therefore reduce complexity.

### DIFF
--- a/xx.md
+++ b/xx.md
@@ -1,0 +1,33 @@
+NIP-XX
+======
+
+Offset Filter Attribute
+-------------------
+
+`draft` `optional` `author:egge7`
+
+Relays may support an additional Filter attribute `offset`, which combined with `limit` allows Clients to get ranges of events.
+
+### Motivation
+
+In order to reduce data usage when fetching historic data, clients need to paginate their subscriptions. With the currently available filter attributes this can become quite complicated, especially when handling subscriptions to multiple relays with different states. By introducing the `offset` attribute we allow Clients to easily build paginated subscriptions.
+
+### Example
+
+Page 1
+```json
+{
+  "authors": [<followedPubkeys>],
+  "kinds": [1],
+  "limit": 25
+}
+```
+Page 2
+```json
+{
+  "authors": [<followedPubkeys>],
+  "kinds": [1],
+  "limit": 25,
+  "offset": 25
+}
+```


fiatjaf:
We already have `since` and `until`. Also `offset` is heavy on databases as the first n rows have to be counted before the next are returned.

staab:
`offset` also introduces the possibility of race conditions, where records can be missed or duplicated if the dataset changes between requests. If we're going to add pagination (which as a client dev I certainly would appreciate), we should use an cursor-based approach, which in nostr's case would be very easy, a simple `"after": "<event-id>"` would probably do it.

I've also considered ways to solve this problem via a multiplexer, where the multiplexer buffers events and implements pagination on top of multiple relay event streams.

Egge7:
> We already have `since` and `until`. Also `offset` is heavy on databases as the first n rows have to be scanned through and counted before the next ones are returned.

Since and until can be used to build pagination, but it introduces a lot of complexity when interacting with many different relays that return different events (as you need to keep track of the events timestamps per relay).

And it would be great to learn more about the performance impact. Offset seems to be a very common filter option in databases, no?

Egge7:
> `offset` also introduces the possibility of race conditions, where records can be missed or duplicated if the dataset changes between requests. If we're going to add pagination (which as a client dev I certainly would appreciate), we should use an cursor-based approach, which in nostr's case would be very easy, a simple `"after": "<event-id>"` would probably do it.

Instead of after we should probably use "until" as a cursor, because you can not assume that a relay has the specific event id from "after"

staab:
>Instead of after we should probably use "until" as a cursor, because you can not assume that a relay has the specific event id from "after"

You're right, I didn't think this through. until/after/limit support pagination already, the complexity comes from dealing with multiple relays

Egge7:
> You're right, I didn't think this through. until/after/limit support pagination already, the complexity comes from dealing with multiple relays

Exactly! If we would have "offset" clients could have a single "until" cursor, then get pages of events from relays without having to worry about timestamps. It would streamline the process drastically.

staab:
Ok, I've had the same thought, but I'm not sure it would, because you'd still have to keep track of the offset (rather than the earliest event returned) for each relay, and not ask that relay for more when fetching the next page, otherwise you would end up paginating by the same number of records regardless of event frequency. So with an offset of 10, an relay with a matching event every minute would give you something 10 minutes old, while a relay with events every hour would give you something 10 hours old. So for proper pagination you still need to look at the timestamps.

I tried at one point to implement this in Coracle, but it meant sending a different filter to each relay, which broke the pool abstraction entirely so I gave up on it. But I now use extremely complicated heuristics to avoid displaying very old events to the user, which only work 90% of the time, so I'll likely re-visit. Anyway, I'm not sure `offset` would solve this problem.

Egge7:
> Ok, I've had the same thought, but I'm not sure it would, because you'd still have to keep track of the offset (rather than the earliest event returned) for each relay, and not ask that relay for more when fetching the next page, otherwise you would end up paginating by the same number of records regardless of event frequency. So with an offset of 10, a relay with a matching event every minute would give you something 10 minutes old, while a relay with events every hour would give you something 10 hours old. So for proper pagination you still need to look at the timestamps.
> 
> 
> 
> I tried at one point to implement this in Coracle, but it meant sending a different filter to each relay, which broke the pool abstraction entirely so I gave up on it. But I now use extremely complicated heuristics to avoid displaying very old events to the user, which only work 90% of the time, so I'll likely re-visit. Anyway, I'm not sure `offset` would solve this problem.

You are absolutely right. Even "offset" will not help when you trying to build a paginated feed, as soon as one relay responds with data that is way older than another relays response... back to the drawing board!

It's good to hear that others are trying to wrap their head around this as well. I did the same thing, but also didn't finish.

PR #412, opened 2023-04-04
Add account_index for infinite NIP-06 key generation

sethforprivacy:
I propose a change to NIP-06 that would allow the generation of infinite root keys by using the `account` field of the 5-level path used to derive Nostr keys under BIP32.

The current 5-level path used for NIP-06 is `m/44'/1237'/0'/0/0`, and only generates one private key. By changing it to `m/44'/1237'/<account>'/0/0`, we can generate multiple keys from a single parent bitcoin seed.

### DIFF
--- a/06.md
+++ b/06.md
@@ -8,8 +8,8 @@ Basic key derivation from mnemonic seed phrase
 
 [BIP39](https://bips.xyz/39) is used to generate mnemonic seed words and derive a binary seed from them.
 
-[BIP32](https://bips.xyz/32) is used to derive the path `m/44'/1237'/0'/0/0` (according to the Nostr entry on [SLIP44](https://github.com/satoshilabs/slips/blob/master/slip-0044.md)).
+[BIP32](https://bips.xyz/32) is used to derive the path `m/44'/1237'/0'/0/<address_index>` (according to the Nostr entry on [SLIP44](https://github.com/satoshilabs/slips/blob/master/slip-0044.md)).
 
-This is the default for a basic, normal, single-key client.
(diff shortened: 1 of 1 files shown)


sethforprivacy:
Happy to modify it to account instead of address_index, that's a very easy change on the implementation side as well.

sethforprivacy:
Thanks for the quick feedback, @benthecarman and @Semisol! Changes have been pushed in [59e5195](https://github.com/nostr-protocol/nips/pull/412/commits/59e51957842a46838901c5e48242cd3a9cfd3185)

benthecarman (approved):

RandyMcMillan:
I was originally using "deterministic aliases" for this stuff...

Appending an additional "password" on to the private key and rehashing could be a part of this NIP - this would be similar to how HWWs add the additional entropy.

sethforprivacy:
> I was originally using the term "deterministic aliases" for this stuff...
> 
> Appending an additional "password" on to the private key and rehashing could be a part of this NIP - this would be similar to how HWWs add the additional entropy.

The other approach to this can certainly be to use passphrases to derive new keys, but that introduces an additional SPOF that we didn't think made for great UX. Because NIP-06 leverages BIP 32 under the hood, however, you don't need anything in this spec to properly derive Nostr private keys using account 0 and passphrases (or any combination of account_index and passphrase)!

sethforprivacy:
Tested and confirmed that keys derived using an incremented `account` instead of `address_index` work perfectly in Amethyst.

Excited to see this more broadly available soon!

Egge7:
> Tested and confirmed that keys derived using an incremented `account` instead of `address_index` work perfectly in Amethyst.
> 
> 
> 
> Excited to see this more broadly available soon!

Will add this soon to Current. We use NIP06 by default, so this will be perfect for sub-accounts.

PR #417, opened 2023-04-07
NIP-81 Modeling Files / Filesharing

lovvtide:
Here's my proposal for censorship resistant file sharing on nostr. The three main goals of this proposal:

- To define standard metadata to describe a file in the nostr ecosystem
- To make it possible for relays to index file metadata, thereby enabling clients to query for files
- To mitigate the problem of dead links and generally increase resilience/censorship-resistance by utilizing BitTorrent as a fallback when there are no servers hosting a file

I'm currently building a file manager app the implements this spec, so I should be able to show that soon as well. What do you guys think?

Here's the proposal https://github.com/lovvtide/nips/blob/NIP-81/81.md

### DIFF
--- a/81.md
+++ b/81.md
@@ -1,0 +1,103 @@
+NIP-81
+======
+
+Modeling Files
+--------------
+
+`draft` `optional` `author:lovvtide`
+
+This NIP introduces a "file" event (kind `2001`) for modeling files on nostr in a way that is simple and extensible. The file event is served by relays like any other event and contains metadata that clients can use to download, verify, and display the file's data. Specifically, this proposal:
+
+- Defines standard metadata to describe a file in the nostr ecosystem
+- Makes it possible for relays to index file metadata, thereby enabling clients to query for files
+- Mitigates the problem of dead links and generally increases resilience/censorship-resistance by utilizing BitTorrent as a fallback when there are no servers hosting a file
+
+
+## Event Model
+
+``` 
+{
+    "content": <description_of_file>,
+    "kind": 2001,
+    "tags": [
+        ["x", <hash_value>, <hash_algorithm>],
+        ["m", <mime_type>],
+        ["size", <byte_length>]
+        ["name", <file_name>],
+        ["url", <url_of_file>]
+    ],
+    ...
+}
+```
+
+The kind `2001` event makes use of three required tags (containing the identifying hash(es), file type, and file size) and two optional tags (containing the file name and url(s) to download the file). Each tag is explained as follows:
+
+- `x` (required) - Hash of the file and the name of the hashing algorithm used to generate the hash. Each event MUST contain at least one `x` tag with the `sha256` hash of the file. The event MAY also include additional `x` tags with different hashes of the same file (such as the torrent `infohash`). For any `x` tag, if the name of the hashing algorithm is omitted, it should be assumed to be a `sha256` hash.
+- `m` (required) - Lowercase mime type of the file as per [the IANA's official list](https://www.iana.org/assignments/media-types/media-types.xhtml)
+- `size` (required) - Integer byte length of the file, formatted as a string
+- `name` (optional) - File name set by creator of event (recommended to improve UX when listing files on the client)
+- `url` (optional) - Link to download the file's data. Note that an event MAY contain multiple `url` tags if the file is being hosted by multiple servers.
+
+
+## Example Event
+
+```
+{
+    "content": "This is a human readable description of the file",
+    "kind": 2001,
+    "tags": [
+        ["x", "719171db19525d9d08dd69cb716a18158a249b7b3b3ec4bbdec5698dca104b7b", "sha256"]
+        ["x", "a19a70552cc801415a6071993c04b3ab21572438", "infohash"],
+        ["m", "video/mp4"],
+        ["size", "26685219"]
+        ["name", "myvideo.mp4"],
+        ["url", "https://cdn.example.com/foo/bar/a19a70552cc801415a6071993c04b3ab21572438.mp4"]
+    ],
+    "created_at": 1679552101,
+    "pubkey": "fbd992dbccdd4186f4ef25b7d72ea47387a59227493f8569b17963afae4e4d33",
+    "id": "faea2a5dbe12c877c1201e91ec909e3161a1ba8714fdca4d3077f6d4b0780191",
+    "sig": "10c0d634ea0b9ae085205c3951367da784268a0cdf4a18b969aba7c06b1a59abf96b7f5f4e2d3791389dff71ff5527c33e3405e8decad9b3e4419406124dc46a"
+}
+```
+
+In this example, you might notice that there are two `x` tags. Both are hashes of the file. To ensure interoperability, this spec requires the presence of a `sha256` hash, while the `infohash` (and any number of additional hashes) may optionally also be included.
+
+Note that the `content` field MAY contain a human-readable description of the file.
+
+
+## Querying File Events
+
+The reason why `x` (file hash) and `m` (file type) are single letter tags is to enable clients to pull file events from relays using generic tag queries (see [NIP-12](https://github.com/nostr-protocol/nips/blob/master/12.md)).
+
+For example, if a client knows any hash of a file it can fetch the corresponding event with a request filter like this:
+
+`{ #x: [ "a19a70552cc801415a6071993c04b3ab21572438" ] }`.
+
+As another example, suppose a client wants to fetch events for all the videos a relay knows about. The request might look something like this:
+
+`{ #m: [ "video/mp4", "video/wav", ... ] }`
+
+In general, clients being able to ask relays for exactly which files they're interested in is a big improvement on the current situation where clients have to download all the events and parse out links to files.
+
+
+## Discussion
+
+### Why BitTorrent?
+
+The BitTorrent protocol (created in 2001) is an old, stable, and widely supported means of decentralized file sharing. One of its biggest drawbacks, however, is that it lacks a way to verify the identity of the creator of a torrent. Nostr fixes the problem of verifying identity. It's therefore reasonable to think that BitTorrent + Nostr could be a very powerful combination.
+
+A torrent `infohash` is a special hash defined by the BitTorrent protocol used to uniquely identify a torrent. Clients can use this hash (along with a list of popular torrent trackers) to construct a magnet link and get the file's data from peers. In fact, some nostr clients have already implemented [WebTorrent](https://github.com/webtorrent/webtorrent) which supports loading torrents directly in the browser.
+
+One peculiarity of how the `infohash` is computed is that its value depends not only on the contents of the file, but also on the `name` of the torrent chosen by the torrent creator. When computing the `infohash` of a file, clients SHOULD, for the sake of consistency, ensure that the `name` passed to the torrent constructor is the same value as the `name` tag in the nostr event.
+
+For the same reason, applications should use the `sha256` hash (not the `infohash`) for deduplicating and uniquely identifying files.
+
+
+### What About IPFS?
+
+IPFS is not included as part of this spec due to its performance issues, unstable API, and de-facto reliance on centralized gateways. Still, if the creator of an event wanted to include the IPFS hash in an event, there's nothing in this spec that would stop him from doing that.
+
+
+### Extensibility
+
+It should be easy to extend this proposal in the future by standardizing additional tags and/or hashing algorithms. For now it seems best to start with a simple spec that can be easily implemented.


v0l:
Same as https://github.com/nostr-protocol/nips/pull/337?

lovvtide:
> Same as #337?

I investigated using NIP-94 for my filesharing app but realized there some pieces missing:

1. In NIP-94, neither the file hash nor the file type are single letter tags, and thus (according to NIP-12) those values are not indexed by relays. This is a big issue because it means that when using NIP-94 clients cannot effectively query relays for media.
2. NIP-94 is missing metadata for the file name and the file size which are important for UX on the client, and so the client can know how big a file is before you start downloading it.
3. NIP-94 has no mechanism for resilence/censorship-resistance or dealing with dead links when a server stops hosting the file

This is an alternative proposal that fixes those problems and provides greater extensibility.

v0l:
I think all of the issues you raised are valid and should be included in NIP-94

fiatjaf:
This seems to be a very well-thought proposal, we should merge it with NIP-94 for sure, or vice-versa.

fiatjaf:
I think we can summarize both this and the other proposals as just pointer to files with optionally multiple ways of accessing these files and optional metadata. We just have to agree on the names of the tags for the metadata.

- `url` -> http or magnet (multiple)
- `e` -> NIP-95 event that contains the actual file, if anyone is that crazy (optional)
- `x` -> sha256
- `t` -> torrent infohash
- `size` -> bytes (optional)
- `secret` -> key and nonce for AES encryption (optional)
- `blurhash` -> for cosmetic purposes (optional)

What do you think? @frbitten

lovvtide:
> We just have to agree on the names of the tags for the metadata

Some thoughts/questions:

- I think there needs to be an `m` tag for the file's mime type so that clients can query based on type of file (the `type` tag was also suggested by NIP-94)
- I think the `size` tag should be required and not optional. If the client is able to obtain the hash of the file, they also presumably can read the size of the file. Is there any situation where the client would not have access to that value when creating the event? File size seems quite important for both the client and also potentially for hosting providers, e.g. for allocating the proper amount of storage
- Having at least an optional tag `name` seems like a good idea for the sake of UX when listing/managing files on the client (I'm imagining a dropbox sort of interface), so the user doesn't see a list of files that are all named "Untitled"
- If the `url` tag was allowed to contain http OR a magnet link, wouldn't that put an additional burden on clients to "detect" somehow what kind of link if was before loading that link? I think if the magnet link is going to be included it should therefore be its own optional `magnet` tag. Or perhaps a better alternative would be to not include the magnet link at all, since it can be easily constructed from the other metadata in the event according the [magnet link format](https://en.wikipedia.org/wiki/Magnet_URI_scheme#Format) like `magnet:?xt=urn:btih:${infohash}&dn=${name}&ws=${url}`
- I wonder, what is the benefit of having the infohash be its own `t` tag and not just another `x` tag with a marker `infohash` as the third element of the array? I agree that `sha256` should definitely be the default hash (which is why, as I wrote in this proposal, `sha256` is required and any `x` tag without a marker is automatically assumed to be `sha256`). The main benefit I see for having only an `x` tag (instead of an `x` tag specifically for `sha256` and a `t` tag for `infohash`) is that having a single indexed tag for all hashes would make the event type easily extensible for applications that for some reason needed to use a different hash *and* need to be able to query using that hash. This would not add any burden on clients that only care about the default since an `x` tag with `sha256` is guaranteed to be present anyway and any clients can just safely ignore other `x` tags with markers they don't know about. Put another way — I think it would be worth it to avoid having to think about modifying this NIP every time some app wishes they could use a new indexed identifier for a file

fiatjaf:
> I think there needs to be an m tag for the file's mime type so that clients can query based on type of file (the type tag was also suggested by NIP-94)

Sorry, I forgot about that one. Modified above.

> I think the size tag should be required and not optional.

What if you just have a file hosted somewhere and want to use a Nostr filesharing client to broadcast it -- and your file is 10GB? You'll have to download the file to check the size?

I don't know about the magnet stuff, remains to be seem, we can't really control that.

Also it's not a big deal to edit the NIP every time. People will certainly add new non-specified tags to the events anyway if the usage catches on. It's better to differentiate tags according to usage, I think. We lose nothing from doing that.

lovvtide:
> What if you just have a file hosted somewhere and want to use a Nostr filesharing client to broadcast it -- and your file is 10GB? You'll have to download the file to check the size?

I suppose in that case (you don't have the file locally to check the size) the same logic would apply to file hash and mime type. It seems likely that whatever source you are getting that metadata from would also provide the file size.

> Also it's not a big deal to edit the NIP every time. People will certainly add new non-specified tags to the events anyway if the usage catches on. It's better to differentiate tags according to usage, I think. We lose nothing from doing that

Ok yeah. If it's no big deal to edit the NIP then I agree it's better to use tags to differentiate usage.

fiatjaf:
> I suppose in that case (you don't have the file locally to check the size) the same logic would apply to file hash and mime type. It seems likely that whatever source you are getting that metadata from would also provide the file size.

Good point.

I guess the file size and mimetype can be fetched from a HEAD request in most cases.

v0l:
`url` -> `r` ?
`t` is already used for hashtag?

blurhash could also be merged into `x` if you use the hash name in the tag `["x", "<some-blurhash>", "blurhash"]`

`x` may contain a `sha256`, `blurhash` or `btih` (or all)

v0l:
@lovvtide If you don't use the full magnet link then you lose out on extra data like `xs` and `ws` tags which make it basically unusable on web

lovvtide:
> url -> r ?

The reason I thought to use `url` instead of `r` is to save relays from having to index that tag. Is there a reason the url needs to be indexed? I am aware that `r` is standardized elsewhere as referring to a url, but as I recall the reason that convention was standardized as a single letter tag is for the sake of building a commenting system where clients can query for events based on the url of the page. Since the url in this proposal refers to just the source of a file (and not really a page per se) I wonder if it makes sense to retain that usage, or would it be better to go with non-indexed `url` to lessen the load on relays?

> t is already used for hashtag?

Ah yes, should probably change that then. Maybe `i` for "infohash"?

> blurhash could also be merged into x if you use the hash name in the tag ["x", "<some-blurhash>", "blurhash"]
> x may contain a sha256, blurhash or btih (or all)

That was my original idea, but as @fiatjaf pointed out it might be better to have a separate tag for each type of hash, i.e. use tags to differentiate usage instead of markers

> If you don't use the full magnet link then you lose out on extra data like xs and ws tags which make it basically unusable on web

Well `xs` (webseed) would just be the value of the `url` (or `r` ... whatever we are calling it) right? As for `xs`, I did not know about that — if there are clients relying on that it does make sense to include it somehow. I suppose `magnet` could be its own tag, but I do think that even if there is a `magnet` tag the infohash and file name should still have their own tags so you can query using infohash and so that the client does not have to parse the magnet link just to display basic metadata. Maybe it's simplest to just have an optional `xs` tag — what do you think?

frbitten:
As I already put it in the NIP-94, I don't see a problem in changing the tags to be single letter and indexable. I didn't put them as indexable on purpose precisely so as not to overload the relays with indexes that we are not even sure if they will be used. I don't think hash lookup is viable. It will just be a way to confirm that the download was done correctly. In most cases, I imagine you won't even have this information at the event.

The m tag really is a good thing to be indexable for searches by type. File name in NIP-94 I left open to use other tags that already exist in the protocol such as "subject" (NIP-14) instead of creating another tag.

They can suggest changes in the NIP-94 that I accept them. Any other tag that you deem useful can be added. I tried to be as minimalist as possible to avoid limiting its use. There are numerous image metadata tags that can be included for example.

frbitten:
> Acho que podemos resumir esta e as outras propostas como apenas um ponteiro para arquivos com várias formas de acessar esses arquivos e metadados. Nós apenas temos que concordar com os nomes das tags para os metadados.
> 
> * `url`-> http ou ímã (vários)
> * `e`-> Evento NIP-95 que contém o arquivo real, se alguém for louco (opcional)
> * `x`-> sha256
> * `m`-> mimetype
> * `t`-> infohash torrent
> * `size`-> bytes (opcional)
> * `secret`-> chave e nonce para criptografia AES-GCM (opcional)
> * `blurhash`-> para fins cosméticos (opcional)
> 
> O que você acha?@frbitten

@fiatjaf 
In NIP-95 it was suggested to have a specific header for it. To avoid mixing file sharing with data storage. That's why in NIP-95 I defined a specific header event for it.
Because the e tag may not be clear when it is the definition of a NIP-95 event or a simple reference to another existing event.

For the NIP-94 to work, it should be limited to only having one event-id in the "e" tag and it must be a NIP-95 event.

I don't know what would be the best approach in this case. What do you think?

v0l:
> The reason I thought to use url instead of r is to save relays from having to index that tag.

Yea maybe its not necessary, in NIP-94 they use `r` tag

> That was my original idea, but as @fiatjaf pointed out it might be better to have a separate tag for each type of hash, i.e. use tags to differentiate usage instead of markers

I think it makes more sense to use just `x`, we have limited single char tags (which are indexable) and we will need all the different hash values indexed. Also they are hashes so there is almost no chance of any collision with some other hash. So you're basically guaranteed to get the correct result you're looking for when you query the relay by hash (maybe except blurhash, not sure how much data they store there) 

> Well ws (webseed) would just be the value of the url 

Maybe, trying to imagine a scenario where you would want some other URL that would be different than the web seed...

> Maybe it's simplest to just have an optional xs tag — what do you think?

Idk, you remove any of the flexibility of just having a plain `magnet:` link with custom trackers or any other supported options in the magnet link, you're not really "supporting" magnet links if you don't support all of the options.

Here is a random image uploaded to void.cat without `xs` and it doesnt seem to work in `Transmission` or on `instant.io` 
```
magnet:?xt=urn:btih:daaa43e9bf35f4fe95d4174d926e7fa9544c1e46&dn=logobtc_purpletrasparent.png&tr=wss%3A%2F%2Ftracker.btorrent.xyz&tr=wss%3A%2F%2Ftracker.openwebtorrent.com&tr=udp%3A%2F%2Ftracker.opentrackr.org%3A1337%2Fannounce&tr=udp%3A%2F%2Ftracker.openbittorrent.com%3A6969%2Fannounce&tr=http%3A%2F%2Ftracker.openbittorrent.com%3A80%2Fannounce&ws=https%3A%2F%2Fvoid.cat%3A443%2Fd%2FHi1YEB7LoBtnsnoHFVGkzd
```

You need the torrent file in order to have the piecehash data, maybe this is a problem for void.cat to solve by hosting the metadata on trackers too idk, what do you think?

fiatjaf:
@v0l 

> we have limited single char tags

But we have enough, right? If we consider that each kind has different meanings for each letter (besides some canonical ones like `e`, `p` and `a` all other tag names are contextual I think) I think it is ok to use different tag names for each kind of hash, and we shouldn't index blurhash.

About the magnet I don't know, I think having the full contents of the magnet split between tags looks good to me, but having the full magnet too.

Ideally file hosts (yes, voidcat.io) that provide webseed should also provide the torrent file at the same path, so while constructing the magnet we could just use the same URL pattern for the `xs` and clients can download it from there.

v0l:
> Ideally file hosts (yes, voidcat.io) that provide webseed should also provide the torrent file at the same path, so while constructing the magnet we could just use the same URL pattern for the `xs` and clients can download it from there.

This is how void.cat does it, do you think that searching by infohash will be a thing? Will searching by any hash ever be a thing? I dont know, seems like just putting the full magnet link is a better option

> I don't think hash lookup is viable 

Im in agreement with @frbitten

frbitten:
In what cases will someone have a hash and want to look up the file that has that hash? Either someone gave me the hash to search or I already have the file and I want to see if it is already shared. If someone passed the hash they might as well pass the event id. Knowing if the file has already been shared can be a feature external to NOSTR that the client or relay implements in another way.

  What if we have the same file shared on more than one link. Are we going to have the same hash in multiple events, how will the indexing be in this case? I don't know how this databases are to know if this could be a problem or not.

lovvtide:
> Idk, you remove any of the flexibility of just having a plain magnet: link with custom trackers or any other supported options in the magnet link, you're not really "supporting" magnet links if you don't support all of the options.

@v0l This is a really good point

> I think having the full contents of the magnet split between tags looks good to me, but having the full magnet too.

@fiatjaf This seems like the best solution. It may result in duplicated information sometimes, but that's a small price to pay for clients not having to parse magnet links to read metadata while at the same avoiding any constraints on the usage of the full magnet spec.

> In what cases will someone have a hash and want to look up the file that has that hash?

@frbitten I'm thinking of the case where you have the file and want to query relays for event(s) containing the hash of the file to find out who published the event and/or verify it's authenticity. I think this is quite an important use case actually. For instance, suppose that all I have is a link to a file (or someone sends me a file as an attachment, or perhaps I downloaded the file from a torrent client that doesn't know about nostr) and, in any case, I want to verify that the file can be trusted. I want a nostr client to be able to download the file, compute the hash, and then query relays using that hash to find out if anyone has created an event the vouches for the authenticity of that file I just downloaded. With deep fakes and AI and all of that I imagine in the future this capability to "verify" media will only become more important. I want it to be possible to build a "trusted torrent client" using nostr that is interoperable with existing torrent clients but uses nostr as a way to get information about credibility/authenticity of files.

> Are we going to have the same hash in multiple events, how will the indexing be in this case? I don't know how this databases are to know if this could be a problem or not.

Won't be a problem at all, for the same reason that many events contain, for example, the same `e` tag and that's fine. If a client queries relays for a hash they will get back an array of events that contain a tag with that hash.

frbitten:
@lovvtide But a file already published does not guarantee its reliability. You will have to do a social analysis of who published it to see if it is a trusted user or not.

If all this is automatic by the application, I understand the usefulness. But I don't see a common user calculating a hash of a file to search if it already exists in the nostr.

But if it's important for your project, I don't see a problem with being an indexable tag.

lovvtide:
> But a file already published does not guarantee its reliability. You will have to do a social analysis of who published it to see if it is a trusted user or not.

Yes, that's an important distinction. Being able to look up the author of an arbitrary file is not sufficient, on its own, to being able to trust the file.

However (regardless of how people and up doing the social analysis part) it is still a necessary component of any such system.

Like, I'm thinking of the file hash being sort of like a "primary key" that makes nostr more easily interoperable with other ways of distributing files, i.e. leaving the door open for applications in different contexts to talk to each other about that file by referencing that hash without all the applications having to know about nostr.

fiatjaf:
This functionality has been incorporated by NIP-94.

PR #420, opened 2023-04-07
make NIP-15 mandatory

sepehr-safari:
sending an `EOSE` message from the relay must be mandatory not optional.

### DIFF
--- a/15.md
+++ b/15.md
@@ -4,18 +4,14 @@ NIP-15
 End of Stored Events Notice
 ---------------------------
 
-`final` `optional` `author:Semisol`
+`final` `mandatory` `author:Semisol`
 
-Relays may support notifying clients when all stored events have been sent.
+Relays MUST support notifying clients when all stored events have been sent.
 
-If a relay supports this NIP, the relay SHOULD send the client a `EOSE` message in the format `["EOSE", <subscription_id>]` after it has sent all the events it has persisted and it indicates all the events that come after this message are newly published.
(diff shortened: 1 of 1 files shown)


15.md
```diff
@@ -4,18 +4,14 @@ NIP-15
...
 ---------------------------
 
-`final` `optional` `author:Semisol`
+`final` `mandatory` `author:Semisol`
 
-Relays may support notifying clients when all stored events have been sent.
+Relays MUST support notifying clients when all stored events have been sent.
 
-If a relay supports this NIP, the relay SHOULD send the client a `EOSE` message in the format `["EOSE", <subscription_id>]` after it has sent all the events it has persisted and it indicates all the events that come after this message are newly published.
+The relay SHOULD send the client an `EOSE` message in the format `["EOSE", <subscription_id>]` after it has sent all the events it has persisted and it indicates all the events that come after this message are newly published.
```
pablof7z, line 11: if it's mandatory, it should be `MUST`

sepehr-safari: am I missing something?

pablof7z: The relay `SHOULD` send the  👉 The relay `MUST` send the

pablof7z: https://www.rfc-editor.org/rfc/rfc2119

```
MUST   This word, or the terms "REQUIRED" or "SHALL", mean that the
   definition is an absolute requirement of the specification.
```

```
SHOULD   This word, or the adjective "RECOMMENDED", mean that there
   may exist valid reasons in particular circumstances to ignore a
   particular item, but the full implications must be understood and
   carefully weighed before choosing a different course.
```

sepehr-safari: thanks for the explanation.
- [X] Done.

PR #421, opened 2023-04-07
NIP-15 must be mandatory

sepehr-safari:
sending an `EOSE` message from the relay `MUST` be `mandatory` not optional.

### DIFF
--- a/15.md
+++ b/15.md
@@ -8,7 +8,7 @@ End of Stored Events Notice
 
 Relays MUST support notifying clients when all stored events have been sent.
 
-The relay SHOULD send the client an `EOSE` message in the format `["EOSE", <subscription_id>]` after it has sent all the events it has persisted and it indicates all the events that come after this message are newly published.
+The relay MUST send the client an `EOSE` message in the format `["EOSE", <subscription_id>]` after it has sent all the events it has persisted and it indicates all the events that come after this message are newly published.
 
 Client Behavior
 ---------------
--- a/15.md
+++ b/15.md
@@ -4,18 +4,14 @@ NIP-15
 End of Stored Events Notice
 ---------------------------
 
-`final` `optional` `author:Semisol`
+`final` `mandatory` `author:Semisol`
 
-Relays may support notifying clients when all stored events have been sent.
+Relays MUST support notifying clients when all stored events have been sent.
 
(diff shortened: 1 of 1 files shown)


pablof7z:
just fyi, you don't need to create a new PR every time you need to make a change, just add a new commit to the branch 😉

#420

sepehr-safari:
Thanks man. it's too late here and I ran out of caffeine 😄

15.md
```diff
@@ -4,18 +4,14 @@ NIP-15
...
 ---------------
 
-Clients SHOULD use the `supported_nips` field to learn if a relay supports end of stored events notices.
-
-Motivation
-----------
-
-The motivation for this proposal is to reduce uncertainty when all events have been sent by a relay to make client code possibly less complex.
+Clients may use the `supported_nips` field to learn if a relay supports end of stored events notices.
+Clients MUST unsubscribe when they get an `EOSE` message from the relay.
```
fiatjaf, line 17:
```suggestion
```

fiatjaf, line 17: Actually it's better to just delete this entirely.

sepehr-safari: You mean all the client behavior section?
no need to unsubscribe and close connection on client side?

sepehr-safari: I've added your suggestions. but I think it's better to unsubscribe on EOSE, and create a new subscription later just if we want to refetch new events based on our client behavior. some clients might want to refetch only on hard page refresh. some might want to refetch every n seconds.

Egge7: > I've added your suggestions. but I think it's better to unsubscribe on EOSE, and create a new subscription later just if we want to refetch new events based on our client behavior. some clients might want to refetch only on hard page refresh. some might want to refetch every n seconds.

nostr events are transmitted in real time. Sounds like you want to force the standard HTTP request pattern over it, but nostr being based on WebSockets real-time is a feature not a bug.

fiatjaf:
Sorry for asking you to open this PR, but I've decided it would be simpler and clearer for everybody to just merge NIP-15 into NIP-01.

sepehr-safari:
> Sorry for asking you to open this PR, but I've decided it would be simpler and clearer for everybody to just merge NIP-15 into NIP-01.

great decision 👍

PR #427, opened 2023-04-09
added pubkey zap-tag

Egge21M:
PR https://github.com/nostr-protocol/nips/pull/402 has recently introduced the zap-tag in NIP-57, allowing clients to reference a static zap-destination in events. This can lead to events having stale zap-destinations in them, when users decide to switch to a new address or provider.

This PR introduces a third type for the zap-tag `pubkey`, with a flow quite similar to the base spec, but allowing clients to specify someone elses pubkey, essentially allowing "redirecting" zaps.

Both PRs lay the groundwork for client-side payment splitting, which will be included in a subsequent PR.

### DIFF
--- a/57.md
+++ b/57.md
@@ -168,7 +168,15 @@ A client can retrieve `zap receipts` on events and pubkeys using a NIP-01 filter
 
 ### Appendix G: `zap` tag on zapped event
 
-When an event includes a `zap` tag, clients SHOULD calculate the lnurl pay request based on it's value instead of the profile's field. An optional third argument on the tag specifies the type of value, either `lud06` or `lud16`.
+When an event includes a `zap` tag, clients SHOULD calculate the lnurl pay request endpoint based on it's value. A third argument on the tag specifies the type of value, either `pubkey`, `lud06` or `lud16`.
+
+- `lud06` : Calculate lnurl pay request endpoint according to [lud06](https://github.com/lnurl/luds/blob/luds/06.md)
+- `lud16` : Calculate lnurl pay request endpoint according to [lud16](https://github.com/lnurl/luds/blob/luds/16.md)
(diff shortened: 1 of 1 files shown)


fiatjaf:
Can't we do just pubkeys and not addresses directly? Having to support all these methods is super cumbersome for clients. @jb55 @v0l @pablof7z

pablof7z:
the idea behind using addresses is that it allows to ad-hoc create prisms that receive the sats

for zapstr, I want artists to be able to say "this is the split for this track"; so that track's event gets a tag

`["zap", "arist1-artist2-artist3@prism-splitting-service" ]`

that way nostr clients don't need to concern themselves with how splits are done, nor do multiple payments

Open to suggestions if consensus is that this is too complex for clients.

Egge7:
> Can't we do just pubkeys and not addresses directly? Having to support all these methods is super cumbersome for clients. @jb55 @v0l @pablof7z 

I agree and this is what I proposed in the PR that introduced zap-tags originally. However @pablof7z pushed for lud16 to be there, as it allows for notes to have "non-user-centric" addresses that point to prisms.

I believe going with a pubkey and handling splits on the client side is the better idea, but it will be more difficult for clients without a wallet or wallet Connect integration.

fiatjaf:
Well, if you all think it's best to do it like this I'm fine with it. It is unnecessarily bloated, but what can I do about it?

jb55:
pubkeys make way more sense to me since clients can show "zaps forwarded to bob" instead of "zaps forwarded to rando-ln-address@lightning.com".

ericholguin:
Should the third type just be `p` instead of `pubkey`, isnt that how its normally represented?

jb55:
> Can't we do just pubkeys and not addresses directly? Having to support all these methods is super cumbersome for clients. @jb55 @v0l @pablof7z

fwiw damus will likely just implement pubkey.

vitorpamplona:
Has been added via another PR

PR #428, opened 2023-04-10
Nip 121 basic joins

jason19710815:
Hi All... This is a proposal for a simple enhancement that I believe would improve the UX of Nostr apps. It could do this by improving screen load times, reducing the jumpiness of screen elements during the initial load, and preventing the problem of missing content (displaying pubkeys instead of user names and robot heads instead of PFPs).  

While this NIP's implementation would be pretty straightforward for relays using relational or nosql databases, Cameri has raised a question as to how relays using non-relational data stores might implement the NIP and has shared that one of the principles behind the Nostr protocol design is that it remains agnostic in terms of technology used for implementations. This makes perfect sense of course. 

I figured I'd go ahead and submit the NIP for discussion so the community can weigh these pros and cons to see if a path forward can be found. Thanks in advance for considering it and for the opportunity to maybe make a small contribution to the protocol.

### DIFF
--- a/121.md
+++ b/121.md
@@ -1,0 +1,75 @@
+# NIP-121
+
+## Basic Joins
+`draft` `optional` `author:j19710815`
+
+### Introduction
+
+This NIP introduces the `join` filter property as a pragmatic solution for common querying patterns performed by clients. The goal of this proposal is to improve UX for users of Nostr clients and the DX for client developers. This NIP enables clients to easily retrieve related, cross referenced notes. It does not seek to solve all querying challenges faced by clients interacting with the Nostr protocol. Rather it seeks to solve the most common use cases with the least amount of surface area in terms of protocol enhancement.
+
+### Motivation
+
+There is presently no way to perform a join among events when retrieving related data as one might do when querying a relational or nosql database. For instance, nearly every client displays a list of users with their profile image and display name. Currently, in order for a client to achieve this (without going out of band), the client must first retrieve a list of pubkeys (e.g., a user’s Kind 3 follows list), then open another filter for Kind 0 profile data that includes that list of users as authors, then filter the response to disregard any irrelevant (stale) data, match it to its use in the interface, then pass it along to a display function. Alternatively, the client could query relays for Kind 0 profile data for each user individually. 
+
+Should the initial list of users change, the process is run again on the client with the updated filters passed along to relays, or otherwise, complex caching is required to optimize for efficiency. The inability to retrieve related data with a single filter leads to an unrefined experience for users and an unnecessarily complex code base for clients. While this can all be optimized for by the client, it requires an inordinate amount of development time and computation put toward a technical concern that has long been solved by query languages.
+
+### Solution
+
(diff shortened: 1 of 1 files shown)


staab:
NACK, joins/relations don't really fit the event-sourcing paradigm. Joins add a lot of complexity by implicitly defining certain kinds of events and new data types. Arguably other features already do this, but still.

Egge7:
NACK. This sounds like offloading tasks to relays that are easily solvable on the client side. Introducing this NIP will also not necessarily reduce complexity on the clients code base, as they would still need to offer fallbacks, to remain compatible with all of the network.

jason19710815:
> NACK. This sounds like offloading tasks to relays that are easily solvable on the client side. Introducing this NIP will also not necessarily reduce complexity on the clients code base, as they would still need to offer fallbacks, to remain compatible with all of the network.

I definitely see the validity in your second point in that the only way to avoid clients needing to offer fallbacks would be to require the NIP on all relays, which is highly unlikely to happen at this stage. 

I would say though, that if the problem were easily solvable by clients, we wouldn't still be seeing pubkeys and robohash pfps popping up on well developed clients like Damus. Not to mention high power devices running hot because they're working so hard. So I think the problem remains unsolved. 

The question is then, what is the solution? If this work remains on the client, then we risk turning off users who expect modern UX and/or alienating those with less powerful devices. Not everybody in the world can afford powerful phones with unlimited data plans and I don't think they should be excluded from Nostr. If we move the work to bespoke backends (the most likely solution for VC funded super apps) that aggregate data and provide querying services, then we risk client and relay centralization. These backends won't emerge out of bad intentions, they'll simply be necessary to provide the best experience for users. For instance, identity on the web was captured because early players found ways to solve a problem unaddressed by the protocol itself. And people and devs gravitated to these providers because they simplified onboarding UX.

This NIP is pragmatic because, like Nostr, it colors outside of the lines a bit to keep the playing field level for all participants. It isn't perfect, but maybe we should keep it in our back pockets. We might see a need for it someday.

jason19710815:
>  NACK, joins/relations don't really fit the event-sourcing paradigm. Joins add a lot of complexity by implicitly defining certain kinds of events and new data types. Arguably other features already do this, but still.

I agree that this would be a whacky idea in any conventional architectural discussion. But so would connecting to a dozen ws services to request a dozen copies of hundreds of the same events. That's what we do though because it serves the purpose of not being reliant on any single source. Nostr also supports mutable events, which is a no-go in most event-driven systems. Point being, Nostr itself doesn't follow a strict event-sourcing paradigm. It is its own paradigm. That being the case, does it make sense to apply conventional constraints to an unconventional protocol? 

You also said that "joins add a lot of complexity by implicitly defining certain kinds of events and new data types..." I take this to mean the shape of the response would change based on if a join was performed. Is that the concern?

Regarding kinds… No new kinds would be introduced by this NIP, implicitly or otherwise.

staab:
Sorry, should have commented on closing. If someone wants to ACK this we can keep talking it over, but my guess is that this won't get much traction from the rest of the community.

jason19710815:
Thanks for the explanation. I see what you mean. Those links could be made explicit, but there's no need to discuss refining a proposal with such little interest. I appreciate your time. :-)

staab:
Reviewing this one more time just to make sure I understood it, it also seems to introduce a new return type? So in other words the return type would use the EVENT verb, but return something other than an event? IOW, the return type can't be signed by the author, because the tags are not what the author signed when publishing.

PR #155, opened 2023-01-07
NIP-05: Fixed some typos

VictorieeMan:
line 9: Added a hyphen
line 53: Fixed the typo by a context appropriate reformulation.
line 67: Removed an "are"
line 79: Added a comma

Great project, looking forward to more nostr :)

PR #443, opened 2023-04-14
Markdown improvements, embelishments, and standardizations

mariano-perez-rodriguez:
# Fixes

- Added missing `content-warning` tag to README

# Markdown Changes

- Use `-` for all unordered lists
- Unify heading styles to use `#` only at the start of lines
- Ensure a single sentence per line to minimize future diffs
- Ensure all fenced code blocks have a (correct-ish) language tag associated to them
- Use bolded unordered lists to implement pseudo-definition lists
- Remove **all** non-ASCII characters
- Remove bare URLs, use `<...>` instead
- Remove all HTML

# Style

- Move tags to top (immediately after NIP title)
- Be conservative with the use of ellipsis (especially in JSON fragments)
- Don't use `code style` to realize _emphasis_ (almost always)
- Use **THIS STYLE** for all occurrences of the keywords on [RFC 2119](https://www.rfc-editor.org/rfc/rfc2119)

# Standardization

- Use `NIP-XX` for **all** references to NIPs (and turn them into links)
- Use relative links for **all** references to NIPs
- Use `kind:NN` for **all** references to kinds
- Use `"T"` for **all** references to tags (both single-letter and multi-letter ones)
- Use `XXX` for **all** references to message names
- Use `field` for NOSTR events' field names (viz. instead of `event.field` or `.field`)

# Typos / Other Changes

- Capitalize proper names (eg. "Schnorr", "Merkle")
- Add links to OpenTimestamps
- Add links to CORS information
- Try our best to have JSON fragments be as valid as possible (eg. pay attention to trailing commas and the like)
- Remove **all** trailing spaces

# Code Changes

- add `;` to JavaScript sources
- add explicit `{...}` to C sources
- enhance and correct typos in Python sources

# Includes

- https://github.com/nostr-protocol/nips/pull/155

:)


fiatjaf:
Did you do all this by hand?

There are some good changes here, but some that I don't want. What do I do now?

mariano-perez-rodriguez:
Yup, all by hand, with a little help from vscode.

Ask and ye shall receive :)
What changes would have me remove?

fiatjaf:
I don't know, there are so many changes here, it's hard to read.

mariano-perez-rodriguez:
No problem, I'll create separate PRs for each file and close this one, does that sound OK?

mariano-perez-rodriguez:
I've created PRs for the README file and the first five NIPs, let me know if this is an approach you're OK with and I'll do my best to create the remaining PRs during the weekend / early next week :)

staab:
These improvements are mostly good and worthwhile, but we should aspire to spend zero time on bike shedding, and introducing and enforcing a style guide will siphon time away from more substantial development. Maybe a better approach would be to find a markdown formatter (that supports embedded json?) and add it to a CI step via Github actions.

In any case though, any global formatting convention change will result in a big commit, so I don't mind working through this diff and merging it, as long as it's consistent with an automatic formatter/checker. If you can introduce a mechanism to do that, we can start merging these changes.

WRT the specific changes you've made here, I think we should be conservative in what conventions to enforce (they should be either obviously correct or automatically enforceable, and not come at the expense of the readability of the raw markdown itself). So, we should skip the following changes:

- Use bolded unordered lists to implement pseudo-definition lists (I don't mind bolding, but bold + code fencing is weird).
- Use **THIS STYLE** for all occurrences of the keywords on [RFC 2119](https://www.rfc-editor.org/rfc/rfc2119) (capitals are enough, bold is just annoying)
- Code Changes to js/C/python unless obviously correct (we shouldn't have an opinion on semicolons)
- Unify heading styles to use # only at the start of lines
- Ensure a single sentence per line to minimize future diffs

mariano-perez-rodriguez:
I agree with the github action approach, seems like the way to go.

With respect to the comments made to the specific conventions: I applied the conventions I was most familiar with, but I'm happy to adapt them to the project's ones :)

Having said that, I'd like to comment on two specific ones:
- Unify heading styles to use `#` only at the start of lines: I don't mind using any header style whatsoever, what I do feel should be avoided is mixing the `=====` and `------` one with the `#` one... we should just pick one and stick to it; in that light, `#` is more flexible, that's why I think we should use that one in particular.
- Ensure a single sentence per line to minimize future diffs: this is one that has 0 impact on the actual rendering of the file, but it can indeed make diffs much shorter (think about fixing a typo on a big paragraph containing many sentences: if the paragraph appears as a single line, the whole line will be highlighted as changed, if we had a single sentence per line, only the sentence containing the typo will be highlighted); additionally, having more granular diffs reduces the conflict rate.

That being said, I'm not trying to impose anything on anyone, whatever the project decides works best, I'll be happy to help realize.

I'll close this PR and the related ones I made, and bring up a new one implementing the github actions approach, we can then decide on the specific linting rules that should apply therein.

:)

PR #449, opened 2023-04-14
Markdown improvements - 05

mariano-perez-rodriguez:
- Unify heading styles to use `#` only at the start of lines
- Ensure a single sentence per line to minimize future diffs
- Move tags to top (immediately after NIP title)
- Use **THIS STYLE** for all occurrences of the keywords on [RFC 2119](https://www.rfc-editor.org/rfc/rfc2119)
- Use `kind:NN` for all references to kinds
- Use `"T"` for all references to tags (both single-letter and multi-letter ones)
- Use `field` for NOSTR events' field names (viz. instead of `event.field` or `.field`)
- Try our best to have JSON fragments be as valid as possible (eg. pay attention to trailing commas and the like)
- Includes https://github.com/nostr-protocol/nips/pull/155

### DIFF
--- a/05.md
+++ b/05.md
@@ -1,16 +1,16 @@
-NIP-05
-======
-
-Mapping Nostr keys to DNS-based internet identifiers
-----------------------------------------------------
+# NIP-05
 
 `final` `optional` `author:fiatjaf` `author:mikedilger`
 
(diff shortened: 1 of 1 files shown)


mariano-perez-rodriguez:
Closing as per https://github.com/nostr-protocol/nips/pull/443#issuecomment-1511502530

PR #396, opened 2023-03-25
Update all NIPs to include 'depends' and 'mentions' tags

blakejakopovic:
PR for open Nostr Bounty: NIP dependency tree https://nostrbounties.com/b/naddr1qq9rzd3h8yerzwfexqmqygpm7rrrljungc6q0tuh5hj7ue863q73qlheu4vywtzwhx42a7j9n5psgqqqw4rsdu2nfu

Reviewed all NIPs and added NIPs to include `depends` and `mentions` tags. `mentions` tags were used where a NIP was mentioned, however if that reference NIP didn't exist, you would be fine. I fixed a few NIP headers, as they had an extra new line, or an inconsistent heading.

For the dependency graph, the following bash script can be used to parse files and generate a `dependency_graph.dot` file, which can be compiled into an output `dependency_graph.svg`.

Example SVG: [https://svgshare.com/i/rSB.svg](https://svgshare.com/i/rSB.svg). It may be a little noisey, as most NIPs point back to NIP-01. 

@fiatjaf I'm not sure how to create the Github Action for this repo, or exactly how it can generate an image and update the Readme.md / repo with the latest graph.

```bash
#!/bin/bash

# Create graph dot file
echo "digraph {" > dependency_graph.dot

# Loop over files in current directory starting with two or more numbers and ending in .md
for file in [0-9][0-9]*.md; do
  # Check if file exists
  if [ -e "$file" ]; then

    current_nip=$(echo $file | grep -o "^[0-9]\+")

    # Add NIP label for node
    title=$(echo "$(head -n 1 "$file"): $(sed '4q;d' "$file" | sed 's/"/\\"/g' | sed 's/\`//g')")
    echo -e "\t${current_nip} [label=\"${title}\"]" >> dependency_graph.dot

    # Extract NIP dependencies
    depends=$(grep -o "depends:[0-9]\+" "$file" | grep -o "[0-9]\+")
    if [ -n "$depends" ]; then
      # Add graph dependency edge
      for dep_nip in $depends; do
        echo -e "\t${current_nip}->${dep_nip}" >> dependency_graph.dot
      done
    fi
  fi
done

# Close graph file
echo -n "}" >> dependency_graph.dot

# Generate dependency graph
dot -Tsvg dependency_graph.dot > dependency_graph.svg
```

PR #454, opened 2023-04-17
Implement Markdownlint GitHub action

mariano-perez-rodriguez:
This PR implements a simple [MarkdownLint](https://github.com/DavidAnson/markdownlint) action, and applies the minimal changes required for it to pass.


mariano-perez-rodriguez:
I think saying something like "becoming a slave" is somewhat hyperbolic...

I completely understand if this is not the direction you want the repository to go, and I respect that, but allow me to present a counterpoint.

Linting is not (in my eyes) something that should "shackle" you, much to the contrary, it takes care of "irrelevant" stuff so that you can concentrate on what maters: the content.
If the rules as defined herein feel too "restrictive", we can certainly start a conversation around them and find the sweetspot that works best for all of us.

Furthermore, linting whatever we can will simplify ramp-up for would-be contributors, giving users utilizing this repository as reference a consistent look and feel, freeing cognitive resources for them to concentrate on what matters: the content.

Even more so, after the great point raised by @staab in https://github.com/nostr-protocol/nips/pull/443#issuecomment-1509850358, we can further extend the linter to add out own "semantic" validation rules.
If we were to have merged https://github.com/nostr-protocol/nips/pull/443 (or this PR for that matter), here's what we could now provide our users that we cannot now:

- linking across NIPs for ease of navigation, maintaining their focus and easing navigation, providing a more concise and unified experience that fosters learning
- all code blocks would have been semantically highlighted, again, easing the learning experience helping new users grok the examples faster
- thanks to the standardizations provided therein, looking for reference NIPs would have been trivial (a mere `grep` away), instead of requiring a manual tagging effort (sinergizing with https://github.com/nostr-protocol/nips/pull/396)
- again, thanks to the standardizations provided therein, all kinds would have been easily `grep`able, as would have been all tags, and message types
- the beauty of this all is that the linter chosen can be extended, and all of these benefits could have been indeed automated and cared for, with minimal vigilance upon PR review (we would have simply needed the NIP author mention the required NIPs)
- the linter providing a `fix` command, most of these decisions need not be final: we could have changed them down the road and have the linter re-adapt our codebase to our new preferences (hell, we could have even had local preferences for each user and used git hooks to convert back and forth if we felt sufficiently strong about it)

This is my attempt to forward rational arguments for as to why there's great benefit to be reaped from merging this PR (or any other PR on the same line, I'm not jealous like that).

Having said that, if you don't find value in these arguments and would rather keep things the way they are, perhaps there's a greater philosophical stance that we, would-be-contributors, should observe; if this is the case, I think explicitly stating it in the README file (perhaps in the "Criteria for acceptance of NIPs" section, perhaps in a CONTRIBUTING document) would save us all a lot of time and frustration.

:)

cardugarte:
I agree with @mariano-perez-rodriguez, for many of us it is not easy to start with these tools and if there is a much clearer standardization we can save a lot of time and effort.

chantzlarge:
I agree with @fiatjaf too much governance and you end up with [ActivityPub](https://www.w3.org/TR/activitypub/).

agustinkassis:
Still waiting for this.

PR #455, opened 2023-04-17
More precise wording on delegation conditions in NIP-26

sistemd:
NIP-26 says that conditions can be combined using `&`, but does not explain what it means to combine multiple conditions. For example, the `&` does not necessarily mean _and_. For `kind=` conditions, it actually means _or_. For multiples of `created_at>` and `created_at<` conditions, it's unclear if `&` should be allowed. I think it's best to say that such conditions should not be combined.


vitorpamplona:
Closing because NIP-26 is deprecated

PR #457, opened 2023-04-17
NIP-68 & NIP-69 - defined vocabulary for content warning & reporting (NIP-36 and NIP-56)

rabble:
This NIP defines a vocabulary and spec for content that users tag with NIP-36 (Content Warning) and NIP-56 (Content Reporting). 

To understand why we need this and how it'll be used: https://s3x.social/nostr-content-moderation

This spec was primarily created by: [S3X-Jay](https://github.com/S3X-Jay) [npub1veeqt66jt2j209y9nv4a90w3dej074lavqygq5q0c57dharkgw0qk738wt](https://snort.social/p/npub1veeqt66jt2j209y9nv4a90w3dej074lavqygq5q0c57dharkgw0qk738wt) with some input by myself and others. 

As stated, this is about a set of public standards for tagging and categorizing content that some users and relays may want to publish and see, or not host or see. Like with all NIP's the implementation is optional and up to the relay and clients.

PR #459, opened 2023-04-18
NIP-43 - Labeling

Gruruya:
Introduces abritrary labeling as thoroughly discussed here https://github.com/nostr-protocol/nips/discussions/438. It's written in the style of [1984](https://github.com/nostr-protocol/nips/blob/master/56.md), but general and with the option of voting against a label.

### DIFF
--- a/31.md
+++ b/31.md
@@ -10,7 +10,7 @@ A label is an event of kind `111` or `31111`, it describes the content of an eve
 It is similar to a `t` tag, but it can be added by others.  
 
 Labels allow for a user-defined level of moderation of arbitrary precision, among other things.  
-A label such as "nsfw" may be filtered to keep nsfw content away and a label such as "nostr" could be used to search for a specific topic.
+A label such as "nudity" may be filtered to keep nudity content away and a label such as "nostr" could be used to search for a specific topic.
 
 The content of a label event MAY contain additional information submitted by the entity labelling the content.
 
@@ -23,7 +23,7 @@ The `31111` label event has the same structure, but it should only reference a s
 A few standardized labels and their meanings follow:
 - `sfw` - Safe, the content is completely non-sexual
 - `questionable` - Potentially sexual content, soft-core porn
-- `nsfw` - Exposed genitalia or woman's chest
+- `nudity` - Exposed genitalia or woman's chest
 
 Example events
 --------------
@@ -32,7 +32,7 @@ Example events
 {
   "kind": 111,
   "tags": [
-    ["e", <event-id-1>, "nsfw"],
+    ["e", <event-id-1>, "nudity"],
     ["e", <event-id-1>, "spam"],
     ["e", <event-id-2>, "my favorite"]
   ],
@@ -61,4 +61,4 @@ Clients should decide how to weigh labels against each other, one option is to w
 Relay behavior
 --------------
 
-Consider nsfw labels the same as nudity reports.
+Consider nudity labels the same as nudity reports.
--- a/31.md
+++ b/31.md
@@ -1,0 +1,64 @@
+NIP-31
+======
+
+Labels
+---------
+
+`draft` `optional` `author:gruruya`
+
+A label is an event of kind `111` or `31111`, it describes the content of an event, according to an entity.  
+It is similar to a `t` tag, but it can be added by others.  
+
+Labels allow for a user-defined level of moderation of arbitrary precision, among other things.  
+A label such as "nsfw" may be filtered to keep nsfw content away and a label such as "nostr" could be used to search for a specific topic.
+
+The content of a label event MAY contain additional information submitted by the entity labelling the content.
+
+Tags
+----
+
+The `111` label event must include at least one `e` or `p` tag referencing an event id or pubkey to label followed by a label.  
+The `31111` label event has the same structure, but it should only reference a single event/pubkey and have that hash as its `d` parameter.
+
+A few standardized labels and their meanings follow:
+- `sfw` - Safe, the content is completely non-sexual
+- `questionable` - Potentially sexual content, soft-core porn
+- `nsfw` - Exposed genitalia or woman's chest
+
+Example events
+--------------
+
+```json
+{
+  "kind": 111,
+  "tags": [
+    ["e", <event-id-1>, "nsfw"],
+    ["e", <event-id-1>, "spam"],
+    ["e", <event-id-2>, "my favorite"]
+  ],
+  "content": "",
+  ...
+}
+
+{
+  "kind": 31111,
+  "tags": [
+    ["d", <pubkey>],
+    ["p", <pubkey>, "sfw"]
+  ],
+  "content": "It's a cute dog!",
+  ...
+}
+```
+
+Client behavior
+---------------
+
+Clients may use labels to filter events.  
+For nudity, labels, hashtags, and nudity reports as described in NIP-56 should be considered together.  
+Clients should decide how to weigh labels against each other, one option is to weigh labels based on if the labeller is someone you're following or indirectly following.  
+
+Relay behavior
+--------------
+
+Consider nsfw labels the same as nudity reports.
--- a/31.md
+++ b/31.md
@@ -4,7 +4,7 @@ NIP-31
 Labels
 ---------
 
-`draft` `optional` `author:gruruya`
+`draft` `optional` `author:michaelhall923` `author:gruruya`
 
 A label is an event of kind `111`, it describes the content of an event, according to an entity.  
 It is similar to a `t` tag, but it can be added by others.  
--- a/31.md
+++ b/31.md
@@ -6,19 +6,18 @@ Labels
 
 `draft` `optional` `author:gruruya`
 
-A label is an event of kind `111` or `31111`, it describes the content of an event, according to an entity.  
+A label is an event of kind `111`, it describes the content of an event, according to an entity.  
 It is similar to a `t` tag, but it can be added by others.  
 
-Labels allow for a user-defined level of moderation of arbitrary precision, among other things.  
-A label such as "nudity" may be filtered to keep nudity content away and a label such as "nostr" could be used to search for a specific topic.
+Labels can be used the same way as hashtags. They can be useful for filtering or searching for something, as labels to draw attention to something, etc.
+Labels can be assigned to events and profiles.
 
 The content of a label event MAY contain additional information submitted by the entity labelling the content.
 
 Tags
 ----
 
-The `111` label event must include at least one `e` or `p` tag referencing an event id or pubkey to label followed by a label.  
-The `31111` label event has the same structure, but it should only reference a single event/pubkey and have that hash as its `d` parameter.
+The `111` label event may include one or more `e` tags followed by a label and must include a single `p` tag, optionally followed by a label.  
 
 A few standardized labels and their meanings follow:
 - `sfw` - Safe, the content is completely non-sexual
@@ -34,16 +33,16 @@ Example events
   "tags": [
     ["e", <event-id-1>, "nudity"],
     ["e", <event-id-1>, "spam"],
-    ["e", <event-id-2>, "my favorite"]
+    ["e", <event-id-2>, "my favorite"],
+    ["p", <pubkey>]
   ],
   "content": "",
   ...
 }
 
 {
-  "kind": 31111,
+  "kind": 111,
   "tags": [
-    ["d", <pubkey>],
     ["p", <pubkey>, "sfw"]
   ],
   "content": "It's a cute dog!",
@@ -55,8 +54,9 @@ Client behavior
 ---------------
 
 Clients may use labels to filter events.  
-For nudity, labels, hashtags, and nudity reports as described in NIP-56 should be considered together.  
-Clients should decide how to weigh labels against each other, one option is to weigh labels based on if the labeller is someone you're following or indirectly following.  
+Clients may display profiles' and events' labels to users.  
+When doing this, clients should decide what labels to show, one option is to weigh labels based on if the labeller is someone you're following or indirectly following. Clients may also want to show who applied what labels or the number of labels applied.  
+Nudity labels, hashtags, and nudity reports as described in NIP-56 should be considered together.  
 
 Relay behavior
 --------------
--- a/31.md
+++ b/31.md
@@ -17,7 +17,10 @@ The content MAY contain additional information submitted by the entity labelling
 Tags
 ----
 
-The `111` label event may include one or more `e` tags followed by a label and must include a single `p` tag, optionally followed by a label.  
+The `111` label event consists of the following tags:
+- exactly one `p` tag, optionally followed by a label
+- optionally one or more `e` tags followed by a label
+If a label is specified, there may be another element following it containing either a `"+"` or `"-"` to specify if the label is a vote for or a vote against. Assume positive if omitted.
 
 Example events
 --------------
@@ -26,7 +29,7 @@ Example events
 {
   "kind": 111,
   "tags": [
-    ["e", <event-id-1>, "orange"],
+    ["e", <event-id-1>, "orange", "-"],
     ["e", <event-id-1>, "france"],
     ["e", <event-id-2>, "my favorite"],
     ["p", <pubkey>]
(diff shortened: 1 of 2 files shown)


Gruruya:
> users can label an account as being an organization, a "benign bot" (i.e. nostr-commits bot), a nsfw account, a celebrity, etc.

Should I add these examples to the NIP?

Also, should the label field be an array of labels? It would compress the size of the message in the event of mass-tagging.

Gruruya:
> I would think we would want to just modify kind 1984 right? Since a report is simply a "type" of label. I don't see the need for two separate kinds? In fact, the reason I proposed this was to consolidate all of these kinds of things under 1 kind.

Okay, it's just one kind now. Take a look. Reporting hasn't been changed, it was simply made a part of a more generic kind.

michaelhall923:
> may want to treat moderation requests specially

My thinking with this is "judgements" shouldn't be made at the protocol level. The data should be as objective as possible and up to the clients/relays to interpret it.

Gruruya:
> My thinking with this is "judgements" shouldn't be made at the protocol level. The data should be as objective as possible and up to the clients/relays to interpret it.

The difference between a label "spam" for a post discussing spam and "spam" in a report is very different. It should be expressed somehow. Though, the difference could be made in a way that doesn't involve the protocol. Something like `report:spam` as the label.
It could also be solved by giving context on what you're labeling. Something like `content:spam` vs `describes:spam` vs `image:spam`.

56.md
```diff
@@ -48,35 +54,72 @@ Example events
...
-Clients can use reports from friends to make moderation decisions if they
-choose to. For instance, if 3+ of your friends report a profile as explicit,
-clients can have an option to automatically blur photos from said account.
-
+Labels can be used the same way as hashtags, they can be used for filtering,
+searching, or displayed to draw attention to something.  
+Clients may want to consider or display the ratio of positive/negative labels.
+You may also want to weigh a label's importance based on how far the labeler is
+from someone you're following.  
+Labels can be used for sensitive content, for instance, if 3+ of your friends report
```
ekzyis, line 93:
```suggestion
Labels can be used for sensitive content, for instance, if 3+ of your friends label
```

56.md
```diff
@@ -2,35 +2,38 @@
...
+The `<label>` may optionally be followed by a `<classification>` such as:
+- `"content"` - The content is `label`
+- `"image"` - The image depicts `label`
+- `"topic"` - The event/profile is about `label`
 
-A `report type` string MUST be included as the 3rd entry to the `e` or `p` tag
-being reported, which consists of the following report types:
+Following the classification, there may be an `<affirmative marker>` which must be one of the following:
+- `"+"` - The `label` is accurate
+- `"-"` - The `label` is inaccurate
```
ekzyis, line 31: What is the reasoning for this "affirmative marker"? I imagine that accuracy of labels can be determined by how many users labeled something with the same label.

Gruruya: Many services use negative voting, I want to avoid multiple different ad-hoc solutions such as prefixing the label with a `!` or a `-`. It doesn't have to be used, I'll add a section that says a client may choose to ignore negative affirmative markers.

Gruruya:
> moved from @ekzyis review

I've included the `<affirmative marker>` field to keep clients from implementing their own ad-hoc solutions for negative voting.

If you want an example of negative/positive labeling in the wild, look at [civitai.com](https://civitai.com) which uses this mechanism to tag AI-generated images and differentiate explicit and safe content.

56.md
ekzyis: Ahh, I see. I thought by mistake that "accurate" means here that the user is sure this is the correct label. And "inaccurate" means that the user is not sure.

But "accurate" means "yes, this is the correct label" vs. "no, this is the wrong label" for "inaccurate"

Not a native English speaker haha

Gruruya: I wasn't sure how to label this either, so that's helpful feedback.

What about "a vote for" or "believes to be accurate"

ekzyis: The new description is great!

michaelhall923:
> Should any standard labels be included in this NIP?
My vote is no. Language is too subjective to be defined in a technical spec.

56.md
```diff
@@ -2,35 +2,37 @@
...
 
-If reporting a note, an `e` tag MUST also be included referencing the note id.
+The `<label>` may optionally be followed by a `<classification>` such as:
+- `"content"` - The content is `label`
+- `"image"` - The image depicts `label`
+- `"topic"` - The event/profile is about `label`  
 
-A `report type` string MUST be included as the 3rd entry to the `e` or `p` tag
-being reported, which consists of the following report types:
+Following the classification, there may be an `<affirmative marker>` which must be one of the following:
```
michaelhall923, line 28: I don't understand what this is. Does this mean "The image is not a picture of Belgium"?
`[ "e", <event id 1>, "belgium", "image", "-"]`

56.md
```diff
@@ -48,35 +53,79 @@ Example events
...
 Client behavior
 ---------------
 
-Clients can use reports from friends to make moderation decisions if they
-choose to. For instance, if 3+ of your friends report a profile as explicit,
-clients can have an option to automatically blur photos from said account.
+Labels can be used the same way as hashtags, they can be used for filtering,
+searching, or displayed to draw attention to something.
 
+Clients should have a threshold for labels to be considered legitimate,
```
michaelhall923, line 90: I'd change "should" to "may"

56.md
```diff
@@ -48,35 +53,79 @@ Example events
...
+- `nudity` - depictions of nudity, porn, etc.
+- `profanity` - profanity, hateful speech, etc.
+- `illegal` - something which may be illegal in some jurisdiction
+- `spam` - spam
+- `impersonation` - someone pretending to be someone else
+
+Some report tags only make sense for profile reports, such as `impersonation`
+
+> Deprecated due to its built-in negativity, emphasizing moderation needs over
+  neutral content classification.
```
michaelhall923, line 131: This section can be removed since we are maintaining backwards compatibility.

56.md
Gruruya: Yes. It's a way for users to vote against a label. You're the second person asking so I need to clarify the wording more.

56.md
Gruruya: Good point

56.md
Gruruya: You're right. I wanted to indicate that this should be done somehow, something analagous to the previous "It is not recommended that relays perform automatic moderation using reports." But that's true of all events in Nostr and it may be alright to assume the reader already knows.

56.md
Gruruya: I'm changing the name, `affirmative marker` is too unknown. Which of these seems clearest to you:
- stance
- alignment
- polarity

56.md
```diff
@@ -2,35 +2,37 @@
...
 
 Tags
 ----
 
-The report event MUST include a `p` tag referencing the pubkey of the user you
-are reporting.
+The label event contains one or more tags:
+- `[e, <event-id>, <label>]` - Label event
+- `[p, <pubkey>]` - Use pubkey to find recommended relays
+- `[p, <pubkey>, <label>]` - Same as above, but also label pubkey
```
ekzyis, line 21: Labeling a pubkey essentially means labeling the relay which are found using this pubkey, right?

I would make this more clear then.

56.md
ekzyis: To be honest, I think `affirmative marker` is actually the best choice. It's the closest to the noun `confirmation`. Or use `confirmation marker` (?)

Maybe it would suffice to describe the examples in a few sentences.

For example:

```json
{
  "kind": 1984,
  "tags": [
    [ "p", <pubkey>, "nudity"] // pubkey to find recommended relays. relays contain events about nudity
    [ "e", <event id 1>, "belgium", "image", "-"], // event is labeled as not containing an image of belgium
    [ "e", <event id 1>, "france", "image"], // event is labeled as containing an image of france
    [ "e", <event id 2>, "my favorite"], // event is labeled as "my favorite"
    [ "p", <pubkey>] // pubkey to find (other) recommended relays
  ],
  "content": "",
  ...
}
```

I don't like how the JSON syntax is malformed though. The red highlighting is hard to look at.

56.md
Gruruya: No, it's the same as reporting the pubkey was so it's labeling the account/profile. I haven't detailed a way to label relays (yet?)

56.md
Gruruya: I'll split that one into a block followed by an explanation.

56.md
ekzyis: Ah, oh, I see. I didn't read how it was before.

56.md
michaelhall923: I prefer vote but am kinda indifferent. I guess I just struggle to see the use case for this tbh. Are we expecting that clients will want to implement an "unreport" feature or something?

56.md
Gruruya: Is it clear the way I ordered it?
https://github.com/nostr-protocol/nips/blob/8e657d0e38dda371217188502e27c0e91f48882a/56.md?plain=1#L21

56.md
michaelhall923: Yeah I agree that it should be done in most cases, but something doesn't feel right about wording it that way in the spec

56.md
Gruruya: Yeah, an "unlabel," I don't see it being used for reports too much as well. It's can be used for other things like if an account is a celebrity or an image is of belgium vs france

56.md
michaelhall923: Labeling relays probably isn't a good idea because ip addresses and domain names change. There is no universal pointer to a relay like there is pubkeys for events/profiles. This seems like a better use of NIP-11 https://github.com/nostr-protocol/nips/blob/master/11.md

Gruruya: Really good point.

56.md
michaelhall923: Ok. I personally would probably never use it but I guess it doesn't hurt. It's good that you put the default state of "content" and "+" because that's how most people will probably use this.

Gruruya: Yeah and for backwards compatability with reports

michaelhall923 (approved):

staab:
I think this is a step in the right direction, here's a few changes I would propose based on the conversation over at https://github.com/nostr-protocol/nips/pull/457#issuecomment-1544035939:

- This might be considered backwards-incompatible, but it would be nice to be able to specify more dimensions to a label, including `quality`, `confidence`, and `context`. I would envision this as additional tags that apply to all `e` or `p` tags mentioned. This would mean multiple labels could no longer be applied on a single event, but I actually see that as an improvement.
- Moving the label from a mark to an `l` tag would make these labels queryable.
- Structured/namespaced labeling as described in the draft for NIP 68 would allow a more granular labeling approach for people that want it, either to refer to an existing standard, or to create a proprietary namespace of some kind. Nothing in this NIP prevents that.

Other suggestions:

- Instead of including a `p` tag, why not directly include one or more `r` tags, if the purpose is to facilitate routing? It would be useful to be able to label pubkeys as well using the same mechanism (as the old version had it).
- I don't see much value in `classification`, but `affirmative marker` is a good idea. Maybe that could be expanded via `quality` as described in the NIP 68 draft to a range of -1 to 1 (i.e., -1 quality says this is _not_ an example of the label).

So NACK on the current draft, but I think the general direction is a good one. As many of the above features as can be included in a backwards-compatible way (probably not `l` tags unfortunately) would get my vote.

Gruruya:
> I don't see much value in `classification`

How should "spam post vs discussing spam vs image of spam" be solved?

> Instead of including a p tag, why not directly include one or more r tags

Addressing via public-key is better as URLs can change or go down and notes can move.

> `l` tags
 
Just extend `t` tags, the only specification they currently have is a single sentence at the bottom of NIP-12.

Gruruya:
I like the idea of using tags instead of fields appended to e/p tags for better filtering, I'll change this NIP to do that. It will have to be moved away from the `1984` kind but that's a bit of a non-feature now that I think about it again.

staab:
> How should "spam post vs discussing spam vs image of spam" be solved?

I would expect "spam post" to be a label, while a discussion or image of spam would be a topic. The structured labeling suggestion from the 68 proposal allows attributing topics to other events.

> Addressing via public-key is better as URLs can change or go down and notes can move.

Yes, but you first need to find the person. Also, relay hints in nostr are always relay urls to my knowledge, not pubkeys.

> Just extend t tags, the only specification they currently have is a single sentence at the bottom of NIP-12.

Hmmm maybe. I'm all for conflating similar use cases normally, but topics are so much a part of UX I think a separate tag type for stuff that shouldn't always be surfaced to the end user would be a good idea.

> It will have to be moved away from the 1984 kind 

I'd rather not introduce another redundant kind, that still results in more complexity for implementers. `l` could be an ok addition to the kind though, I'm not sure.

Gruruya:
Yeah there's not really a deprecation path for Nostr is there?

staab:
Nope haha

Gruruya:
Is there a simple way to do an unnamespaced search with that system? That's something that will come up a lot. If there some other way to introduce AND queries that would be great.

michaelhall923:
> I'd rather not introduce another redundant kind, that still results in more complexity for implementers. l could be an ok addition to the kind though, I'm not sure.

Yes the purpose of this NIP was to decrease complexity not increase it lol

Gruruya:
We'll just include a formal syntax for namespaces and remove the classification field. We could even prefix labels with a `-` instead of having the `affirmative marker` field. In fact, if we remove all the extra fields we could end up in a place where multiple labels could be included in a single array.

Gruruya:
My problem with namespaces is that it's impossible to ignore. Allowing any namespace is great for evolution, but as we know people will never agree so it's going to be impossible to find anything when we're all using different namespaces. The purpose of having it as a seperate field was to have an escape hatch so that you can query everything if you want to. Also, it allows for filtering for multiple namespaces at the same time which can also help to alleviate the problem.

staab:
That's a compelling reason, I'd be fine with adding the namespace as an extra field

Gruruya:
> `quality`

I don't want to change the affirmative marker to a floating point number as then it couldn't be filtered and @michaelhall923 earlier said that he would want to completely ignore negative labels.  
I was thinking about adding a `~` option which would be equivilent to a 0 in the -1 to 1 range but decided it wasn't needed.

> `confidence`

It's no different from `quality`.

> `context`

I see `classification` as accomplishing the same as a `namespace` or a `context` field would, am I missing something?  
I could rename `classification` to `context`, that's what I originally called it after all.

> but topics are so much a part of UX I think a separate tag type for stuff that shouldn't always be surfaced to the end user would be a good idea.

`t` tags are fine, but they're severely limited - they can only be set by the person making the post and have no context. Part of the reason I want labels in my client is for other users to be able to vote down inaccurate hashtags.

staab:
> It's no different from quality.

Quality is objective, confidence is subjective. Quality is about how closely something fits a definition (I'm 80% libertarian), confidence is how sure you are about your assessment (you're 50% sure I'm 80% libertarian). This is an approach [New Founding](https://www.newfounding.com/guide) took (unfortunately their guide is no longer online).

> I see classification as accomplishing the same as a namespace or a context field would, am I missing something?

I think that's right, namespace and classification seem the same. Context would just be additional events meant to help a human moderator make a decision.

> t tags are fine, but they're severely limited - they can only be set by the person making the post and have no context. Part of the reason I want labels in my client is for other users to be able to vote down inaccurate hashtags.

If you look at the PR for NIP 68, they do this using qualified labels, e.g. `#t>butter` would be a way to assign the "butter" tag to some other event. In this version, I would think something along the lines of `["e", <eventid>, "butter", "t"]` would be equivalent.

Gruruya:
> Quality is objective, confidence is subjective.

Your objective opinion? Seems like an interesting way of going about it, dunno how I would make it work with the affirmative marker field. A catch-all stringified JSON field like the other NIP seems convenient but messy.

> Context would just be additional events meant to help a human moderator make a decision.

I see, I see. I think the content is what's supposed to do something similar in this version.

> In this version, I would think something along the lines of ["e", <eventid>, "butter", "t"] would be equivalent.

Yep, that would be acceptable. Could also use `"topic"` for the last field.

staab:
Ok, I just created a new PR similar to this one, but which pays a little more attention to backwards compatibility, and adds some of the stuff in from 68/69. https://github.com/nostr-protocol/nips/pull/524

Issue #461, opened 2023-04-20
NIP-15: shipping should specify regions, not countries


ibz:
Event 30017

```
...
    "shipping": [
        {
...
            "countries": [<String, countries included in this zone>],
        }
```

I am feeling that using the "countries" terminology here is very limiting.

Not only that there is not even a good definition of a "country" out there, but I might have different shipping costs to different regions around a country (US states for example, or a lower cost for the city I am in and I can meet face to face).

Also, I might want to have some "meta-regions" like "Europe", so I don't need to specify all included countries (although I agree this latter case could be emulated using presets in the UI).

What if we renamed "countries" to "regions" instead, which is a more flexible term?

@motorina0 agree?

motorina0:
Yes, I agree `regions` is better than `countries`. 
Please open a PR.

raiseprogrammer:
Hi @ibz @motorina0, For me it would make sense to say "location" or "locate". For example,

```
...
    "shipping": [
        {
...
            "locate": [<String, countries included in this zone>], // countries, regions etc
        }
```

or:


```
...
    "shipping": [
        {
...
            "location": [<String, countries included in this zone>], // countries, regions etc
        }
```


Location" or "locate" can be state, region, country, etc. What do you all think of this idea of ​​inserting a meta tag called: "locate" or "location"? it makes sense?

motorina0:
@ibz I think your PR was merged. Can this issue be closed?

PR #465, opened 2023-04-21
Rename countries to regions.

ibz:
#461


fiatjaf:
Looks good to me.

motorina0:
How can we coordinate with other users/interested parties of NIP15?
Chances that they will see this PR are slim.

Are there telegram channels per NIP now? :)

ekzyis:
I think `location` would be better since it's more generic than `region` and `country`. This was mentioned by @ghost here: https://github.com/nostr-protocol/nips/issues/461#issuecomment-1517908873

ibz:
I think `region` is more generic than `location`. You can call Paris a `region` (although it is a small region) but you can't call some bigger region, like France or the EU a `location`. `location` is ways very *local*, hence the name.

motorina0:
> @motorina0 We finally have a Telegram channel for NIP-15! Should I ask there as well?

For whoever might be concerned: 

Channel: https://t.me/+praQQFXCDiFlZGJk

PR #17, opened 2022-07-09
private dm events

Giszmo:
This nip is the result of https://github.com/nostr-protocol/nostr/issues/69 and was tried out in https://github.com/Giszmo/NostrPostr/blob/master/nostrpostrlib/src/main/java/nostr/postr/events/PrivateDmEvent.kt

PR #468, opened 2023-04-23
NIP-59 Gift Wrap

v0l:
A stupid simple solution to DM metadata leakage.

Is it terrible? Yes! 

Does it work? Also Yes!

https://github.com/v0l/nips/blob/59/59.md

### DIFF
--- a/59.md
+++ b/59.md
@@ -1,0 +1,55 @@
+NIP-59
+======
+
+Gift Wrap
+-------------------------------
+
+`draft` `optional` `author:kieran`
+
+This NIP defines a simple method to protect event metadata leakage by using randomly generated keys.
+
+A gift wrap event is a `kind 1059` event which wraps another event with a single use key.
+
+This event kind can be used to wrap DM's or other events which are considered private.
+
+This is similar in concept to Onion Routing.
+
+To construct a wrapped `kind 4` you first start by producing the `kind 4` event as normal, then create a random key and produce the `kind 1059` by encrypting the `kind 4` as a JSON string and storing it in `content` of the `kind 1059` using the same `p` tag to notify the recipient.
+
+Example wrapped event:
+```json
+{
+    "id": "3731dc639ada388aa5ee8c3404351868d4dd0fdcac00183e540669ddc25e8ad4",
+    "pubkey": "4d02c0b4e16db38b80fa3b46548417a4ccc975337e29bc22da72132cf69f3b4a",
+    "content": "7sW2yHvC5o/x6b0xCKPpH8KZJQeQYxt3Y53mxMV53yGQ5kn/jgMM74k+vfwegYBY9zjlQeTvyF5gYuJIHM4KN2SZq7zaCIKqTqMjYhLPI4VjIvowhmMuuZgRlB7cestrYh6OSSteICye6sxkS1NdpON31u+JNK/GHFVRJqfYF4Fnzaka2u9WHQKmm9/nN040OHZ41eT3YUWpaZHupVx1fuN6UoeyPjb2PW4yiwxD72LhlAoNGjR+otpWZ3kD6LU3Btrz8btjSD272wJXT/FBX5Gs59BMCDeJ4eReNfbbVBB5mlXvVqVC/tnHntVgNvAP5GsZE56/4t+kLqql0dqZU/haC8wpOxqtzqzhzxTHxS/IQ3HYpDnU8XuXUqNw1zOQhp0F65rXZUf9nN0gIa/zGa35fK2ek2p8K/3exKssO7fVxaMeTVxP7jEoCmn47gsy5mv4fH7MtxoWvWMapvGFLZf58NPy54ktvldOdGp9BJLO+6/2P7KJqb7bvu0EIPdYCcUt+muYwiJCUoHNFfFQwKrrpr23XuPQlPYNPUs6orbNBs/Oj8vMfZC5hb/AUaNanGtsqsBgqfEufyonBtihMzLZqn9OZNWHE7lYfnhSOZnUjStl03Bd94Qx1rp7IHZN?iv=cIiZuLb2FMsLzc+yYf+3IQ==",
+    "kind": 1059,
+    "created_at": 1682286690,
+    "tags": [
+        [
+            "p",
+            "63fe6318dc58583cfe16810f86dd09e18bfd76aabc24a0081ce2856f330504ed"
+        ]
+    ],
+    "sig": "48e50d7c2d301a494e0cf250685fe4a1d7d03bdb0efae38a0e9575d2dea04b8c227f8bc35fd0451f7e1d7828a8c5f7b2f81fdab135d7a3ef92c4364aaaa8514f"
+}
+```
+
+The `content` is NIP-04 encrypted JSON string with the temporary key, this is the inner event:
+```json
+{
+    "id": "08eb6f89e916b0bd747c039d33c11be0575c4c6d2c08363ada1243ea771a92d2",
+    "pubkey": "63fe6318dc58583cfe16810f86dd09e18bfd76aabc24a0081ce2856f330504ed",
+    "content": "73ZaB2WzPlIwwC55YnhsFA==?iv=1R2WlyrAk9VQ1NdhNuWmsw==",
+    "kind": 4,
+    "created_at": 1682286690,
+    "tags": [
+        [
+            "p",
+            "63fe6318dc58583cfe16810f86dd09e18bfd76aabc24a0081ce2856f330504ed"
+        ]
+    ],
+    "sig": "703bbc5e15208b106ae0885c6c3f20ee97c1f9e6e45e617d49ea6200f3437d74cd5cb186df3656a18fa9b23c362c39676229ab98a323b29112b33cbbd8a910ca"
+}
+```
+
+Clients can decrypt the gift wrapped event and can continue to work as normal.
--- a/59.md
+++ b/59.md
@@ -22,6 +22,14 @@ The key used to sign the wrapper and encrypt its contents can be created using a
 - For the strongest security guarantees, it SHOULD be randomly generated.
 - The sender MAY use `sender private * event id mod curve.n` instead of a completely random key. If this is done, the sender can "retain access" to the sent message and is capable of decrypting, deleting or otherwise operating on the outer message.
 
+A `p` tag for the recipient MUST be added to the `tags` field on the outer event. To protect recipient
+identity, this may be a regular pubkey, or one representing a group of people sharing a private key for
+decryption.
+
+The sender MAY copy tags from the inner event to the outer event. This weakens privacy guarantees,
+but may be useful to accommodate server-side filtering. A `k` tag MAY be added to the outer event to
+denote the `kind` of the inner event.
+
 Example wrapped event:
 ```json
 {
--- a/59.md
+++ b/59.md
@@ -12,43 +12,49 @@ A gift wrap event is a `kind 1059` event which wraps another event with a single
 
 This event kind can be used to wrap DM's or other events which are considered private.
 
-This is similar in concept to Onion Routing.
+This is similar in concept to [Onion Routing](https://en.wikipedia.org/wiki/Onion_routing).
 
-To construct a wrapped `kind 4` you first start by producing the `kind 4` event as normal, then create a random key and produce the `kind 1059` by encrypting the `kind 4` as a JSON string and storing it in `content` of the `kind 1059` using the same `p` tag to notify the recipient.
+To wrap an event, sign it as usual then serialize it to JSON. This string should then be encrypted using NIP-44 with a different private key. 
+
+Next, add the signed payload to the `content` field of a new `kind 1059` event, signed with the same key used to encrypt the content.
+
+The key used to sign the wrapper and encrypt its contents can be created using any appropriate strategy:
+- For the strongest security guarantees, it SHOULD be randomly generated.
+- The sender MAY use `sender private * event id mod curve.n` instead of a completely random key. If this is done, the sender can "retain access" to the sent message and is capable of decrypting, deleting or otherwise operating on the outer message.
 
 Example wrapped event:
 ```json
 {
-    "id": "3731dc639ada388aa5ee8c3404351868d4dd0fdcac00183e540669ddc25e8ad4",
-    "pubkey": "4d02c0b4e16db38b80fa3b46548417a4ccc975337e29bc22da72132cf69f3b4a",
-    "content": "7sW2yHvC5o/x6b0xCKPpH8KZJQeQYxt3Y53mxMV53yGQ5kn/jgMM74k+vfwegYBY9zjlQeTvyF5gYuJIHM4KN2SZq7zaCIKqTqMjYhLPI4VjIvowhmMuuZgRlB7cestrYh6OSSteICye6sxkS1NdpON31u+JNK/GHFVRJqfYF4Fnzaka2u9WHQKmm9/nN040OHZ41eT3YUWpaZHupVx1fuN6UoeyPjb2PW4yiwxD72LhlAoNGjR+otpWZ3kD6LU3Btrz8btjSD272wJXT/FBX5Gs59BMCDeJ4eReNfbbVBB5mlXvVqVC/tnHntVgNvAP5GsZE56/4t+kLqql0dqZU/haC8wpOxqtzqzhzxTHxS/IQ3HYpDnU8XuXUqNw1zOQhp0F65rXZUf9nN0gIa/zGa35fK2ek2p8K/3exKssO7fVxaMeTVxP7jEoCmn47gsy5mv4fH7MtxoWvWMapvGFLZf58NPy54ktvldOdGp9BJLO+6/2P7KJqb7bvu0EIPdYCcUt+muYwiJCUoHNFfFQwKrrpr23XuPQlPYNPUs6orbNBs/Oj8vMfZC5hb/AUaNanGtsqsBgqfEufyonBtihMzLZqn9OZNWHE7lYfnhSOZnUjStl03Bd94Qx1rp7IHZN?iv=cIiZuLb2FMsLzc+yYf+3IQ==",
+    "id": "0bb58c5e646ff1a310d888347cd392f0490d80d9ce963a43d10f0af0b3a92d58",
+    "pubkey": "67813274a3e44cc5f939cc29f1faa4e6b38f1ee4a22469262d6d9d19c377f027",
+    "content": "{\"ciphertext\":\"0s44ii9QNQMmKl/+Bmo0dUoVqCI3dqyPweoA1H06qkguf8u3Hr8sK7lUYKi/DHrR2PiLKuJRHC42mBn+e4jyM49yj+QL0cg3p5MbGiSmExEQPnofaBS6PtU/6Fa30XPNoywC9ptaPiDpEPNTxsI+LgCoCgUc3o9iJzKJRgzfaUVhjugjXmGw6Un8DgX7yfY3JBraxZClfC6FNV3yTtmRdi9AHmphJa5xW8l/XeI1k3C+Qv5z4mMI4GfiYXnqLbHtaMSeTaI31zmyr4nX9Bpnb14PSdiKpHEN1aAmJXgs/mXRw5KS44EDnjSJwItwow/yWNlGZZ2lT21ay9uc7LeD2rNOIOpW3xO5Pqd51NpYgjk+1lD9L8djQCpWRzsJoCr2/G1+qeGre+q/5g7re2aQbA3M9iSqEzBlkw841fq5AJYEDcsCXzFFQwo0ydJutBhzP2YOFuPbGgvWZat2bLRu26PZMMB3Z6ZprzyBrLIgZ4uOJ74/Rpflh4I5xl4r9Yvp37bLkxq25diG/78x0/AsBlgIJzMeNivZ7LCxm9a/nu5Ygj1H8g==\",\"nonce\":\"CvLX+kBOEBM5ZRm/I+4fncvGqbrcb6Dy\",\"v\":1}",
     "kind": 1059,
-    "created_at": 1682286690,
+    "created_at": 1686840217,
     "tags": [
         [
             "p",
             "63fe6318dc58583cfe16810f86dd09e18bfd76aabc24a0081ce2856f330504ed"
         ]
     ],
-    "sig": "48e50d7c2d301a494e0cf250685fe4a1d7d03bdb0efae38a0e9575d2dea04b8c227f8bc35fd0451f7e1d7828a8c5f7b2f81fdab135d7a3ef92c4364aaaa8514f"
+    "sig": "cdcc561865cc4f028311159a9417b717d93a681f957621052335bb3c581f85490cb3f596d251e92f7174d34eeb406ff009acde37588322fcb75c380e335a1b73"
 }
 ```
 
-The `content` is NIP-04 encrypted JSON string with the temporary key, this is the inner event:
+The `content` is NIP-44 encrypted JSON string with the temporary key, this is the inner event:
 ```json
 {
-    "id": "08eb6f89e916b0bd747c039d33c11be0575c4c6d2c08363ada1243ea771a92d2",
+    "id": "a3b3d167f560e9d15434534ee7bf51cf272b15e40e3e9a64501fb83fd0f06326",
     "pubkey": "63fe6318dc58583cfe16810f86dd09e18bfd76aabc24a0081ce2856f330504ed",
-    "content": "73ZaB2WzPlIwwC55YnhsFA==?iv=1R2WlyrAk9VQ1NdhNuWmsw==",
-    "kind": 4,
-    "created_at": 1682286690,
+    "content": "example",
+    "kind": 1,
+    "created_at": 1686840217,
     "tags": [
         [
             "p",
             "63fe6318dc58583cfe16810f86dd09e18bfd76aabc24a0081ce2856f330504ed"
         ]
     ],
-    "sig": "703bbc5e15208b106ae0885c6c3f20ee97c1f9e6e45e617d49ea6200f3437d74cd5cb186df3656a18fa9b23c362c39676229ab98a323b29112b33cbbd8a910ca"
+    "sig": "675a8cd35925928d6623ba18aea0eeb9691c462c59cac4a6441f9495eaec7182107ece7b7012bb628e28029f059362fcbce9bce3a00c05817480603d8f8112f5"
 }
 ```
 


Giszmo:
The use of ephemeral keys was proposed before in different ways and was always the weak point. Relays will not be too happy to accept ephemeral keys, especially when the content is encrypted. So the relay you are talking to would have to learn it's you which brings us back to what fiatjaf and others favor: the relay learns the meta data but promises the sender and receiver of DMs to not let unauthorized clients request these DMs.

The advantage of using ephemeral keys is that in theory, events could be mirrored to other relays without damage but that holds DOS and spam potential.

v0l:
> How is the temporary key communicated?

Its not, it's used only wrap the actual DM event, but as it turns out this has been suggested before (forgot to check closed PR's/Issues).

And for reasons which @Giszmo has suggested it makes sense that this would be bad for relays, although as it currently stands free relays already have to deal with random keys spamming them.

d-krause:
This is essentially what I have been working on with nip 76 (https://github.com/d-krause/nostr-nips/blob/nip76-draft-2/76.md)

It does not get any interest, but I am still pressing on with Draft 3 that addresses more concerns.   I guess you could say that Nip 76 Keys are "ephmeral", but they are also deterministic, using one determined key for the author and another determined key for the encryption.  We could let relay operators in on the author determined key, but still withhold the encryption determined key.

Something along these lines needs to happen and it will happen one way or another despite what relay operators like and dislike. It is high time we stop storing or data in the clear, both in terms of whether it readable AND whether it belongs to us or not.

If we can't get relay operators that control this repo on board,  I can see a Nostr fork coming where there will be operators that are on board.  I don't understand their resistance anyway - they are just "NIPS" not mandates.  They can approve a NIP without agreeing to conforming to it.

Giszmo:
> although as it currently stands free relays already have to deal with random keys spamming them.

Well, nip-42 is only now being rolled out. No client supported it a week ago.

Giszmo:
I said it many times and I will repeat it here, too. I don't like any approaches that require limiting event replication across relays. A new relay should be easy to spin up and saturated with past events and other relay operators should only charge for what it costs them to stream over all that data. There should not be information privy to one relay or another. That is why I am in favor of plausible deniability in the DM meta data using a nip like my #17 

With approaches that limit DMs to one relay to not leak too much meta data make parties to the conversation dependent on that relay.

starbackr-dev:
Hello Everyone, This is Arun from the Current App team. 

There are several open pull requests aimed at solving the privacy issues surrounding Direct Messages on Nostr. Our team has conducted research on this matter over the past few weeks. Many of these requests involve the exchange of keys between two users to ensure security or create secure channels. They are complex in terms of preventing leaks and ensuring privacy measures are in place.

We have also examined the use case from the perspective of end users. Almost all the proposed solutions address the scenario of "We are friends and need to communicate securely, so let's use Nostr DM." We believe this is not a valid use case for Nostr.

The use case for DMs on Nostr is quite different. It begins with a message like, "Hey, I saw your post about attending this meetup. Do you have time to meet up in person?" The content of the event is already encrypted. The only issue here is the leakage of metadata, which could be used by anyone to create a social graph or for spamming profiles that receive DMs from trending profiles.

This NIP-59 proposal, which wraps DMs as a gift into another event, works well for this use case by addressing the metadata leakage. After reviewing other solutions, we kept returning to this NIP because of its simplicity and effectiveness within Nostr.

Here are the different levels of DM privacy that I see:

Level 1 - Kind 4: Content is encrypted, but I have no concern about metadata leaks.
Level 2 - Kind 1059: I am comfortable trusting my paid relay to view my metadata, but not others.
Level 3 - Kind XX: I will use onion routing to obfuscate metadata and send it to routing bots.
Level 4 - I need NSA-level super-private secure, encrypted DMs to use with my friends. 

We should begin by implementing level 2, which requires changes to the client, and then move on to implementing level 3. Level 4 should never be implemented within Nostr and direct them to Signal or SimpleX. 

Of course, this does not solve everything, and a couple of issues were raised during this discussion. Let me address them.

1. **There is an argument that "throwaway" keys will not work with paid relays.** However, the majority of relays are still open and will automatically work with this NIP. The reason paid relays block throwaway keys is that they perform authentication during write rather than on connection. At least, this is what I observed with paid relays built on Nostream. I modified the code to authenticate on connection instead of write, and now it works with throwaway keys as well.  
Here’s a code snippet that connects to our DM relay using NIP-42 AUTH and then sends DM using throw away keys. [https://gist.github.com/starbackr-dev/558c30c126bcb2e10d39e1dc9c58d8a6](https://gist.github.com/starbackr-dev/558c30c126bcb2e10d39e1dc9c58d8a6)

As of today, There are 120 free relays Vs 65 paid relays. Using any of the free relays with NIP-59 will stop the metadata leakage and make it anonymous. 


2. **Relay operators can still link throwaway keys** to the original profile using IP addresses or by linking the authentication public key and the throwaway key. We can always enhance the NIP to prevent this by using routing relays/bots that act as intermediaries to pass the message, thereby addressing the concern. We can also create relays specifically designed for ingesting DMs and declare that they will not perform any data mining. In fact, we have already built one for Current App users.


This is also backward compatible and no user flow changes required. Clients query for both kind 4 and 1059 and unwraps kind 1059 into kind 4 automatically. When starting a new conversation they can choose a less(kind 4) or more secure (kind 1059) way to send DMs. Overtime we believe all Nostr clients will start to support kind 1059. 


Couple of weeks ago we released a version on Current App with DM and push notifications. @v0l  we are ready to start supporting and releasing a test version to see how the users react to it. 

@Semisol  - Your requested changes has been addressed. Can you review it one more time? 

One reason why I enjoy developing on Nostr is its simplicity. We should not compromise that for complex solutions that no one will use in the end.

It's a simple, straightforward solution. Simple always wins... not SimpleX.

paulmillr:
Why not 444?

starbackr-dev:
Yes. 444 is even better..!

paulmillr:
In the meantime i've introduced NIP-44 that fixes broken cryptography of NIP-4 and introduces versioning:

https://github.com/nostr-protocol/nips/pull/574

arthurfranca:
> Why not 444

Because of [NIP-16](https://github.com/nostr-protocol/nips/blob/master/16.md). I think kinds below 1000 are considered legacy, because it isn't clear how relays should handle them.

arthurfranca:
> @starbackr-dev: [...] the majority of relays are still open and will automatically work with this NIP. The reason paid relays block throwaway keys is that they perform authentication during write rather than on connection. At least, this is what I observed with paid relays built on Nostream. [...]

Afaik they don't use NIP-42. Paid relays are just checking the pubkey of received events, so the side-effect is that one can only send events of allowed pubkeys. That's why NIP-59 won't work for now.

This probably has to do with the fact that NIP-42 flow is unpredictable in the sense that it is hard for the client to know when and if the relay 'AUTH' message will arrive. So just a few clients enable this, which makes relay don't require it.

Please support [NIP-43 - Fast Authentication](https://github.com/nostr-protocol/nips/pull/571) PR for a better alternative.

earonesty:
this is exactly how we are doing encrypted group chat invitations in arclib.   we're also using a "fixed" nip4, which is basically the same as nip44.   i see no problem with this proposal and would love it to be a standard

Egge7:
> this is exactly how we are doing encrypted group chat invitations in arclib.   we're also using a "fixed" nip4, which is basically the same as nip44.   i see no problem with this proposal and would love it to be a standard.  relay authentication can be trivially handled in many ways:   - a special "auth" tag that the relay strips on send for example.
> 
> 
> 
> i would think that a simple json tag that the relay provider agrees to remove immediately is the best way to go for auth
> 
> 
> 
> it's easy to understand, and if they fail to remove it, they've only doxxed one ephemeral key... now you can ditch them and move to someone that honors it

Interesting thought. Some trust in the relay is always required anyways unless you use different proxies for every connection to cloak your up address... so you envision an auth-tag that has the "main" pubkey as well as a signature on it?

starbackr-dev:
@staab  Great idea. Can you point me to where I can read more on nsec bunker ?

madelainemichele:
> @staab Great idea. Can you point me to where I can read more on nsec bunker ?

Here is link to an article and utube video
https://www.nobsbitcoin.com/nsecbunker-announced/
nsecBunker setup demo    https://youtu.be/2oojFSoQ-So

59.md
```diff
@@ -0,0 +1,55 @@
...
+
+`draft` `optional` `author:kieran`
+
+This NIP defines a simple method to protect event metadata leakage by using randomly generated keys.
+
+A gift wrap event is a `kind 1059` event which wraps another event with a single use key.
+
+This event kind can be used to wrap DM's or other events which are considered private.
+
+This is similar in concept to Onion Routing.
```
earonesty, line 15: if the sender of the event would like to decrypt it at a later date, they can use their "tweaked" private key, rather than a random key when producing the event.   the "tweak value" can be located in an optional implementation dependent tag, or it can simply use the "created_at" field.  tweaking secp256k1 private keys is a well-known mechanism, and should be considered if a client wants to be able to read these events later from a message store, for example.

earonesty (approved):

59.md
```diff
@@ -0,0 +1,55 @@
...
+
+Example wrapped event:
+```json
+{
+    "id": "3731dc639ada388aa5ee8c3404351868d4dd0fdcac00183e540669ddc25e8ad4",
+    "pubkey": "4d02c0b4e16db38b80fa3b46548417a4ccc975337e29bc22da72132cf69f3b4a",
+    "content": "7sW2yHvC5o/x6b0xCKPpH8KZJQeQYxt3Y53mxMV53yGQ5kn/jgMM74k+vfwegYBY9zjlQeTvyF5gYuJIHM4KN2SZq7zaCIKqTqMjYhLPI4VjIvowhmMuuZgRlB7cestrYh6OSSteICye6sxkS1NdpON31u+JNK/GHFVRJqfYF4Fnzaka2u9WHQKmm9/nN040OHZ41eT3YUWpaZHupVx1fuN6UoeyPjb2PW4yiwxD72LhlAoNGjR+otpWZ3kD6LU3Btrz8btjSD272wJXT/FBX5Gs59BMCDeJ4eReNfbbVBB5mlXvVqVC/tnHntVgNvAP5GsZE56/4t+kLqql0dqZU/haC8wpOxqtzqzhzxTHxS/IQ3HYpDnU8XuXUqNw1zOQhp0F65rXZUf9nN0gIa/zGa35fK2ek2p8K/3exKssO7fVxaMeTVxP7jEoCmn47gsy5mv4fH7MtxoWvWMapvGFLZf58NPy54ktvldOdGp9BJLO+6/2P7KJqb7bvu0EIPdYCcUt+muYwiJCUoHNFfFQwKrrpr23XuPQlPYNPUs6orbNBs/Oj8vMfZC5hb/AUaNanGtsqsBgqfEufyonBtihMzLZqn9OZNWHE7lYfnhSOZnUjStl03Bd94Qx1rp7IHZN?iv=cIiZuLb2FMsLzc+yYf+3IQ==",
+    "kind": 1059,
+    "created_at": 1682286690,
```
earonesty, line 26: `created_at` can be made optional to save space, verifier assumes `created_at` is the same, and verifies the sig as if it is present (not as if it is excluded!).

59.md
```diff
@@ -0,0 +1,55 @@
...
+```
+
+The `content` is NIP-04 encrypted JSON string with the temporary key, this is the inner event:
+```json
+{
+    "id": "08eb6f89e916b0bd747c039d33c11be0575c4c6d2c08363ada1243ea771a92d2",
+    "pubkey": "63fe6318dc58583cfe16810f86dd09e18bfd76aabc24a0081ce2856f330504ed",
+    "content": "73ZaB2WzPlIwwC55YnhsFA==?iv=1R2WlyrAk9VQ1NdhNuWmsw==",
+    "kind": 4,
+    "created_at": 1682286690,
```
earonesty, line 44: created_at can be made optional to save space, verifier assumes created_at is the same, and verifies the sig as if it is present (not as if it is excluded!).

earonesty:
our metadata-hidden encrypted group chat uses this spec: https://github.com/ArcadeLabsInc/arcade/wiki/NIP-112:-Encrypted-Group-Chat

staab:
Based on the conversation [here](https://github.com/nostr-protocol/nips/pull/580#issuecomment-1577178529) and elsewhere, I'd like to request removing the dm-specific language from this PR. I think nip-44 gift wrap is a big deal, and is applicable to many other parts of the protocol.

59.md
staab: I don't think the additional complexity is worth the space saving.

59.md
v0l: Any change in how key material is handled needs to be added to NIP-07 or else it's not possible for me to use, this is why ephemeral keys work for me because i create the secret material.

v0l:
> Agreed. Kind-1004 may be a good one to pick.
> 
> Kind 4 - legacy DM encryption with metadata leaks Kind 1004 - Upgraded encryption with a fix to metadata leaks.
> 
> @v0l - your comments?

I think of this PR as just general purpose, it's not meant to replace DM's its just one use-case, so that is why the kind is just the nip number, i don't think there is any need to change it.

59.md
earonesty: agreed.  that would have to happen first!

mikedilger:
I like the simplicity of this NIP.

I like that the temporary private key can be immediately incinerated because it is the public key (in the event) and the recipient's private key that create the AES key for decryption.

I like that anything can be gift wrapped.

As for relay acceptance, at least in the case of NIP-65 "read" inbox relays, using my "personal relay" idea these would be accepted (along with spam and logic bombs and anything else that is addressed to me, but none of it allowed to be read back out).  I know this is probably a small sliver of the cases, but I thought I'd make the point.

I wonder if paid relays allow users AUTHenticated as `A` to push notes authored by `B`.  If so, while the relay can record this association, it could in principle keep that association secret.

59.md
arthurfranca: What do you think of adding a nonce key with a random string to the wrapped event JSON (not as a tag, but as an extra field)?

So for the DM example, you could send the gift wrapped DM with 2 1059 events from different random keys and 2 nonces, one to the recipient and another one to yourself (so you can keep chat history). With 2 nonces, the 2 gift wrap `content`s won't match, so no one can tell it is the same message.

This way you don't need to keep track of the wrapping key for decrypting chat history.

The down side is the created_ats will be close to each other and your copy won't have an `e` tag with the recipient so you cant fetch just the messages sent to a specific user. =(

arthurfranca: @staab nice so the sender should encrypt the receiver pubkey using the shared secret

But from NIP-04, is `secp256k1.getSharedSecret(SENDER_privkey, '02' + RECEIVER_pubkey).slice(1, 33)` the same value as `secp256k1.getSharedSecret(RECEIVER_privkey, '02' + SENDER_pubkey).slice(1, 33)`? So the receiver can use it to also encrypt his own pubkey and query with `{ #p: [encrypted_receiver_pubkey] }`?

staab: No idea, encrypted tags would rely on signing with a shared secret. This PR uses ephemeral keys, but it would be easy to convert it to using a shared secret (since one is necessary for decryption anyway). Encrypted tags could then be encrypted symmetrically.

earonesty:
fyi #580 uses this extensively.

earonesty:
from this comment:
https://github.com/nostr-protocol/nips/pull/580#issuecomment-1582998010

the sender can optionally use `sender private * event id mod curve.n` as the ephemeral key.   if this is done, the sender can "retain access" to the sent message and is capable of decrypting, deleting or otherwise operating on the outer message.

earonesty:
want to merg this already.   really using it a lot

59.md
```diff
@@ -0,0 +1,55 @@
...
+
+This NIP defines a simple method to protect event metadata leakage by using randomly generated keys.
+
+A gift wrap event is a `kind 1059` event which wraps another event with a single use key.
+
+This event kind can be used to wrap DM's or other events which are considered private.
+
+This is similar in concept to Onion Routing.
+
+To construct a wrapped `kind 4` you first start by producing the `kind 4` event as normal, then create a random key and produce the `kind 1059` by encrypting the `kind 4` as a JSON string and storing it in `content` of the `kind 1059` using the same `p` tag to notify the recipient.
```
staab, line 17:
```suggestion
To wrap an event, sign it as usual then serialize it to JSON. This string should then be encrypted using
NIP-44 with a different private key. Next, add the signed payload to the `content` field of a new
`kind 1059` event, signed with the same key you used to encrypt the content.

A `p` tag for the recipient MUST be added to the `tags` field on the outer event. To protect recipient identity, this may be a regular pubkey, or one representing a group of people sharing a private key for decryption.
```

Rephrased to make it clearer, and less about kind 4.

staab:
@earonesty @v0l let me know your thoughts on the suggested changes. All of it is optional, and solves some problems with other use cases I have in mind, albeit with weaker security guarantees. It will probably affect how #580 is implemented as well.

59.md
earonesty: this means if alice loses her key, all the messages she sent to bob can be read

v0l: Do we really need all these suggestions on tags? If somebody comes to read this NIP they will be confused about why there are so many suggested tags without more context, they should come to their own realization that you can do these things, or they can be explicitly called out, its too vague imo.

59.md
```diff
@@ -0,0 +1,55 @@
...
+        [
+            "p",
+            "63fe6318dc58583cfe16810f86dd09e18bfd76aabc24a0081ce2856f330504ed"
+        ]
+    ],
+    "sig": "48e50d7c2d301a494e0cf250685fe4a1d7d03bdb0efae38a0e9575d2dea04b8c227f8bc35fd0451f7e1d7828a8c5f7b2f81fdab135d7a3ef92c4364aaaa8514f"
+}
+```
+
+The `content` is NIP-04 encrypted JSON string with the temporary key, this is the inner event:
```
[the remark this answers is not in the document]
v0l: Not sure if we should call out NIP-44 because then we force people to implement that NIP first before being able to use this NIP.

Also I myself won't even be able to use it in snort because I rely on NIP-7 extension to support NIP-44 first.

59.md
staab: I'm not sure I understand your objection, the only specific I'm adding is `k`, which should be defined in the main table on the readme. But then you say it's too vague at the end? @earonesty and I had talked about this a bit [here](https://github.com/nostr-protocol/nips/pull/580/files#r1218625058), I added the suggestion about optionally copying tags because there are other alternatives (like encrypting tag values) that would be pretty bad. Are you looking for more examples here?

v0l: There should not be any requirements on tags, maybe you don't even need a p tag

staab: None of the above are requirements, just MAY depending on whether the use case for the wrapper requires filtering (I want to build large private groups on this NIP, which would). Maybe `p` isn't strictly required, but without it wouldn't you have to attempt to decrypt all wrapped messages to find the ones you want?

earonesty: without a "p" tag, it will be impossible to find the messages you want.   but the "p" tag could, technically, be for a derived key of some kind.   for example, alice may have a public key that she uses only when talking to bob.   in that case, she would only subscribe to events of people in her contact list, and couldn't get dm's from strangers.

so yes, a "p" tag is definitely required, but it need-not be the "p" tag of the recipient.    gift-wrap is a very generic construct

earonesty:
> @earonesty @v0l let me know your thoughts on the suggested changes. All of it is optional, and solves some problems with other use cases I have in mind, albeit with weaker security guarantees. It will probably affect how #580 is implemented as well.

imo:

- we don't need to go into detail about the "p" tag.  but i agree, it's required
- listing possible derivations for the ephemeral key is useful to show how the protocol is flexible

earonesty:
Suppose i want to build a search engine for nostr.   First, I index every note I can find, with the origin relay, and the keywords, and prunable content.

Then I hang up a shingle "post your search requests to me!"

 - I accept NIP-59 structured encrypted requests (KIND 10444 Search)
 - I reply with NIP-59 structured encrypted responses (KIND 10445 Search Response)

Gift-wrap is a good way of standardizing for any NOSTR RPC!

v0l:
Gift wrap code added on Snort: https://git.v0l.io/Kieran/snort/commit/a9f3216102521d18655d2e41864e12f9321895e8

59.md
```diff
@@ -0,0 +1,69 @@
...
+
+This NIP defines a simple method to protect event metadata leakage by using randomly generated keys.
+
+A gift wrap event is a `kind 1059` event which wraps another event with a single use key.
+
+This event kind can be used to wrap DM's or other events which are considered private.
+
+This is similar in concept to [Onion Routing](https://en.wikipedia.org/wiki/Onion_routing).
+
+To wrap an event, sign it as usual then serialize it to JSON. This string should then be encrypted using NIP-44 with a different private key.
```
staab, line 17: Are we actually switching to nip 44? It's not yet merged, and I'm not sure it's a good idea. Although this would be a good place to start, gift wrap doesn't have to match nip 04 DMs

v0l: I think it's a good place to start using it, though it does make it harder for people to support gift wrap, vs using the current nip-04 encryption.

I like NIP-44 simply because its versioned, we really need more things to be versioned

59.md
```diff
@@ -0,0 +1,69 @@
...
+The sender MAY copy tags from the inner event to the outer event. This weakens privacy guarantees,
+but may be useful to accommodate server-side filtering. A `k` tag MAY be added to the outer event to
+denote the `kind` of the inner event.
+
+Example wrapped event:
+```json
+{
+    "id": "0bb58c5e646ff1a310d888347cd392f0490d80d9ce963a43d10f0af0b3a92d58",
+    "pubkey": "67813274a3e44cc5f939cc29f1faa4e6b38f1ee4a22469262d6d9d19c377f027",
+    "content": "{\"ciphertext\":\"0s44ii9QNQMmKl/+Bmo0dUoVqCI3dqyPweoA1H06qkguf8u3Hr8sK7lUYKi/DHrR2PiLKuJRHC42mBn+e4jyM49yj+QL0cg3p5MbGiSmExEQPnofaBS6PtU/6Fa30XPNoywC9ptaPiDpEPNTxsI+LgCoCgUc3o9iJzKJRgzfaUVhjugjXmGw6Un8DgX7yfY3JBraxZClfC6FNV3yTtmRdi9AHmphJa5xW8l/XeI1k3C+Qv5z4mMI4GfiYXnqLbHtaMSeTaI31zmyr4nX9Bpnb14PSdiKpHEN1aAmJXgs/mXRw5KS44EDnjSJwItwow/yWNlGZZ2lT21ay9uc7LeD2rNOIOpW3xO5Pqd51NpYgjk+1lD9L8djQCpWRzsJoCr2/G1+qeGre+q/5g7re2aQbA3M9iSqEzBlkw841fq5AJYEDcsCXzFFQwo0ydJutBhzP2YOFuPbGgvWZat2bLRu26PZMMB3Z6ZprzyBrLIgZ4uOJ74/Rpflh4I5xl4r9Yvp37bLkxq25diG/78x0/AsBlgIJzMeNivZ7LCxm9a/nu5Ygj1H8g==\",\"nonce\":\"CvLX+kBOEBM5ZRm/I+4fncvGqbrcb6Dy\",\"v\":1}",
```
fiatjaf, line 38: Instead of JSON this should be a simpler format, easier to parse. Maybe prefix the base64 it with the version and add the nonce at the end with a simpler separator.

staab:
What do you guys think about making encryption optional? I can think of a lot of use cases when someone might want to wrap an event to "validate" it (e.g. moderation), or change its tags to make it match certain queries. We could also create a new NIP for that.

59.md
fiatjaf: Or maybe not.

earonesty:
> What do you guys think about making encryption optional? I can think of a lot of use cases when someone might want to wrap an event to "validate" it (e.g. moderation), or change its tags to make it match certain queries. We could also create a new NIP for that.

well, then you should just refer to the event id, right?

arthurfranca:
@staab you are describing the Event Copy PR.
[This comment](https://github.com/nostr-protocol/nips/pull/528#issuecomment-1556030505) describes synced (on relays) DM inbox use case.

staab:
@arthurfranca wow, ok, somehow I missed that one, thanks.

vitorpamplona:
Here's a different use case for this. 

We want to do paid-subscriber-only content and make it difficult (not impossible) to leak the paid content. We could use this Gift Wrap but with a twist: the `.content` is just any other **UNSIGNED** event kind and the wrap is signed by the regular private key of the author.

So, for Long Form content, for instance, the author creates a regular `Kind:30023` but *unsigned, and not id-ed with the hash*. The author would then take the unsigned payload, encrypt it with the author's regular key and place it on the Gift Wrap with the `p` tag of the receiver. The receiving client verifies the wrap and considers the inner event as valid. 

Why keep the `.content` unsigned? Because it forbids the receiver from re-broadcasting the inner event. The receiver can still expose all contents, but no one would be able to verify that the content is real. The only way to verify is to have the receiver's private key and verify the entire gift wrap itself. 

Clients would then consider any inner event of a GiftWrap as a real one, even though the inner event is not verifiable in itself, reusing all the rendering code of the usual event kind. We can send kind1s, zaps, private likes, etc. 

**Then all event types will have a Private option without even thinking about it.** 

It essentially generalizes kind-4 DMs for all event kinds. 

Now, this will leak metadata in the same way kind-4 does these days. But people wouldn't know which type the wrap actually is. 

To make it really private, we could then Wrap the Wrap with a new random key and send it to the destination. The first unwrap reveals who the sender is and the second wrap the original unsigned event.  

At this point, all people are seeing is when receivers receive a package. But they have no idea what type it is, when it was created (if dates don't match) or, more importantly, who is it coming from.

jb55:
On Fri, Jul 21, 2023 at 10:54:28AM -0700, Vitor Pamplona wrote:
>Here's a different use case for this.
>
>Why keep the `.content` unsigned? Because it forbids the receiver from re-broadcasting the inner event. The receiver can still expose all contents, but no one would be able to verify that the content is real. The only way to verify is to have the receiver's private key and verify the entire gift wrap itself.

yes this is a good idea

>Clients would then consider any inner event of a GiftWrap as a real one, even though the inner event is not verifiable in itself, reusing all the rendering code of the usual event kind. We can send kind1s, zaps, private likes, etc.
>**Then all event types will have a Private option without even thinking about it.**
>
>It essentially generalizes kind-4 DMs for all event kinds.

I've always wanted a generalized kind-4 for private likes, etc. If we can do it via gift-wrapping then that would be great.

I haven't looked at the spec until this post reminded me of it. I just took a look and it's very simple! Although I find the algorithm for determinism a bit confusing. In the private zap spec I got determinism by sha256ing the id, created_at and privkey. Maybe this is better but I don't know why.

>To make it really private, we could then Wrap the Wrap with a new random key and send it to the destination. The first unwrap reveals who the sender is and the second wrap the original unsigned event.
>At this point, all people are seeing is when receivers receive a package. But they have no idea what type it is, when it was created (if dates don't match) or, more importantly, who is it coming from.

:+1: probably best to do it this way

vitorpamplona:
I think I am getting a good grasp on this + https://github.com/nostr-protocol/nips/pull/574's encryption scheme. But I would bring them together in a different way. Let's take private DMs as a use case. 

This protocol obfuscates the DM metadata from relays and the general public WHILE being recoverable in any client with the receiver or the sender's private key. 

## 1. A new Kind 9 for Private DMs. 

This kind is always unsigned and will never be seen in the wild without a signed wrap. 

Complete Direct Message Payload: 
```js
{
    "content": "This is a DM. Notice that there is no sender or receiver here, no signatures, and no ids",
    "kind": 9,
    "created_at": 1686840217,
    "tags": [],
}
```

## 2. A new kind 10: Encrypted Unsigned Event

Kind 10s sign **any** unsigned event for distribution in the protocol. The unsigned event is placed with the XChaCha encryption scheme (NIP-44) inside the `.content`. Clients should treat inner fields as if they were fields of the kind 10 event (merge them). 

Encrypted Unsigned Event: 
```js
{
    "id": "<The usual hash>",
    "pubkey": "<Real Author's PubKey>",
    "content": "<NIP-44-Encrypted Kind 9>",
    "kind": 10,
    "created_at": 1686840217,
    "tags": [],
    "sig": "<Real Author's PubKey Signature>"
}
```

This event can be broadcasted, but, for DMs, this event should **not** be broadcasted automatically. 

Notice that there is no receiver exposed here even though the encrypted procedure encrypts only to the receiver's key. If this event leaks, it will only link to the sender. One will have to have the private key of the receiver to prove that this was a message for the receiver. 

## 3. A Gift Wrap to hide the sender from the public. 

Public Wrap (Kind 1059): 
```js
{
    "id": "<The usual hash>",
    "pubkey": "<Random, one-time-use PubKey>",
    "content": "<NIP-44-Encrypted Kind 10>",
    "kind": 1059,
    "created_at": 1686840217,
    "tags": [
        [ "p", "<Receiver>" ]
    ],
    "sig": "<Random, one-time-use PubKey Signature>"
}
```

The receiver is now exposed, but the sender is hidden. 

## 4. A Gift Wrap to recover the sender's DM history when changing clients. 

Public Wrap (Kind 1059): 
```js
{
    "id": "<The usual hash>",
    "pubkey": "<Random, one-time-use PubKey>",
    "content": "<NIP-44-Encrypted Kind 10>",
    "kind": 1059,
    "created_at": 1686840217,
    "tags": [
        [ "p", "<Sender>" ]
    ],
    "sig": "<Random, one-time-use PubKey Signature>"
}
```

By broadcasting this event, the private key of the sender can unwarp sent and received messages and reassemble the DM history in any client. 

The two wraps can be made separate with separate times to avoid time collision. And if we allow GiftWraps to contain more than one event (which I strongly recommend), then the 4th step can be made once at night with a sum of all messages sent during the day. 

The implementation is fairly straightforward: 
1. Implement NIP 44 encryption, serialized in a Json style like @v0l did in Snort
2. Implement kind 10s with overriding functions to pull from the inner event fields if present. 
3. Render kind10 overridden by kind9s as DMs. 
4. Implement Kind 1059, with overriding functions for the inner event. 
6. Pull all gift wraps with p = logged in user from relays and process the inner event accordingly.  

Interesting benefits. 

- Sender and receiver are not linked in public.
- Kind, tags, and created_at are hidden from the public. 
- Wraps create a bigger anonymity set because they can be anything, not only DMs, not only messages to me, but also from me.  
- The generic unsigned event allows us to take any Nostr event private. That's why we don't need an encrypted DM event anymore.

mikedilger:
Did you mean "NIP-44 encrypted kind 13"?  I didn't see reference to kind 10.  Then you mention kind9 too, which I don't see other reference to.

It appears to me that this scheme does indeed hide the metadata and correlation from the sender to receiver, while letting both sender and receiver be aware of the DM. I really like that part.

Is it is going to be a challenge to post from random pubkeys? Will relays just allow this based on the event kind? And if so, can that be abused, e.g. flooding relays with these events to store movies in RAR pieces?  Like I don't care if people store nasty stuff in encrypted form... that's their business.  But relays may not like being overloaded.

vitorpamplona:
> Did you mean "NIP-44 encrypted kind 13"? I didn't see reference to kind 10. Then you mention kind9 too, which I don't see other reference to.

I just fixed the references. I had to move away from 9 and 10 because somebody else is using. 

> I really like that part.

But yes, I have been playing with this for the past few days and I am starting to be a big fan of this approach. 

Here's a fully signed example: 

```json 
[
   {
      "content":"Hi There!",
      "created_at":1690496057,
      "kind":14,
      "tags":[
         [
            "p",
            "460c25e682fda7832b52d1f22d3d22b3176d972f60dcdc3212ed8c92ef85065c"
         ]
      ]
   },
   {
      "id":"7e0c882cb34e854e4f0cef259d6bc13dff2c971d1331f6fa0295c72aa356bf22",
      "pubkey":"0297bb9fbfa1342e297c12700cbaf2bccdc2719d6f33efc4c8c43b4ba5c015ff",
      "sig":"1c8254dad808c3c07c89c8b6f5c31662baa701969298444a87670eea4ef593cfb34d2662b854acf024bbc5a6ba1298931c7728958e5f10aabc843d3defc29e67",
      "content":"{\"ciphertext\":\"lIlnqsNPY/NciqdacncD/4dnEpbWuFI8grwjZdvnolemqUkKIrzqjptKVefVPGEgovBKDOGrrgtGtExv/vEmc6n0kXG4i0TOkbZ1R1T9JoUHYyPZrHZTtCrZgoQ1ts8BYK4vJFCxa5bb+XIdpFlSmqNcP6dCuvG6Pwyt7IvZ2oAfIcBPw2MX974kDrwnvDR1f25IoQGD/ANM0Vw=\",\"nonce\":\"kV4L/sbldT51QZssNx6DzT2tV7zFPiO8\",\"v\":\"XChaCha20\"}",
      "created_at":1690496057,
      "kind":13,
      "tags":[
         
      ]
   },
   {
      "id":"44d28b1a3af7eff3b0dc14a88c5695b8c41e2a9a29a05456d59d78e3f2e70d78",
      "pubkey":"d92134424f616305d77d2606b4a704799e9a02622b53adcd31f75a6b70131298",
      "sig":"b4441de5adabab484744c0bc841902a59d9cb554560a29faee9f0522d124385c30d03c3a0aac2822af5363e582e2e2aa86aa2c36e08780bd6603802cf41b937a",
      "content":"{\"ciphertext\":\"zBlNKoeQJOuNiv/F4WBAzjFjWPpWHnsSzcgQpwvu8kJ27E5DSjk7JbITJm8M939KLWduvw9W88a+eQPdvYHQE08VTJgViOIcQnbynSyImch2wLFKqjBdgVE0sS7fdwmRqD0+JYQ3cBjJExiwndlqPIcAa6+kT1w/0V4rDKpMHXBxY/NGB0SndYAIkdBNRSMkqIT6HjhnQsamHgK6UyT3kCd8syKuWviqgDmcWN14I3pPT2sQzzTCF2BqQFD7jmABXqmisJHg9ceWWXjyPPFT8m634Kf57hd8JXhCLIL/Sj4N1QcW2l9zolmtE1ToDdN+97C8AbJrfYlk8BRCINtrF/enQKlSYIfWW1ZhckpgutvuGz0u1OwOE0SIYa3Ec3242Z+cv9lm/rR3Fy3DzxbFUgXRmamh+jeZoOHMrmSsGYLzqE07fVpxBG5HeRXy7u6hEaYvu7YD7KBoAxzJZY+UnNP1ZNBAXYyc4JvE1iINmrxp8Jhqf1ATH1EX+EOq925+UFyE/Od86JsqRJADeQ0Xj5kKCaZ1K0ZHm59hJ1vS0QgAZb+tdcHgXpEf2i6w2VhIsf24zhfCZ0mULw40AvL4ckKTZWAHwNe0/bt8f4yZMlyzExQs+qNl6ui44Uc5w8vFcVkOioKiAOEzkFM9bubb29wAh1ritzIDOAjvbNspST5Lj1CDOpjCHCRZ2UXtQS1yJMRZ17ldCxwPCI0t7c+5q88quwdhzcLmehRzRYOuFXi0sGpAZcQNA17Idvtg802wwI5SMG98Hf0k58ULzBHiwScuyQp1MTL9J2ahzSuQy6kdQFq3yu7L4YWomnlZ1n1ne2Xas7P9JW5s1lXuWzIc/O9W/S7E9j/vPlUxAC1wsQ==\",\"nonce\":\"gcfRZJoMI32mq252LBBqre94wnwimn71\",\"v\":\"XChaCha20\"}",
      "created_at":1690496057,
      "kind":1059,
      "tags":[
         [
            "p",
            "460c25e682fda7832b52d1f22d3d22b3176d972f60dcdc3212ed8c92ef85065c"
         ]
      ]
   },
   {
      "id":"4a61e4ee7c60e7df1eac23b33d54761ad93f6003c7a02efe84da418c4b0f2925",
      "pubkey":"781c599bc1e2a351b6c240c9f3772a18dcc80a3ff861a9ba74911fd789aa0412",
      "sig":"cf8a0db24deb777108261c306af7d8fe3de288f9ae20793c37f154df9ddd1ff23bc9a2a94510a2ffacb0fa0e976dc1cb4e387818a74a7558025d629f88151adf",
      "content":"{\"ciphertext\":\"2Y2J4PQv4S+1RxJEfYVPqt47SXyQtNBlLsTs/BqDQjQ57alK3YUF/TBv3i5xDT86Ne46hlS9nL38NLQzsc0PfynqomgaixugJiEDk0U5mSr1eaEKI4jr8SKiwnXe3axCy4W0CBUpa7GO/UY87B8sq0w4oMJLK5y1cbuVQdGccIvYbUP12BNrnS8FmFFg/QxyPklpkhZqU9YvZGfZKNaRGuQVqC9uxopvXZ+moMBZGUulqGRe0ajo/KZXCFBbJBuOt2m+B6NR13PaP2Qur6PuHnOrSE/rQwaEU6Bd6af4wM0duA1v9AG56MUrlsLNwlCG4/jsXYqBdVRo+Rniv86Inkvwm1WgXpeK24J6LzIjZHp2Qf47JbbGNVeKDtb9IstVF07Bd/6ZBS9U4Ux665XqqLbNxDReZTg3KwtZWAVuLkv22dz1CmMCRgs8xLkeufszY63EVeZakBcZKFYKQLhkNuX1XG2CVodGTBHs/WbyR8ms74mPR12v1LlvBbV6j3BOAwp4MPkxX3E7SYsmK8ibrikGsVg9qwBiwoswD9TzJU+yNftXVIRVjBkcJOBzAMj0lWgTDsAjzO2kHlAU5OfHYEw8aBlJFURnPUwAwclzdTwUhsoTx2RNnBW2m3zNpzShepzXLE+oqIs7XUZFzUBZDbNJq/pXNCY9XoH0//qiwPxsrEsPDxBMZUvJq/VkVVM6z/0Cy2FfuJwdghGDcOAnM4iBclgB+pLn5RvgVkFdlEQUf7jPN+S8dE8iDr/d1oZ2efd2zYAIN16y0+rmPxjPm9e3pkRertCX9/fgdoUM0UmRwDXYxBUkrRuSmoLVKVPb2j+YcDQRqgstxBlLkw07DGJHbWs3Y4ECZmm0CYeCEw==\",\"nonce\":\"JVmEX3iHlyWelnkeadLWxiuSITg5ddUZ\",\"v\":\"XChaCha20\"}",
      "created_at":1690496057,
      "kind":1059,
      "tags":[
         [
            "p",
            "0297bb9fbfa1342e297c12700cbaf2bccdc2719d6f33efc4c8c43b4ba5c015ff"
         ]
      ]
   }
]

mikedilger:
I don't like that the inner encrypted data is described as a nostr event which does not have an id, pubkey, or signature.  It isn't a nostr event, doesn't require the same structure, and shouldn't be called an "event".

vitorpamplona:
We can create another name. The only point is that any event can be unsigned and put it into a warp to hide it's nostr properties from the public.

earonesty:
its not necessary to leak any keys

you calculate the shared secret between sender and receiver

then you use that as the private key and post an encrypted message

you can calculate this for all your follows

and send a subscription for the authors

now you have two way blinded dms

earonesty:
blind dms are implemented here

https://github.com/ArcadeLabsInc/arclib

also encrypted group chat that uses gift wrap

mmalmi:
> you calculate the shared secret between sender and receiver

I agree. Would be cool if window.nostr had a method for getting a hash(shared secret + your pubkey). That hash could be used as the DM recipient, so the actual recipient is not revealed.

That's actually what I used to do in the old Iris which used gundb, and it worked nicely.

earonesty:
yeah, we do this with the arcade app, and it allows blinded dm's .. we don't use gift wrap for that, because that leads to outbound subscription issues for state reloads.

vitorpamplona:
> outbound subscription issues for state reloads

Hum... what do you mean by that? I seem to be running fine.

earonesty:
> > outbound subscription issues for state reloads
> 
> Hum... what do you mean by that? I seem to be running fine.

you cant load the history of outbound messages when you log out and log back in... because with gift-wrap you have no idea who sent it.   but if you use the hash(shared-secret + pub) as a private key... then you can easily subscribe to that author on the relay, and see all the fully blinded dm's

vitorpamplona:
> you cant load the history of outbound messages when you log out and log back in... because with gift-wrap you have no idea who sent it. 

ohh I see. Yes, I agree. That's why I am sending a GiftWrap to myself as well. 

> hash(shared-secret + pub) as a private key

How do you communicate the shared-secret to the other side? 

> because you don't need any tags

There should be at least one tag of the subscription id, no?

earonesty:
> > you cant load the history of outbound messages when you log out and log back in... because with gift-wrap you have no idea who sent it.
> 
> ohh I see. Yes, I agree. That's why I am sending a GiftWrap to myself as well.
> 
> > hash(shared-secret + pub) as a private key
> 
> How do you communicate the shared-secret to the other side?

it's the standard shared secret (my priv * your pub)

> 
> > because you don't need any tags
> 
> There should be at least one tag of the subscription id, no?

no need to have tags on blinded dm's.   no receipient, no sender.   everyone just subs to the join well-known shared-secret key

purely private dm

vitorpamplona:
> no need to have tags on blinded dm's. no receipient, no sender. everyone just subs to the join well-known shared-secret key

I am confused on this. Can you provide a Nostr filter example? Do they filter by author or by ptag?

vitorpamplona:
> it's the standard shared secret (my priv * your pub)

In this case, you won't receive messages from strangers, right? Because the app woundn't be able to subscribe to each other active pubkey out there.

earonesty:
correct, our system does not allow messages from strangers.   also we allow private follows too (encrypted in a replaceable event)

here's the sub we use:

https://github.com/ArcadeLabsInc/arclib/blob/542ececf9ea03e9d375d2905820376e98ecba2aa/src/private.ts#L161

vitorpamplona:
> correct, our system does not allow messages from strangers

Yeah, that's a no go here. 

> https://github.com/ArcadeLabsInc/arclib/blob/542ececf9ea03e9d375d2905820376e98ecba2aa/src/private.ts#L161

Oh I see. So, you do have a "tag", but your tag is the pubkey. It's just that you are using that pubkey as a central point.

earonesty:
> 

no I think that code is just to deal with various legacy stuff

basically you can tag or not tag depending on how much metadata you want to leak vs whether or not you want to allow strangers to message people.  we chose more private but I imagine a good write-up of the protocol would allow for both

vitorpamplona:
> no I think that code is just to deal with various legacy stuff

If two users are communicating with the same Private and Public Key, how do you know which one sent each message without tags?

vitorpamplona:
ohh I see what you are doing. You have a custom payload where you place the pubkey of the writer together with the message in the `.content` of the event. So, you this scheme, you have to trust the counterparty is not adding your pubkey in the message.

jb55:
@earonesty 

> the sender can optionally use `sender private * event id mod curve.n` as the ephemeral key (derived key, not ephemeral). if this is done, the sender can "retain access" to the sent message and is capable of decrypting, deleting or otherwise operating on the outer message. this may not be strictly _needed_ but could be useful if a sender wants to "retract" a statement in a group-chat... for example. I would propose adding this as an OPTION to the NIP-59 gift wrap.

I just re-read over this spec again and this is the only glaring thing that appears to be missing. Why no mention of deterministic gift wrap keys? Are clients just supposed to fire off random gift wraps with random keys and have no way of fetching them again? Maybe I'm missing something.

vitorpamplona:
You make an extra one for yourself firing from yet another random key. It's actually really good. You can choose "Disappearing Messages" if you don't give yourself a gift for your own DMs, for instance. It happens to make a lot of sense. 

In summary, you only need one filter p=your key to ALL gift wraps. Including those from and to yourself.

jb55:
nm I see it in there now, not sure how I missed it

jb55:
I see why I was confused, I was looking at the modified one by @vitorpamplona which doesn't mention deterministic ephemeral keys. who's the real gift wrap spec ?

https://github.com/nostr-protocol/nips/blob/1c9c9719e068b4dbd405133decb4bf1ec8d367fe/59.md

vitorpamplona:
We have gift wraps now in 2-3 PRs, but I think they are mostly the same, just applied in different ways to add protections. The [one I did](https://github.com/nostr-protocol/nips/pull/686) is basically a double wrap of an unsigned event to make sure the wrapped event cannot leak and expose metadata. 

Then @staab created a version for large groups with Group management events to try to reduce the amount of GiftWrapped messages in those large group implementations (1,000+ people).

staab:
Right, pull request #706 is my version. It distills double wrapping from Vitor's version down to its pure form (and has a distilled form of nip 44 as well). NIP 87 in the same PR is largely un-reviewed. If there's interest in getting 44 and 59 in, I can separate those into separate pull requests, or we can update these. I think the DM-specific stuff in the current versions should be superseded by Vitor's PR.

staab:
I've created a new version of this PR at #716, @v0l please let me know if you'd like to close this PR or incorporate those changes into this one.

v0l:
Replaced by #716

PR #476, opened 2023-04-27
NIP-123 Epheremal bitcoin transaction broadcasting

benthecarman:


pablof7z:
this is awesome, I'll be able to retire https://psbt.io

we'd need a tag like

`["network", "signet"]`

benthecarman:
good idea! Might be better to use network magic so we can delineate between signets as well.

melvincarvalho:
Nice work.  JS Implementation:

https://gist.github.com/melvincarvalho/70c29313554ce9d1a22ce9bf954909d9

arcbtc:
Excellent work Ben!
We already have a bunch of things we wont to do with it, very useful indeed.

melvincarvalho:
It may be worth pointing out that since every nostr account is a taproot address, this NIP allows the sending of on chain coins from any nostr account to any other, using only existing existing nostr infrastructure and potentially with increased privacy.

It would then be a fairly simple task to look up utxos for a given nostr npub and build, for example, a wallet.

melvincarvalho:
> this is awesome, I'll be able to retire https://psbt.io
> 
> we'd need a tag like
> 
> `["network", "signet"]`

If you add testnet and signet, it would be possible to build a web wallet and demo.

fiatjaf:
Can you give it a sub-100 NIP number?

ronaldstoner:
This is great to see. TeamMunstr (@ronaldstoner, @satsie,  @0xBEEFCAF3) has been working on a similar NIP proposal this week for use cases such as multisignature. Since bitcoin transactions can be multiple types (MuSig requiring nonce[s for example) we wanted to ensure the data payload and tags encompass all transaction use cases. Munstr would need something like nonce sharing in order for this type of data type to work.

### **NIP-XX
Partially Signed Bitcoin Transactions**

This NIP defines an event type for partially signed bitcoin transactions (PSBT) that can be transferred via the Nostr relay network. `XXXX` is the event kind for a PSBT, representing either a partially signed or fully signed (meets signature thresholds) bitcoin transaction. 

PSBTs transmitted via the Nostr network provide a decentralized communication layer, encryption and digital signature benefits with no username and password required for participants to coordinate and sign transactions using separate sovereign bitcoin private keys and hardware wallet devices. 

### Reference and Examples
Munstr - https://github.com/0xBEEFCAF3/munstr
JoinStr (Coinjoin over nostr) - https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2022-August/020875.html

**Example**
```
{
"id": <32-bytes lowercase hex-encoded sha256 of the serialized event data>,  
"pubkey": <32-bytes lowercase hex-encoded public key of the multisig creator>,
"kind": <integer>, 
"tags": [ 
	         [“n”, <something something integer>],
             ], 
"content": <arbitrary string>, 
"sig": <64-bytes hex of the signature of the sha256 hash of the serialized event data, which is the same as the "id" field> }
```

### Explanation of fields:

```
id: The sha256 hash of the serialized multisig payload.
pubkey: The public key of the creator of the multisig.
created_at: A Unix timestamp indicating the time when the multisig transaction was created.
kind: An integer value used to indicate the type of event. For multisig transactions, the kind is set to XXXX.
tags: An array of tags to capture additional information about the multisig transaction:
    “n”: (optional) Aggregated nonce. tag provided by musig coordinator 
“relays” (optional) tag provides relays to broadcast the transactions
content: Additional information about the transaction in string format.
sig: The signature of the sha256 hash of the serialized event data, which is the same as the "id" field.
```

**Note:**
The multisig transaction is a new event type, with the "kind" field set to XXXX. This uses a specific event type so that clients do not need to render PSBTs the same as they do notes, invoices, and other event kind types. If clients wish to display PSBTs they can choose to do so in whatever format they wish (string, QR Code, view render). 

The optional “n” tag provides a nonce(s) for certain multisignature transactions.

The "relays" field is optional and can be used to provide relay URL recommendations to clients for improved censorship resistance.

**Current Questions:**
Can we encrypt the tag fields similar to kind 4 DMs to protect the bitcoin transaction and privacy further?

benthecarman:
> Can you give it a sub-100 NIP number?

Changed to NIP-89

benthecarman:
Updated to now include network magic so we can identify which network to broadcast to

benthecarman:
@ronaldstoner that looks like you are serving a different usecase, this is purely for broadcasting transactions, not constructing them.

motorina0:
I have updated the [Republish Events by Peers (NIP-705)](https://github.com/motorina0/nips/blob/republish_events/705.md) proposal to cover any ephemeral key events (`bitcoin tx broadcasting` included).
`NIP-705` defines a way for a client to indirectly publish events to a relay with the help of its peers.
`NIP-705` improves privacy by hiding from the relay(s) the IP of the client who produced the event (`bitcoin tx` in this case)

PR: https://github.com/nostr-protocol/nips/pull/430

ekzyis:
Sorry if being dumb, but I expected that the example content is a valid transaction. So I would expect that it has already been broadcast. However, I don't seem to find it. I did the following:

```
❯ echo -n "AgAAAAABAtkXo8t2LfqIwQ+BPQg/LZdqAtAb61bY0zIjFvvxMTtGAAAAAAD9////j8wvOaFL5Dz+0CTKXNKxxv1GrusMps3aodsXQGp95xgAAAAAAP3///8CLAPTBQAAAAAiUSBIKoy1/JK1GdfWhii8KLfBooZg3qT7ImuwwwC6t2eT6oiKAQAAAAAAIlEgZtFI4Bx2rDqBe4X9+lzEhh7Kxca3tNE9l/QfH/ZghQECRzBEAiBV2O5+jskV8/H+UuzgRuku2gCAwlmErieXrS5scbongAIgLWD6uN8BGpyE7gNtX4jxnkNShY8yjhdcjhsHR4RAWJEBIQLlCnhSR+K9YoAngUWj94mptHhcVYJGi+rT3lwApIekkQJHMEQCIDwrK0YHftZ2BQsBjHdJb89Tlpzvc3z8E1wGgbqIVRBQAiB6J0UXuPjib+geVmaZeoq4vYD+fDz835o5OJRg9ITFDAEhA1oq2Zj0d4XbATjc3OUSXSJDvvIosDeAsHyegwk4fkcTNAslAA==" | base64 -d > tx.dat
$ xxd tx.dat 
00000000: 0200 0000 0001 02d9 17a3 cb76 2dfa 88c1  ...........v-...
00000010: 0f81 3d08 3f2d 976a 02d0 1beb 56d8 d332  ..=.?-.j....V..2
00000020: 2316 fbf1 313b 4600 0000 0000 fdff ffff  #...1;F.........
00000030: 8fcc 2f39 a14b e43c fed0 24ca 5cd2 b1c6  ../9.K.<..$.\...
00000040: fd46 aeeb 0ca6 cdda a1db 1740 6a7d e718  .F.........@j}..
00000050: 0000 0000 00fd ffff ff02 2c03 d305 0000  ..........,.....
00000060: 0000 2251 2048 2a8c b5fc 92b5 19d7 d686  .."Q H*.........
00000070: 28bc 28b7 c1a2 8660 dea4 fb22 6bb0 c300  (.(....`..."k...
00000080: bab7 6793 ea88 8a01 0000 0000 0022 5120  ..g.........."Q 
00000090: 66d1 48e0 1c76 ac3a 817b 85fd fa5c c486  f.H..v.:.{...\..
000000a0: 1eca c5c6 b7b4 d13d 97f4 1f1f f660 8501  .......=.....`..
000000b0: 0247 3044 0220 55d8 ee7e 8ec9 15f3 f1fe  .G0D. U..~......
000000c0: 52ec e046 e92e da00 80c2 5984 ae27 97ad  R..F......Y..'..
000000d0: 2e6c 71ba 2780 0220 2d60 fab8 df01 1a9c  .lq.'.. -`......
000000e0: 84ee 036d 5f88 f19e 4352 858f 328e 175c  ...m_...CR..2..\
000000f0: 8e1b 0747 8440 5891 0121 02e5 0a78 5247  ...G.@X..!...xRG
00000100: e2bd 6280 2781 45a3 f789 a9b4 785c 5582  ..b.'.E.....x\U.
00000110: 468b ead3 de5c 00a4 87a4 9102 4730 4402  F....\......G0D.
00000120: 203c 2b2b 4607 7ed6 7605 0b01 8c77 496f   <++F.~.v....wIo
00000130: cf53 969c ef73 7cfc 135c 0681 ba88 5510  .S...s|..\....U.
00000140: 5002 207a 2745 17b8 f8e2 6fe8 1e56 6699  P. z'E....o..Vf.
00000150: 7a8a b8bd 80fe 7c3c fcdf 9a39 3894 60f4  z.....|<...98.`.
00000160: 84c5 0c01 2103 5a2a d998 f477 85db 0138  ....!.Z*...w...8
00000170: dcdc e512 5d22 43be f228 b037 80b0 7c9e  ....]"C..(.7..|.
00000180: 8309 387e 4713 340b 2500                 ..8~G.4.%.
$ sha256sum tx.dat | cut -d' ' -f1 | sha256sum
03bac0c86cae47f7813ec5841a53287c54eb43b3a153cce180b583c2b637f49d  -
```

But there seems to be no transaction with this ID: https://mempool.space/de/tx/03bac0c86cae47f7813ec5841a53287c54eb43b3a153cce180b583c2b637f49d

Am I doing something wrong?

89.md
```diff
@@ -0,0 +1,53 @@
...
+
+Nostr provides a better opportunity for users to broadcast transactions in a private way. Nostr relays are at most
+bitcoin adjacent and have a much lower risk of being run by a chain analysis company, thus providing a possibly better
+way to broadcast transactions.
+
+## Recommendations
+
+Clients should generate an ephemeral keypair for each transaction they want to broadcast. This keypair should only ever
+be used for broadcasting that transaction. This is to best preserve the privacy of the user and to not link any extra
+metadata to the transaction.
```
TonyGiorgio, line 53: Not sure if it's worth mentioning, but less is really more when it comes to broadcasting the transaction, especially if it's for privacy reasons.

If someone is doing this for privacy reasons, broadcasting to a lot of relays will maximize the chance to broadcast to a relay that is monitoring IPs and associating events with them. For now, it might not be the case that CA companies are running relays just to eavesdrop on bitcoin transactions being broadcasted, but over time there could be a bunch of events that are juicy from an bitcoin analysis perspective.

benthecarman:
@ekzyis you can't just hash it with segwit txs, also it is on segwit, this is the transaction

https://mempool.space/testnet/tx/94f46759c50272034cfb00b944a8e5ef3bbff12c90b3f889acaf9bfa61c58969

89.md
benthecarman: Added a section about it

benthecarman:
Added a section about recommendation around selecting relays.

ekzyis (approved):

kristapsk:
> Hacked together the transaction broadcaster real quick: https://github.com/benthecarman/nostr-tx-broadcast

Why mempool.space not local bitcoind node? Doing `sendrawtransaction` RPC seems even simpler than mempool.space API.

monty888:
A rough version of this at https://github.com/monty888/txbroadcastr
added endpoint to blockstream, and will add to broadcast via local bitcoind 
also a poster that will monitor a given directory and post raw tx files that you drop in that directory and post them to nostr to be broadcast, for example after signing on coldcard.

benthecarman:
> Why mempool.space not local bitcoind node? Doing sendrawtransaction RPC seems even simpler than mempool.space API.

Plan to add that, was just hacking it together and didn't want to deal with having to add config options

89.md
```diff
@@ -0,0 +1,62 @@
...
+
+`draft` `optional` `author:benthecarman`
+
+A special event with kind `28333`, that contains an encoded bitcoin transaction.
+The bitcoin transaction is encoded as a base64 string, and is contained in the `content` field.
+
+The network magic is contained in the `tags` field. This is to identify the network that the transaction is for.
+The common ones are:
+
+- mainnet: f9beb4d9
```
melvincarvalho, line 15: Slight nit:

| Network        | Magic value | Sent over wire as |
|----------------|-------------|-------------------|
| main           | 0xD9B4BEF9  | F9 BE B4 D9        |
| testnet/regtest| 0xDAB5BFFA  | FA BF B5 DA        |
| testnet3       | 0x0709110B  | 0B 11 09 07        |
| signet(default)| 0x40CF030A  | 0A 03 CF 40        |

https://en.bitcoin.it/wiki/Protocol_documentation#Common_structures

There are so far 3 testnets : perhaps align with the descriptions above ie testnet3 instead of testnet

benthecarman: updated testnet to testnet3, otherwise I think this is enough information for the user to understand

89.md
```diff
@@ -0,0 +1,62 @@
...
+
+This is an ephemeral event, and is not stored on the server. So relays or other custom clients could receive these
+events, and broadcast them to the bitcoin network.
+
+For example:
+
+```json
+{
+  "kind": 28333,
+  "content": "AgAAAAABAtkXo8t2LfqIwQ+BPQg/LZdqAtAb61bY0zIjFvvxMTtGAAAAAAD9////j8wvOaFL5Dz+0CTKXNKxxv1GrusMps3aodsXQGp95xgAAAAAAP3///8CLAPTBQAAAAAiUSBIKoy1/JK1GdfWhii8KLfBooZg3qT7ImuwwwC6t2eT6oiKAQAAAAAAIlEgZtFI4Bx2rDqBe4X9+lzEhh7Kxca3tNE9l/QfH/ZghQECRzBEAiBV2O5+jskV8/H+UuzgRuku2gCAwlmErieXrS5scbongAIgLWD6uN8BGpyE7gNtX4jxnkNShY8yjhdcjhsHR4RAWJEBIQLlCnhSR+K9YoAngUWj94mptHhcVYJGi+rT3lwApIekkQJHMEQCIDwrK0YHftZ2BQsBjHdJb89Tlpzvc3z8E1wGgbqIVRBQAiB6J0UXuPjib+geVmaZeoq4vYD+fDz835o5OJRg9ITFDAEhA1oq2Zj0d4XbATjc3OUSXSJDvvIosDeAsHyegwk4fkcTNAslAA==",
```
joostjager, line 30: Is it useful to make this an array so that the event can contain transaction packages right from the start?

benthecarman: Very interesting idea! We could make the content a json array of transactions to support this. Another alternative could be to use tags, either using an `e` tag to reference another bitcoin transaction event or add a new tag for a parent transaction.

fiatjaf: It's better to use tags for everything so the content can be something stupid like the string `"a bunch of bitcoin transactions to be broadcasted"` that can be shown on nonsupporting clients.

fiatjaf:
This is super cool, but is there a real use case?

89.md
joostjager: Not sure about the `e` tag to reference another bitcoin transaction event. It seems to complicate implementations again similar to bitcoin p2p, because they need to look for / wait for the referred transaction before making a decision whether the complete package is worth including in the mempool.

joostjager:
As a bypass for bitcoin p2p transaction relay, this is potentially a quicker way to get package relay in place.

Package relay will be enormously helpful if not security critical for protocols such as lightning that use presigned transactions with a fixed fee rate.

fiatjaf:
Getting package relay up would be great, but is anyone actually in fact willing to implement and use this in the real world or is it just a potentially good idea for now?

89.md
```diff
@@ -0,0 +1,62 @@
...
+Today, users have to connect to the bitcoin network directly or use an api like mempool.space to broadcast transactions.
+This is not private, as you can leak your ip address to the service or a bitcoin node.
+
+Nostr provides a better opportunity for users to broadcast transactions in a private way. Nostr relays are at most
+bitcoin adjacent and have a much lower risk of being run by a chain analysis company, thus providing a possibly better
+way to broadcast transactions.
+
+## Recommendations
+
+Clients should generate an ephemeral keypair for each transaction they want to broadcast. This keypair should only ever
```
fiatjaf, line 55: What if people just used `G` as the key always instead of creating new keypairs all the time?

ekzyis: Was thinking about using a static key, too.

But I think creating a new keypair (and deleting them afterwards) is more secure for individuals. With a known public keypair (with secret key also known since we need to sign events), there could be the chance that one may use this accidentally for other purposes. Like if the last used key is reused for something sensitive and then everyone else will be able to decrypt it.

However, not sure if this is a valid concern. But that I can even come up with something like this also tells me, that maybe the solution to just create keypairs and delete them is safer. Better safe than sorry in privacy/cybersecurity. I don't think it's too inconvenient to generate a new keypair every time to start using a less common mechanism for privacy.

Or are there other protocols which implement privacy by letting everyone use the same key?

Also, if I ever use this, I would go with the "generate my own keypairs for every event"-solution. So if enough people think like me, the people who use the same keypair will naturally be at a disadvantage (anonymity set becomes smaller and smaller). So they will be incentivized to generate their own keypairs, too.

So I think in the long run, `G` won't be used.

benthecarman:
> Getting package relay up would be great, but is anyone actually in fact willing to implement and use this in the real world or is it just a potentially good idea for now?

This is something we want to support in mutiny. Obviously I am biased but I think it could other projects implementing this given all the recent issues with tor.
https://github.com/MutinyWallet/mutiny-node/pull/459

joostjager:
>Getting package relay up would be great, but is anyone actually in fact willing to implement and use this in the real world or is it just a potentially good idea for now?

Implementation would probably be not much more than mining pools listening in to these events and including the transactions in their infinite mempools for maximum profit?

If the incentives are right, it will happen? Otherwise it at least is a potentially good idea.

89.md
ekzyis: > Since we can't ban people from not using G it could create a fingerprint of clients that use G and ones that don't.

Well said, that's basically what I wanted to say, too, haha

benthecarman:
Added a `parent_txs` tag to support package relay.

89.md
```diff
@@ -0,0 +1,76 @@
+NIP-89
+======
+
+Bitcoin Transaction Broadcasting Events
+-------------------------
+
+`draft` `optional` `author:benthecarman`
+
+A special event with kind `28333`, that contains an encoded bitcoin transaction.
```
joostjager, line 9: plural

README.md
```diff
@@ -88,6 +89,7 @@ They exist to document what may be implemented by [Nostr](https://github.com/fia
 | `10002` | Relay List Metadata        | [65](65.md) |
 | `22242` | Client Authentication      | [42](42.md) |
 | `24133` | Nostr Connect              | [46](46.md) |
+| `28333` | Bitcoin Transaction        | [89](89.md) |
```
joostjager, line 92: plural?

89.md
```diff
@@ -0,0 +1,76 @@
+NIP-89
+======
+
+Bitcoin Transaction Broadcasting Events
+-------------------------
+
+`draft` `optional` `author:benthecarman`
+
+A special event with kind `28333`, that contains encoded bitcoin transactions.
```
joostjager, line 9: Do you want to use the term 'package relay' anywhere in this NIP? I think it would be nice.

benthecarman: Referred to them as transaction packages in the recommendation section. Probably can add more color though

joostjager: Ah yes, missed that. Nostr package relay sounds good too though.

joostjager: Add 'package' to the NIP title? In my view, this is a great selling point for the NIP.

benthecarman:
Okay sorry for all the little pushes, should be updated for package relay now

benthecarman:
Okay sorry for all the little pushes, should be updated for package relay now

benthecarman:
> What do you think about an update to the title ([#476 (comment)](https://github.com/nostr-protocol/nips/pull/476#discussion_r1186137578)) ?
> 
> `NIP-89 Ephemeral bitcoin transaction package broadcasting`

I changed it to `Bitcoin Transaction Package Relay` seems that is clear enough

benthecarman:
This is ready for re-review/merge

fiatjaf:
I think this is great and I dream with the day in which the Bitcoin mempool will live on Nostr, but before merging I think we should wait to see it working in the real world first, just in case. Meanwhile this stays here reserved.

89.md
```diff
@@ -0,0 +1,81 @@
...
+The common ones are:
+
+- mainnet: f9beb4d9
+- testnet3: 0b110907
+- signet (default): 0a03cf40
+- regtest: fabfb5da
+
+If another value is present it is likely a custom signet.
+
+The `tags` field MUST contain a `transactions` tag that contains a list of bitcoin transactions encoded in base64.
```
joostjager, line 22: Why base64? Isn't bitcoin core's standard for transactions hex?

fiatjaf: Hex is so much more pleasant!

benthecarman: Yeah did to save on bandwidth, since these are ephemeral I guess it is not a big deal since they shouldn't be stored.

89.md
```diff
@@ -0,0 +1,81 @@
...
+
+A special event with kind `28333`, that contains encoded bitcoin transactions.
+The bitcoin transactions are encoded as a hex string, and are contained in a `transactions` tag.
+
+The network magic is contained in the `tags` field. This is to identify the network that the transactions are for.
+The common ones are:
+
+- mainnet: f9beb4d9
+- testnet3: 0b110907
+- signet (default): 0a03cf40
```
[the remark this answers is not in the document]
ekzyis: I think since money is involved, I think it's good to be explicit when using mainnet.
Don't want someone send something on mainnet by accident. However, even addresses contain net information:

> Invoice addresses on the Bitcoin Testnet are generated with a different prefix. See [List of address prefixes](https://en.bitcoin.it/wiki/List_of_address_prefixes) and [Testnet](https://en.bitcoin.it/wiki/Testnet) for more details.

-- https://en.bitcoin.it/wiki/Invoice_address#Testnet

But I think for signet and regtest this is not the case? :thinking:

89.md
```diff
@@ -0,0 +1,81 @@
...
+- regtest: fabfb5da
```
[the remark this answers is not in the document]
ekzyis: > Might be better to use network magic so we can delineate between signets as well.

-- https://github.com/nostr-protocol/nips/pull/476#issuecomment-1525038776

89.md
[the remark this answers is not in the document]
joostjager: No objection to being explicit, but maybe then always be explicit? Signet doesn't look like such a natural choice for a default to me.

89.md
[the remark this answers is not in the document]
joostjager: Right. But I suppose those alternative signets will also get their own name?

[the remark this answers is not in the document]
benthecarman: I think this is better, this way we don't get any overlap, this could even be used for things like litecoin without needing new specifications.

I do agree just writing "mainnet" is more clear but I figured just enumerating them here would make it easy enough. Also, most bitcoin libraries will have these values available anyways.

ATXMJ:
> > Can you give it a sub-100 NIP number?
> 
> Changed to NIP-89

This conflicts with pre-existing PR https://github.com/nostr-protocol/nips/pull/262

ATXMJ (changes requested):
There is a [pre-existing PR for a NIP-89](https://github.com/nostr-protocol/nips/pull/262).

Please change this one.

89.md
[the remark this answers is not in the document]
weex: https://github.com/nostr-protocol/nips/pull/476#discussion_r1191278173

melvincarvalho:
Gentle bump on this.  Did this ever move forward.  I'm at the point where I could now build a whole wallet on top of nostr, nip-07 and a nostr best send like nip89 would be the final step.

benthecarman:
> Gentle bump on this. Did this ever move forward. I'm at the point where I could now build a whole wallet on top of nostr, nip-07 and a nostr based send like nip89 would be the final step.

looks like nip 89 was taken, think that was all that was really needed

stutxo:
I am using this NIP in my project here https://github.com/stutxo/bostr. Would be cool to get this merged with a new nip number

PR #480, opened 2023-04-28
Add imeta tag to NIP 94

staab:
This is a different way to solve https://github.com/nostr-protocol/nips/pull/478

### DIFF
--- a/94.md
+++ b/94.md
@@ -4,9 +4,11 @@ NIP-94
 File Metadata
 -------------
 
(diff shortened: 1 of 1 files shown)


vitorpamplona:
I don't mind either way. I just want to standardize it.

vitorpamplona:
Can we keep it open, remove the 2nd paragraph and just merge the new `dim` tag? I could use that.

staab:
For sure, force push didn't update the PR, so I made a new one: https://github.com/nostr-protocol/nips/pull/481

PR #481, opened 2023-04-28
Add dim tag to NIP 94

staab:


v0l (commented):
I wont be able to create this tag from web easily, but i think that file hosts should create the NIP94 anyway if they are nostr native and they can fill this.

PR #483, opened 2023-04-28
add nostr groups draft

DanConwayDev:


bu5hm4nn:
I think voting is one special kind of group management. There should be owned groups as well.

In one of his interviews @rabble says the group encryption scheme he likes best out of his database of social protocols is the one employed by P2Panda. Here is their spec: https://p2panda.org/specification/encryption

DanConwayDev:
> I think voting is one special kind of group management. There should be owned groups as well.

Interesting point @bu5hm4nn. Voting prevents a single compromised key from removing the other administrators and  taking over the group. 

Are there instances where removing the inconvenience of having to prompt other admins to remove an admin outweighs the compromised key threat?

Regular nostr user admins may be prompted by one of their clients to vote on the proposal should this be widely adopted.

By 'owned groups' you mean groups with a single admin account? This can be achieved by adding an admin group with a single member. 
 

> In one of his interviews @rabble says the group encryption scheme he likes best out of his database of social protocols is the one employed by P2Panda. Here is their spec: https://p2panda.org/specification/encryption

Thanks for sharing that link. I didn't even consider encrypted content or private group membership as it is so far out of the wheelhouse of the GitHub replacement use case. I think a model for safer encryption through nostr would need to come first before something like that could be attempted.

DanConwayDev:
Something I didn't make explicit in the text is that public keys can appear multiple times to increase their vote share.

Perhaps a 'simple majority' should be replaced by 'at least 50%' and a 'super majority' replaced by 'at least 66%'. This would enable a better workflow for 2 or 3 administrators which I expect would be common place.

A single user could have an offline key with 2 votes and their standard key with 1 vote. This would enable them to regain control of their groups if their standard key got compromised.

staab:
I think this NIP is coming at the problem from the wrong direction. The real challenge is not assigning group admins, it is determining an architecture that supports group privacy. The approach that emerges to solve that problem will dictate the level of trust required (and available) to solve these more peripheral problems.

Access-limited groups aren't possible without either relay support or magic cryptography. The latter seems to be an open question for academic researchers, so for now groups have to involve relays in some way or another. Luckily, that's the whole point of nostr — to distribute trust across multiple relays.

Once a set of trusted relays is selected, events published to a group would only go to those relays, and include a `g` tag (or something) that specifies the group they belong to. Those relays would then implement access control, only relaying events after a client passes an AUTH challenge. This is a fair amount of additional complexity, but is also a real step up in functionality.

Once that framework is decided, moderation, member management, admin voting, trusted relay selection editing, etc. become much more concrete problems.

alexgleason:
That magic cryptography solution is... megolm??

I've been thinking the same things @staab, and same conclusions.

DanConwayDev:
This NIP should really be called public groups as 'groups' is misleading. Public and private groups require two very different models.

I agree that a solid implementation of private groups would be extremely beneficial and would have wider application than public groups. I haven't been following the developments in private messaging closely enough to express a considered opinion. 

@staab, what you are suggesting is to significantly increase the level of trust placed in relays so they also act as control systems.  Intuitively that feels like it goes against the trust-minimised philosophy of nostr. I'll ruminate on this... 

> Access-limited groups aren't possible without either relay support or magic cryptography.

I agree that this is the case for read access but not for write access. 

There are quite a few usecases that require validation that messages were issued by members of a group, who's membership is public.

This draft NIP achieves this without increasing trust in relays or require them to add support for the NIP.

This would also work in private relay context, where the group membership and message content are only accessible to those that could access the relay. Originally I was thinking this would be behind an enterprise firewall, but I suppose using auth messages to achieve this is exactly the model you are proposing. It would only be a matter of time before a relay was hacked and the messages revealed without the 'magic cryptography' that you speak of. If we had that, auth messages would just protect metadata (no bad thing).

I'll update the NIP to call this public groups.

staab:
> @staab, what you are suggesting is to significantly increase the level of trust placed in relays so they also act as control systems. Intuitively that feels like it goes against the trust-minimised philosophy of nostr.

I think it's actually precisely consistent with the philosophy of nostr, which is not to minimize trust, but to distribute it across as many parties as possible. In nostr, you're not any less reliant on 3rd parties than you are in traditional systems, you're just reliant on more of them, reducing the chances of data loss, but actually increasing the chances that your trust will be betrayed by some relay or other. I only mention this here because it's a very interesting concept to me.

At the same time, you're right that in this scenario a relay's betrayal will have more impact, since the data is private.

DanConwayDev:
> I only mention this here because it's a very interesting concept to me.

In the spirit of exploring interesting concepts , let me put to you the contrary position :)
Nostr isn't spreading trust to many parties, but minimising the level of trust required for any given entity. If a relay betrays our trust that it will, well, relay our messages, it doesn't matter because we have redundancy from the other relays we push to.

If we start trusting that a relay wont relay certain messages to certain parties or will only relay authorised messages, then we are increasing, rather than decreasing our risk exposure for every new relay we push to. 

> Doesn't this require consistent state to work?

Yes. It doesn't require all the messages related to group activity but it definitely requires a specific small subset of messages; those related to group membership / admin changes. This is likely to be less than 1%.

>  And not only see all messages, but also know they have seen all messages

It _only_ needs to see a message sent after one it hasn't received to realise it doesn't have the complete state. That's a very big _only_. 

The context I had in mind for this was maintainers groups on code repositories where their may be 1 or 2 changes a year.

> That is what goes against the philosophy of Nostr,

To be honest, I'm inclined to agree with you. You trusting that:

1. Group admins are pushing to all the specified relays, those messages are getting through and the relay isn't rejecting them.
2. At least one of these relays, which have been selected to reliably deliver all messages related to the group, are responding to requests for appropriately tagged events with the messages.
3. The messages don't get lost in transit. 

There is actually a prescient for a trust model like this on nostr in NIP-16 and NIP-33 replaceable events.

I was trying to achieve distributed consensus without worrying about honey badgers and byzantine generals. Perhaps we should just inscribe the full events onto the bitcoin blockchain  :laughing:

staab:
> If we start trusting that a relay wont relay certain messages to certain parties or will only relay authorised messages, then we are increasing, rather than decreasing our risk exposure for every new relay we push to.

Agreed, I hadn't thought of it that way.

> Doesn't this require consistent state to work?

This is also somewhat true of @earonesty's encrypted group chat proposal #580. My feeling is that instead of baking the key management into the proposal, we should define a way to use gift wrap generically as a way of creating an "encrypted nostr subnet". More thoughts on that [here](https://github.com/nostr-protocol/nips/pull/580#issuecomment-1576618688). The keys for these subnets could then be managed in one of many ways, either using the key rotation strategy specified in these nips, or using an access-controlled nsec bunker as described [here](https://habla.news/a/naddr1qqwkzttswfhhqmmnv9kz6en0wgkhqunfweshgefdvaex7atswvq3samnwvaz7tmjv4kxz7fwwdhx7un59eek7cmfv9kqzrthwden5te0dehhxtnvdakqzynhwden5te0wp6hyurvv4cxzeewv4esz9rhwden5te0wfjkccte9ejxzmt4wvhxjmcpzemhxue69uhhyetvv9ujumn0wd68ytnzv9hxgqgawaehxw309ahx7um5wgkhqatz9emk2mrvdaexgetj9ehx2aqpz3mhxue69uhhyetvv9ujumn0wd68ytnpdyq32amnwvaz7tmwdaehgu3wxpur2vpww3jkx6qprpmhxue69uhkummnw3ezu6m0d3kxjer9wgh8s7t6qyfhwumn8ghj7un9d3shjctzd3jjummjvuq3yamnwvaz7tm0venxx6rpd9hzuur4vgqswcm0wfskxmr9qgsf03c2gsmx5ef4c9zmxvlew04gdh7u94afnknp33qvv3c94kvwxgsrqsqqqa2884vczr).

earonesty:
all of nostr requires state.   if i follow someone and i can't find their kind 0, then... i've lost their info.   what addresses state is cross posting to "many" relays.   imo: we probably should work on discoverability protocols.   for example, if i want to know "all the kind 40's that exist in every relay", do i have to connect to all of them?   or can we have relays start talking to each other in a mesh net to lazily, slowly, but eventually fulfill queries more completely.

PR #484, opened 2023-04-29
NIP-30: Custom Emoji

alexgleason:
Adds a NIP for custom emoji, similar to Mastodon. It is [already implemented by the Mostr bridge](https://gitlab.com/soapbox-pub/mostr/-/merge_requests/56).

A real-world event can be found here: https://www.nostr.guru/e/5e1e0e524432e524f9584f296f72dda5559fc9fe6116a744baa43ea6769c7b65

### DIFF
--- a/30.md
+++ b/30.md
@@ -1,0 +1,56 @@
+NIP-30
+======
+
+Custom Emoji
+------------
+
+`draft` `optional` `author:alexgleason`
+
+Custom emoji may be added to **kind 0** and **kind 1** events by including one or more `"emoji"` tags, in the form:
+
+```
+["emoji", <shortcode>, <image-url>]
+```
+
+Where:
+
+- `<shortcode>` is a name given for the emoji, which MUST be comprised of only alphanumeric characters.
+- `<image-url>` is a URL to the corresponding image file of the emoji.
+
+For each emoji tag, clients should parse emoji shortcodes (aka "emojify") like `:shortcode:` in the event to display custom emoji.
+
+Clients may allow users to add custom emoji to an event by including `:shortcode:` identifier in the event, and adding the relevant `"emoji"` tags.
+
+### Kind 0 events
+
+In kind 0 events, the `name` and `about` fields should be emojified.
+
+```json
+{
+  "kind": 0,
+  "content": "{\"name\":\"Alex Gleason :soapbox:\"}",
+  "tags": [
+    ["emoji", "soapbox", "https://gleasonator.com/emoji/Gleasonator/soapbox.png"]
+  ],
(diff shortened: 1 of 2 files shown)


jb55:
LGTM

benthecarman (approved):

verbiricha (approved):

alexgleason:
I've also implemented this [in Ditto](https://gitlab.com/soapbox-pub/ditto/-/commit/837a3ae1dda3457535bd83d5d27934f92f393423) now, so it's interoperating between two Nostr projects. MR has 3 approvals. Any other feedback? Merge?

fiatjaf:
This should be a different event kind.

fiatjaf:
Can we change?

mikedilger:
I think a new event kind would just cause notes with custom emojis in them to drop entirely.

But IMHO every special thing inserted into a kind-1 note should IMHO ideally be inserted by a tag that specifies the offset to insert at, rather than infecting the content with a placeholder.  That includes ':emoji:', '#[1]', 'nostr:nevent1...', etc.  That way the content can still be sensibly rendered by clients that do not understand the tag, and there would be no need to parse the content.

fiatjaf:
Nevermind, I thought you would define your emoji pack once and reuse everywhere, but since you have to specify the emoji URL every time it's good as it is.

alexgleason:
I thought about that @mikedilger, but with NIP-27 there's nothing left that uses the `#[0]` scheme anymore. Also, a placeholder like `:nostrich:` is able to be typed by a human (with good autocomplete support) and displays nicely in unsupported clients.

fiatjaf:
> But IMHO every special thing inserted into a kind-1 note should IMHO ideally be inserted by a tag that specifies the offset to insert at, rather than infecting the content with a placeholder. That includes ':emoji:', '#[1]', 'nostr:nevent1...', etc. That way the content can still be sensibly rendered by clients that do not understand the tag, and there would be no need to parse the content.

What if we had both?

fiatjaf:
The emoji tag could be like `["emoji", "ablobcatrainbow", "https://gleasonator.com/emoji/blobcat/ablobcatrainbow.png", 7]`, 7 being the offset.

mikedilger:
> I thought about that @mikedilger, but with NIP-27 there's nothing left that uses the `#[0]` scheme anymore. Also, a placeholder like `:nostrich:` is able to be typed by a human (with good autocomplete support) and displays nicely in unsupported clients.

Yes when composing you could type :nostrich: and have that replaced by a tag by the client before sending.  Similar to @alexgleason or #3e470a etc.

Maybe we can have a new textnote kind that does this.  Current text notes will always need to be parsed as there will always be older software around.

staab:
Maybe we should have a convention of saying "last call" before merging something. Or maybe merging a PR just creates the urgency needed to actually have the discussion.

PR #486, opened 2023-04-30
RFC: Extending Nostr to support x.509 certificates

djschnei21:
Hi all - Very possible this has already been discussed, but had an idea last night that I'd appreciate some feedback on.  I was thinking about nip-26 and didn't _love_ how the only way for a subordinate key to become invalid was for a revocation message to be successfully propagated to all clients.  It seems to me that a configurable TTL would be ideal for subordinate key pairs, especially when these identities are tied to an organization or anything high profile where simply generating a new root key and starting fresh might not be an option.  adopting the x.509 standard would also unlock a LOT of existing PKI tooling which could unlock other additional possibilities.

### DIFF
--- a/95.md
+++ b/95.md
@@ -10,7 +10,15 @@ This NIP introduces an optional enhancement to the existing Nostr protocol by al
 
 ## Motivation
 
-While the Nostr protocol is designed to be completely decentralized, with anyone able to generate their own private keys, the inclusion of X.509 certificates can provide additional benefits. Most certificates would be self-signed X.509 certificates by design, but this feature opens up the possibility for the root key (or the original self-signed X.509 certificate) to create and sign additional subordinate key pairs with time-to-live (TTL) values. This allows users who follow the root's pubkey to validate the identity and validity of child/subordinate keys, offering an improvement over NIP-26.
+The Nostr protocol, as described in NIP-01, relies on public-private keypairs for user identification, event signing, and event validation. However, it does not currently support any mechanism for including additional pubkey metadata, such as issuer and validity time, within events. This limitation makes it challenging to manage key rotation, validity periods, and subordinate key delegation in a more structured and secure manner.
+
+The proposed NIP aims to enhance Nostr's security and flexibility by allowing clients to embed an optional X.509 certificate in addition to the existing pubkey field. Including an X.509 certificate would enable additional pubkey metadata to be embedded, such as issuer and validity time. Since Nostr and the issuance of private keys are entirely decentralized, most certificates would be "self-signed" X.509 certificates, which is by design.
+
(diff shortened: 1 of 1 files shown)


fiatjaf:
Why is it called x.509?

djschnei21:
hey @fiatjaf , just so I am clear, are you asking why "x.509" is included in the title of the nip?

djschnei21:
to poke some holes in my own idea, these are some critiques I can think of:

- Certificate size: X.509 certificates are typically larger than raw public keys due to the inclusion of metadata, such as issuer information and validity periods. This increased size could lead to higher storage and bandwidth requirements for Nostr clients and relays.

- Certificate revocation: The current proposal does not address how to handle certificate revocation or situations where a key has been compromised. Without a proper revocation mechanism, revoked or compromised certificates might still be considered valid by clients, potentially leading to security risks.  Could potentially incorporate something similar to nip-26

djschnei21:
misunderstood nip-26's delegation conditions! closing, not needed

fiatjaf:
I don't think it would be a bonus. My question was a depiction of my total ignorance about x.509 stuff even though I have used it before multiple times. It is probably a legacy thing impossible to modernize that we should reinvent instead of reuse.

mikedilger:
NACK

History Lesson (imprecise and possibly wrong):

x.509 is part of the x.500 directory specification, a directory system that pretty much nobody uses (it became a bloated nightmare so a lightweight alternative LDAP took over the industry). We are not an x.500 directory.

BTW, neither are web servers, but SSL choose to use x.509 (against the advice of some of the smartest commentators) for reasons similar to those expressed in this NIP.

x.509 digital certificates were designed to:
- Distribute/Find public keys
- Be hierarchical:  If two parties trust the same parent, that parent can help them learn each other's keys
- Use a well defined namespace: e.g. fir SSL it is DNS domain names
- Work offline: Certificates are just digitally signed documents indicating what somebody's key is (and this requires the well defined name space to work).  If they didn't need to be offline, you could just connect to the CA and ask, and there would be no need for a digitally signed document.

Then people wanted to verify that a certificate had not been revoked. To do that, you have to connect to the CA and ask (over OCSP), so the whole benefit of these Certificates being offline documents that could be served by the person they talk about... kinda fell apart as we have to connect to the CA anyways.  Now we are stuck with the legacy of them.

I am NACKing because:
* Nothing about nostr is hierarchical
* Nostr neither utilizes nor provides a well defined name space (it uses DNS but not necessarily, you could use IP addresses)
* X.509 has a massive amount of legacy cruft, much of which is now pointless (as I point out with the OSCP comment)
* I can't imagine any benefit to using x.509

djschnei21:
Thanks all, appreciated you taking a look. I probably jumped the gun and should have done some more research before submitting... Great points all around.

PR #491, opened 2023-05-02
NIP-0077: nostr-share deeplink/uri scheme for sharing images and text

mandelmonkey:
A NIP to allow app developers to share text and image with nostr clients via a uri scheme

Wasn't sure on what NIP number to assign so will let the editor do that


andrerfneves:
Big fan. Is IMG and MSG enough? What about video URLs? I bring up videos as they're super commonplace too. TikTok etc.

mandelmonkey:
Would make sense to see what other socials apps do,  video makes sense
But most social apps limit to text and 1 media i.e. img or video not both

Maybe it makes sense to re name img to media to cover everything

No technical reason why this nip can't support multiple media though

andrerfneves:
Agree here. But the open ended "media" URL would need to be rendered differently depending if it's a video or URL or whatever. So then one needs a "mediaType". And we go back to MIME types etc. I guess one could try and grasp from the .png or .mp4 of the file but that's assuming all clients will have that handled by the extension which is far from ideal. Almost think an open ended Media and MediaType properties alongside Text property would suffice and be more extensible? Clients can dictate which types of media THEY can handle/render?

fiatjaf:
I would say just `msg` is enough. Videos and image URLs go on the body of the message, as in any Nostr note. And I would add an optional relay hint property in case the website or game has a preferred set of relays they want to see the shared note in.

andrerfneves:
Id second the relays prop. Wouldn't "msg" alone lead to forever parsing that message in many ways? I guess that's how Nostr already does things...

mandelmonkey:
hmm I think given the non centralised nature of nostr clients compared to centralised single apps such as twitter or tiktok it makes sense for this to be more "general" in the approach, rather than having one ```msg``` and one ```img```  just having a ```msg``` and ```hint``` parameter where the ```msg``` parameter can contain a generic url encoded ```body```  for clients to parse as they wish or just a simple piece of text that is posted is probably best

@fiatjaf what is the format of a relay hint? is it just a bech32 encoded pubkey?

fiatjaf:
> @fiatjaf what is the format of a relay hint? is it just a bech32 encoded pubkey?

No, it's just the relay URL.

fiatjaf:
We need some clients implementing this. I think they don't even know about it. Seems to be an easy thing to implement.

mandelmonkey:
I have a sneaky plan to add a "share to nostr" button in my game which when pressed, if nothing opens it preloads an email to various nostr wallets support asking them to implement

arcbtc:
I want this bad, but I also want to be able for it to generate a DM.
Client scan qr and the client asks "Do you want to send <this message> to <some public key>"
In my application it would be for and offline nostr powered bitcoin ATM, and after the DM is sent the receiver client will zap the sender.
I imagine there are a tonne of other use cases as well. Even just a DM me QR people can connect to me with "Do you want to send "Pur Vida" to <some public key>".
Can you bend this NIP to facilitate that use case, and I will build the atm?

vitorpamplona:
Closing due to inactivity. The idea is needed but this doesn't seem to address all the requirements.

PR #493, opened 2023-05-03
NIP-97 Event Recommendations

vitorpamplona:
This NIP defines a simple protocol for NOSTR-native feed recommendation services as ephemeral requests between two Nostr pub keys: the user and a recommendation provider.

Client applications allow users to subscribe to such services, request recommendations at will and display them on the screen in any way they see fit. 

Let me know your thoughts.

### DIFF
--- a/97.md
+++ b/97.md
@@ -41,7 +41,7 @@ Clients can request recommendations at will using a kind `20020` ephemeral event
   "tags": [
     ["p", "<service pubkey>"],
     ["p", "91cf9..4e5ca", "wss://relay1.com"],
-	["p", "30223..14aeb", "wss://relay2.com"],
+    ["p", "30223..14aeb", "wss://relay2.com"],
   ],
   "content": "{ \"filters\": [<FILTER1>, <FILTER2>, ...]}",
   ...other fields
--- a/97.md
+++ b/97.md
@@ -18,14 +18,14 @@ For example:
 
 ```json
 {
-  "kind": 10020,
-  "tags": [
-    ["p", "91cf9..4e5ca", "wss://provider1.com/"],
-    ["p", "14aeb..8dad4", "wss://provider2.com/nostr"],
-    ["p", "612ae..e610f", "ws://provider3.com/ws"]
-  ],
-  "content": "<private list>",
-  ...other fields
+  "kind": 10020,
+  "tags": [
+    ["p", "91cf9..4e5ca", "wss://provider1.com/"],
+    ["p", "14aeb..8dad4", "wss://provider2.com/nostr"],
+    ["p", "612ae..e610f", "ws://provider3.com/ws"]
+  ],
+  "content": "<private list>",
+  ...other fields
 }
 ```
 
@@ -37,13 +37,13 @@ Clients can request recommendations at will using a kind `20020` ephemeral event
 
 ```json
 {
-  "kind": 20020,
-  "tags": [
-    ["p", "<service pubkey>"],
-    ["p", "91cf9..4e5ca", "wss://relay1.com"],
-  ],
-  "content": "{ \"filters\": [<FILTER1>, <FILTER2>, ...]}",
-  ...other fields
+  "kind": 20020,
+  "tags": [
+    ["p", "<service pubkey>"],
+    ["p", "91cf9..4e5ca", "wss://relay1.com"],
+  ],
+  "content": "{ \"filters\": [<FILTER1>, <FILTER2>, ...]}",
+  ...other fields
 }
 ```
 
@@ -59,17 +59,17 @@ Providers respond with an ephemeral event with kind `20021`. This event MUST con
 
 ```json
 {
-  "kind": 20021,
-  "tags": [
-    ["p", "<Request Pub Key>"]
-	["e", "<Request Event ID>"]
-    ["e", "91cf9..4e5ca", "wss://alicerelay.com/", <score>],
-    ["a", "30223:14aeb..8dad4:myblogpost", "wss://bobrelay.com/nostr", <score>],
-    ["p", "612ae..e610f", "ws://carolrelay.com/ws", <score>],
-	["r", "http://nostr.com", "", <score>],
-	["t", "bitcoin", "", <score>],
-  ],
-  ...other fields
+  "kind": 20021,
+  "tags": [
+    ["p", "<Request Pub Key>"]
+    ["e", "<Request Event ID>"]
+    ["e", "91cf9..4e5ca", "wss://alicerelay.com/", <score>],
+    ["a", "30223:14aeb..8dad4:myblogpost", "wss://bobrelay.com/nostr", <score>],
+    ["p", "612ae..e610f", "ws://carolrelay.com/ws", <score>],
+    ["r", "http://nostr.com", "", <score>],
+    ["t", "bitcoin", "", <score>],
+  ],
+  ...other fields
 }
 ```
 
--- a/97.md
+++ b/97.md
@@ -33,12 +33,13 @@ Providers can offer as many algorithms as they way as long as they offer one per
 
 # Recommendation Request
 
-Clients can request recommendations at will using the ephemeral event with kind `20020`. `p` tags in this event contain the group of users to make the recommendation for and the `.content` field contains an array of NOSTR `REQ` filters ([NIP-1 - Basic Protocol](1.md)) that describe the type and constraints of the requested recommendation events.
+Clients can request recommendations at will using a kind `20020` ephemeral event. `p` tags MUST contain the group of users to make the recommendation for as well as the pub key of the service provider. The `.content` field MUST contains an array of NOSTR `REQ` filters ([NIP-1 - Basic Protocol](1.md)) that describe the type and constraints of the requested recommendation events.
 
 ```json
 {
   "kind": 20020,
   "tags": [
+    ["p", "<service pubkey>"],
     ["p", "91cf9..4e5ca", "wss://relay1.com"],
   ],
   "content": "[<FILTER1>, <FILTER2>, ...]",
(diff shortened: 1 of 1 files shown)


97.md
```diff
@@ -0,0 +1,82 @@
...
+Clients can request recommendations at will using a kind `20020` ephemeral event. `p` tags MUST contain the group of users to make the recommendation for as well as the pub key of the service provider. The `.content` field MUST contains an array of NOSTR `REQ` filters ([NIP-1 - Basic Protocol](1.md)) that describe the type and constraints of the requested recommendation events.
+
+```json
+{
+  "kind": 20020,
+  "tags": [
+    ["p", "<service pubkey>"],
+    ["p", "91cf9..4e5ca", "wss://relay1.com"],
+  ],
+  "content": "[<FILTER1>, <FILTER2>, ...]",
```
brugeman, line 45: Should this be an object w/ 'filters' array to allow for further extensibility? I miss that in the REQ request, let's not re-create the issue here.

97.md
```diff
@@ -0,0 +1,82 @@
...
+  ...other fields
+}
+```
+
+Clients may specify several recommendation filters, i.e. `[{ "kinds": [1, 1065]}, { "#p": ["91cf9..4e5ca"], limit 200}, ...]`. Clients may include `kinds`, `tags` and other filter fields to restrict recommendation results to particular event kinds.
+
+Providers SHOULD interpret the recommendation request to the best of their ability and return events that it deems interesting to the requesting pubkey group. Recommendations can also include trending events as long as they are deemed of interest.
```
brugeman, line 52: It's good that you mention trending here, bcs 'recommendation' can be 'personalized' or 'global' (and probably many more types too). You're probably ok merging trending stuff from several providers, and personalized stuff from several providers, but mixing 'trending' with 'personalized' will be confusing. So a user either has to specify several different recommendation subscription lists (say one for trending, one for personalized), which means it should become 30k event, or we should be able to add some standardized labels of what kind of recommendation is being requested. I'd prefer the first option - let users evaluate recommendation algos and add them to different lists and let them switch btw those lists.

97.md
```diff
@@ -0,0 +1,82 @@
...
+```json
+{
+  "kind": 20021,
+  "tags": [
+    ["p", "<Request Pub Key>"]
+	["e", "<Request Event ID>"]
+    ["e", "91cf9..4e5ca", "wss://alicerelay.com/", <score>],
+    ["e", "14aeb..8dad4", "wss://bobrelay.com/nostr", <score>],
+    ["e", "612ae..e610f", "ws://carolrelay.com/ws", <score>]
+  ],
```
brugeman, line 69: Maybe we should allow any tags, not just 'e' as a recommendation? I might wish to recommend p profiles, r urls, t hashtags etc.

97.md
vitorpamplona: Agreed.

97.md
vitorpamplona: The only issue here is that there is no good way to use REQ filters to ask for only events or only URLs.

brugeman: Again, back to the question of 'type of algo' - how does a client know which filters to provide in the request? An algo might work let's say for music tracks only, how would client discover that? I'm guessing at least the appropriate kinds must be somehow discover-able by a client?

Or would it be like: this client only works with kind1, so it specifies kind1, and if results are empty - user will see empty results and will not subscribe to the list? 

Maybe there should be some other replaceable event announcing an algo? Like 'this profile is NIP-97 service and we support these kinds' type of thing, with a human-readable description? Not only could it solve the compatibility issue, but it could also allow clients to discover the algos and offer them to users for subscription?

vitorpamplona: I was hoping to keep service discoverability events out of this PR for now, but it might be a parameter in the User Metadata for automatic services offered (obviously a larger scope than just Recommendations).

vitorpamplona: The interface dictates the type of event. The client will just ask the subscription list for the types each part of the screen can display. If nothing comes back then it displays nothing (or a message asking the user to subscribe to a service)

brugeman: Agree on making discoverability a separate NIP, and the 'just display nothing if kind not supported' approach.

97.md
brugeman: So should the subscription list become 30xxx kind, so that users could have several of them?

97.md
```diff
@@ -0,0 +1,92 @@
...
+
+Providers SHOULD interpret the group of users as watching the feed together. If individual users within the group have safety filters in their profiles (e.g. parents and kids together), providers SHOULD apply the sum of safety filters to the response. 
+
+Providers SHOULD interpret the recommendation request to the best of their ability and return events that they deem interesting to the requesting group. Recommendations can also include trending events as long as they are deemed of interest.
+
+Clients MAY send requests to any provider or to just the select few the user is interested in at the moment.
+
+# Recommendation Response
+
+Providers respond with an ephemeral event with kind `20021`. This event MUST contain the tags to the recommended `e` event ids, `a` addresses, `p` profiles, `r` urls and/or `t` hashtags followed by a relay address to find them. An optional normalized score from 0 (least interesting) to 1 (most interesting) SHOULD be added as the 4th element of each tag. The `p` tag of the requester and the `e` tag for the requesting event SHOULD be added as well.
```
dyng, line 61: Must the Recommendation Response be a response to a previous Recommendation Request? I think it's useful that  when users have subscribed to a recommendation provider successfully, the provider could push recommendation event to users continuously (a push-style recommendation).

If the push-style is not always the case, we can have a configuration field in subscription event that allows users to explicitly specify what they need.

vitorpamplona: If we turn the request into a registration event, then I think it's better if these events become simple replaceable events (non-ephemeral). In that way, a client can register it's touch points and just download the recommendations using regular filters (as opposed to a request/reponse protocol)

The issue is that these registrations are client dependent. The request's filters is defined by the app (by the UI), not by the client. Apps can have n registrations for different parts of the app.

s3x-jay:
Just trying to get my head around this NIP. I don't have an opinion yet. Some basic questions…

I don't understand the utility of requesting recommendations for a group of people. Shouldn't recommendations be for an individual? Having the recommendation filter based on the most strict user seems like it's suppressing content. Why not just share recommendations and subscribe to other people's recommendations? (And don't share / subscribe if the recommendation parameters aren't appropriate.)

If you're talking about a curated list of posts… Why not just spin up an account and have a feed of reposts and reposts with comments? What does doing it the way you propose give you that that solution doesn't give you?

If this is about content discovery - then why not beef up Nostr's search capabilities and have a mechanism for user's to save their favorite queries - even grouping them so clients could create a feed of the combination of a set of queries? 

Again - I have no opinion. Rather, I'm just trying to understand _why_ this particular way of doing things…

vitorpamplona:
Thanks for the questions!

> I don't understand the utility of requesting recommendations for a group of people. Shouldn't recommendations be for an individual?

Imagine a Nostr video feed playing on a TV. You can get recommendations for everyone watching it AT THE SAME TIME. 

> If you're talking about a curated list of posts… Why not just spin up an account and have a feed of reposts and reposts with comments? What does doing it the way you propose give you that that solution doesn't give you?

Because it's individualized to the viewer instead of a one-size-fits-all feed. 

> If this is about content discovery

Search is good. But most content discoveries work via recommendations. It's hard to search for things user don't know exist yet.

arthurfranca:
> Providers are not required to be time-consistent. Every Recommendation Response can be a new list.

I think the filters from recommendation **request** event should use a `page` (default "0") attribute along with optional `page_limit` meaning number of recommendation tags (the recommended data) per page. Like this:
```js
{
  // ..., 
  "content": "{ \"page\": 3, \"page_limit\": 200, \"filters\": [<FILTER1>, <FILTER2>, ...]}"
}
```


The recommendation **response** should have the following:
```js
{
  // ..., 
  "tags": [
    // ..., 
    ["page_limit", "10"], // the actually used limit by the provider (maybe I requested 200 but the provider max page limit is 10)
    ["has_next_page", "false"] // (optional) true if has more pages
  ]
}
```

Considering `Providers are not required to be time-consistent` and that the sort order isn't chronological, `since` and `until` aren't useful for next page requests. With this in place, one would be really able to use this NIP as [algorithmic choice](https://blueskyweb.xyz/blog/3-30-2023-algorithmic-choice) for kind 1 feed, for instance. Instead of selecting which relay to read global feed from, the user selects from which provider pubkey.

vitorpamplona:
Nice idea @arthurfranca, what do you think @brugeman, does pagination make sense in recommendation systems? 

To be clear, this would require consistency between pages which creates a stateful session in the provider.

brugeman:
> Nice idea @arthurfranca, what do you think @brugeman, does pagination make sense in recommendation systems?
> 
> To be clear, this would require consistency between pages which creates a stateful session in the provider.

I think this NIP is already an algo feed. 

Implementing a stateful personalized algo feed is very non-trivial. Provider would have to maintain each user's feed potentially indefinitely, to allow for consistent experience btw sessions, btw different clients. It's like we're trying to reimplement what's really only possible with a centralized network. We can only scroll your algo-feed down until the end of time on Twitter bcs we have an account there, and each acc has a table with all new events written to it by the algo.  

Provider returns event ids, not full events. Providers could return say 500 ids (~50Kb), and then client does it's own pagination - downloads the events if user scrolls down. 500 entries is a huge number for a user to consume, they won't really want more 99.9% of the time. Google only returns a couple hundred results (try to paginate behind that), bcs it's too expensive and useless to do more than that. 

If some client has a database it can build the longer feed itself, by appending new stuff to the existing log.

I suggest we try to implement it as is and see if the experience is somehow lacking. I think the main issue will be latency, not lack of pagination.

arthurfranca:
> Providers could return say 500 ids (~50Kb), and then client does it's own pagination - downloads the events if user scrolls down.

You are right. This will work if it is highlighted in the NIP like `"Due to lack of pagination, it is advisable to return around 500 items at once."`

When I first read the NIP, I imagined the provider would return around 10 items without any pagination mechanism so that it would be enough to fill just a small recommendation panel, not a feed.

PR #495, opened 2023-05-04
NIP-90: Bitcoin PSBTs

ronaldstoner:
This NIP defines an event type for partially signed bitcoin transactions (PSBT) that can be transferred via the Nostr relay network. `1174` is the event kind representing a Bitcoin transaction in standard PSBT format, regardless if the transaction is partially signed or fully finalized.

We are submitting as draft and opening this up for comments from the greater community.

### DIFF
--- a/90.md
+++ b/90.md
@@ -1,0 +1,73 @@
+NIP-90
+======
+
+Partially Signed Bitcoin Transactions
+--------------------------------------------
+`draft` `optional` `author:ronaldstoner` `author:Armin Sabouri` `author:satsie` 
+
+# Abstract
+
+This NIP defines an event type for partially signed bitcoin transactions (PSBT) that can be transferred via the Nostr relay network. `1174` is the event kind representing a Bitcoin transaction in standard PSBT format, regardless if the transaction is partially signed or fully finalized.
+
+A special event with kind `1174` which contains an encoded PSBT. PSBTs used under this spec are version 0 [BIP-0174](https://github.com/bitcoin/bips/blob/master/bip-0174.mediawiki).
+
+PSBTs are encoded as a base64 string, and are contained in the `content` field. These events use a specific kind identifier so that clients do not need to render PSBTs the same as they do notes, invoices, and other event kind identifiers. If clients wish to display PSBTs they can choose to do so in whatever format they wish (string, QR Code, view render).
+
+The network magic is contained in the `tags` field. This is to identify the network that the transaction is for. The [common ones](https://en.bitcoin.it/wiki/Protocol_documentation#Message_structure) are:
+```
+- mainnet: f9beb4d9
+- testnet: 0b110907
+- signet: 0a03cf40
+- regtest: fabfb5da
+``` 
+
+PSBTs transmitted via the Nostr network provide a decentralized communication layer, encryption, and digital signature benefits with no username and password required for participants to coordinate and sign transactions using separate sovereign bitcoin private keys.
+
+# Protocol Flow
+The following are two protocol flow examples using the 1174 event kind. Other flows that could be used encompass various interactive/multi-party solutions such as coin joins, swaps or Schnorr bases multisig solutions (FROST/Musig2).
+
+## Multisig P2WSH
+2-of-2 Multisignature setup with Alice and Bob using a central coordinator
+
+1.  The coordinator has published a PSBT that spends from a multi-signature output controlled by Alice and Bob
+2.  Alice listens for event type 1174. Alice provides a signature for each input and publishes the updated PSBT referencing the original note using the ref tag.
+3.  Bob performs the same steps
+4.  Once the coordinator has received a quorum of signatures the transaction is finalized and broadcast 
+
+## Payjoin
+
+1. Sender publishes PSBT as specified by BIP-078.
+2. Receiver adds inputs and outputs to original PSBT and publishes the updated PSBT referencing the original note using the ref tag as specified in the “e” event reference tag detailed in the **tags** section of NIP-01.
+3. Sender verifies the final PSBT, and broadcasts the transaction
+
+## Event Format
+```json
+{
+	"id": <32-bytes lowercase hex-encoded sha256 of the serialized event data>,
+	"pubkey": <32-bytes lowercase hex-encoded public key>,
+	"created_at": <unix timestamp in seconds>,
+	"kind": 1174 <hardcoded integer>,
+	"tags": [
+		[“e”, <32-bytes lowercase hex-encoded sha256 of the previous serialized event data> ], // provided when replying
+		[“magic”, <4-byte hex value representing the Bitcoin network identifier>],
+		[“relays”, <list of relays to broadcast event>],
+	],
+	"content": <base 64 encoded PSBT>,
+	"sig": <64-bytes hex of the signature of the sha256 hash of the serialized event data, which is the same as the "id" field>
+}
+```
+
+## Explanation of fields:
+**id**: The sha256 hash of the serialized event payload.
+**pubkey**: The public key of the creator of the event.
+**created_at**: A Unix timestamp indicating the time when the event was created.
+**kind**: An integer value used to indicate the type of event. For PSBTs, the kind is set to 1174.
+**tags**: An array of tags to capture additional information about the multisig transaction:
+**“e”**: (optional): the 32 byte lowercase hex encoded sha256 event id of a prior event id. Only required when replying
+**“magic”**: the 4 byte hex representation of the network that the transaction belongs to. See <link to section above that lists magic bytes>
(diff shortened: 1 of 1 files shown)


thesimplekid:
Should these coordination type transactions be encrypted?  For example in the payjoin scenario having the metadata that two parties exchanged PSBTs to create the payjoin reduces the privacy benefit. [NIP-89](https://github.com/nostr-protocol/nips/pull/476) uses unencrypted PTSBs as the goal of that is to broadcast the transaction anyway, but to me these intermediate coordinating steps should leak as little metadata as possible.

ronaldstoner:
> Should these coordination type events be encrypted? For example in the payjoin scenario having the metadata that two parties exchanged PSBTs to create the payjoin reduces the privacy benefit. [NIP-89](https://github.com/nostr-protocol/nips/pull/476) uses unencrypted PTSBs as the goal of that is to broadcast the transaction anyway, but to me these intermediate coordinating steps should leak as little metadata as possible.

Agreed. If we can encrypt the content (the PSBT base64 string) such as NIP-04 DMs do then it provides a greater level of user privacy. I didn't know the best way to approach that though. Is it as easy as duplicating what NIP-04 is doing?

thesimplekid:
> Is it as easy as duplicating what NIP-04 is doing

NIP-04 is how NIP 46 and 47 encrypt their content.

90.md
```diff
@@ -0,0 +1,75 @@
...
+- mainnet: f9beb4d9
+- testnet: 0b110907
+- signet: 0a03cf40
+- regtest: fabfb5da
+``` 
+
+PSBTs transmitted via the Nostr network provide a decentralized communication layer, encryption, and digital signature benefits with no username and password required for participants to coordinate and sign transactions using separate sovereign bitcoin private keys.
+
+# Protocol Flow
+The following are two protocol flow examples using the 1174 event kind. Other flows that could be used encompass various interactive/multi-party solutions such as coin joins, swaps or Schnorr bases multisig solutions (FROST/Musig2).
```
satsie, line 27:
```suggestion
The following are two protocol flow examples using the 1174 event kind. Other flows that could be used encompass various interactive/multi-party solutions such as coin joins, swaps or Schnorr based multisig solutions (FROST/Musig2).
```

90.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Multisig P2WSH
+2-of-2 Multisignature setup with Alice and Bob using a central coordinator
+
+1.  The coordinator has published a PSBT that spends from a multi-signature output controlled by Alice and Bob
```
satsie, line 32:
```suggestion
1.  The coordinator publishes a PSBT that spends outputs controlled by Alice and Bob's multi-signature.
```

Suggestions: 
- Change step 1 to present tense so it matches the other steps. 
- Since step 2 notes Alice could be signing more than one input, update step 1 to reflect this possibility.

ronaldstoner: This makes sense. Good updates!

90.md
```diff
@@ -0,0 +1,75 @@
...
+A special event with kind `1174` which contains an encoded PSBT. PSBTs used under this spec are version 0 [BIP-0174](https://github.com/bitcoin/bips/blob/master/bip-0174.mediawiki).
+
+PSBTs are encoded as a base64 string, and are contained in the `content` field. These events use a specific kind identifier so that clients do not need to render PSBTs the same as they do notes, invoices, and other event kind identifiers. If clients wish to display PSBTs they can choose to do so in whatever format they wish (string, QR Code, view render).
+
+The network magic is contained in the `tags` field. This is to identify the network that the transaction is for. The [common ones](https://en.bitcoin.it/wiki/Protocol_documentation#Message_structure) are:
+```
+- mainnet: f9beb4d9
+- testnet: 0b110907
+- signet: 0a03cf40
+- regtest: fabfb5da
```
satsie, line 21:
```suggestion
- mainnet: f9beb4d9
- testnet3: 0b110907
- signet: 0a03cf40
- testnet/regtest: fabfb5da
```

Suggestion to bring list of magic bytes in alignment with this review comment from NIP-89: https://github.com/nostr-protocol/nips/pull/476#discussion_r1183356937

90.md
```diff
@@ -0,0 +1,75 @@
...
+
+Partially Signed Bitcoin Transactions
+--------------------------------------------
+`draft` `optional` `author:ronaldstoner` `author:Armin Sabouri` `author:satsie` 
+
+# Abstract
+
+This NIP defines an event type for partially signed bitcoin transactions (PSBT) that can be transferred via the Nostr relay network. `1174` is the event kind representing a Bitcoin transaction in standard PSBT format, regardless if the transaction is partially signed or fully finalized.
+
+A special event with kind `1174` which contains an encoded PSBT. PSBTs used under this spec are version 0 [BIP-0174](https://github.com/bitcoin/bips/blob/master/bip-0174.mediawiki).
```
benthecarman, line 12: PSBTs have versioning built in and in the spec it recommends to stop parsing if you dont support that version. Libraries should already have handling for this so I think it is fine to allow all kinds of PSBTs for the NIP

90.md
```diff
@@ -0,0 +1,75 @@
...
+tags: An array of tags to capture additional information about the multisig transaction:
+	“e”: (optional): the 32 byte lowercase hex encoded sha256 event id of a prior event id. Only required when replying.
+	“magic”: the 4 byte hex representation of the network that the transaction belongs to.
+	“relays”: (optional) tag provides relays to broadcast the transactions.
+content: PSBT encoded as base64.
+sig: The signature of the sha256 hash of the serialized event data, which is the same as the "id" field.
+```
+
+### Note:
+The "relays" field is optional and can be used to provide relay URL recommendations to clients for improved censorship resistance.
```
benthecarman, line 75: can mention the e tag is optional too.

90.md
```diff
@@ -0,0 +1,75 @@
...
+		[“e”, <32-bytes lowercase hex-encoded sha256 of the previous serialized event data> ], // provided when replying
+		[“magic”, <4-byte hex value representing the Bitcoin network identifier>],
+		[“relays”, <list of relays to broadcast event>],
+	],
+	"content": <base 64 encoded PSBT>,
+	"sig": <64-bytes hex of the signature of the sha256 hash of the serialized event data, which is the same as the "id" field>
+}
+```
+
+## Explanation of fields:
```
benthecarman, line 60: i dont see the point of this section, seems the same text is above already

benthecarman (commented):
I think this also needs to include p tags. Having to listen to all PSBTs to just find yours would be a lot of wasted resources. However, you can also make the argument if that you need that, then just do it over dms

90.md
ronaldstoner: Agreed. I'll remove it.

90.md
ronaldstoner: Including this in the next update

90.md
ronaldstoner: Thanks @benthecarman! I've added you as a co-author and incorporated all of your feedback.

ronaldstoner:
This has been updated to the newest revision after incorporating the comments and feedback.

benthecarman:
> Per @fiatjaf's reply in https://github.com/nostr-protocol/nips/pull/476#discussion_r1190066144 I have updated the `content` field to use hex as opposed to base64. 

PSBTs are typically encoded in base64, I think that should be used here

ronaldstoner:
> > Per @fiatjaf's reply in [#476 (comment)](https://github.com/nostr-protocol/nips/pull/476#discussion_r1190066144) I have updated the `content` field to use hex as opposed to base64.
> 
> PSBTs are typically encoded in base64, I think that should be used here

Good point - I'll revert. I talked to the team and they agree. I was trying to be a good nostr citizen and may have drifted too far from spec. I appreciate you @benthecarman !

90.md
```diff
@@ -0,0 +1,75 @@
...
+		[“magic”, <4-byte hex value representing the Bitcoin network identifier>],
+		[“relays”, <list of relays to broadcast event>],
+	],
+	"content": <base64 encoded PSBT>,
+	"sig": <64-bytes hex of the signature of the sha256 hash of the serialized event data, which is the same as the "id" field>
+}
+```
+
+### Additional Encryption
+The encryption used by DMs in NIP-04 can be utilized to further protect the base64 encoded data in the `content` field. This adds an extra layer of protection from those attempting to get information regarding transaction creation via nostr event data similar to how NIP-46 and NIP-47 implement NIP-04.
```
thesimplekid, line 61: If the content is only the PSBT does this need the params? I think it could be just like NIP-04 where the content is the encrypted PSBT and there is a `p` tag of the receiver. NIP46 and 47 have multiple message types so the JSONish structure is needed but here there is only one. Having the `p` tag would also allow clients to subscribe to events related to them.

PR #498, opened 2023-05-04
NIP-53 Live Activities

vitorpamplona:
This NIP defines a simple replaceable event for NOSTR-native advertisements of live activities (events, rooms, workshops, etc). The event is posted by the system's pub key and cites participants in a queriable fashion. 

Interested clients are expected to listen to `"kind": 30311` events in general or for given follow lists. 

Read [here](https://github.com/vitorpamplona/nips/blob/nip102-live-activities/53.md)


102.md
```diff
@@ -0,0 +1,53 @@
+NIP-102
+=======
+
+Live Activities
```
fernandolguevara, line 4: @vitorpamplona can we rename this to make it more abstract?

would be nice to have something like https://schema.org/Event

vitorpamplona: I just don't want to use Event because the basic building block of Nostr is an Event... So... But it should work for all events (arghhh)

jacany:
Shouldn't this kind of event eventually expire to prevent rogue clients from hijacking someone's profile with this? Ex client needs to rebroadcast every 3 minutes for it to be valid

vitorpamplona:
> Shouldn't this kind of event eventually expire to prevent rogue clients from hijacking someone's profile with this? Ex client needs to rebroadcast every 3 minutes for it to be valid

My original idea was to use trusted services: User adds the pub key of the system's service to a list of followers. For NostrNests for instance, if the user follows their account, the client will display current rooms other follows are participating on. This is separate from a usual feed (no need to sort by created_at, for instance)

jacany:
Got it, makes sense.

Second concern: I believe the title may be a bit broad considering the actual implementation. "Live Activities" could be something more specific like Audio Events as this seems to be very focused on a nostr-nests like application.

vitorpamplona:
> Second concern: I believe the title may be a bit broad considering the actual implementation. "Live Activities" could be something more specific like Audio Events as this seems to be very focused on a nostr-nests like application.

I was hoping we could use it for any live event (maybe live streaming stuff?). Why would it be related to Audio only? If a user Follows TED.com, for instance, they could see their follows showing up in the interface as video live stream.

staab:
I wonder if this could actually be broadened to form a sort of "calendar event" primitive, then add use-case-specific features (like streaming/participants) in on top of that.

102.md
```diff
@@ -0,0 +1,55 @@
...
+  "kind": 30311,
+  "tags": [
+    ["d", "<unique identifier>"],
+    ["title", "<name of the event>"],
+    ["summary", "<description>"],
+    ["image", "<preview image url>"],
+    ["t", "hashtag"]
+    ["streaming", "<url>"],
+    ["starts", "<unix timestamp in seconds>"],
+    ["ends", "<unix timestamp in seconds>"],
```
shawndearmond, line 30: A concern I have here is for events that might have multiple occurrences, or maybe occurs across multiple days. I think I would prefer something more like:
`["time", [["<start 1 unix timestamp in seconds>", "end 1 unix timestamp in seconds"],["<start 2 unix timestamp in seconds>", "end 2 unix timestamp in seconds"]],`

shawndearmond: Okay, well reading NIP-01, which says: "The tags array can store a tag identifier as the first element of each subarray, plus arbitrary information afterward (always as strings)."
So, we can't have arrays as tag values.

fiatjaf:
I think this NIP would be more popular if we made it possible to use it to link to more mundane live activities like a Twitch stream or a stream or things like that.

vitorpamplona:
That is inside what I was thinking when I proposed it. What's giving people pause that it is not for that? We should fix it right away.

fiatjaf:
What if this was just a set of extra tags in a https://github.com/nostr-protocol/nips/pull/597 calendar event?

vitorpamplona:
It could be. I was trying to keep them separate to make things simpler (and to filter by streaming events, not just calendar events), but there is no need to.

fiatjaf:
What if it was just a set of extra tags in a kind 1?

Clients implementing this NIP would display the extra streaming stuff, if any.
The content could contain a web link to the stream that non-augmented clients would just display.

fiatjaf:
Now I see it is a replaceable event. I think it makes sense to be a separate kind if you want to display "what is happening now" in a sidebar like Twitter does with Spaces. I guess I will shut up for now.

vitorpamplona:
Because we need to stop making everything about kind 1s. 

Also, the idea for this is exactly to avoid having a kind 1 post annoying the feed. Clients would figure out which of your follows are watching streams without requiring them to advertise to their feeds.

vitorpamplona:
Amethyst v60 has support for this kind now. It won't have a section dedicated to live events your follows are attending just yet but if you quote it in a kind 1, it will display the event.

fiatjaf:
Can you move this to a sub-100 NIP number?

(I'm still hoping we'll stop at NIP-99 and somehow transition to a better NIP scheme afterwards that I still don't know how will look.)

fiatjaf:
I think this NIP could also define another kind number to be used for in-app live chat in case someone is doing a native Nostr broadcast.

Someone could just host a livestream somewhere using HLS and the Nostr people could watch that from inside their clients and chat about the livestream. This could be done with kind:1 but would get spammy fast, so it's better to use another kind that is also pure text, with a reference to the original livestream event or address.

This also opens space for the streamers to easily integrate comments made with this new kind and zaps directly into their stream.

v0l:
Maybe we can include `kind: 1311` as regular event used to reply to live activity `kind: 30311`

vitorpamplona:
I think we are going to need some good moderation capabilities in Streaming chats.

vitorpamplona:
Also, @v0l, it feels like we could just reuse NIP-28 for streaming chats. Each streaming is a new chat channel with some metadata for the chat + mute functionalities. Do we really need to recreate that in 1311?

lylepratt:
> Also, @v0l, it feels like we could just reuse NIP-28 for streaming chats. Each streaming is a new chat channel with some metadata for the chat + mute functionalities. Do we really need to recreate that in 1311?

At Vida, we've been working on integrating our streams/rooms chat as NIP-28 public chat channels. We were thinking the actual stream location (HLS) could just be included as meta data in the NIP-28 declaration. Happy to implement it any way, just sharing this has been our thinking so far.

lylepratt:
@v0l is the chat here NIP-28 or something else? https://snort.social/live/naddr1qq8xgetddukkxe3dwd68yetpd5pzq9vhy34vytmazd6sgyz572jfs67ew8vdr9khn9ly39ejvwkfs70vqvzqqqrkvu0w7qmx

v0l:
I should probably read NIP-28 lol Snort doesn't support it at all (yet)

fiatjaf:
NIP-28 has a different purpose and different implementation details that are completely unnecessary here. There is no need for explicit moderation here. The messages on a live event are basically the same as kind:1 replies, you pick the relay from where to read and nothing more. In fact kind:1 would be very appropriate here, the only reason to pick a different kind is so these things do not spam the normal feed.

vitorpamplona:
> NIP-28 has a different purpose and different implementation requirements that are completely unnecessary here. 

How so? I don't think they are that different at all. And it would be better if we could re-use the moderation structures between public chats and streaming chats. Large streaming chats are a huge target for spam. If Nostr streaming gets any bigger, I would expect to have to deal with more spam than what we have seen in Public chats in Nostr.

fiatjaf:
What about reply spam in normal notes? How are we dealing with that? It's the same problem to me.

I don't think the person making a livestream should have control over what people are commenting. This is a decision that should be done by whoever is reading that chat of how to properly moderate (i.e. they can connect to safer relays or they can moderate themselves by blocking spammers or they can download a list of reported keys and moderate those automatically.)

vitorpamplona:
In a normal feed, you only see the people you follow. In streaming, you see everyone that comes into the website. They are very different. Imagine dick pics being posted in all streams. We can, of course, reduce to "Follows only". But that reduces too much. Even Follows of Follows in too small for streaming.

fiatjaf:
On a reply feed you don't see only people you follow. You see whoever has replied to that event according to the relays you query.

vitorpamplona:
Right, my on the reply feed of a given thread, we don't have 80K people watching it the replies being updated.

fiatjaf:
What do you mean there is no spam? The entirety of my Nostr experience is seeing threads with spam replies.

vitorpamplona:
I didn't say that the owner should be moderating. I said we need moderating tools. Otherwise, streaming will never take off in Nostr because as soon as it gets big, it gets filled with idiots trying to send a message to viewers.

fiatjaf:
You said NIP-28. I think that implies the owner moderates.

But if you want something like a livestream of moderation events that each client can receive and apply automatically to their channel feed by willing users then that is more reasonable, but that maybe could be made more generic so it could also help with reply spam, perhaps this could be mixed with NIP-56.

vitorpamplona:
I think you have not read NIP-28 in a while. This is no owner-based moderation predicted in NIP-28: https://github.com/nostr-protocol/nips/blob/master/28.md The only option is for chat users to moderate themselves.

vitorpamplona:
Added live chat features with @v0l 's kind 1311 and some examples.

arthurfranca:
The host is able to know current and total participant count by IPs requesting the media. However it can't know who are the participants because they may be using different clients. 

Consider using [Member List](https://github.com/nostr-protocol/nips/pull/607) for that. Each participant client is expected to update each user status. It is also useful to know if host and speaker are online or dropped connection.

arthurfranca:
I've added bespoke Member List config for Live Activities [here](https://github.com/arthurfranca/nips/blob/nip-35/35.md#live-activities). Please check if it is a good fit.

vitorpamplona:
So hard to say if Member List should be used or not. It can work well for small streamers, but It will be hard to download 10K events every 10 minutes just to keep the list of 10K people online.

arthurfranca:
Yes the idea is that each user client will dynamically build a custom member list, which should be small with possibility of fetching more if user wants to see more. Instead of adding many `["p", "612ae..e610f", "ws://provider3.com/ws", "Participant"]` tags to the live event, which the author probably won't be able to know who are these, the Member List NIP would be used. But just if the client UI wants to show online users (some UIs may show just the chat conversations).

First slots would be filled with host and speaker if present on the Live event. Then follows. Then the next would be retrieved using `limit` filter with a small value. So if you filled 10 slots (depends on UI), the client keep polling for just these 10 status updates. If 2 get offline, the client fetch 2 status events to replace the offline ones on UI.

arthurfranca:
Considering a limited mobile view, just knowing the Host and Speakers online status is alright.

Participants constantly updating status events would still be useful to know if my friends are participating on any live event, for discovery.

v0l:
We need to put something in the NIP about writing live comments to specific relays because ive seen different chats on different clients and probably this is not good for interaction because peoples comments might get missed and users will feel like they are being ignored.

Maybe we can add some host relays in a `relays` tag:

```js
"tags": [
  ["relays", "wss://one.com", "wss://two.com", ...]
]
```

This is where you should write live event comments to plus any other relays you want.

v0l:
Another suggestion is to have a `recording` tag.

This will be a way to link to a recorded video from a previous live stream.

Once the stream is `finished` state then we can expect to look for `recording` tag.

v0l:
Also a note about custom emoji NIP supported in Live Stream Comment kind @verbiricha Will have more detail

verbiricha:
> Also a note about custom emoji NIP supported in Live Stream Comment kind @verbiricha Will have more detail

[NIP-30](https://github.com/nostr-protocol/nips/blob/master/30.md) emoji tags are currently being used in kind `1311`. Since Amethyst already support NIP-30 it already works there. Probably worth updating NIP-30 by adding `1311` to supported kinds.

Also using the proposal from [Categorized Custom Emoji Lists](https://github.com/nostr-protocol/nips/pull/559) for autocompleting channel emojis.

102.md
michaelhall923: "Sessions" maybe.

vitorpamplona:
We should merge this and add new features in new PRs.

fiatjaf (approved):

shawndearmond:
If this is now named NIP-52, what about #597 ?
@tyiu is this equivalent to what you're proposing?

fiatjaf:
It is not equivalent, I just picked a random number for this. Let's rename it again.

vitorpamplona:
Updated to add relays and the recording tag. 

Let's merge this, please. We have other things to discuss once this is done (how to subscribe to events and how to verify hosts have agreed to host).

v0l:
Yea maybe we could include a type of streaming after because ive been looking to WebRTC streaming and it just looks like a regular URL without extension so it might be hard to identify without tagging it.

`["streaming", "http://webrc.com/play", "webrtc"]`

We could also tag `hls`.

What do you think about this.

fiatjaf:
What about pointing to a Twitch or https://vida.live/ stream and having that show up? Can I still do this on this NIP or has this idea been trashed and now we want people to do "native" streams only?

v0l:
> What about pointing to a Twitch or https://vida.live/ stream and having that show up? Can I still do this on this NIP or has this idea been trashed and now we want people to do "native" streams only?

vida.live streams work already its a regular HLS stream, there have been questions about youtube live and twitch embeds too, i dont see why not

vitorpamplona:
The proposition for the event kinds introduced here is to offer "things happening right now"-type of UI elements. We use it for video streaming most of the time nowadays, but it could be used for anything that is currently happening. The `planned` and `ended` statuses are just side effects of that UI need. 

That being said, we could mix this with a "Video/Audo Catalog" of some sort with a new `status` option. Catalogs seem to require different UI experiences for me, but I could be convinced otherwise.

> We are using HLS streaming in Stemstr. Not for live streaming, but to stream static audio files.

Amethyst plays HLS files in any post. It doesn't need to be a live event. The technical use of a "streaming" protocol is less relevant than the UI expectation of its contents. 

> youtube live and twitch embeds

Impossible to do in native apps without support for HTML, CSS and JS. Also, I don't feel good about encouraging the level of audience control/centralization of these players.

michaelhall923:
@vitorpamplona our specific use case is that we have kind 1 posts with embedded audio files as kind 1063's that can be either downloaded as .wavs or streamed via hls. we were considering just using a single kind 1063 for the wav file with an additional "streaming" tag for the streaming url. any thoughts on that approach?

vitorpamplona:
@fiatjaf you might want to see this addition of a proof for each `p` tag. I am not sure if this is used anywhere else in the protocol but could easily be extended to other event kids that require confirmation: https://github.com/nostr-protocol/nips/pull/498/commits/5477719989d211598ad16c8869a28075dec1bec2

fiatjaf:
Yes, looks like a good idea.

fiatjaf:
So, let's merge this already or not?

vitorpamplona:
Ready to merge.

v0l (approved):

michaelhall923:
> > we were considering just using a single kind 1063 for the wav file with an additional "streaming" tag for the streaming url. any thoughts on that approach?
> 
> This works for me.

Would be better to have a separate 1063 for the wav and the m3u8, since its techically two files?

PR #499, opened 2023-05-05
NIP-103:  Onion Routed Direct Messages

threeseries:
This NIP should help improve privacy with respect to kind 4 events on nostr.


v0l:
Same as https://github.com/nostr-protocol/nips/pull/468

Giszmo:
Ok, read your nip.

Can you re-word it all, avoiding the "another key pair" or "random"? These are things many don't want on nostr as to prevent spam.

I would call it "Onion Routed DMs" and use ephemeral messages where possible.

Allow for more than one hop.

Consider paying the routing bots.

Consider routing bot discoverability.

Routing bot advertisement:

* kind 14174 for bots to advertise their offering
* DMs could contain "nuts" for bots along the way. @callebtc any ideas here? bot advertisement could advertise a fee.
* Clients could filter out bots that were not recently updated
* Sybil attacks are an issue to be solved later. Bots could commit to staked BTC or whatever. In the beginning, known bots from the community could run the service.

DMs:

* Using nip-4's encryption is problematic as it does not allow the kind of onion routing you had in mind - the recipient needs to know the sender's pubkey to decrypt the message but as that is ideally not known to the intermediary on the last leg, you can't take that route. But as the recipient has to be ready to receive DMs of this kind, we could signal this readiness by adding an RSA public key to the bot and recipient profile. This key would ideally be deterministically derived from the users nsec.
* Send onions as ephemeral kind 20174 events.
* Send the core as non-ephemeral kind 174.
* As Carol receives the core from Bot, she needs to learn the real origin in the encrypted part, so the content could be an RSA encrypted nip-4 DM but I would avoid it being a valid event as you wouldn't want some clients to store them on relays in the decrypted form.

103.md
```diff
@@ -0,0 +1,25 @@
...
+
+# Motivation and usage
+
+Despite being encrypted direct messages on nostr have very poor privacy properties since anyone can see who is messaging whom and when. One solution to this problem is for the entire event including its metadata to be encrypted before being sent, and for the final recipient to be further obfuscated by adding additional hops between sender and receiver. In order to provide additional privacy for users RSA keys are used for encryption since these messages can be decrypted without knowledge of the encrypting user's nostr pubkey.
+
+The flow works as follows: when Bob wishes to send Alice an onion-routed DM he must first identify a set of intermediate pubkeys that can be used for routing and obtain their corresponding RSA public keys. Once done Bob creates a kind 4 event addressed to Alice using his nsec and then encrypts the whole event JSON using Alice's public RSA key. This becomes the content for the outer kind 174 event. The sender of this outer event is not Bob in general, but is rather the pubkey immediately before Alice in the chain. Events are then iterately wrapped in kind 174, working back up the chain until finally reaching Bob.
```
Giszmo, line 15:
```suggestion
The flow works as follows: when Bob wishes to send Alice an onion-routed DM he must first identify a set of intermediary pubkeys that can be used for routing and obtain their corresponding RSA public keys. Once done, Bob creates a kind 4 event addressed to Alice using his nsec and then encrypts the whole event JSON using Alice's public RSA key. This becomes the content for the outer kind 174 event. The sender of this outer event is not Bob in general, but is rather the pubkey immediately before Alice in the chain. Events are then iterately wrapped in kind 174, working back up the chain until finally reaching Bob.
```

103.md
```diff
@@ -0,0 +1,25 @@
...
+
+When Bob sends this kind 174 event to the first hop in the chain, the user or bot decrypts the content using their private RSA key. The decrypted content will be either kind 174 or kind 4, and the message is forwarded to the recipient pubkey. In order to provide additional privacy time delays can be added, or messages not forwarded until enough are in a queue.
+
+# Intermediate hops
+
+Intermediate nodes can be one of two types: always-online bots that exist solely to perform onion-routing, or ordinary users who have opted into forwarding messages for others (this also provides plausible deniability to the users themselves who are participating in forwarding). In the former case it may be desirable to use kind 20174 to make tracing more difficult, however there needs to be a way for bots to signal that they're online to ensure that such a message will be received. Hence it may be useful to have an ephemeral "heartbeat" event for sending these types of signals.
```
Giszmo, line 21: In my proposal I suggested to update a replaceable event regularly. A bot that updated an hour ago is probably online. A heartbeat would have to be every minute or the user firing up his client would have wait a long time before knowing an online bot is online.

You could also probe by sending a message to yourself.

103.md
```diff
@@ -0,0 +1,25 @@
...
+
+# RSA keys
+
+RSA keys should be derived deterministically from the user's nsec. They should also be advertised in the metadata of a pubkey for any account that can perform onion routing.
```
Giszmo, line 25: I'm afraid you have to be more prescriptive of how this should work. While profiles must expose the public RSA key, clients must know how to independently derive it from the Schnorr key. Some

```
entropy = sha256(npub+"RSA")
rsa = getRsaKeyPair(entropy)
```

but you'd have to research this yourself for some real-world code example.

Also you have to prescribe how and where to store the RSA pubkey. In the profile? In a new replaceable event?

Giszmo: Nah, the key derivation is not compatible in a sense that an RSA pub key wouldn't have anything to do with the Schnorr pubkey. The determinism is in my opinion only important, so a user doesn't have to backup yet another key. As the event is signed with the Schnorr key though, you can be certain that the RSA key was approved.

103.md
Giszmo: I'm arguing for lower relay load here. Updating the replaceable event is way more expensive than an ephemeral event but the latter you would need once per minute I would say and I consider the replaceable event to be preferable.

There is network, CPU and storage. 60 ephemeral events would be maybe x60 the network, x20 the CPU and x0 the storage. Not sure what will be the bottleneck.

103.md
Giszmo: Not a cryptographer. If RSA is done by brute-forcing a prime number in the right range, yes, this could work and should not make things slower than any other tool that would work with RSA. Now if we should store the RSA private key in a nip-4 DM style enrypted event I don't know. That would take away the requirement for it to be deterministic.

vitorpamplona:
This was superseded by GiftWraps.

PR #500, opened 2023-05-05
nip-31: dealing with custom unknown events

fiatjaf:
@pablof7z please take a look and fix this.

NIP text: https://github.com/nostr-protocol/nips/blob/nip31/31.md


staab:
What happens with existing events that have JSON in their `content`? I think this is a really good idea, but maybe we should specify an `alt` tag instead that clients SHOULD provide for all events. This would be useful for nip 94/95 image metadata, and might be a good place to put a text summary of long-form content.

pablof7z:
I like the `alt` tag idea a lot, actually, I think that's very clean.

vitorpamplona:
Are tags a good place for large amounts of data? I am thinking NIP-95 issues.

As a client, I would never render `.content` of unknown events by default. Even if there is guidance to make `.content` human-readable, the fact that anyone can simply write whatever they want in there is a UX hell. 

An "Event not supported" message is probably a better answer for this.

pablof7z:
> Are tags a good place for large amounts of data? I am thinking NIP-95 issues.
> 
> As a client, I would never render `.content` of unknown events by default. Even if there is guidance to make `.content` human-readable, the fact that anyone can simply write whatever they want in there is a UX hell.
> 
> An "Event not supported" message is probably a better answer for this.

An event that explicitly has an `alt` tag *can* be supported, though, as it explicitly informs `kind:1`-centric clients that they can safely render that value as an indication to the user of what that event is.

I wonder if an "Event not supported" message might feel like an error of some sort to users?

fiatjaf:
Why can't I write "your client sucks, get a better client" on a normal kind 1 note?

fiatjaf:
Makes sense. I think we need some more concrete examples to think better about this.

pablof7z:
this is `content` straight from a `31337` track on zapstr.live;

https://gateway.nostr.com/e/fd2ad0dcfe3bad28f88151f874b637f0009008b23cfc64dbed22760fcd8af13e

```
Seventh Hour Of The 24hr Frog Talk Ft. Iefan & Derek Rosss & Walker & Roya & Tatum

https://zapstr.live/t/naddr1qqtrvatwxe6xjen3wakx7amp8pex6eryv9uhy6czqqpsgqqq0f5s2ftxn4

https://nostr.build/p/nb5250.jpeg
```

I wrote those in the `content`, but for an `alt` tag, I would write it more specifically targeted as content for a client that doesn't know how to render this, something like:

```
This is an audio track called "Seventh Hour Of The 24hr Frog Talk Ft. Iefan & Derek Rosss & Walker & Roya & Tatum", you can listen to it on https://zapstr.live/t/naddr1qqtrvatwxe6xjen3wakx7amp8pex6eryv9uhy6czqqpsgqqq0f5s2ftxn4

https://nostr.build/p/nb5250.jpeg
```

If/Once we have something like note16trxqa2wwfvcgzy62q0gew247z3m9a60nnvya06g0fsy7h3remnsw2c9xd, we could leave the URL out and let clients query the user's follows for a handler for that kind.

mikestaub:
All clients should have a graceful fallback UX. "Unsupported Message Found" -> "View raw text" or "Search marketplace for plugin". VSCode nailed this with their extension ecosystem. Nostr could have a similar public registry of clients that list all the events they support.

pablof7z:
> All clients should have a graceful fallback UX. "Unsupported Message Found" -> "View raw text" or "Search marketplace for plugin". VSCode nailed this with their extension ecosystem. Nostr could have a similar public registry of clients that list all the events they support.

No user would click on "View raw text"; that feels too scary and implies that the experience will be probably broken. The first use of an `alt` tag would be providing context to the discussion that follows that unsupported, root event.

But that's up to clients to decide.

mikedilger:
I'm fine with things like "Event not supported" or what gossip says "NON FEED RELATED EVENT", geeky as that may be for normies.

I'm also fine with rendering whatever is in the "alt" tag should such a tag exist, along with a red ALT-TEXT marker (or something similar) indicating to the user that this is in fact alt text, and not the original content, and that we don't support this event directly.

So carry on.

staab (approved):

tyiu (approved):

PR #502, opened 2023-05-05
NIP-07 friendly description

elidyweaver:
When people use web clients for the first time, many are redirected to this page to learn about signing in with a browser extension. This page keeps a good list of options, but it is technical. Adding a "Nostr How To" here may make it easier for non-devs to understand NIP-07.


vitorpamplona:
Closing due to not being the place for these types of How Tos.

PR #503, opened 2023-05-06
Document multiple queries in a COUNT request

Giszmo:


staab:
I believe @fiatjaf preferred to remove the positional arguments. If someone wants multiple counts, they can send multiple COUNT requests.

Giszmo:
Uhm ... what? Then remove the multiple queries from the request maybe? Or define what the relay is supposed to do with those. As it stands, the client may send a thousand queries and the relay may aggregate the counts, ignore all but the first or whatever.

staab:
Are you referring to `["COUNT", <subscription_id>, <filters JSON>...]`? That works the same way as REQ, and returns a single count with the filters OR'd together. That could definitely be clarified though if you want to update the PR.

Giszmo:
Nah, I'm good

PR #506, opened 2023-05-07
NIP-01 Rephrase Markdown special rule

schmijos:
I was a bit confused about the Markdown remark and had to read up https://github.com/nostr-protocol/nips/commit/379252f992c2528e287bc4ce7faee5631aa3f73c

I guess @fiatjaf did not want to forward reference, so I tried to keep both statements:
* No markdown by example
* Mentioning other event kinds

I added the interpretation that clients should not need to infer content in IE6 manner.


fiatjaf:
I am more confused with that stuff about "inferring content". What does that mean? But if you say that is clearer I'm ok with this change;

schmijos:
Thank you for the feedback. I rephrased again.

fiatjaf:
I don't know about that either. I have the fear that people will glance over it without reading and assume you can and should do `[]()` stuff on kind 1, which I think should be avoided at all costs.

What if we said "Do not use Markdown."?

schmijos:
Since you say "at all costs", I added an exclamation mark.

PR #509, opened 2023-05-09
Reactions: require explicit like/dislike for emoji reactions

Aliervo:
This will allow clients to more easily decide how to display reactions. If a client receives a reaction event with an emoji it doesn't recognize, or doesn't want to display emoji at all, it can still be added to the up/down counts.

It also ensures consistency between clients. For instance, if a client receives a 🤣 reaction, is it a positive reaction because the post was meant to be funny, or a negative because the post is being ridiculed via emoji?

### DIFF
--- a/25.md
+++ b/25.md
@@ -18,8 +18,8 @@ downvote or dislike on a post. A client MAY also choose to tally likes against
 dislikes in a reddit-like system of upvotes and downvotes, or display them as
 separate tallies.
 
-The `content` MAY be an emoji, in this case it MAY be interpreted as a "like" or "dislike",
-or the client MAY display this emoji reaction on the post.
+The `content` MAY inlcude an emoji, in this case the client MAY display this
+emoji reaction on the post.
 
 Tags
 ----


fiatjaf:
Sounds good to me, but I don't know.

staab:
I don't think so, `include` just broadens the amount of potential context clients have to ignore. The ambiguity is real, but I don't think this is the way to solve it.

fiatjaf:
Yeah, nevermind, this breaks everything actually.

fiatjaf:
> I must be missing something here, because I was thinking the counting logic would just change from .content == "+" to .content[0] == "+".

This is a breaking change.

staab:
> Maybe adding a tag will be better, but it's unclear what client implementors will think of that.

In the past I've advocated for a numeric rating, -1 to 1, which could go in a tag. This would be more machine-readable, but of course the problem is how to get users to provide that, or infer it from the reaction, which is the same problem in a different place.

Aliervo:
> This is a breaking change.

I see your point, technically the latter works for both versions of the spec, but let's leave semantics for another day.

Looking at how snort.social currently handles reactions, it seems only one emoji is handled, 👎. It and "-" are treated as dislikes and everything else is a like. (see [Utils.ts line 141](https://github.com/v0l/snort/blob/88c4ea65eff57aca32dfd3de861c38679d238d19/packages/app/src/Util.ts#L141)) We can certainly figure out something better.

For my own reference, are breaking changes ever allowed in NIPs, or is the preferred method to write a new NIP and deprecate the old as with NIP-08 and NIP-27?

Aliervo:
>In the past I've advocated for a numeric rating, -1 to 1, which could go in a tag. This would be more machine-readable, but of course the problem is how to get users to provide that, or infer it from the reaction, which is the same problem in a different place.

That's an interesting thought. I suppose from there clients could implement their own visualizations of what a 1 or a -0.25 is. Emoji could be completely client side, 👎 for -.5, 🎉 for 1, completely up to the client. The event would be submitted simply as a number.

This could, however, lead to differences between clients which might confuse some users. It also would not allow use of arbitrary emoji, although I don't think any client currently supports that.

PR #512, opened 2023-05-09
NIP-88: binary event encoding

dr-orlovsky:
At the present moment nostr (NIP-1) uses text serialization of the event data, represented in a form of a JSON object. While being simple-to-implement and web-compatible, it has some significant drawbacks. This proposal tries to address them in a backward-compatible fashion, also providing encoding determinism and DoS attack mitigation.

### DIFF
--- a/88.md
+++ b/88.md
@@ -99,8 +99,8 @@ where:
 * `tlv` stream is skipped;
 * `tag` entries are provided in lowercase hexadecimal (base64) encoding.
 
-Binary protocol should operate using different network protocol and port number,
-which will be described in a dedicated NIP.
+The proposal is not compatible with the current nostr protocol and requires a
+binary protocol, which will be described in a dedicated NIP.
 
 
 [1]: https://nostrexplorer.com/e/9d89ef468ec690f2177635b8c0fd589aca840de1a749a08a44de5b148654bb49
--- a/88.md
+++ b/88.md
@@ -1,0 +1,107 @@
+NIP-88
+======
+
+Binary event encoding
+---------------------
+
+`draft` `optional` `author:dr-orlovsky`
+
+At the present moment nostr (NIP-1) uses text serialization of the event data,
+represented in a form of a JSON object. While being simple-to-implement and 
+web-compatible, it has some significant drawbacks:
+
+* Speed: serialization from JSON consumes most of the app time, according to [1]
+* Complex handling of binary data, which have to be encoded into Base64 or some
+  other binary-to-ascii format, increasing the problem described above and 
+  leading to more traffic consumption.
+* No explicit limits for the size of the data (i.e. it is not clear what is the
+  maximal size of the message, or the maximal number of tags), leading to DoS 
+  vulnerability and interoperability. While each relay is free to choose the
+  limits to prevent DoS attacks, the limits are not yet there, opening an attack
+  surface which may down the network for some time. Also, absence of the 
+  standard will lead that some of the relays will stop propagating some events 
+  (as a part of DoS prevention), while others will support all of them, making 
+  it hard to get the predictable and consistent user experience.
+* Protocol extensibility: with the current version of the protocol it is 
+  prohibited to add new fields to the JSON format.
+
+These all problems can be solved with introducing binary encoding to the 
+messages with strict limits to the maximal size of the elements and optional TLV
+extension stream.
+
+Specification
+-------------
+
+The proposed binary encoding must follow this spec (in pseudocode):
+```haskell
+Event ::
+    -- fields matching NIP-1:
+    Id, Pubkey, Timestamp, Kind, Tags, Content,
+    -- Type-length-value encoded stream:
+    TLVs,
+    -- NIP-1 signature
+    legacy Signature
+    -- new signature over binary data (except legacy sig)
+    binary Signature
+
+Id :: [Byte ^ 32]     -- indicates fixed-size byte string of exactly 32 bytes
+Pubkey :: [Byte ^ 32]
+Timestamp :: U32 -- indicates 32-bit unsigned integer
+Kind :: U64 -- indicates 64-bit unsigned integer
+
+Tags :: [Tag ^ 0..0xFFF] -- indicates array of tags with up to 4096 elements
+Tag :: TagName, TagData
+TagName :: [AsciiPrintable ^ 1..0xFF] -- indicates string from 1 to 255 chars
+                                      -- containing only printable ASCII chars
+                                      -- (bytes 0x20-0x7e)
+TagData :: [Entry ^ 0..0xFF]
+Entry :: [Byte ^ 1..0xFF]     -- inticates non-empty byte string up to 255 bytes
+
+Content :: [Byte ^ 0..0xFFFFFF] -- inticates byte string up to 2^24 bytes
+Signature :: [Byte ^ 64] -- indicates Shnorr signature
+
+-- Type-length-value encoded stream
+TLVs :: [TLVEntry ^ 0..0xFF]
+TLV :: type U64, value [Byte ^ 0..0xFFFF]
+```
+
+The pseudocode represents in-memory and network serialization layout of the 
+binary data, organized in form of fields listed after `::`. All integers are 
+serialized in little-endian format (to match the bitcoin protocol); hashes, 
+public keys and signatures match bitcoin consensus rules for serialization. 
+Variable-size collections (i.e. excluding id, pubkey and signatures) are 
+prefixed with either a single byte, a 16-bit or 24-bit word (in little-endian 
+encoding) matching the bit dimensionality for the maximum number of elements and 
+specifying the number of elements in the collection. The used notation follows
+[strict type] system notation for a memory binary data layout, which can't be
+deterministically specified if a protocol buffers or similar notation is used.
+
+The binary events are provided with two signatures: a legacy one, created over
+JSON-serialized message without TLV stream (see [Backward Compatibility](#backward-compatibility)
+section below); and a new signature, created over binary-serialized event data,
+including TLV stream but excluding the legacy signature.
+
+The maximal total length of the binary-serialized tags must not exceed 16MB 
+(0xFFFFFF bytes); thus the total length of the event can't be larger than 
+50 268 360 bytes (~50 MB).
+
+
+Backward compatibility
+----------------------
+
+Clients and relays not supporting the current proposal must ignore the events.
+
+Clients and relays supporting the proposal must provide a NIP-1-compatible 
+encoding of the binary data for the legacy (pre-NIP-88) clients and relay,
+where:
+* `content` is given in the Base64 encoding;
+* `signature` over binary data is skipped;
+* `tlv` stream is skipped;
+* `tag` entries are provided in lowercase hexadecimal (base64) encoding.
+
+Binary protocol should operate using different network protocol and port number,
+which will be described in a dedicated NIP.
+
+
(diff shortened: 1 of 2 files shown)


88.md
```diff
@@ -0,0 +1,107 @@
...
+
+Clients and relays supporting the proposal must provide a NIP-1-compatible 
+encoding of the binary data for the legacy (pre-NIP-88) clients and relay,
+where:
+* `content` is given in the Base64 encoding;
+* `signature` over binary data is skipped;
+* `tlv` stream is skipped;
+* `tag` entries are provided in lowercase hexadecimal (base64) encoding.
+
+Binary protocol should operate using different network protocol and port number,
```
dpc, line 102: Couldn't this binary protocol be delivered over existing web-socket, either on a different URL's path, or when certain http header is present?

88.md
```diff
@@ -0,0 +1,107 @@
...
+messages with strict limits to the maximal size of the elements and optional TLV
+extension stream.
+
+Specification
+-------------
+
+The proposed binary encoding must follow this spec (in pseudocode):
+```haskell
+Event ::
+    -- fields matching NIP-1:
```
dpc, line 38: Any reason not to start with a one-byte version?

88.md
benthecarman: It could be over the same websocket just sent as a bytes message instead of a text message

88.md
dr-orlovsky: TLVs. With just a version you either support it - or not - in full. With TLVs a gradual improvements are possible.

The only thing which version does (and TLVs can't) is signalling of the overall encoding standard. However, it can't do much even here: you can't migrate text/binary encodings with version byte (since it can't be present in the text encoding); it can't be use to migrate to protobufs or other common standard (since they prefix all fields with their own flags). So I do not see any use for the version byte here.

88.md
mutatrum: It would make sense if a relay/client decide on binary support, that they both start using the binary message protocol.

mutatrum:
How would this work for a relay? If a relay has a textmode client and a binary client, then the relay has to transcode every note from one to the other, and back.

I also have my doubts about JSON being the heaviest load on a client. You linked to a note from jb55, so I assume his remarks are about Damus. I looked into the code, and as far as I can see it's using the standard Swift JSON decoder.

Apparently, the standard JSON decoder used by Swift is not really made for performance. There are definitely ways to improve handling of JSON, especially if you expect to do all the heavy lifting in app. A high performance JSON library can come really close to binary encoding, but this might be language dependant if it's feasible or not.

dr-orlovsky:
> If a relay has a textmode client and a binary client, then the relay has to transcode every note from one to the other, and back.

Nope, the relay will just verify one signature, not both (i.e. binary for binary messages, text for JSON messages). If the other sig is invalid - it is not a problem, since it does not invalidates the event (a single valid sig already proofs authenticity of the event).

> There are definitely ways to improve handling of JSON, especially if you expect to do all the heavy lifting in app. 

JSON is well-known for code injection attacks, so using non-system decoder creates a tradeoff with security.
The other JSON problem is the absence of embedded limits to the max number of items, opening a huge surface for DoS attacks. I listed these reasons in the proposal introduction as a rationale for it

fiatjaf:
Although you can think of this as backwards-compatible it really isn't.

First because just doing a super optimized encoding this early in the protocol will immediately cause people to start prematurely optimizing JSON out of the their thing, split the protocol in two, and then probably making the JSON people second-class citizens and eventually JSON will not be supported anymore and the binary encoding will be a requirement.

Now, even if somehow people accept the weird situation of having both JSON and binary living together forever, that wouldn't work either, as you're introducing new fields, changing the event format and basically doing a new protocol, so I don't know what to say, maybe I misunderstood this part.

It is clear to me that Nostr wouldn't have had a tenth of the success it had if the original protocol had used a custom binary encoding, or even some standard efficient encoding method like protobufs or cap'n'proto. My opinion is that losing JSON compatibility would basically guarantee the failure of Nostr.

fiatjaf:
About the other points this NIP is trying to solve:

- speed: of course nothing will ever be as fast as the most optimized binary encoding you can come up with, but I believe there are other ways of speeding up JSON deserialization with custom parsers, if people really want that, specially since we have a relatively simple and static event format.
- binary data: Nostr is not meant to replace all other forms of internet communication, so there is barely any need for including raw binary data on Nostr events, for most cases just text should be enough. I think it would be nice to be able to include binary data sometimes -- for example, for encrypted message --, but life is made of tradeoffs, and in the cases where we're doing binary data it should be enough to do base64 as it shouldn't be a lot of data.
- limits for the size: I don't understand the low level details here, but my understanding is that relays are already limiting events by size, by number of tags and other criteria. A server or client should be able to know how many bytes it is reading from a socket, so JSON is not making things worse here.
- protocol extensibility: what is the problem of kind numbers and tags? Aren't they enough for extensibility? Adding new fields to events and changing the signature verification and all that is recipe for disaster, creation of incompatibilities and will only result in centralization or the death of the protocol.

dr-orlovsky:
> First because just doing a super optimized encoding this early in the protocol will immediately cause people to start prematurely optimizing JSON out of the their thing, split the protocol in two, and then probably making the JSON people second-class citizens and eventually JSON will not be supported anymore and the binary encoding will be a requirement.

I see that as an advantage and not a negative thing. I do not see why you see a value in keeping JSON and not abandoning it eventually. Even javascript can deal with binary data...

Just for the record, here are some of my comments you your points:

* binary data: file sharing and many other apps (which can run self-hosted cross-device synchronization) would benefit from the binary encoding;
* limits for the size: correct, these are the limits to the maximal (and sometimes minimal) number of items. And this is not about knowing "how many bytes to read from the socket" but for preventing DoS attacks

Anyway, I understand that you give a concept NACK here. You clearly have a different preferences on the tradeoffs between DoS protection, speed, scalability than me.

88.md
dpc: > since it can't be present in the text encoding)

Is the text encoding meant to be "the default" (or at least a standard option) forever, or only kept until ecosystem has a time to migrate?


> The only thing which version does (and TLVs can't) is signalling of the overall encoding standard.

I guess I see your point. Both T in TLV and possibly even Kind, could act as a "version" But then ... wouldn't you want NIP-1 fields to be enclosed in a TLV?

I'm thinking - what if eventually it seems desirable to change NIP-1 fields? Right now there's no way to obsolete them. If they lived in their own TLV, clients can start creating events with one TLV for NIP-1, and another with whatever new format.

dr-orlovsky: I think we can close the topic since this PR is was concept NACKed. But let me answer just of scientific interest.

Yes, I was thinking about making all fields TLVs - but this makes base implementation more complex without any win. Even if all the fields are TLVs, you can't eventually get rid of them, since there is no rule in TLV saying "an required field may become obsolete and not required".

dpc: Not each. All NIP-1 fields become one TLV with some "NIP-1" type. Then you eventually drop them all, as nothing is supposed to use them anymore.

fiatjaf:
I don't know what a "concept NACK" means. I just expressed my opinion. The NIPs do not reflect my preferences, if people start adopting this I will eventually have to merge it. But yes, I don't think this should be adopted in its current form.

dr-orlovsky:
"Concept NACK" comes from bitcoin PR review world and means what you have said: you do not agree with the proposal conceptually and do not acknowledge it.

BTW I think it is complete fine to shape the project basing on some vision and not allowing to not distracted from that. While I do not agree with your vision, I accept this approach.

I think at this stage it would be better to have a separate protocol which will use a different set of trade-offs for the use cases not covered by nostr - and if the data under this new protocol will be representable in nostr format, than it will be up to the market to chose the protocol it needs (or the one which will be more robust/scalable - or just have larger networking effect).

ursuscamp:
Why a brand new binary encoding, rather than a widely supported once that already exists and is designed to be JSON compatible, like msgpack?

dr-orlovsky:
@ursuscamp I am happy with any encoding which is deterministic and confined (i.e. controls the number of items in collection and have a memory layout which doesn't vary from message to message, like in bitcoin consensus due to `VarInt`-kind of things). Never heard about msgpack before you wrote.

dpc:
Msgpack is not deterministic. Though there's a concept of "Canonical Msgpack encoding", which was used in some crypto projects (from the top of my head, IIRC: Algorand).

One one hand I'm not sure if nostr needs deterministic encoding. On the other hand: self-describing formats like msgpack are most useful to exchange free-form, arbitrarily shaped data, which seem like needless overhead in a harder to change, more specified up front application.

ursuscamp:
@dr-orlovsky Pretty much what @dpc said. msgpack is more Nostr-y, because it's basically binary JSON (using key-value pairs instead of guaranteed layout). Another option would be protocol buffers, perhaps. Lots of support across most major and minor languages there, and I believe it has guaranteed layouts.

hsjoberg:
The most attractive thing about nostr is in my opinion its simplicity.
It has managed to build a strong grassroots movement of developers and enthusiasts and I strongly believe one of the biggest reasons for this is because it's so easy to understand and to program for.

Adding another encoding introduces unnecessary burden for the whole ecosystem and potentially destroys the simplicity as it could drive the JSON encoding out, leaving only the binary format left.
Also, I'm not sure it's even fair to call it nostr, because it's really its own thing. It's replacing one of the most fundamental parts of the protocol, the messaging structure and format.

Furthermore, If we look at the Internet today, there are many protocols that are text based; HTML, HTTP, SMTP, IRC and more.
They have all managed to scale much larger than nostr is today.
Nostr is just 2 years old. We don't need to tear it apart. If JSON performance is a bottleneck, we can focus on fixing that bottleneck, instead of throwing away everything we have already.

I strongly advise the nostr community to not adopt NIP-88.
Concept NACK.
But... I'm instead wondering if it's possible to look at binary use-cases and try to fix that either out-of-band or some semi out-of-band thing, rather than replacing whole of nostr.

dr-orlovsky:
Related comment: https://github.com/nostr-protocol/nips/pull/515#issuecomment-1543772833

88.md
```diff
@@ -0,0 +1,126 @@
...
+    -- NIP-1 signature
+    legacy Signature
+    -- new signature over binary data (except legacy sig)
+    binary Signature
+
+Id :: [Byte ^ 32]     -- indicates fixed-size byte string of exactly 32 bytes
+Pubkey :: [Byte ^ 32]
+Timestamp :: U32 -- indicates 32-bit unsigned integer
+Kind :: U64 -- indicates 64-bit unsigned integer
+OtsHash :: [Byte ^ 32]
```
RCasatta, line 53: OpenTimestamps use the document id as the hash for the proof, what's inside the ots field defined in nip3 is the base64 of the proof (basically the .ots file) and is variable length.
for more info see https://github.com/RCasatta/nostr-ots

Sammypit: Thank you for getting back to me I really appreciate this solution
Programming scientists etc reverse engineering is very creative and it's always a wormhole to get out of an in-out when it comes to ethical hacking lol having good day nice to hear from you

dr-orlovsky:
Closing this in favor of https://github.com/renostr/nrps effort - the NIP88 has moved into NRP-1.

88.md
dr-orlovsky: @RCasatta good point. 

NB: Since the concept doesn't fit the current #nostr approach (which I appreciate and which is completely fine) I moved this experimental proposal into a dedicated initiative to https://github.com/renostr

I will add the amendment you suggested there.

PR #515, opened 2023-05-11
NIP-93: NSON

fiatjaf:
I am unsure if this is a really dumb idea and I'm missing some obvious issue with it, but anyway.

text: https://github.com/nostr-protocol/nips/blob/nip93-nson/93.md

---

**update (Set 2024)**: I still think this is a valuable idea worth exploring, but I think specific encoding I used here could be improved.

### DIFF
--- a/93.md
+++ b/93.md
@@ -1,0 +1,184 @@
+NIP-93
+======
+
+NSON
+----
+
+`draft` `optional` `author:fiatjaf`
+
+### Preamble
+
+Some [benchmarks](https://github.com/fiatjaf/nostr-json-benchmarks/tree/2f254fff91b3ad063ef9726bb4a3d25316cf12d8) made using all libraries available on Golang show that JSON decoding is very slow. And even when people do assembly-level optimizations things only improve up to a point (e.g. for decoding a Nostr event, the decoding time is 50% smaller).
+
+Meanwhile, doing a simple TLV encoding reduces the decoding time to 35% and a simpler static binary format for Nostr events reduces makes that number drop to 4%. However, it would be bad for Nostr if a binary encoding was introduced, as it would be likely to cause compatibility issues, centralize the protocol and/or increase the work for everybody, more about this at [this comment](https://github.com/nostr-protocol/nips/pull/512#issuecomment-1542368664).
+
+### The actual NIP
+
+NSON is a crazy idea that, according to the benchmarks above, reduces the decoding time to 14% of that of the standard library. It works by having the JSON sender encode the event _as JSON_, but in a specific, very strict, order of fields (taking advantage of the fact that Nostr events have static fields, static lengths for some fields, and an overall rigid structure) and include in the JSON object a new field called `"nson"` that contains metadata the JSON receiver can read to help in the decoding process.
+
+Here's an example of a NSON-encoded Nostr event:
+
+`{"id":"57ff66490a6a2af3992accc26ae95f3f60c6e5f84ed0ddf6f59c534d3920d3d2","pubkey":"79be667ef9dcbbac55a06295ce870b07029bfcdb2dce28d959f2815b16f81798","sig":"504d142aed7fa7e0f6dab5bcd7eed63963b0277a8e11bbcb03b94531beb4b95a12f1438668b02746bd5362161bc782068e6b71494060975414e793f9e19f57ea","created_at":1683762317,"nson":"2801000b0203000100400005040001004000000014","kind":1,"content":"hello world","tags":[["e","b6de44a9dd47d1c000f795ea0453046914f44ba7d5e369608b04867a575ea83e","reply"],["p","c26f7b252cea77a5b94f42b1a4771021be07d4df766407e47738605f7e3ab774","","wss://relay.damus.io"]]}`
+
+The idea is that `"id"` comes first, so it can be accessed by reading a slice of the string from character `7` to character `71`, `pubkey` from character `83` to `147` and so on. `"content"`, `"kind"` and `"tags"` have dynamic sizes, so these are given by the values inside the `"nson"` field (which is also dynamic, its size by its first byte).
+
+### Anatomy of the `"nson"` field
+
+It is hex-encoded. Some fields are a single byte, others are two bytes (4 characters).
+
+Each explanation starts at the same line as the field it is referring to.
+
+                    number of tags (let's say it's two)
+                      number of items on the first tag (let's say it's three)
+                        number of chars on the first item
+                            number of chars on the second item
+                                number of chars on the third item
+                                    number of items on the second tag (let's say it's two)
+                                      number of chars on the first item
+                                          number of chars on the second item
+    "nson":"xxkkccccttnn111122223333nn11112222"
+            nson size
+              kind chars
+                content chars
+
+### Reference implementation
+
+Beware, all Rust maniacs, the following reference implementation is written in Go:
+
+```go
+func decodeNson(data string) *Event {
+	evt := &Event{}
+
+	// static fields
+	evt.ID = data[7 : 7+64]
+	evt.PubKey = data[83 : 83+64]
+	evt.Sig = data[156 : 156+128]
+	ts, _ := strconv.ParseInt(data[299:299+10], 10, 64)
+	evt.CreatedAt = Timestamp(ts)
+
+	// nson values
+	nsonSizeBytes, _ := hex.DecodeString(data[318 : 318+2])
+	nsonSize := int(nsonSizeBytes[0])
+	nsonDescriptors, _ := hex.DecodeString(data[320 : 320+nsonSize])
+
+	// dynamic fields
+	// kind
+	kindChars := int(nsonDescriptors[0])
+	kindStart := 320 + nsonSize + 9 // len(`","kind":`)
+	evt.Kind, _ = strconv.Atoi(data[kindStart : kindStart+kindChars])
+
+	// content
+	contentChars := int(binary.BigEndian.Uint16(nsonDescriptors[1:3]))
+	contentStart := kindStart + kindChars + 12 // len(`,"content":"`)
+	evt.Content, _ = strconv.Unquote(`"` + data[contentStart:contentStart+contentChars] + `"`)
+
+	// tags
+	nTags := int(nsonDescriptors[3])
+	evt.Tags = make(Tags, nTags)
+	tagsStart := contentStart + contentChars + 9 // len(`","tags":`)
+
+	nsonIndex := 3
+	tagsIndex := tagsStart
+	for t := 0; t < nTags; t++ {
+		nsonIndex++
+		tagsIndex += 1 // len(`[`) or len(`,`)
+		nItems := int(nsonDescriptors[nsonIndex])
+		tag := make(Tag, nItems)
+		for n := 0; n < nItems; n++ {
+			nsonIndex++
+			itemStart := tagsIndex + 2 // len(`["`) or len(`,"`)
+			itemChars := int(binary.BigEndian.Uint16(nsonDescriptors[nsonIndex:]))
+			nsonIndex++
+			tag[n], _ = strconv.Unquote(`"` + data[itemStart:itemStart+itemChars] + `"`)
+			tagsIndex = itemStart + itemChars + 1 // len(`"`)
+		}
+		tagsIndex += 1 // len(`]`)
+		evt.Tags[t] = tag
+	}
+
+	return evt
+}
+
+func encodeNson(evt *Event) string {
+	// start building the nson descriptors (without the first byte that represents the nson size)
+	nsonBuf := make([]byte, 256)
+
+	// build the tags
+	nTags := len(evt.Tags)
+	nsonBuf[3] = uint8(nTags)
+	nsonIndex := 3 // start here
+
+	tagBuilder := strings.Builder{}
+	tagBuilder.Grow(1000) // a guess
+	tagBuilder.WriteString(`[`)
+	for t, tag := range evt.Tags {
+		nItems := len(tag)
+		nsonIndex++
+		nsonBuf[nsonIndex] = uint8(nItems)
+
+		tagBuilder.WriteString(`[`)
+		for i, item := range tag {
+			v := strconv.Quote(item)
+			nsonIndex++
+			binary.BigEndian.PutUint16(nsonBuf[nsonIndex:], uint16(len(v)-2))
+			nsonIndex++
+			tagBuilder.WriteString(v)
+			if nItems > i+1 {
+				tagBuilder.WriteString(`,`)
+			}
+		}
+		tagBuilder.WriteString(`]`)
+		if nTags > t+1 {
+			tagBuilder.WriteString(`,`)
+		}
+	}
+	tagBuilder.WriteString(`]}`)
+	nsonBuf = nsonBuf[0 : nsonIndex+1]
+
+	kind := strconv.Itoa(evt.Kind)
+	kindChars := len(kind)
+	nsonBuf[0] = uint8(kindChars)
+
+	content := strconv.Quote(evt.Content)
+	contentChars := len(content) - 2
+	binary.BigEndian.PutUint16(nsonBuf[1:3], uint16(contentChars))
+
+	// actually build the json
+	base := strings.Builder{}
+	base.Grow(320 + // everything up to "nson":
+		2 + len(nsonBuf)*2 + // nson
+		9 + kindChars + // kind and its label
+		12 + contentChars + // content and its label
+		9 + tagBuilder.Len() + // tags and its label
+		2, // the end
+	)
+	base.WriteString(`{"id":"` + evt.ID + `","pubkey":"` + evt.PubKey + `","sig":"` + evt.Sig + `","created_at":` + strconv.FormatInt(int64(evt.CreatedAt), 10) + `,"nson":"`)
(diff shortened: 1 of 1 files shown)


93.md
```diff
@@ -0,0 +1,184 @@
...
+
+Also, other event fields that may be present (for example, the NIP-03 `"ots"` field) can be added at the end, after `"tags"`, with no loss.
+
+### Other points worth mentioning
+
+- Relays can receive non-nsonified events from clients, then reformat and store them nsonified so they can serve future clients better by sending them NSON always.
+
+### Open questions to be edited out of the NIP
+
+- How to signal NSON support? I thought it would work to have an initial field `"n":1` (before `"id"`) on the JSON, which could be read very fast, but I don't know.
```
v0l, line 184: Can it simply be signaled in the relay info document once its listed as supporting this nip?

fiatjaf: I think that's too restrictive, introduces the need for clients to query these relays and keep track of these settings in the code. Also I was thinking relays should be able to mix both nson and non-nson events (for whatever reason).

Another possibility is signaling on the websocket handshake, but that is also kinda annoying for everybody.

fiatjaf: Actually we don't need an initial field at all, the readers can just check if the static position `x...y` has the word `"nson"` in it.

dr-orlovsky:
This introduces quite a lot of attack vectors. I think using binary encoding is still simpler solution: by trying to avoid introduction of "complexity" of binary encoding (which is actually simpler than JSON) you actually add complexity and increase the attack surface.

Just switch to CBOR or anything else out there with TLV support to remain flexible in message structure and plan depreciation of JSON encoding. I do not see how keeping JSON makes protocol simpler to use from a developer perspective. JSON is actually _harder_ to use from my perspective.

fiatjaf:
I did not ever claim binary encoding was more complex, in fact this own NIP mentions a binary encoding that is referred to as "simpler", but it turns out that most people are not comfortable with binary stuff and having to learn these things and write a decoder before starting to develop a Nostr app is a barrier probably too high. Also backwards-compatibility is very important, even though you don't care about it at all.

I'm interested in learning what attacks this introduces.

habibitcoin:
ACK. I think @dr-orlovsky has some very real concerns/points. JSON is easier to wrap the head around, but obviously makes great sacrifices on scalability

I think NSON is a nice marginal move in the right direction vs JSON, and can keep the binary encoding convo ongoing

ekzyis:
Concept ACK. This looks like a good approach in the right direction.

Did you just come up with this on your own?
I tried to read more about "NSON" but I haven't found anything. ChatGPT also doesn't know what NSON is, lol

> Prompt: Do you know what NSON is?
> Answer: I'm sorry, but I couldn't find any specific information about "NSON" in my training data up until September 2021. It's possible that NSON is a relatively new or specialized term that has emerged after my knowledge cutoff. Could you please provide more context or information about NSON so that I can better assist you?

> Just switch to CBOR or anything else out there with TLV support to remain flexible in message structure and plan depreciation of JSON encoding. I do not see how keeping JSON makes protocol simpler to use from a developer perspective. JSON is actually harder to use from my perspective.

I never heard of CBOR and I think many other people also didn't. JSON has become the de-facto standard for data exchange between languages, for configs, for APIs (`application/json`), ... This is just how it is and we have to deal with it.
Using something else than JSON necessarily means that there will be less developers working on nostr since it's always a hurdle to learn something new. One if not the most important part of the nostr spec was how easy it was to understand. I agree with @fiatjaf there:

> It is clear to me that Nostr wouldn't have had a tenth of the success it had if the original protocol had used a custom binary encoding, or even some standard efficient encoding method like protobufs or cap'n'proto. My opinion is that losing JSON compatibility would basically guarantee the failure of Nostr.

-- https://github.com/nostr-protocol/nips/pull/512#issuecomment-1542368664

ekzyis:
> ACK. I think @dr-orlovsky has some very real concerns/points. JSON is easier to wrap the head around, but obviously makes great sacrifices on scalability

Maybe we just need ASICs to decode/encode JSON, lol (not sure if serious)

hoytech:
In order to keep back-compat, the nson field is not signed, right? This means that relays and other middle-men can alter it. Now we need to worry about how clients handle nson that is inconsistent with the normal event structure. Just off the top of my head: what happens if a decoder trusts the nson's number of tags, and the middle-man reduces the actual number of tags by 1: Then the attacker would've successfully "deleted" the final tag from the message.

What are we even trying to solve here though? Yes, JSON parsing is slow, but compared to verifying the signature it's insignificant. Since we're talking about the *external* format of nostr events, everyone is verifying signatures upon receipt, right? After you've validated it, you're free to store it in whatever efficient format you want for later access. strfry uses flatbuffers, and yes @Nuhvi is right -- take me out for a beer and I'll be happy to evangelise that format to you at length. :)

If we were going to create a new incompatible external format (which, to be clear, I *don't* think we should do, it would be a disaster for the ecosystem), I would focus on reducing the payload size. I believe the absolute minimal valid nostr event is 342 bytes (pre-compression). Especially since a lot of events are literally like 1 byte of payload (an emoji reaction or whatever), this is a ton of overhead. Most of the suggested binary formats don't even improve this very much, especially when factoring in compression. Things like msgpack and CBOR (which is basically just a rip-off of msgpack that a guy named C. Bormann named after himself) are especially pointless because they are schema-less, which means just pure overhead for a static format like this.

Here's what I would do in a brand-new format:

* Get rid of `id` field. It's redundant, and must be re-computed upon receipt in order to validate the signature anyway.
* Don't use schnorr signatures, but instead use a format where the pubkey can be recovered from the signature. This would let us get rid of `pubkey` field

These 2 changes would cut message overhead by 40% with no additional processing overhead (again, you *are* validating sigs on message receipt right?)

Note: I think schnorr signatures in theory do allow some cool batch verification stuff, which might actually be a big win when importing many events in a batch.

93.md
hoytech: This would have false positives in the case of a non-nson even where the content field or a tag spans `x...y` and happens to have `nson` in those positions.

fiatjaf:
>  and the middle-man reduces the number of tags in the nson by 1

This will be spotted once the receiver checks the id. The attacker can already modify the JSON directly, no need to modify the NSON field.

> If we were going to create a new incompatible external format

But the purpose of NSON is to be compatible!

> I don't think we should do

Me neither!

> Especially since a lot of events are literally like 1-4 bytes of payload

I was think shrinking the event size is useless as the `.content` dominates everything, but I always forget these reactions, which to me are useless and I think any client should never request or create these. But I guess it's fair to say that this kind of event is the actual overhead, not the event format.

> you are validating sigs on message receipt right?

I think it is not a bad approach for clients to assume relays are verifying, then sample event id and signature validation in the background in order to detect if there is some relay relaying fake messages eventually. If that happens the bad relay will be caught and expunged from society.

hoytech:
> I always forget these reactions

These are like 50% of all events now!

OK now I understand: You *are* optimising for the case where the client trusts its relays and isn't validating signatures, and the JSON decoding could plausibly be a bottleneck. Re-reading my message, I kind of came off a condescending jerk about that use-case, sorry about that hehe.

Actually in this case there's no point in even checking the id either, and obviously the fact that the `nson` field is unauthenticated doesn't matter. I think the NSON offset "hints" are a pretty clever idea, although it's yet more event-size overhead for clients that don't use them (which will be most of them, especially at first).

If JSON decode is a bottleneck, TBH I would push the libraries harder. Can you add [rapidjson](https://rapidjson.org/) to your benchmarks too? I see there are some [https://github.com/amosmzhang/rapidjson](go bindings), but they will have some overhead too of course. There is a *lot* of room for optimising this. Previously I have worked on a code-base that parsed JSON (mostly flat with static fields, similar to nostr) using a custom [ragel](http://www.colm.net/open-source/ragel/)-based parser that was faster than rapidjson still.

93.md
mikedilger: yeah I think you should start with the signalling field of a fixed length to not screw the offsets. Then the variable length nson field after all the fixed length fields somewhere. I can't imagine how it could work otherwise.

93.md
```diff
@@ -0,0 +1,180 @@
...
+
+### The actual NIP
+
+NSON is a crazy idea that, according to the benchmarks above, reduces the decoding time to 14% of that of the standard library. It works by having the JSON sender encode the event _as JSON_, but in a specific, very strict, order of fields (taking advantage of the fact that Nostr events have static fields, static lengths for some fields, and an overall rigid structure) and include in the JSON object a new field called `"nson"` that contains metadata the JSON receiver can read to help in the decoding process.
+
+Here's an example of a NSON-encoded Nostr event:
+
+`{"id":"57ff66490a6a2af3992accc26ae95f3f60c6e5f84ed0ddf6f59c534d3920d3d2","pubkey":"79be667ef9dcbbac55a06295ce870b07029bfcdb2dce28d959f2815b16f81798","sig":"504d142aed7fa7e0f6dab5bcd7eed63963b0277a8e11bbcb03b94531beb4b95a12f1438668b02746bd5362161bc782068e6b71494060975414e793f9e19f57ea","created_at":1683762317,"nson":"2801000b0203000100400005040001004000000014","kind":1,"content":"hello world","tags":[["e","b6de44a9dd47d1c000f795ea0453046914f44ba7d5e369608b04867a575ea83e","reply"],["p","c26f7b252cea77a5b94f42b1a4771021be07d4df766407e47738605f7e3ab774","","wss://relay.damus.io"]]}`
+
```
mikedilger, line 22: In the 'nson' here it starts with 28.  But the event has 2 tags.

mikedilger:
ACK in principle.

IMHO JSON is actually harder than binary in non-javascript languages, but we could squabble about that all day so I'm not here to debate it. But getting people and lots and lots of installed code base to all switch to binary is fucking impossible. So no switch to binary. But if we were going to do that, please involve me I have written a lot about what nostr should change if it could start over (in cold storage). And of course include @hoytech. The real world is not where you get to create your idealistic perfect notions of how things should be, it is a messy place where you have to navigate the best imperfect path.

Giszmo:
Concept ACK. Can we agree to merge this only after two relay implementations picked up on it? What does @jb55 say?

hoytech:
@fiatjaf - I decided to verify your assumption that web clients check the `id`. If they aren't validating signatures, maybe they don't bother to do this step either. I modified my relay to return events with an incorrect `id` and tried to load this relay up in various clients:

* snort: Displays the event
* iris: Does not display the event
* coracle: Inconclusive: can't add custom relays?

So at least snort is not bothering to validate the `id`. In the case of a client that trusts the relay and therefore does not validate signatures, this is actually fine security-wise (the relay could easily recompute the `id` anyway) and desirable from an efficiency perspective.

However, consider the following case:

* Honest relay that doesn't know anything about the `nson` field, but simply relays it unchanged to users
* Client like snort that trusts the relay and therefore does not validate the `id`, but is upgraded to use the `nson` field for parsing

A malicious user could take a legit event from a popular user, add or modify the `nson` field and upload it to the honest relay, which would accept it since the signature is valid. The client would download this event and use the unauthenticated `nson` to parse the data. The event could then be parsed and displayed incorrectly. For example, the attacker could truncate the beginning/ending of the content, remove tags, etc.

There are several possible mitigations:

* Require even non-signature validating clients to validate the `id`, at least when they intend to use NSON (which means they now have to re-encode and hash the event, reducing whatever efficiency benefit they got from NSON).
* Relays can be required to parse and validate the NSON
* Alternatively, relays can be required to strip out unknown and therefore authenticated fields (strfry has always done this)

Any of the above mitigations could work, but in general I think using unauthenticated data at all is a bad idea. It's a variation of the [cryptographic doom principle](https://moxie.org/2011/12/13/the-cryptographic-doom-principle.html).

jb55:
> Concept ACK. Can we agree to merge this only after two relay implementations picked up on it? What does @jb55 say?

if anything it should just be a different wire format that is just TLVs with null-terminated strings (like flatbuffers but we don't need that level of complexity). I'm not sure I understand the point of putting nson in the json. The biggest slowness in json is dealing with escaping and copying. If it was binary zero-copy over the wire then it would be insanely trivial to decode, and this would be backwards compatible and optional since you could signal it in the websocket request.

fiatjaf:
> I'm not sure I understand the point of putting nson in the json.

It makes decoding the JSON insanely faster than using normal JSON libraries. Check the benchmarks. That includes unescaping strings and parsing hex.

jb55:
this was discussed 3 weeks ago on nostr as well:

note1rtra39wn6ke985uzsxj5y0xpyfgjtvc2e4g6jg4ghqgctmnj8m3sp3753q

fiatjaf:
In other news, this is how nson and my simple binary codec compare against `JSON.parse` in JavaScript:

2023-05-12-145627_700x542_scrot

Remember that in the Go benchmarks nson runs in 14% of the speed of the standard library, and the binary codec in 4%. In JavaScript it looks like the worse performance is on the binary codec by far.

mikedilger:
I put my vote behind this:

> Require even non-signature validating clients to validate the id, at least when they intend to use NSON (which means they now have to re-encode and hash the event, reducing whatever efficiency benefit they got from NSON).

That is, in the NSON NIP, it should specify that clients MUST validate the id.  This isn't a must if they don't use nson.

dr-orlovsky:
@jb55 I proposed a binary encoding some time ago with #512, but we came to conclusion that it is better to experiment with it in a dedicated protocol. That's why we have started http://github.com/renostr initiative. It will provide:
* binary encoding
* end-to-end encryption
* much faster sigs and SSH/GPG compatible keys with Ed25519
* combination of both RPC-style of communications with PubSub (over HTTP+WebSocket and TCP)

All of these will be backward-compatible, i.e. it is an extension protocol over nostr. Main goal - help mobiles reduce power consumption, increase speed, better & standardized protection from DoS attacks.

Announcement: https://damus.io/note1d2y8jyrkzu46neuk8806nghx8vdutqznxkmglvqytzh0s7xtrjyq5rlnxa

fiatjaf:
> why not just make this an optional negotiated wire format that is pure binary?

Because the speed gains of binary over NSON do not justify the added complexity of keeping two formats around forever and the (very high) risk that some people decide to implement just the binary encoding, thus splitting the network.

93.md
```diff
@@ -0,0 +1,180 @@
+NIP-93
+======
+
+NSON
```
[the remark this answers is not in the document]
fiatjaf: but it's specific to Nostr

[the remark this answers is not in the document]
Giszmo: NOSN?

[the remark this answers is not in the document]
fiatjaf: but that sounds bad

boreq:
>but it turns out that most people are not comfortable with binary stuff and having to learn these things and write a decoder before starting to develop a Nostr app is a barrier probably too high.

I think writing a custom JSON encoder is most likely a more complex task than concatenating bytes together. One of the problems with Secure Scuttlebutt that I perceived was serious was for example the need to reimplement the complex JSON-based event validation algorithm in every single implementation. This lead to people burning out before doing anything meaningful with the protocol and I know that what we are proposing is nowhere near that complex but concatenating bytes together still seems like an easier alternative.

fiatjaf:
@boreq I know, the only point here is that this is backwards-compatible so you can keep calling `JSON.parse()` in your code.

93.md
```diff
@@ -0,0 +1,180 @@
...
+
+Also, other event fields that may be present (for example, the NIP-03 `"ots"` field) can be added at the end, after `"tags"`, with no loss to anyone.
+
+### Other points worth mentioning
+
+- Relays can receive non-nsonified events from clients, then reformat and store them nsonified so they can serve future clients better by sending them NSON always.
+
+### Open questions to be edited out of the NIP
+
+- How to signal NSON support? I thought it would work to have an initial field `"n":1` (before `"id"`) on the JSON, which could be read very fast, but I don't know.
```
kehiy, line 180: Why not just check the presence of `"nson" field?

kehiy:
I checked the benchmarks and go-nostr implementation, I found out that the jsoniter is not involved in these benchmarks. Since easyjson was very close to NSON, maybe JSON iter is the same or faster? I think we have to add a benchmark for it.

kehiy:
Results:

```
goos: linux
goarch: amd64
pkg: github.com/nbd-wtf/go-nostr/nson
cpu: 13th Gen Intel(R) Core(TM) i7-13620H
BenchmarkNSONEncoding/easyjson.Marshal-16                   9391            109022 ns/op
BenchmarkNSONEncoding/jsoniter.Marshal-16                   7382            140099 ns/op
BenchmarkNSONEncoding/nson.Marshal-16                       3633            293041 ns/op
BenchmarkNSONDecoding/easyjson.Unmarshal-16                10000            104739 ns/op
BenchmarkNSONDecoding/jsoniter.Unmarshal-16                 3948            278600 ns/op
BenchmarkNSONDecoding/nson.Unmarshal-16                    15740             68405 ns/op
BenchmarkNSONDecoding/easyjson.Unmarshal+sig-16              112          10092620 ns/op
BenchmarkNSONDecoding/jsoniter.Unmarshal+sig-16              100          11253574 ns/op
BenchmarkNSONDecoding/nson.Unmarshal+sig-16                  100          10335135 ns/op
PASS
ok      github.com/nbd-wtf/go-nostr/nson        12.039s
```

kehiy:
@fiatjaf this nip example is removed from go-nostr and i think it won't be logical to use this. i think we can close this pr.

PR #517, opened 2023-05-11
NIP-99: Prediction markets

ekzyis:
Hi, I have been working on this for some days now and I thought I will just create a draft PR to get some early feedback.

I mentioned this idea on SN for the first time: https://stacker.news/items/173759

I also posted about this on nostr: https://snort.social/e/nevent1qqs8g0jrsd8uszpgsmtzt6gp4gw64re6avn5mjycg8hrg6drsq86vtcpz4mhxue69uhkummnw3ezummcw3ezuer9wchsygx3nlk9gqnap9cx2jpzszhp0f5dpt2h7r6u97n5uf8nmwa3umz5j5psgqqqqqqskp8r0v

Any feedback is appreciated! I have no experience with betting and prediction markets (or writing specs) so maybe someone with more experience can chime in and become a co-author.

I have been reading and learning a lot about prediction markets in the last week and I am starting to think that they would be a very good addition to nostr (and lightning) and could become a big thing.

However, I am not sure if I totally lost touch with reality now so any feedback is indeed very appreciated!

text: https://github.com/ekzyis/nips/blob/nip-prediction-markets/99.md

### DIFF
--- a/99.md
+++ b/99.md
@@ -1,0 +1,170 @@
+# NIP-99
+
+## Prediction Markets
+
+`draft` `optional` `author:ekzyis`
+
+This NIP specifies how prediction markets can be implemented using nostr and lightning.
+
+Any user ("the creator") can create a _predictable event_ (`kind 1099`) such that other users ("the participants") can buy and sell shares of possible outcomes using zap requests as defined by NIP-57 and reverse zap requests (TBD).
+Services ("the betting exchanges") the required events for buying and selling shares, set the prices per share (based on odds) and handle payouts.
+Participants can freely choose in which exchange(s) they want to participate, e.g. based on liquidity or reputation since participants must trust the exchanges with payouts and reporting the honest outcome of the event.
+Exchanges can charge commissions on winning bets to incentivize continuous honest behaviour.
+
+For a general overview on what prediction markets are, see Appendix A.
+
+### Creation of predictable events
+
+To create a predictable event, `kind 1099` events are used as defined by NIP-01 like this:
+
+```json
+{
+  "content": <description about what should be predicted>,
+  "tags": [
+    ["outcome": <description of possible outcome>],
+    ["outcome": <description of another possible outcome>],
+    ...
+  ],
+  "kind": 1099,
+  "id": <event_id>
+  "pubkey": <pubkey of creator>,
+  "created_at": <timestamp>,
+  "sig": <signature>
+}
+```
+
+Example for a binary option event:
+
+```json
+{
+  "content": "Will Twitter go bankrupt in 2023?",
+  "tags": [
+    ["outcome": "Yes"],
+    ["outcome": "No"]
+  ],
+  "kind": 1099,
+  ...
+}
+```
+
+Exchanges run clients which have subscribed to these events.
+Per received event, they create parametrized replaceable events as defined by NIP-33 per possible outcome:
+
+<!--
+Working on a better explanation. Exchanges don't have to provide services for every event.
+
+Also, I want to provide an example flow from start to finish.
+This example flow should especially answer the following questions regarding "market bootstrapping":
+  Q: How are initial prices set?
+  A: Without any participants, there is no information regarding odds so every outcome is equally probable. So every share should cost the same.
+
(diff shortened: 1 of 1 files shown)


99.md
```diff
@@ -0,0 +1,170 @@
...
+  Q: How are initial prices set?
+  A: Without any participants, there is no information regarding odds so every outcome is equally probable. So every share should cost the same.
+
+  Q: What happens if there will only be one participant? Do they not always lose money because of potential commission fees?
+  A: ???
+
+  Q: Why would the first participant buy more than a single share? If they buy multiple shares, the next participant has an advantage because the next participant can just buy a single share to profit the most with the least risk (assuming no other participant will follow).
+  Basically this: Why would you bet 100 sats to win 100 sats if you can just bet 1 sat to win 100 sats?
+  A: ??? Maybe I am just confused with shares vs. odds. I need to learn more about prediction markets first.
+-->
```
ekzyis, line 67: I have iterated a few times over this NIP already but I still have some important questions to answer.

I added them as HTML comments to the NIP so I don't forget them

99.md
```diff
@@ -0,0 +1,170 @@
...
+}
+```
+
+These events include `zap` tags to link zapping the event with buying a share of that specific outcome with the current given odds.
+
+<!-- how to sell shares using NIP57 ??? -->
+
+Outcomes are specified by using plus addresses and zero-based indexing the `outcome` tags in the referred predictable event. Participants must validate that the plus address local part references an outcome with the same description as in `content` to prevent ambiguity.
+
+<!-- plus addresses are not supported by LUD16 ... yet? -->
```
ekzyis, line 92: > The idea here is that a SERVICE can offer human-readable addresses for users or specific internal endpoints that use the format <username>@<domainname>, e.g. [satoshi@bitcoin.org](mailto:satoshi@bitcoin.org). A user can then type these on a WALLET. The <username> is limited to a-z0-9-_.. Please note that this is way more strict than common email addresses as it allows fewer symbols and only lowercase characters.

-- https://github.com/lnurl/luds/blob/luds/16.md

Can a `+` be added as a valid character to allow plus addressing? This would be backwards compatible.

If not, I need to find another way to refer to specific outcomes in an event within a LN address. I could just use `_` instead of `+`, I guess. But I liked how plus addressing nicely fits the use case here.

fiatjaf:
I will read this more carefully later, but it is looking pretty good.

99.md
```diff
@@ -0,0 +1,170 @@
...
+
+## Reference
+
+### Appendix A
+
+<!--
+I think this video is a really good explanation of what I am trying to achieve with this NIP.
+Is this "good style" to include links to videos instead of writing it out? -->
+
+[![](https://img.youtube.com/vi/xA27x7GRMZQ/0.jpg)](https://www.youtube.com/watch?v=xA27x7GRMZQ)
```
ekzyis, line 144: Add your comments regarding this to this thread

99.md
```diff
@@ -0,0 +1,170 @@
+# NIP-99
+
+## Prediction Markets
+
+`draft` `optional` `author:ekzyis`
+
+This NIP specifies how prediction markets can be implemented using nostr and lightning.
+
+Any user ("the creator") can create a _predictable event_ (`kind 1099`) such that other users ("the participants") can buy and sell shares of possible outcomes using zap requests as defined by NIP-57 and reverse zap requests (TBD).
```
ekzyis, line 9: With "reverse zap requests" I mean an event you create and the other party does zap.

Basically my idea to implement something like LNURL-withdraw on nostr

99.md
ekzyis: Oh, looking at https://github.com/ekzyis/nips/blob/nip-prediction-markets/99.md, this doesn't seem to render the image on Github so one can't click on anything. So maybe just add the link without any image?

ekzyis:
Going to close this since I don't think I'll be able to work on this in the near future anyway.

I've decided to experiment with prediction market stuff (UX, oracle design etc.) on https://delphi.market/ for now.

See this announcement: https://stacker.news/items/337637

PR #518, opened 2023-05-11
Add user experience recommendation about NIP-07

vivganes:
Invoking NIP-07 methods as a part of page-load event listener creates intermittent issues due to race conditions between the browser's extension script and the client's page scripts.

### DIFF
--- a/07.md
+++ b/07.md
@@ -22,10 +22,9 @@ async window.nostr.nip04.encrypt(pubkey, plaintext): string // returns ciphertex
 async window.nostr.nip04.decrypt(pubkey, ciphertext): string // takes ciphertext and iv as specified in nip-04
 ```
 
-### Recommendation to Implementers
-When `window.nostr` is injected by a browser extension, it may not be fully available until a certain point in the page's lifecycle.  Hence, it is a good idea to invoke the first `window.nostr` based functionality after a user-action on the web page (like a button click), and not during the page load.
+### Recommendation to Extension Authors
+To make sure that the `window.nostr` is available to nostr clients on page load, the authors who create Chromium and Firefox extensions should load their scripts by specifying `"run_at": "document_end"` in the extension's manifest.
 
-Calling the `window.nostr` based methods during the page load runs the risk of intermittent issues due to race conditions between the client page's script and the extension's script
 
 ### Implementation
 


vivganes:
Bumping this up, in case it missed the review.  Do you have any comments on this?

07.md
```diff
@@ -22,6 +22,11 @@ async window.nostr.nip04.encrypt(pubkey, plaintext): string // returns ciphertex
 async window.nostr.nip04.decrypt(pubkey, ciphertext): string // takes ciphertext and iv as specified in nip-04
 ```
 
+### User experience recommendation
+Due to the way browsers load the extensions, it may take some time for `window.nostr` object will be fully ready to accept requests.  Hence, it is a good idea to run the first `window.nostr` interaction on a user action on the web page (like a button click), and not during the page load.
```
alexgleason, line 26: I'm not sure this is true. You do want to wait for the rest of the page to load before you call it the first time, like with `window.addEventListener('load', ...)`. But Alby seems to work better if you access `window.nostr` (even just checking for its existence) at least once before any user action.

Also, NIP-07 is not exclusive to browser extensions. For example, a custom web browser could implement NIP-07 and it would be available before the page is fully loaded.

If we were to add this text, I would name the headline "Recommendations for implementers" and lead with "When `window.nostr` is injected by a browser extension, it may not be available until a certain point in the page's lifecycle. So it's a good idea to [...]" 

It would also be good for any Alby (or other extension) developers to chime in.

vivganes: @alexgleason  Thanks a lot for your inputs.  This text makes a lot of sense.  I have edited the wording as you suggested.

bumi: I actually would add a UX notice here. Imo this is not only a technical detail but imo it is a good practice to not automatically do window.nostr calls on page load but always at least for the first time do this on user interaction. 

there is this in the [webln guide](https://www.webln.guide/building-lightning-apps/best-practices):
```
Before running window.webln.enable() or other methods, make sure the user knows what your app does, and why they should allow your calls to run. Popping up a window as soon as they load a page will cause users to reject WebLN requests, or worse yet, bounce from your page. Explicit buttons (e.g. to login) are helpful, but alternatives could be hiding pages behind loaders while requesting the user's provider, or keeping it to sub-pages and having your homepage not require WebLN.
```

alexgleason:
This is related to: https://github.com/fiatjaf/nos2x/pull/62

erskingardner (approved):

alexgleason:
Now that we know the trick, and browser extensions have been updated, I'm not sure we need this change to the NIP anymore. It should "just work".

vivganes:
> Now that we know the trick, and browser extensions have been updated, I'm not sure we need this change to the NIP anymore. It should "just work".

Thanks for the suggestion.  I have updated the file by deleting the existing suggestion to client authors and ADDING a suggestion to the extension authors.

This way, any new extension authors will save themselves the hassle.

staab (approved):

07.md
```diff
@@ -22,6 +22,10 @@ async window.nostr.nip04.encrypt(pubkey, plaintext): string // returns ciphertex
 async window.nostr.nip04.decrypt(pubkey, ciphertext): string // takes ciphertext and iv as specified in nip-04
 ```
 
+### Recommendation to Extension Authors
+To make sure that the `window.nostr` is available to nostr clients on page load, the authors who create Chromium and Firefox extensions should load their scripts by specifying `"run_at": "document_end"` in the extension's manifest.
```
bumi, line 26: why actually not `document_start` ?

Issue #131, opened 2022-12-29
Embedding decentralized media


grantwest:
Relays can't be expected to host media, the infrastructure requirements would be immense. URLs break and many image/video hosts do annoying things that prevent simple unobtrusive viewing of media. Use of decentralized URN to embed media in messages should be normalized. 

Example URN:
```
urn:ipfs:bafybeigdyrzt5sfp7udm7hu76uh7y26nf3efuylqabf3oclgtqy55fbzdi
```

There is an [open issue](https://github.com/ipfs/in-web-browsers/issues/161) on the IPFS repo to register this URN with IANA, not that being registered really matters and nostr could adopt either way.

It would be up to the individual nostr implementations/clients to decide how to implement. They could embed an IPFS client or just issue http requests to public gateways. So this is not truly part of the nostr protocol, rather just a suggestion to clients.

fiatjaf:
We can definitely standardize this, but I think it would be nice to see it implemented and used with success first because I am not sure people will prefer this over https://gateway.cloudflare.com/ipfs links (or whatever is the best gateway these days).

grantwest:
A client could convert URNs to cloudflare gateway links on the fly. The signed nostr message should use the URN because it is permanent. How the client wants to access the referenced file and how it wants to present the link to the user is client preference.

fiatjaf:
I agree. What is the format of that URL?

grantwest:
```
urn:ipfs:<IPFS CID>
```
More info on IPFS CID: https://docs.ipfs.tech/concepts/content-addressing/#what-is-a-cid

Example:
```
urn:ipfs:bafybeigdyrzt5sfp7udm7hu76uh7y26nf3efuylqabf3oclgtqy55fbzdi
```
Which can be accessed using the cloudflare gateway for example with this link:
https://cloudflare-ipfs.com/ipfs/bafybeigdyrzt5sfp7udm7hu76uh7y26nf3efuylqabf3oclgtqy55fbzdi

burdiyan:
I think it would be better to standardize on URLs instead of URNs, as it's equally flexible and probably more portable. Also, IPFS community has [settled on](https://github.com/ipfs/kubo/issues/1678#issuecomment-492977659) `ipfs://<cid>` URL scheme for these portable linking capabilities. There was even some effort to support this in browsers, among which I think Brave does support this kinds of links by default.

grantwest:
@burdiyan Thanks for your input! I think that is a very good idea!

PR #521, opened 2023-05-12
NIP-54 - Inline Resource Metadata

arthurfranca:
It extracts the best from the [previous discussion](https://github.com/nostr-protocol/nips/pull/478).

One additional upside is that when the regular user right clicks an image and copies then pastes the url on a new note, it preserves all added metadata. Also, it can be appended to any url like this NIP-21 one: `nostr:nevent1qq...epm#a-metadata-name=a-metadata-value`


vitorpamplona:
Sorry, why are we moving the description to alt?

arthurfranca:
@vitorpamplona  It's because alt is basically for blind people, so it should be a correct description. While the other one (NIP-94 event content) could be a loose description, like a funny caption that not really describes the picture, just like [html image titles](https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Images_in_HTML#image_titles).

vitorpamplona:
We used the `.content` as the descriptor for the blind. I don't know why there would be a need for another descriptor.

arthurfranca:
> [...] unless you are trying to do two things in a single event type.

It is exactly that. I imagined a photo gallery of NIP-94 kind 1063 events.

On one of the photos, the `.content` description says "This day at Copacabana was awesome!". In the abscence of assistive text (alt), you could use `.content` as such. But alt tag text of the same photo could be "A sunset at the beach with a group of four people looking at it". This has the benefit of not requiring kind 1 to have a photo caption and not showing up on kind 1 feed.

Now if you want this on regular social feed, the 1063 event would be referenced on a kind 1 note.

vitorpamplona:
So, the stories in Amethyst are exactly that. But without any additional descriptors. 

I understand that it makes things easier to not have another kind, but I am not sure if it's the right approach. We have to keep in mind that NIP-94 and NIP-95 are generalized file references and not posts. I believe another type is required to host them correctly if your goal is to make another non-kind1 feed. 

Otherwise, you will start seeing 1063s that were included in replies in a Kind 1 on your image feed, disconnected from the Kind 1 context. Or files that were added to a GitHub on Nostr project. Of your feed is based on Kind 1063 alone (like ours is) you will see a bunch of images and posts out of context. 

And keep in mind, 1063 is for any file. Not only for images or videos.

arthurfranca:
> Otherwise, you will start seeing 1063s that were included in replies in a Kind 1 on your image feed

Yes, but just put a `t` tag like photo_gallery or something and load 1063 only with that.

arthurfranca:
I consider caption part of metadata but I'm ok reverting the NIP-94 change. I will just wait if someone has anything to add.

fiatjaf:
It doesn't make much sense to have this NIP given that the Damus way already exists and is being used in practice. We should all use that. Yes, that's not the best way to write a spec ("just follow whatever Damus is doing") but that is probably the best way to make an existing protocol interoperable.

arthurfranca:
This is because imeta tag alone lacks some features.

Different from imeta tag, inlining with # fragment makes it possible to:
- sharing an url on DM `.content` without leaking metadata;
- can be used to extend any url on DM, no matter the protocol (even NIP-21 ones without adding new TLV types, like  e.g.: nostr:nevent1...#any=a&metadata=b);
- on web clients, user can right click/long tap on an rendered image (using default browser context menu)>copy url>paste url in another note keeping metadata

imeta tag is better for:
- having dimension and blurhash at hand faster (especially if you are delaying `.content` parsing) so to prepare the image/video container.

Giszmo:
Concept ACK

jb55:
LGTM

arthurfranca:
Just rebased, no changes

alexgleason:
IMO using URL fragments is not a good solution. See #751 which uses a JSON string.

arthurfranca:
Replaced `imeta` with `resource` tag (non-indexable) to keep same encoding. Also added some tag examples:

A kind 30023 event JSON file (with `nostr/30023` MIME type):
`["resource", "https://xyz.com/example.json#m=nostr%2F30023"]`

A PDF file without extension with added MIME type information:
`["r", "https://xyz.com/article02#m=application%2Fpdf"]`

arthurfranca:
@alexgleason your solution 1) force tag usage that may not be suitable to current DMs/encrypted events (metadata leakage). 2) make copy/pasting urls lose metadata 3) force kind 1/30023 clients showing events to not only parse `.content` (looking for mentions/urls/hashtags) but also parse tags.

⚔️ The clash of the url metadata PRs

alexgleason:
Most clients don't display inline media anyway, they pull it down to the bottom. Enumerating tags instead is an easier solution to what they're already doing.

Leaking metadata in DMs is already an issue, and it's why giftwrap is being proposed. The tags would be hidden with giftwrap.

arthurfranca:
**Most** clients, not all of them

arthurfranca:
@alexgleason This note on Amethyst gets inlined at the intended position: [https://nostr.com/nevent1qqsdnu95d6lxct2k6nx5nprauwhauchad544sctdjmrtfv33y334lfgpz3mhxue69uhhyetvv9ujumn0wd68ytnzvupzq3huhccxt6h34eupz3jeynjgjgek8lel2f4adaea0svyk94a3njdqvzqqqqqqyvseq6w](https://nostr.com/nevent1qqsdnu95d6lxct2k6nx5nprauwhauchad544sctdjmrtfv33y334lfgpz3mhxue69uhhyetvv9ujumn0wd68ytnzvupzq3huhccxt6h34eupz3jeynjgjgek8lel2f4adaea0svyk94a3njdqvzqqqqqqyvseq6w)

alexgleason:
Amethyst does everything under the sun currently possible on Nostr. It is most prepared to deal with things both ways, while the majority of clients just want to stick media at the bottom.

alexgleason:
@vitorpamplona Is there already a content-addressable solution in use?

alexgleason:
@vitorpamplona NIP-95 could work with a `media` tag with a `nostr:naddr1...` URI instead of a URL.

vitorpamplona:
> @vitorpamplona NIP-95 could work with a `media` tag with a `nostr:naddr1...` URI instead of a URL.

`nostr:addr1` in tags could be a little weird (we usually just add the event-id as an `e` tag), but sure.

501st-alpha1:
> Is there already a content-addressable solution in use?

I'd love to be able to use IPFS for this; see #131.  Some quotes from there:

> IPFS community has settled on `ipfs://<cid>` URL scheme for these portable linking capabilities.

Example:

```
ipfs://bafybeigdyrzt5sfp7udm7hu76uh7y26nf3efuylqabf3oclgtqy55fbzdi
```

> A client could convert URNs to cloudflare gateway links on the fly. The signed nostr message should use the URN because it is permanent. How the client wants to access the referenced file and how it wants to present the link to the user is client preference.

Same applies to the URL scheme, so a Nostr client that doesn't want to deal with IPFS could replace the above IPFS URL with https://cloudflare-ipfs.com/ipfs/bafybeigdyrzt5sfp7udm7hu76uh7y26nf3efuylqabf3oclgtqy55fbzdi.

However, I don't know of any Nostr clients that implement this (yet).

---

If we still need to upload files to some `https://domain.name` for now, one mitigation in the meantime would be to include the hash of the file somewhere in the signed content, see #300.

vitorpamplona:
Ideally, we should focus on solutions that don't centralize on another server. I don't think it's a good idea to resolve IPFS links through an external service. It gives too much tracking power to that server.

vitorpamplona:
@arthurfranca yep. It is an option. Unfortunately, in NIP-96, the decentralizing functions are optional (x is not required) and there is still an emphasis on hardcoding `http://server.com` addresses into signed events, which is not good IMHO. But it's better than what we have today. 

The main issue we have is that for both NIP-96 and NIP-97, we need servers and relays to get the implementation done before apps adopt it. Amethyst can't drive adoption for these ones otherwise I would have already coded them.

arthurfranca:
> The main issue we have is that for both NIP-96 and NIP-97, we need servers and relays to get the implementation done before apps adopt it

[There were 2 devs waiting to implement NIP-96](https://github.com/nostr-protocol/nips/pull/547#discussion_r1313122034). Don't know how they are feeling now.

501st-alpha1:
> Ideally, we should focus on solutions that don't centralize on another server. I don't think it's a good idea to resolve IPFS links through an external service. It gives too much tracking power to that resolving server. Maybe we can add 100+ potential IPFS resolvers to be used at random by the app.

Oh definitely; my preference would be to have an option to allow routing through my personal IPFS node, but I was just pointing out that even without doing that, it's very easy to route through a public resolver instead (or even one pulled at random from a list as you suggest).  All the spec should say is files on IPFS can be referenced with `ipfs://<cid>`, and then it's up to the client to decide how to display them.

vitorpamplona:
Should we put `ox` in the fragment as well? That could be a way to tell the receiving client that this author is using a Client with NIP-96 support and might have sent the file to other servers. 

The other way to do it is to have a list of all NIP-96 servers and parse the `ox` tag from the full URL if it matches a known provider.

arthurfranca:
> Should we put ox in the fragment as well? 

The `ox` hash is useful when paired with the author's kind:10096 "File Server Preference event" to discover up-to-date fallback urls. But this won't work well if upon updating kind:10096 entries, media files aren't moved to new servers (similar to using NIP-65 and searching for an event on a secondary write relay requires the user events to be re-broadcasted when adding new relay entries to kind:10002).

arthurfranca:
But thinking about it, how feasible it is to download and re-upload all media files to a new server when adding a new server url to `kind:10096` event? =o

One may however choose to do `https://url.xyz/file#ox=<hash>` along with `kind:10096` event to find out not-necessarily-up-to-date fallback urls instead of using the bigger `https://url.xyz/file#url=<fallback-URL-1>&url=<fallback-URL-2>`

arthurfranca:
> test case: https://www.nostr.guru/e/6c5021f86e012305a4d05c4416fcf3537bcca85f738e4ae643c7a758c911d7c1?relays=wss://nostr.bitcoiner.social

My decoder fn returned this:
```
{
    "m": "image/jpeg",
    "dim": "485x500",
    "alt": "Testing+NIP-54+inline+metadata",
    "blurhash": "|LCF#99XB-XnH??HRQjIag*H9s-B-CD%tlV@afWCAng3#Vw1buOAW:fjkBQm?HAC9@-:Vss;o0s:9F-qV@RP%MM{jvWWofELogwdnioyWBahS2fkwhVtWqX7WBWXoLs.axnnRQj[j[n+R*kCoeR%nkjbW.W.s:aKofWqRj",
    "x": "a9049db58c7284608313112cbf089fae0d960af7fcfdda1a7bffadc86be77224",
    "witdh": "485",
    "height": "500"
}
```

vitorpamplona:
So, I just coded both the imeta idea (#904) and this one to test them side by side. Both were assembled from the new NIP-96 (#547) results. I believe this one is slightly better. It's all minor things but if we are looking closely, this one wins. 

1. The implementation here is simpler/more straightforward because the URL already has everything you need to prepare the image. There is no need to pass the tag list to the Image component in the UI so that it can then search for the appropriate tags OR there is no need to pre-compute URLs with the tag list to pass it along to the UI's image component. It comes in ready to go. 

2. There is a one-to-one relationship between tags and URLs. In the imeta idea, the dev has to deal with more edge cases (bugs) like having 2 equal urls in the tag with separate params for just one URL in the `.content`. It's not a huge deal, but I can see how clients might find themselves having to deal with incorrect implementations of the imeta spec. 

3. There are minor performance improvements since the UI is not searching for imeta tags in all URLs inside `.content`, which you have to do if you are rendering images/videos from URLs that do not provide file extension cues to know what type of file the link is pointing to. With this standard, the code knows right away if there is a content type tag and can immediately discard everything else that doesn't (because the tags also wouldn't have the info). 

4. It is possible to use the same structure for `nostr:` and `magnet:` urls, which does open some possibilities of improvement when inserting other events inside `.content`. If you are pointing to a nip94 image, you can add the `dim` and `blurhash` right in the nostr URI to prepare the space for the event when it shows up.

staab:
All those things are true, but the reason this is a non-starter for me is basically the same reason fiatjaf doesn't like markdown in kind 1's. Notes need to be human-readable, and while markdown is technically human-readable, it's not a nice experience if a client doesn't parse and display it for you. Same thing with inline resource metadata. Here's an example:

```
Look at my lunch!

https://i.nostr.build/whatever#m=image/jpeg&dim=485x500&alt=Testing+NIP-54+inline+metadata&blurhash=|LCF#99XB-XnH??HRQjIag*H9s-B-CD%tlV@afWCAng3#Vw1buOAW:fjkBQm?HAC9@-:Vss;o0s:9F-qV@RP%MM{jvWWofELogwdnioyWBahS2fkwhVtWqX7WBWXoLs.axnnRQj[j[n+R*kCoeR%nkjbW.W.s:aKofWqRj&x=a9049db58c7284608313112cbf089fae0d960af7fcfdda1a7bffadc86be77224&width=485&height=500
```

In contrast:

```
Look at my lunch!

https://i.nostr.build/whatever
```

I think this is a pretty compelling reason for users to choose one client or another, either for reading or publishing this stuff. I personally use nostr.build instead of nostrcheck.me because the url looks nicer.

vitorpamplona:
But to be fair, Amethyst does have a mode where we don't load images and all huge URLs like that are shown. We already have pretty ugly URLs right now. This makes it worse, but it was already pretty bad with today's posts. 

Also, I would argue that whoever is just displaying URLs should crop them. Even without this change. That's what I am going to do. Users can copy the URL if they want to see the whole thing.

arthurfranca:
Yes even before nip54 its a good idea to remove http(s)://, put some char limit and add  '…' to the url when rendering the note if not an image/video.

vitorpamplona:
> Do we? nostr.build and primal's image urls are extremely tidy, what else are you thinking of?

On mobile, most servers return 3-line URLs. today. The extreme test (super large blurhash) give is it a 10-line (4 is the blurhash alone). 

|   |   |
|---|---|
|Screenshot_20231211_173945|Screenshot_20231211_174009|

staab:
🤮 they should fix that. I personally switched from nostr.build from nostrcheck for exactly this reason. Tidy urls are nice.

vitorpamplona:
> 🤮 they should fix that. I personally switched from nostr.build from nostrcheck for exactly this reason. Tidy urls are nice. But maybe I'm a nerd.

With NIP-96 everything is a long URL :(

staab:
Not necessarily, NIP 96 allows the host the pass any url they like: https://github.com/nostr-protocol/nips/pull/547/files#diff-b27fe23ed5ff40d3f85bf18f65ecadaeaf905d363a6df3c20cb2dffd30f903d9R164

jb55:
> So, I just coded both the imeta idea (#904) and this one to test them side by side. Both were assembled from the new NIP-96 (#547) results. I believe this one is slightly better. It's all minor details but if we are looking closely, this one wins.

weird to me you would give up metadata from one of the largest clients over these minor differences, but alright. I guess we have two competing standards in the wild now.

jb55:
> 🤮 they should fix that. I personally switched from nostr.build from nostrcheck for exactly this reason. Tidy urls are nice. But maybe I'm a nerd.

would have to strip the # if metadata is detected I guess

vitorpamplona:
> weird to me you would give up metadata from one of the largest clients over these minor differences, but alright. I guess we have two competing standards in the wild now.

Nothing personal. I go with what's best at the time.

vitorpamplona:
BTW, For completeness, I also implemented shoving the entire NIP-94 event in a tag as a serialized json for each URL and then using a `nostr:nevent...` to point to the image in the `.context` instead of the URL. 

It's ok. But it ended up still URL-based (it just transferred the limitations of dealing with a URL to the NIP-94 event). And since this PR allows the use of any nip-94 tag in the fragment, this version is just shorter.

alexgleason:
> would have to strip the # if metadata is detected I guess

```ts
function renderLink(url: string): string {
  const url = new URL(attributes.href);
  const search = new URLSearchParams(url.search);

  if (isMediaParams(search)) {
    url.hash = '';
  }

  return `<a href="${url.href}">${url.href}</a>`;
}

function isMediaParams(search: URLSearchParams): boolean {
  // how am I supposed to detect this?
  return search.has('m'); // ?
}
```

alexgleason:
You can't remove the hash unconditionally, otherwise people can't post links like: `https://en.wikipedia.org/wiki/Nostr#See_also`

vitorpamplona:
You should remove it just for display purposes, not from the actual link.

alexgleason:
I don't think it's a good solution to hide the hash from URLs.

vitorpamplona:
> I don't think it's a good solution to hide the hash from URLs.

You mean the file at the end of the path? 

For instance, for 

`https://i.nostr.build/n5AQ.jpg#m=image%2Fjpeg&dim=1080x1626&blurhash=%5EE986*Z%24DQr.nMkY8w%25f-mWtS%24ngvfoecEn*r%3Fav.mVtMxs8WTo%7ED*t6%24eS%24g4aL%3FEb0K6v%7DnNRiIBxY%24%25bdW%3Bxtx%5Br%5Daeoet6Rk%252kDWYi%5EV%40j%3F&x=4d9ad2dd6635fc937427ed0fbb723c0b925d1e0b304b925a65a1185581596dfe`

I remove the `https://` scheme, all query params, fragments and just show `i.nostr.build/n5AQ.jpg`. 

For long hashes, like: 

`https://image.nostr.build/10f7dc11a35d6fb0f8372666e6392287cd272d543a3832fb3e5796a547d83172.jpg#m=image%2Fjpeg&dim=1920x1440&blurhash=%23BGHuG0K%23R%7Ep9F%25zr%3Dxtxuv%7D%7EV9ar%3Dx%5DD%25tRi%7BkC00m-cE%24fNxn%2CkCR*o001E1-%3BEMsmtRVskWsmxaXmr%3FIp%250j%40RjxZSN%7EWtQR5WXjYRPozRjRktQs%2BRjWBofV%40R*WCV%5B&x=fd8b2438b21bfc35a560c1ac6496f8c3fc8cca085ebd734d98ecd7ac22279df3`

I add a `...` between the host and the end of the path: 

`image.nostr.build/...a547d83172.jpg`

Assuming of course that I can't preview it for some reason.

staab:
Hashes also seems to break imgproxy. Still NACK on this, imeta is better.

vitorpamplona:
> Hashes also seems to break imgproxy. Still NACK on this, imeta is better.

Any examples of the issue? If it is breaking in these simple cases, we better fix it before the real content servers come in.

staab:
Oh, nevermind, the bug was my own

v0l:
Snort is using the `imeta` tag i don't link these long URL's

jb55:
some of the links seem to have evaded my url parser, leading to broken images in a few cases :cry:

Yonle:
Most image parser will expecting `.jpg`, `.png`, `.gif`, or similiar at the end of the URL. But this completely break the parsee that most nostr client use.

fiatjaf:
I think this was a good try but the big URLs create a big burden on everybody and we should stop.

I guess it's correct to say that the URL parsers and regexes _should have been ready_ for this, but in practice in the real world no one ever does such big URLs with # and * and { and whatnot, and when they do it's not a big deal if they fail to get recognized, but now it's thousands of notes with them and their authors don't even know they're publishing such giant monsters.

Yonle:
There are even posts that put `.mp4` (or similiar at the end of url) after `#` for attachments.

vitorpamplona:
> Most image parser will expecting .jpg, .png, .gif, or similiar at the end of the URL.

This is the dumbest way to check what the content type of a URL is. There are no file extensions in URIs. Some things look like them that show up in URIs, but that's only because some servers map URIs to files internally. They are meaningless for anything outside those servers.

The only way to reliably establish if a resource identified by a particular URI is an image or not is to make a HEAD or GET request and look at the Content-Type response header.

Take the famous site 500px, for instance. None of their photo URLs end with an extension.  https://drscdn.500px.org/photo/79694843/m%3D2048_k%3D1_a%3D1/v2?v=0&sig=5cd241cf922efd12d8aa1dac2722e261b006648cd0b7e3902bb6f4360e17d450

If you are relying on extensions, you are offering a terrible user experience already.

Yonle:
> > Most image parser will expecting .jpg, .png, .gif, or similiar at the end of the URL.
> 
> This is the dumbest way to check what the content type of a URL is. There are no file extensions in URIs. Some things look like them that show up in URIs, but that's only because some servers map URIs to files internally. They are meaningless for anything outside those servers.

> Take the famous site 500px, for instance. None of their photo URLs end with an extension.  https://drscdn.500px.org/photo/79694843/m%3D2048_k%3D1_a%3D1/v2?v=0&sig=5cd241cf922efd12d8aa1dac2722e261b006648cd0b7e3902bb6f4360e17d450
> 
> If you are relying on extensions, you are offering a terrible user experience already.

The thing is that **most nostr clients** did it this way for quickly recognizing the URL as an attachment.

vitorpamplona:
> The thing is that most nostr clients did it this way for quickly recognizing the URL as an attachment.

Sure. They can keep doing it. Just remove everything after `?` and after `#` before checking. I will work for the images from the PR, but it won't work for many image servers out there.

fiatjaf:
> If you are relying on extensions, you are offering a terrible user experience already.

We should standardize and enforce that extensions are required for media content.

vitorpamplona:
> We should standardize and enforce that extensions are required for media content.

Why? We have NIP-94, `imeta` and this PR to offer proper content types. Bringing back an outdated file extension model from the 90s on top of them just duplicates everything yet again. 

Keep in mind that clients can't add extensions on links that don't already offer them. There is not much we can do. 

If we want to truly fix this, we should not use URLs at all. Everything else is just a half-ass hack.

fiatjaf:
> If we want to truly fix this, we should not use URLs at all. Everything else is just a half-ass hack.

I agree, but Nostr itself is a half-ass hack. The URL system is very broken and I wanted to fix it too, but is it worth fixing? We have tradeoffs here at hand.

> Why? We have NIP-94, imeta and this PR to offer proper content types. Bringing back a flimsy outdated file extension model from the 90s on top of them just duplicates everything yet again.
>
> Keep in mind that clients can't add extensions on links that don't already offer them. There is not much we can do.

But yeah, maybe let's not try to force-standardize this for now, since it wouldn't work. But I still think people should try to put extensions in URLs whenever they can, as a kind gesture to the dumber clients that rely on that flawed hackish property of URLs.

vitorpamplona:
> is it worth fixing?

Yes. The decentralization benefits alone are a huge win. And clients get a better set of preview tags to optimize their UX before the content is loaded. It's a win-win situation. 

This feels very similar to the NIP-04 discussion. Everybody knew it was wrong but complacency and laziness kept us stuck in that model for far too long. 

Sometimes the easiest design and the simplest design are vastly different from one another. This is one of those cases. We can keep piling into an easy design the URL overlords want us to use or design our own content naming and resolution scheme to make it 10x better (more aligned with Nostr needs). 

In the end, our design will be simpler to implement than trying to make URLs work.

arthurfranca:
> Most image parser will expecting .jpg, .png, .gif, or similiar at the end of the URL. But this completely break the parsee that most nostr client use.

You telling me kind:1 client devs that are already dealing with decentralized relay quirks and other harder NIPs are incapable of updating an url regexp to a valid one? =o

If `#...` is breaking them, `?...` was already breaking these parsers too

jb55:
On Wed, Dec 27, 2023 at 09:39:42AM -0800, arthurfranca wrote:
>You telling me kind:1 client devs that are already dealing with
>decentralized relay quirks and other harder NIPs are incapable of
>updating an url regexp to a valid one? =o

This is the code in damus:

func classify_url(_ url: URL) -> UrlType {
     let str = url.lastPathComponent.lowercased()
     
     if str.hasSuffix(".png") || str.hasSuffix(".jpg") || str.hasSuffix(".jpeg") || str.hasSuffix(".gif") || str.hasSuffix(".webp") {
         return .media(.image(url))
     }
     
     if str.hasSuffix(".mp4") || str.hasSuffix(".mov") || str.hasSuffix(".m3u8") {
         return .media(.video(url))
     }
     
     return .link(url)
}

What are we arguing about exactly? what is wrong with using extensions?
It works perfectly fine and is simple for devs to implement.

arthurfranca:
It works alright no need to change. Clients will only fail to load an image/video if the parser use some poor regex not accounting for valid urls (that may include `?` query params or `#` hash) and this has nothing to do with this NIP because the file extension is not stripped from the url.

> some of the links seem to have evaded my url parser

Probably just media with no extension. Nothing related to this NIP.

On NIP-96 i have added the advice to append the media extension to urls so that nip54 or imeta isnt required to detect mime type.

Nip54 is totally optional and beautifying long urls or not when no media is detected is a dev choice.

Issue #522, opened 2023-05-12
algorithms on nostr that have a chance of working


huumn:
Nostr already has an "algo" all relays use to order events. The algo is "sort by created_at in descending order." What I want is a way to ask relays for events using a variety algos such that algorithms are
1. provided in a censorship resistant way
2. transparent
3. compete

I'd like to get some initial feedback before drafting a NIP.

I'm unsure of the nostr idioms but we need a way for clients to specify an algorithm and its arguments.

What I'm thinking:
```json
{
    kinds: [1],
    algo: [ALGO_NAME, ALGO_ARG1_NAME, ALGO_ARG1_VALUE, ..., ALGO_ARGN_NAME, ALGO_ARGN_VALUE]
}
```

Where ALGO_NAME is an algorithm, and the rest of the array are argument name-value pairs spread into an array of strings. Algorithms should be rigourously specified as their own NIPs and clients may check for ALGO_NAME support and request events using them. All normal REQ arguments should be respected like normal and operate on the algo's input or output set (where the REQ args are applied, if relevant, should be specified in the algo's NIP probably).

e.g.
```json
{
    kinds: [1]
    algo: ["TrustRank", "degree", "3"]
}
```

Anyway, that's the gist of what I'd like to propose. There's more to consider like how clients reconcile different results from relays, etc, but I'd like to guage if there's interest in this approach to algorithms first.

Is this the right idiom? Is there prior work I should be using instead/considering?

Alternatively, a new verb, much like `COUNT` could be proposed. wyt

alexgleason:
I've been thinking about this too, since seeing bluesky's feed generator: https://github.com/bluesky-social/feed-generator

Instead of every relay implementing support for an algo filter, I was thinking it makes sense to create purpose-built relays, and then clients should add the ability to view a specific relay's feed. It doesn't even have to be a full relay, it could just proxy events from other relays (like an aggregator) and display specific content in a specific order.

Coracle already lets you view the feed of individual relays: https://coracle.social/relays/d3NzOi8vcmVsYXkuc3RvbmVyLmNvbQ==

staab:
I'd like to see this too, my preference would be a new verb that asks for recommendations so that relays can implement it if they want to (mostly so the NIP can be extended with more functionality over time), although the above `algo` property on `REQ` could also work.

[...]

PR #524, opened 2023-05-13
Update NIP-56 to handle more detailed content moderation and generic labeling

staab:
# Purpose

Content reporting is a type of labeling. In https://github.com/nostr-protocol/nips/pull/459, @gruruya proposed expanding NIP-56 to encompass labeling.

In https://github.com/nostr-protocol/nips/pull/457, @s3x-jay and @rabble proposed a nomenclature and more comprehensive mechanism for content reporting/labeling. The consensus seems to be that a nomenclature is not a good idea, but there were some good ideas to include.

In https://github.com/nostr-protocol/nips/issues/522#issuecomment-1546286149 I realized that the same labeling mechanism could be used to leave reviews for relays, which I am eager to add to Coracle so people can start to find a better relay selection.

This PR may also render https://github.com/nostr-protocol/nips/pull/46 obsolete.

# Justification

Maybe this will seem like I'm trying to make a single NIP do too many things. But reviews, labels, and reports all have two things in common: they refer to an object, and are an expression of someone's opinion. In each case, there is a value judgment involved in assigning the label, involving both imprecision and uncertainty. Expressing this opinion using either a mark, or content, or both, allows either a machine or human to make a judgment call as needed.

One nice thing I'd also like to point out about this PR is if clients follow the bi-directional tagging rule, it allows for associating any two entities (e/p/r/t) in nostr, and searching those associations. This gives us the beginning of graph-database-like functionality that can form the basis for both WoT and topical moderation and recommendations.

Confidence/quality may seem like the same thing, but they're two different axes along which to measure certainty. The first is subjective (the label author may not be 100% sure), the second objective (the label may not perfectly fit). This has been suggested in a few places, but was best executed by guide.newfounding.org (unfortunately their site is no longer available).

# Compatibility

This PR should be backwards-compatible "enough" with the old version of NIP-56, but if there's disagreement on it, I can re-draft as a separate NIP. I'm particularly interested here in @jb55's feedback since he wrote the original version and is probably the only person who is relying on its implementation. 

The only backwards-incompatible change I had to make is that context must now be marked by a `context` tag. My sense is that the `p` tag was originally required so that a given event could be located.  I don't expect that anyone has implemented any automatic routing based on the `p` tag, and clients that continue to include a `p` tag will simply be reporting the pubkey in addition to the event, which seems to be acceptable semantically.

Other minor changes:

- `e` tag is now optional
- The report type (now "label") is now optional, and is no longer required to use a given vocabulary. Tags without labels are allowed, since sometimes it might make sense to only leave a review in `content`, or if the `quality` and `confidence` tags are sufficient.

# Implementation

None of this is implemented, but my main motivation of drafting this PR is so I can add relay ratings and reviews to Coracle soon.

### DIFF
--- a/56.md
+++ b/56.md
@@ -2,66 +2,126 @@
 NIP-56
 ======
 
-Reporting
+Labeling
 ---------
 
-`draft` `optional` `author:jb55`
+`draft` `optional` `author:jb55` `author:staab` `author:michaelhall923` `author:gruruya`
 
-A report is a `kind 1984` note that is used to report other notes for spam,
-illegal and explicit content.
+A label is a `kind 1984` note that is used to report other notes for spam, illegal, explicit, or any other type of content.
 
-The content MAY contain additional information submitted by the entity
-reporting the content.
+The content MAY contain additional information submitted by the entity reporting the content.
 
-Tags
+Label Tag
 ----
 
-The report event MUST include a `p` tag referencing the pubkey of the user you
-are reporting.
+The label event MUST include one or more tags representing the object or objects being
+labeled: `e`, `p`, `r`, or `t` tags. This allows for labeling of events, people, relays,
+or topics respectively.
 
-If reporting a note, an `e` tag MUST also be included referencing the note id.
+A `label` string MAY be included as the 3rd entry to each object tag. This MAY be an unqualified string indicating the type of content based on convention, or a qualified string referring to a nomenclature external to nostr. Some examples of report types:
 
-A `report type` string MUST be included as the 3rd entry to the `e` or `p` tag
-being reported, which consists of the following report types:
+- `#t/footstr` - the publisher thinks the given event should have the `footstr` topic applied.
+- `#p/<pubkey>` - the publisher things the given event should be tagged with with `<pubkey>`
+- `MeSH/D005528` - ["Foot"](https://meshb.nlm.nih.gov/record/ui?ui=D005528) from NIH's Medical Subject Headings vocabulary
+- `GeoNames/3173435` - [Milan, Italy](https://www.geonames.org/3173435/milan.html) using the GeoNames coding system
+- `ISO-3166-2/IT-MI` - Milano, Italy using ISO 3166-2.
+- `nudity`, `profanity`, `illegal`, `spam`, `impersonation` - informal, but potentially useful content-based reports
 
-- `nudity` - depictions of nudity, porn, etc.
-- `profanity` - profanity, hateful speech, etc.
-- `illegal` - something which may be illegal in some jurisdiction
-- `spam` - spam
-- `impersonation` - someone pretending to be someone else
+As much as possible, fully-qualified labels should be used if the goal is precision, for
+example for moderation. Otherwise, convention should be the rule.
 
-Some report tags only make sense for profile reports, such as `impersonation`
+When associating standard nostr tags with one another via a label, clients SHOULD provide
+two tags defining the same association in reverse order so that the association is searchable.
+For example, if you wanted to say `<pubkey>` is related to `#pizza`, you would include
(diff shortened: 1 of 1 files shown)


staab:
Hang on, for some reason this is pulling in way more commits, let me fix it.

Gruruya:
I see you kept quality between 0 to 1, do you want to include a seperate field like `[negative, true/false]` that could be better queried?

mikedilger:
The meaning of the fields in a tag has not depended on the event kind before.  Here we have an 'e' tag where the meaning of the fields is different then every other e tag in every other kind.  This complicates my code lots.  So I'm against unless this can be changed.

Gruruya:
That isn't new, it's how kind 1984 is already implemented. 
https://github.com/nostr-protocol/nips/blob/00491aa9facdcb91a88abd78080e76dccd5eda03/56.md?plain=1#L22-L25

It may be better to eschew kind 1984 for a different structure, but it's been kept this way for backwards compatability.  
I'm starting to think it would be better to just use a new kind and do it right as 1984 seems mostly unimplemented (rightfully so.)

staab:
Yes, the intention was for backwards compatibility, but I agree that the structure used in 1984 is abhorrent. I'll draft a new PR this week with a cleaner implementation.

staab:
> I see you kept quality between 0 to 1, do you want to include a seperate tag like [negative, true/false] that could be better queried?

As I was putting this PR together, I came to the conclusion that negation isn't necessary if you think in terms of a venn diagram rather than vectors to/away from a target. In the former metaphor, a negative match makes no sense, since 0 indicates no overlap, IOW a total negation. The latter might indicate toward (+1) or away from (-1) some target or standard, but that metaphor only applies to some concepts (you might be able to justify negative quality, but not negative confidence). So my preference is now a spectrum of 0 to 1, I think it's much cleaner.

Gruruya:
Here's what I'm thinking so far; single reference, multiple labels.
```JSON
{
  kind: ???
  tags: [
    ["p", <pubkey>],
    ["t", "label"],
    ["t", "label2"]
  ]
}
```
Any tags attached to the event besides tags in the spec like `e`/`p`/`r` could be interpreted as applying to what's referenced.  
So you could apply a `description` tag to a relay for instance, or even an updated `relay` tag.

I see two ways in which `l` tags extend `t` tags, namespaces and quality. I say we make `namespace` the third field and `quality` the fourth, assuming the default `namespace: topic` and `quality: 1`. I think simply extending `t` tags seems like the right way to go instead of introducing a new, strict superset, tag and it seems you agree. I want to include `quality` on the label-level so that different labels in the same event can have different qualities.

Keeping `confidence` as a special tag for labeling events is fine as `confidence` doesn't make much sense when tagging your own event, and in this case the downside of applying to all labels in the event is acceptable IMO.

Other thoughts:  
I think we should keep the `context` tag you introduced, I like it.
With my described changes, you would lose the ability to specify the `quality` of a reference without a label, but you could use a blank label to achieve something similar.

staab:
Closed in favor of https://github.com/nostr-protocol/nips/pull/532

staab:
@Gruruya I don't mind any suggestions you made above, but I went a slightly different direction with the new PR. There is a trade-off between being able to apply multiple labels to a single entity vs applying a single label to multiple entities, I went with the latter in the new PR. I removed `context`, since that was mostly to support NIP 56, which I've decided not to touch for now. I do think namespace is helpful, I'm ambivalent as to whether it should be included in a label, except that defining fewer tags keeps things simple, and a slash is trivial to parse.

Gruruya:
I'll write something, I want multiple labels for one reference as I expect to be doing a lot of mass-tagging. It seems two seperate implementations is necessary for both.

staab:
Ok, I can take another look. I suppose we could just support multiple labels and multiple targets, with the implication being that all labels are applied to all entities.

PR #525, opened 2023-05-13
Idea, add "name" tag to channel create

earonesty:
It's probably going to be very helpful to search for channels by name.   Placing a recommended "d" tag, so a user can search channels by name would probably be helpful in chat implementations.

### DIFF
--- a/28.md
+++ b/28.md
@@ -28,6 +28,7 @@ In the channel creation `content` field, Client SHOULD include basic channel met
 ```json
 {
     "content": "{\"name\": \"Demo Channel\", \"about\": \"A test channel.\", \"picture\": \"https://placekitten.com/200/200\"}",
+    "tags": [["d", <channel name>]],
     ...
 }
 ```


staab:
I appreciate this, but is there an implementation that does this? Also, a `d` tag completely changes the semantics of the event. Closing this.

staab:
Feel free to re-PR for the name issue, but `search` currently works only on content, so you don't get search unless you have an exact match. Also, new functionality usually should be supported by an implementation, is there a client that has solved this problem? NIP 33 isn't explicit, but `d` is only defined for replaceable events as far as I can tell. https://github.com/nostr-protocol/nips/blob/master/33.md

earonesty:
relays allow searching on prefixes, so you can search for "cat" or "car" or something and expect a decent list.  probably we need a full-text search standard.   i think the invoiceable RPC-request is the way to go for that!

earonesty:
yes   nip 01

```json
{
  "ids": <a list of event ids or prefixes>,
  "authors": <a list of pubkeys or prefixes, the pubkey of an event must be one of these>,
  "kinds": <a list of a kind numbers>,
  "#e": <a list of event ids that are referenced in an "e" tag>,
  "#p": <a list of pubkeys that are referenced in a "p" tag>,
  "since": <an integer unix timestamp, events must be newer than this to pass>,
  "until": <an integer unix timestamp, events must be older than this to pass>,
  "limit": <maximum number of events to be returned in the initial query>
}
```

badge names are not searchable.   NIP-50 is fine.   it will, weirdly, search the content field as if it's one big string though.   try doing a REQ with search kind 40 events and the keyword "about" ... you get all of them!

staab:
> it will, weirdly, search the content field as if it's one big string though

Yes, which is why `content` should be plain text for future NIPs. I'm not sure what `#k` accomplishes, maybe just a generic search index? But if a client is looking for something in particular it should be able to either target it or rely on `search`. I think `search` covers this, relays just need to be smarter about the implementation.

I was asking about "relays allow searching on prefixes", I don't see anything in NIP 01 about that.

earonesty:
``"ids": <a list of event ids or prefixes>`

presumably it's not too much of a burden to say other things can be searched by prefixes, if they're already doing it for ids and authors

staab:
Are they? Again, I have not heard that. Any nip or implementation

earonesty:
nip01 includes ids and author prefixes

staab:
Wow I have never noticed that. I guess it would be reasonable to update NIP 01 to allow prefix matching on all tags

earonesty:
kinda suprised there's no keyword-list feature on notes in general.   i get that SEARCH sort-of handles this.   but self-tagging is probably the best way to  organize stuff.  way better than expecting every relay to operate a full text index.

arthurfranca:
startsWith queries are harder on tags because tags are sets of values.
Considering the event `{ ..., tags: [["x", "help, David"], ["x", "hello, John"]] }`, tag `x` is in fact `["help, David", "hello, John"]`, so a prefix search like `{ #x: ["hel"] }` is harder than on ids and authors.

Adding it to NIP-01 may rule out some relay databases in use.

earonesty:
is "x" a legit tag value?   do we have a keyword tag?   i care less about prefixes than having *any tag at all*

arthurfranca:
For now there is `t` for hashtags. Any one-letter tag is valid for searching. But just for exact match searches, not prefix searches.

staab:
NIP 32 labels would technically work, but we should probably have an actual tag. `i` or `r` could potentially fit. `t` is definitely the wrong choice.

arthurfranca:
The "flaw" of NIP-28 is that it was created before NIP-33 existed (parameterized replaceable events - PRE) so the updated channel name is at kind `41` that is similar to PRE but uses the `e` tag instead of `d` to dedupe (or one just sends deletion event for old kind 41 and then sends a new one to replace it).

I think you could use any tag like `n` meaning channel "n"ame but it would have to be present at both kind 40 and kind 41 events?

earonesty:
> I think you could use any tag like `n` meaning channel "n"ame but it would have to be present at both kind 40 and kind 41 events?

it definitely should be a replacement, and should be present on all events.  need multiple labels though.  t sounds fine.

PR #527, opened 2023-05-13
add nip101 - Alias exchange

wcat7:
Alias exchange

implement: [https://github.com/0xchat-app/nostr-dart/blob/main/lib/src/nips/nip_101.dart](https://github.com/0xchat-app/nostr-dart/blob/main/lib/src/nips/nip_101.dart)

### DIFF
--- a/101.md
+++ b/101.md
@@ -132,6 +132,9 @@ For example:
 
 A reference implementation for this protocol can be found [here](https://github.com/0xchat-app/nostr-dart/blob/main/lib/src/nips/nip_101.dart).
 
+## Note
+
+The final events will be packaged and sent via [Gift Wrap](https://github.com/nostr-protocol/nips/blob/fd914c600fd590e6188534cd7555116da7855503/59.md), so there is no need to worry about information leakage.
 
 ## Use Cases
 
--- a/101.md
+++ b/101.md
@@ -1,123 +1,145 @@
-# NIP-101: Alias Exchange
+# NIP-101: Alias Key Exchange
 
 ## Abstract
 
-This NIP proposes the 'Alias Exchange' protocol, designed to facilitate the secure exchange of alias keys in private communication, thus protecting users' real npub keys. The protocol includes four kinds of events, each serving a specific function in the exchange process.
-
-## Motivation
-
-The Alias Exchange protocol enhances the security of NIP04 (private chat) by providing a means to protect real npub keys. By exchanging alias keys, users can communicate without revealing their real keys, reducing the risk of unauthorized access and potential data breaches. This protocol can be particularly useful in scenarios such as friend requests, acceptance of friend requests, rejection of friend requests, and removal of friends.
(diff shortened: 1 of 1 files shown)


101.md
```diff
@@ -0,0 +1,111 @@
...
+
+- **10103 - Remove Alias**: This event is used to remove the alias. It is signed with the initiator's alias and sent to the recipient's alias public key.
+
+### Implementation
+
+The protocol can be implemented following the steps below. Note that the content of the content field in this protocol is encrypted following the NIP04 protocol:
+
+1. User A (the initiator) generates an alias key pair derived from their own private key and User B's (the recipient) public key.
+
+2. User A sends a 10100 kind event to User B, signed with their alias private key. The 'p' parameter is User B's real pubkey. The `encrypted_text` can be in json format: `{"p":"A's real pubkey", "content":"your content"}`.
```
Redpug111, line 30: Shouldn't the encrypted_text contain some form of verification that A is actually sending this message? Any attacker could just put A's pubkey into the event and B would have no way to tell if it was actually sent by A.

water783: Thank you, you're absolutely right. We can add a "sig" field in the text:

`{
    "p":"A's real pubkey",
    "content":"your content", 
    "sig":"<64-bytes hex of the signature of the event id, signed by A's real private key>"
}`

Redpug111: What is the "event id" in this scenario? We only have p and content fields, is it just some hash of that?

water783: Apologies for any confusion earlier. Let me clarify: it's the request event id send to B, we will have two signatures. 

The first signature of the event id is generated with A's alias private key and is included in the `sig` field of the event(NIP01)

The second signature of the event id is generated with A's actual private key and is included within the encrypted `content`. 

Here's how the event would look:

```json
{
 "id": "event id",
 "pubkey": "A's alias public key for B",
 "kind": 10100,
 "tags": [
    ["p", "B's real public key"],
  ],
 "content": "<encrypted_text>?iv=<initialization_vector>",
 "sig": "signature generated with A's alias private key",
 ...other fields
}
```
In the content field, <encrypted_text> would be a JSON string:
```json
{ 
  "p":"A's real pubkey", 
  "content":"your content",  
  "sig":"<The second signature , signed by A's real private key>" 
}
```

water783: Apologies for any confusion earlier. Let me clarify: 

data: `[eventCreateTime, p, content]` , where eventCreateTime  corresponds to the time the event was created
sig: 64-bytes hex of the signature of the sha256 hash of the `data` mentioned above, generated with A's real private key.

vitorpamplona:
Does it make sense to transfer aliases without encryption?

water783:
> Does it make sense to transfer aliases without encryption? 

I removed the encryption because the entire event will be sent via gift wrapped.

It seems I should add this note to the document. :)

vitorpamplona:
ohhh makes sense.

PR #529, opened 2023-05-14
Minor Fixes of NIP-57

jiftechnify:
- Replaced `zap note` with `zap receipt`
- Fixed inconsistent quoting of `zap request` & `zap receipt` (quoted all occurrences of them)
- Fixed description about the `nostr` query parameter in the Appendix D
  - "the contents of the event" would be misread as "the `content` field of the event", though the `nostr` query param includes the whole event.


fiatjaf:
Looks good to me.

PR #530, opened 2023-05-14
NIP-89: Recommended Application Handlers

pablof7z:
This PR describes what we've been discussing for a couple of weeks: kind-specific application handlers.

https://github.com/pablof7z/nips/blob/application-handlers/89.md

cc @fiatjaf @staab @brugeman

### DIFF
--- a/89.md
+++ b/89.md
@@ -1,0 +1,107 @@
+NIP-89
+======
+
+Recommended Application Handlers
+--------------------------------
+
+`draft` `optional` `author:pablof7z`
+
+This NIP describes `kind:31989` and `kind:31990`: a way for unknown event-kinds application-handlers to be discovered.
+
+## Rationale
+Nostr's discoverability and transparent event interaction is one of its most interesting/novel mechanics. This NIP seeks to provide the glue for cross-`kind` client interaction.
+
+### Parties involved
+There are three actors to this workflow:
+
+* application that handles a specific event kind
+    * Publishes `kind:31990`, detailing how apps should redirect to it
+* user who recommends an app that handles a specific event kind
+    * Publishes `kind:31989`
+* user who seeks a recommendation for an app that handles a specific event kind
+    * Queries for `kind:31989` and, based on results, queries for `kind:31990`
+
+# Events
+
+## Recommendation event
+```json
+{
+    "kind": 31989,
+    "pubkey": <recommender-user-pubkey>,
+    "tags": [
+        [ "d", <event-kind-handler> ],
+        [ "p", <app-pubkey> ]
+    ]
+}
+```
+
+## Handler information
+```json
+{
+    "kind": 31990,
+    "pubkey": <application-pubkey>,
+    "tags": [
+        [ "d", <event-kind-handler> ]
+        [ "web", "https://example.com/", "root" ],
+        [ "web", "https://..../a/<bech32>", "nevent" ],
+        [ "web", "https://..../p/<bech32>", "nprofile" ],
+        [ "web", "https://..../e/<bech32>" ],
+        [ "ios", ".../<bech32>" ]
+    ]
+}
+```
+
+`bech32` in a URL MUST be replaced by clients with the NIP-19-encoded entity that should be loaded by the application.
+
+Multiple tags might be registered by the app, following NIP-19 nomenclature as the second value of the array.
+
+A special `root` tag MIGHT be included with the main page of the application.
+
+A tag without a second value in the array SHOULD be considered a generic handler for any NIP-19 entity that is not handled by a different tag.
+
+# User flow
+A user A who uses a non-`kind:1`-centric nostr app could choose to announce/recommend a certain kind-handler application.
+
+When user B sees an unknown event kind, e.g. in a social-media centric nostr client, the client would allow user B to interact with the unknown-kind event (e.g. tapping on it).
+
+The client MIGHT query for the user's and the user's follows handler.
+
+# Example
+
+## User A recommends a `kind:31337`-handler
+User A might be a user of Zapstr, a `kind:31337`-centric client (tracks). Using Zapstr, user A publishes an event recommending Zapstr as a `kind:31337`-handler.
+
+```json
+{
+    "kind": 31989,
+    "tags": [
+        [ "d", "31337" ],
+        [ "p", "1743058db7078661b94aaf4286429d97ee5257d14a86d6bfa54cb0482b876fb0" ],
+    ]
+}
+```
+
+## User B interacts with a `kind:31337`-handler
+User B might see in their timeline an event referring to a `kind:31337` event
+(e.g. a `kind:1` tagging a `kind:31337`).
+
+User B's client, not knowing how to handle a `kind:31337` might display the event
+using its `alt` tag (as described in NIP-31). When the user clicks on the event,
+the application queries for a handler for this `kind`:
+
+`["REQ", <id>, '[{ "kinds": [31989], "#d": [31337], 'authors': [<user>, <users-contact-list>] }]']`
+
+User B, who follows User A, sees that `kind:31989` event and fetches the `p`-tagged profile for handler information.
+
+```json
+["REQ", <id>, '[{
+    "kinds": [31990],
+    "#d": [31337],
+    "authors": ["1743058db7078661b94aaf4286429d97ee5257d14a86d6bfa54cb0482b876fb0"]
+}]']`
+```
+
+User B's client sees the application's `kind:31990` which includes the information to redirect the user to the relevant URL with the desired entity replaced in the URL.
+
+## Alternative query bypassing `kind:31989`
+Alternatively, users might choose to query directly for `kind:31990` for an event kind. Clients SHOULD be careful doing this and use spam-prevention mechanisms to avoid directing users to malicious handlers.
--- a/89.md
+++ b/89.md
@@ -39,7 +39,7 @@ There are three actors to this workflow:
 
 The `d` tag in `kind:31989` is the supported event kind this event is recommending.
 
-Multiple `a` tags can appear on on the same `kind:31989`.
+Multiple `a` tags can appear on the same `kind:31989`.
 
 The second value of the tag SHOULD be a relay hint.
 The third value of the tag SHOULD be the platform where this recommendation might apply.
--- a/89.md
+++ b/89.md
@@ -29,7 +29,7 @@ There are three actors to this workflow:
     "kind": 31989,
     "pubkey": <recommender-user-pubkey>,
     "tags": [
-        [ "d", <event-kind-handler> ],
+        [ "d", <supported-event-kind> ],
         [ "p", <app-pubkey> ]
     ]
 }
@@ -41,7 +41,7 @@ There are three actors to this workflow:
     "kind": 31990,
     "pubkey": <application-pubkey>,
     "tags": [
-        [ "d", <event-kind-handler> ]
+        [ "d", <supported-event-kind> ]
         [ "web", "https://example.com/", "root" ],
         [ "web", "https://..../a/<bech32>", "nevent" ],
         [ "web", "https://..../p/<bech32>", "nprofile" ],
@@ -94,11 +94,11 @@ the application queries for a handler for this `kind`:
 User B, who follows User A, sees that `kind:31989` event and fetches the `p`-tagged profile for handler information.
 
 ```json
-["REQ", <id>, '[{
+{
     "kinds": [31990],
     "#d": [31337],
     "authors": ["1743058db7078661b94aaf4286429d97ee5257d14a86d6bfa54cb0482b876fb0"]
-}]']`
+}
 ```
 
 User B's client sees the application's `kind:31990` which includes the information to redirect the user to the relevant URL with the desired entity replaced in the URL.
--- a/89.md
+++ b/89.md
@@ -37,6 +37,8 @@ There are three actors to this workflow:
 }
 ```
 
+The `d` tag in `kind:31989` is the supported event kind this event is recommending.
+
 Multiple `a` tags can appear on on the same `kind:31989`.
 
 The second value of the tag SHOULD be a relay hint.
@@ -46,7 +48,7 @@ The third value of the tag SHOULD be the platform where this recommendation migh
 ```json
 {
     "kind": 31990,
-    "pubkey": <application-pubkey>,
+    "pubkey": <pubkey>,
     "content": "<optional-kind:0-style-metadata>",
     "tags": [
         [ "d", <random-id> ],
@@ -62,7 +64,7 @@ The third value of the tag SHOULD be the platform where this recommendation migh
 * `content` is an optional `set_metadata`-like stringified JSON object, as described in NIP-01. This content is useful when the pubkey creating the `kind:31990` is not an application. If `content` is empty, the `kind:0` of the pubkey should be used to display application information (e.g. name, picture, web, LUD16, etc.)
 
 * `k` tags' value is the event kind that is supported by this `kind:31990`.
-Multiple `k` tags can exist in the same event if the application supports more than one event kind.
+Multiple `k` tags can exist in the same event if the application supports more than one event kind and their handler URLs are the same.
 
 * `bech32` in a URL MUST be replaced by clients with the NIP-19-encoded entity that should be loaded by the application.
 
--- a/89.md
+++ b/89.md
@@ -6,19 +6,20 @@ Recommended Application Handlers
 
 `draft` `optional` `author:pablof7z`
 
-This NIP describes `kind:31989` and `kind:31990`: a way for unknown event-kinds application-handlers to be discovered.
+This NIP describes `kind:31989` and `kind:31990`: a way to discover applications that can handle unknown event-kinds.
 
 ## Rationale
-Nostr's discoverability and transparent event interaction is one of its most interesting/novel mechanics. This NIP seeks to provide the glue for cross-`kind` client interaction.
+Nostr's discoverability and transparent event interaction is one of its most interesting/novel mechanics.
+This NIP provides a simple way for clients to discover applications that handle events of a specific kind to ensure smooth cross-client and cross-kind interactions.
 
 ### Parties involved
 There are three actors to this workflow:
 
-* application that handles a specific event kind
+* application that handles a specific event kind (note that an application doesn't necessarily need to be a distinct entity and it could just be the same pubkey as user A)
     * Publishes `kind:31990`, detailing how apps should redirect to it
-* user who recommends an app that handles a specific event kind
+* user A, who recommends an app that handles a specific event kind
     * Publishes `kind:31989`
-* user who seeks a recommendation for an app that handles a specific event kind
+* user B, who seeks a recommendation for an app that handles a specific event kind
     * Queries for `kind:31989` and, based on results, queries for `kind:31990`
 
 # Events
@@ -30,19 +31,26 @@ There are three actors to this workflow:
     "pubkey": <recommender-user-pubkey>,
     "tags": [
         [ "d", <supported-event-kind> ],
-        [ "p", <app-pubkey> ]
+        [ "a", "31990:app1-pubkey:<d-identifier>", "wss://relay1", "ios" ],
+        [ "a", "31990:app2-pubkey:<d-identifier>", "wss://relay2", "web" ]
     ]
 }
 ```
+
+Multiple `a` tags can appear on on the same `kind:31989`.
+
+The second value of the tag SHOULD be a relay hint.
+The third value of the tag SHOULD be the platform where this recommendation might apply.
 
 ## Handler information
 ```json
 {
     "kind": 31990,
     "pubkey": <application-pubkey>,
+    "content": "<optional-kind:0-style-metadata>",
     "tags": [
-        [ "d", <supported-event-kind> ]
-        [ "web", "https://example.com/", "root" ],
+        [ "d", <random-id> ],
+        [ "k", <supported-event-kind> ],
         [ "web", "https://..../a/<bech32>", "nevent" ],
         [ "web", "https://..../p/<bech32>", "nprofile" ],
         [ "web", "https://..../e/<bech32>" ],
@@ -51,11 +59,14 @@ There are three actors to this workflow:
 }
 ```
 
-`bech32` in a URL MUST be replaced by clients with the NIP-19-encoded entity that should be loaded by the application.
+* `content` is an optional `set_metadata`-like stringified JSON object, as described in NIP-01. This content is useful when the pubkey creating the `kind:31990` is not an application. If `content` is empty, the `kind:0` of the pubkey should be used to display application information (e.g. name, picture, web, LUD16, etc.)
+
+* `k` tags' value is the event kind that is supported by this `kind:31990`.
+Multiple `k` tags can exist in the same event if the application supports more than one event kind.
+
+* `bech32` in a URL MUST be replaced by clients with the NIP-19-encoded entity that should be loaded by the application.
 
 Multiple tags might be registered by the app, following NIP-19 nomenclature as the second value of the array.
-
-A special `root` tag MIGHT be included with the main page of the application.
 
 A tag without a second value in the array SHOULD be considered a generic handler for any NIP-19 entity that is not handled by a different tag.
 
@@ -76,7 +87,7 @@ User A might be a user of Zapstr, a `kind:31337`-centric client (tracks). Using 
     "kind": 31989,
     "tags": [
         [ "d", "31337" ],
-        [ "p", "1743058db7078661b94aaf4286429d97ee5257d14a86d6bfa54cb0482b876fb0" ],
+        [ "a", "31990:1743058db7078661b94aaf4286429d97ee5257d14a86d6bfa54cb0482b876fb0:abcd", <relay-url>, "web" ]
     ]
 }
 ```
@@ -89,19 +100,13 @@ User B's client, not knowing how to handle a `kind:31337` might display the even
 using its `alt` tag (as described in NIP-31). When the user clicks on the event,
 the application queries for a handler for this `kind`:
 
-`["REQ", <id>, '[{ "kinds": [31989], "#d": [31337], 'authors': [<user>, <users-contact-list>] }]']`
+`["REQ", <id>, '[{ "kinds": [31989], "#d": ["31337"], 'authors': [<user>, <users-contact-list>] }]']`
 
-User B, who follows User A, sees that `kind:31989` event and fetches the `p`-tagged profile for handler information.
-
-```json
-{
-    "kinds": [31990],
-    "#d": [31337],
-    "authors": ["1743058db7078661b94aaf4286429d97ee5257d14a86d6bfa54cb0482b876fb0"]
-}
-```
+User B, who follows User A, sees that `kind:31989` event and fetches the `a`-tagged event for the app and handler information.
 
 User B's client sees the application's `kind:31990` which includes the information to redirect the user to the relevant URL with the desired entity replaced in the URL.
 
 ## Alternative query bypassing `kind:31989`
 Alternatively, users might choose to query directly for `kind:31990` for an event kind. Clients SHOULD be careful doing this and use spam-prevention mechanisms to avoid directing users to malicious handlers.
+
+`["REQ", <id>, '[{ "kinds": [31990], "#k": [<desired-event-kind>], 'authors': [...] }]']`
--- a/89.md
+++ b/89.md
@@ -64,7 +64,9 @@ The third value of the tag SHOULD be the platform where this recommendation migh
 * `content` is an optional `set_metadata`-like stringified JSON object, as described in NIP-01. This content is useful when the pubkey creating the `kind:31990` is not an application. If `content` is empty, the `kind:0` of the pubkey should be used to display application information (e.g. name, picture, web, LUD16, etc.)
 
 * `k` tags' value is the event kind that is supported by this `kind:31990`.
-Multiple `k` tags can exist in the same event if the application supports more than one event kind and their handler URLs are the same.
+Using a `k` tag(s) (instead of having the kind onf the NIP-33 `d` tag) provides:
+    * Multiple `k` tags can exist in the same event if the application supports more than one event kind and their handler URLs are the same.
+    * The same pubkey can have multiple events with different apps that handle the same event kind.
 
 * `bech32` in a URL MUST be replaced by clients with the NIP-19-encoded entity that should be loaded by the application.
 


arthurfranca:
How about we bring relays to the table too? I mean, maybe my friends have not set any recommendation (although, yes, clients can try to create such recommendation events on behalf of their users).

If relays could advertise clients it wants, so to draw traffic to his own/partner's clients, it is one more incentive to run a free relay.

Like you could add to the NIP that for relays to participate, they can make available kind 31989 events with a specific pubkey (for instance one letter pubkey 'g' so to not be hex, and its ok one char as from nip-01 its allowed a prefix/incomplete pubkey for querying).

We can imagine that relay implementations like nostream would set default list of recommended clients for many kinds, which would help with reach.

staab:
Could labels as described in https://github.com/nostr-protocol/nips/pull/524 be used to support the same use case? I don't necessarily think they should, but that PR introduces a primitive that allows for generic recommendation/review events, which can be applied to relays as written. Introducing a `c` tag could pretty easily add support for client recommendations/reviews as well.

pablof7z:
if a user wants to trust a relay to provide these recommendations, they can follow the relay's pubkey, or clients could implement querying the relay's pubkey regardless of whether the user is following it or not, but I don't think this needs (nor should) be a particular case handled differently.

arthurfranca:
> if a user wants to trust a relay to provide these recommendations, they can follow the relay's pubkey, or clients could implement querying the relay's pubkey regardless [...]

@pablof7z This brings friction. The user would have to manually search and follow the relay's pubkey or the client would have to code http connection (instead of ws) to get Relay Information Document to find out relay's pubkey.

I mean, If the relay is already configured on user's client ([NIP-65](65.md)), it means the user already trusts the relay enough.

Considering that the client is already connected to the relay, the only thing it needs is show intent `["REQ", <id>, '{ ..., p: ['g', <my friend 1>, <my friend 2>] }]`. With a `g` pubkey always meaning "currently connected relay's pubkey", the relay doesn't even need to bother having a real pubkey/nostr account.

brugeman:
> Could labels as described in #524 be used to support the same use case? I don't necessarily think they should, but that PR introduces a primitive that allows for generic recommendation/review events, which can be applied to relays as written. Introducing a `c` tag could pretty easily add support for client recommendations/reviews as well.

We want an efficient way to query for "handlers of kind X recommended by a set of pubkeys" for WoT check. Labels as suggested don't seem to provide that. And then to do a good WoT check I need an opinion of every person I trust, and if they only have one event w/ all their recomms then I can query for them with one REQ (authors:[a,b,c]). But with labels, a single spammy person would flood a single REQ with their opinions, I would have to do REQ per person (authors:[a], then authors:[b]...) to get a good sample of recomms. These are sort of technical issues and conceptually labels are the same thing, but we want to solve a very specific need (WoT recomms for apps), and generic labels as described in #524 don't seem to work well for that.

Does this make sense?

brugeman:
> How about we bring relays to the table too? I mean, maybe my friends have not set any recommendation (although, yes, clients can try to create such recommendation events on behalf of their users).

> If relays could advertise clients it wants, so to draw traffic to his own/partner's clients, it is one more incentive to run a free relay.

@staab you had some NIP about relays recommending other relays/services? I'd say the above idea should go there, wdyt?

89.md
```diff
@@ -0,0 +1,107 @@
...
+
+When user B sees an unknown event kind, e.g. in a social-media centric nostr client, the client would allow user B to interact with the unknown-kind event (e.g. tapping on it).
+
+The client MIGHT query for the user's and the user's follows handler.
+
+# Example
+
+## User A recommends a `kind:31337`-handler
+User A might be a user of Zapstr, a `kind:31337`-centric client (tracks). Using Zapstr, user A publishes an event recommending Zapstr as a `kind:31337`-handler.
+
```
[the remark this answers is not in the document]
brugeman: One random idea is that maybe signing apps (Alby etc) could have a blacklist of kinds such that if an app asks to use it, they'd ask for user confirmation every single time. At least it would add some friction and would make it obvious to the first users so they would raise the alarms.

89.md
```diff
@@ -0,0 +1,107 @@
...
+```
+
+## Handler information
+```json
+{
+    "kind": 31990,
+    "pubkey": <application-pubkey>,
+    "tags": [
+        [ "d", <supported-event-kind> ]
+        [ "web", "https://example.com/", "root" ],
```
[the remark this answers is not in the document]
pablof7z: yeah, but then you lose the querying ability, which is the most interesting part of this; what's the point of having a range if you are only going to be able to query for one of them?

imo what's interesting is discovering apps that are kind-specific, not apps that provide a generic interface to all kinds of events like nostr.guru.

I think if an app supports multiple kinds it should create one `kind:31989` per kind it supports, which dissuades apps from advertising they support 10k different event kinds.

What I have in mind is, an app that is GREAT at supporting event `kind:X` can be discovered.

89.md
```diff
@@ -0,0 +1,107 @@
...
+        [ "web", "https://..../a/<bech32>", "nevent" ],
+        [ "web", "https://..../p/<bech32>", "nprofile" ],
+        [ "web", "https://..../e/<bech32>" ],
+        [ "ios", ".../<bech32>" ]
+    ]
```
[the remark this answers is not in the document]
pablof7z: yeah, this is the part I was the most uncertain of as I'm not familiar with how these platforms are defined but could intuit that there would be a formal definition of this. 100% this should follow whatever standard there is to refer to these platforms.

89.md
```diff
@@ -0,0 +1,107 @@
...
+    "kinds": [31990],
+    "#d": [31337],
+    "authors": ["1743058db7078661b94aaf4286429d97ee5257d14a86d6bfa54cb0482b876fb0"]
+}
+```
+
+User B's client sees the application's `kind:31990` which includes the information to redirect the user to the relevant URL with the desired entity replaced in the URL.
+
+## Alternative query bypassing `kind:31989`
+Alternatively, users might choose to query directly for `kind:31990` for an event kind. Clients SHOULD be careful doing this and use spam-prevention mechanisms to avoid directing users to malicious handlers.
```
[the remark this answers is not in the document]
pablof7z: the `kind` discovery is the interesting thing here, a range can't be queried on relays.

A client-side can perfectly well use trust hints based on the user's follows, and perhaps show a shortlist of app handlers it came up with for the user to choose from; a WoT-type algorithm could be used, but it's not required.

89.md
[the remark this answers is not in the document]
brugeman: Ok I agree that maybe all-kinds-apps aren't in scope for this.

89.md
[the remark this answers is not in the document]
pablof7z: this is just like nostr: we don't need to solve this in the same way nostr chooses not to solve for DIDs and stuff like that; we can let the free market decide; apps that are abusive will naturally be shunned and clients/users can choose to blacklist the recommendation of apps.

Ultimately, all nostr events are signed by the user's key, which means is the users' responsibility, in the same way that you might give a friend your email address and a virus in their computer spam the hell out of you, it's up to you and the user to figure out, not a concern of SMTP whom/how you should give your email address to 😂

89.md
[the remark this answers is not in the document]
fiatjaf: Global is still useful if you're doing it on filtered relays.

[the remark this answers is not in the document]
pablof7z: ultimately any "trust" heuristic (trust rank, WoT, filtered relays, human judgement, whatever) we come up with will apply to this recommendations as well

fiatjaf:
The only thing I don't like here is the "application pubkey" part. What does that mean? To be considered as an app an application must have its own public key and publish an event with their handlers? What if they don't have one but I still want to recommend it? Since it's me that am recommending shouldn't it be me creating the handler templates for the application?

Can both models can coexist?

arthurfranca:
> I think you're referring to [this](https://github.com/nostr-protocol/nips/pull/457#issuecomment-1544574452)?

@staab I think it is about your [NIP-11#extensions](https://github.com/nostr-protocol/nips/pull/259/files) PR. Although it is currently related to recommending relays (to handle a NIP) instead of recommending apps/clients (to handle a kind).

The downside of using NIP-11 "Relay Information Document" is that it is http instead of websocket. So the client must code the additional request/response logic and integrate to the kind-handler recommendation fetcher code (this NIP, that uses ws).

Also, if using NIP-11 just to learn the relay's pubkey, the relay need to manage a nostr account (store secret key somewhere).

That's why I suggested using a special purpose `g` pubkey (that can't be a prefix of any real pubkey cause it isn't `/^[0-9A-F]{1,64}$/i`) that could be simply added to the same websocket `REQ` filter ("authors" filter). But I'm just trying to better explain the idea... not trying to force anything.

staab:
No, that's a good call. I definitely want to see that functionality, and having it in the relay doc doesn't support more dynamic/granular usage. This NIP handles specific kind support and is distributed, so yeah, I'd love to see relay recommendations folded into this if possible.

fiatjaf:
@brugeman the only issue I see is that many apps may not bother with having their own keys. I certainly would see that as an annoying hassle for some small stupid apps I made, so I imagine many people wouldn't want to do it at all.

I was thinking I could create these events with my own keys in these cases, which works fine, I think, we add a `d` tag and... hmm, now that I am looking we're already using the `d` tag for the kind. This is getting weird.

pablof7z:
> @brugeman the only issue I see is that many apps may not bother with having their own keys. I certainly would see that as an annoying hassle for some small stupid apps I made, so I imagine many people wouldn't want to do it at all.
> 
> I was thinking I could create these events with my own keys in these cases, which works fine, I think, we add a `d` tag and... hmm, now that I am looking we're already using the `d` tag for the kind. This is getting weird.

Just use your key to create both events, the 31990 event also signed with your key and not a dedicated app-key.

pablof7z:
> @pablof7z Have you thought about having the URL handler templates specified on a HTML page or `/.well-known/` HTTP response in the app URL? If we're talking about **URLs** then it's logical to think the apps have websites.

what would be gained from the extra roundtrip though? 🤔

fiatjaf:
> what would be gained from the extra roundtrip though? 

I don't know, that's the way our ancestors would do this. But I see three immediate advantages:

1. It's very simple for app developers to provide and update URL templates for their apps, it fits in their workflow very nicely.
2. Relays don't have to be bothered, we don't have to wonder in what relay a given event is or is not.
3. It's very straightforward for app recommenders to get consensus.

On point 3: if I'm recommending an app, how do I know where is that app's 31990 event? How do I decide if that app has one or not? If everybody follows your suggested strategy above ("Just use your key to create both events, the 31990 event also signed with your key and not a dedicated app-key") we'll be left in a situation with a lot of data duplication -- people querying recommendations will have to do a lot of work to get a lot of crappy data in response (in this case it is actually better to put all the handlers in the 31989 event).

On the other hand if the 31989 event is just a URL then it's simple, people querying for recommendations will get a bunch of responses pointing to the same URL.

fiatjaf:
On the other hand this won't work at all with applications that don't do the URL templates themselves.

So what about we have just the 31989 event and it includes the URL, but also optionally the handlers in itself?

brugeman:
@fiatjaf 

> @brugeman the only issue I see is that many apps may not bother with having their own keys. I certainly would see that as an annoying hassle for some small stupid apps I made, so I imagine many people wouldn't want to do it at all.

Honestly I don't think it makes sense to have stupid apps/prototypes recommended at all. If an app wants to be auto-discovered, it should put some work to enable good UX for users who see this app for the first time. 

>  @pablof7z Have you thought about having the URL handler templates specified on a HTML page or /.well-known/ HTTP response in the app URL? If we're talking about URLs then it's logical to think the apps have websites.

nostr:xxx were also URLs in my head (well some of them are URIs I guess), so a mobile/desktop app doesn't have to have a website set up, and forcing them to host .well-known seems like a bigger hassle then generating an app pubkey. We should probably mention that nostr: handlers are the way to go for non-web apps.

> Relays don't have to be bothered, we don't have to wonder in what relay a given event is or is not.

31989 should have relay after the app pubkey in tags, we should mention that in the NIP. The fact that it's uncertain is the inherent nostr problem, we should solve it elsewhere, this case isn't special.

> On point 3: if I'm recommending an app, how do I know where is that app's 31990 event? 

From the relay hint in 31989.

> How do I decide if that app has one or not? 

It should be a MUST per this NIP if app want's to be auto-discovered.

> If everybody follows your suggested strategy above ("Just use your key to create both events, the 31990 event also signed with your key and not a dedicated app-key") we'll be left in a situation with a lot of data duplication -- people querying recommendations will have to do a lot of work to get a lot of crappy data in response (in this case it is actually better to put all the handlers in the 31989 event).

That's a bad suggestion, @pablof7z . How should I present this new app to a user? I was going to get app's kind 0 to get name, picture, description (would fetch it along with 31990), but if app doesn't have it's own pubkey then how should I get that info? Just showing people a url makes zero sense. 

And putting a .well-known url to 31989 and expecting that thing to host the app info still feels ugly - we have a good place for that info (kind0), most big apps already have their profiles set up, why force them to keep the same info updated in several places?

So I propose to mention in the NIP:
1. To have auto-discovery, apps should have their own pubkeys, and app info in kind0, and handlers in kind31990.
2. Apps should add themselves to users' 31989 and should specify the relay along with their pubkey. 
3. Apps should prefer nostr: URLs in their 31990.

WDYT?

fiatjaf:
> Honestly I don't think it makes sense to have stupid apps/prototypes recommended at all.

This is irrelevant. People have different opinions. We're just trying to come up with the best standard for recommendations here, one that is open for everybody to use.

> nostr:xxx were also URLs in my head (well some of them are URIs I guess), so a mobile/desktop app doesn't have to have a website set up

Mobile and desktop apps will necessarily always have to handle URLs in the format `nostr:<nip19code>` so for these it's already straightforward. They don't have to have websites at all.

@brugeman you didn't address the fundamental question:

what is easier for web apps: to have a page describing its own routes (which they can create in one minute and forget about until some route changes), or to a create and keep track of a standalone secret key forever?

brugeman:
> > Honestly I don't think it makes sense to have stupid apps/prototypes recommended at all.
> 
> This is irrelevant. People have different opinions. We're just trying to come up with the best standard for recommendations here, one that is open for everybody to use.

A pubkey is more open than url, why should I maintain a url somewhere if there is an open network that stores stuff for me? 

> 
> > nostr:xxx were also URLs in my head (well some of them are URIs I guess), so a mobile/desktop app doesn't have to have a website set up
> 
> Mobile and desktop apps will necessarily always have to handle URLs in the format `nostr:<nip19code>` so for these it's already straightforward. They don't have to have websites at all.

That's why we shouldn't force them to have .well-known url.

> 
> @brugeman you didn't address the fundamental question:
> 
> what is easier for web apps: to have a page describing its own routes (which they can create in one minute and forget about until some route changes), or to a create and keep track of a standalone secret key forever?

Maybe 10% easier - no need to copy the secret to your key store, but I disagree that this is a fundamental question. If keys are so hard, why build nostr - we could just use websites? If app is represented by a URL, then it's second-class citizen on nostr. I can't zap it, can't mention it, can't DM it. We should either make app an event or a profile, making it a url is a "convenience" for web apps today, and an inconvenience for everyone forever.

fiatjaf:
> Maybe 10% easier - no need to copy the secret to your key store, but I disagree that this is a fundamental question. If keys are so hard, why build nostr - we could just use websites? If app is represented by a URL, then it's second-class citizen on nostr. I can't zap it, can't mention it, can't DM it. We should either make app an event or a profile, making it a url is a "convenience" for web apps today, and an inconvenience for everyone forever.

We've had this discussion already and I don't see a way to resolve it. The entire point sums up to you believing in the ideal world of everything-p2p in which we don't need any of the current internet's infrastructure. Nostr's fundamental design is about not trying to do everything p2p and relying on URLs and servers, so I can only see a contradiction in you liking Nostr and at the same time hating its core design.

fiatjaf:
> A pubkey is more open than url, why should I maintain a url somewhere if there is an open network that stores stuff for me?

Because you are a web app with a URL! Aren't we talking about URL handler templates? Aren't these dependent on the existence of URLs?

> That's why we shouldn't force them to have .well-known url.

I'm not trying to force them. If they are native apps they don't need that at all, the NIP should make this clear. But they also don't need a keypair.

fiatjaf:
@brugeman It seems like you're thinking of a future in which apps will not have URLs, but will be served as Nostr events directly, somehow. I think this is cool, but it is not the reality today.

In any case, we can still divide apps in three categories:

1. webapps -- apps that have URLs and require URL path templates
2. native apps -- apps that are expected to handle `nostr:<nip19>` URLs no matter what
3. nostr-apps -- apps that are hosted inside Nostr (these do not exist today)

Kind 31989 notes can account for these 3 categories:

1. for webapps they can just point to the root URL of the app and the app can have a  `/.well-known/` resource talking about its own handlers -- or the recommender can specify the path templates themselves, as a fallback.
2. for native apps, the recommendation can contain a website (or a URL from which the app can be downloaded, I'm not sure about what is the best way to _link_ to a native app), but it's implied that the app will handle `nostr:<nip19>` URLs in any case.
3. for nostr-apps I don't know, probably link to their pubkey? Or maybe their `naddr`? We'll have to wait until these exist to decide.

In any case the recommendation may mention a pubkey for the app, but this should not be a requirement.

brugeman:
@fiatjaf this is very constructive, thank you. 

> @brugeman It seems like you're thinking of a future in which apps will not have URLs, but will be served as Nostr events directly, somehow. I think this is cool, but it is not the reality today.

I think apps need an "app description" (not talking about "handlers", but name, picture, description), which I think are better off being a nostr event. Without description, I can't suggest an app to the user. Where the app is served from when user decides to launch it resolves from the url handlers that it provides (and it might not be nostr:, at least for mobile, see below). 

> 
> In any case, we can still divide apps in three categories:
> 
> 1. webapps -- apps that have URLs and require URL path templates
> 2. native apps -- apps that are expected to handle `nostr:<nip19>` URLs no matter what
> 3. nostr-apps -- apps that are hosted inside Nostr (these do not exist today)
> 
> Kind 31989 notes can account for these 3 categories:
> 
> 1. for webapps they can just point to the root URL of the app and the app can have a  `/.well-known/` resource talking about its own handlers -- or the recommender can specify the path templates themselves, as a fallback.

That's right. But I need a name, picture, description, which web apps of course could host in well-known.

> 2. for native apps, the recommendation can contain a website (or a URL from which the app can be downloaded, I'm not sure about what is the best way to _link_ to a native app), but it's implied that the app will handle `nostr:<nip19>` URLs in any case.

For native apps I still need a description, should they also host it as well-known? Also, should they also have a root-url serving as an identity? Bcs without an "identity" of this app, it's hard to provide WoT check (this app is used by 10 friends). 

How native apps are downloaded and/or launched is a great question. In theory Android has App Links and iOS has Universal Links which could drive the user through app store->install->specific screen that handles the url, I guess we should consult iOS/Android/win/linux experts here, maybe they'd have some input. 

> 3. for nostr-apps I don't know, probably link to their pubkey? Or maybe their `naddr`? We'll have to wait until these exist to decide.

I like your pure-nostr-apps idea, I wasn't thinking about anything like that. All I imagined were new app apks distributed as special nostr events w/ links to the apk. But I digress. 

> 
> In any case the recommendation may mention a pubkey for the app, but this should not be a requirement.

If we address the "app identity" and "app description" questions for native apps then I'm fine w/ optional app pubkeys.

arthurfranca:
> 2. Apps should add themselves to users' 31989 and [...]

@brugeman whoaaaaa! I thought clients generating these events on behalf of the user (with users pubkeys) would be an unintended side-effect of using clients that are evil behind the curtains. Now I see it is the intended feature =o in practice this won't be recommendation from my friends, but from my friends' clients.

This kinda is incentive (a bad one) for some devs creating free clients of specific domains (domains capable of bringing big user bases). Just to then generate thoundands to millions of 31989 events (with their purpose-specific app recommendatins they intend to make money with) and broadcast these events to all relays it knows or atleast to user configured write relays. Then other clients will end up using said recommended apps when seeing an embedded event they can't load.

I like my idea of empowering relays better (the `g` service relay pubkey) xD Specially because there won't be all these events being broadcasted and also cause I suppose we need to add more incentives to running **free** relays. Free relays are what make it possible for anyone to choose nostr instead of regular web to create apps without high storage/db costs.

brugeman:
> > 2. Apps should add themselves to users' 31989 and [...]
> 
> @brugeman whoaaaaa! I thought clients generating these events on behalf of the user (with users pubkeys) would be an unintended side-effect of using clients that are evil behind the curtains. Now I see it is the intended feature =o in practice this won't be recommendation from my friends, but from my friends' clients.
> 

This is addressed here: https://github.com/nostr-protocol/nips/pull/530/files#r1193530209

brugeman:
@fiatjaf @pablof7z we could do this:

- 31989 recomm event can contain either 'p' - app pubkey or an 'r' - url to a json
- if recomm event has 'r', the URL is treated as app identity for WoT, that json should have two fields - "app" and "handlers", app contains name, picture etc as in kind0 format, handlers contains an array of tags as in 31990 in this nip.

This is a way to make both models work like was suggested. WDYT?

89.md
```diff
@@ -0,0 +1,107 @@
...
+
+## Recommendation event
+```json
+{
+    "kind": 31989,
+    "pubkey": <recommender-user-pubkey>,
+    "tags": [
+        [ "d", <supported-event-kind> ],
+        [ "p", <app-pubkey> ]
+    ]
```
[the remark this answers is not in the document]
erskingardner: Agree, guessing the examples are that way for simplicity, but the idea of recommending multiple apps for a given event type makes sense.

erskingardner:
Finally had a chance to catch up on this (thanks @fiatjaf for [the bump](https://snort.social/e/note1m8twrfglzekpc83hwzv8ygtapava7jlcaq7w8ra8flyp7g4crrds95x4fm))

Apologies for the length, but hopefully a recap for future readers can help make sure we're on the same page:

### The Goals of this NIP
Make it easier for clients to discover the right app(s) to handle a _specific_ event `kind`. The focus is on specific, and not ranges of events, so that it's possible to query relays for the `kind` in question. This is done via:
1. Breadcrumbs "recommendations" published by users (kind 31989 events) that help other clients to find the right/best app to handle a given `kind`.
2. A canonical event (kind 31990) that allows anyone to publish the URL/URI scheme of the app in question to allow for deep linking into a specific resource/page of the app. Ideally these are published by the app itself.

### Risks Identified
1. Some apps might not want/need to have their own keys, which makes publishing a canonical/first-party 31990 event impossible.
2. SPAM: it's possible for bad actors to flood relays with 31989 events pointing to their own apps (which might be malicious). This then leaves relays or clients to sort out who is good/bad and implement blacklists, etc.

### Something no one has asked
1. If we're having users publish recommendations, why wouldn't we include more/higher-signal data in the recommendation? A numeric rating or a text review or something else. Everyone seems to agree on the spam risk so why wouldn't we think about giving relays and clients more data points/signal to help them sift through. Not that rating and review content isn't easily fakeable but maybe there is something better to include...


### My thoughts, as an app dev and user.

1. Almost nothing is _required_ in Nostr. I think it's completely reasonable to have several optional things that apps can do to improve UX and discoverability. All the incentives align well here.
    1. I think it's safe to assume that, long-term, all apps of any importance will have their own keys. And those apps will want to do what they can to promote themselves and improve discoverability/UX. Think how much effort businesses currently go to with SEO or google page rank improvements.
    2. I think it's also reasonable that, while it's slightly backwards from the future direction we want Nostr to go, it's totally fine to give apps the option of publishing a `.well-known` style doc detailing the same information that they publish in their `31990` event. This feels like a fallback option that clients can check only if they can't find a `31990` event or for whatever other reason. The difference between a WS req or a HTTP roundtrip feels trivial to me and not worth splitting hairs over.
    4. @brugeman's point about how how to show users info about the apps that they might suggest for a specific event type; again, apps of any importance will take the time to create a kind `0` event. Those that don't can expect to not be discovered as easily. 
2. Ok – now for the anti-patterns.
    1. I don't think that we can or should try and control spam. That said, there is something really unsatisfying about the incentives here. This system of recommending apps is completely based on the premise that we can build good enough heuristics (and maintain them through time) to overcome a near infinite quantity of spam. I tend to agree with @pablof7z that this isn't the protocol's problem but, if the protocol can help avoid it, I do believe we should try.
    2. How are apps going to ask users to publish these kind 31989 recommendations? Is this going to become the new app store review dialog that plagues literally every mobile app? And if so, should these recommendations contain any other potentially helpful information? Ratings, text content reviews?
    3. Another anti-pattern here is that I can imagine most apps will do very basic/cursory queries for these recommendations based on a small subset of relays. Back to the heuristics question of how do you quickly build conviction that you have the right app(s) for handling a specific kind that you can then show users? This is a lot of work and I can see lots of clients stopping at the simplest implementation. This leaves the door open for a lot of abuse that wouldn't be obvious to end-users. I could see a more centralized verification/indexing service pop up around doing this work...is that a centralizing force we want to avoid?

pablof7z:
> SPAM: it's possible for bad actors to flood relays with 31989 events pointing to their own apps (which might be malicious). This then leaves relays or clients to sort out who is good/bad and implement blacklists, etc.

This is trivially solvable by the recommended default method of querying `kind:31989` from the users' follow list

erskingardner:
> This is trivially solvable by the recommended default method of querying kind:31989 from the users' follow list

@pablof7z Certainly that solves it in the short term but I think longer-term it will breakdown and we'll need more complex heuristics.

pablof7z:
> If we're having users publish recommendations, why wouldn't we include more/higher-signal data in the recommendation? A numeric rating or a text review or something else. Everyone seems to agree on the spam risk so why wouldn't we think about giving relays and clients more data points/signal to help them sift through. Not that rating and review content isn't easily fakeable but maybe there is something better to include...

Out of scope. This kind of thing should be solved in a more generic way and not be pushed to each "thing" (e.g. `kind:31989` in this case).

pablof7z:
> > This is trivially solvable by the recommended default method of querying kind:31989 from the users' follow list
> 
> @pablof7z Certainly that solves it in the short term but I think longer-term it will breakdown and we'll need more complex heuristics.

Why?

erskingardner:
> Out of scope. This kind of thing should be solved in a more generic way and not be pushed to each "thing" (e.g. kind:31989 in this case).

I agree, but wanted to flag it so 1) people knew it came up and 2) to double check what other's thoughts were.

pablof7z:
> I think it's also reasonable that, while it's slightly backwards from the future direction we want Nostr to go, it's totally fine to give apps the option of publishing a .well-known style doc detailing the same information that they publish in their 31990 event. This feels like a fallback option that clients can check only if they can't find a 31990 event or for whatever other reason. The difference between a WS req or a HTTP roundtrip feels trivial to me and not worth splitting hairs over.

I don't have a strong opinion either way so I'm fine with either but not both (why implement the same thing in two different ways).

To me, the in-nostr event `kind:31990` *feels* better, but again, I don't have a strong opinion on this. The strongest argument against it I've seen so far is that a client might not find the `31990`, which is a general problem with nostr that gossip-style and kind-specific relays would solve (i.e. a purplepag.es for `kind:31990`)

A positive thing about doing in-protocol is that `kind:31990` could be referred for reviews of the app, which would be much more fault tolerant that doing reviews of an `r` tag.

[**off-topic**: github not having threaded discussions of PRs is absolute garbage; this discussion is confusing as fuck because you can't really respond in thread -- can't wait for nostr to support PR discussions..)

erskingardner:
> Why?

Maybe i'm over-indexing on solving future problems here. But I think unless client developers are thoughtful about how they are finding and presenting "options", it's very likely that you'll get situations where users are presented with two seemingly legit options where one is malicious. TBH, I'm more than fine not changing anything with regards to this now – just wanted to consider the likely long-term direction a bit.

pablof7z:
> > Why?
> 
> Maybe i'm over-indexing on solving future problems here. But I think unless client developers are thoughtful about how they are finding and presenting "options", it's very likely that you'll get situations where users are presented with two seemingly legit options where one is malicious. TBH, I'm more than fine not changing anything with regards to this now – just wanted to consider the likely long-term direction a bit.

This is a non-problem imo; if two seemingly identical options are presented, one will likely have more recommendations from my follow list; just present the one with the most recommendations and always display the URL you are actually pointing to.

This is just like how Snort currently autocompletes with two "Jeff Booths" regardless of the fact that I only follow one of them and that my whole social graph points to one of them. This is trivial and obvious and this is not specific to `kind:31989`s, this is something all clients must be better at across the board.

pablof7z:
> brugeman's point about how how to show users info about the apps that they might suggest for a specific event type; again, apps of any importance will take the time to create a kind 0 event. Those that don't can expect to not be discovered as easily.

furthermore, wrt to using `r` tags; this NIP is to recommend apps that support a specific nostr kind. It follows that the app is nostr-aware, so they can easily be expected to publish a 31990. I think kind-specific nostr developers who want to develop a nostr app and don't want to be bothered with creating a key for the app + publish a 31990 will be a small minority and, even if that's the case, a user who likes the app can just create the 31990 on the behalf of the developer/app.

pablof7z:
> I don't think that we can or should try and control spam. That said, there is something really unsatisfying about the incentives here. This system of recommending apps is completely based on the premise that we can build good enough heuristics (and maintain them through time) to overcome a near infinite quantity of spam. I tend to agree with @pablof7z that this isn't the protocol's problem but, if the protocol can help avoid it, I do believe we should try.

Just to make sure this doesn't look like it went unanswered 👇

https://github.com/nostr-protocol/nips/pull/530#issuecomment-1566918079

i.e. this is a non-problem, in the same way all nostr apps default to `REQ`, `{"kinds":[1], "authors": [<your-follow-list>]}` apps, don't default to querying for all `kind:1` without a follow-list scope and display the `kind:1`s nudging the user to think they are seeing high quality/relevant content.

erskingardner:
yeah I think these are all fair points – especially with regards to Github PR discussions being a shitcoin. 😂 

If I had to sum up your perspective here, it's that we should optimize for "now" with something that is single-use focused and clear, rather than thinking too much about possible future problems or trying to solve for issues that are even slightly different from "As a client, how do I find what apps I might direct a user to in order to handle a specific kind of event that I don't natively support."

89.md
```diff
@@ -0,0 +1,107 @@
+NIP-89
+======
+
+Recommended Application Handlers
+--------------------------------
+
+`draft` `optional` `author:pablof7z`
+
+This NIP describes `kind:31989` and `kind:31990`: a way for unknown event-kinds application-handlers to be discovered.
```
erskingardner, line 9: "A way to discover applications that can handle unknown event-kinds."

89.md
```diff
@@ -0,0 +1,107 @@
...
+
+## Rationale
+Nostr's discoverability and transparent event interaction is one of its most interesting/novel mechanics. This NIP seeks to provide the glue for cross-`kind` client interaction.
```
erskingardner, line 12: This NIP provides a simple way for clients to discover applications that handle events of a specific `kind` to ensure smooth cross-client and cross-`kind` interactions.

erskingardner (commented):
two tiny language tweaks to make it touch more clear.

erskingardner (commented):
two tiny language tweaks to make it touch more clear.

pablof7z:
> Another anti-pattern here is that I can imagine most apps will do very basic/cursory queries for these recommendations based on a small subset of relays. Back to the heuristics question of how do you quickly build conviction that you have the right app(s) for handling a specific kind that you can then show users? This is a lot of work and I can see lots of clients stopping at the simplest implementation. This leaves the door open for a lot of abuse that wouldn't be obvious to end-users. I could see a more centralized verification/indexing service pop up around doing this work...is that a centralizing force we want to avoid?

This is the raison d'être of NDK: naive clients will always present a substandard view/experience of nostr. I think this is so obvious that all clients will end up adopting a gossip-style implementation. This is not specific to this NIP.

Also, kind-specific relays like purplepag.es does for `kind:0,3,10002` could very easily aid in solving this.

brugeman:
> The strongest argument against it I've seen so far is that a client might not find the `31990`, which is a general problem with nostr that gossip-style and kind-specific relays would solve (i.e. a purplepag.es for `kind:31990`)

I think fiatjaf's (major?) point was that managing keys for small apps is a burden. https://github.com/nostr-protocol/nips/pull/530#issuecomment-1549425462

brugeman:
> Another anti-pattern here is that I can imagine most apps will do very basic/cursory queries for these recommendations based on a small subset of relays. Back to the heuristics question of how do you quickly build conviction that you have the right app(s) for handling a specific kind that you can then show users? This is a lot of work and I can see lots of clients stopping at the simplest implementation. This leaves the door open for a lot of abuse that wouldn't be obvious to end-users. I could see a more centralized verification/indexing service pop up around doing this work...is that a centralizing force we want to avoid?

This is the raison d'être of nostrapp.link - not only will it allow you to manage the 31989 that you've published, but also could serve as a solution for small apps that don't bother with proper app discovery logic - just redirect user to nostrapp.link/#bech32 and it will handle things properly, including - redirect user straight to their app of choice if it was saved already.

erskingardner:
After thinking about it most of the day, my vote would be to merge this PR as is. I think most of what I was voicing is probably over-optimizing for where we are now. I'd love to see how this starts to work in the wild.

pablof7z:
updated the PR:

* `kind:31989` no longer uses `p`; it directly `a` tags the event with the handler information.
* `kind:31990` no longer depends on the `d` tag expressing the supported event kind (because of the previous point making it explicit):
  * now `kind:31990` uses one ore more `k` tags to describe the event kinds it supports
* `kind:31990` can now include a `kind:0`-like `content` describing the metadata of the application (name, description, lud16, etc)
  * this obviates the need of requiring an application-specific set of keys to publish the `kind:31990` or having to store the handler information on a URL. (e.g. @fiatjaf can publish any application without his own pubkey without storing new keys) and nostr clients don't need to support two different ways of fetching this information

89.md
```diff
@@ -0,0 +1,112 @@
...
+        [ "a", "31990:app2-pubkey:<d-identifier>", "wss://relay2", "web" ]
+    ]
+}
+```
+
+Multiple `a` tags can appear on on the same `kind:31989`.
+
+The second value of the tag SHOULD be a relay hint.
+The third value of the tag SHOULD be the platform where this recommendation might apply.
+
```
brugeman, line 44: Could we interpret the third value as "user A is using this app on this platform", instead of "user A recommends this app for this platform"? I think recommendations/endorsements should be a separate thing like nip-32.

89.md
```diff
@@ -0,0 +1,112 @@
...
+## Handler information
+```json
+{
+    "kind": 31990,
+    "pubkey": <application-pubkey>,
```
brugeman, line 49: Not necessarily <application-pubkey>, could be user A pubkey.

89.md
```diff
@@ -0,0 +1,112 @@
...
+        [ "web", "https://..../e/<bech32>" ],
+        [ "ios", ".../<bech32>" ]
+    ]
+}
+```
+
+* `content` is an optional `set_metadata`-like stringified JSON object, as described in NIP-01. This content is useful when the pubkey creating the `kind:31990` is not an application. If `content` is empty, the `kind:0` of the pubkey should be used to display application information (e.g. name, picture, web, LUD16, etc.)
+
+* `k` tags' value is the event kind that is supported by this `kind:31990`.
+Multiple `k` tags can exist in the same event if the application supports more than one event kind.
```
brugeman, line 65: Should we say:
"Multiple `k` tags can exist in the same event if the application supports more than one event kind and their `handler` URLs are the same."
?

89.md
brugeman: Also, if I'm using this app on several platforms, can I use forth and fifth value for "a" tag? Or maybe comma-separated list of platforms as third value?

89.md
```diff
@@ -0,0 +1,112 @@
...
+
+# Events
+
+## Recommendation event
+```json
+{
+    "kind": 31989,
+    "pubkey": <recommender-user-pubkey>,
+    "tags": [
+        [ "d", <supported-event-kind> ],
```
brugeman, line 33: We could comment that we're using d tag as kind, instead of separate k tag, so make sure user only has on app list per kind. Or, if it's not our intention, then we should consider not assigning special meaning to d-tag and using k tag instead.

89.md
```diff
@@ -0,0 +1,112 @@
...
+
+* `k` tags' value is the event kind that is supported by this `kind:31990`.
+Multiple `k` tags can exist in the same event if the application supports more than one event kind.
+
+* `bech32` in a URL MUST be replaced by clients with the NIP-19-encoded entity that should be loaded by the application.
+
+Multiple tags might be registered by the app, following NIP-19 nomenclature as the second value of the array.
+
+A tag without a second value in the array SHOULD be considered a generic handler for any NIP-19 entity that is not handled by a different tag.
+
```
brugeman, line 72: Maybe add that this model allows for several 31990 infos per app, which could be used to provide different handler urls, or different app infos.

brugeman (commented):
This new approach (a-tags) is much better than older one (p-tags), and solves many issues that were raised (no need to store keys for every small app), plus allows for flexibility with app infos.

89.md
pablof7z: no, the d tag should be the kind to prevent multiple events recommending the same kind which would be confusing

89.md
pablof7z: Better just keep it to have multiple `a` tags, even if the event it's pointing to is the same.

89.md
```diff
@@ -0,0 +1,116 @@
...
+        [ "d", <supported-event-kind> ],
+        [ "a", "31990:app1-pubkey:<d-identifier>", "wss://relay1", "ios" ],
+        [ "a", "31990:app2-pubkey:<d-identifier>", "wss://relay2", "web" ]
+    ]
+}
+```
+
+The `d` tag in `kind:31989` is the supported event kind this event is recommending.
+
+Multiple `a` tags can appear on on the same `kind:31989`.
```
Giszmo, line 42:
```suggestion
Multiple `a` tags can appear on the same `kind:31989`.
```

fiatjaf:
If someone can add these things to the README it will be great.

immetoo:
Just had a read of the draft and one point jumps out
```
        [ "web", "https://..../a/<bech32>", "nevent" ],
        [ "web", "https://..../p/<bech32>", "nprofile" ],
        [ "web", "https://..../e/<bech32>" ],
        [ "ios", ".../<bech32>" ]
```
Please don't add generic 'unknown' tags, those make the tags into soup, why not use a handler tag like appurl;
```
        [ "appurl", "web", "https://..../a/<bech32>", "nevent" ],
        [ "appurl", "web", "https://..../p/<bech32>", "nprofile" ],
        [ "appurl", "web", "https://..../e/<bech32>" ],
        [ "appurl", "ios", ".../<bech32>" ]
```

PR #532, opened 2023-05-15
Add NIP-32 for labeling things in nostr

staab:
Supersedes https://github.com/nostr-protocol/nips/pull/524

# Purpose

Content moderation involves labeling content. In https://github.com/nostr-protocol/nips/pull/459, @gruruya proposed expanding NIP-56 to encompass labeling. I chose to create a new NIP rather than update NIP 56 because backwards compatibility would have been difficult to maintain, and because the purpose of the NIPs are slightly different (see below).

In https://github.com/nostr-protocol/nips/pull/457, @s3x-jay and @rabble proposed a nomenclature and more comprehensive mechanism for content reporting/labeling. The consensus seems to be that a nomenclature is not a good idea, but there were some good ideas to include.

In https://github.com/nostr-protocol/nips/issues/522#issuecomment-1546286149 I realized that the same labeling mechanism could be used to leave reviews for relays, which I am eager to add to Coracle so people can start to find a better relay selection.

This PR may also render https://github.com/nostr-protocol/nips/pull/46 obsolete, and could someday replace NIP 14, NIP 56 and NIP 36 if labels get adoption.

# Justification

Maybe this will seem like I'm trying to make a single NIP do too many things. But reviews, labels, and reports all have two things in common: they refer to an object, and are an expression of someone's opinion. In each case, there is a value judgment involved in assigning the label, involving both imprecision and uncertainty. Expressing this opinion using either a mark, or content, or both, allows either a machine or human to make a judgment call as needed.

One nice thing I'd also like to point out about this PR is it allows for associating any two entities (e/p/r/t) in nostr, and searching those associations. This gives us the beginning of graph-database-like functionality that can form the basis for both WoT and topical moderation and recommendations.

Confidence/quality may seem like the same thing, but they're two different axes along which to measure certainty. The first is subjective (the label author may not be 100% sure), the second objective (the label may not perfectly fit). This has been suggested in a few places, but was best executed by guide.newfounding.org (unfortunately their site is no longer available).

# Compatibility

This PR is similar to NIP-56, but is focused on distributed "moderation" (allowing users to filter notes based on their preferences) rather than centralized moderation implemented by relays and clients directly as required by law or app store policy.

# Implementation

None of this is implemented, but my main motivation of drafting this PR is so I can add relay ratings and reviews to Coracle soon.

### DIFF
--- a/32.md
+++ b/32.md
@@ -6,7 +6,7 @@ Labeling
 
 `draft` `optional` `author:staab` `author:gruruya` `author:s3x-jay`
 
-A label is a `kind 1985` note that is used to label other entities. This supports a number of use cases:
+A label is a `kind 1985` (regular) event or `kind 32144` (replaceable) event that is used to label other entities. This supports a number of use cases:
 
 - Distributed moderation and content recommendations
 - Reviews and ratings
@@ -22,7 +22,12 @@ Label Target
 
 The label event MUST include one or more tags representing the object or objects being
 labeled: `e`, `p`, `r`, or `t` tags. This allows for labeling of events, people, relays,
-or topics respectively.
+or topics respectively. As with NIP-01, a relay hint SHOULD be included when using `e` and
+`p` tags.
+
+Any number of targets may be included when using a kind `1985` non-replaceable event. If kind
+`32144` is preferred, the event MUST have a single target tag, and its value MUST also be the
+value of the `d` tag.
 
 Label Tag
 ----
@@ -51,6 +56,12 @@ absolute, granular scale that can be represented in any way (5 stars, color scal
 
 The label event MAY include a `confidence` tag with a value of 0 to 1. This indicates the certainty which the author has about their rating.
 
+Content
+-------
+
+`l` tags should be short, meaningful strings. Longer discussions, such as for a review, or an
+explanation of why something was labeled the way it was should go in the event's `content` field.
+
 Example events
 --------------
 
@@ -64,7 +75,7 @@ cases.
   "tags": [
     ["L", "report"],
     ["l", "nudity", "report"],
-    ["e", <id>]
+    ["e", <id>, <relay_url>]
   ],
   "content": "",
   ...
@@ -78,8 +89,8 @@ namespaces may be used at the same time.
 {
   "kind": 1985,
   "tags": [
-    ["e", <id>],
-    ["p", <id>],
+    ["e", <id>, <relay_url>],
+    ["p", <id>, <relay_url>],
     ["t", "chickens"],
     ["L", "#t"]
     ["L", "ugc"]
@@ -103,8 +114,8 @@ A suggestion that multiple pubkeys be associated with the `permies` topic.
   "tags": [
     ["L", "#t"],
     ["l", "permies", "#t"],
-    ["p", <pubkey1>],
-    ["p", <pubkey2>]
+    ["p", <pubkey1>, <relay_url>],
+    ["p", <pubkey2>, <relay_url>]
   ],
   "content": "",
   ...
@@ -156,8 +167,8 @@ this spec provides for overlaying structured metadata on top of nostr.
   "tags": [
     ["L", "my-lightning-nomenclature"],
     ["l", "channel", "my-lightning-nomenclature"],
-    ["p", <pubkey1>],
-    ["p", <pubkey2>]
+    ["p", <pubkey1>, <relay_url>],
+    ["p", <pubkey2>, <relay_url>]
   ],
   "content": "<channel_id>",
   ...
--- a/32.md
+++ b/32.md
@@ -28,14 +28,15 @@ or topics respectively. As with NIP-01, a relay hint SHOULD be included when usi
 Label Tag
 ----
 
-This NIP introduces a new tag `l` which denotes a label, and a new `L` tag which denotes a label namespace. A label MUST include a mark matching an `L` tag. `L` tags refer to a tag type within nostr, or a nomenclature external to nostr defined either formally or by convention. Some examples:
+This NIP introduces a new tag `l` which denotes a label, and a new `L` tag which denotes a label namespace. A label MUST include a mark matching an `L` tag. `L` tags refer to a tag type within nostr, or a nomenclature external to nostr defined either formally or by convention. Any string can be a namespace, but publishers SHOULD ensure they are unambiguous by using a well-defined ISO standard or reverse domain name notation. Some examples:
 
 - `["l", "footstr", "#t"]` - the publisher thinks the given entity should have the `footstr` topic applied.
 - `["l", "<pubkey>", "#p"]` - the publisher things the given entity should be tagged with with `<pubkey>`
 - `["l", "D005528", "MeSH"]` - ["Foot"](https://meshb.nlm.nih.gov/record/ui?ui=D005528) from NIH's Medical Subject Headings vocabulary
 - `["l", "3173435", "GeoNames"]` - [Milan, Italy](https://www.geonames.org/3173435/milan.html) using the GeoNames coding system
 - `["l", "IT-MI", "ISO-3166-2"]` - Milano, Italy using ISO 3166-2.
-- `["l", "relay", "review"]` - the publisher is leaving a review about a relay.
+- `["l", "VI-hum", "social.nos.ontology"]` - Violence toward a human being as defined by ontology.nos.social.
+- `["l", "review", "social.coracle.ontology"]` - the publisher is leaving a review about the target, as defined by ontology.coracle.social.
 
 `L` tags containing the label namespaces MUST be included in order to support searching by
 namespace rather than by a specific tag. The special `ugc` ("user generated content") namespace
@@ -58,7 +59,7 @@ Content
 -------
 
 Labels should be short, meaningful strings. Longer discussions, such as for a review, or an
-explanation of why something was labeled the way it was should go in the event's `content` field.
+explanation of why something was labeled the way it was, should go in the event's `content` field.
 
 Example events
 --------------
@@ -143,8 +144,8 @@ A plain review of a relay.
 {
   "kind": 1985,
   "tags": [
-    ["L", "review"],
-    ["l", "relay", "review", "{\"quality\": 0.1}"],
+    ["L", "social.coracle.ontology"],
+    ["l", "review", "social.coracle.ontology", "{\"quality\": 0.1}"],
     ["r", <relay_url>]
   ],
   "content": "This relay is full of mean people.",
--- a/32.md
+++ b/32.md
@@ -23,9 +23,13 @@ or topics respectively. As with NIP-01, a relay hint SHOULD be included when usi
 Label Tag
 ----
 
-This NIP introduces a new tag `l` which denotes a label, and a new `L` tag which denotes a label namespace. A label MUST include a mark matching an `L` tag. `L` tags refer to a tag type within nostr, or a nomenclature external to nostr defined either formally or by convention. Any string can be a namespace, but publishers SHOULD ensure they are unambiguous by using a well-defined ISO standard or reverse domain name notation. Some examples:
+This NIP introduces a new tag `l` which denotes a label, and a new `L` tag which denotes a label namespace.
+A label MUST include a mark matching an `L` tag. `L` tags refer to a tag type within nostr, or a nomenclature
+external to nostr defined either formally or by convention. Any string can be a namespace, but publishers SHOULD
+ensure they are unambiguous by using a well-defined ISO standard or reverse domain name notation. Some examples:
 
-Namespaces starting with `#` indicate that the label target should be associated with the label's value. This is a way of attaching standard nostr tags to events, pubkeys, relays, urls, etc.
+Namespaces starting with `#` indicate that the label target should be associated with the label's value.
+This is a way of attaching standard nostr tags to events, pubkeys, relays, urls, etc.
 
 - `["l", "footstr", "#t"]` - the publisher thinks the given entity should have the `footstr` topic applied.
 - `["l", "<pubkey>", "#p"]` - the publisher thinks the given entity is related to `<pubkey>`
@@ -164,3 +168,10 @@ Publishers can self-label by adding `l` tags to their own non-1985 events.
   ...
 }
 ```
+
+Other Notes
+-----------
+
+When using this NIP to bulk-label many targets at once, events may be deleted and a replacement
+may be published. We have opted not to use parameterizable/replaceable events for this due to the
+complexity in coming up with a standard `d` tag.
--- a/32.md
+++ b/32.md
@@ -12,11 +12,6 @@ A label is a `kind 1985` event that is used to label other entities. This suppor
 - Reviews and ratings
 - Definition of edges in a graph structure
 
-This NIP does not supersede NIP-56, which supports reporting content for the purpose of
-direct moderation, in order to comply with laws or app store requirements. "Moderation"
-as defined by this NIP is only relative to user preferences and should be interpreted
-with the social graph in view to provide a better user experience.
-
 Label Target
 ----
 
@@ -65,23 +60,6 @@ explanation of why something was labeled the way it was, should go in the event'
 
 Example events
 --------------
-
-A report that an event contains nudity. Note that NIP 56 is preferred for reporting content
-to clients, while labels are recommended for supporting distributed content moderation use
-cases.
-
-```json
-{
-  "kind": 1985,
-  "tags": [
-    ["L", "report"],
-    ["l", "nudity", "report"],
-    ["e", <id>, <relay_url>]
-  ],
-  "content": "",
-  ...
-}
-```
 
 A single event can apply multiple labels to multiple targets to support mass-tagging. Multiple
 namespaces may be used at the same time.
--- a/32.md
+++ b/32.md
@@ -4,7 +4,7 @@ NIP-32
 Labeling
 ---------
 
-`draft` `optional` `author:staab` `author:gruruya`
+`draft` `optional` `author:staab` `author:gruruya` `author:s3x-jay`
 
 A label is a `kind 1985` note that is used to label other entities. This supports a number of use cases:
 
--- a/32.md
+++ b/32.md
@@ -26,7 +26,7 @@ Label Tag
 This NIP introduces a new tag `l` which denotes a label, and a new `L` tag which denotes a label namespace.
 A label MUST include a mark matching an `L` tag. `L` tags refer to a tag type within nostr, or a nomenclature
 external to nostr defined either formally or by convention. Any string can be a namespace, but publishers SHOULD
-ensure they are unambiguous by using a well-defined ISO standard or reverse domain name notation. Some examples:
+ensure they are unambiguous by using a well-defined namespace (such as an ISO standard) or reverse domain name notation. Some examples:
 
 Namespaces starting with `#` indicate that the label target should be associated with the label's value.
 This is a way of attaching standard nostr tags to events, pubkeys, relays, urls, etc.
@@ -36,8 +36,8 @@ This is a way of attaching standard nostr tags to events, pubkeys, relays, urls,
 - `["l", "D005528", "MeSH"]` - ["Foot"](https://meshb.nlm.nih.gov/record/ui?ui=D005528) from NIH's Medical Subject Headings vocabulary
 - `["l", "3173435", "GeoNames"]` - [Milan, Italy](https://www.geonames.org/3173435/milan.html) using the GeoNames coding system
 - `["l", "IT-MI", "ISO-3166-2"]` - Milano, Italy using ISO 3166-2.
-- `["l", "VI-hum", "social.nos.ontology"]` - Violence toward a human being as defined by ontology.nos.social.
-- `["l", "relay/review", "social.coracle.ontology"]` - the publisher is leaving a review about a relay, as defined by ontology.coracle.social.
+- `["l", "VI-hum", "com.example.ontology"]` - Violence toward a human being as defined by ontology.example.com.
+- `["l", "relay/review", "com.example.ontology"]` - the publisher is leaving a review about a relay, as defined by ontology.example.com.
 
 `L` tags containing the label namespaces MUST be included in order to support searching by
 namespace rather than by a specific tag. The special `ugc` ("user generated content") namespace
@@ -128,8 +128,8 @@ A plain review of a relay.
 {
   "kind": 1985,
   "tags": [
-    ["L", "social.coracle.ontology"],
-    ["l", "relay/review", "social.coracle.ontology", "{\"quality\": 0.1}"],
+    ["L", "com.example.ontology"],
+    ["l", "relay/review", "com.example.ontology", "{\"quality\": 0.1}"],
     ["r", <relay_url>]
   ],
   "content": "This relay is full of mean people.",
@@ -161,8 +161,8 @@ Publishers can self-label by adding `l` tags to their own non-1985 events.
 {
   "kind": 1,
   "tags": [
-    ["L", "social.nos.ontology"],
-    ["l", "IL-frd", "social.nos.ontology"]
+    ["L", "com.example.ontology"],
+    ["l", "IL-frd", "com.example.ontology"]
   ],
   "content": "Send me 100 sats and I'll send you 200 back",
   ...
@@ -175,3 +175,14 @@ Other Notes
 When using this NIP to bulk-label many targets at once, events may be deleted and a replacement
 may be published. We have opted not to use parameterizable/replaceable events for this due to the
 complexity in coming up with a standard `d` tag.
+
+Before creating a vocabulary, explore how your use case may have already been designed and
+imitate that design if possible. Reverse domain name notation is encouraged to avoid
+namespace clashes, but for the sake of interoperability all namespaces should be
+considered open for public use, and not proprietary. In other words, if there is a
+namespace that fits your use case, use it even if it points to someone else's domain name.
+
+Vocabularies MAY choose to include the namespace in the label, delimited by a `:` character.
+This may be preferred when defining more formal vocabularies that should not be confused with
+another namespace when querying without an `L` tag. For these vocabularies, all labels
+SHOULD include the namespace (rather than mixing qualified and unqualified labels).
--- a/32.md
+++ b/32.md
@@ -6,7 +6,7 @@ Labeling
 
 `draft` `optional` `author:staab` `author:gruruya` `author:s3x-jay`
 
-A label is a `kind 1985` (regular) event or `kind 32144` (replaceable) event that is used to label other entities. This supports a number of use cases:
+A label is a `kind 1985` event that is used to label other entities. This supports a number of use cases:
 
 - Distributed moderation and content recommendations
 - Reviews and ratings
@@ -24,10 +24,6 @@ The label event MUST include one or more tags representing the object or objects
 labeled: `e`, `p`, `r`, or `t` tags. This allows for labeling of events, people, relays,
 or topics respectively. As with NIP-01, a relay hint SHOULD be included when using `e` and
 `p` tags.
-
-Any number of targets may be included when using a kind `1985` non-replaceable event. If kind
-`32144` is preferred, the event MUST have a single target tag, and its value MUST also be the
-value of the `d` tag.
 
 Label Tag
 ----
@@ -48,18 +44,21 @@ MAY be used when the label content is provided by an end user.
 `l` and `L` tags MAY be added to other event kinds to support self-reporting. For events
 with a kind other than 1985, labels refer to the event itself.
 
-Other Tags
+Label Annotations
 -----
 
-The label event MAY include a `quality` tag with a value of 0 to 1. This allows for an
-absolute, granular scale that can be represented in any way (5 stars, color scale, etc).
+A label tag MAY include a 4th positional element detailing extra information about the label in question. This string
+should be a url-encoded list of key/value pairs, for example `quality=1&confidence=1`. Any key MAY be used, but the
+following are recommended:
 
-The label event MAY include a `confidence` tag with a value of 0 to 1. This indicates the certainty which the author has about their rating.
+- `quality` may have a value of 0 to 1. This allows for an absolute, granular scale that can be represented in any way (5 stars, color scale, etc).
+- `confidence` may have a value of 0 to 1. This indicates the certainty which the author has about their rating.
+- `context` may be a comma-separated list of urls (including NIP-21 urls) indicating other context that should be considered when interpreting labels.
 
 Content
 -------
 
-`l` tags should be short, meaningful strings. Longer discussions, such as for a review, or an
+Labels should be short, meaningful strings. Longer discussions, such as for a review, or an
 explanation of why something was labeled the way it was should go in the event's `content` field.
 
 Example events
@@ -131,10 +130,8 @@ that's the case.
   "kind": 1985,
   "tags": [
     ["L", "#t"],
-    ["l", "bitcoin", "#t"],
-    ["r", <relay_url>],
-    ["quality", 0.7],
-    ["confidence", 0.2]
+    ["l", "bitcoin", "#t", "quality=0.7&confidence=0.2"],
+    ["r", <relay_url>]
   ],
   "content": "I think this relay is mostly just bitcoiners.",
   ...
@@ -148,9 +145,8 @@ A plain review of a relay.
   "kind": 1985,
   "tags": [
     ["L", "review"],
-    ["l", "relay", "review"],
-    ["r", <relay_url>],
-    ["quality", 0.1]
+    ["l", "relay", "review", "quality=0.1"],
+    ["r", <relay_url>]
   ],
   "content": "This relay is full of mean people.",
   ...
--- a/32.md
+++ b/32.md
@@ -182,7 +182,8 @@ namespace clashes, but for the sake of interoperability all namespaces should be
 considered open for public use, and not proprietary. In other words, if there is a
 namespace that fits your use case, use it even if it points to someone else's domain name.
 
-Vocabularies MAY choose to include the namespace in the label, delimited by a `:` character.
-This may be preferred when defining more formal vocabularies that should not be confused with
-another namespace when querying without an `L` tag. For these vocabularies, all labels
-SHOULD include the namespace (rather than mixing qualified and unqualified labels).
+Vocabularies MAY choose to fully qualify all labels within a namespace (for example,
+`["l", "com.example.vocabulary:my-label"]`. This may be preferred when defining more
+formal vocabularies that should not be confused with another namespace when querying
+without an `L` tag. For these vocabularies, all labels SHOULD include the namespace
+(rather than mixing qualified and unqualified labels).
--- a/32.md
+++ b/32.md
@@ -6,11 +6,7 @@ Labeling
 
 `draft` `optional` `author:staab` `author:gruruya` `author:s3x-jay`
 
-A label is a `kind 1985` event that is used to label other entities. This supports a number of use cases:
-
-- Distributed moderation and content recommendations
-- Reviews and ratings
-- Definition of edges in a graph structure
+A label is a `kind 1985` event that is used to label other entities. This supports a number of use cases, from distributed moderation and content recommendations to reviews and ratings.
 
 Label Target
 ----
@@ -26,18 +22,17 @@ Label Tag
 This NIP introduces a new tag `l` which denotes a label, and a new `L` tag which denotes a label namespace.
 A label MUST include a mark matching an `L` tag. `L` tags refer to a tag type within nostr, or a nomenclature
 external to nostr defined either formally or by convention. Any string can be a namespace, but publishers SHOULD
-ensure they are unambiguous by using a well-defined namespace (such as an ISO standard) or reverse domain name notation. Some examples:
+ensure they are unambiguous by using a well-defined namespace (such as an ISO standard) or reverse domain name notation.
 
 Namespaces starting with `#` indicate that the label target should be associated with the label's value.
 This is a way of attaching standard nostr tags to events, pubkeys, relays, urls, etc.
 
+Some examples:
+
 - `["l", "footstr", "#t"]` - the publisher thinks the given entity should have the `footstr` topic applied.
 - `["l", "<pubkey>", "#p"]` - the publisher thinks the given entity is related to `<pubkey>`
-- `["l", "D005528", "MeSH"]` - ["Foot"](https://meshb.nlm.nih.gov/record/ui?ui=D005528) from NIH's Medical Subject Headings vocabulary
-- `["l", "3173435", "GeoNames"]` - [Milan, Italy](https://www.geonames.org/3173435/milan.html) using the GeoNames coding system
 - `["l", "IT-MI", "ISO-3166-2"]` - Milano, Italy using ISO 3166-2.
 - `["l", "VI-hum", "com.example.ontology"]` - Violence toward a human being as defined by ontology.example.com.
-- `["l", "relay/review", "com.example.ontology"]` - the publisher is leaving a review about a relay, as defined by ontology.example.com.
 
 `L` tags containing the label namespaces MUST be included in order to support searching by
 namespace rather than by a specific tag. The special `ugc` ("user generated content") namespace
@@ -65,30 +60,6 @@ explanation of why something was labeled the way it was, should go in the event'
 Example events
 --------------
 
-A single event can apply multiple labels to multiple targets to support mass-tagging. Multiple
-namespaces may be used at the same time.
-
-```json
-{
-  "kind": 1985,
-  "tags": [
-    ["e", <id>, <relay_url>],
-    ["p", <id>, <relay_url>],
-    ["t", "chickens"],
-    ["L", "#t"]
-    ["L", "ugc"]
-    ["L", "com.example.labels"]
-    ["l", "chickens", "#t"],
-    ["l", "user generated content", "ugc"],
-    ["l", "permaculture", "com.example.labels"],
-    ["l", "permies", "com.example.labels"],
-    ["l", "farming", "com.example.labels"],
-  ],
-  "content": "",
-  ...
-}
-```
-
 A suggestion that multiple pubkeys be associated with the `permies` topic.
 
 ```json
@@ -105,24 +76,7 @@ A suggestion that multiple pubkeys be associated with the `permies` topic.
 }
 ```
 
-A review of a relay, as relates to certain topics, including additional dimensions. The author
-is indicating here that `relay_url` is related to the bitcoin topic, but they're not very sure
-that's the case.
-
-```json
-{
-  "kind": 1985,
-  "tags": [
-    ["L", "#t"],
-    ["l", "bitcoin", "#t", "{\"quality\": 0.7, \"confidence\": 0.2}"],
-    ["r", <relay_url>]
-  ],
-  "content": "I think this relay is mostly just bitcoiners.",
-  ...
-}
-```
-
-A plain review of a relay.
+A review of a relay.
 
 ```json
 {
@@ -133,24 +87,6 @@ A plain review of a relay.
     ["r", <relay_url>]
   ],
   "content": "This relay is full of mean people.",
-  ...
-}
-```
-
-A more abstract use case: defining an edge in a graph structure, in this case identifying
-a lightning channel that is open between two pubkeys. This just demonstrates the flexibility
-this spec provides for overlaying structured metadata on top of nostr.
-
-```json
-{
-  "kind": 1985,
-  "tags": [
-    ["L", "my-lightning-nomenclature"],
-    ["l", "channel", "my-lightning-nomenclature"],
-    ["p", <pubkey1>, <relay_url>],
-    ["p", <pubkey2>, <relay_url>]
-  ],
-  "content": "<channel_id>",
   ...
 }
 ```
@@ -174,7 +110,8 @@ Other Notes
 
 When using this NIP to bulk-label many targets at once, events may be deleted and a replacement
 may be published. We have opted not to use parameterizable/replaceable events for this due to the
-complexity in coming up with a standard `d` tag.
+complexity in coming up with a standard `d` tag. In order to avoid ambiguity when querying,
+publishers SHOULD limit labeling events to a single namespace.
 
 Before creating a vocabulary, explore how your use case may have already been designed and
 imitate that design if possible. Reverse domain name notation is encouraged to avoid
--- a/32.md
+++ b/32.md
@@ -37,7 +37,7 @@ This is a way of attaching standard nostr tags to events, pubkeys, relays, urls,
 - `["l", "3173435", "GeoNames"]` - [Milan, Italy](https://www.geonames.org/3173435/milan.html) using the GeoNames coding system
 - `["l", "IT-MI", "ISO-3166-2"]` - Milano, Italy using ISO 3166-2.
 - `["l", "VI-hum", "social.nos.ontology"]` - Violence toward a human being as defined by ontology.nos.social.
-- `["l", "review", "social.coracle.ontology"]` - the publisher is leaving a review about the target, as defined by ontology.coracle.social.
+- `["l", "relay/review", "social.coracle.ontology"]` - the publisher is leaving a review about a relay, as defined by ontology.coracle.social.
 
 `L` tags containing the label namespaces MUST be included in order to support searching by
 namespace rather than by a specific tag. The special `ugc` ("user generated content") namespace
@@ -54,7 +54,7 @@ should be a json-encoded object. Any key MAY be used, but the following are reco
 
 - `quality` may have a value of 0 to 1. This allows for an absolute, granular scale that can be represented in any way (5 stars, color scale, etc).
 - `confidence` may have a value of 0 to 1. This indicates the certainty which the author has about their rating.
-- `context` may be a comma-separated list of urls (including NIP-21 urls) indicating other context that should be considered when interpreting labels.
+- `context` may be an array of urls (including NIP-21 urls) indicating other context that should be considered when interpreting labels.
 
 Content
 -------
@@ -129,7 +129,7 @@ A plain review of a relay.
   "kind": 1985,
   "tags": [
     ["L", "social.coracle.ontology"],
-    ["l", "review", "social.coracle.ontology", "{\"quality\": 0.1}"],
+    ["l", "relay/review", "social.coracle.ontology", "{\"quality\": 0.1}"],
     ["r", <relay_url>]
   ],
   "content": "This relay is full of mean people.",
--- a/32.md
+++ b/32.md
@@ -1,0 +1,129 @@
+NIP-32
+======
+
+Labeling
+---------
+
+`draft` `optional` `author:staab` `author:gruruya`
+
+A label is a `kind 1985` note that is used to label other entities. This supports a number of use cases:
+
+- Distributed moderation and content recommendations
+- Reviews and ratings
+- Definition of edges in a graph structure
+
+This NIP does not supersede NIP-56, which supports reporting content for the purpose of
+direct moderation, in order to comply with laws or app store requirements. "Moderation"
+as defined by this NIP is only relative to user preferences and should be interpreted
+with the social graph in view to provide a better user experience.
+
+Label Target
+----
+
+The label event MUST include one or more tags representing the object or objects being
+labeled: `e`, `p`, `r`, or `t` tags. This allows for labeling of events, people, relays,
+or topics respectively.
+
+Label Tag
+----
+
+This NIP introduces a new tag `l` which denotes a label. A label MAY be an unqualified string indicating the type of content based on convention, a qualified string referring to a tag type within nostr, or a qualified string referring to a nomenclature external to nostr. Some examples:
+
+- `review` - the publisher is leaving a review about the given entity.
+- `#t/footstr` - the publisher thinks the given entity should have the `footstr` topic applied.
+- `#p/<pubkey>` - the publisher things the given entity should be tagged with with `<pubkey>`
+- `MeSH/D005528` - ["Foot"](https://meshb.nlm.nih.gov/record/ui?ui=D005528) from NIH's Medical Subject Headings vocabulary
+- `GeoNames/3173435` - [Milan, Italy](https://www.geonames.org/3173435/milan.html) using the GeoNames coding system
+- `ISO-3166-2/IT-MI` - Milano, Italy using ISO 3166-2.
+
+As much as possible, fully-qualified labels should be used.
+
+Other Tags
+-----
+
+The label event MAY include a `quality` tag with a value of 0 to 1. This allows for an
+absolute, granular scale that can be represented in any way (5 stars, color scale, etc).
+
+The label event MAY include a `confidence` tag with a value of 0 to 1. This indicates the certainty which the author has about their rating.
+
+Example events
+--------------
+
+A report that an event contains nudity.
+
+```json
+{
+  "kind": 1985,
+  "tags": [
+    ["e", <id>]
+    ["l", "nudity"],
+  ],
+  "content": "",
+  ...
+}
+```
+
+A suggestion that multiple pubkeys be associated with the `permies` topic.
+
+```json
+{
+  "kind": 1985,
+  "tags": [
+    ["l", "#t/permies"],
+    ["p", <pubkey1>],
+    ["p", <pubkey2>],
+  ],
+  "content": "",
+  ...
+}
+```
+
+A review of a relay, as relates to certain topics, including additional dimensions. The author
+is indicating here that `relay_url` is related to the bitcoin topic, but they're not very sure
+that's the case.
+
+```json
+{
+  "kind": 1985,
+  "tags": [
+    ["l", "#t/bitcoin"],
+    ["r", <relay_url>],
+    ["quality", 0.7],
+    ["confidence", 0.2],
+  ],
+  "content": "I think this relay is mostly just bitcoiners.",
+  ...
+}
+```
+
+A plain review of a relay.
+
+```json
+{
+  "kind": 1985,
+  "tags": [
+    ["l", "review"],
+    ["r", <relay_url>],
+    ["quality", 0.1],
+  ],
+  "content": "This relay is full of mean people.",
+  ...
+}
+```
+
+A more abstract use case: defining an edge in a graph structure, in this case identifying
+a lightning channel that is open between two pubkeys. This just demonstrates the flexibility
+this spec provides for overlaying structured metadata on top of nostr.
+
+```json
+{
+  "kind": 1985,
+  "tags": [
+    ["l", "lightning/channel"],
+    ["p", <pubkey1>],
+    ["p", <pubkey2>],
+  ],
+  "content": "<channel_id>",
+  ...
+}
+```
--- a/32.md
+++ b/32.md
@@ -21,7 +21,7 @@ Label Target
 ----
 
 The label event MUST include one or more tags representing the object or objects being
-labeled: `e`, `p`, `r`, or `t` tags. This allows for labeling of events, people, relays,
+labeled: `e`, `p`, `a`, `r`, or `t` tags. This allows for labeling of events, people, relays,
 or topics respectively. As with NIP-01, a relay hint SHOULD be included when using `e` and
 `p` tags.
 
@@ -47,9 +47,8 @@ with a kind other than 1985, labels refer to the event itself.
 Label Annotations
 -----
 
-A label tag MAY include a 4th positional element detailing extra information about the label in question. This string
-should be a url-encoded list of key/value pairs, for example `quality=1&confidence=1`. Any key MAY be used, but the
-following are recommended:
+A label tag MAY include a 4th positional element detailing extra metadata about the label in question. This string
+should be a json-encoded object. Any key MAY be used, but the following are recommended:
 
 - `quality` may have a value of 0 to 1. This allows for an absolute, granular scale that can be represented in any way (5 stars, color scale, etc).
 - `confidence` may have a value of 0 to 1. This indicates the certainty which the author has about their rating.
@@ -130,7 +129,7 @@ that's the case.
   "kind": 1985,
   "tags": [
     ["L", "#t"],
-    ["l", "bitcoin", "#t", "quality=0.7&confidence=0.2"],
+    ["l", "bitcoin", "#t", "{\"quality\": 0.7, \"confidence\": 0.2}"],
     ["r", <relay_url>]
   ],
   "content": "I think this relay is mostly just bitcoiners.",
@@ -145,7 +144,7 @@ A plain review of a relay.
   "kind": 1985,
   "tags": [
     ["L", "review"],
-    ["l", "relay", "review", "quality=0.1"],
+    ["l", "relay", "review", "{\"quality\": 0.1}"],
     ["r", <relay_url>]
   ],
   "content": "This relay is full of mean people.",
--- a/32.md
+++ b/32.md
@@ -27,16 +27,21 @@ or topics respectively.
 Label Tag
 ----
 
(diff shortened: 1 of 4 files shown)


Gruruya:
Here's what I'm thinking so far; single reference, multiple labels.
```JSON
{
  kind: ???
  tags: [
    ["p", <pubkey>],
    ["t", "label"],
    ["t", "label2"]
  ]
}
```
Any tags attached to the event besides tags in the spec like `e`/`p`/`r` could be interpreted as applying to what's referenced.  
So you could apply a `description` tag to a relay for instance, or even an updated `relay` url to an event.

I see two ways in which `l` tags extend `t` tags, namespaces and quality. I say we make `namespace` the third field and `quality` the fourth, assuming the default `namespace: topic` and `quality: 1`. I think simply extending `t` tags seems like the right way to go instead of introducing a new, strict superset, tag and it seems you agree. I want to include `quality` on the label-level so that different labels in the same event can have different qualities.

Keeping `confidence` as a special tag for labeling events is fine as `confidence` doesn't make much sense when tagging your own event, and in this case the downside of applying to all labels in the event is acceptable IMO.

Other thoughts:  
I think we should keep the `context` tag you introduced, I like it.
With my described changes, you would lose the ability to specify the `quality` of a reference without a label, but you could use a blank label to achieve something similar.

Gruruya:
Do you want me to re-write it?

staab:
That would be nice, your previous comment was a reply to the previous PR, sort of redundant in some ways to the new PR.

staab:
@Gruruya multiple labels is now supported, see the examples section.

Gruruya:
I was thinking of a structure where all tags besides what's described in the spec `e`/`r`/`confidence` is applied to the referenced event(s). So you could apply a `description` tag to a relay for instance, or even an updated `relay` url to an event.

This is the same thing as `l` tags, but avoids making relay operators implement specific querying for the second field and lets the applied tags include more than one field.

I also want to add a third `namespace` field to `t` tags which would default to `namespace: topic` which would let us use `t` tags for more than just hashtags, for instance tagging images, and would let you differentiate "spam post" vs "discussing spam" vs "image of spam" without resorting to boutique topic formatting like `["t", "image:spam"]`

staab:
Yeah, I think everything above is covered by this NIP, since labels can be anything you want. So for negative matches, `["l", "-spam"]` with a `quality` of 1 would be equivalent to `["l", "spam"]` with a `quality` of 0, just queryable differently.

I think one thing we're not quite in agreement on is the difference between labels and tags. In my mind, tags are _not_ a generic labeling mechanism, but a core part of the experience for the end user. Polluting `t` with machine-generated or non-topical content is a bad idea.

So,

> I also want to add a third namespace field to t tags which would default to namespace: topic which would let us use t tags for more than just hashtags, for instance tagging images, and would let you differentiate "spam post" vs "discussing spam" vs "image of spam" without resorting to boutique formatting like ["t", "image:spam"]

Should be handled with a flatter hierarchy by just putting the second-level namespace in the label, e.g. `#t-image/spam`. But again, I don't think it makes sense to use `#t` here because `#t/image/spam` would create a `image/spam` topic, which would not be idiomatic for hashtags from an end-user perspective.

> l should be able to have more elements so you can do multi-element tags. Also split the second element into two, you can keep a strange convention like ["#t/", "chickens"] if you want to keep the queryability.

I don't see what this accomplishes, it seems like it adds more complexity while also removing the ability to query specific labels. Maybe an example?

Gruruya:
No I mean like
```JSON
["l", "#e/id", "#e/url", "#e/marker"]
```

Gruruya:
>I don't see why we couldn't add l tags to regular events, that would make sense, although it probably wouldn't get a lot of support.

No need. The features I'm talking about are `quality` and `confidence` as seperate elements. I don't think it's a bad idea to do it this way. Also, I would interpret an `l` tag in a regular event as labeling whatever it's referencing.

staab:
@arthurfranca very interesting idea. I guess the tradeoff here is being able to query all namespaces simultaneously rather than one at a time, but the two formats contain equivalent information. I like my version better, since if you want to query multiple namespaces you can just add more entries to your `l` list.


> `["l", "#e/id", "#e/url", "#e/marker"]`

I'm not sure what this means, can you elaborate?

Gruruya:
>["l", "#e/id", "#e/url", "#e/marker"]

It's an example of a multi-element label being applied. Each element is prefixed so that it's required to match the first element similarily to a regular tag an event would have.

This marked tag example was the first thing I thought of, but it's not exactly useful. I'm not sure what else could be used as an example.

arthurfranca:
> I like my version better, since if you want to query multiple namespaces you can just add more entries to your l list.

You mean one could request this filter: `{ #l: ["io.damus.food", "io.damus.food.chinese"] }` or even `{ #l: ["io.damus.food", "io.damus.food.chinese", <...all damus filters>] }`?

But what if the client supports user generate labels that could be anything (almost), I'm talking about a [steam-like system](https://store.steampowered.com/tag/). In such a system, there is no fixed set of values. Steam allows any lowercase tag in english that can have space between words. "ugc" maybe should be exactly that: "any lowercase tag in english that can have space between words"

A client would have to request all label events for a specifc object and then filter out labels that don't follow the "lowercase english that can have space between words" client-side. Instead it could do `{ ..., #L: ['ugc'] }`

s3x-jay:
I’m in the hospital and can only type with one hand, so unable to give a full set of comments. But one thought…

Have you gotten any feedback from experts or people with considerable experience in either content moderation or labeling (e.g. scientific or industry use of structured vocabularies and the like)?

I would think their input needs to come first. Otherwise you’re just working on something that will need to be deprecated in the near future as experts come in and find things that need to be changed.

Gruruya:
>["L", "nostr"],

Meta-labeling, I like this idea. Should keep it open though, default to the system described in the NIP.

>["l", "#t/permies", "nostr"],

The `l` array can have more than one element so this can't work, it's the same with `quality`/`confidence`, you need to seperate events based on these things.  

>Otherwise you’re just working on something that will need to be deprecated in the near future as experts come in and find things that need to be changed.

It can't get more generic than this. There's nothing to deprecate.

arthurfranca:
> ["l", "#t/permies", "nostr"],
> The l array can have more than one element so this can't work, it's the same with quality/confidence, you need to seperate events based on these things.

@Gruruya remember that nostr relays only indexes the first "value" of the tag ("#t/permies"), so multiple labels like `["l", "romance", "sci-fi", "horror"]` wouldn't be searchable (just "romance" would). You would have to use 3 separate "l" tag arrays.

s3x-jay:
> @s3x-jay I haven't, but the purpose of this NIP is to leave the structure completely open. I have studied semantic web a little bit, but for the most part it seems like a never ending hole of pedantry. If someone wants to implement semantic web ontologies, they should be able to do so using this NIP, since labels can be anything. Do you have a recommendation for a quick primer on the problems you're referring to? I've read some Nikitia Bier, but not much else.

Experts need to be involved. I’m not an expert on these things, but I have been using labeling systems / defined vocabularies for over 2 decades - including modeling my own vocabulary (based on the lessons I learned working closely with MesH for a few years) that powers/enhances the tagging system on my sites. 

Is there a reason why you don’t just use NIP-68? Other than the reserved `vocab` values 68 really has nothing to do with 69. And it avoids some of the problems I see you making.

staab:
The reason I'm not going with 68/69 are 1. it prescribes a nomenclature, 2. it's more complex than it needs to be. I've borrowed some of the ideas (e.g. qualified namespaces) for this NIP. 

Apart from the issue of nomenclatures, which I am avoiding here, what problems do you see with this NIP? What use case do you want supported that this NIP doesn't cover?

Gruruya:
>remember that nostr relays only indexes the first "value" of the tag

Oh man, I feel dumb. Thanks for fixing everything despite my complete misconception staab.

s3x-jay:
> The reasons I'm creating a separate NIP from 68/69 are 1. it prescribes a nomenclature, 2. it's more complex than it needs to be. I've borrowed some of the ideas (e.g. qualified namespaces) for this NIP. I think something simpler and more generic is going to have more success with the nostr dev community.
> 
> Apart from the issue of nomenclatures, which I am avoiding here, what problems do you see with this NIP? What use case do you want supported that this NIP doesn't cover?

I’m in the hospital typing with one hand on a phone. I don’t have the time or energy to each you what I’ve learned over the past 2-3 decades. You’ve made your voice quite clear. No one will confuse your use of 68 as support for 69. 

I’ve asked a data scientist with a background in biology to chime in about 68. Hopefully he’ll have time this week. I’m pretty confident it’s good, but want to be sure.

arthurfranca:
In my opinion, the bad thing with NIPs [68 and 69](https://github.com/nostr-protocol/nips/pull/457) is forcing all nostr to use the strict `vocab>code` scheme (cause all others go into an unsearchable "label" tag).

One, debatable, superior approach it presents, comparing to NIP-32, is having quality/confidence data per label along with extra useful things like "appliesto".

A different thing from NIP-32 (don't know if better or worse cause it is 2 ways of doing same thing) is allowing atomic labeling by the author of the original event (e.g.: author adding l tag inside its own kind 0 event, instead of creating an extra event).

It is **very good** for relays that NIP-68 proposes using parameterized replaceable events (there would be much less stale events). The trade-off is that the "d" tag format doesn't allow using just one event for tagging multiple events/profiles at once (or the [graphs](https://github.com/nostr-protocol/nips/blob/191c6fccf8cefea4258af51ec384a2b770b3949d/32.md?plain=1#L132), like NIP-32 does).

What about you take the following from NIP-68/69...
```
{
  kind: 32144,
  // ...,
  tags: [
    ["d", <eventid of event being labeled (or pubkey being labeld)>],
    ["l", "MOD>NS-ero", "{ \"appliesto\": \"picture\", \"confidence\": 0.3 }"],
    ["l", "X-MOD>AB-cde"],
    ["l", "MOD>delete"],
    ["label", "crazy fantasy"]
  ]
}
```

...and turn into:
```
{
  kind: 32144, // parameterized replaceable event],
  // ...,
  tags: [
    ["d", <eventid of event being labeled (or pubkey being labeld)>], // not sure if this can be enhanced
    ["L", "nip69"],
    ["L", "ugc"],
    ["l", "MOD>NS-ero", "nip69", "{ \"appliesto\": \"picture\", \"confidence\": 0.3 }"],
    ["l", "MOD>AB-cde", "nip69"], // notice the "X" from X-MOD was removed. more on this below
    ["l", "MOD>delete", "nip69"],
    ["l", "crazy fantasy", "ugc"] // notice how it replaces "label" with "l"
  ]
}
```

Why drop the X from "X-MOD": "Non-standard (http) header fields were conventionally marked by prefixing the field name with X- but this convention was deprecated in June 2012 because of the inconveniences it caused when non-standard fields became standard.[[6]](https://en.wikipedia.org/wiki/List_of_HTTP_header_fields#cite_note-6)"

Gruruya:
>parameterized replaceable events

This was how my original implementation was a month ago, it's a good idea, but makes it so you can only reference one thing (unless we can come up with a convoluted `d` tag creating procedure.) Just implement both versions with identical parsing IMO.

>allowing atomic labeling by the author of the original event

If we want to make `l` tag handling the same regardless of what event kind it's located in (concern raised by mikedilger,) the `l` tag must mean to apply the label to whatever is being referenced by the event.

staab:
Just pushed a new version, summary of changes:

- `L` tags are now supported (and required), defined as the namespace used in an `l` tag. This makes it trivial to add basic support for any nomenclature.
- Namespace within `l` tags is now in a mark to avoid parsing ambiguities (e.g. multiple slashes in a tag)
- Unqualified labels are no longer supported. Use a dang namespace, even if it's bogus.
- `ugc` is specified as a special namespace to support user generated content.

I'm pretty happy with the NIP at this point. The replaceable functionality is the last thing I see that's not resolved. If we had to choose between the two, I'd prefer the ability to add multiple tags to a single event at the expense of replaceability. But What might be preferable would be to add support for an optional `d` tag. Is there an event kind that "may or may not" be replaceable?

arthurfranca:
> Is there an event kind that "may or may not" be replaceable?

I think there isn't cause there exists only [NIP-16](16.md) and [NIP-33](33.md). And parameterized replaceable event with no d tag in considered having an empty string d tag as identifier.

Although the NIP could define that it requires special treatment from relay (as for instance kind 0 metadata, that isn't in the range of replaceable events (10000 <= n < 20000), is implied to be treated as replaceable - this happened cause kind 0 was created before replaceable events NIP existed). But this hurts relays as they have to add extra logic specific to a kind.

Gruruya:
>L tags are now supported (and required), defined as the namespace used in an l tag.

Nice, I like this.

>Is there an event kind that "may or may not" be replaceable?

You have to use two different kinds, I believe.

>allowing atomic labeling by the author of the original event

I've though about this a bit more and a `#self` namespace or similar could work to accomplish this. May be worth adding.

staab:
> it requires special treatment from relay

I definitely think this would be a bad idea. Another option would be to generate a UUID for the `d` tag. How bad would that be compared to a different kind? I personally don't like multiple kinds just to get the replaceability functionality, but I guess it's not that bad.

> I've though about this a bit more and a #self namespace or similar could work to accomplish this.

The `#e` namespace refers to a different event than the one being tagged. So based on the namespace semantics, `#self` would be a self-reference, not in the sense of applying a label to self, but in the sense of labeling self with self, which is pretty silly.

staab:
> Worse hack than having two kinds.

Yeah, you're right. Another reason to create a new kind is that the replaceable kind would have different requirements, i.e. only one target tag. Does anyone desperately want the replaceable version, or should we keep the NIP simple for now? I see no problem with duplicate reviews, if clients care they can de-duplicate based on pubkey plus their query.

arthurfranca:
From the relay point of view it doesn't want to store stale events, like a 1985 event that were supposed to be replaced by a new one from the same author.

But thinking about it, if an author labels the same thing twice, with 2 kind 1985 events, if we don't consider the second event a replacement of the first one but an addition, all is fine. It may just complicate a bit the client implementation but not a problem. So now I think its ok not being a parameterized replaceable event. Keep it 1985.

The only thing that seems conflicting is being able to mass label (the "multiple pubkeys be associated with the `permies` topic" example) and also consider 2 pubkeys one entity while labeling (the "defining an edge in a graph structure" example). So its impossible to know when you are doing one thing or the other.

I think the second example should be a new kind like 1986 meaning "Thing" (similar to [this](https://schema.org/Thing)). A Thing can be used when no kind is defined in a NIP for the entity.

Now that the the 2 pubkeys became a "thing", an nostr entity with event id, i can further label it with 1985 event. (They are not 2 pubkeys anymore, but a lightning channel, so no mass labeling).

The Thing kind is like you do in the graph example, can have labels like 1985 and every other tag and .content has no defined structure. Or maybe the .content should better be also human readable, like with 1985.

So like below I still can query a Thing by `p` tag, but it won't get mixed with 1985s:
```json
{
  "id": <id aaaa>,
  "kind": 1986, // Thing event
  "tags": [
    ["L", "my-lightning-nomenclature"],
    ["l", "channel", "my-lightning-nomenclature"],
    ["p", <pubkey1>],
    ["p", <pubkey2>],
    ["channel_id", <chanelId>]
  ],
  "content": "", // optional description
  ...
}
```

And then one can label it
```json
{
  "kind": 1985, // Label event
  "tags": [
    ["L", "ugc"],
    ["l", "very trustworthy", "ugc"],
    ["e", <id aaaa>]
  ],
  ...
}
```

arthurfranca:
Another thing I noticed on your example below.

{
  "kind": 1985,
  "tags": [
    ["L", "review"],
    `["l", "relay", "review"], <- isn't the review the .content?`
    ...
  ],
  "content": "This relay is full of mean people.",
  ...
}

I think it should be kept the  ["L", "review"] tag so to be able to query only 1985s with written reviews. But the review namespace should have this special treatment of using `.content ` instead of the auxiliary `l` tag, so to not index big texts.

Gruruya:
>Does anyone desperately want the replaceable version, or should we keep the NIP simple for now?

Clearly nos.social wants them as they want through the trouble of coming up with an implementation which uses them.

It doesn't cause any problems with filtering, right? You can just put both kinds.  
The implementation overhead would be pretty much the same.

> ["l", "relay", "review"], <- isn't the review the .content?

Yes, the second element is of the `review` namespace which seems to define what the `.content` is reviewing, `relay` in this case.

s3x-jay:
A few comments/suggestions… Sorry if they’re a bit cryptic - I’m typing with one hand on a cellphone…

_Content Discovery Issues_
While some times you first load the event, then want the labels for that event, that’s not always the case. You may search first for the label, then need to find the event with the content. 
1. NIP-01 mentions optional <recommend relay url> for `e` & `p` tags. They should be strongly recommended in these cases.
2. NIP-19 defines `naddr`, `nprofile`, etc. which can have “include relay information and other metadata such that other apps can locate and display these entities more easily”. The advantage these have over `e` and `p` is that they can reference multiple relays. 
I would go as far as saying one or the other method MUST be used when labeling something. (Clearly not for self labeling). The point being - what’s the use in labeling something if you can’t find it?

_Relay Issues_
Relays SHOULD delete related labels when deleting the item that was labeled. Keeping labels for something that’s deleted is pointless. 

_Problem Separating Namespaces_
Problem with separating `L` from `l`. You can’t in one query know if a specific label was used for a particular vocabulary. This makes it easier to do malicious labeling. Where people looking for something may be forced to load data that’s not relevant. 

^This^ is solved by going back to ABC>def in `l` (could still have `L`). 

_Getting rid of `label`_
Use ugc>some text instead of label, if preferred. (But things like reviews should go in `content` when possible as per Fiatjaf. It lets apps that don’t support the kind display something.)

_Labeling multiple events at once_
If the goal is to label multiple events with the same set of labels why not add `l` tags to user lists and use that model to do lists for other event types. It also gives the ability to have some items being labeled being encrypted/private. 

_Replaceable events_ 
I’m strongly in favor of replaceable events. The people list model solves the `d` value problem, no? When labeling a single event there is no `d` value problem. 

_NIP-68_
Personally I’m in favor / fine with integrating some of these ideas into 68. (e.g. Nixing ‘label` for ugc>some text, adding `L`, etc). And spawning a new compatible NIP for labeling  a group of things (see people lists above). 

Ask questions if any of that doesn’t make sense. Apologies if I misunderstood anything.

Gruruya:
>Problem with separating `L` from `l`

I believe seperating to be best, you can choose to query
- Based on label
- Based on label + namespace
- Or even based only on namespace

If we really want to prevent abuse, we can get rid of the third `l` namespace field and just apply the `L` tag to all of the labels in the event. But this would make it so you could only have one namespace per event and remove some neat use-cases I'm sure. The `>` or `/` seperator system creates some developer overhead (potential bugs,) and only allows for filtering "Based on label + namespace"

>Relays SHOULD delete related labels when deleting the item that was labeled. Keeping labels for something that’s deleted is pointless.

I see use in it; you can still get the event from other relays and a label could be useful there. Of course if space is a large concern to the relay developer he should make all of the optimizations he can, but there's trade offs between the two approaches so I don't see the sense in enforcing one or the other.

>_Content Discovery Issues_

I don't believe this applies, you can't label something without referencing it in the current system. Also NIP-19 `naddr`, `nprofile`, etc. should be avoided for stuff like this as it states "These formats are not meant to be used anywhere in the core protocol."

>I’m strongly in favor of replaceable events.

I'm completely onboard as well.

s3x-jay:
Some of my points don’t seem to have been clear…

I’m fine with having both `l` an `L` if someone sees value in it (I don’t) but I can’t see the point of ever searching for the some label in _every_ namespace. E.g. searching for “LA” would return Los Angeles, Louisiana, Laos, and god knows whatever else. Why would you want all of that junk in your search results? It doesn’t scale well.  IMHO you only ever search within namespaces, never across them. That, and the fact that it invites abuse…

Oh, and I’d bet there are coding systems which use /, which is why I didn’t use it. Much less likely that they use >. 

Deletions & SHOULD - I think the issue is that if a delete was issued it should be honored anywhere. I’m very uncomfortable on us planning to do something against the users’ will. And it’s basic relational database stuff to not leave orphaned records lying around. I don’t think you are really grasping the scale issues… Scaling something messy is inviting disaster. Again, this is a vector for abuse. Spam gets labeled. The spam gets deleted but the labels (which take 10x the space remain). I stand by SHOULD on this. 

I was a bit unclear on the use of `nevent`, etc. They can’t be used for queries, so they should only be used redundantly with the corresponding `p` or `e` or whatever. 

But you absolutely need to be able to find the content being labeled. So either you need a suggested relay url or you need `nevent` et al. which specify multiple relays where the data can be found. Leaving the client to do a needle in a haystack search and hope they find it on whatever relays they use is a highly inefficient use of Nostr.

s3x-jay:
BTW, in an ideal world where more than one parameter of a tag was indexed, yes, I’d be 100% on your side with separating the namespace and code. (Which I think was your original intention). It’s messy combining them. But the only other way to resolve the query problem is to only have one label per labeling event. I wouldn’t dismiss that idea. But we’ve all considered both options and wound up wanting multiple labels per event. Still, one per event is a workable/viable option.

s3x-jay:
Question… NIP-12  doesn’t mention “starts with” queries (e.g. `SomeColumn` LIKE “SomeVale%” in SQL). Are they not supported? They’re pretty basic and not intensive to serve. If they’re not an option it does change how we approach things here.

s3x-jay:
> Yeah, querying is very basic. You can only query with full matching OR across one tag type and with AND across multiple tags. So a filter could be `l: label1 OR label2 AND L: namespace1 OR namespace2`
> 
> > But the only other way to resolve the query problem is to only have one label per labeling event.
> 
> I think we could recommend using a single namespace per event for a similar effect.

It’s sucks there’s no starts with searches. Each time I think up a path forward it doesn’t work. Single namespace breaks on self-reporting in the event. Having the option of multiple namespaces but a single label (eg sone of the ISO codes that are the same in similar coding systems) - that would complicate a `d` tag. (The namespace probably has to be in the `d` tag). 

> > E.g. searching for “LA” would return Los Angeles, Louisiana, Laos, and god knows whatever else.
> 
> It's a niche use-case but it's something. Also seperating `l` and `L` makes querying across multiple namespaces at once simpler.
> 
> > So either you need a suggested relay url
> 
> Ah, I see what you're saying. You're right we should recommend that it's included.

So more to ponder on…

Gruruya:
>It’s sucks there’s no starts with searches. 

How am I supposed to implement that in LMDB, y'know?

>The namespace probably has to be in the d tag

That's a neat idea. Would require only allowing a single namespace like you said. Technically there's possibility of collision as the reference also included in the `d` tag can be an arbitrary `t` tag, but I think it can be overlooked.

s3x-jay:
> > It’s sucks there’s no starts with searches.
> 
> How am I supposed to implement that with LMDB, y'know?

Just as not all relays support NIP-12, starts with queries is something that just certain relays could support.

staab:
Catching up here:

> NIP-01 mentions optional for e & p tags. They should be strongly recommended in these cases.

Added that, same way as NIP-01 handles it, rather than how NIP-19 recommends, just for simplicity/compatibility with existing tag parsing.

> Relays SHOULD delete related labels when deleting the item that was labeled. Keeping labels for something that’s deleted is pointless.

Maybe, but I don't think that belongs in this NIP. It's more of a heuristic for deletion in general.

> E.g. searching for “LA” would return Los Angeles, Louisiana, Laos, and god knows whatever else.

Then clients shouldn't do that. But I do think more non-namespaced label queries could be useful in a less-formal context. E.g. "spam" is likely to exist in multiple vocabularies, and is relatively well defined. I don't see it as a problem at any rate.

> reviews should go in content

Agreed, clarified this.

> why not add `l` tags to user lists

Users have expectations about lists. It would be weird for a client to label a bunch of stuff, and some strange generated list appears in user lists. Like topics, lists should probably be considered part of the end-user experience. 

— unless the user is labeling the list itself, in which case that label would implicitly refer to the list contents as well. But that would be the self-assign use case already supported. At any rate, I'm not sure how that's different from the current state of this NIP, the two events are structured very similarly, except that we have to figure out some reasonable way to assign `d` when multiple targets are involved.

Maybe I just don't understand the use case you guys have in mind for replaceable labeling. Obviously, it could make sense to use a replaceable event when leaving a review, so subsequent reviews could replace older ones (i.e. review edits), in which case the `d` tag could simply be the target's value. But it sounds like you're talking about some kind of named list, where multiple targets are labeled at once, but it's replaceable? I just don't know what that would look like. But I also don't know if it matters, clients can come up with whatever `d` tag they want, the heuristic for these complex cases doesn't necessarily have to be defined. Feel free to clarify with an example.

I've added the following stuff on `d` tags to the PR, but I have the feeling it's not what you guys are fishing for, since the way I have it set up is multi-tagging and replaceability are mutually exclusive:

> Any number of targets may be included when using a kind `1985` non-replaceable event. If kind
> `32144` is preferred, the event MUST have a single target tag, and its value MUST also be the
> value of the `d` tag.

Gruruya:
The way I see it:

>["L", "#t"], ["l", "bitcoin", "#t"]

The event contains hashtag labels, no bearing on `.content`. The referenced things have a bitcoin hashtag.

>["L", "review"], ["l", "relay", "review"]

The event contains review labels, `.content` must contain the review. The referenced things have a relay review.

The only difference is that the `review` namespace defines the `.content`, which I don't see the problem with.

If you want to get rid of the somewhat redundant `l` tag you would have to describe a different system, but I think it's good as is because including the `l` tag lets you review things other than "what I'm referencing."

As I don't see how you would structure it differently and we already have a way to distinguish different types of labels, I don't think you need to add another kind. Reviews are analogous to labels in describing another thing, so they both benefit from tags like `quality`/`confidence`. Seem similar enough to keep together IMO.

s3x-jay:
Another way to look at it…

1. Structured labels from a defined vocabulary 
2. Unstructured tagging (short, no spaces - e.g. hashtags)
3. Short phrases (no longer than “alt text”)
4. Longer text (reviews)

The first three could go in an `l` tag. But you wouldn’t want the 4th. Even the third isn’t ideal given that there’s no starts with or contains searching. 

The other way to look at it is reviews are something that could be labeled. They aren’t labels. Except if you define a simple vocabulary for them. Something as simple as what’s defined for reactions could work. +, -, or a single emoji. 

So imagine a restaurant review. The label is 🤮, and then in `contents` the person could give their full review.

Gruruya:
We would probably want to add ~ for between + and - as reviews have more reason to be "inbetween" than reactions.

staab:
I think I'll remove the `review` example from this NIP, but if someone (me) wants to misuse labels to implement reviews, I guess no one will be too mad. Are we generally happy with this PR? I'll implement relay reviews at some point and then we can talk about merging it.

Gruruya:
If we're talking about misuse, I plan on labeling labels, specifically with descendent labels (label1 implies label2) and alias labels (label1 is the same as labelOne.) Probably with the namespaces `#l`, `#l=`, and `target` for the referenced label. Regardless, the structure of the event looks good to me.

staab:
Mad lad

Gruruya:
The obvious problem left-in right now is querying events with multiple namespaces with colliding labels.  
The `l` `namespace:label` hack doesn't have this problem, and is the most obvious solution. I assume people will do this, so we should recommend when doing it to ALSO include an `l` `label` tag without the namespace for compatibility.  
Maybe put it as a footnote, most events with multiple namespaces won't have labels collide often enough to care.

Gruruya:
The problem left in right now is querying events with multiple different namespaces and the same labels will return extra results.  
The solution is the `namespace:label` layout, we should put something in a footnote so people use the same separator/format.
Most events with multiple namespaces won't have many conflicting labels, so they won't care.

32.md
```diff
@@ -0,0 +1,176 @@
...
+This NIP does not supersede NIP-56, which supports reporting content for the purpose of
+direct moderation, in order to comply with laws or app store requirements. "Moderation"
+as defined by this NIP is only relative to user preferences and should be interpreted
+with the social graph in view to provide a better user experience.
+
+Label Target
+----
+
+The label event MUST include one or more tags representing the object or objects being
+labeled: `e`, `p`, `r`, or `t` tags. This allows for labeling of events, people, relays,
```
verbiricha, line 24: Should we also include `a` tags for NIP-33 parameterized replaceable events?

nostrband:
I've added 1985 labels to nostr.band, just fyi, needed labels for some other thing

Gruruya:
Namespaces are potentially conflicting, how should this be solved? Special namespace formats like ISOTime ISODate?  
Should people opt to use the `l` `namespace/label` format? A third queryable field? I guess we can just leave it as implementation dependant if we want. Fixing this was the intention behind the `L` tag originally.

It can be worked around by using people lists and querying based on `pubkey` and such, but there's flaws with that. Not sure, WDYT?

staab:
Spoke with @mplorentz this afternoon and hammered out a few more things:

- `d` tags are inherently problematic. There might be multiple clients with different use cases that want to let a user label the same target, so target id isn't sufficient. Adding label in to the d tag makes the event effectively non-replaceable. Maybe id + namespace could work, but that just reduces rather than eliminates the possibility that one client will nuke another's label. The only scheme that I think could work would be something like `client/usecase/targetid`. If that's the case, we should just let people do arbitrary `d` tags. There's no good solution, so I've left replaceable events completely out. If clients want to only display the most recent label for something, they can do that. Or, they can issue delete requests for old events. Not a great solution, but lets us move forward.
- I think that @s3x-jay is right about quality/confidence/etc being applied to each individual label. I had moved those out because json in json is gross. I've put them back into the `l` tag, this time as URL encoded query strings to avoid the json-in-json problem, although I admit it's probably equally ugly. I've also clarified that any key can be used in the additional data part.
- I added `a` to the list of target tag types.

Gruruya:
>URL encoded query strings

We should probably avoid introducing new parsing requirements, JSON-in-JSON is already used in a few places.

staab:
😢 very well

Gruruya:
> :cry: very well

Look at the bright side, we can use `["p", <pubkey>]` instead of NIP-19 urls.

>`d` tags are inherently problematic

Good thinking.

I'll summarize my past two posts: we should warn about namespaces conflicting, or come up with some other solution to that problem. One solution is a *cough* another field for namespace groups similar to the original `L` tag.

Also I think nostr.band is making posts without any `L` tags, should we describe that in the NIP?

staab:
That could be improved, I'll add a recommendation to "make sure" namespaces are unique. Reverse domain names are a common strategy for this. If there's a collision, the person defining the namespace chose poorly, and I think that's fine.

Leaving out L tags is already prohibited, since even if you assume `ugc` you can't search using the namespace. @nostrband do you mind adding L tags? If L tags are omitted, clients that query using them simply won't find the tags they're looking for.

staab:
Oh, and from my conversation with @mplorentz, we should probably wait to merge this PR until it's implemented by nos.social for moderation, and by coracle.social for relay reviews, which will probably take a couple weeks to happen.

32.md
```diff
@@ -0,0 +1,172 @@
...
+- `quality` may have a value of 0 to 1. This allows for an absolute, granular scale that can be represented in any way (5 stars, color scale, etc).
+- `confidence` may have a value of 0 to 1. This indicates the certainty which the author has about their rating.
+- `context` may be a comma-separated list of urls (including NIP-21 urls) indicating other context that should be considered when interpreting labels.
+
+Content
+-------
+
+Labels should be short, meaningful strings. Longer discussions, such as for a review, or an
+explanation of why something was labeled the way it was, should go in the event's `content` field.
+
```
mplorentz, line 63:
```suggestion
Label Deletion
---------------

Clients that support labels SHOULD also support their deletion via [NIP-09](https://github.com/nostr-protocol/nips/blob/master/09.md) events. Labels added at the time of publishing can only be deleted by deleting the original event.
```

mplorentz: Maybe this is superfluous? But given all the debate between using replaceable events for labeling maybe it's worth calling out? I could go either way.

mplorentz (commented):
As @staab mentioned we are going to use labels with NIP-69 vocabulary for our first pass on decentralized moderation and content warnings. I left a few suggestions to make this more compatible with that use case. 

Also for transparency, some conversation regarding this PR is happening over on [Zulip](https://nips.zulipchat.com/#narrow/stream/389765-NIP-32.2F68-Labels-.28PR-.23532.29/topic/combining.2032.20and.2068).

brugeman:
> Leaving out L tags is already prohibited, since even if you assume `ugc` you can't search using the namespace. @nostrband do you mind adding L tags? If L tags are omitted, clients that query using them simply won't find the tags they're looking for.

Done, added "ugc" category, is this correct? e7f66df9562f89aef8ce07bc8c18e8de1aab8e30519c48a42e17d829a7f39e67 at wss://relay.nostr.band

32.md
```diff
@@ -0,0 +1,172 @@
...
+  "tags": [
+    ["L", "my-lightning-nomenclature"],
+    ["l", "channel", "my-lightning-nomenclature"],
+    ["p", <pubkey1>, <relay_url>],
+    ["p", <pubkey2>, <relay_url>]
+  ],
+  "content": "<channel_id>",
+  ...
+}
+```
```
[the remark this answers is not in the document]
arthurfranca: Allowing labels directly on event kinds other than 1985 means that "L" and "l" will have a reserved meaning, no matter the kind. So, for example, nobody may use `l` meaning "language" on a kind xx.

Maybe in your example its best to create a kind 1 and also a kind 1985 event.

32.md
```diff
@@ -0,0 +1,172 @@
...
+
+`draft` `optional` `author:staab` `author:gruruya` `author:s3x-jay`
+
+A label is a `kind 1985` event that is used to label other entities. This supports a number of use cases:
+
+- Distributed moderation and content recommendations
+- Reviews and ratings
+- Definition of edges in a graph structure
+
+This NIP does not supersede NIP-56, which supports reporting content for the purpose of
```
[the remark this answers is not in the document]
staab: I mostly wanted to avoid pushback based on adding a second way to do the same thing. I agree this paragraph is sort of BS, but it seems worth keeping.

32.md
```diff
@@ -0,0 +1,172 @@
...
+labeled: `e`, `p`, `a`, `r`, or `t` tags. This allows for labeling of events, people, relays,
+or topics respectively. As with NIP-01, a relay hint SHOULD be included when using `e` and
+`p` tags.
+
+Label Tag
+----
+
+This NIP introduces a new tag `l` which denotes a label, and a new `L` tag which denotes a label namespace. A label MUST include a mark matching an `L` tag. `L` tags refer to a tag type within nostr, or a nomenclature external to nostr defined either formally or by convention. Any string can be a namespace, but publishers SHOULD ensure they are unambiguous by using a well-defined ISO standard or reverse domain name notation. Some examples:
+
+- `["l", "footstr", "#t"]` - the publisher thinks the given entity should have the `footstr` topic applied.
```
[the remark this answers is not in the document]
staab: The second argument, which is `#t`. This is a special namespace indicating "nostr topic".

32.md
staab: I think it is, this is just the way events work.

32.md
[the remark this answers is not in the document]
staab: From the NIP:

> `l` and `L` tags MAY be added to other event kinds to support self-reporting. For events with a kind other than 1985, labels refer to the event itself.

If someone wants to designate language, a new tag isn't necessary, they can do `["l", "en", "ISO 639-1"]`.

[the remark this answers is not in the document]
arthurfranca: "Language" was just an example, it could mean anything in another kind context. It was just a side note for you to consider. Not against it I'm neutral on this specific topic.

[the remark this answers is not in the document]
staab: Just added an example for self-tagging.

32.md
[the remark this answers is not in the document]
staab: Added a clarification for this.

32.md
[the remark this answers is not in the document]
staab: > "Language" was just an example

Gotcha, I think this is ok, but if you see a bigger potential problem let me know. Do we already use "language" tags?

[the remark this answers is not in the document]
arthurfranca: > Do we already use "language" tags?

There is [this PR](https://github.com/nostr-protocol/nips/pull/182). There's other open issue with `l` meaning "tree (l)evel/depth". But nothing using the `l` tag that I know became a NIP yet, so no problem.

32.md
```diff
@@ -0,0 +1,172 @@
...
+Content
+-------
+
+Labels should be short, meaningful strings. Longer discussions, such as for a review, or an
+explanation of why something was labeled the way it was, should go in the event's `content` field.
+
+Example events
+--------------
+
+A report that an event contains nudity. Note that NIP 56 is preferred for reporting content
```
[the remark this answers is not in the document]
erskingardner: Agree, I think it would be better to use a different example.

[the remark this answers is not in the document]
staab: Sure, removed the example.

32.md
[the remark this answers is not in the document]
staab: Alright, I've removed it, we'll leave it up to people to interpret.

32.md
```diff
@@ -0,0 +1,166 @@
...
+{
+  "kind": 1,
+  "tags": [
+    ["L", "social.nos.ontology"],
+    ["l", "IL-frd", "social.nos.ontology"]
+  ],
+  "content": "Send me 100 sats and I'll send you 200 back",
+  ...
+}
+```
```
fiatjaf, line 166:
```suggestion
```

For simplicity, we should have a single way of doing things.

mplorentz: I see your point - semantically these are two ways of arriving at the same state. However I think being able to label something in the original event is desirable, and for some use cases it's necessary.

The immediate need in my mind is using labels for content warnings. If I'm publishing a note with a content warning on it I don't want that label stored separately from the event. It creates lots of failure modes like: 
- some relays could store only the note and not the 1985 label event. Either accidentally or maliciously.
- clients that support content warnings would need to wait until all relays have confirmed that they don't have any associated 1985 events before displaying any note - creating a huge overhead. 

Content warnings are one case but there are others where having the label and event in the same place is desirable. For instance I could use labels to publish notes under different "streams": one for my notes on gardening, one for my notes about video games, etc. Being able to fetch the notes for a stream in a single REQ is much preferable to needing to fetch all the 1985 events for a stream and then individually request all the events in the stream.

fiatjaf: There is a NIP for content warnings already. Is it part of this NIP here that we will deprecate that one?

For "streams" shouldn't you be using `t` tags?

staab: This NIP informally intrudes on NIP 32 and 52's territory. The distinction is that 32 and 52 are designed for creating a very simple flag for clients/relays to manually moderate or hide stuff. This NIP *could* be used for that as well, but is intended to support more robust decentralized moderation — where users choose what vocabularies they care about. NIP 32 and 52 should still be used to flag stuff for centralized-style moderation or hiding in user feeds.

That's not to say I don't think 32 and 52 are bad and should be deprecated, but I don't want to force client authors to change how they do things if it's working for thm.

mplorentz: I haven't seen the other NIP for content warnings. I'll check it out.

earonesty:
note a lot of the conversation took place here: https://github.com/nostr-protocol/nips/pull/46

32.md
mplorentz: NIP-36 doesn't have a space for a machine-readable category, which is important for how we want to implement them. So I'm keen to keep the self-tagging in this NIP. 

(Since NIP-36 is already being used by amethyst we'll probably use the `content-warning` tag in addition to `l` and `L` tags when publishing a note with a content warning for compatibility.)

s3x-jay:
Sorry for being absent from the discussion, but getting back up to speed is taking forever after getting out of the hospital. Here's my thoughts...

1. The lack of starts with searches is a real problem, but it will get fixed eventually. That said, there's a lot in this NIP that's working around that problem. I think we should consider what will be best when starts with is possible. In particular changing parameter 2 to "Namespace>Code" would help immensely once starts with is supported by some relays.
2. Matt explained to me how delete + new event is as good/better than PREs... I just think we should mention the process.
3. Suggestion - reviews are pretty universal so a reverse domain namespace doesn't make sense to me. How about namespace = "review", code = "ugc"?
4. Line 19 - `a`tag should be taken out of list since it's not searchable. A separate section should encourage it's use with `nevent`, etc. for finding labeled items.
5. Small - lines 18-21 need to be combined.

32.md
fiatjaf: What about editing NIP-36 to add the machine-readable category on an optional tag item and link it to this NIP? I think that would be the best, unless I'm missing something.

s3x-jay: > What about editing NIP-36 to add the machine-readable category on an optional tag item and link it to this NIP? I think that would be the best, unless I'm missing something.

NIPs 36 & 56 don’t use single letter tags, hence they are searchable. Ergo neither really is capable of doing the job it needs to do.

staab:
1. I think we're good here, this topic has been hashed out pretty well.
2. Cool, added a note at the bottom.
3. I played with designating a special purpose `review` namespace, but reviews are just a weird thing I want to do with this NIP, and not what it's about, so I didn't want to give it special treatment. Feel free to use an informal `review` namespace as well (I likely will do that as well).
4. I don't understand removing `a`, it seems perfectly valid to label a PRE
5. Convention for formatting markdown in this repo seems to be converging on breaking lines manually (they are automatically rendered as a word-wrapped paragraph).

I've also updated NIP 36 to refer to this NIP to qualify the `content-warning` tag per @fiatjaf's request.

32.md
staab: I think I get what @fiatjaf means, I've added a note to NIP 36 and 56.

staab:
> Right now you can’t query for a particular code in a particular namespace.

What's wrong with `{"#l": ["tag"], "#L": ["namespace"]}`?

> understanding is that a is only used for things like nevent, nprofile, etc.

I think you're confusing NIPs [33](33.md) and [19](19.md). `a` is for stably referencing a parameterized replaceable event.

staab:
> Is a ever used without also specifying e or p?

Yes, see nip 33.

> It may return things other than what you’re looking for.

I don't see how, multiple filter keys are AND'd together, so the above query is equivalent to a single `l` query with a fully namespaced tag.

Gruruya:
If you MUST have multiple labels with the same value and different namespaces put in the same event you could duplicate it  
`tags: [["l", "label", "namespace"], ["l", "namespace>label", "namespace"], ["L", "namespace"]]`  
or something or other. There wouldn't be much point in keeping the current `["L", "namespace"]` without `["l", "label"]`

>multiple filter keys are AND'd together

He's talking about not being able to filter `namespace/label` in an event with multiple namespaces and duplicate labels.  
You should note somewhere that should only be done when you WANT that filtering behavior.

s3x-jay:
@Gruruya - I need to be able to use labels created by others and can't know how they specified them.

@staab - Case: Since "IT-MI" can't be queried to find IT% = Italy, people will label for country, state & city. But then when they query for "LA" they get everything for Laos, Louisiana, Los Angeles, and more. There's no way to get just one.

staab:
I still don't understand, can you write a detailed example?

s3x-jay:
> I still don't understand, can you write a detailed example?

No, I can’t. As you know I’ve had a stroke and typing is difficult for me.

staab:
I don't understand what 23 is doing, but from 58:

> Zero or more ordered consecutive pairs of a and e tags referencing a kind 30009 Badge Definition and kind 8 Badge Award, respectively.

That means the `a` tag is sufficient to refer to a 30009 event. You can see the `Example of Badge Award event` for an example of this.

Per the requested example, let me try. If I were to tag louisiana and los angeles in the same event, I would do `[["l", "22", "fips.statefp"], ["L", "fips.statefp"], ["l", "02410877", "fips.placefp"], ["L", "fips.placefp"]]` in which case there's no conflict. If for the publisher is more informal and instead uses `[["l", "LA", "state"], ["L", "state"], ["l", "LA", "city"], ["L", "city"]]`, you can still REQ `{"#l": "LA", "#L": "state"}` to find the event, then `tags.find(t => t[0] === "l" && t[2] === "state")` to get the desired tag. Is there something I'm missing?

[Source](https://www.census.gov/library/reference/code-lists/ansi.html)

s3x-jay:
> I still don't understand, can you write a detailed example?

Try to find one of the LA codes without returning the others...

```json
{
  "kind": 1985,
  "tags": [
    ["e", <id>, <relay_url>],
    ["p", <id>, <relay_url>],
    ["L", "country"]
    ["L", "province"]
    ["L", "#t"]
    ["l", "LA", "country"],
    ["l", "HO", "province"],
    ["l", "XamNeua", "#t"]
  ],
  "content": "",
  ...
}
```

```json
{
  "kind": 1985,
  "tags": [
    ["e", <id>, <relay_url>],
    ["p", <id>, <relay_url>],
    ["L", "country"]
    ["L", "province"]
    ["L", "#t"]
    ["l", "US", "country"],
    ["l", "LA", "province"],
    ["l", "NewOrleans", "#t"]
  ],
  "content": "",
  ...
}
```

```json
{
  "kind": 1985,
  "tags": [
    ["e", <id>, <relay_url>],
    ["p", <id>, <relay_url>],
    ["L", "country"]
    ["L", "province"]
    ["L", "#t"]
    ["l", "US", "country"],
    ["l", "CA", "province"],
    ["l", "LA", "#t"]
  ],
  "content": "",
  ...
}
```

Gruruya:
You would have to seperate the labeling events based on namespace as the label formats are potentially conflicting and you don't want them to be equivilent.

s3x-jay:
> You would have to separate the labeling events based on namespace as the label formats are potentially conflicting and you don't want them to be equivalent.

Or change parameter 2 of `l` to Namespace>Code. Then only search on `l`.

Gruruya:
What benefit does that have? I only see downsides. Separating based on namespace isn't that big a deal.

s3x-jay:
> > Or change parameter 2 of l to Namespace>Code. Then only search on l.
> 
> What benefit does that have? I only see downsides. Separating based on namespace isn't that big a deal.

Benefit is huge - clean, efficient query results.

Can you list the downsides? I can’t think of any.

Gruruya:
>Can you list the downsides?

You can't query a label across multiple namespaces at once.  
If you don't understand why you would want to, take reverse-domain name namespaces for an example.

It requires parsing/formatting and that format could be ambiguous with your label.

That said, nothing is stopping you from doing it. If you feel the need to put conflicting namespaces in the same event and you don't want them to conflict, a unique incompatible `l` format is how you would do it. I simply recommend also including the plain `l` tag so that people may still query for it across namespaces.

Here's an example

staab:
I see what you mean, that's a very good point. @Gruruya is right though, it would be easy to create a namespace that had to versions of each label, qualified and unqualified, and we can have our cake and eat it too.

s3x-jay:
> Reverse-domain name namespaces as in equivalent labels from different clients. So `social.coracle.recipe` vs `social.nos.recipe` or whatever. Searching across multiple standards is necessary if we want people to properly separate their labels. For standards we all agree on exactly it's not necessary, but in practice this will rarely happen and similar labels will be put under different namespaces.

Unless `social.coracle.recipe` and `social.nos.recipe` use the same codes, that example makes no sense. If they do, then they shouldn't use different namespaces. 

> > In what way is it ambiguous?
> 
> Ambiguity is only an issue if we don't have one single standard, you're right. But again it's useless when we can just separate events based on namespaces that could conflict.
> 
> > > Separating label events based on potentially conflicting namespaces does this as well.
> > 
> > Agreed, but everyone must do it for it to work.
> 
> Lets agree and put a recommendation for it then.

Just remembered it won't work for labeling on the original event.

Gruruya:
>Unless social.coracle.recipe and social.nos.recipe use the same codes, that example makes no sense. If they do, then they shouldn't use different namespaces.

They would have mostly the same codes. In some contexts it would be better to be more strict and limit exclusively to the local namespace and others you would want to be more broad.

>Just remembered it won't work for labeling on the original event.

Yes that's right, in the situation where the original event needs multiple potentially conflicting namespaces at creation-time due to race conditions AND the labels conflict, it would return an extra event.

Gruruya:
>Unless social.coracle.recipe and social.nos.recipe use the same codes, that example makes no sense. If they do, then they shouldn't use different namespaces.

They would have mostly the same codes. In some contexts it would be better to be more strict and limit exclusively to the local namespace and others you would want to be more broad. For instance imagine coracle and nos vehemently disagree on labeling pizza as pie so they can't use the same namespace.

>Just remembered it won't work for labeling on the original event.

When a filter matches an event which was posted with multiple, potentially conflicting namespaces which had to be included with the event to avoid race conditions and the labels conflict, the filter would return an extra event which your client would ignore as its third fields would not match your query. This exception is rare and the namespace>label optimization is not necessary as even in this situation, the extra matches are easy to deal with.

s3x-jay:
> > Unless social.coracle.recipe and social.nos.recipe use the same codes, that example makes no sense. If they do, then they shouldn't use different namespaces.
> 
> They would have mostly the same codes. In some contexts it would be better to be more strict and limit exclusively to the local namespace and others you would want to be more broad. For instance imagine coracle and nos vehemently disagree on labeling pizza as pie so they can't use the same namespace.

Change `l` = (nos|coracle) and `L` = SharedCode
to `l` = (nos>SharedCode|coracle>SharedCode)
and problem solved. No?

I still only see advantages to namespace>code.

> > Just remembered it won't work for labeling on the original event.
> 
> When a filter matches an event which was posted with multiple, potentially conflicting namespaces which had to be included with the event to avoid race conditions and the labels conflict, the filter would return an extra event which your client would ignore as its third fields would not match your query. As this exception is rare and easy to deal with, I don't think the namespace>label optimization is necessary.

That's the dirty search results I was talking about. If you're searching for something rare it can get overwhelmed with stuff you don't want.

arthurfranca:
Maybe the ideal would be this:
```json
{
  "kind": 1985,
  "tags": [
    ["e", <id>, <relay_url>],
    ["L", "country"]
    ["L", "province"]
    ["l", "country>LA"],
    ["l", "province>LA-HO"]
  ],
  ...
}

{
  "kind": 1985,
  "tags": [
    ["e", <id>, <relay_url>],
    ["L", "country"]
    ["L", "province"]
    ["l", "country>US"],
    ["l", "province>US-LA"]
  ],
  ...
}
```

`["L", "xyz"]` "L" is needed cause an app may want so show all xyz labels applied to an event id. startsWith query wouldn't be available on all relays, so "L" tag is good enough and works.

`["l", "country>LA"]` won't have the problem pointed by @s3x-jay of requesting `{ #l: ['LA'] }` and receiving things from other namespaces. In the example, `province>LA-HO` will make sure only HO from LA country will be received when searching for HO province because `{ #l: ['country>US', 'province>LA'] }` is an OR query so the correct would be `{ #l: ['province>US-LA'] }`.

I may be wrong but clients searching for keywords, that don't care about different meanings, already can rely on hashtags `{ #t:  'la' }` or NIP-50 `{ search: 'la' }`. Labels could be a thing for searching for a specific meaning `{ #l: ['province>US-LA'] }`.

The namespaces in the example could even ve like this:

```json
{
  "kind": 1985,
  "tags": [
    ["e", <id>, <relay_url>],
    ["L", "geo"]
    ["l", "geo>LA"],
    ["l", "geo>LA-HO"]
  ],
  ...
}

**Comment by Gruruya** [comment]
>keywords, that don't care about different meanings, already can rely on hashtags

Hashtags have a specific meaning as topics, labels are machine-readable and categorized.

**Comment by s3x-jay** [comment]
> > and problem solved. No?
> 
> Yes, but that will quickly get messy if you're searching many namespaces or labels.

I'm still waiting for a real-world use case...

> Rather than argue whether returning some extra events or being able to easily search across namespaces is more important, you can do both as I said before. However I don't think it needs to be standardized, namespaces' 'l' formats are already unique and you can make it a convention of your namespace to do `namespace>label`. It's only useful in namespaces used together with other namespaces whose values may conflict which can be predicted by the client dev.

I agree. I think we should recommend a value of "namespace>code" for coding systems that are highly unique / rigid / industry standards, and just "code" for more informal labeling.

> Maybe the ideal would be this: ...

I've always been thinking in terms of coding systems that have been around for decades. You don't tell those folks to change to suit Nostr.

**Comment by arthurfranca** [comment]
@Gruruya you are right current spec seems flexible enough. As I added US prefix to the state/province, the previous example could be:
```json
{
  "tags": [
    ["e", "<id>", "<relay_url>"],
    ["L", "country"]
    ["L", "province"]
    ["L", "city"]
    ["l", "US", "country"],
    ["l", "US-LA", "province"]
    ["l", "US-LA-New Orleans", "city"]
  ],
  ...
}
```

motorina0:
Regarding his comment at the end of the NIP-32:
>We have opted not to use parameterizable/replaceable events for this due to the complexity in coming up with a standard d tag.

- The `d` tag can be used to assign `rating mass` (importance/intensity) to a rating and also to prevent spam.
- I think the `rating mass` goes hand-in-hand with this PR.
- Details about `rating mass` can be found here: https://gist.github.com/motorina0/874912282d2419be4eaf89de8d0c4464

arthurfranca:
@motorina0 very clever math. Maybe I misunderstood but isn't this creating a system where only rich people ratings matter?

Unless when I buy an item or use some service, the store/service provider use part of the money to buy me a rating mass tree leaf/root registration.

But yet how do we stop a rich person from buying many rating mass tree root registrations and creating fake positive ratings to his crappy product or fake negative ratings to a good competitor product?

motorina0:
> @motorina0 very clever math. Maybe I misunderstood but isn't this creating a system where only rich people ratings matter?

That is a very good point. I have open this PR: https://github.com/nostr-protocol/nips/pull/604
Can you please add your comment there (so we do not pollute this PR)

staab:
I have implemented reviews on coracle using this NIP. Any final thoughts before I ask another maintainer to merge this?

staab:
Not really, it's sort of a special purpose namespace. We could use `nostr.t` or something, but that seems worse to me. Why would another namespace starting with # be a problem?

Gruruya:
If a `namespace` tag is introduced, legacy `#namespace` labels could be wrongfully interpreted. Also I guess `#t` seems unobvious to me, could do `ntag:t` which looks alright

s3x-jay:
Changes still needed...

Take out all references to content moderation _proposals_ as that will be handled in a separate NIP and differences will cause confusion. Plus, I'm the primary author of that vocabulary, so "social.nos.ontology" would be incorrect since I'm not part of NOS (just working closely with them).

> ensure they are unambiguous by using a well-defined ISO standard or reverse domain name notation.

Suggested change: "ensure they are unambiguous by using a well-known namespace or reverse domain name notation."

Need something along these lines added...

"Before labeling with a namespace one should explore how that namespace has been used previously. In many cases the standard will be to have a label value of 'namespace>label' rather than just 'label' since that allows for more precise queries. This tends to be the case with well-established labeling systems that are industry standards while just using 'label' is often preferred for less formal labeling systems where there may be several systems using similar codes."

Then the examples for MESH, GeoNames, and ISO-3166-2 should all be changed to "namespace>label".

staab:
Was about to switch to `ntag:t` format, when I re-read this line in the PR:

> Namespaces starting with `#` indicate that the label target should be associated with the label's value.

Just seems nice to be able to arbitrarily do `#preimage` or whatnot and have it make sense.

> content moderation proposals

Sure, to be clear I've removed them. I've also added your suggestion to qualify labels at the bottom of the NIP. Any particular reason for the `>` character? I hate to bikeshed, but `.`, `/`, and `:` seem more common for this kind of thing. I personally like `:` since labels can then double as urls using custom url handlers (unlikely to happen, but still, it seems to fit the paradigm). Of course, it really doesn't matter since the bare label can be inferred by just doing `label.replace(namespace, '')`.

s3x-jay:
> > content moderation proposals
> 
> Sure, to be clear I've removed them. I've also added your suggestion to qualify labels at the bottom of the NIP. Any particular reason for the `>` character? I hate to bikeshed, but `.`, `/`, and `:` seem more common for this kind of thing. I personally like `:` since labels can then double as urls using custom url handlers (unlikely to happen, but still, it seems to fit the paradigm). Of course, it really doesn't matter since the bare label can be inferred by just doing `label.replace(namespace, '')`.

The reason I chose > was because it implied the idea and because it was not common. I was hoping it was not likely to be used it the name of a namespace.

staab:
Unfortunately, choosing a delimiter runs up against the problem of distinguishing data and code, hence the idea of separating the namespace and the label in the first place. Luckily, since the namespace can define its vocab, it can use any delimiter it wants. I think we should just not recommend a delimiter.

earonesty:
this is what i was looking for in kind-40 channel creation.   use labels to supply keywords and enable people to search for channels that match their interests by searching for labels

s3x-jay:
> Unfortunately, choosing a delimiter runs up against the problem of distinguishing data and code, hence the idea of separating the namespace and the label in the first place. Luckily, since the namespace can define its vocab, it can use any delimiter it wants. I think we should just not recommend a delimiter.

There are literally 100s (if not 1000s) of existing coding systems that haven’t declared a Nostr delimiter. Our not declaring a default makes it incredibly difficult to use those coding systems efficiently. To me the whole point of this NIP is to be able to use those coding systems. So not declaring a default is a big problem.

staab:
No, it doesn't. If we assume there is a one character delimiter, `qualifiedLabel.replace(/^namespace/, '').slice(1)` gives you the unqualified label.

s3x-jay:
> No, it doesn't. If we assume there is a one character delimiter, `qualifiedLabel.replace(/^namespace/, '').slice(1)` gives you the unqualified label.

Breaks only having to query on `l` - which is why you'd use a delimiter.

SMH...

staab:
NIPs are not meant to be exhaustive, nor prescriptive. Why don't you just go build your implementation and we can update the NIP to suit.

Gruruya:
> NIPs are not meant to be exhaustive, nor prescriptive. Why don't you just go build your implementation and we can update the NIP to suit.

To be honest I've been wanting to shorten the NIP when I've been re-reading it. We could do without so many examples.

s3x-jay:
> NIPs are not meant to be exhaustive, nor prescriptive. Why don't you just go build your implementation and we can update the NIP to suit.

 Because I should be able to use what others build and vice versa. That's kinda the whole point IMHO.

Gruruya:
>Breaks only having to query on l - which is why you'd use a delimiter.

What the hell? Why would you do that? Just use `L`
Anyways the `l` field has its namespace as its third parameter so you're point is moot.

s3x-jay:
> > Breaks only having to query on l - which is why you'd use a delimiter.
> 
> What the hell? Why would you do that? Just use `L` Anyways the `l` field has its namespace as its third parameter so you're point is moot.

Because using `L` returns “dirty” results, as I’ve explained many times now. 

`L` exists because relays don’t support starts with searches (and many never will). `L` is needed when you want to find all the labels using a specific namespace (e.g. for a particular event). It’s not the best way to find all uses of a particular namespace>code combination. For that searching on just `l` is far more efficient.

Gruruya:
>For that searching on just l is far more efficient.

And it call comes down to this. It's simply more efficient for your use-case to use `namespace>label` and for mine to use `label`. Wouldn't it be great if we had a protocol that could accomodate both of us and have them both work?

At this point we're arguing over examples and it's probably better to just merge the thing and let us focus on implementing rather than arguing with each other. Any changes to examples can be made in a separate PR anyways.

s3x-jay:
> NIPs are not meant to be exhaustive, nor prescriptive. Why don't you just go build your implementation and we can update the NIP to suit.

Some additional thoughts…

NIPs are absolutely meant to be both prescriptive and exhaustive - or at least enough so that people all do things in the same way. NIPs are not needed where someone just wants to do their own thing. 

There is a rich history of labeling systems that existed before Nostr - systems that have a long history of collaborative use where people used the same coding system in the same way so they could benefit from each other’s use. The point of this NIP is to be prescriptive and exhaustive enough that that collaboration can continue on Nostr.

staab:
So my question is: in what scenario can you generically construct a namespace-merged `l` query like you're talking about without knowing anything about the vocabulary? In other words, the prefixed `l` tag fixes ambiguity, but you don't know a priori if a given namespace is prefixed or not. So you have to have a list of valid labels within a namespace, either explicitly (the client is coupled to the namespace, or there's a namespace definition document somewhere) or implicitly (by keeping track of labels that have been seen for a given namespace).

I really feel like we're over-thinking this. If clarity is required, bulk tagging across multiple namespaces should just be done using multiple events. Can we just go with that? I admire your thoroughness, but it feels like you're inventing new edge cases in an effort to create the perfect spec, which isn't going to happen in any event, rather than being practical.

> NIPs are absolutely meant to be both prescriptive and exhaustive

This is the W3C way, not the nostr way. Nostr is permissionless, do it how you want, and reconcile differences later. The implementation is the standard, not the spec. Just wait until people start crafting deliberately malformed events.

s3x-jay:
> It's simply more efficient for your use-case to use `namespace>label`. 

I’ve been working with a variety of coding systems/vocabularies for 20-30 years. Every one I’ve encountered benefits from what I’m talking about. 

I have asked repeatedly for a good example one one that doesn’t and haven’t gotten any specific examples.

Gruruya:
Yeah you did?

s3x-jay:
> So my question is: in what scenario can you generically construct a namespace-merged `l` query like you're talking about without knowing anything about the vocabulary? 

I agree which is why I was initially pushing for everything to be prefixed then (when there was pushback) figured a note saying that well-defined coding systems (especially industry standards) assume prefixing. 

>  bulk tagging across multiple namespaces should just be done using multiple events. Can we just go with that? 

The database admin in me balks at the additional db records, but given how tags are indexed, I’m not sure if it’s as big of a problem as I might think it is. So sure…

> Just wait until people start crafting deliberately malformed events.

Trust me - that’s a big concern of mine.

staab:
> Trust me - that’s a big concern of mine.

Me too, should be fun ;)

tyiu:
I read the proposal a few times but I'm not sure if I fully understand it, especially around label queries and how namespaces play into it. It might be useful to see examples of how [NIP-12](https://github.com/nostr-protocol/nips/blob/master/12.md) queries would work with this.

staab:
You can see my implementation of relay reviews [here](https://github.com/search?q=repo%3Acoracle-social%2Fcoracle+review%2Frelay&type=code), which is a nice concrete example. Namespaces exist to disambiguate between similar labels being used for different purposes, and to point to an authority or convention that is the basis for the label semantics. `l` and `L` are separate so you can query by label only (to query multiple namespaces at once at the risk of ambiguity), namespace only (to get all labels for a given vocabular), or both at once (for maximum precision). There are some edge cases @s3x-jay has pointed out having to do with clients tagging with multiple clashing namespaces at once, but they can be solved with some client-side filtering and/or not mixing namespaces with creating an event.

staab:
@fiatjaf would you mind merging this? @tyiu is planning to use it for calendar events as well.

fiatjaf:
I came here to ask if this was in a mergeable state. Are the nos.social people onboard with it too in its current form?

s3x-jay:
> I came here to ask if this was in a mergeable state. Are the nos.social people onboard with it too in its current form?

I’m not technically part of the NOS team, but build 51 of NOS (available since yesterday) uses 32 + 69 for content reporting. I haven’t had a chance to study how NOS’ implementation works exactly but 32’s current form seems “good enough” to me personally. 

I’m currently reworking 69 to reflect the changes from 68 -> 32.

staab:
Just pushed a small update to remove some redundant/exotic examples. I think this is good to go.

earonesty:
All right I'm just still going to use the l tag on my channel creation

I just don't know if it should be uppercase or lowercase

so there's like less work

If people choose to index it and honor a search for keywords cool

mplorentz (approved):
This looks good to me. We have shipped a basic report UI in Nos using this format. It will definitely go through some more iterations but this NIP has everything we need. 

~~I do wish the convention of embedding the namespace in the string with the label (i.e. `["l", "MOD>NS", "MOD"]`) had made it into the NIP, but we're just going to do it anyway and it won't break anything for others.~~ I see that this did make it in at the bottom. Thanks for pointing that out @s3x-jay.

earonesty:
is there some way, in this PR, to give someone a score?   i don't see any ability to give a numeric label

ArcadeCityMayor:
Cool, great work! We'll implement this in Arcade.

Yes this supersedes my placeholder NIP-32 from #46, closed that.

PR #540, opened 2023-05-19
Specify replacement behaviour when replaceable events have the same timestamp

hoytech:
- This is so that relays can converge on a deterministic sets of events, no matter the order they were received
- Otherwise, clients or relays that sync their sets of events could continually retransmit events they think are missing on the other side, wasting bandwidth

### DIFF
--- a/16.md
+++ b/16.md
@@ -20,6 +20,8 @@ Upon a replaceable event with a newer timestamp than the currently known latest 
 effectively replacing what gets returned when querying for
 `author:kind` tuples.
 
+If two events have the same timestamp, the event with the lowest id (first in lexical order) SHOULD be retained, and the other discarded.
+
 Ephemeral Events
 ----------------
 An *ephemeral event* is defined as an event with a kind `20000 <= n < 30000`.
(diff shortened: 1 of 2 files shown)


weex:
Is there a related issue or other discussion pertaining to this behavior?

hoytech:
Not yet! I'll paste what I sent to the strfry telegram chat just now:

> I think we should specify the relay replacement behaviour when two replaceable events have the exact same timestamp. This is so that two relays that are exchanging events (directly via syncing/streaming or indirectly via clients) will always converge on the same set of events, no matter the order they are received.
>
> The current master branch of strfry will just retain whatever event it already has in this case. However, with negentropy syncing, this could cause a relay cluster to constantly be syncing events across and never converging. So, in the next branch I arbitrarily decided to retain the event with the lowest id (first, in lexical order), and discard the other.

"negentropy syncing" is referring to the following effort:

https://github.com/hoytech/negentropy
https://github.com/hoytech/strfry/blob/next/docs/negentropy.md

fiatjaf (approved):

Cameri (approved):

scsibug:
This looks like a good solution to make this more deterministic than simply "first one the relay sees wins".

PR #542, opened 2023-05-19
NIP-60 Zap Gates - Interoperable distributed zap / pay walls 

Egge7:
We propose this flow for protecting resources (visuals, audio, text, html/iframes etc.) on web servers and unlocking them using NIP-57 zaps. This nip also proposes a new kind for sharing pointers and metadata for these resources throughout nostr.

Notes: We are aware that there is currently an ongoing discussion about using `url` vs `u` for tagging URI/URL. ZapGates started out with `url` like in NIP-94, but we changed it to `u` after feedback from @melvincarvalho. This PR is marked as a draft until at least NIP-98 is merged, because it heavily relies on it. Its probably fair to assume that this will by then.

Additional resources:
- Blog post about the implementation: [ZAPGATE - A new way to pay for content on nostr (habla)](https://habla.news/a/naddr1qqxnzd3cxs6rydpnx5mnyvf4qy08wumn8ghj7mn0wd68yttsw43zuam9d3kx7unyv4ezumn9wshszrnhwden5te0dehhxtnvdakz7qghwaehxw309aex2mrp0yhxummnw3ezucnpdejz7q3qmz3vx0ew9le6n48l9f2e8u745k0fzel6thksv0gwfxy3wanprcxqxpqqqp65whugw2f)

- WIP Demo [ZapGate.link](https://zapgate.link)


fiatjaf:
Looks good from a quick glance, but I hope we get more feedback from other people that may be interested in this.

Egge7:
> Looks good from a quick glance, but I hope we get more feedback from other people that may be interested in this.

Yeah, would be great to get some feedback from @verbiricha or @pablof7z or anyone else building a content focused client.

verbiricha:
Pablo will have more insight than me here since zapstr implements a similar thing to this. cc @v0l

fmaul:
Is the note limited to a single resource (i.e. only one u tag) or multiple resources?  IMHO it would be good to be able to post multiple images as a set, or images in combination with a video. It would be more convenient for the zapper to unlock those resources as a package with a single zap.

https://lightning.video/ for example even has multiple unprotected preview images.

Egge7:
> Is the note limited to a single resource (i.e. only one u tag) or multiple resources?  IMHO it would be good to be able to post multiple images as a set, or images in combination with a video. It would be more convenient for the zapper to unlock those resources as a package with a single zap.
> 
> 
> 
> https://lightning.video/ for example even has multiple unprotected preview images.

Yeah that's a great point. I guess it wouldn't be difficult to make it work. Multiple files would then be tied to the same zapgate event and a zap would be unlocking access to all the listed resources.

arthurfranca:
@Egge7 I've added zap gates integration to #547 NIP-96 File Storage PR. But i used the NIP-94 1063 event instead of the 1211 one you use here. A 1063 with an `amount` tag should be considered a zap gated file (similar to how simply a `aes-256-gcm` tag indicates an encrypted file).
Used `url` from 1063 instead of `u` tag. The `preview` tag is a little different (has a third argument with MIME type).

Could you update your PR?

Egge7:
> @Egge7 I've added zap gates integration to #547 NIP-96 File Storage PR. But i used the NIP-94 1063 event instead of the 1211 one you use here. A 1063 with an `amount` tag should be considered a zap gated file (similar to how simply a `aes-256-gcm` tag indicates an encrypted file).
> 
> Used `url` from 1063 instead of `u` tag. The `preview` tag is a little different (has a third argument with MIME type).
> 
> 
> 
> Could you update your PR?

Awesome! I'll look at it asap and update the PR accordingly.

basantagoswami:
I like the idea. Another use case of it can be to create [crowdwalls](https://thebiz.pro/about#crowdwall) like it's done for [The Biz podcast](https://thebiz.pro/). The content gets unlocked for everyone once it's accumulated enough sats. I need to think how we might integrate it with NIP-75 Zap Goals.

But I am also worried about creating more and more non nostr services that interact with the nostr protocol in very specialized ways (not as clients with key pairs, or relays).

vitorpamplona:
@Egge21M are you still working in this idea? If not can we close the PR?

Egge21M:
Have not looked at this for ages tbh. I am fine with closing it.

PR #550, opened 2023-05-23
NIP-23: drafts

verbiricha:
I've implemented drafts in Habla by using the `30024` kind. Using an adjacent kind number seemed like the most straightforward solution but open to other suggestions.

### DIFF
--- a/23.md
+++ b/23.md
@@ -6,7 +6,7 @@ Long-form Content
 
 `draft` `optional` `author:fiatjaf`
 
-This NIP defines `kind:30023` (a parameterized replaceable event according to [NIP-33](33.md)) for long-form text content, generally referred to as "articles" or "blog posts".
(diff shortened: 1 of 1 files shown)


arthurfranca:
It seem alright.

Hey @verbiricha may I ask you how are you dealing on Habla with people editing articles which makes the new event have a more recent `created_at` field, which will mess with sort order, specially if the edit doesn't contain/honor the `published_at` tag?

It is certainly a way to keep one's article always on top of the front page.

verbiricha:
> Hey @verbiricha may I ask you how are you dealing on Habla with people editing articles which makes the new event have a more recent created_at field, which will mess with sort order, specially if the edit doesn't contain/honor the published_at tag?

Screenshot 2023-05-22 at 10 44 34 I display both `published_at` and latest `created_at` if they don't match, otherwise just `published_at`.


>It is certainly a way to keep one's article always on top of the front page.

You are correct. If this becomes an issue I can have separate sections for "New" and "Updated" articles.

PR #552, opened 2023-05-23
Adds Zap splits to NIP-57

vitorpamplona:
1. Changes NIP-57's `zap` tag in events from one to many 
2. Adds an optional percentage for each zap receiver. 
3. Adds dynamic resolution of the lightning address of the receiver by `zap` tagging with pubkey
4. Adds the suggestion to clients to display the Zap split on the screen.

### DIFF
--- a/57.md
+++ b/57.md
@@ -173,9 +173,9 @@ When an event includes one or more `zap` tags, clients SHOULD calculate the lnur
 ```json
 {
     "tags": [
-		[ "zap", "LNURL..", "lud06", 10 ],  // 10%
+        [ "zap", "LNURL..", "lud06", 10 ],  // 10%
         [ "zap", "pablo@f7z.io", "lud16", 65 ], // 65%
-		[ "zap", "460c25e682fda7832b52d1f22d3d22b3176d972f60dcdc3212ed8c92ef85065c", "pubkey", 25 ] // 25%
+        [ "zap", "460c25e682fda7832b52d1f22d3d22b3176d972f60dcdc3212ed8c92ef85065c", "pubkey", 25 ] // 25%
     ]
 }
 ```
--- a/57.md
+++ b/57.md
@@ -180,7 +180,7 @@ When an event includes one or more `zap` tags, clients SHOULD calculate the lnur
 }
 ```
 
-When the hex code is used, clients MAY inform the user the zap split configuration in the note. 
+Clients MAY display the zap split configuration in the note.
 
 ## Future Work
 
--- a/57.md
+++ b/57.md
@@ -168,15 +168,19 @@ A client can retrieve `zap receipt`s on events and pubkeys using a NIP-01 filter
 
 ### Appendix G: `zap` tag on zapped event
 
-When an event includes a `zap` tag, clients SHOULD calculate the lnurl pay request based on it's value instead of the profile's field. An optional third argument on the tag specifies the type of value, either `lud06` or `lud16`.
+When an event includes one or more `zap` tags, clients SHOULD calculate the lnurl pay request based on their value instead of the profile's field. The tag's second argument is a `lud06` address, `lud16` identifier or `hex` string of the receiver's pub key. An optional third argument specifies the type of the previous value: `lud06`, `lud16` or `pubkey`. An optional fourth parameter specifies the weight (a generalization of a percentage) assigned to the respective receiver. Clients should parse all weights, calculate a sum, and then a percentage to each member. If weights are not present, CLIENTS should equally divide the zap amount to all zap receivers. If weights are only partially present, receivers without a weight should not be zapped (`weight = 0`).
 
 ```json
 {
     "tags": [
-        [ "zap", "pablo@f7z.io", "lud16" ]
+		[ "zap", "LNURL..", "lud06", 10 ],  // 10%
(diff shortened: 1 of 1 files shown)


Egge7:
Love it. Aligned with all of these changes. I proposed something similar in the original zap-tag PR and also made it part of [ZapGates](https://github.com/nostr-protocol/nips/pull/542)

I even think that the use of lud06 and 16 should be discouraged because it can lead to stale zap targets on non-replaceable events.

57.md
```diff
@@ -168,16 +168,20 @@ A client can retrieve `zap receipt`s on events and pubkeys using a NIP-01 filter
...
 ### Appendix G: `zap` tag on zapped event
 
-When an event includes a `zap` tag, clients SHOULD calculate the lnurl pay request based on it's value instead of the profile's field. An optional third argument on the tag specifies the type of value, either `lud06` or `lud16`.
+When an event includes one or more `zap` tags, clients SHOULD calculate the lnurl pay request based on their value instead of the profile's field. The tag's second argument is a `lud06` address, `lud16` identifier or `hex` string of the receiver's pub key. An optional third argument specifies the type of the previous value: `lud06`, `lud16` or `pubkey`. An optional fourth parameter specifies the weight (a generalization of a percentage) assigned to the respective receiver. Clients should parse all weights, calculate a sum, and then a percentage to each member. If weights are not present, CLIENTS should equally divide the zap amount to all zap receivers. If weights are only partially present, receivers without a weight should not be zapped (`weight = 0`).
 
 ```json
 {
     "tags": [
-        [ "zap", "pablo@f7z.io", "lud16" ]
+        [ "zap", "LNURL..", "lud06", 10 ],  // 10%
```
Cameri, line 176: LGTM but the number must be a string as tags can only contain strings
```suggestion
        [ "zap", "LNURL..", "lud06", "10" ],  // 10%
```

staab:
Can we not do integer percents? Percents are really weird because they're essentially for display. I'd strongly prefer a number from 0 to 1.

vitorpamplona:
> Can we not do integer percents? Percents are really weird because they're essentially for display. I'd strongly prefer a number from 0 to 1.

Then we need to come up with rules for what happens when the amounts do not match 100%. The weight system is just a generalization of percentages to avoid worrying about mistakes in the client creating those weights.

Egge7:
> Can we not do integer percents? Percents are really weird because they're essentially for display. I'd strongly prefer a number from 0 to 1.

They are not percents, but weights. So you could do 1, 1, 2 which would mean 0.25, 0.25, 0.5

57.md
vitorpamplona: Good catch @Cameri I just fixed this.

jb55:
why do we need lud16,lud06? can't we just have pubkey so we don't have to handle N different UI variations for this?

Egge7:
> Interesting! Leave it as is then, that's a clean solution. Maybe explain that in the NIP.

@vitorpamplona maybe using weights that add up to exactly 100 is not ideal, because some people might think that those actually are numbers in percent. If you make it 1,1,2 for example and explain that that equals 0.25,0.25,0.5 it gets much clearer

vitorpamplona:
> why do we need lud16,lud06? can't we just have pubkey so we don't have to handle N different UI variations for this?

I am in favor in killing lud16 and lud06 options.

pablof7z:
yeah, I like option 1 too and I'm in favor of killing luds, which will make it easier to display the splits

vitorpamplona:
@v0l @staab let us know your position in terms of removing lud06 and lud16 from the zap tag. I can quickly modify this and start changing Amethyst's code to do hexes instead.

staab:
I'm in favor of option 1 as well

v0l:
> > What is the relay tag for
> 
> A tip for where to get the author's metadata from

Option 2 for me

Cameri (approved):
Option 1

rolznz:
This would be one lnurl pay request per zap tag, right? Does this require mobile clients to have NWC to work practically? (otherwise a user would need to manually pay multiple invoices?)

jb55:
On Thu, May 25, 2023 at 03:28:35AM -0700, Egge wrote:
>> This would be one lnurl pay request per zap tag, right? Does this require mobile clients to have NWC to work practically? (otherwise a user would need to manually pay multiple invoices?)
>
>Yes, that is correct.

yes, lnurl prisms are insecure. it's better if clients know who they are zapping.

vitorpamplona:
We should advocate with wallets to allow a uri scheme were we can send many invoices to be paid at the same time for those who don't want to use NWC.

arthurfranca:
Regarding zap splits without NWC, I propose [a simple solution](https://github.com/nostr-protocol/nips/blob/b9424fd454b124b5a61941d611f71336623cf582/96.md#monetization:~:text=If%20NIP%2D47%20isn%27t%20configured%2C). Taking advantage of the popularity of this thread, I ask client authors to review the related NIP-96 PR part (previous link) that talks about how file storage servers should get a share of the money on zap splits.

vitorpamplona:
@Semisol @fiatjaf can we merge this?

pablof7z (approved):

jb55:
I see that it says "SHOULD", so I guess it's fine.

PR #556, opened 2023-05-25
NIP-57 - Zap Proofs / Partially complete zap receipts

Egge7:
Todays zap flow puts a lot of trust in the Zap Provider publishing the zap receipt. There is pretty much no way to proof a zap has been sent, when the provider refuses or fails to publish the receipt (even a preimage wont help you here). In order to enable a broader zap-based ecosystem this needs to change.

This PR propses a fairly simply addition to the Zap flow in which the provider sends a incomplete Zap Receipt along with the payment requests that can be completed and published by the sender only after the preimage has been revealed.

Instead of modifying the Zap Receipt, the provider could also encrypt it using the preimage and send it along, but it feels like this adds too much complexity on both sides.

This is marked as draft because it propbably is not mature enough yet and needs feedback.

### DIFF
--- a/57.md
+++ b/57.md
@@ -178,6 +178,30 @@ When an event includes a `zap` tag, clients SHOULD calculate the lnurl pay reque
 }
 ```
 
+### Appendix H: Partially Complete Zap Receipt
+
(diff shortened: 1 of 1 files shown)


Egge7:
> I feel like this is unnecessary.
> 
> This also requires the zap service have the preimage available ahead of time.

Right now a zap provider can always stop a payment from becoming a zap, even after successful payment. I feel like this is stopping us from building more complex applications on Zaps.

You mean in the case that Zap Provider and wallet provider are two different entities?

weex:
@Egge7 Do you have a response or want to close this?

PR #492, opened 2023-05-03
New specification for NIP-51 to accept curated articles (edited from NIP-24 proposition for topics)

mostafa-khaldi:
A NIP that represents topics that group a selection of articles of a `kind:30023` of the NIP-23.

PR #559, opened 2023-05-26
Categorized Custom Emojis List

SnowCait:
## Purpose

To share custom emojis list between clients and/or devices.

## Summary

Use [NIP-51](https://github.com/nostr-protocol/nips/blob/a56d1c2877e5c40640350fe0b21748642258baaf/51.md) ([NIP-33](https://github.com/nostr-protocol/nips/blob/a56d1c2877e5c40640350fe0b21748642258baaf/33.md)).  
In addition to `emoji` tags, use `a` tags to import other custom emojis list. It is useful not to copy list.

## Specification

- kind `30002` (increment) or `30030` (from NIP-30)
- tags
  - `d`: Category name (e.g. `["d", "ostrich"]`, `["d", "cat"]`)
  - `emoji`: Defined in [NIP-30](https://github.com/nostr-protocol/nips/blob/a56d1c2877e5c40640350fe0b21748642258baaf/30.md) (e.g. `["emoji", "<shortcode>", "<image-url>"]`)
  - `a`: Reference to other custom emojis list (e.g. `["a", "30030:<pubkey>:ostrich", "<relay-url>"]`)

## Note

Please hide whitespace for table diff.


fiatjaf:
Looks good to me.

verbiricha (approved):
❤️‍🔥

erskingardner:
Sweet! PR looks good to me but... it's made me wonder about the long-term scalability of lists in their current format. If we continue on this path we'd likely end up creating a different list `kind` for every potentially definable type of list. e.g. movies, music, stars, etc. etc. etc.

I wonder if it would be a good idea to explore how we might generalize lists a bit. 

For instance, why do we have specific defined list kinds at all? Would it be better to assume that lists can always contain a mix of any type of event and just leave it to users/clients to use the `d` tag to give meaning to the list? I've seen plenty of lists misused (10000 for pins, 30000 for mute, etc) but it never seems to cause issues. 

Additionally, `bookmark` and `people` both seem to be transplants from the social media use-case. It's not obvious to me that we should hang on to that model when thinking about how to make lists most useful in the long-term.

I don't (yet) have a strong opinion here so I would love feedback, thoughts.

staab:
@erskingardner consider #532 for a more general "labeling" approach. Not lists per se, but it might cover some of the things you're thinking about here.

SnowCait:
@erskingardner 
That is a good point.
To be honest, I thought it is enough to be a single list using kind `30001` + `["d", "custom-emoji"]` at first.
It will work fine with a REQ `{"kinds": [30001], "#d": ["custom-emoji"]}`.

However, considering custom emojis increase unlimited, multi (categorized) lists will be required.
The REQ will become `{"kinds": [30001]}` because we don't know which `d` tags are  custom emojis.
This query result is large and includes many unrelated events. Client cannot allow this.
Therefore I've concluded independent kind is suitable to retrieve neither too much nor too little events of custom emojis lists.

erskingardner:
@staab Yeah - thanks for the nudge towards #532. I hate to take another few steps back here but something you said in there made me think of lists in a completely different way. I want to noodle on this a bit more but I don't think it should necessarily block this change. 

@fiatjaf I'd love to hear a bit more about your thoughts on the infinitely growing space of list types based on [NIP-51](https://github.com/nostr-protocol/nips/blob/master/51.md), where you mention that "A distinct event kind should be used for each list type created." Maybe I'm overthinking this...

fiatjaf:
Whenever you need lists that serve a specific function that must be standardized I think they should be strictly defined and use a dedicated kind. Mute lists, for example, that you want to share across apps. That's what kinds are for.

But if you just want to _show_ lists that will have no programmatic behavior based upon them, like the list of your favorite movies or a collection of links to your favorite geojson libraries, then I think we need one special kind that is generic.

erskingardner:
Thanks. I feel like that makes a case to put emoji lists in the 10k range of kinds then, instead of the 30k range, since they're definitely something I could see clients using programmatically. Again – I actually don't have too strong an opinion here so I think if we're all happy with new kinds for each new different type of list, then go with this as is.

fiatjaf:
I don't like emojis, so I think we need the opinion of more people who would be interested in implementing support for such a list.

ttiiggss:
I'm neither here nor there for emoji's. They can be very annoying... but at the time create great community if the emotes get shared and become 'main-scene'. What if some Nostr relays supported them, but the majority didn't? Perhaps emoji-centric Relays could be a thing for more inclusive communities?

erskingardner:
I've started implementing emojis in Listr already but in a meta way – not (at least to start) in a way to use these emojis in notes). Like I said in my earlier message, there is a lot of inconsistent usage of lists across clients at the moment which is what got my mind going in the direction of how we can make sure we're building something simple but consistent going forward.

### To recap my understanding and make sure others are on the same page: 
1. Lists in the 10-20k kind range are for lists where there should be 1 (and only 1) copy of a given list for a given user. Mute and Pin are the only two in this category so far. These should have a **universal** meaning that is shared across all clients that interact with them. That meaning isn't up for interpretation.
2. Lists in the 30-40k kind range are for lists with a name given by the user. There is 1 (and only 1) copy of a given list with a given name for a given user. New events from the same user, of the same kind, with the same `d` tag (name) replace all previous lists with the same name. The meaning of these lists is **not** universal across all clients. The meaning is ascribed by the user via the name of the list.

So, that means that if we want clients to support a set of custom emoji's in a consistent way for each user, we should be using a kind in the 10000-19999 range. That way there is 1 and only 1 master list for custom emojis for a given user. Otherwise, we have multiple lists of custom emojis for a user in the 30000s range and clients don't know if they should be looking at one, a subset, or all of those lists when detecting custom emojis that a user might want to use.

To get consistent usage and support for custom emojis in clients, we likely need both kinds. A PRE kind in the 30000s for emoji lists (like proposed in this PR) but we'd ALSO need a simple replaceable event for custom emojis in the 10000s. This list could reference emoji tags directly or could reference other PRE lists of emojis, but would stand as the canonical list of custom emojis that a user expects to find and use in clients.

@SnowCait @verbiricha @pablof7z @vivganes @staab @v0l As people doing stuff with lists (and potentially emojis) I would love to hear your thoughts.

verbiricha:
>So, that means that if we want clients to support a set of custom emoji's in a consistent way for each user, we should be using a kind in the 10000-19999 range. That way there is 1 and only 1 master list for custom emojis for a given user. Otherwise, we have multiple lists of custom emojis for a user in the 30000s range and clients don't know if they should be looking at one, a subset, or all of those lists when detecting custom emojis that a user might want to use.

Makes sense, we could use kind 10030 as User Emoji List event, containing a list of `emoji` or `a` tags pointing to `30030` Custom Emoji Lists. These are the emojis the user wants to use and clients can fetch this event and related 30030 lists to show appropiate emoji autocompletion for a user.

>To get consistent usage and support for custom emojis in clients, we likely need both kinds. A PRE kind in the 30000s for emoji lists (like proposed in this PR) but we'd ALSO need a simple replaceable event for custom emojis in the 10000s. This list could reference emoji tags directly or could reference other PRE lists of emojis, but would stand as the canonical list of custom emojis that a user expects to find and use in clients.

I'd restrict 30030 lists to only contain `emoji` tags and not be able to point to other emoji lists. We don't need that if we have kind 10030 as User Emoji List. Thoughts @SnowCait ?

SnowCait:
@erskingardner @verbiricha 
I also think kind `10030` works fine (and prefer it personally) but I'm afraid it becomes complex or fetter (i.e, lost extendability).
I can't decide a consistent way is better at the moment. I'll more consider...

e.g. Client authors or users may wander how to use. (overthinking?)

e.g. Doesn't single list of kind  `10030` let us difficult to extend in the future?

e.g. We cannot create nested lists if not be able to point to other emoji lists in kind `30030`. (but it is simple. I like simple)

<details>
<summary>nested list example</summary>

```
{
  pubkey: <pubkey-A>,
  kind: 30030,
  tags: [
    ["d", "cat"],
    ["emoji", "abyssinian", "https://example-a.com/abyssinian"]
  ]
}
```
```
{
  pubkey: <pubkey-B>
  kind: 30030,
  tags: [
    ["d", "cat"],
    ["a", "30030:<pubkey-A>:cat"]
    ["emoji", "siamese", "https://example-b.com/siamese"]
  ]
}
```

</details>

### Another Idea

@verbiricha's idea is realistically and good. A bit different idea is here.
Restrict `10030` lists to only contain `a` tags and `30030` lists contain only `emoji` tags.
I don't know this is good but each kind's roles will be clear. Instead, require both kinds.

erskingardner:
> I'd restrict 30030 lists to only contain emoji tags and not be able to point to other emoji lists. We don't need that if we have kind 10030 as User Emoji List. 

@verbiricha is the limit on just using emoji tags here just to limit the amount of recursive queries we'd need to make? Or were you thinking about something else too?

verbiricha:
> Restrict 10030 lists to only contain a tags and 30030 lists to only contain emoji tags.
I don't know this is good but each kind's roles will be clear. Instead, require both kinds.

My reasoning behind allowing `a` and `emoji` in `10030` is because some clients might want to allow to add emoji to your list (10030) without caring about reusable lists (30030). For instance a client might allow you to add custom emoji individually but not follow other people's emoji. It is also possible to add emojis from other people/lists to your list individually without having to reference the full list.

>@verbiricha is the limit on just using emoji tags here just to limit the amount of recursive queries we'd need to make? Or were you thinking about something else too?

Yes, we don't need nesting.

---

Proof of concept implementation in https://emojis-iota.vercel.app/ Haven't implemented adding individual emojis to your list but shouldn't be too hard to add.

erskingardner:
I'm all for the restriction on `30030` lists having only `emoji` tags and `10030` having both `emoji` and `a` tags. It's a little quirky but I think the reasons are valid. 

Anyone have a strong aversion to this? Or will we move forward with that proposal?

fiatjaf:
So let me make sure I understand this: we're defining 2 kinds, 10030 is a singleton for each user, 30030 can be many for each user, it's basically an "archive".

Is the meaning of these that 10030 contains your "current" emoji set, while 30030 contains lists you might want to switch to or share with others?

erskingardner:
@fiatjaf that is basically it.

### The options:

1. Use just a single `10030` list. This makes it easiest for clients because there is always only one that matters. Users would be able to put `emoji` tags in this list and use them. Drawbacks, it doesn't make sharing lists (sets) of emojis between users easy.
2. Use just kind `30030` lists. Users can give lists of emoji's names, lists can contain references to `emoji` tags and also `a` tags pointing to other emoji lists. Drawbacks, this makes it hard for clients to pull and rationalize a user's custom emojis. There may be references to many duplicates and potential circular references that aren't immediately apparent to the user.
3. Use both kind `10030` and `30030`. `30030` is for organizing, sharing, and curating lists of emojis. These are limited to only containing `emoji` tags to limit circular and recursive craziness. The `10030` list, on the other hand, is the only one that clients have to worry about. Users would decide explicitly what emoji set they want to use (vs curating lists for others) and would put those emojis in the `10030` list. These lists could contain `emoji` and `a` tags.

fiatjaf:
I like that many clients can just request a kind 10030 and use that, and never bother about kind 30030. While other clients or specialized emoji manager applications can deal with 30030.

I don't understand why we need to reference lists inside other lists. Can't 30030 lists just have emojis and not references to other lists? I think this is better, simpler.

Is the use case for having lists inside lists that someone may want to curate lists of emojis? In that case I think we need another kind. We could define a kind for "anything", or we could define a kind for "curated lists" or we could have a kind for "curated lists of emoji lists".

erskingardner:
> I don't understand why we need to reference lists inside other lists. Can't 30030 lists just have emojis and not references to other lists? I think this is better, simpler.

This is already happening with parameterized replaceable lists. @pablof7z is doing it with highlighter, I've seen a few other instances around as well. It takes an extra step to support it well in clients for sure, but it's not terrible. The use case is really helpful. For highlighter, it allows for nested lists of highlights or notes, for emoji lists it allows for creating curated lists of lists. In general, I've been thinking of it like folders for your lists. 

It's definitely possible that it would be better/simpler to have a specific kind for "lists of lists" to make sure we are clear about where devs will need to handle recursivity and where they won't. @pablof7z I think you've implemented this more than anyone else, what do you think about that?

fiatjaf:
Well, it may make sense in highlighter.com, but should we just shove that into lists of emojis just because it looks cool for other use cases?

pablof7z:
oh yeah, I was talking in general terms; I absolutely don't think it makes sense for this one particular use case.

SnowCait:
I have implemented https://github.com/nostr-protocol/nips/pull/559#issuecomment-1566104988 in practice. (Sorry for late)
It is simple to use (one list and no recursive 'a' tag is better) and we can share of roles (use pack / create pack).
I'd like to agree with this proposal.
In that case, can I add commit or could @verbiricha add?

verbiricha:
@SnowCait go ahead and update the proposal, sounds good to me. Updated the text in https://github.com/SnowCait/nips/pull/1 to reflect this.

SnowCait:
@verbiricha Thank you!

erskingardner:
I'm still not convinced this is a good way forward. Given lists are so flexible (and it's so easy to handle the types of content in lists) this feels like we'll just be setting ourselves up for a huge number of list types for any imaginable type of list item.

One counterpoint to this is that it's obviously helpful for filtering/indexing lists from relays. Is that worth the potential added complexity implicit here?

erskingardner:
Bringing this back up to close it out, hopefully. 

#492 proposes adding a new PRE event for a curated list of long-form articles. That seems to support the direction that this discussion went which was to just add a single PRE kind for collection of emoji tags `kind: 30030`. 

I've come around to the point that we shouldn't have a `kind:10030` as well.

What needs to be done for this to get merged?

51.md
```diff
@@ -123,12 +123,14 @@ Then the user would create a 'Categorized Bookmarks' list event like below:
...
-| ------ | ----------------------- |
-| 10000  | Mute                    |
-| 10001  | Pin                     |
-| 30000  | Categorized People      |
-| 30001  | Categorized Bookmarks   |
+| kind   | list type                 |
+| ------ | ------------------------- |
+| 10000  | Mute                      |
+| 10001  | Pin                       |
+| 10030  | Custom Emoji List         |
```
erskingardner, line 130: @SnowCait i think if we remove the references to the 10030 list this is probably mergeable? @pablof7z what do you think?

SnowCait: @erskingardner Do you mean to use #492 instead of 10030? (I've not read #492.)

erskingardner: No – I think that #492 should also just be a new `kind` listed on NIP-51 just like the categorized emoji list. There isn't a need for an entirely new NIP there. I've talked to a good few folks recently and the use of the 10000's lists is actually very low among clients. e.g. Most are using 30000 lists for Mute lists. I think the best move going forward is to try and normalize/maximize the usage of the kind numbers in the 30k-40k range and use the kinds + standard tags to build lists that have more context. That would mean...for example, 

* 30000 - categorized people
* 30001 - categorized bookmarks
* 30030 - categorized emoji 
* ect...

fiatjaf: What `d` tag are they using for mute lists?

fiatjaf: Why even have the kind numbers, then? If the lists will only be indexed by the `d` tag and there is no fundamental difference between them then they could just use the same kind number for "generic list".

erskingardner: > What d tag are they using for mute lists?

Usually just "mute" or "Mute". 🤷‍♂️ 

> Why even have the kind numbers, then? If the lists will only be indexed by the d tag and there is no fundamental difference between them then they could just use the same kind number for "generic list".

So that you can search/filter for them. Without something that identifies them you'd need to fetch all lists from a user, loop through them, and do the filtering yourself based on the items in the lists.

SnowCait: >the use of the 10000's lists is actually very low among clients.

That is client problem, not NIP one. Why does NIP follow regacy clients which doesn't follow NIPs?
[My client](https://nostter.vercel.app/) uses 10000, 10001 and 10030 in fact.

fiatjaf: > Usually just "mute" or "Mute". 🤷‍♂️

Well, if clients have not bothered to agree among themselves when doing a simple thing as a mute list or to follow the NIP, what makes you think they will from now on?

Honestly, if people don't want to be compatible with other apps or they don't care, there is nothing we can do.

I was expecting you to tell me most clients had agreed on some standard that deviated from the NIP, in which case I would say we should change the NIP to reflect that -- but also ask where are they having these discussions, because that is a very important part of Nostr. But it looks like they haven't agreed on anything and apparently do not care.

verbiricha: > I've talked to a good few folks recently and the use of the 10000's lists is actually very low among clients. e.g. Most are using 30000 lists for Mute lists.

> Most are using 30000 lists for Mute lists.

How many is "good few folks"? This is wrong in any case, they should be using 10000 for mute lists as specified in NIP-51. I'm guilty of this for implementing NIP-51 proposal in snort before it was merged (sorry) and it seems other clients have copied the code without checking the NIP. As Snowcait says "that is client problem, not NIP one." NIP is very clear on the kind used for mute list.

Going back to the topic of the conversation:

> I've come around to the point that we shouldn't have a kind:10030 as well.

> I think the best move going forward is to try and normalize/maximize the usage of the kind numbers in the 30k-40k range and use the kinds + standard tags to build lists that have more context

The point of the 10030 event is to have a list of current emoji packs (30030 events by you or others) that clients can fetch for the user and show autocompletion. Is like your Telegram sticker collection, you can browse sticker packs created by others and add them to your collection. The expectation is that you'll have the same emoji packs available across all clients and since it will only be one per pubkey I chose a replaceable event.

If using two different kinds is the problem we could use a 30001 list with well-known "d" tag like "emoji" for the same purpose. For instance satellite.earth does this to store a list of communities you are "subscribed" to using kind 30001 and "d" tag "communities" (this is not in the communities NIP yet, not sure why).

SnowCait:
Relates to #880

fiatjaf:
I believe this is addressed by recent changes in https://github.com/nostr-protocol/nips/pull/880.

SnowCait:
Perfect💯 Thank you!

Issue #545, opened 2023-05-20
NIP-XX: NIP proposal process


weex:
As a newcomer I find it confusing to follow the various NIPs and ideas that have been proposed. The README.md doesn't give any guidance on process which means that submitters are somewhat less likely to make proposals and it's also harder for maintainers to follow drafts in a consistent way. The graph of discussion in PRs and issues also becomes interwoven and hard to know when every related bit of discussion has been seen.

**Proposed solution**
Define a process for proposing a NIP such as:
1. Create an issue describing the problem and proposed solutions. The title should be something like NIP-XX Title where title is descriptive and less than 80 chars. This is a tracking issue which should exist for all NIP drafts and ideas.
2. If the author has a draft prepared, they create a pull request with that draft and mention the issue. If a draft is asking to reserve a kind, it should make such proposal near the top so maintainers can see it easily and offer the requested or other kind.
3. Maintainers assign a NIP number and edit the tracking issue and PR to include that number. This makes it easier to search in a consistent way (i.e. by NIP-XX).
4. Discussion ensues in tracking issue, links added to tracking issue, and in the PR.
5. When the author feels that issues raised by the community have been addressed, they request in the tracking issue to finalize the NIP and tag the maintainers.
6. After some period with no additional blocking feedback, the NIP is finalized.

Once this process is adopted there's some work to backfill tracking issues. I see this as a step on the way to address #162 

Thoughts? If you'd like to see this as a PR, please give a thumbs up.

xeruf:
also wrote about this in https://github.com/nostr-protocol/nips/issues/1216

fiatjaf:
This is a good idea, but we don't have the personnel required for keeping track of these things.

fiatjaf:
I think we should go the opposite route.

PR #560, opened 2023-05-26
Add submission process

weex:
Update the README.md to include a new NIP submission process.

Addresses #545


fiatjaf:
Thank you your attempts at bringing order to this repository, but I think for now we should keeping living in chaos.

PR #567, opened 2023-05-29
Update 07.md Implementation

haorendashu:
When user vist website in nostrmo. The website can access and use nostr object.


AsaiToshiya:
Please tell me how to use it as an implementation of NIP-07. At least on Android it looks like it's just a client to me.

haorendashu:
Nostrmo provider NIP-07 with webview.

You can see these screenshot when visting kind3.xyz :

AsaiToshiya:
I see, it is provided to the WebView in the app.

Issue #419, opened 2023-04-07
Accountability for publishers and relays


nostronaut:
### Aim: permit accountability for publishers and relays. 

follows the discussion [in this issue](https://github.com/nostr-protocol/nostr/issues/148)

### Publisher's weight using key-longevity

This proposal requires a chain of events. This can be an optional protocol built on top of Nostr.

Instead of chaining events directly, using the event_hash like in Secure Scuttlebutt, use merkle mountain range (MMR) as an append only vector commitment for events (the same data structure used by opentimestamps in NIP-03) to enable fast verification (only a few events needed for proof of membership in the MMR).

In a Nostr event, add a field in `tags` that includes the `previous_event_hash` and the `previous_root_hash` of the MMR.

Special relays and clients implement the extra protocol on top of Nostr. To enable the goodness, a relay must at a minimum store the last prev_hash and MMR prev_root_hash for each pubkey and its sequence number, and verify the new event follows from the last one, else reject event, etc, being super lightweight, and able to detect missing events from publisher. In contrast, an archival client might store and verify the entire event history, and may serve data to light clients to construct their light proofs. A client implementing this protocol behind a relay dropping events can tell events are being lost.

The idea is to calculate a weight for the author's pubkey in a similar way as one calculates the amount of POW for a blockchain, here the `weight of the author` being the sequence number of the event, that is the number of events that key signed sequentially. Up on the stack, one might estimate the `weight of the event` based on the `weight of the author` and of the people that references the event. For example, when long-lived, active identities interact, the event thread becomes a heavy-weight event-thread. This might be useful for page ranking.

A basic form of moderation is still needed, for example, relays shouldn't replicate events from a spammer with a long and provable career of spamming etc.

When correctly combining the current protocol with NIP-03, one might be able to create a notarised "side-chain". As analogy to Bitcoin, one could think of this chain as containing one UTXO (the last event) that gets consumed when a new event builds on top of the last event.

nostronaut:
since it looks like there is some interest in this, we implemented a demo client here https://github.com/nostronaut/mmr-nostr

nostronaut:
first and foremost, the proposed authentication system should not be used by everyone, because it is too commiting, and may be exploited as legal evidence against normal people. 

on the other hand, it could be used by companies, journalists, polititians, or in general reliable sources of information ("oracles") trying to attest to the state of the world.

> note: we already have previous event hash in fields. it's called "reply to" and an "e" tag. 

[...]

PR #569, opened 2023-05-29
draft: references to previous events

milahu:
[rendered](https://github.com/milahu/nips/blob/references-to-previous-events/references-to-previous-events.md)

fix https://github.com/nostr-protocol/nostr/issues/102


references-to-previous-events.md
```diff
@@ -0,0 +1,119 @@
...
+- random distances, for example: 1, 15, 32, 58, 93, 139, 185, ...
+
+```json
+{
+  "id": ...,
+  "kind": ...,
+  ...,
+  ...,
+  "chain": [
+    <previous event distance>, <previous event ID>,
```
milahu, line 35: maybe allow adding relay URLs to each previous event

```json
{
  "chain": [
    [<previous event distance>, <previous event ID>, <previous event relay URL>],
    [<previous event distance>, <previous event ID>, <previous event relay URL>],
    ...
  ]
}
```

staab:
This is not a bad idea, what might be better (and more compact) would be [vector clocks as implemented by polycentric](https://gitlab.futo.org/polycentric/polycentric/-/blob/master/doc/protocol.md). Basically, each device a user signs into generates a unique id for itself, and assigns an incrementing integer to each event. So you would have `instance, ordinal` properties added to events. You would then be able to find missing events (by device) by numeric range. I also proposed this as part of [blazepoint](https://github.com/blazepoint/docs/blob/master/protocol.md).

As a nice bonus, nostr would then have ordinals! I think it's worth doing it this way just to make that joke.

milahu:
more compact would be nice indeed.
one argument in favor of the (less compact) event IDs: all relays support lookup by event ID

> You would then be able to find missing events (by device) by numeric range.

this (search events by vector time) could be cheaper than my proposed forward search (search events by references to previous events)

nostronaut:
we are proposing something that is redundant to this NIP proposal, but with a larger scope and better understood properties, using MMR as the vector commitment to refer to previous events, with $log_2$ scalling properties as a fn of nr of events. please find it [here](https://github.com/nostr-protocol/nips/issues/419)

milahu:
> we are proposing something that is redundant to this NIP proposal, but with a larger scope and better understood properties

granted, my proposal is rather chaotic. this can be seen as a strength, because users are free to choose the number of "backlinks" to defeat an arbitrary attack (arbitrary number of missing events)

what would be the worst-case-scenario in your case? "worst case" as in: some events are missing, but a current event can be linked back to some previous event in the chain? can this limitation be exploited to break the chain of events?

at first sight, your proposal looks less censorship-resistant, when you store only these values per event

```rs
    Mmr {
        prev_event_id: Hash,
        prev_mmr_root: Hash,
        prev_event_pos: i64,
    }
```

i would expect more redundancy (more references to previous events), which also allows "adopting" the chain-of-events for previous events

("adoption" in lack of a better word)

nostronaut:
> users are free to choose the number of "backlinks" to defeat an arbitrary attack (arbitrary number of missing events)
> i would expect more redundancy (more references to previous events), which also allows "adopting" the chain-of-events for previous events

the MMR root works as an cryptographic accumulator of events (its hash is derived from all events), thus through the root you have short "backlinks" (i.e., merkle paths) to **all** events all at once (again, it is a set accumulator, or if u like a commitment to the set of all events)

> at first sight, your proposal looks less censorship-resistant, when you store only [these values](https://github.com/nostronaut/mmr-nostr) per event

that really is the beauty of it :D store so little and be able to verify everything. a MMR validator node may be storing a lot more hashes than that, but they are able to compute the missing ones from the ones published

if you'd like to really understand how this works, including prunning, I suppose these two links may be of help:

https://docs.grin.mw/wiki/chain-state/merkle-mountain-range/

https://eprint.iacr.org/2019/226.pdf

> for me, the worst case would be a relay that blocks all events after a censored event. in that case, a recipient has no hard "proof of censorship", only a weak suspicion of "either the sender was censored, or he stopped sending events" (assuming the recipient still has trust in the relay, and has no better relay)

that is a protocol liveness problem impossible to resolve with a single untrusted relay, that is why one needs many relays

nostronaut:
of course, i'm trying to convince you that prunable MMR is the optimal solution, and you should come work on it :P

MMR stores data only on leave nodes

but if you'd like something like your proposal, storing events on all nodes, but more formal, there is [bamboo](https://github.com/AljoschaMeyer/bamboo)

ssb will use it, and p2panda currently uses it

milahu:
does MMR tolerate missing events?

example: initially, 2 events are missing between `E0` and `E3`. we ask a different relay for events between `E0` and `E3`, and we get `E1` but not `E2`. can we verify this broken chain `[E0, E1, E3]`?

```
      H3
    /    \
  H1      H2
 /  \    /  \
E0  E1  xx  E3
```

when `E2` is missing, we cannot verify the position of `E3`, also not by receiving more events (no?)

> [bamboo](https://github.com/AljoschaMeyer/bamboo)

it seems that bamboo has only one backlink per entry, while i propose multiple backlinks per entry (redundancy for better fault-tolerance)

nostronaut:
https://docs.grin.mw/wiki/chain-state/merkle-mountain-range/#pruning

nostronaut:
on ur example, xx is a member of H2 and H3

milahu:
> Pruning
> In Grin, a lot of the data that gets hashed and stored in MMRs can eventually be removed.

can the MMR be pruned when some leaves are missing? (my guess: no)

> Merkle proof
> 
> To construct a Merkle proof to show inclusion of
> node at pos 8 (green) in the peak at pos 15
> (yellow) we need the hashes of "sibling" nodes
> (blue) up the branch connecting 8 to 15.
> 
> - 9 proves 8 is "in" 10
> - 13 proves they [10 and13] are "in" 14
> - 7 proves they [7 and 14] are "in" 15 (peak)

does the proof work when some leaves are missing? (my guess: no)

> when `E2` is missing, we cannot verify the position of `E3`, also not by receiving more events (no?)

`E3.prev_mmr_root` = `H2` (or `H3`?)
`E3.prev_event_id` = `E2`
`E3.prev_event_pos` = `3`

can you please show how this would play out?
where do we store the hashes `[H1, H2, H3]`?
what are the values of `E.prev_mmr_root`?

nostronaut:
MMR is not the simplest data struct, you will have to study it carefully and let it sink in, if you want to fully appreciate it, and then you can stop guessing

```
E3.prev_mmr_root = H1
E3.prev_event_id = E2
E3.prev_event_pos = 3
```

```
E4.prev_mmr_root = H3
E4.prev_event_id = E3
E4.prev_event_pos = 4
```

then 
```
E0.pos = 0
E1.pos = 1
H1.pos = 2
E2.pos = 3
E3.pos = 4
H2.pos = 5
H3.pos = 6
```

`H2`, for example, never makes it into an event explicitly, but the verifier can compute it from `E2` and `E3`, and it is included in `H3`, which is commited

i do think its more meaningful to go through the references above, because quickly drafted examples may include mistakes

nostronaut:
link to the MMR issue https://github.com/nostr-protocol/nips/issues/419 being discussed above

milahu:
```
                      H7
                /            \
              H3              H6
            /    \          /    \
          H1      H2      H4      H5      H8
         /  \    /  \    /  \    /  \    /  \
event:  E0  E1  E2  E3  E4  E5  E6  E7  E8  E9

  pos:   0 2 1 6 3 5 4 e 7 9 8 d a c b   f h g

eprev:   -  E0  E1  E2  E3  E4  E5  E6  E7  E8

mmrpr:   -  E0  H1  H1  H3  H3  H3  H3  H7  H7
```

when one event is missing (for example `E2`), we get `E2.id = E3.eprev`.
we use `E2.id` to get `E2` from other relays.

---

when two consecutive events are missing (for example `[E1, E2]`), then the chain is broken.
we still can get `E2.id = E3.eprev`, but we dont have `E1.id = E2.eprev`.
so we ask some relays: "i need events between `E0` and `E3`"

---

what happens when two consecutive events are missing forever?
for example, `[E1, E2]` are missing forever.
we still can get `E2.id = E3.eprev`, but we dont have `E1.id = E2.eprev`.

```
                      H7
                /            \
              H3              H6
            /    \          /    \
          H1      H2      H4      H5      H8
         /  \    /  \    /  \    /  \    /  \
event:  E0  xx  xx  E3  E4  E5  E6  E7  E8  E9

  pos:   0 2 1 6 3 5 4 e 7 9 8 d a c b   f h g

eprev:   -  E0  E1  E2  E3  E4  E5  E6  E7  E8

mmrpr:   -  E0  H1  H1  H3  H3  H3  H3  H7  H7
```

without `E1.id`, we cannot compute `H1` = we cannot validate `E0`.

we have `H1 = E3.mmrpr`, so we can validate `[H2, E2, E3]`

---

what happens when three consecutive events are missing forever?
for example, `[E1, E2, E3]` are missing forever.
we still can get `E3.id = E4.eprev`, but we dont have `[E1.id, E2.id]`.

```
                      H7
                /            \
              H3              H6
            /    \          /    \
          H1      H2      H4      H5      H8
         /  \    /  \    /  \    /  \    /  \
event:  E0  xx  xx  xx  E4  E5  E6  E7  E8  E9

  pos:   0 2 1 6 3 5 4 e 7 9 8 d a c b   f h g

eprev:   -  E0  E1  E2  E3  E4  E5  E6  E7  E8

mmrpr:   -  E0  H1  H1  H3  H3  H3  H3  H7  H7
```

without `E1.id`, we cannot compute `H1` = we cannot validate `E0`.
without `E2.id`, we cannot compute `H2`, but we can validate `E3` with `E3.id = E4.eprev`.

---

what happens when four consecutive events are missing forever?
for example, `[E1, E2, E3, E4]` are missing forever.
we still can get `E4.id = E5.eprev`, but we dont have `[E1.id, E2.id, E3.id]`.

```
                      H7
                /            \
              H3              H6
            /    \          /    \
          H1      H2      H4      H5      H8
         /  \    /  \    /  \    /  \    /  \
event:  E0  xx  xx  xx  xx  E5  E6  E7  E8  E9

  pos:   0 2 1 6 3 5 4 e 7 9 8 d a c b   f h g

eprev:   -  E0  E1  E2  E3  E4  E5  E6  E7  E8

mmrpr:   -  E0  H1  H1  H3  H3  H3  H3  H7  H7
```

without `E1.id`, we cannot compute `H1` = we cannot validate `E0`.
without `E2.id`, we cannot compute `H2`, but we can validate `E3` with `E3.id = E4.eprev`.

---

and so forth ...

nice. but i also like my idea, because it allows

- adoption for older events
- long jumps = jumps to previous events with large distance

nostronaut:
u r taking a data structure that can hold millions of event commitments, you delete the second event, and claim you cant validate the first one! not very fair

what you really describe is not what happens if u delete 2, 3 or 4 events consecutively, but mostly that u cant validate the first event if you delete the second. And if the first is the genesis event, you cannot validate genesis

now shift your deletion start one to the right, from the 3rd leaf, and everything works again :P, you have a merkle path 

when you delete the second leaf, that certainly makes it impossible to create a merkle path to validate the genesis event through a binary merkle tree. 

i think you have to be specific what you want to prove at each time. if you want to be able to prove everything all the time, and still be able to delete the entire past, then good luck

fiatjaf:
If it's too complex no one is going to implement. Also if it uses too much space no one is going to implement. I'm interested in the idea @staab linked to, although I didn't read it.

milahu:
most comments here talk about MMR

MMR (proposed by nostronaut)

- has a fixed origin `E0` = append-only
- has backlinks only to the first-previous event
- next to the backlink, it stores a hash derived from event IDs. multiple hashes build a hash tree
- no redundancy = space-optimal solution

"soft chain" (proposed by milahu)

- has one or more backlinks to the previous events
- there is no fixed origin, the chain can grow to "adopt" historical events
- the chain is mutable, events can be inserted by using the same distance as a sibling event
- user is free to choose the level of redundancy

in case of missing events (broken chain), both solutions require relays to implement a "search between events"

vitorpamplona:
Closing due to inactivity.

Issue #72, opened 2022-11-17
nip-04 key material not uniformly random


mikedilger:
Encrypted messages are encrypted according to nip-04 using a Diffie-Hellman shared secret generated from a private key and a public key.  This shared secret is then used as an AES key.  But such a shared secret is not uniformly random.  AES requires a uniformly random key to guarantee its level of security.  Cryptographers recommend using an HKDF (HMAC-based extract and expand key derivation function) ideally with a salt in order to get a uniformly random key suitable for symmetric encryption by, e.g. AES.  SHA256 can be used as the HKDF.

I would suggest generating a random salt, SHA256 the shared secret and the salt together, and include the salt with the message in the same way that the IV is included with the message.  Alternatively, ditch the salt entirely.  Or it can be a fixed well-known string.

Lest you mistake me for a cryptographer (albeit I did take a crypto class from Phil Rogaway in the early 1990s), I'm getting my information from this page: https://docs.rs/elliptic-curve/latest/elliptic_curve/ecdh/struct.SharedSecret.html

I'll leave it to others to suggest how to move forward.  IMHO event kind 4 should not change but should be superseded.

fiatjaf:
I did ask at least one cryptographer before writing the spec for NIP-04 if it was ok to use the shared key directly without hashing it first and he said it was ok.

But even if this is super bad I agree we shouldn't change NIP-04.

By the way NIP-04 was never supposed to be _the thing_ in the first place, it was made as a prototype to show it was possible and expecting someone to supersede it with a better protocol (I personally think we must still find some other more p2p more private way to send direct messages between Nostr contacts that doesn't involve broadcasting to relays).

mikedilger:
It's encrypted either way. The HKDF just makes it less susceptible to cryptanalysis.

p2p encrypted messaging would avoid public knowledge of who was talking to who (would hide the metadata) but requires both parties online at the same time and at least one able to listen. Many clients can't listen due to corporate networks, home firewalls, CGNAT, etc.  I was thinking maybe having messages both encrypted and ephemeral. Clients who would invite encrypted messaging could publish an event disclosing their IP address, without any intent as to who they might want to talk to. As long as enough of them did, it wouldn't be clear who among them might be talking to whom.

Welteam:
[...]

PR #574, opened 2023-05-31
NIP-44: Encrypted Direct Message (Versioned), replaces NIP-4

paulmillr:
Simple solution that fixes broken cryptography and introduces versioning. [Preview](https://github.com/paulmillr/nips/blob/master-1/44.md), [PR to nostr-tools](https://github.com/nbd-wtf/nostr-tools/pull/221)

### tl;dr

- confidentiality is achieved by `xchacha(key: sha256(ecdh(userPriv, partnerPub)), nonce: random(24)).encrypt(message)`
- integrity is achieved by ciphertext getting signed with user's secp256k1 schnorr key
- versioning for future-proofness
- profiles signal which relays should be used to communicate with them
- no one could fetch DM events besides "the authenticated user"

### Benchmark

Round-trip of encrypting-decrypting a basic event with "hello" message, on Apple M2.

```
nip4 x 19,419 ops/sec @ 51μs/op
nip44 x 101,235 ops/sec @ 9μs/op
```

### The goals

1. Security
2. Quick and dirty first version, without advanced features. We can iterate further after the adoption.
3. Simplicity. We need to get maximum adoption by nostr client and relay implementors. NIPs are optional and no one is forcing anyone to implement them. If we overcomplicate things, no one would write code for it.
4. As few building blocks as possible. Implementors should not be able to shoot themselves into the foot: "dumb cryptography". Users should not be able to leak their nostr private key.

NIP4 was bad, because it used AES-CBC with non-hashed, non-IND ECDH key, was not versioned, did not have relay signalling.

Because it's simple, the protocol would not be as good as Signal or others. It's okay, because nostr is public DM system. Ideally you'll exchange Signal (etc) credentials in nostr DMs. If you think the approach is wrong, write down your thoughts in this thread.

### Rationale

- One static shared key is used to encrypt **all** messages between person A and person B
    - We could have rotated keys: encrypting every message with a new one. The key could be generated from some kind of salt, that would be attached to message
        - If user key is leaked, the attacker would be able to decrypt the message anyway. Complication for nothing
- encrypt-then-sign, [XChaCha20](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-xchacha-03) is extended-nonce construction of ChaCha20, [RFC8439](https://datatracker.ietf.org/doc/html/rfc8439)-standardized algorithm
    - It allows us to use same encryption key for all messages. The safe threshold of key wear-out for XChaCha is up to 2^80 messages to be sent under a single key. For AES it's just 2^32!
    - It is simpler, allowing us to use 24-byte nonces and same encryption key for all messages. AES-GCM uses 12-byte nonces which are too short to pick randomly, because of collision risk. Which means a separate key needs to be used per-message, which means we'll need to bring KDF, and complicate the scheme
    - It is faster than AES on devices which don't have hardware acceleration: those include many cheap phones.
    - It is also much faster (14x) for our precise case than AES because there is no KDF step
- No AEAD (Authenticated encryption) aka Poly1305
    - Any tampering, bit flipping will invalidate the ciphertext's signature
    - Poly1305 is not perfect, its integrity limit for single-key messages [is 2^46](https://datatracker.ietf.org/doc/draft-irtf-cfrg-aead-limits/)
    - Polynomial, non-committing MACs are vulnerable to invisible salamanders attack
- SHA256 is used to hash shared secret instead of HKDF
    - sha256 is already standardized for this purpose in libsecp256k1
    - Achieves key IND 
    - HKDF is easy to misuse (salt? info?)
    - HKDF is 10x slower than sha256
    - HKDF introduces a new algorithm into the lightweight nostr protocol, which means more code
    - HKDF is awesome if we were to construct a new version with forward secrecy, but not for now
- Only one encryption algorithm for now, instead of a choice between xchacha and aes. Makes it mandatory for clients to implement without diverging

### Out-of-scope

- Ephemeral anonymous keys. Prototyped in [NIP-444](https://github.com/nostr-protocol/nips/pull/468) and NIP-59 gift wrapping
- Forward secrecy. It's great feature: when a user leaks their key, adversary won't be able to read their previous messages. But it also means users themselves won't be able to read their previous messages. nostr clients need to be able to fetch all historical conversations for UX purposes; nostr clients are immature and it's more-or-less common for users to switch between them. We can get back to forward secrecy prototyping in the future algorithm versions.
- Really private and secure messaging. nostr will always leak metadata. It's better to keep this in mind and only use nostr for exchanging contacts for more secure messenger

### Questions

- Should `content` be hex string, or CSV base64 string?
- Is AEAD really not needed?
- Should we just prepend `02` to B in `getSharedSecret(privkeyA, '02' + pubkeyB)`?
- How could this work with group chats? Signal group chats are simply a bunch 1-to-1 conversations. Maybe we could think of something similar here

Live code: https://github.com/nostr-protocol/nips/pull/574


paulmillr:
@fiatjaf sounds good. How does the general thing look like? And WDYT about https://github.com/nostr-protocol/nips/pull/468?

44.md
```diff
@@ -0,0 +1,80 @@
...
+  ver = 1
+): string {
+  if (ver !== 1) throw new Error('NIP44: unknown encryption version')
+  let key = getConversationKey(privkey, pubkey)
+  let nonce = randomBytes(24)
+  let plaintext = new TextEncoder().encode(text)
+  let ciphertext = xchacha20(key, nonce, plaintext, plaintext)
+  let ctb64 = base64.encode(ciphertext)
+  let nonceb64 = base64.encode(nonce)
+  return JSON.stringify({ciphertext: ctb64, nonce: nonceb64, v: 1})
```
earonesty, line 66: maybe just use "c", "n" and "v", similar to how nostr deals with tags

44.md
```diff
@@ -0,0 +1,80 @@
...
+}
+
+export function decrypt(privkey: string, pubkey: string, data: string): string {
+  let dt = JSON.parse(data)
+  if (dt.v !== 1) throw new Error('NIP44: unknown encryption version')
```
earonesty, line 71: should probably have something in the nip specification that says how user interfaces should deal with versions they don't know about:   let users know they got a message, tell them the version isn't supported and suggest appropriate action (upgrade, etc.)

earonesty:
minor comments, looks great.   ACK

44.md
```diff
@@ -0,0 +1,80 @@
...
+}
+
+export function encrypt(
+  privkey: string,
+  pubkey: string,
+  text: string,
+  ver = 1
+): string {
+  if (ver !== 1) throw new Error('NIP44: unknown encryption version')
+  let key = getConversationKey(privkey, pubkey)
```
earonesty, line 60: i see no reason not to pass in the nonce here, and include it in the hash.   marginal speed difference, and that way if a conversation key gets leaked, it doesn't unlock all messages.

paulmillr: if main key gets leaked, the attacker would be able to calculate sub keys. The security is no different.

starbackr-dev:
> AUTH (NIP-42) requirement for relays to return chat messages (even if relays don't implement that at first, clients should be ready for when they do).

if we make AUTH (NIP-42)  mandatory then we can drop NIP-59 #468 completely because there is no use for it.

earonesty:
making AUTH mandatory is a nonstarter, imo.  also the disclaimers in this nip are misleading.  if metadata is not a concern (and there's no reason we should blanket assume that it is), there is no need to use AUTH.   

also, there are other solutions ... it would be easy to have a system send spurious messages regularly to obscure this, preventing harvesting of this info.

the enryption provided in this leaks the sender and the recipient public keys.   that's the only disclaimer we need.   it is correctly done and scaring people away from using it isn't the right move.

44.md
earonesty: if the post-hash aes key gets leaked, the attacker can't calculate anything if we use the nonce in the hash.

paulmillr: > if the post-hash aes key gets leaked, the attacker can't calculate anything

Why would the sub-key get leaked and not the main key? 

Leaked main key would allow attacker to forge messages and read ALL messages if they access all historical events, since we don't do forward secrecy. How would it be different under the proposed scheme?

starbackr-dev:
We got to pick one option or the other to stop metadata leaks with this. 

Option 1: Make NIP-42 mandatory and explicit relay signaling in the profile. 
or
Option 2: Encrypt metadata as per NIP-59 but free to post on any relay. 

Option 2 wastes compute and bandwidth resources but less restrictive. 

But we have to do either one and close this topic for a while. 

one more. I am new to this NIP process. Shouldn't @earonesty name be on the NIP-44 since he started this discussion and did initial work?

paulmillr:
@starbackr-dev

> he started this discussion

The initial discussion was started by mikedilger in #72, adiabat in #107, AdamISZ in #303. I also don't know how nip authoring process works. Does anyone know?

44.md
paulmillr: Agree.

44.md
paulmillr: I know we're back and forth on this.

Just trying to understand by which standards it's considered a good design to use new key for every mesage, when no ratcheting or FS is in place.

XChaCha clearly says re-using 1 key for 2^80 messages would bring nonce collision probability to 2^-32. Which is miniscule. For AES it's just 2^32 messages, 14 orders of magnitude smaller.

The design is clear. We calculate key once and then decrypt all 5000 conversation messages, using different IVs (IVs may not be public if auth is used).

The suggestion is to re-calculate key for every message. For which purpose? For FS, ratcheting or anon messages that would make sense.

earonesty: i suppose it's fine.   i'll be using ephemeral keys (nip59 style) anyway.   the only reason i can think that the aes key might leak is that the crypto.subtle library is getting called over and over with the same key

earonesty: pls resolve this convo.

mikedilger:
I like this NIP.  I think it makes the right decisions. We can't do forward-secrecy or metadata secrecy in nostr without a lot of other junk (if at all) and we need a better but still simple DM system.

>     * Explicit relay signaling: users must signal via NIP-65 that they want to receive DMs only on some specific relays, and whenever one wants to message these users they should only always send DMs to these relays and only those.

I like this.  I could specify my personal relay that doesn't allow anybody to read back what was written.

> The initial discussion was started by mikedilger in https://github.com/nostr-protocol/nips/issues/72, ...

Don't put me as an author. I didn't think you were going to, but just for clarity.

earonesty:
NIP-59 handle metadata secrecy elegantly and very well.   It does exactly what nostr  should do.  We just need to update NIP-59 to specify that it uses NIP-44 encryption instead of NIP-04 and now we've got the building blocks of a a lot of great protocols!

BlowaterNostr:
@paulmillr Do you have an working example of the padding oracle attack?

paulmillr:
@BlowaterNostr no

iloveyumyumshrimpnoodles:
Question, how does this perform against on-the-fly modifications of the public key, signature and content?


If `Mallory` would change `Alice`'s message, right before it reaches the relay, and changes the public key so it matches hers, changes the ciphertext to something else and recomputes the signature, is there any way to validate that the message actually came from `Alice`?

Why not pass that data into Poly1305 as AAD, so it's protected from being modified?
Or am I wrong?

earonesty:
> 

at this point Mallory has changed the entire message to something she felt like sending

it is just a message from Mallory now

she might as well just posted it from the outset

If she's changing the content and the public key and the event hash and everything

iloveyumyumshrimpnoodles:
Ah yeah, true.
My bad, haven't got much sleep

earonesty:
simple feature.   instead of signing with his public key, a client signs with (hash(shared-secret)*client-private-key) instead.   then the recipient of the dm can verify the message, but nobody knows who sent it, and also the recipient can't prove who sent it without releasing his private key.

paulmillr:
Stealth addresses are hard and require complicated analysis to ensure discrete logarithm problem is not stupidly solvable. i've been around an audit of one such system. Better to not have it in the initial release.

earonesty:
> Stealth addresses are hard and require complicated analysis to ensure discrete logarithm problem is not stupidly solvable. i've been around an audit of one such system. Better to not have it in the initial release.

i would hope that signing with a tweaked private key even if everyone knows how you tweaked it, cant possibly be broken or basically all derivations are broken

fiatjaf:
Sorry, @paulmillr, I forgot to answer https://github.com/nostr-protocol/nips/pull/574#issuecomment-1570493068.

I think "gift wrap" and all the other similar proposals that use random keys or anonymous keys are bad because for the incentives to work on Nostr relays ultimately will have to know something about who is writing to their database, so these solutions won't scale or may require additional complexities (despite already being complex and resource-intensive for clients).

earonesty:
agreed, i suspect payment is best handled at auth-time.   nip-65 is a decent stopgap for scalability issues that can help things right now.   we should also have a "set of relays" on NIP-28 "channel create" messages and recommend they be broadcasted widely, even if the main channel is locates on a smaller set of relays.

earonesty:
i'd like to point out that it is very likely we will only be using this for the "encryption type", and not for actual DM's.   instead we will likely be doing the following:

alice wants to sends bob a message
alice creates a shared secret with bob.   alice uses a hash of this as a new "shared public / private key pair", and sends bob a message appearing to come from "alice public key tweaked by the shared key" and encrypted for "bob's private key".   

bob does the same when he replies.   

these messages don't identify the sender, so there's no way to distinguish them from "spam" ... unless you're alice or bob... 

but they are easily "subscribeable" by either the "fake author" or the "p tag destination"  

the advantage is full metadata hiding and deniability of all messages, while maintaining ability to do subscription

44.md
```diff
@@ -0,0 +1,80 @@
...
+  let nonce = randomBytes(24)
+  let plaintext = new TextEncoder().encode(text)
+  let ciphertext = xchacha20(key, nonce, plaintext, plaintext)
+  let ctb64 = base64.encode(ciphertext)
```
v0l, line 64: This example code didnt work for me, i think `Uint8Array` is weird, need to copy it here:
`const ctb64 = base64.encode(Uint8Array.from(ciphertext))`

I get this error: `radix2.encode input should be Uint8Array`

earonesty: the PR to nostr-tools is written and tested, check that code first.

earonesty:
> I'm concerned about forking dms here. This NIP definitely seems like a necessary upgrade, but will we ever really be free of NIP 04, or will we forever have to support two encryption algorithms? This is a question not just for clients, but the extensions everyone is using too (since most people log in using a nip 07 extension).


we can deprecate nip04.   people will switch pretty quickly.   it was generally decided that 2 kinds was better than 1 kind with implicit versioning (decryption fails if you don' t support the version, and/or you can show a 'version unsupported' message).

i am also concerned with forking.   we will be supporting: NIP04, NIP44, and metadata hidden DM's.   Using the shared secret as the sender's private key is an easy extension.   Maybe we can include that in NIP44?

```
Sender computes the shared secret.  The X coordinate of this is used as the private key to sign the message, rather than signing with the sender's regular public key.   This allows the recipient to verify the sender, but no one else.
```

44.md
```diff
@@ -0,0 +1,80 @@
...
+  if (dt.v !== 1) throw new Error('NIP44: unknown encryption version')
+  let {ciphertext, nonce} = dt
+  ciphertext = base64.decode(ciphertext)
+  nonce = base64.decode(nonce)
+  let key = getConversationKey(privkey, pubkey)
+  let plaintext = xchacha20(key, nonce, ciphertext, ciphertext)
+  let text = new TextDecoder().decode(plaintext)
+  return text
+}
+```
```
earonesty, line 80: ## Sender masking

Instead of signing with the private key, the sender computes a shared secret.  The X coordinate of this is used as the private key to sign the message.   This allows the recipient to verify the sender, but no one else can verify it.   If the recipient's client does not support this, then the client will show it as an unknown sender.

44.md
v0l: I also have tests, and it fails, unless i copy the ciphertext

staab:
> people will switch pretty quickly

No, they won't. Old messages will never be migrated, unmaintained clients will never be updated. Clients focused on other things (all of them) will wait for someone else to implement it and cause a bad UX for their users.

44.md
earonesty: `let ciphertext = xchacha20_stream(key, nonce, plaintext, plaintext)`

earonesty:
> > people will switch pretty quickly
> 
> No, they won't. Old messages will never be migrated, unmaintained clients will never be updated. Clients focused on other things (all of them) will wait for someone else to implement it and cause a bad UX for their users.

oh, for sure, it's the nature of data-at-rest cryptosystems that old versions have to be maintained for many years.   that's why it's critical to version them early on.

44.md
paulmillr: Inventing custom protocols (x as private key) sounds like a really bad idea.

I've saw security audit of a well-thought stealth address protocol, and in the end there were a few non-obvious vulnerabilities that leaded to uncovering hidden addr relationship.

I wouldn't add it unless there's a consensus/audit of at least a few cryptography folks. It's not as easy as it looks like.

v0l:
> > people will switch pretty quickly
> 
> No, they won't. Old messages will never be migrated, unmaintained clients will never be updated. Clients focused on other things (all of them) will wait for someone else to implement it and cause a bad UX for their users.

I have added support for NIP-44 on Snort, could simply enable support for receiving `kind:444` very easily 
https://git.v0l.io/Kieran/snort/commit/c2a3a706def58828d6a962834778d7c2e8e11dbc#diff-5f3569743bbd644c14b36eb1624a9f8c981434b1

44.md
v0l: Idk, it's weird in browser context, I've had this problem with `Uin8Array` before, sometimes you cannot pass it between functions and you need to copy it, otherwise the array gets deleted and you're left with `undefined`

fiatjaf:
This NIP to me without any additional (very simple) requirements such as explicit relay signaling on the receiver side and mandatory `AUTH` is a waste of everybody's time. The cryptography improves a little, but we throw away an opportunity to drastically improve metadata privacy, and then tomorrow everybody will be complaining again about the metadata leakage.

The fact that Snort has already implemented it probably without the mandatory relay stuff (since it isn't in the spec text), and that @earonesty passionately dislikes the idea of trusting relays to hide metadata make are hints to me that this NIP will end up not being a good idea.

paulmillr:
@fiatjaf I don't mind your features, just need to understand the threat model.

What happens if a user never sends AUTH? Do they never receive DMs? What happens if all relays which user send AUTH to, went down? Would messages disappear?

earonesty:
I literally proposed a way to hide the metadata just up there

fiatjaf:
> What happens if a user never sends AUTH? Do they never receive DMs? What happens if all relays which user send AUTH to, went down? Would messages disappear?

Yes, they never receive DMs. Yes, messages disappear.

earonesty:
note: this scheme is best attacked using a birthday attack on the 192 bit nonce in a extremely busy message stream (or across all known streams).   that means it's (192/2lg(n)) bits of security (a million messages == 152 bit, a billion == 132 bit. trillion = 112 bit... and so on).   if someone uses this in an automated system of some kind to encrypt trillions of messages... it begins to degrade quickly below 128 bits

paulmillr:
1. I've added NIP-42 and NIP-65. cc @fiatjaf. However, one case is unclear to me. What if I want to communicate with Alice, via specific, "more secure" relay that we both agreed upon, which is not in the list of our preferred relays? Also, how does this play with wrapper "hidden" messages that have been mentioned above?
2. With regards to content encoding, our options are:

    - hex, with splits at specific indexes ((0, 2), (2, 34)). Pros: consistency with sig and pubkey fields `1c35e24e540566e8464dfe2d5c43453d936207fdfe`
    - base64+csv. It is 1.5x shorter than hex and can save some traffic. Disadvantages would include the need to include base64 implementation (currently no other NIP requires it, besides NIP4, which we're replacing). `1,hkMpXxHhTPZA3gFYrsJPWEfGO6YBc0Lw,5EZtlb1tF4uf`

3. With regards to algorithm choice:

    - iOS has ChaCha20-Poly1305 https://developer.apple.com/documentation/cryptokit/chachapoly
    - Android has ChaCha20, ChaCha20-Poly1305 https://developer.android.com/reference/kotlin/javax/crypto/Cipher
    - HPKE rfc9180 has ChaCha20-Poly1305 https://datatracker.ietf.org/doc/rfc9180/
    - MLS has ChaCha20-Poly1305 https://datatracker.ietf.org/doc/draft-ietf-mls-protocol/
    - Sodium has box/sealedbox/secretbox xsalsa20-poly1305, secretstream xchacha20-poly1305
    - tweetnacl has xsalsa20-poly1305

Before merge, I also want to have a second opinion from someone experienced, who can do security evaluation / audit. If anyone knows such people, it would massively help.

BlowaterNostr:
Is metadata leakage and cryptographic security 2 several problems? Can we should them independently?

fiatjaf:
Thinking again about this, I think people won't switch easily, so it may be better to try to implement SimpleX instead? Even though it is too complicated, someone can take an OpenSats grant and write a SimpleX JS library and libraries for other languages, lobby the SimpleX people to enable websocket support on their servers or write a separate implementation, and allow easily hosting paid SimpleX relays or community whitelisted SimpleX relays, then we have a solid feature to build upon (assuming SimpleX is actually good).

paulmillr:
I strongly suggest we finalize NIP-44, explicitly state our threat model, do some sort of audit of the result scheme, and release it to public. This is because Simplex is extremely complicated. I will open a separate discussion with regards to Simplex estimates and amount of work, it's likely 100x more than finalizing NIP-44.

Regarding "people won't switch easily" - on a contrary, I don't think so. We just need to sell it. If it's too complex, then yes, they won't switch. If it's as simple as it gets, then switching should take no time. It's important to understand that, in the worst case, all the primitives (xchacha, whatever) are still available in extremely popular libsodium, which works on all platforms. So, it's good. For JS i've released noble-ciphers, so it's good too.

To reiterate deficiencies of NIP4: 1) cbc padding oracle attacks 2) multi-user security of AES is shit 3) cross-platform key reuse, it's easy to accidentally leak private key 4) and aes key not being IND. Maybe 4 is minor. 1 and 2 are medium. 3 is really bad, since nostr uses one key for everything.

@BlowaterNostr metadata leakage of some sort always happens. It happens in signal, it happens in tor, it happens everywhere. The best we can do is minimize it. I think fiatjaf's solution would radically decrease leakage, so it's cool.

paulmillr:
SimpleX discussion => https://github.com/nostr-protocol/nips/discussions/658

earonesty:
we're using the shared secret to sign/encrypt messages.   but are otherwise using nip-44 for encryption.   that prevents metadata leakage better than most of the above.   and its plausibly deniable.   and it's understandable/easy to use.

paulmillr:
@earonesty 

> and its plausibly deniable

how so?

44.md
```diff
@@ -0,0 +1,96 @@
...
+
+## Metadata leakage
+
+All nostr events are passed over the internet, which means some metadata would always leak.
+
+All compatible clients MUST implement following flow to limit leakage:
+
+1. Alice authenticates on `wss://nostr.example.com` using NIP-42. This would mean sending and receiving
+   messages would only be available to proper, authenticated users.
+2. On success, Alice creates NIP-65 event specifying `wss://nostr.example.com` as her preferred relay.
```
mikedilger, line 33: How about mentioning that these are Alice's `read` relays, the ones she reads from (in case she uses relays for just read or just write, instead of both).

... or better yet, since people are concerned about metadata, we extend NIP-65 to add a flag indicating a DM relay.  That way it would be possible for people concerned with metadata leakage to setup a personal relay for receiving DMs and be confident that those DMs are not being read by interlopers (at least the receipient is confident it doesn't leak from their specified read relay, but not that the sender didn't put it elsewhere; and the sender can't really be confident of this).

vitorpamplona:
Folks, I can't find a reliable xchacha20 implementation for Android. Do you have any tips? It's going to be hard to support this PR if the cryptographic function is not easily available in all languages. :(

paulmillr:
@vitorpamplona libsodium is available on all platforms.

paulmillr:
To clarify, in libsodium the function is called `crypto_stream_xchacha20`. There are bindings to basically all languages, that includes java, js, swift, and everything else.

vitorpamplona:
I got it to work with one of the bindings: https://github.com/nostr-protocol/nips/pull/468#issuecomment-1654746517

 3/4 of the libs I tested were crashing for no reason. Things are still very raw.

earonesty:
we're using a variant of this in arcade, running on android.   it doesn't use chacha, but it does the "right thing" for aes.   (new salt, new key per message):   https://github.com/ArcadeLabsInc/arclib/blob/9857e1ae1558b9e265c1461a53d281ecbe4fae77/src/ident.ts#L107

works fine in react/ios/andriod and is def better than nip04.   also we support blinded-dms as well (no metadata leaks) on android

paulmillr:
your solution is still:

- much less secure against multi-key attacks
- vulnerable to padding oracle attacks
- not compatible with nip4. if an upgrade is required, it should be a proper upgrade

> i had various problems using chacha

Which problems, exactly?

capitalistdog:
Dropping an idea I had around metadata concerns that would not alter this NIP as proposed, because if this is more cryptographically sound than NIP04, it should move forward on that alone. 

Metadata concerns can be addressed in a separate NIP, using this one as a dependency as those concerns are separate.

**"Chaum Chat"**

The objection to ephemeral conversation keys ties back to paid relays not allowing writes for new keys. With all he stupidity being pushed by NGO's to use ECash in representing custodial Bitcoin, why not re-purpose those libraries into something useful, like relay access tokens?

An additive NIP for paid relays could define how clients receive ECash instead of a whitelisting, with the ECash then being used for the whitelisting of keys. 

Additionally, free relays could reward ECash to identity keys based on reputation, to keep rate limits loose without abuse.

This process would decouple identity keys and their ephemeral children from the Lightning payment. 

Since's it's only for service access, so there's no dishonest NGO narrative of it enhancing the security/privacy of Bitcoin transactions themselves, and a demonstrated of how ECash _should_ be used.

vitorpamplona:
> "Chaum Chat"

I can't evaluate the merits of this idea without a full description. But I am happy to do so if you decide to make a NIP for it. It feels like Large Group chats could use this Ecash allowlist idea.

xz-cn:
Correct me if I am wrong, but I have a question regarding this NIP as well as NIP-04. So the ECDH part involves computation using the private key. Where exactly should this computation `getConversationKey` take place?

Ideally, users should be able to store their private key in an extension wallet such as Metamask of Ethereum. But Metamask won't let a Dapp export your private key for an ECDH computation. For Nostr, idealy it should be same—web app can't touch your private key directly. Maybe we should have a Nostr wallet which does signature and ECDH computation within the wallet instead of exporting it to a web app. Thoughts?

paulmillr:
Of course. That was one of ideas behind getConversationKey. It can be abstracted in a proper way.

staab:
I've created a new version of this PR at #715, @paulmillr please let me know if you'd like to close this PR or incorporate those changes into this one.

staab:
Closing this since the conversation is continuing at #715

PR #576, opened 2023-06-01
Update 18.md for reposting parameterized replaceable events

digi-monkey:
Reposting is a key mechanism that allows for the discovery of quality content. Therefore, I believe it would be beneficial to apply the repost feature to NIP-23 long-form articles as well.

### DIFF
--- a/18.md
+++ b/18.md
@@ -19,6 +19,10 @@ to indicate where it can be fetched.
 The repost SHOULD include a `p` tag with the `pubkey` of the event being
 reposted.
 
+## Reposting Parameterized Replaceable Events
+
+When reposting events of kind `30000 <= n < 40000`, a parameterised replaceable event as defined by [NIP-33](https://github.com/nostr-protocol/nips/blob/master/33.md), the repost SHOULD also include an `a` tag with fields `<kind>:<pubkey>:<d-identifier>`  and `relay-url`. This allows the most recent version of the parameterised event to be retrieved even if the original event is no longer available.
+
 ## Quote Reposts
 
 Quote reposts are `kind 1` events with an embedded `e` tag (see [NIP-08](08.md) and [NIP-27](27.md)).


hazeycode:
Concept ACK. But should we consider making the `e` tag mandatory such that the repost can always be traced to the version of the event that the reposter actually ACK'd? Clients and relays can simply ignore it but the information is retained.

hazeycode (changes requested):

digi-monkey:
> Are you doing this somewhere?

yes I plan to do this in the upgrade version on flycat https://github.com/digi-monkey/flycat-web/blob/develop/src/service/nip/18.ts#L8-L9

18.md
```diff
@@ -7,13 +7,12 @@ Reposts
...
 A repost is a `kind 6` note that is used to signal to followers
-that another event is worth reading.
+that another event(a kind 1 short-note or a [NIP-23](https://github.com/nostr-protocol/nips/blob/master/23.md) long-form article) is worth reading.
 
 The `content` of a repost event is empty. Optionally, it MAY contain
 the stringified JSON of the reposted note event for quick look up.
 
-The repost event MUST include an `e` tag with the `id` of the note that is
-being reposted. That tag MUST include a relay URL as its third entry
+The repost event MUST include either an `e` tag with the `id` of the note or an `a` tag with the `<kind>:<pubkey>:<d-identifier>` of the parameterized replaceable note(see [NIP-33](https://github.com/nostr-protocol/nips/blob/master/33.md)) that is being reposted. That tag MUST include a relay URL as its third entry
```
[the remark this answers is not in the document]
digi-monkey: updated

hazeycode (changes requested):
I don't think NIP-23 needs specific mention, nor that content can be used for this purpose.

How about moving the `a` tag recommendtion into a subsection titled "Reposting Parameterized Replaceable Events"?

digi-monkey:
> I don't think NIP-23 needs specific mention, nor that content can be used for this purpose.
> 
> How about moving the `a` tag recommendtion into a subsection titled "Reposting Parameterized Replaceable Events"?

sure, that is more clean. updated.

18.md
```diff
@@ -19,6 +19,12 @@ to indicate where it can be fetched.
 The repost SHOULD include a `p` tag with the `pubkey` of the event being
 reposted.
 
+## Reposting Parameterized Replaceable Events
+
+An `a` tag with `<kind>:<pubkey>:<d-identifier>`
+MAY also be included in the repost event for client to retrieve the most recent
+version of the parameterized replaceable event(see [NIP-33](https://github.com/nostr-protocol/nips/blob/master/33.md)) being reposted.
```
hazeycode, line 26: When reposting events of kind `30000 <= n < 40000`, a parameterised replaceable event as defined by [NIP-33](https://github.com/nostr-protocol/nips/blob/master/33.md), the repost SHOULD also include an `a` tag of the form `<kind>:<pubkey>:<d-identifier>`. This allows the most recent version of the parameterised event to be retrieved.

hazeycode (changes requested):

18.md
```diff
@@ -19,6 +19,10 @@ to indicate where it can be fetched.
 The repost SHOULD include a `p` tag with the `pubkey` of the event being
 reposted.
 
+## Reposting Parameterized Replaceable Events
+
+When reposting events of kind `30000 <= n < 4000`, a parameterised replaceable event as defined by [NIP-33](https://github.com/nostr-protocol/nips/blob/master/33.md), the repost SHOULD also include an `a` tag with fields `<kind>:<pubkey>:<d-identifier>`  and `relay-url`. This allows the most recent version of the parameterised event to be retrieved even if the original event is no longer available.
```
hazeycode, line 24: missing a 0 on 4000

hazeycode (approved):

vitorpamplona:
Amethyst supports reposting of any event, replaceable events included. I would just remove the word "also" from the new text. Ideally, users and clients should choose between reposting an `a` tag and thus showing the latest info of that event OR reposting the `e` tag that fixes a given version of the replaceable event in the repost event. Having both in a single repost makes it confusing.

hazeycode:
What's confusing to have both? They are both well defined. Keeping it the 'e' tag means there doesn't need to be trust, there is no reason to drop it IMO.

vitorpamplona:
If you have both, which one are you reposting? The original `e` event or an updated event through `a`?

vitorpamplona:
The issue is having 1 repost with an `e` AND an `a` tag at the same time. It's unclear which version the author wants to repost.

hazeycode:
Always the original. Including an 'a' tag simply allows finding updated versions. An end user will see no difference but we should not throw away information because of this.

vitorpamplona:
ok, but if you have the two tags in a single repost, which one do I display?

hazeycode:
> ok, but if you have the two tags in a single repost, which one do I display?

That's a good question. It depends on the client. Either displaying the latest or displaying the "original" but also that there is a new version available would be acceptable.

The point is that even if one client ignores the 'e' tag, the information is still there. I don't think there is good reason to throw the information away.

vitorpamplona:
And we can do that - Amethyst will always display the newest event in the entire list (you can have multiple `e`s and `a`s together). But that is not ideal as guidance. 

The NIP guidance should be for the author to make it clear which version he/she wants to display their followers. Don't let clients guess.

hazeycode:
You're right. The guidance should be to show the reposted version and a call-out to the updated version. But clients are free to ignore the guidance.

vitorpamplona:
I really don't understand the insistence on having the 2 tags in the same repost. But sure. Just write it there.

staab:
I'm with @vitorpamplona, `a` should be canonical for replaceables

fiatjaf:
I think this is somewhat absurd as an idea and possibly non-backwards compatible, breaking and evil. We shouldn't encourage this kind of thing.

Who on earth would repost a long-form content into a feed of text notes? Is this about trying to trick basic social clients into displaying Markdown?

I think a "mention" is the only way to go. You make the mention, add a link there, people can click and open the long-form articles elsewhere or even internally if the client supports.

If your clients really wants these things to feel like reposts, then make they show the mentions right away in full screen, it's simple and doesn't break everybody else.

vitorpamplona:
@fiatjaf I would agree but the boosted post is stringified in the `.content` of the repost event and thus if the client does not support certain event types, they can see it right away and not show it. There is not even a second request to the relay to download the event. 

Also, frankly, people think Nostr is only about kind 1s. What if I have a feed of NIP94 images and people want to repost them? What about reposting inside chats? What about reposting lists? Do I need to create a new Repost Event type for every event type out there? 

That doesn't seem right.

fiatjaf:
I think it is reasonable to create a new repost kind for each type of content.

kind:1 stuff gets more visibility because it was the first use and the more popular etc, but having separate kinds for different things is good, keeps queries small and clean and options optional.

That is actually probably what we should do here: specify a new event kind for reposts of kind:30023 events. Probably put that on NIP-23.

vitorpamplona:
@arthurfranca is adding the `k` tag for event kind in many cases like this. Could he solve this by simply identifying the kind in the repost event in such a way that clients can filter for only kind 1 reposts?

arthurfranca:
Ignoring the possible problem of breaking social clients that may not check if the repost is really of a kind 1 event (but hey this feels like an implementation mistake that can be abused)...

Adding a `k` tag seems like a good fit here to help query exactly the reposts a client support. Similar to the use on gift wraps, the member list PR and Moderated Communities too.

fiatjaf:
Composability is bad for interoperability. It's better to make a new kind.

fiatjaf:
Just to clarify my point above and not sound like a jerk, let me say this:

What is the practical difference between having a query with `{"kinds": [10023]}` and a `{"kinds": [6], "#k": [30023]}`? The second one is more elegant, I grant you that, but it's also less efficient on the relay querying side and requires a type conversion.

The first one, while ugly, must necessarily be specified somewhere in a NIP, which makes it clear for everybody how it is supposed to be used -- and shows that someone is using it; while the second, once we have such ergonomic queries, will be used without being included in any specs, leading to chaos.

hazeycode (commented):
> I think it is reasonable to create a new repost kind for each type of content.

I agree with this.

@fiatjaf has made good points, I revoke my ACK, FWIW.

staab:
> Composability is bad for interoperability. It's better to make a new kind.

This makes me sad, I wonder how true it is. 🤔

fiatjaf:
More problems with using a `k` tag instead of a new kind:
- existing clients will all break because they all assume kind 6 contain text notes inside (even if they don't break they will still have to downloads tons of data they don't want and filter that data out on the client, only after parsing everything etc)
- uses more data

vitorpamplona:
> existing clients will all break because they all assume kind 6 contain text notes inside

Can we do a quick survey to see which ones actually break? 

Don't get me wrong, I am all for being overzealous to keep things working if it is actually needed. But if clients already deal with this (I believe no one actually thinks reposts are only for kind-1s) then making decisions **assuming** that they are not is highly damaging to the evolution of the protocol.

vitorpamplona:
Habla has been using Kind 6 for Long-form content for a long time now. We should have heard people complaining at some point in the past if clients aren't ready for it. 

I just boosted Jack's post on Habla to see if anyone complains.

fiatjaf:
Indeed it does. I still think we shouldn't do it and we should ask Habla to move to the new kind. I just tried it too and it doesn't break things.

But what is probably happening is that normal kind:1 social clients are having to download these giant notes and then hide them from the feed. It's just bad manners.

staab:
I don't support re-posts, but I do support nevent embeds. My goal is to be able to handle _any_ event kind and show a reasonable summary (even if it's pretty minimal). I also want to be able to show many types of event kinds in user feeds. For example, I recently added 1985-based relay reviews. I think it would be awesome for those to show up in your normal social feed — your friend rated a restaurant, that's actually pretty valuable information, and you'll be interested. It certainly beats sending pictures of your food.

Tangential to this actual issue, but germane to the discussion. Interoperability is of course paramount, but having a powerful protocol is also extremely important.

fiatjaf:
> That's why I think the k is a good addition.

Sure, we can do it with the `k` tag, but we go back to my previous point: is using `k` better than using a new kind? No, it is not, as I argued above.

> As a client, you cannot expect Nostr payloads to be well-formatted or to follow any particular guidance no matter how well-defined it is.

I know, but still the guidance should be as best as possible for those who want to follow them. And you can always unfollow people that are trying to screw you by sending bogus events to your feed. If the spec mandates bogus events then that becomes impossible.

fiatjaf:
Other two points in favor of a new kind:

1. Existing clients that are sending kind:6 events for reposts don't have to update their implementations to add a `k` tag with value of `1`, they can just keep existing.
2. And to @staab's point, clients should be able to pick what type of content they want to display in their feed. It's very easy today for a client to use a filter like `{"kinds": [1, 6], "authors": [...]}`. If they want, they can add 1985 or whatever other kind they want in that same array -- but if now we have kind:6 events that require an extra `k` tag then they will need a new filter, it puts more burden on relays, more work everywhere, more data to send, more possibilities of failure etc.

@pablof7z you asked to be mentioned here.
@verbiricha your webapp is destroying the world.

pablof7z:
I disagree with this position; nostr gives us incredible composability, and NIP-31+NIP-89 make it so this composability is not bad for interoperability; quite the opposite is GOOD for interoperability, as a `kind:1`-centric client all of the sudden gains access to marketplaces, song distribution, note taking, and whatever (~all) other use cases we end up building in nostr.

Why create kind silos now that we have these primitives for cross-kind interoperability? I would have agreed with @fiatjaf's heart-wrenching "Composability is bad for interoperability. It's better to make a new kind." but with those two NIPs this is no longer true, and in fact, this seemingly-chaotic composability can (will) unlock functionalities that were simply not possible before nostr. (yes, I repeat myself)

----

I do like the `k` tag, in fact, I had the same idea while writing highlighter; beyond being able to quickly find `kind:9802` reposts with a `{kinds:[6], "#k":[9802]}` filter, the `k` tag also would allow a `kind:1`-centric client to decide whether to fetch the reposted event or to simply ignore it (which it shouldn't imo).

fiatjaf:
> the k tag also would allow a kind:1-centric client to decide whether to fetch the reposted event or to simply ignore it (which it shouldn't imo).

How can the kind:1 client ignore it? All kind:6 events will come when the kind:1 client requests `{"kinds": [1,6]}`, no matter what is their `k` tag, because the kind:1 client doesn't know about these tags, also all the other kind:1 clients are not writing `k` tags on their repost events.

fiatjaf:
image

fiatjaf:
If you really want to use the `k` tag, then at least make a new kind that supports the `k` tag and leave the current behavior of kind:6 unchanged.

vitorpamplona:
> If you really want to use the `k` tag, then at least make a new kind "generic_repost" that supports the `k` tag and leave the current behavior of kind:6 unchanged.

Not to disappoint but I'd say kind:6 has already changed (months ago). The NIP is just very outdated and doesn't match events in the network anymore. Asking it to go back won't delete the events that are already there.

staab:
> If you really want to use the k tag, then at least make a new kind "generic_repost" that supports the k tag and leave the current behavior of kind:6 unchanged.

The draft for #468 could support this use case very easily.

fiatjaf:
> Not to disappoint but I'd say kind:6 has already changed (months ago). The NIP is just very outdated and doesn't match events in the network anymore. Asking it to go back won't delete the events that are already there. We might as well embrace it by now.

What has changed about it?

vitorpamplona:
> > Not to disappoint but I'd say kind:6 has already changed (months ago). The NIP is just very outdated and doesn't match events in the network anymore. Asking it to go back won't delete the events that are already there. We might as well embrace it by now.
> 
> What has changed about it?

The fact that we have 1000s of reports on replaceable events, other multiple 'e' stuff, and reposts of non-kind 1s.  

If clients have been using it without any issue for months, why should we change the client and not the NIP? 

I am against any policy document that goes against the practice of it.

vitorpamplona:
It's also good to remind everyone that NIP-18 never claimed that it was designed for Kind 1s only. Clients using it for any other kind are actually fully complying with the NIP.

fiatjaf:
You don't have to be such a purist, you can be more pragmatic.

Let's forget the past mistakes and create a new `generic_repost` event.

pablof7z:
> Let's forget the past mistakes and create a new `generic_repost` event.

ACK

vitorpamplona:
> You don't have to be such a purist, you can be more pragmatic.

Pragmatic person in me says: "Shit is working. Stop regulating things that work". 

> Let's forget the past mistakes and create a new generic_repost event.

People are free to create one if they think it's better. Amethyst will fully support it. Amethyst will also fully support the current use of Kind 6 for anything people will actually use it for, complying or not to a NIP.

fiatjaf:
> Pragmatic person in me says: "Shit is working. Stop regulating things that work".

You're not thinking about the long term. Also I just literally posted an example of a thing that is _not_ working.

> People are free to create one if they think it's better. Amethyst will fully support it. Amethyst will also fully support the current use of Kind 6 for anything people will actually use it for, complying or not to a NIP.

We must keep the ecosystem open for all the people who want to make apps and don't have a full-time team dedicated to their apps like you have on Amethyst (yes, you are the team). Also we should try to not bother people who may not be interested in implementing every new thing someone else comes up with and just wants to keep using their app as they always did.

vitorpamplona:
> We must keep the ecosystem open for all the people who want to make apps and don't have a full-time team dedicated to their apps like you have on Amethyst (yes, you are the team).

I am nowhere near full-time. 

> Also we should try to not bother people who may not be interested in implementing every new thing someone else comes up with and just wants to keep using their app as they always did.

I don't think we have any choice in this. Anyone can do anything in nostr. 

In this kind6 case, they have been and will continue to receive non-kind-1s boosts no matter what we do here.

mikedilger:
> Can we do a quick survey to see which ones actually break?

A quick peek at the code shows that gossip was assuming the events inside of kind-6 were "feed related" events. When it gets around to displaying them, if they are not supported by gossip, they get displayed as "NON FEED RELATED EVENT" rather than dropped entirely, an issue I could fix.  But I would still be downloading events that I couldn't display.

> Not to disappoint but I'd say kind:6 has already changed (months ago).

Existing kind-6 events don't have 'k' tags.

There will always be these "bad" kind-6 events that embed weird-undisplayable-kinded events, and gossip for example will just drop them. But I'd rather not download them in the first place.  Kind-16 https://github.com/nostr-protocol/nips/pull/610 let me deal with 'k' tags and avoid downloading them, and creates the composability everybody is clamoring on about, while fixing future kind-6 events so less and less of them will be wastefully downloaded.

fiatjaf:
@digi-monkey I hope https://github.com/nostr-protocol/nips/pull/610 is solving all your problems.

https://habla.news/ has also already stopped sending kind:6 and is sending kind:16 now: https://github.com/verbiricha/habla.news/commit/026b678faa1445a7fe994c49201a1e1dbbb0eea2

digi-monkey:
> @digi-monkey I hope #610 is solving all your problems.
> 
> https://habla.news/ has also already stopped sending kind:6 and is sending kind:16 now: [verbiricha/habla.news@026b678](https://github.com/verbiricha/habla.news/commit/026b678faa1445a7fe994c49201a1e1dbbb0eea2)

cool, a new kind is better for this

PR #580, opened 2023-06-04
NIP-112: Encrypted Group Chat

earonesty:
This NIP builds on the encryption from NIP-44, the gift-wrap of NIP-59 and the distributed chat channel management of NIP-28.   It is similar to NIP38, with the exception that gift-wrap is used on every message, and issues like metadata leakage and optional forward secrecy are addressed.    Also lists of participants are not published.

It works by creating a shared secret that is used to encrypt the group or channel messages. Only the participants know that secret, so only such members can read/write to this group chat.  This effectively hides metadata from external users.

### DIFF
--- a/112.md
+++ b/112.md
@@ -8,11 +8,11 @@ Encrypted Group Chat
 
 This NIP defines new event kinds for encrypted group chat with optional forward secrecy guidelines. 
 
-This NIP builds on the encryption from NIP-44, the gift-wrap of NIP-59 and the distributed chat channel management of NIP-28.   It is similar to NIP38, with the exception that gift-wrap is used on every message, and issues like metadata leakage and optional forward secrecy are addressed. 
+This NIP builds on the encryption from NIP-44, the gift-wrap of NIP-59 and the distributed chat channel management of NIP-28.   It is similar to NIP38, with the exception that gift-wrap is used on every message, and issues like metadata leakage and optional forward secrecy are addressed.   
 
 It works by creating a shared secret that is used to encrypt the group or channel messages. Only the participants know that secret, so only such members can read/write to this group chat.  This effectively hides metadata from external users.   See drawbacks below.
 
-It reserves 4 event kinds (104) for immediate use.
+It reserves 8 event kinds (1059, 400-406) for immediate use.
 
 - [`1059`](https://github.com/nostr-protocol/nips/pull/468) - encrypted gift-wrap, can be used by many protocols
 - `400`  - wrapped kind: create encrypted channel
@@ -20,6 +20,8 @@ It reserves 4 event kinds (104) for immediate use.
 - `402`  - wrapped kind: change channel metadata
 - `403`  - wrapped kind: send encrypted message
 - `404`  - wrapped kind: moved to new channel
+- `405`  - wrapped kind: delegate new owner
+- `406`  - wrapped kind: accept owner delegation
 
 
 ## Kind 1059: Encrypted Gift Wrap
--- a/112.md
+++ b/112.md
@@ -1,0 +1,240 @@
+NIP-112
+=======
+
+Encrypted Group Chat
+--------------------
+
+`draft` `optional` `author:earonesty` 
+
+This NIP defines new event kinds for encrypted group chat with optional forward secrecy guidelines. 
+
+This NIP builds on the encryption from NIP-44, the gift-wrap of NIP-59 and the distributed chat channel management of NIP-28.   It is similar to NIP38, with the exception that gift-wrap is used on every message, and issues like metadata leakage and optional forward secrecy are addressed. 
+
+It works by creating a shared secret that is used to encrypt the group or channel messages. Only the participants know that secret, so only such members can read/write to this group chat.  This effectively hides metadata from external users.   See drawbacks below.
+
+It reserves 4 event kinds (104) for immediate use.
+
+- [`1059`](https://github.com/nostr-protocol/nips/pull/468) - encrypted gift-wrap, can be used by many protocols
+- `400`  - wrapped kind: create encrypted channel
+- `401`  - wrapped kind: invite to encrypted channel
+- `402`  - wrapped kind: change channel metadata
+- `403`  - wrapped kind: send encrypted message
+- `404`  - wrapped kind: moved to new channel
+
+
+## Kind 1059: Encrypted Gift Wrap
+
+For all encrypted group chat messages, the client first generates a kind 4XX message, as below.   
+
+Then they generate a new public/private keypair, and uses this to sign and encrypt the message.
+
+The `content` is [NIP-44](https://github.com/nostr-protocol/nips/pull/574) encrypted JSON string with the temporary key, this is the inner event.
+
+All events called "wrapped-kind XX" refer to these kind 1059 gift wrapped events.
+
+## Wrapped Kind 400: Create Encrypted channel
+
+Create a Encrypted chat channel.
+
+In the channel creation `content` field, Client SHOULD include basic channel metadata (`name`, `about` and `picture`).   
+
+This is akin to NIP-28, kind 40 but using gift-wrap encryption.
+
+On creating a channel, the creator MUST also generate a unique/new private-public key pair which will serve as the `shared-secret` for a given channel. 
+
+This is the "destination pubkey" for the channel creation.
+
+```json
+{
+    "content": "{\"name\": \"Demo Channel\", \"about\": \"A test channel.\", \"picture\": \"https://placekitten.com/200/200\"}",
+}
+```
+
+The wrapped INNER event id for the channel creation should be saved.   This is the event id that will be used in all future INNER references to the channel creation.
+
+Channel creation events MUST be sent to the "destination pubkey" of the channel itself, not to any specific user.
+
+As a rule, this "destination pubkey" of the channel MUST used as the sole identifier and external filter that clients use to retrieve events for the channel.
+
+There are no other public tags or metadata, all other kinds and tags are "inner" (inside the wrapped message).  
+
+## Wrapped Kind 401: Invitation to Encrypted Channel
+ 
+Invitations are sent directly to the participants of the channel.
+
+```json
+{
+    "content": "optional personalized invitation message",
+    "tags": [["e", <channel-create-event-id> <relay-urls>], ["privkey": [<channel-private-key>]],
+}
+```
+
+When wrapping invitation messages, the destination pubkey is the user being invited, the relay used should be one of that user's well-known relays.  
+
+More than one relay can be recommended in the "e" tag.
+
+## Wrapped Kind 402: Change channel metadata
+
+Update an encrypted channel's public metadata.
+
+Clients and SHOULD handle wrapped-kind 402 events similar to kind 33 `replaceable` events.   Relays cannot distinguish this from other kind 1059 events.
+
+Clients SHOULD ignore kind wrapped-402 from pubkeys other than the current owner.
+
+The current owner of a channel is the pubkey that created the given channel using the corresponding wrapped-kind 400 event, or the current owner based on the 405-406 channel ownership delegation flow. 
+
+In other words, only the owner of a channel can *effectively* send kind wrapped-402 events; kind 402 for any channel sent by other people, even participants of that channel, are ignored.
+
+Clients SHOULD support basic metadata fields:
+
+- `name` - string - Channel name
+- `about` - string - Channel description
+- `picture` - string - URL of channel picture
+
+Clients MAY add additional metadata fields.
+
+Clients MUST specify an "e" tag to identify the channel id.
+Clients SHOULD mark the "e" tag to recommend a relay.  
+
+```json
+{
+    "content": "{\"name\": \"Updated Demo Channel\", \"about\": \"Updating a test channel.\", \"picture\": \"https://placekitten.com/201/201\"}",
+    "tags": [["e", <channel-wrapped-400-event-id> <relay-url>]],
+    ...
+}
+```
+
+
+## Kind 403: Send encrypted message to encrypted group channel
+
+Clients SHOULD use a marked "e" tags to specify if it is is a reply message.
+
+Clients SHOULD use [NIP-10](10.md) marked "e" tags to recommend a relay and specify whether it is a reply or root message.
+
+Clients SHOULD append [NIP-10](10.md) "p" tags to replies.
+
+### Format
+
+Inner-root message:
+
+```json
+{
+    "content": <string>,
+    "tags": [["e", <kind_400_event_id>, <relay-url-where-it-was-seen>, "root"], <optional expiration tag>]
+    ...
+}
+```
+
+Inner-reply to another message:
+
+```json
+{
+    "content": <string>,
+    "tags": [
+        ["e", <kind_400_event_id>, <relay-url>, "root"],
+        ["e", <kind_403_reply_to_event_id>, <relay-url>, "reply"],
+        ["p", <pubkey>, <relay-url>],
+        ...
+    ],
+    ...
+}
+```
+
+### Wrapped-kind 404: Removing Participants
+
+Send a Wrapped Kind 404 (similar to 401)
+
+Once a group is created, all participants know the shared-secret, so to remove a participant, we need to create a new group.
+
+To do, this the creator of the group issues a new 400 event.
+
+Then the creator sends out a kind 404 individually to every member of the new group, *excluding the removed participants*.
+
+```json
+{
+    "content": "optional friendly invite message",
+    "tags": [["e", <previous-kind-400>, <relay-url>], ["e", <new-kind-400>, <relay-url>], ["privkey", [<channel-private-key>]]
+}
+```
+Clients SHOULD ignore wrapped-kind 404 messages from anyone but the current owner of a group. 
+
+Clients SHOULD ignore messages to the previous group that occur after their receive the first kind 404.  In addition there is no longer a need to subscribe to the public keys of prior groups.
+ 
+Clients use kind 404 to find out which was the precursor channel(s) to this new channel, and should attempt to display all old and new channels in a chain seamlessly to the user as a unified channel history.
+
+Clients SHOULD consider the new channel metadata as authoritative.  
+
+Clients MAY discard previous channel information, since it will no longer be used.
+
+### Wrapped-kind 405: Delegate new owner
+
+To do, this the creator of the group sends out a wrapped-kind 405 to the main group public key.
+
+```json
+{
+    "content": "optional friendly invitation to own the group",
+    "tags": [["e", <400-event-id-of-group>, <relay-url>], ["p", <public-key-of-delegate>]]
+}
+```
+Clients SHOULD ignore wrapped-kind 405 messages from anyone but the current "owner" of a group. 
+Clients SHOULD ignore wrapped-kind 405 messages until they receive a 406 CLAIM message from the new "owner" of the group. 
+
+### Wrapped-kind 406: Accept ownership transfer
+
+To do, this the creator of the group sends out a wrapped-kind 405 to the main group public key.
+
+```json
+{
+    "content": "optional friendly hey, i'm in charge",
+    "tags": [["e", <event-id-of-405>, <relay-url>]]
+}
+```
+Clients SHOULD ignore wrapped-kind 406 messages unless they are signed by the delegate in the associated 405 message
+Clients SHOULD treat accepted delegates as the new owner of the group for all 404 and 405 messages.
+
+## NIP-10 relay recommendations
+
+For [NIP-10](10.md) relay recommendations on replies and on channel creation info, clients generally SHOULD use the relay URL where the event was first seen, if known.
+
+
+Future extensibility
+--------------------
+
+We reserve wrapped-event kinds 407-419 for other events related to chat, to potentially include new types of media (photo/video), delegation, moderation, mute, hide, etc.
+
+Motivation
+----------
+This is the easiest way to allow the use of group chat with select group of people, while hiding metadata.
+
+Optional Forward Secrecy
+------------------------
+
+Claims of "forward" secrecy in mobile applications can be disingenuous.   Many mobile applications allow users to purchase a new phone and restore their account with historic messages intact - therefore this is not forward secrecy.
+
+True forward secrecy relies on servers reliably discarding the intermediate data - in addition to frequent key rotations, or prev-key ratcheting.   A trusted server, for example, can be requested to discard old messages, by adding an optional NIP-40 expiration.
+
+### OPTIONAL forward secrecy extension
+
+The owner of a channel can add a "expiration" tag to the channel creation message.
+
+The owner of a channel who adds a "expiration" tag SHOULD issue a 404 message before the expiration.   
+
+All members of a channel SHOULD also adding a NIP-40 "expiration" tag to all messages sent to the channel using the duration (`expiration - created_at`) of the wrapped-400 create.
+
+Members of a channel MUST NOT send new messages if a 404 message has not been received before expiration.
+
+Servers SHOULD obey NIP-40 expiration tags.
+
+Provided that servers obey NIP-40 expiration tags, this will prevent an attacker from reading old group messages before the expiration.
+
+
+Drawbacks
+---------
+Any member of the group can, implicitly, invite new members to the group, since they have the private key.  Any member of a group can "dox" other members by publishing their wrapped messages.
+
+Any member of the group can spam the group, or otherwise DOS the group.
+
+If any single participant of the group chat leaks the shared secret ( whether intentionally or by accident), all the messages can then be decrypted by others until the next 400 and 404 rotation events.
+
+The use of optional frequent forward secrecy rotation events can mitigate these attacks, provided the server is compliant with the expiration times,
+


fyookball:
Is it correct that this doesn't hide who is posting to the group?   If so , can we also use nip76 style encapsulation to hide the pubkeys of who is posting? As long as we're combining several nips, might as well add this valuable feature.

Also, it should be made explicit in the NIP how to specify (multiple) relays for the chat... This was not adequately documented in nip-28, seems like the correct way is multiple e-tags in the metadata.

earonesty:
> Is it correct that this doesn't hide who is posting to the group?

please look at draft nip59

Maybe I should reproduce it here

The sender *always rolls a new public private key pair every time they post to the group*.  indeed every message is sent this way even the invitations.

there's an outer message (using an ephemeral public private keypair) and an inner message (which is signed by the actual sender)

Only members can verify the inner events and clients must validate signatures, since the relays cannot do it

I will make this more clear
 
> Also, it should be made explicit in the NIP how to specify (multiple) relays for the chat... This was not adequately documented in nip-28, seems like the correct way is multiple e-tags in the metadata.

agreed I will make that modification that is the correct way

fyookball:
So the gift wrap is with a temporary key?  That's fine but where do the other chat participants get the key from?  I would understand if the wrapping for each message is done with the same key as is used to get access to the encrypted room, but if a new key is used each time, then where do the others get the key from?

earonesty:
> So the gift wrap is with a temporary key? That's fine but where do the other chat participants get the key from? 

 - The sender's key is ephemeral.   
 - The recipient's key is the known public key of the chatroom.  
 - The other participants receive invitations from the room creator (or other participants)
 - All messages use ephemeral sender keys but well-known recipient keys.   There's no purpose or value to obscuring recipient keys in any protocol.

paulmillr:
keep in mind that Signal group chats are basically a bunch of 1-to-1 chats. There is no shared secret. I wonder which solution is better.

fyookball:
@earonesty ok i get it. sorry for the confused question. thanks for explaining.  I like this nip.  Would implement.  +1 on the relay thing.

earonesty:
> keep in mind that Signal group chats are basically a bunch of 1-to-1 chats. There is no shared secret. I wonder which solution is better.

agreed, it's all tradeoffs.   i think nostr has excellent persistence, stronger censorship resistance, and is far easier to understand and build on top of.  i like the idea of a full encrypted group chat lib, replete with all the subscriptions, persistence layer, etc , in easy-to-understand layers of code that anyone can make a cli, or an app, or a webapp on top of.   and i love the fact that the servers are "dumb" nostr relays, although this proposal does push a lot of work onto the clients (including key rotation), which could be expensive.   hard to say.   it seems to be working so far.   code isn't complete, but you can build the app right now off main and start chatting.

staab:
This is a very well written nip, but it states:

> This is the easiest way to allow the use of group chat with select group of people, while hiding metadata.

Which I don't think is true. I recently wrote a note about this topic, which you can read [here](https://habla.news/a/naddr1qqwkzttswfhhqmmnv9kz6en0wgkhqunfweshgefdvaex7atswvq3samnwvaz7tmjv4kxz7fwwdhx7un59eek7cmfv9kqzrthwden5te0dehhxtnvdakqzynhwden5te0wp6hyurvv4cxzeewv4esz9rhwden5te0wfjkccte9ejxzmt4wvhxjmcpzemhxue69uhhyetvv9ujumn0wd68ytnzv9hxgqgawaehxw309ahx7um5wgkhqatz9emk2mrvdaexgetj9ehx2aqpz3mhxue69uhhyetvv9ujumn0wd68ytnpdyq32amnwvaz7tmwdaehgu3wxpur2vpww3jkx6qprpmhxue69uhkummnw3ezu6m0d3kxjer9wgh8s7t6qyfhwumn8ghj7un9d3shjctzd3jjummjvuq3yamnwvaz7tm0venxx6rpd9hzuur4vgqswcm0wfskxmr9qgsf03c2gsmx5ef4c9zmxvlew04gdh7u94afnknp33qvv3c94kvwxgsrqsqqqa2884vczr). It shares the gift-wrap approach, but makes different tradeoffs.

The main trade-off of my idea is that it relies on an access-controlled nsec bunker rather than a private key shared directly over the network. I tend to think that's an improvement, because you get a lot more flexibility and control, not to mention consistency, without making clients do super complicated things to get the right channel metadata.

My main point here though is that my approach basically creates a generic cryptographic subset of the entire nostr protocol using gift wrap, rather than specifying each individual component independently. Private chat is an obvious start, but with this PR every time a feature is added to chat, two NIPs need to be updated. Likewise, every time we want encryption for some other part of the protocol (e.g. kind 1s), a new NIP needs to be added.

I would strongly prefer to see a new NIP that specifies a generic way to 1. define channels based on a shared public key, 2. communicate over gift-wrapped channels, and 3. specifies multiple ways to manage the key rotation (nsec bunker and DM'd private keys are both viable alternatives, as well as FROSTR potentially).

earonesty:
> I would strongly prefer to see a new NIP that specifies a generic way to 1. define channels based on a shared public key, 2. communicate over gift-wrapped channels, and 3. specifies multiple ways to manage the key rotation (nsec bunker and DM'd private keys are both viable alternatives, as well as FROSTR potentially).

genuinely confused.   how does this not satisfy that?   #1 check, #2 check #2 ... "bunker" based approach can be used if you simply don't invite people.   you can make an invitation to the bunker instead.   probably it's not more secure though,

staab:
> "bunker" based approach can be used if you simply don't invite people. you can make an invitation to the bunker instead.

Yes, I just think the authentication method should be decoupled from the data structure so you can more easily choose your solution.

> probably it's not more secure though, also not decentralized.

Also, true, but more flexible and simpler. It doesn't make sense to me to re-implement centralized administration features on top of a decentralized, eventually consistent network. 

My main point was the "a generic way" bit, which probably should have been its own bullet point. This NIP solves chat, but it doesn't solve private kind 1 for example, or anything else. I think we should start by creating a way to generically create "groups" within which a nostr sub-network can be run. That automatically solves groups, along with many other problems. Again, I know the nsec bunker vs in-protocol key management is contentious, specifying the auth strategy when creating the group would be a good way to decouple and keep possibilities open for future improvements. I will try to create a NIP later this week so I can explain myself more concretely.

earonesty:
#468 does what you want.  it does solve private kind 1.

staab:
Awesome, I see, so apart from the nsec bunker vs in-protocol group management tradeoff (which we agree can coexist), we are on the same page. To make things clearer, I'd suggest removing the term "chat" entirely from this PR (except in the motivation/use cases section), and defining the term "channel" or "group" at the top of the NIP as something like the nostr subnet concept. I can then create my own PR for nsec-bunker-controlled channels at some future date.

earonesty:
that sounds reasonable, thanks.

staab:
Ok, now that we have that figured out, let me propose a slightly different design. Instead of following the NIP 40 channel definition model, what about overloading pubkeys? IOW if a pubkey represents an "entity" rather than a "person", you could create a "channel" by publishing a kind 0 (for public metadata), or a gift-wrapped kind 0 (for private metadata). Then all these 4xx events would be used to extend/revoke the ability for channel "members" to publish data on behalf of/to the group. This seems simpler, since you need a shared key anyway, you might as well refer to self rather than a sort of arbitrary event id.

The same NIP could also be used to implement key rotation for regular accounts. I don't think it would increase the attack surface area for rogue group members by widening the scope of published events to public, since they have the private key anyway in your scenario and could publish weird public claims on behalf of the whole group.

One other small thing, might 405/406 be expanded to support multiple admins?

earonesty:
> Ok, now that we have that figured out, let me propose a slightly different design. Instead of following the NIP 40 channel definition model, what about overloading pubkeys? IOW if a pubkey represents an "entity" rather than a "person", you could create a "channel" by publishing a kind 0 (for public metadata), or a gift-wrapped kind 0 (for private metadata). Then all these 4xx events would be used to extend/revoke the ability for channel "members" to publish data on behalf of/to the group. This seems simpler, since you need a shared key anyway, you might as well refer to self rather than a sort of arbitrary event id.

what key would you encrypt with?

> One other small thing, might 405/406 be expanded to support multiple admins?


it could be, but it's hard in some ways.   because then you have admins that can "disagree" and "fork the group" into two groups, depending who you saw first.   i think it's probably better to start small

staab:
Yeah, you would just do symmetric encryption using the group's key.

> any time you add/remove members

I see what you mean. The approach I propose would make the nsec bunker approach simpler, but wouldn't help at all with the in-protocol approach since you have to rotate keys anyhow.

Thanks for working through my questions, I believe I can retrofit a custodial key approach onto either this nip or the gift wrap one directly if I end up going that direction.

112.md
```diff
@@ -0,0 +1,248 @@
...
+=======
+
+Encrypted Group Events
+----------------------
+
+`draft` `optional` `author:earonesty` 
+
+This NIP defines new event kinds for encrypted group events, with chat as the primary example, and with optional forward secrecy guidelines. 
+
+This NIP builds on the encryption from NIP-44, the gift-wrap of NIP-59 and the distributed chat channel management of NIP-28.   It is similar to NIP38, with the exception that gift-wrap is used on every message, and issues like metadata leakage and optional forward secrecy are addressed.
```
staab, line 11:
```suggestion
This NIP builds on the encryption from NIP-44, the gift-wrap of NIP-59 and the distributed chat channel management of NIP-28. It is similar to NIP-38, with the exception that gift-wrap is used on every message, and issues like metadata leakage and optional forward secrecy are addressed.   
```

112.md
```diff
@@ -0,0 +1,248 @@
...
+Encrypted Group Events
+----------------------
+
+`draft` `optional` `author:earonesty` 
+
+This NIP defines new event kinds for encrypted group events, with chat as the primary example, and with optional forward secrecy guidelines. 
+
+This NIP builds on the encryption from NIP-44, the gift-wrap of NIP-59 and the distributed chat channel management of NIP-28.   It is similar to NIP38, with the exception that gift-wrap is used on every message, and issues like metadata leakage and optional forward secrecy are addressed.   
+
+It works by creating a shared secret that is used to encrypt the group or channel messages. Only the participants know that secret, so only such members can read/write to this group chat.  This effectively hides metadata from external users.   See drawbacks below.
```
staab, line 13:
```suggestion
It works by creating a shared secret that is used to encrypt the group or channel messages.
Only the participants know that secret, so only such members can read/write to this group chat.
This effectively hides metadata from external users. See drawbacks below.
```

112.md
```diff
@@ -0,0 +1,248 @@
...
+
+## Kind 1059: Encrypted Gift Wrap
+
+For all encrypted group chat messages, the client first generates a kind 4XX message, as below.   
+
+Then they generate a new public/private keypair, and uses this to sign and encrypt the message.
+
+The `content` is [NIP-44](https://github.com/nostr-protocol/nips/pull/574) encrypted JSON string with the temporary key, this is the inner event.
+
+All events called "wrapped-kind XX" refer to these kind 1059 gift wrapped events.
```
staab, line 35:
```suggestion
For all encrypted group chat messages, the client first generates a kind 400 message, using a fresh public/private keypair.

The `content` is [NIP-44](https://github.com/nostr-protocol/nips/pull/574) encrypted JSON string with the temporary key, this is the inner event.

All events called "wrapped-kind XX" refer to kind XX events wrapped using kind 1059 gift wrapped events.
```

112.md
```diff
@@ -0,0 +1,248 @@
+NIP-112
+=======
+
+Encrypted Group Events
+----------------------
+
+`draft` `optional` `author:earonesty` 
+
+This NIP defines new event kinds for encrypted group events, with chat as the primary example, and with optional forward secrecy guidelines.
```
staab, line 9:
```suggestion
This NIP defines new event kinds for encrypted channels with optional forward secrecy guidelines.
A channel is a way to create a private nostr sub-network, within which the protocol can operate normally
visible only to users entrusted with a given decryption key.
```

112.md
```diff
@@ -0,0 +1,248 @@
...
+}
+```
+
+The wrapped INNER event id for the channel creation should be saved.   This is the event id that will be used in all future INNER references to the channel creation.
+
+Channel creation events MUST be sent to the "destination pubkey" of the channel itself, not to any specific user.
+
+As a rule, this "destination pubkey" of the channel MUST used as the sole identifier and external filter that clients use to retrieve events for the channel.
+
+There are no other public tags or metadata, all other kinds and tags are "inner" (inside the wrapped message).
```
staab, line 64: What about optionally adding tags to the outer event, but encrypted? So `["p", <pubkey>]` would be added to the wrapper as `["p", "<encrypted pubkey>"]`. This would allow users in the know to still execute filters, supporting larger groups (though still not scaling as well as native nostr). A new `k` tag could be used to denote the kind of the inner event as well.

A 1k member group, each publishing 1k events over the course of a year would result in 1MM events, which would be far too many to process without filters.

112.md
```diff
@@ -0,0 +1,248 @@
...
+Invitations are sent directly to the participants of the channel.
+
+```json
+{
+    "content": "optional personalized invitation message",
+    "tags": [["e", "channel-create-event-id", "relay-url"], ["privkey", "channel-private-key"]]
+}
+```
+
+When wrapping invitation messages, the destination pubkey is the user being invited, the relay used should be one of that user's well-known relays.
```
staab, line 77:
```suggestion
When wrapping invitation messages, the destination pubkey is the user being invited, not the pubkey of the channel. The relay used should be one of that user's well-known relays.  
```

112.md
```diff
@@ -0,0 +1,248 @@
...
+```json
+{
+    "content": "optional personalized invitation message",
+    "tags": [["e", "channel-create-event-id", "relay-url"], ["privkey", "channel-private-key"]]
+}
+```
+
+When wrapping invitation messages, the destination pubkey is the user being invited, the relay used should be one of that user's well-known relays.  
+
+More than one relay can be recommended in the "e" tag.
```
staab, line 79:
```suggestion
Multiple "e" tags can be used to recommend multiple relays.
```

Not sure about this one, but it seems better than putting a relay url where a mark usually is. Another good option would be to use separate `e` and `r` tags.

112.md
```diff
@@ -0,0 +1,248 @@
...
+```
+
+
+## Kind 403: Send encrypted message to encrypted group channel
+
+Clients SHOULD use a marked "e" tags to specify if it is is a reply message.
+
+Clients SHOULD use [NIP-10](10.md) marked "e" tags to recommend a relay and specify whether it is a reply or root message.
+
+Clients SHOULD append [NIP-10](10.md) "p" tags to replies.
```
staab, line 121:
```suggestion
Clients SHOULD use marked "e" tags to specify if it is a reply message.

Clients SHOULD use [NIP-10](10.md) marked "e" tags to recommend a relay and specify whether it is a reply or root message.

Clients SHOULD append [NIP-10](10.md) "p" tags to replies.
```

See above about encrypting tags. Leaking `e` tags is ok, but leaking `p` tags seems like a problem. Maybe group members wouldn't even want anyone to know they're members of a particular group.

112.md
```diff
@@ -0,0 +1,248 @@
...
+
+### Format
+
+Inner-root message:
+
+```json
+{
+    "content": <string>,
+    "tags": [["e", "<kind_400_event_id>", "<relay-url-where-it-was-seen>", "root"], "<optional expiration tag>"]
```
staab, line 130:
```suggestion
    "tags": [["e", "<kind_400_event_id>", "<relay-url-where-it-was-seen>", "channel"], "<optional expiration tag>"]
```

It doesn't seem like a good idea to overload the `root` mark. Also, do we need to know the channel id in the wrapped event? The wrapper is addressed to the channel already, which seems like enough.

112.md
```diff
@@ -0,0 +1,248 @@
...
+ 
+Clients use kind 404 to find out which was the precursor channel(s) to this new channel, and should attempt to display all old and new channels in a chain seamlessly to the user as a unified channel history.
+
+Clients SHOULD consider the new channel metadata as authoritative.  
+
+Clients MAY discard previous channel information, since it will no longer be used.
+
+### Wrapped-kind 405: Delegate new owner
+
+To do, this the creator of the group sends out a wrapped-kind 405 to the main group public key.
```
staab, line 179:
```suggestion
To do this the creator of the group sends out a wrapped-kind 405 to the main group public key.
```

112.md
```diff
@@ -0,0 +1,248 @@
...
+Clients SHOULD ignore wrapped-kind 405 messages until they receive a 406 CLAIM message from the new "owner" of the group. 
+
+### Wrapped-kind 406: Accept ownership transfer
+
+To do, this the creator of the group sends out a wrapped-kind 405 to the main group public key.
+
+```json
+{
+    "content": "optional friendly hey, i'm in charge",
+    "tags": [["e", "<event-id-of-405>", "<relay-url>]"]
```
staab, line 197:
```suggestion
    "tags": [["e", "<event-id-of-405>", "<relay-url>"]]
```

112.md
```diff
@@ -0,0 +1,248 @@
...
+    "content": "optional friendly invitation to own the group",
+    "tags": [["e", "<400-event-id-of-group>", "<relay-url>"], ["p", "<public-key-of-delegate>"]]
+}
+```
+Clients SHOULD ignore wrapped-kind 405 messages from anyone but the current "owner" of a group. 
+Clients SHOULD ignore wrapped-kind 405 messages until they receive a 406 CLAIM message from the new "owner" of the group. 
+
+### Wrapped-kind 406: Accept ownership transfer
+
+To do, this the creator of the group sends out a wrapped-kind 405 to the main group public key.
```
staab, line 192:
```suggestion
To do this the creator of the group sends out a wrapped-kind 405 to the main group public key.
```

112.md
```diff
@@ -0,0 +1,248 @@
...
+
+
+Future extensibility
+--------------------
+
+We reserve wrapped-event kinds 407-419 for other wrapped group events, such as delegation, moderation, mute, hide, etc.
+
+Motivation
+----------
+This is the easiest way to allow the use of group events with select group of people, while hiding metadata, and preserving nostr's censorship resistence
```
staab, line 215:
```suggestion

This is the easiest way to allow the use of group events with select group of people, while hiding metadata, and preserving nostr's censorship resistence.
```

112.md
```diff
@@ -0,0 +1,248 @@
...
+
+True forward secrecy relies on servers reliably discarding the intermediate data - in addition to frequent key rotations, or prev-key ratcheting.   A trusted server, for example, can be requested to discard old messages, by adding an optional NIP-40 expiration.
+
+### OPTIONAL forward secrecy extension
+
+The owner of a channel can add a "expiration" tag to the channel creation message.
+
+The owner of a channel who adds a "expiration" tag SHOULD issue a 404 message before the expiration.   
+
+All members of a channel SHOULD also adding a NIP-40 "expiration" tag to all messages sent to the channel using the duration (`expiration - created_at`) of the wrapped-400 create.
```
staab, line 230:
```suggestion
All members of a channel SHOULD also add a NIP-40 "expiration" tag to all messages sent to the channel using the duration (`expiration - created_at`) of the wrapped-400 create.
```

112.md
```diff
@@ -0,0 +1,248 @@
...
+
+Drawbacks
+---------
+Any member of the group can, implicitly, invite new members to the group, since they have the private key.  Any member of a group can "dox" other members by publishing their wrapped messages.
+
+Any member of the group can spam the group, or otherwise DOS the group.
+
+If any single participant of the group chat leaks the shared secret ( whether intentionally or by accident), all the messages can then be decrypted by others until the next 400 and 404 rotation events.
+
+The use of optional frequent forward secrecy rotation events can mitigate these attacks, provided the server is compliant with the expiration times,
```
staab, line 247:
```suggestion
The use of optional frequent forward secrecy rotation events can mitigate these attacks, provided the server is compliant with the expiration times.
```

staab (commented):
Some fairly minor suggestions to try to make things clearer. Other thoughts:

- What about creating two new key pairs? One for group moderators, and one for group members. Posting _about_ the group would be done by keypair A, posting _to_ the group would be done by keypair B. This would make it easier to ignore malicious events.

fyookball:
What's the deal with event deletion in this context?  I think the owner of a channel should be able to delete events.  Why not extend nip-09 so that channel owner pubkey is additionally valid to delete any event in the channel, not just the author of the event.

112.md
earonesty: this is not correct.   the "wrapped kind 4XX" messages are always using the same keypair.   only the kind 1059 is fresh.   i will make this more clear.

112.md
earonesty: such a large group should be filtered on time, primarily.  nobody can look at 1 million events at a time.   no matter what you put in the outer event, it can and will leak metadata.   kinds won't help much, since 99% of the messages are going to be non-admin messages.   encryption won't work, because iv's are random.  

lets suppose you hash a "p" tag, so you can filter by "p" tag.   now everyone, externally, can also know that there are X participants and that Y posts more than anyone else.    It's not much info, but in my experience these seemingly minor metadata leakages can quickly get exploited by experienced cryptographers  (just knowing the difference between top posts and replies can tell you something about the contents, for example)

earonesty:
> What's the deal with event deletion in this context? I think the owner of a channel should be able to delete events. Why not extend nip-09 so that channel owner pubkey is additionally valid to delete any event in the channel, not just the author of the event.

if the author wants to be able to delete an event at the relay, then the author needs to use a "derived key" instead of an ephemeral key.   we can write-up what this means, but consider something very simple:

`ephemeral key = event id * private key`

this then allows the author to delete an event later (at the relay level!)

112.md
earonesty: these tags are inside the gift wrap, so no problem with leaks.

112.md
earonesty: is this overloading?   i thought the "root" mark was the root of a channel?   in this case, it's the event-id of the channel root message.  i think it should be the outer event id.

earonesty:
at what point should we just "go live with this"?  we're ready to deploy this (update code to match requested changes, etc.), but i don't want to waste time doing that if there's no chance that others will use the same protocol.   interoperability is key!

112.md
staab: I see, yeah nip 28 does use root this way. But I do think things would be simpler if we could remove this entirely.

112.md
```diff
@@ -0,0 +1,257 @@
...
+
+Then they generate a new public/private keypair, and uses this to sign new kind 1059 message.
+
+The `content` of that message is a [NIP-44](https://github.com/nostr-protocol/nips/pull/574) encrypted JSON string.
+
+The content is encrypted using the public key of the destination (the chat room public key, or the invitee public key.
+
+All events called "wrapped-kind XX" refer to these kind 1059 gift wrapped events.
+
+For more information on gift-wrapping, see: [NIP-59](https://github.com/nostr-protocol/nips/pull/468)
```
staab, line 43:
```suggestion
## Kind 1059: Encrypted Gift Wrap

All events called "wrapped-kind XX" refer to kind `1059` gift wrapped events, using an ephemeral key for signing, and a known public key for the recipient (a chat room public key, or invitee public key depending on the event kind).

For more information on gift-wrapping, see: [NIP-59](https://github.com/nostr-protocol/nips/pull/468)
```

fyookball:
@earonesty I don't really understand your logic regarding deletion.  Why should the author have to use a derived key or do anything special?  I'm suggesting that the standard practice be that the room creator should have the power to delete events.  If this is something that makes sense, then we don't necessarily need extra logic -- just specify in the nip that the creator of the room pubkey can delete events that are part of the channel.

arthurfranca:
@fyookball The relay does not know that the inner event (gift-wrapped) is a chat message. So you can't edit NIP-09 (event deletion) the way you want.

There are only 2 solutions to this:
1) @earonesty way of deriving each time the gift-wrap key from the shared channel key + inner event id
2) Using my proposal of the `!` tag by [editing NIP-09 slightly](https://github.com/nostr-protocol/nips/pull/539/files), which is generic enough to be useful to other use cases and is very easy for relays to implement

For 2), the gift wrap event now instead of (or in addition to) using the tag `["p", "<shared channel pubkey>"]` should use `["!", "<shared channel pubkey>"]`. This will tell the relay this pubkey can delete the gift wrap event.

fyookball:
@arthurfranca @earonesty i'm fine with the deletion being done client-side. As long as its nip-standardized it will work across any clients who implement.  It doesn't have to be strict, but if it's there, it will improve the user experience.

arthurfranca:
@earonesty taking advantage of your encryption knowledge, is there a way any group member could derive a pubkey from the channel owner's pubkey (+ shared group key? or kind 403 event id? or kind 403 event hash? or channel create event id? or channel create hash) in a way that the channel owner with its privkey could derive the corresponging privkey? Sorry that I have no idea on this.

If it was possible, the kind 403 wrapping events could have:
```
{
  ...,
  tags: [
    ["!", "<channel owner derived pubkey>"], <-- is this possible?
    ...
  ]
}
```
If possible, the ! tag derived value would hide the real channel owner pubkey while allowing the channel owner to delete kind 403 channel messages by signing kind 9 ([updated with this relay rule](https://github.com/nostr-protocol/nips/pull/539/files)) events using corresponding privkey.

Currently the channel owner lacks ability to delete kind 403 messages.

earonesty:
the only way a user can delete a message is if they prove they own the key that produced the message.  this can be done deriving keys

arthurfranca:
Today yes, but look at the ! tag proposal I linked on comment above. That's why I asked if such a keypair could be derived so to be used as ! tag.

fyookball:
maybe there can be a boolean flag whether to allow owner to delete events.  So you could have moderated channels like Telegram and open channels like Reddit without centralized moderation.

rabble:
My reading says that there can be one and only one group owner?

112.md
```diff
@@ -0,0 +1,257 @@
...
+There are no other public tags or metadata, all other kinds and tags are "inner" (inside the wrapped message).  
+
+## Wrapped Kind 401: Invitation to Encrypted Channel
+ 
+Invitations are sent directly to the participants of the channel.
+
+```json
+{
+    "content": "optional personalized invitation message",
+    "tags": [["e", "channel-create-event-id", "relay-url"], ["privkey", "channel-private-key"]]
```
vishalxl, line 81: I am missing what is 'privkey' mentioned here in kind 401? There is 'channel-private-key', which is the shared secret private key used commonly by all to encrypt/decrypt channel messages. But what is 'privkey'?

staab: That's just a literal tag string, `"privkey"`. Square bracket notation should probably be used for non-literals.

vishalxl:
The proposal does not seem to have a 'member's list'. One result of that would be that the members would not 'know', or the clients will not be able to find out, whether the messages that are being received for the group are from invited members, or whether some interloper is also sending messages ( by getting the shared secret somehow). I think for this kind of sanity check, a membership-list is desirable; NIP 38 has such a list, and I find that this kind of list will help people only interact with members who are invited, and will reduce bad-use cases of interlopers using group etc. 

A 'members list' will also improve UI, in that the clients can show who the members of a group are. 

So my suggestion would be that only the admin should send invites, and at that time also update the members list. So when user is added to group ( by admin), then the members list is updated in kind 402, etc.

112.md
vishalxl: Ah yes. I missed that, very basic thing. Am back looking at json after many weeks!..

fyookball:
@vishalxl I don't like the membership-list idea because it's not going to be accurate.  Being in the list is no guarantee those people have even viewed the channel once.  It wouldn't be so bad if it was a one-time thing, but the overhead of continually updating a list with poor accuracy doesn't seem great.

d-krause:
I see the following issues with this NIP Pull Request.

1. **Channels do not need so many new event kinds.**  Fewer kinds make for simpler implementation.

    - Since events are replaceable, one does not need all of _"create channel"_, _"change channel"_, _"move channel"_, _"delegate new owner"_ and _"accept owner delegation"_ kinds.   
    
    - These can all be handled by **ONE** kind with all the required channel metadata properties that verifies all of the properties of these "kinds" as simply properties of the channel metadata itself.  

    - Imagine one channel metadata kind that had properties like :
        ```
        {
           "name": "",
           "about": "",
           "picture": "",
           "owner": "",
           "ownerSig": "",
           "replacedBy": "", 
           "transferTo": "",
           "moderators": []
        }
        ```
    - 

1. **The event kind 403 "_send encrypted chat message_"** is not needed.   An Encrypted chat message should simply be a giftwrapped kind of the kind of event desired.   If it is a text message, then use kind 1, a reaction? then kind 7, an article? then kind 30023.  Let the Gift wrap do the work it is intended to do.   We don't need new kinds to do this.

1. **All events should be deleteable by their creator.**  There is no reason why any replaceable event should not be deleteable.

1. **Each event should use a unique pubkey and separate unique encryption key.**

    - By changing the author pubkey and encryption key on each event, we make it nearly impossible for outsiders to breach the entire channel if any message is compromised.   We also make it exponentially more difficult for outsiders to make targeted eavesdrops on channels without having the channels keys.

    - I had previous pull requests for NIP 76 ([here](https://github.com/nostr-protocol/nips/pull/413) and [here](https://github.com/nostr-protocol/nips/pull/260)) that demonstrated with actual functioning code that this is not only very possible, but very powerful.   While not all parts of NIP 76 may be worth keeping, I believe this aspect of ephemeral pubkeys and encryption keys is worthy of careful consideration.

    - The NIP 76 wording was a bit cumbersome, so please consider it in these simplest of terms:

    - A Private Channel in NIP-76 is created from 2 extended public keys - the signingKey and the encryptKey. An event on the channel chooses the event keys where

        - _authorPubKey_ = `(userPrivateKey + signingKey.ChainCode)/create_date`,
        - _encryptionkey_ = `encryptKey/create_date`.
        - Channel events are identified by a derived #e tag that is solely an unintelligent identifier.

    - An approach like this results in events that cannot be tracked by any third party, and even if one event is compromised, others will not be at risk.

earonesty:
> I see the following issues with this NIP Pull Request.
> 
> 1. **Channels do not need so many new event kinds.**  Fewer kinds make for simpler implementation.

agreed, simpler is better

>    
>    * Since events are replaceable, one does not need all of _"400: create channel"_, _"402: change channel"_, and _"404: move channel"_ kinds.
>    * Kind 400 could simply become "_channel metadata_".  402 & 404 can be handled by **ONE** kind with all the required channel metadata properties that verifies all of the properties of these "kinds" as simply properties of the channel metadata itself.
>    * Imagine one channel metadata kind 400 that had properties like :
>      ```
>      {
>         "name": "",
>         "about": "",
>         "picture": "",
>         "owner": "",
>         "replacedBy": "", 
>         "moderators": []
>      }
>      ```

agree.  ownersig not needed (handled with giftwrap).   not sure camel is consistent with other nips, but i agree.


>    * Do we really need  _"405: delegate new owner"_ and _"406: accept owner delegation" kinds?   Isn't enough to "move" the channel to a new channel using the "replacedBy" attribute (or "movedTo" ) in the channel metadata?  Less moving parts means easier to implement and understand.

yes, we can do that, replaced-by can just point to the new owner's channel

>    * I introduced a "moderators" array above as food for thought only.   This could be a simple list of pubkeys that can make moderator metadata.  This metadata could be a new kind that limits posts (either by poster pubkey, or whatever rules you want to implement.)

moderator is a nice feature.   list of pubkeys

> 2. **The event kind 403 "_send encrypted chat message_"** is not needed.   An Encrypted chat message should simply be a giftwrapped kind of the kind of event desired.   If it is a text message, then use kind 1, a reaction? then kind 7, an article? then kind 30023.  Let the Gift wrap do the work it is intended to do.   We don't need new kinds to do this.

agreed.   kind 1 is fine

> 3. **All events should be deleteable by their creator.**  There is no reason why any replaceable event should not be deleteable.  It is Kind 5 - just gift wrap it and post the event on the channel.   Working Demo here [nostr-nip76.web.app/private-channels](https://nostr-nip76.web.app/private-channels)

this is not possible.   the relay cannot verify deletion without the sender remembering the key.   you can have the client "agree to delete it", but you cannot have the relay delete it unless the sender uses a tweaked "sending key" and then can delete later.   i am implementing this.   sender will not use emphemeral key, an will instead use tweaked (private key * event id)... so we can delete for real.

> 4. **Each event should use a unique pubkey and separate unique encryption key.**

agreed.  that's baked in to the gift-wrap.  no need to change it.

>    * By changing the author pubkey and encryption key on each event, we make it nearly impossible for outsiders to breach the entire channel if any message is compromised.   We also make it exponentially more difficult for outsiders to make targeted eavesdrops on channels without having the channels keys.

again, that's baked into gift wrap, all public keys are unique for each message.

earonesty:
if everyone likes the idea of simplifying by overloading the metadata kind, let me know.   im ok with that if people really like it.

earonesty:
> > * Do we really need  _"405: delegate new owner"_ and _"406: accept owner delegation" kinds?   Isn't enough to "move" the channel to a new channel using the "replacedBy" attribute (or "movedTo" ) in the channel metadata?  Less moving parts means easier to implement and understand.
> 
> yes, we can do that, replaced-by can just point to the new owner's channel

but the new owner needs to say { "replaces": "original channel id" }

that way they are consenting up front

doing it without consent is a recipe for problems.

PR #589, opened 2023-06-08
"c" for content-type

mattn:
Currently nostr has clients that interpret content as Markdown or HTML on their own.

* Snort as Markdown
* Satelite as HTML with sanitizer
* Primal as HTML with sanitizer

Rendering with HTML is wrong since the santizer might not work correctly in all clients. Satelite and Primal drops following content after `<script>` tag.

image

It is different between Snort and others. Snort displays `<g>`  or `&lt;g&gt;` as we can see. (`<g>` OMG)

image

image

So we should add way to specify content-type. Client does not need to guess the content, and it can give up rendering

### DIFF
--- a/01.md
+++ b/01.md
@@ -99,7 +99,7 @@ This NIP defines no rules for how `NOTICE` messages should be sent or treated.
 ## Basic Event Kinds
 
   - `0`: `set_metadata`: the `content` is set to a stringified JSON object `{name: <username>, about: <string>, picture: <url, string>}` describing the user who created the event. A relay may delete past `set_metadata` events once it gets a new one for the same pubkey.
-  - `1`: `text_note`: the `content` is set to the plaintext content of a note (anything the user wants to say). Do not use Markdown! Clients should not have to guess how to interpret content like `[]()`. Set MIME content-type to `"c"` tag if the note should be displayed as Markdown or HTML.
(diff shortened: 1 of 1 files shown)


fiatjaf:
This will incentivize people to use these weird content-types, when they should not, they should use plaintext.

pablof7z:
Concept NACK

staab:
This shouldn't be done on kind 1. I wouldn't necessarily mind seeing it on a new "rich text note" kind that can be optionally supported, although that would also be a soft fork.

mattn:
I understand. What this pull-request meant was not only text notes and rich text required for SNS, but also future use of Event as a mashup without increasing the type of kind.

staab:
The problem is that it's not backwards compatible with client implementations and old events, which don't specify a `c` tag.

mattn:
In any case, clients which already interpret text as Markdown or HTML will have to make incompatible changes to render old content as plaintext. I think that existing renderers can be leveraged if c is given. As you said, events that do not use the c tag must be rendered as plain text, but after this PR, the clients themselves will be possible to check the c tag and select renderers.

mikedilger:
> ... clients which already interpret text as Markdown or HTML ...

We can't let clients that do things outside of the agreed upon NIPs bully the rest of the nostr ecosystem into doing it their way.  It is only the existing clients who are following the NIPs as they are (or were) which must be respected and not caused to break.

> I think that existing renderers can be leveraged if c is given

Don't just think about the positive case. Think about the negative case. I get kind-1 content with a 'c' tag, my client doesn't have code for that tag, my client renders it as plain text, but it looks like shit and takes 20 screens of scrolling for me to get past. That is not acceptable.

All new proposals for nostr should pass this test: Old clients no longer being developed that follow the NIPs as they are (or were!) should not suddenly start having problems when your new feature idea goes lives.

PR #591, opened 2023-06-08
NIP-1317: Secure Communication with Deniability

root-hardenedvault:
The current spec of NIP-04 messaging protocol are signed by the sender/publisher, the protocol itself is undeniable. In order to implement deniable instant messaging based on Nostr, users can generate a new ephemeral key pair as a "temporary identity" specifically for instant messaging, even each time. The key pair used for posting content is the "permanent identity". However, which permanent identity a temporary identity belongs to should only be known by the communicating parties, and it can be deniable if the other party is treacherous. This requires the communicating parties to authenticate the temporary identities they are about to use with each other's permanent identities, but they cannot provide proof to third parties. We proposed to a secure communication with deniability by extending NIP-04.


shawndearmond:
If new temporary identities are created for the secure communication, how will paid relays with npub whitelists allow the messages to be published?

root-hardenedvault:
Unfortunately, you can either avoid using relay servers with npub restrictions, or stick to the original NIP-04 which lacks both forward secrecy and deniability.

earonesty:
> If new temporary identities are created for the secure communication, how will paid relays with npub whitelists allow the messages to be published?

they can use the AUTH protocol or a simple "auth" tag.  npub whitelists are not really a good way to do things.

earonesty:
this is basically the same as NIP-59

root-hardenedvault:
In the draft [NIP-59]( https://github.com/nostr-protocol/nips/blob/b1d042ed9a57cb3cd6bcdc12fbd283d4f8fc6eea/59.md ), the first request is sent from and signed by Alice's permanent identity key, thus is basically undeniable. In order to keep deniability, sending messages from one's permanent identity should be avoided, while mentioning one's permanent identity from a message sent from a temporary identity is totally deniable.

earonesty:
in NIP-59 all messages are sent using an ephemeral public key.   only the recipient can decrypt and prove identity.

root-hardenedvault:
As shown in [the latest draft NIP-59](https://github.com/nostr-protocol/nips/pull/468/files),  the kind 4 message wrapped inside the kind 1059 message seems still signed by the sender's permanent identity, making it hardly deniable.

earonesty:
it's encrypted, so only the recipient can see it.  that's the "wrapped" part.   in your protocol bob can prove the temp id belonged to alice, for example.  it's the same level of protection.

root-hardenedvault:
>  in your protocol bob can prove the temp id belonged to alice

but unlike signature, bob cannot prove it to third party.

earonesty:
he can if he leaks his private key

root-hardenedvault:
> he can if he leaks his private key. signing a message with a shared secret is cool. you can even do it simply. just generate a shared secret, and sign a regular nostr message with (hash(shared-secret)*priv) as the private key. bob can get a DM of this kind at any time, and there's no need to dance back and forth. instead he can just verify the protocol was followed - know who sent it to him, even subscribe to it.... but not prove it without leaking the PK. then you don't even have to change nip04 (44) that much. a client who doesn't know this protocol will see it as an untrusted DM. a client that knows it will see it as a trusted DM.

If Bob leaks his private key, Alice can say that anyone holding Bob's  private key can forge such a message, so how can they prove the temp id belongs to Alice?

earonesty:
bob can prove it belongs to Alice because he has the signed message from
her proving it came from her.

root-hardenedvault:
> bob can prove it belongs to Alice because he has the signed message from her proving it came from her.

MAC is different from signature. Because Bob can also generate the shared secret and MAC the message with the same shared secret as Alice, he cannot prove to others that the message is MAC'ked by Alice instead of himself, which is the basic point of deniability.

earonesty:
ok, i see.  this is equivalent to "signing the message with the shared secret" instead of signing it with the ephemeral key.   that saves a step, gets you a fixed author field which can be useful, and reduces the message size.

earonesty:
>  is only used to exchange temporary identities

ok, so all the messages should be marked as "kind ephemeral", where they get deleted after a certain amount of time.   if i spin up a new client... i've lost alice's temp id, and need to re-establish it.

root-hardenedvault:
> As shown in [the latest draft NIP-59](https://github.com/nostr-protocol/nips/pull/468/files), the kind 4 message wrapped inside the kind 1059 message seems still signed by the sender's permanent identity, making it hardly deniable.

NIP-59 is trying to solve different problems, while NIP-1317 only provides the deniability. The question here is more like whether Nostr community is interested in having deniability as an option in the long-run.

earonesty:
you can easily sign inner message with the shared secret, instead signing with the private key - providing the same deniability, and the same level of proof (could only be from yourself or from the other person)

root-hardenedvault:
> you can easily sign inner message with the shared secret, instead signing with the private key

Cryptographically, how? Unlike the private key, the shared secret is not a scalar, let alone the notation for the recipient to know it is "signed with a shared secret between who and whom".

earonesty:
> > you can easily sign inner message with the shared secret, instead signing with the private key
> 
> Cryptographically, how? Unlike the private key, the shared secret is not a scalar, let alone the notation for the recipient to know it is "signed with a shared secret between who and whom".

you can just use the x coordinate or, to be safe, the hash of the x coordinate.   now you have a secure, shared scalar, and you have the deniability - without the problem of being able to follow the conversation at all.   easy enough to include sender in the envelope.   nostr clients are generally stateless, and preserving this seems important. 

you can modify nip-44   so that the outer message is signed this way, and the encrypted content contatins {a: author: c: content).

now you have deniability and metadata hiding without interactivity

root-hardenedvault:
> ok, i see. this is equivalent to "signing the message with (a hash of) the shared secret" instead of signing it with the ephemeral key. 

You always mix up signature and MAC, in which signature is fundamentally undeniable, while MAC is deniable.

> nostr clients are generally stateless, and preserving this seems important.

That is why forward secrecy has seldom been concerned in Nostr and it doesn't make sense for modern cryptography engineering without forward secrecy.

> then again, i still feel that the gift-wrap (59), which full hides metadata, is enough. no need to do a live/hot exchange of id's. have you looked at it? does it get you what you want?

It's one of reasons why NIP-59 is not enough for us.

earonesty:
I'm not mixing anything up You have to read what I wrote very carefully

there's no fundamental difference between a message authentication and a signature if both use the shared secret as the identifier

The advantage of using a signature is that you don't have to perform multiple authentications to get the same fully deniable result

NalaGinrut:
> there's no fundamental difference between a message authentication and a signature if both use the shared secret as the identifier

Just pass by and interrupt for a moment. ;-)

Message authentication and signature schemes are two distinct cryptographic techniques that serve different purposes.

Message authentication schemes are used to verify the authenticity of a message, ensuring that it has not been tampered with or altered in transit. They typically use a shared secret key between the sender and recipient to generate a MAC that is sent along with the message. The recipient can then use the same shared secret key to verify the MAC and confirm the message's integrity.

On the other hand, signature schemes are used to provide non-repudiation, meaning that the sender cannot deny having sent a message. Signature schemes are typically used in scenarios where the authenticity and integrity of a message need to be guaranteed, and the sender needs to be held accountable for the message's contents.

While both message authentication and signature schemes may use a shared secret as part of their cryptographic operations, they serve different purposes and have distinct security properties. Therefore, it is not accurate to claim that there is no fundamental difference between the two if they use the shared secret as the identifier.

PR #592, opened 2023-06-08
NIP-11: add optional image

suhailsaqan:
Add optional image to Relay Information Document.

### DIFF
--- a/11.md
+++ b/11.md
@@ -255,6 +255,18 @@ Relays that require payments may want to expose their fee schedules.
 }
 ```
 
+### Image ###
+
+A URL pointing to an image to be used to represent the relay.
+
+```json
+{
+...
+  "image": "https://nostr.build/i/53866b44135a27d624e99c6165cabd76ac8f72797209700acb189fce75021f47.jpg",
+...
+}
+```
+
 ### Examples ###
 As of 2 May 2023 the following `curl` command provided these results.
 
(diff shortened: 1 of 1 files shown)


jeffthibault:
LGTM

alexgleason (approved):

alexgleason:
I've been thinking for a while we should have this. cc @vitorpamplona

vitorpamplona:
Amethyst uses `/favicon.ico` on the relay address to get the image. 

But this could also work. It's all about compliance. 

We show the relay icon upfront because we want to give relays traffic. But most relays don't bother to add a logo... :(

fiatjaf (approved):

11.md
```diff
@@ -255,6 +255,18 @@ Relays that require payments may want to expose their fee schedules.
 }
 ```
 
+### Image ###
+
+A URL pointing to an image to be used to represent the relay.
```
alexgleason, line 260: It occurred to me this isn't very specific, which could cause the field to be used in competing ways by different applications. Is the use-case an icon, like I imagine? If so I would call it "icon" and recommend it should be square. That would leave room for us to include fields like "banner" in the future.

vitorpamplona: Agreed.

suhailsaqan: True. Will change now.

alexgleason:
I think this is good to go. 👍

vitorpamplona (approved):

PR #597, opened 2023-06-11
NIP-52: Calendar Events

tyiu:
There are many useful types of applications that can be built on top of this NIP. For example:
- calendar app
- meetup app
- conference app


tyiu:
Amended my commit to mention that recurring calendar events are an unsolved limitation. Somebody reminded me that I didn't tackle it. I'll think about how to add it to the NIP.

52.md
```diff
@@ -0,0 +1,130 @@
...
+  "created_at": <unix timestamp in seconds>,
+  "kind": "31923",
+  "tags": [
+    ["d", "<UUID>"],
+
+    ["summary", "<summary of calendar event>"],
+    ["description", "<description of calendar event>"],
+
+    // Timestamps
+    ["start", "<unix timestamp in seconds>"],
```
tyiu, line 42: In order to support recurring calendar events, I wonder if it would make sense to switch over from unix timestamps to ISO 8601 timestamps instead.

As a reference, iCalendar formats use ISO 8601.
https://en.wikipedia.org/wiki/ICalendar

vitorpamplona:
How do we do private events?

52.md
```diff
@@ -0,0 +1,130 @@
...
+    ["t", "calendar"]
+  ]
+}
+```
+
+## Unsolved Limitations
+
+* No private events
+* No support for time zones
+* No support for recurring events
```
tyiu, line 130: I think there's a case to be made to say recurring events just aren't supported. It adds a lot of complexity around RSVP'ing on individual instances, cancelling individual instances, modifying individual instances as one-offs.

tyiu:
> How do we do private events?
> 
> Secret-based encryption? Build a secret, share with invitees individually and they use it to decrypt.

I feel like there's case to be made for a separate NIP that specifies how to do this with any kind, not just for calendar events. I'm not sure how you would share the secret with invitees.

nolim1t:
looks cool. 

Maybe find a way to include payment for an event too, as nostr events can be zapped too.

And you can verify proof of payment according to NIP57

(replied on nostr as well)

52.md
fiatjaf: yes, we also don't need timezones.

nostronaut:
how would you signal if a `calendar event` got cancelled?

MiningSC:
> How do we do private events?
> 
> Secret-based encryption? Build a secret, share with invitees individually and they use it to decrypt.

Could this be done with NIP-04?  When a private calendar event is created, send a kind 4 event to anyone who is an invitee, the 31923 json data could be included in the content of the kind 4 message. When invitees RSVP, they could respond via a calendar app to the creator, or all invitees with kind 4 messages with the 31924 data being included in the message.  A calendar app could require the user to login to see their private events.  If a user logs in, the app would parse their kind 4 messages for calendar events to be shown, in addition to any other public calendars they are subscribed to.

pablof7z:
> how would you signal if a `calendar event` got cancelled?

Delete kind

nostronaut:
> > how would you signal if a `calendar event` got cancelled?
> 
> Delete kind

that is a bit too implicit

the user of a stateless client searches for a `calendar event` they planned on going and the event is not there anymore -- confused user

a stateful client upon seeing the event being deleted before its due date displays to the user calendar event cancelled

pablof7z:
> looks cool. 
> 
> 
> 
> Maybe find a way to include payment for an event too, as nostr events can be zapped too.
> 
> 
> 

You don't need to do anything special for zaps; any nostr event can be zapped so there's no need to explicitly call that out on each nip

pablof7z:
> > > how would you signal if a `calendar event` got cancelled?
> 
> > 
> 
> > Delete kind
> 
> 
> 
> that is a bit too implicit
> 
> 
> 
> the user of a stateless client searches for a `calendar event` they planned on going and the event is not there anymore -- confused user
> 
> 
> 
> a stateful client upon seeing the event being deleted before its due date displays to the user calendar event cancelled  

All clients should be stateless (other than a cache)

There're no guarantees in nostr and all clients should take this into account; even if you replaced the event itself to signal it was cancelled a client might not see the replaced event.

tyiu:
> > > > how would you signal if a `calendar event` got cancelled?
> > 
> > > 
> > 
> > > Delete kind
> > 
> > 
> > 
> > that is a bit too implicit
> > 
> > 
> > 
> > the user of a stateless client searches for a `calendar event` they planned on going and the event is not there anymore -- confused user
> > 
> > 
> > 
> > a stateful client upon seeing the event being deleted before its due date displays to the user calendar event cancelled  
> 
> All clients should be stateless (other than a cache)
> 
> There're no guarantees in nostr and all clients should take this into account; even if you replaced the event itself to signal it was cancelled a client might not see the replaced event.

How deletions work currently only accept `e` tags which doesn't seem to work well with kind 3xxxx. I think we need to expand the deletion NIP to support `a` tags.

52.md
tyiu: > Recurring events are a shitcoin and you don't need them for your use case, right?
> 
> I think it's good mentioning they are not supported in this version (hopefully never) and roll with it

I don't need them for my conference app use case, no. That's good with me. Don't want to open this can of worms.

52.md
```diff
@@ -0,0 +1,130 @@
...
+{
+  "id": <32-bytes lowercase hex-encoded sha256 of the the serialized event data>,
+  "pubkey": <32-bytes lowercase hex-encoded public key of the event creator>,
+  "created_at": <unix timestamp in seconds>,
+  "kind": "31924",
+  "tags": [
+    ["a", "31923:<calendar event author pubkey>:<d-identifier of calendar event>", "<optional relay url>"],
+    ["d", "<UUID>"]
+
+    ["t", "<yes/no/maybe>"],
```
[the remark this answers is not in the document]
tyiu: I was thinking that some clients may find it useful to filter based on RSVP status. "Give me only the calendar events I'm attending". But maybe that's a niche use case and instead that can be filtered on the client side since clients may want to show all RSVPs, regardless of status. What do you think?

ArcadeCityMayor:
Looks cool, will use it!

> How do we do private events?
> 
> Secret-based encryption? Build a secret, share with invitees individually and they use it to decrypt.

Perhaps [NIP-112: Encrypted Group Events](https://github.com/nostr-protocol/nips/pull/580)

52.md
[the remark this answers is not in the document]
pablof7z: Ah yeah, that's probably a good use case for an index

pablof7z:
> > > > > how would you signal if a `calendar event` got cancelled?
> > > 
> > > > 
> > > 
> > > > Delete kind
> > > 
> > > 
> > > 
> > > that is a bit too implicit
> > > 
> > > 
> > > 
> > > the user of a stateless client searches for a `calendar event` they planned on going and the event is not there anymore -- confused user
> > > 
> > > 
> > > 
> > > a stateful client upon seeing the event being deleted before its due date displays to the user calendar event cancelled  
> > 
> > All clients should be stateless (other than a cache)
> > 
> > There're no guarantees in nostr and all clients should take this into account; even if you replaced the event itself to signal it was cancelled a client might not see the replaced event.
> 
> How deletions work currently only accept `e` tags which doesn't seem to work well with kind 3xxxx. I think we need to expand the deletion NIP to support `a` tags.

There's nothing special about the e vs a kind on the delete event. Highlighter does a tags on delete kinds. Probably habla does a tag deletes on NIP-23s.

tyiu:
> > > > > > how would you signal if a `calendar event` got cancelled?
> > > > 
> > > > 
> > > > > 
> > > > 
> > > > 
> > > > > Delete kind
> > > > 
> > > > 
> > > > that is a bit too implicit
> > > > the user of a stateless client searches for a `calendar event` they planned on going and the event is not there anymore -- confused user
> > > > a stateful client upon seeing the event being deleted before its due date displays to the user calendar event cancelled
> > > 
> > > 
> > > All clients should be stateless (other than a cache)
> > > There're no guarantees in nostr and all clients should take this into account; even if you replaced the event itself to signal it was cancelled a client might not see the replaced event.
> > 
> > 
> > How deletions work currently only accept `e` tags which doesn't seem to work well with kind 3xxxx. I think we need to expand the deletion NIP to support `a` tags.
> 
> There's nothing special about the e vs a kind on the delete event. Highlighter does a tags on delete kinds. Probably habla does a tag deletes on NIP-23s.

I agree, but https://github.com/nostr-protocol/nips/blob/master/09.md as written today requires `e` tags.

>A special event with kind 5, meaning "deletion" is defined as having a list of one or more e tags, each referencing an event the author is requesting to be deleted.
>Each tag entry must contain an "e" event id intended for deletion.

pablof7z:
> > > > > > > how would you signal if a `calendar event` got cancelled?
> 
> > > > > 
> 
> > > > > 
> 
> > > > > > 
> 
> > > > > 
> 
> > > > > 
> 
> > > > > > Delete kind
> 
> > > > > 
> 
> > > > > 
> 
> > > > > that is a bit too implicit
> 
> > > > > the user of a stateless client searches for a `calendar event` they planned on going and the event is not there anymore -- confused user
> 
> > > > > a stateful client upon seeing the event being deleted before its due date displays to the user calendar event cancelled
> 
> > > > 
> 
> > > > 
> 
> > > > All clients should be stateless (other than a cache)
> 
> > > > There're no guarantees in nostr and all clients should take this into account; even if you replaced the event itself to signal it was cancelled a client might not see the replaced event.
> 
> > > 
> 
> > > 
> 
> > > How deletions work currently only accept `e` tags which doesn't seem to work well with kind 3xxxx. I think we need to expand the deletion NIP to support `a` tags.
> 
> > 
> 
> > There's nothing special about the e vs a kind on the delete event. Highlighter does a tags on delete kinds. Probably habla does a tag deletes on NIP-23s.
> 
> 
> 
> I agree, but https://github.com/nostr-protocol/nips/blob/master/09.md as written today requires `e` tags.
> 
> 
> 
> >A special event with kind 5, meaning "deletion" is defined as having a list of one or more e tags, each referencing an event the author is requesting to be deleted.
> 
> >Each tag entry must contain an "e" event id intended for deletion.

I'll update that NIP to also mention a tags, don't let the fact that it's not named deter you from using 🤙

tyiu:
One situation that I've left undefined is how to handle situations where a calendar event substantially changes after an RSVP has been submitted. I'm leaning towards letting this behaviour be undefined, and letting the client determine what to do with that. Clients can observe the sequence of changes based off of the `created_at` timestamp.

tyiu:
> One situation that I've left undefined is how to handle situations where a calendar event substantially changes after an RSVP has been submitted. I'm leaning towards letting this behaviour be undefined, and letting the client determine what to do with that. Clients can observe the sequence of changes based off of the `created_at` timestamp.
> 
> I'll add an amendment to make that explicit.

Actually, should RSVPs use the calendar event `e` tag?

tyiu:
> > One situation that I've left undefined is how to handle situations where a calendar event substantially changes after an RSVP has been submitted. I'm leaning towards letting this behaviour be undefined, and letting the client determine what to do with that. Clients can observe the sequence of changes based off of the `created_at` timestamp.
> > I'll add an amendment to make that explicit.
> 
> Actually, should RSVPs use the calendar event `e` tag, in addition to the `a` tag? So that we can tell what version of the calendar event it was in response to.

Thinking about it more. It can easily get out of hand because the invitee list is encoded in the calendar event. If someone wants to invite someone additional without changing the calendar event, we don't want clients to unnecessarily invalidate the RSVP.

Never mind! Leaving the `e` tag off.

alexgleason:
This NIP would allow bridging ActivityPub calendar events to Nostr. See: https://docs.joinmobilizon.org/contribute/activity_pub/#event_1

At a glance it looks compatible enough.

tyiu:
Addressed all the feedback so far. Please take another look.

52.md
```diff
@@ -0,0 +1,142 @@
...
+Unlike the term `calendar event` specific to this NIP, the term `event` is used broadly in all the NIPs to describe any Nostr event. The distinction is being made here to discern between the two terms.
+
+## Calendar Events
+
+### Format
+
+The format uses a parameterized replaceable event kind `31923` with a list of tags as described below:
+* `d` (required) universally unique identifier (UUID). Generated by the client creating the calendar event.
+* `summary` (required) name of the calendar event
+* `description` (optional) detailed description of the calendar event
```
nostronaut, line 20: should `description` not be in `content`?

52.md
```diff
@@ -0,0 +1,142 @@
...
+    // Reference links
+    ["r", "<url>"],
+    ["r", "<url>"]
+  ]
+}
+```
+
+Calendar events are not required to be part of a [calendar](#calendar).
+
+## Calendar Event RSVP
```
nostronaut, line 66: as this is spec, and since an implementation could implement calendar events and possibly not  RSVP, it would be more accurate to place the latter in another RFC. just my 21 sats.

nolim1t:
> Amended my commit to mention that recurring calendar events are an unsolved limitation. Somebody reminded me that I didn't tackle it. I'll think about how to add it to the NIP.

Probably better not to overcomplicate this.

Also as an immutable platform theres a [lot of things to consider](https://damus.io/note149sha97a6fykrr5t9xg7006a5psfjtmw3uqnhhhre0rqu8hg0meq522kmk)

52.md
```diff
@@ -0,0 +1,142 @@
...
+
+## Intentionally Unsupported Scenarios
+
+### Recurring Calendar Events
+
+Recurring calendar events come with a lot of complexity, making it difficult for software and humans to deal with. This complexity includes time zone differences between invitees, daylight savings, leap years, multiple calendar systems, one-off changes in schedule or other metadata, etc.
+
+This NIP intentionally omits support for recurring calendar events and pushes that complexity up to clients to manually implement if they desire. i.e., individual calendar events with duplicated metadata represent recurring calendar events.
+
+### Time Zones
```
rabble, line 138: This is a mistake. Timezones matter, and we can't simply ignore their existence. What's more, timezones as unfun as they can be, are a solved problem.

rabble:
I love the idea of a "calendar event" event type... And this gets close. But this is all very well understood territory. 

iCal is not a terrible format, it's not very hard to implement, and it works with many systems. There are easy ways to translate ical in to json, it's also an IETF standard RFC-7265. I suggest that we adapt some of these fields to match the ical standard. For example, don't use epoch seconds, instead use a date-time format which is compact and handles issues like timezones cleanly.

https://www.rfc-editor.org/rfc/rfc7265#section-3.6.5

Using RFC7265 would also give us a model for how we could add recurring events.

https://datatracker.ietf.org/doc/rfc7265/

Let's not re-invent the wheel here. If we slightly adapt the NIP to follow IETF standards, it makes interop easy. There are tons of bad standards out there, but for what we need, the calendaring ones are petty easy to use.

52.md
nolim1t: Timestamps should be a nice universal way to solve that without doing too much at the base protocol.

The client can convert that timestamp into whatever the local time is set to

tyiu:
One thing that I realized we haven't addressed is all day events. If someone wanted to add a "New Year's Eve" calendar event, there's no time component to it, only date. This situation isn't able to be handled by a Unix timestamp. Need to think more about how that situation plays into it.

fiatjaf:
iCal has "start time" and "end time" for events.

nolim1t:
> One thing that I realized we haven't addressed is all day events. If someone wanted to add a "New Year's Eve" calendar event, there's no time component to it, only date. This situation isn't able to be handled by a Unix timestamp. Need to think more about how that situation plays into it.

new years eve is different everwhere. The client can easily handle that calculation.

tyiu:
> > One thing that I realized we haven't addressed is all day events. If someone wanted to add a "New Year's Eve" calendar event, there's no time component to it, only date. This situation isn't able to be handled by a Unix timestamp. Need to think more about how that situation plays into it.
> 
> 
> 
> new years eve is different everwhere. The client can easily handle that calculation.

Yes, I agree, but it cannot be represented by a Unix timestamp is what I'm implying.

tyiu:
> I love the idea of a "calendar event" event type... And this gets close. But this is all very well understood territory.
> 
> iCal is not a terrible format, it's not very hard to implement, and it works with many systems. There are easy ways to translate ical in to json, it's also an IETF standard RFC-7265. I suggest that we adapt some of these fields to match the ical standard. For example, don't use epoch seconds, instead use a date-time format which is compact and handles issues like timezones cleanly.
> 
> https://www.rfc-editor.org/rfc/rfc7265#section-3.6.5
> 
> Using RFC7265 would also give us a model for how we could add recurring events.
> 
> https://datatracker.ietf.org/doc/rfc7265/
> 
> Let's not re-invent the wheel here. If we slightly adapt the NIP to follow IETF standards, it makes interop easy. There are tons of bad standards out there, but for what we need, the calendaring ones are petty easy to use.

You bring up interesting points. Aside from easy interoperability (still mostly achievable with my proposal imo), what problems do you foresee by using epoch seconds instead of a date-time format that includes timezones at the protocol layer that clients cannot otherwise handle well enough themselves? Similar question for recurring events.

I brought up one example where an "all-day" calendar event like New Year's Eve could be shared globally and a Unix timestamp would not be able to represent it because each client has its own time zone. You'd need to have a date (without time or time zone) format. Assuming that's something we want to support.

I agree that usually reinventing the wheel for established standards isn't the best choice. But also, it's worthwhile exploring if those standards still make sense today, especially when we're reinventing the web with Nostr where simplicity seems to be a core tenet.

52.md
```diff
@@ -0,0 +1,142 @@
...
+## Calendar Events
+
+### Format
+
+The format uses a parameterized replaceable event kind `31923` with a list of tags as described below:
+* `d` (required) universally unique identifier (UUID). Generated by the client creating the calendar event.
+* `summary` (required) name of the calendar event
+* `description` (optional) detailed description of the calendar event
+* `start` (required) inclusive start Unix timestamp in seconds. Must be less than `end`, if it exists.
+* `end` (optional) exclusive end Unix timestamp in seconds
```
shawndearmond, line 22: Kinda think that `end` should be required. Maybe clients should default it to be equivalent to `start` if there is no other value. 
Also, `start` should probably say "Must be less than or equal to `end`."
When querying relays, it would be convenient if relays only need to index one of the fields.

52.md
vitorpamplona: There is no need for anything extra. 

The UNIX timestamp we use is on GMT. If you want to schedule an appointment, apps need to convert from the UI (say, EST) to GMT and send it. When receiving, the UI parses the event and converts back to the timezone of the receiver. If you are traveling, the UI should have a timezone selector so you can mark the event at 5pm JPN time, for instance. The UI converts from JPN to GMT and sends the event.

vitorpamplona:
Ok, I am going to complicate things a bit, but I think private events are important. 

I suggest a protocol where the client creates a shareable secret and hides that secret in each `p` tag, encrypted by each p-tag's pubkey (NIP-04 DM style). The secret decrypts `.content` which has a stringified version of a tags-list structure (similar to NIP-51 encrypted Lists). 

Clients can then choose to move to `.content` any part of the event and also create Anon `p` tags if users want to also hide the metadata. `start`, `end`, and `d` are required to not be encrypted. 

Detail: I propose adding the secret as a 5th `p` tag term for consistency with other NIPs that use a 4th option for "role" in the event. 

```js
{
  "id": //<32-bytes lowercase hex-encoded SHA-256 of the the serialized event data>,
  "pubkey": //<32-bytes lowercase hex-encoded public key of the event creator>,
  "created_at": //<Unix timestamp in seconds>,
  "kind": "31923",
  "tags": [
    ["d", "<UUID>"],

    // Timestamps
    ["start", "<Unix timestamp in seconds>"],
    ["end", "<Unix timestamp in seconds>"],

    // Participants
    ["p", "<32-bytes hex of a pubkey>", "<optional recommended relay URL>", "host", "secret-encrypted-by-this-pubkey"],
    ["p", "<32-bytes hex of a pubkey>", "<optional recommended relay URL>", "guest", "secret-encrypted-by-this-pubkey"],

    ["t", "<tag>"],
    ["t", "<tag>"],
  ], 
  content: "[
    ["summary", "<summary of calendar event>"],
    ["description", "<description of calendar event>"],
    ["location", "<location>"],
    ["g", "<geohash>"],
    ["r", "<url>"],
    ["r", "<URL>"]
  ]"
}
```

52.md
shawndearmond: Definitely. NIP-23 already defines how `content` in long-form notes is handled (ie. markdown) so clients would already have that worked out.

shawndearmond:
This NIP looks very similar to the one proposed in #498 by @vitorpamplona 

There's a better discussion going on here, and I think it's taking shape a little better than the other one. What do you say @vitorpamplona ? Do you want to close #498 in favor of this one?

52.md
tyiu: Having `end` be optional and exclusive is intentional and inspired by the iCal spec which does the same thing.
https://en.wikipedia.org/wiki/ICalendar#Date_and_time_format

`end` is exclusive so that it doesn't cause scheduling conflicts if another calendar event has `start` at the same time, and also makes it less ambiguous where the end boundary actually is (is it at the beginning of the minute, or is it at the end of the minute). That also means that `start` must be less than `end` if it exists. If `end` is omitted, that means the calendar event is just an instantaneous moment.

52.md
tyiu: Sure. I was following the naming from the iCal spec, but `content` seems more in line with idiomatic Nostr.

52.md
```diff
@@ -0,0 +1,142 @@
...
+This NIP is intentionally not defining who is authorized to attend a calendar event if the user who RSVP'd has not been tagged. It is up to the calendar event creator to determine the semantics.
+
+This NIP is also intentionally not defining what happens if a calendar event changes after an RSVP is submitted.
+
+### Format
+
+The format uses a parameterized replaceable event kind `31924` with a list of tags as described below:
+* `a` (required) reference tag to kind `31923` calendar event being responded to per [NIP-33](33.md)
+* `d` (required) universally unique identifier. Generated by the client creating the calendar event RSVP.
+* `t` (required) attendance status. It can be one of the following values: `yes`, `no`, `maybe`
```
tyiu, line 83: Changing `yes`, `no`, `maybe` to `accepted`, `declined`, `tentative` respectively to follow the iCal spec naming more closely.

tyiu:
> Ok, I am going to complicate things a bit, but I think private events are important.

@vitorpamplona I agree private events are important.

However, can we handle private events separately just to optimize for getting in what we have now (which is getting quite beefy)?

Also, it would be cool if we could generalize the proposal for making any kind of Nostr event private (not just calendar events and RSVPs) when involving multiple people.

I do like the idea of adding roles to participants, which also happens to be part of the iCal spec.

tyiu:
I'm kind of thinking that all-day / multi-day events (e.g., New Year's Eve, Hanukkah) should be its own kind where instead of Unix timestamps, you use ISO8601 dates (without time or time zone), but the rest of the fields and tags should be the same as `31923`. RSVPs would work with this other kind as well. Thoughts?

52.md
tyiu: I swapped the kind numbers between calendars and RSVPs so that calendar events and calendars could be sequentially next to each other in the same NIP-52, and then I split off RSVPs into NIP-53.

52.md
```diff
@@ -0,0 +1,112 @@
...
+
+## Calendar Event RSVP
+
+A calendar event RSVP is a response to a calendar event to indicate a user's attendance intention.
+
+RSVPs can be optionally implemented and are defined in [NIP-53](53.md).
+
+## Unsolved Limitations
+
+* No private events
```
staab, line 98: I think private events can be solved using a generic private group implemented with gift wrap. It would just be a public event, published semi-privately.

staab:
Terry, you should write more NIPs, this one is very good.

> all day events

In my experience, dates are completely different from timestamps, because they are essentially timezone-local. They should be represented using `yyyy-mm-dd` format, and should have a timezone component. This is different enough it should be handled with a separate event kind.

> How do we do private events?

I second #468 and #580. Neither NIP is completely satisfactory yet, so if anyone's interested in that problem please give them a read and suggest changes.

staab:
> I recommend using a specific kind for wrapping for every event kind -- otherwise clients download and decrypt events they have no need for.

Agreed, I've proposed a change that allows wrapped events to optionally publish a `k` tag representing the kind of the wrapped note. But I digress.

fiatjaf:
I think we should do this without private events, after there has been experimentation with public events we can edit the NIP to include private events.

53.md
```diff
@@ -0,0 +1,45 @@
...
+
+The format uses a parameterized replaceable event kind `31925`.
+
+The `.content` of these events is optional and should be a free-form note that adds more context to this calendar event response.
+
+The list of tags are as follows:
+* `a` (required) reference tag to kind `31923` calendar event being responded to per [NIP-33](33.md)
+* `d` (required) universally unique identifier. Generated by the client creating the calendar event RSVP.
+* `t` (required) attendance status. It can be one of the following values: `accepted`, `declined`, `tentative`
+* `t` (optional) availability status. It can be one of the following values: `free`, `busy`
```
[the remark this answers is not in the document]
tyiu: Sounds like you're referring to your NIP-32 proposal here:
https://github.com/nostr-protocol/nips/pull/532

You make a reasonable point. Happy to incorporate labels once it's merged. I really just want a way to index / search by attendance status and availability status. Will your proposal be merged soon?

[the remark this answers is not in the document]
staab: Yep, that's the one. It should be merged very soon, it's implemented in two places already.

vitorpamplona:
> I think we should do this without private events, after there has been experimentation with public events we can edit the NIP to include private events.

Sadly. I agree.

53.md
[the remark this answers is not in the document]
tyiu: So for the RSVP use case, is it something like this?
```json
["L", "status"],
["L", "freebusy"],
["l", "<accepted / declined / tentative>, "status"],
["l", "<free / busy>"]
```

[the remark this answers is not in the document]
staab: Yes, although I would namespace the L tags, something like `nip54.status`

52.md
tyiu: > Maybe this is insane, but what about making both optional? I.e. for draft calendar events that are not yet scheduled. Not sure what other use cases.

I feel like if something like that is desired, it should be its own kind. As far as traditional calendar software goes, I'm not sure if I've seen that sort of use case before.

53.md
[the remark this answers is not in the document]
tyiu: Committed the change for labels.

53.md
```diff
@@ -0,0 +1,49 @@
...
+### Format
+
+The format uses a parameterized replaceable event kind `31925`.
+
+The `.content` of these events is optional and should be a free-form note that adds more context to this calendar event response.
+
+The list of tags are as follows:
+* `a` (required) reference tag to kind `31923` calendar event being responded to per [NIP-33](33.md)
+* `d` (required) universally unique identifier. Generated by the client creating the calendar event RSVP.
+* `L` (required) label namespace of `nip54.status` per [NIP-32](32.md)
```
tyiu, line 28: As of writing, NIP-32 has not yet been merged to master, so at the moment, this links to nothing. Here's the PR:
https://github.com/nostr-protocol/nips/pull/532

fiatjaf:
These two things belong in the same NIP, no need for two NIPs. One NIP can describe multiple event kinds and flows if they are related.

53.md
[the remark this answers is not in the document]
fiatjaf: Why all this namespacing? If the spec defines a hard set of values we don't need namespaces. The event kind is the namespace.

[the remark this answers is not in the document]
tyiu: @staab thoughts?

rabble:
We're planning on implementing events in Nos and we'd love this sorted. 

Screenshot 2023-07-12 at 10 09 42 AM


And i still don't think ignoring timezones is an option. Seriously, calendar specs have been stable since the early 2000's and we should use those standars for date time formats.

vitorpamplona:
Good damn.. that's a good-looking UI.

52.md
tyiu: I reversed this decision. @fiatjaf said the two could be merged into one.

tyiu:
> We're planning on implementing events in Nos and we'd love this sorted.
> 
> Screenshot 2023-07-12 at 10 09 42 AM
> And i still don't think ignoring timezones is an option. Seriously, calendar specs have been stable since the early 2000's and we should use those standars for date time formats.

I'm coming to the same conclusion about time zone. If we added a Timezone ID field (e.g. America/Los_Angeles) to the calendar event kind, would that satisfy your concern?

I'm not in agreement with the ISO8601 format though. Unix timestamp seems much better in almost all ways. I'd love to hear more from you though, @rabble

shawndearmond:
Adding an optional timeline field would be good. It would give the client a suggestion if there is otherwise no timeline support or preference set in the app.

yuvadm:
@tyiu @rabble re: iso8601 vs unix epoch, I think this post sums the discussion up nicely https://dev.to/stripe/how-stripe-designs-for-dates-and-times-in-the-api-3eoh - tl;dr; unless optimizing for human readability, epoch is a fine choice.

As for lack of timezone field, clients can always fallback to presenting in user local time. But calculating "correct" timezone from a geolocation is not a trivial task. Most services that do this are online and paid APIs. Some open source libraries offer this functionality offline, but with mixed results. I'm +1 for having an optional timezone field to resolve that.

fiatjaf (approved):

52.md
tyiu: Ended up adding an optional start and end time zone identifier in 2ce758a5be60d7fbda142192b87b57ce552aee50

Use case: I'm in North America and would like to know the schedule of Nostrasia in Tokyo before I get there in Asia/Tokyo time, but I don't want to need to compute it from the geohash as it's non-trivial and implicit.

tyiu:
> I'd still like to see better support for [all-day events](https://github.com/nostr-protocol/nips/pull/597#issuecomment-1591125802). Requiring granular timestamps doesn't always make sense, since they're not always applicable or known in advance. For example "get oil changed" or "beach day, come whenever" might not have a specific time range, and neither one lasts from midnight to midnight.

@staab Added support for date-based calendar event kind `31922` in c44ea4aa1f02e7f0fdbc460ac2beb8980979250f

tyiu:
I think arguments that can be made for using ISO 8601 for time-based calendar events is that developers can interact with the raw data more easily without machine conversion. Moreover, ISO 8601 is the standard for the RFC 5545 iCalendar spec.

And now that we've created date-based calendar events of kind `31922` and time-based calendar events of kind `31923`, it might just make sense make `start` and `end` be consistently ISO 8601 formatted rather than have one in Unix format and the other in ISO 8601 format.

tyiu:
Downside of ISO 8601: there are timestamps that simply cannot be represented properly in ISO 8601, specifically around daylight savings time, that can be represented using Unix timestamps.

See https://datatracker.ietf.org/doc/html/rfc5545#section-3.3.5

>FORM #3: DATE WITH LOCAL TIME AND TIME ZONE REFERENCE
>
>The date and local time with reference to time zone information is identified by the use the "TZID" property parameter to reference the appropriate time zone definition.  "TZID" is discussed in detail in [Section 3.2.19](https://datatracker.ietf.org/doc/html/rfc5545#section-3.2.19).  For example, the following represents 2:00 A.M. in New York on January 19, 1998:
>
>TZID=America/New_York:19980119T020000
>
>If, based on the definition of the referenced time zone, the local time described occurs more than once (when changing from daylight to standard time), the DATE-TIME value refers to the first occurrence of the referenced time.  Thus, TZID=America/New_York:20071104T013000 indicates November 4, 2007 at 1:30 A.M. EDT (UTC-04:00).

That's pretty messed up and makes me convinced that Unix timestamp representation is far superior.

52.md
```diff
@@ -0,0 +1,212 @@
...
+
+#### Format
+
+The format uses a parameterized replaceable event kind `31923`.
+
+The `.content` of these events is optional and should be a detailed description of the calendar event.
+
+The list of tags are as follows:
+* `d` (required) universally unique identifier (UUID). Generated by the client creating the calendar event.
+* `summary` (required) name of the calendar event
```
nostronaut, line 88: `name` is more intuitive than `summary` for a calendar event name

tyiu: Agreed. Changed `summary` to `name` in 1058d1bc9cf4259d4386629f3169290dc940052f

tyiu:
I squashed the commits on this branch and have rebased to latest master.

At this point, I think the NIP is in a relatively good state. There were some unaddressed questions from me to @rabble about keeping Unix timestamps but with time zone support but most people involved seem to be fine with it. Otherwise, we're just waiting on fulfilling the first acceptance criteria of `They should be implemented in at least two clients and one relay -- when applicable.`

@GBKS for nosta-me and nos.social both seem to be building integrations, and I'm trying to find spare time to implement it in one of my other clients (Confstr and perhaps a generic calendar client).

fiatjaf:
What is the difference between the two types of calendars? I think it is terrible, awful, very bad idea to have two kinds of events that are the same thing, but with different ways of representing dates.

Why not just use Unix timestamps? If the event is not meant to happen at an exact timestamp we just need another tag there saying "full-day event" or "approximated". I don't know, that may feel ugly, but I believe for implementors it makes things much simpler.

vitorpamplona:
@fiatjaf It's actually necessary for calendars. Unix timestamps can't represent full-day events that are independent of the timezone. A New Year's holiday, for instance, cannot be coded with start-end time unless you create a different event for each timezone on the planet and then only invite people in that timezone for each event. Unnecessary complexity.

52.md
```diff
@@ -0,0 +1,212 @@
...
+  "tags": [
+    ["d", "<UUID>"],
+
+    ["name", "<name of calendar event>"],
+
+    // Dates
+    ["start", "<YYYY-MM-DD>"],
+    ["end", "<YYYY-MM-DD>"],
+
+    ["start_tzid", "<IANA Time Zone Database identifier>"],
```
tyiu, line 54: `start_tzid` and `end_tzid` got copy-pasta'd from time-based calendar events into date-based calendar events. I'm going to remove them because time zones do not make sense for date-based calendar events.

tyiu: Removed in 4042ab14cc77c833e56cbf9929e0b767d006de58

tyiu:
> @fiatjaf It's actually necessary for calendars. Unix timestamps can't represent full-day events that are independent of the timezone. A New Year's holiday, for instance, cannot be coded with start-end time unless you create a different event for each timezone on the planet and then only invite people in that timezone for each event. Unnecessary complexity.

Agreed. There are real-world use cases where a time zone agnostic date-based calendar event is necessary. Majority of people around the world operate on these mental models. This was written in the NIP but I'll reiterate it here:
>This kind of calendar event starts on a date and ends before a different date in the future. Its use is appropriate for all-day or multi-day events where time and time zone hold no significance. e.g., anniversary, public holidays, vacation days.

fiatjaf:
OK, I see the point.

But is the difference between the two kinds just that? I think in that case they should be the same kind then. We gain nothing by using two different kinds.

shawndearmond:
It would be really hard to query the relay for calendar events with a date greater than or equal to today if the field might contain Unix timestamp or ISO.

fiatjaf:
You can't do that anyway already.

staab:
Just implemented a subset of this nip [here](https://calendar.coracle.social/). I don't know if it counts as a client implementation, but I wanted to mention it here since you all are interested in this topic. If you want to fork or contribute, be my guest, I won't be able to spend much time maintaining it.

tyiu:
> OK, I see the point. Thank you.
> 
> 
> 
> But is the difference between the two kinds just that? I think in that case they should be the same kind then. We gain nothing by using two different kinds.

The data model could get messy. Basically each implementer would need to check a field to see if it’s a date-based calendar event or a time-based calendar event. Conditional on that field, they would have to deserialize the calendar event in a different way.  We would probably have to make sure to name the start and end time fields differently between the two kinds of calendar events. It could get even messier if there’s a different type of calendar event that we didn’t anticipate (e.g., I know we said we weren’t going to do recurring events, but if that happens to be a horrendous mistake and change our minds, that would be something to consider)

On the other hand, there’s a large overlap of the same tags, so one could make the argument that it’s worthwhile.

But I feel like the former argument in my first paragraph is a compelling enough reason to keep the separate, though I could be convinced otherwise. Curious to hear other opinions.

shawndearmond:
> You can't do that anyway already.

Why do you say that?

GBKS:
As an implementor, I would prefer a single event kind. It's a (IMHO) subtle difference in a couple of tags. Clients already have to do a lot of parsing to piece events together into a coherent presentation (live activities, for example, which you want to treat very differently when they are currently happening). Feel free to ignore my opinion though, not a pro on these things.

fiatjaf:
> > You can't do that anyway already.
> 
> Why do you say that?

Because you can only query for exact tags.

fiatjaf:
> Unix timestamps can't represent full-day events that are independent of the timezone. A New Year's holiday, for instance, cannot be coded with start-end time unless you create a different event for each timezone on the planet and then only invite people in that timezone for each event. Unnecessary complexity.

Revisiting this, I retract what I said. I think these date-based events are the real unnecessary complexity. What is the use case for creating a single "new year" event that happens in the entire world? This doesn't seem to be a real use case to me.

tyiu:
> > Unix timestamps can't represent full-day events that are independent of the timezone. A New Year's holiday, for instance, cannot be coded with start-end time unless you create a different event for each timezone on the planet and then only invite people in that timezone for each event. Unnecessary complexity.
> 
> 
> 
> Revisiting this, I retract what I said. I think these date-based events are the real unnecessary complexity. What is the use case for creating a single "new year" event that happens in the entire world? This doesn't seem to be a real use case to me.
> 
> 
> 
> For that reason I think it is fine to keep it as a separate event kind, and hope no one implements that.

Think about it in the context of a shared company calendar. It’s a company holiday and its employees are in different time zones. It’s not possible to represent accurately as a time-based calendar event unless you create a 24-hour long time-based calendar event per time zone, which is way more infeasible than a single date-based calendar event.

Another example, my birthday is on a specific day but I was born on the other side of the world. I’ve since moved. So now my birthday is represented as a day earlier because time zones. That makes no sense.

There are a ton of use cases.

shawndearmond:
> > > You can't do that anyway already.
> > 
> > 
> > Why do you say that?
> 
> Because you can only query for exact tags.

Seems pretty necessary that relays be queried for future calendar events only. Should we add a bit in this nip for how that subscription would be formatted, and how relays should behave?

PeerRich:
i run cal.com and for the love of god please dont invent a new calendar standard

the existing ones cover everything we need here

also timezones are a b*tch, don't underestimate them

iCal is good

also remember a CalDav interface would make instant compatibility with pretty much most calendars nowadays

how are we planning to subscribe to calendars? or invite people? what is Nostr-Cal solving right now?

i think saving a calendar event as a note is just the tip of the iceberg. not sure if this even needs a new nip vs just a normal note 🤔

figuring out how inviting, editing, moving dates around is where the value is and i dont think this nip gets close to it

tyiu:
> i run cal.com (https://github.com/calcom/cal.com) and for the love of god please dont invent a new calendar standard
> 
> the existing ones cover everything we need here
> 
> also timezones are a b*tch, don't underestimate them
> 
> iCal is good
> 
> also remember a CalDav interface would make instant compatibility with pretty much most calendars nowadays
> 
> how are we planning to subscribe to calendars? or invite people? what is Nostr-Cal solving right now?
> 
> i think saving a calendar event as a note is just the tip of the iceberg. not sure if this even needs a new nip vs just a normal note 🤔
> 
> figuring out how inviting, editing, moving dates around is where the value is and i dont think this nip gets close to it

Can you elaborate? The proposed NIP answers all your questions so I’m not sure what the concern is. Have you read it?

Also, what does it even mean when you say it can just be a normal note? That doesn’t make sense to me.

I understand the appeal of not inventing a new standard. But we are also rebuilding the web with Nostr, so some things that feel natural in one standard might not necessarily fit with how Nostr works.

alexgleason (approved):
Lol can we merge this already? It's excruciating seeing all the panicked comments about destroying the universe by reinventing calendars. It's important to do it right, but at this point if it's still wrong just do a follow-up pull request. Even NIP-01 is still getting changes.

tyiu:
> > > > You can't do that anyway already.
> > > 
> > > 
> > > Why do you say that?
> > 
> > 
> > Because you can only query for exact tags.
> 
> Seems pretty necessary that relays be queried for future calendar events only. Should we add a bit in this nip for how that subscription would be formatted, and how relays should behave?

@shawndearmond Like fiatjaf said, Nostr allows searching for only exact tag matches at the moment. I'm open to exploring other types of subscriptions, but I think we can explore that later on as a general problem space independent of this NIP. For now, a poor man's subscription could be made by filtering on only calendar events that were created or updated after a given timestamp. Further filtering can be done on the client for now if they truly want to show only future calendar events.

tyiu:
@fiatjaf Is this PR mergeable or do you need more concrete examples of client implementations?

alexgleason (approved):

shawndearmond (approved):

richgarner-nsw-gov:
Wanna get this in my (somewhat basic and early-days) client:

https://github.com/grubstarstar/nostr-client

Events and DMs are the only thing keeping me tied to Facebook 🤌.

Do you know any relays using this?

alexgleason:
ablobcatrainbow

PR #600, opened 2023-06-12
NIP-09: Add `a` tag deletion

pablof7z:
Simple change to explicitly mention NIP-33 event deletion

### DIFF
--- a/09.md
+++ b/09.md
@@ -21,7 +21,7 @@ For example:
   "tags": [
     ["e", "dcd59..464a2"],
     ["e", "968c5..ad7a4"],
-    ["a", "968c5..ad7a4"]
+    ["a", "<kind>:<pubkey>:<d-identifier>", "<relay url>"]
   ],
   "content": "these posts were published by accident",
   ...other fields
--- a/09.md
+++ b/09.md
@@ -8,7 +8,7 @@ Event Deletion
 
 A special event with kind `5`, meaning "deletion" is defined as having a list of one or more `e` tags, each referencing an event the author is requesting to be deleted.
 
-Each tag entry must contain an "e" event id intended for deletion.
+Each tag entry must contain an "e" event id and/or NIP-33 `a` tags intended for deletion.
 
 The event's `content` field MAY contain a text note describing the reason for the deletion.
 
@@ -20,7 +20,8 @@ For example:
   "pubkey": <32-bytes hex-encoded public key of the event creator>,
   "tags": [
     ["e", "dcd59..464a2"],
-    ["e", "968c5..ad7a4"]
+    ["e", "968c5..ad7a4"],
+    ["a", "968c5..ad7a4"]
   ],
   "content": "these posts were published by accident",
   ...other fields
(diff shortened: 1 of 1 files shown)


arthurfranca:
Imo it isn't needed. Just fetch the parameterized replaceable event to know its id before sending a deletion event.

pablof7z:
that would just delete the latest event; an `a` tag is to delete the nip-33 itself

fiatjaf:
I think I'll side with @arthurfranca for the first time here.

pablof7z:
imagine deleting an event without `a`:

fetch `"#d": [...], "author"..., "kinds"..`
publish kind:5 with `['e', <latestId>]`
fetch `"#d": [...], "author"..., "kinds"..` + beg to the relay gods that they send you the prior version
publish kind:5 with `['e', <priorId>]`
fetch `"#d": [...], "author"..., "kinds"..` + beg some more
publish kind:5 with `['e', <priorId>]`

this sounds like a nightmare, what if you missed an event? did you just "rebased" the nip-33 to a prior revision?

fiatjaf:
I just heard you saying deletions are a lie on the Miami Bitcoin Conference, @pablof7z.

tyiu:
If you can send a request to replace events with an `a` tag, you should be able to request for deletion of `a` tags. It's the same thing and it should mirror that functionality. I agree with @pablof7z

09.md
```diff
@@ -20,7 +20,8 @@ For example:
   "pubkey": <32-bytes hex-encoded public key of the event creator>,
   "tags": [
     ["e", "dcd59..464a2"],
-    ["e", "968c5..ad7a4"]
+    ["e", "968c5..ad7a4"],
+    ["a", "968c5..ad7a4"]
```
tyiu, line 24: Instead of a hex key as `e` has it, `a` should be in the format of

```
["a", "<kind>:<pubkey>:<d-identifier>", "<relay url>"]
```

fiatjaf:
I agree. This should be good then. The only problem is that it creates an extra requirement for relays and confusion on clients about how to properly delete things.

pablof7z:
precisely, given that it would be weird to store all versions, but you can't assume they don't either; the safe approach would be deleting NIP-33s via `a`.

arthurfranca:
xd I wasn't considering this scenario. Your point is valid.
Just need to remove the  "<relay url>" part that makes no sense as you are already sending de deletion event to a specific relay.

09.md
```diff
@@ -20,7 +20,8 @@ For example:
   "pubkey": <32-bytes hex-encoded public key of the event creator>,
   "tags": [
     ["e", "dcd59..464a2"],
-    ["e", "968c5..ad7a4"]
+    ["e", "968c5..ad7a4"],
+    ["a", "<kind>:<pubkey>:<d-identifier>", "<relay url>"]
```
arthurfranca, line 24:
```suggestion
    ["a", "<kind>:<pubkey>:<d-identifier>"]
```
small fix

Issue #315, opened 2023-03-02
Sensitive content needs to be handled better (NIP-36 & elsewhere)


s3x-jay:
I've been working in the porn industry for roughly 15 years and I'm a "Community Ambassador" at XBiz.net (the leading B2B discussion forum for the adult industry). Based on that experience - I'm worried about how sensitive content is handled in Nostr. NIP-36 as spec'd won't be sufficient and doesn't cover all the places someone might encounter sensitive content. This is important to the adult industry because we in the adult industry do not want our content seen/accessed by people who shouldn't or don't want to see it. 

**#1 NIP-36 needs a defined vocabulary**

NIP-36 is incredibly important, but it can't achieve what it needs to achieve because there's no defined vocabulary for "reason". Not that "reason" should be completely locked down with a strict vocabulary, but 99% of the use cases are predictable and those cases would benefit greatly from a defined vocabulary that can be translated, etc.

Without going into too much detail there have been numerous attempts to implement classification systems over the years and all have failed. Many were culturally biased (e.g. "Unsuitable for persons under 18 years of age" is different in Sweden and Saudi Arabia). Other classification systems were just too complicated for the technology at the time.

To me the classification system that had the most promise was ICRA, but it also failed. You can see it's vocabulary here…
https://web.archive.org/web/20080622002259/http://www.icra.org/vocabulary/

My suggestion would be to start with ICRA as a starting point and adjust it to fit the needs of Nostr.

For end users a Nostr client could just let them choose from the primary categories (which I've tweaked a bit):

    (Casual) Nudity
    Erotica ("soft core")
    Sex ("hard core")
    Violence
    Language
    Other potentially harmful topics

A Nostr client could show sub-categories once the user has chosen a main category - but that would be optional. (Keeping it simple is one of the lessons learned from the failure of ICRA.)

Automated systems would be encouraged to use a more detailed vocabulary (see below). The detailed vocabulary could go into a Nostr client's "preferences" section so users could specify whether certain types of content should 1) always be blocked, 2) never be blocked, or 3) the client should confirm with the user each time on a case-by-case basis.

With the vocabulary below, I'm envisioning an automated system would have a comma delimited list of one or more of the items in the vocabulary.

So for example, if I were posting "Someone in Chicago just faved this photo…" and I had an idea of what was in the photo I would use something like:

    Sex - Penetrative sex acts, Context - Porn - Gay male

[...]

PR #602, opened 2023-06-12
NIP-172: Moderated communities

vitorpamplona:
The goal of this NIP is to create moderator-approved public communities around a topic and provide a simple solution to Reddit clients that might want to come into Nostr. 

Let the best curators win. 

Read it [here](https://github.com/vitorpamplona/nips/blob/moderated-communities/172.md)

### DIFF
--- a/172.md
+++ b/172.md
@@ -80,7 +80,7 @@ The post-approval event MUST include `a` tags of the communities the moderator i
 
 It's recommended that multiple moderators approve posts to avoid deleting them from the community when a moderator is removed from the owner's list. In case the full list of moderators must be rotated, the new moderator set must sign new approvals for posts in the past or the community will restart. The owner can also periodically copy and re-sign of each moderator's approval events to make sure posts don't dissapear with moderators. 
 
-Replaceable events can be submitted for approval in two ways: (i) by their `e` tag if moderators want to approve each individual change to the repleceable event or (ii) by the `a` tag if the moderator trusts the original post author to not modify the event beyond community rules. 
+Post Approvals of replaceable events can be created in three ways: (i) by tagging the replaceable event as an `e` tag if moderators want to approve each individual change to the repleceable event; (ii) by tagging the replaceable event as an `a` tag if the moderator authorizes the replaceable event author to make changes without additional approvals and (iii) by tagging the replaceable event with both its `e` and `a` tag which empowers clients to display the original and updated versions of the event, with appropriate remarks in the UI. Since relays are instructed to delete old versions of a replaceable event, the `.content` of an `e`-approval MUST have the specific version of the event or Clients might not be able to find that version of the content anywhere.
 
 Clients SHOULD evaluate any non-`34550:*` `a` tag as posts to be included in all `34550:*` `a` tags.  
 
--- a/172.md
+++ b/172.md
@@ -59,7 +59,7 @@ Community management clients MAY filter all mentions to a given `kind:34550` eve
 
 # Post Approval by moderators
 
-The post-approval event SHOULD include a stringified `new post request` event inside the `.content` of the approval ([NIP-18-style](18.md)).
+The post-approval event MUST include `a` tags of the communities the moderator is posting into (one or more), the `e` tag of the post and `p` tag of the author of the post (for approval notificaitons). The event SHOULD also include the stringified `post request` event inside the `.content` ([NIP-18-style](18.md)) and a `k` tag with the original post's event kind to allow filtering of approved posts by kind.
 
 ```js
 {
--- a/172.md
+++ b/172.md
@@ -1,0 +1,85 @@
+NIP-172
+=======
+
+Moderated Communities (Reddit Style)
+------------------------------------
+
+`draft` `optional` `author:vitorpamplona`
+
+The goal of this NIP is to create moderator-approved public communities arount a topic. It defines the replaceable event `34550` to define the community and the current list of moderators/administrators. Users that want to post into to the community, simply tag any Nostr event with an `a` tag. Moderators the issue an approval event `34551` that links the community with the new post. 
+
+# Community definition
+
+Kind 34550 should include any field that helps define the community and the set of moderators. 
+
+```js
+{
+  "id": "<32-bytes lowercase hex-encoded SHA-256 of the the serialized event data>",
+  "pubkey": "<32-bytes lowercase hex-encoded public key of the event creator>",
+  "created_at": "<Unix timestamp in seconds>",
+  "kind": "34550",
+  "tags": [
+    ["d", "<community_name>"],
+	["description", "<community_description>"],
+
+    // moderators
+    ["p", "<32-bytes hex of a pubkey>", "<optional recommended relay URL>", "moderator"]
+	["p", "<32-bytes hex of a pubkey>", "<optional recommended relay URL>", "moderator"]
+	["p", "<32-bytes hex of a pubkey>", "<optional recommended relay URL>", "moderator"]
+  ]
+}
+```
+
+# New Post Request
+
+Any Nostr event can be a post request. Clients should simply add the community's `a` tag to be presented for the moderator's approval. 
+
+```js
+{
+  "id": "<32-bytes lowercase hex-encoded SHA-256 of the the serialized event data>",
+  "pubkey": "<32-bytes lowercase hex-encoded public key of the event creator>",
+  "created_at": "<Unix timestamp in seconds>",
+  "kind": "1",
+  "tags": [
+    ["a", "34550:<community event author pubkey>:<d-identifier of the community>", "<optional relay url>"],
+  ], 
+  "content": "<my content>"
+}
+```
+
+Community management clients can filter all mentions to the kind-34550 event and request moderators to approve each submission. The same moderator can remove his/her approval of the post at anytime. 
+
+# Post Approval by moderators
+
+The post approval includes a stringified `new post request` event inside the content's of the approval (NIP-18-style). 
+
+```js
+{
+  "id": "<32-bytes lowercase hex-encoded SHA-256 of the the serialized event data>",
+  "pubkey": "<32-bytes lowercase hex-encoded public key of the event creator>",
+  "created_at": "<Unix timestamp in seconds>",
+  "kind": "34551",
+  "tags": [
+    ["a", "34550:<community event author pubkey>:<d-identifier of the community>", "<optional relay url>"],
+	["e", "<Post Request ID>", "<optional relay url>"],
+	["p", "<Post Request Author ID>", "<optional relay url>"],
+  ], 
+  "content": "{ <New Post Request JSON> }"
+}
+```
+
+It's recommended that multiple moderators approve posts to avoid unapproving them when a given moderator is removed from the owner's list. In case the full list of moderators must be rotated, the new moderator set must sign post approvals for the past post or the community will restart. 
+
+# Displaying
+
+Community clients can display posts that have been approved by at least 1 moderator or by the community owner. 
+
+The following filter displays the approved posts. 
+
+```js
+{
+  "authors": ["<author>", "moderator1", "moderator2", "moderator3", ...],
+  "kinds": 34551,
+  "#a": ["34550:<community event author pubkey>:<d-identifier of the community>"],
+}
+```
--- a/172.md
+++ b/172.md
@@ -21,9 +21,9 @@ Kind 34550 should include any field that helps define the community and the set 
   "tags": [
     ["d", "<community_name>"],
     ["description", "<community_description>"],
-	["image",  "<communityimagen>", "WidthxHeight"]
+    ["image",  "<community_image>", "WidthxHeight"],
 	
-	//.. other tags relevant to defining the community
+    //.. other tags relevant to defining the community
 
     // moderators
     ["p", "<32-bytes hex of a pubkey>", "<optional recommended relay URL>", "moderator"],
--- a/172.md
+++ b/172.md
@@ -8,9 +8,9 @@ Moderated Communities (Reddit Style)
 
 The goal of this NIP is to create moderator-approved public communities around a topic. It defines the replaceable event `kind:34550` to define the community and the current list of moderators/administrators. Users that want to post into the community, simply tag any Nostr event with the community's `a` tag (See [NIP-33](33.md)). Moderators issue an approval event `kind:4550` that links the community with the new post.
 
-# Community definition
+# Community Definition
 
-`Kind:34550` should include any field that helps define the community and the set of moderators.
+`Kind:34550` SHOULD include any field that helps define the community and the set of moderators.
 
 ```js
 {
@@ -19,9 +19,9 @@ The goal of this NIP is to create moderator-approved public communities around a
   "created_at": "<Unix timestamp in seconds>",
   "kind": 34550,
   "tags": [
-    ["d", "<Community_name>"],
-    ["description", "<Community_description>"],
-    ["image", "<Community_image_url>", "<Width>x<Height>"],
+    ["d", "<Community name>"],
+    ["description", "<Community description>"],
+    ["image", "<Community image url>", "<Width>x<Height>"],
 	
     //.. other tags relevant to defining the community
 
@@ -40,7 +40,7 @@ The goal of this NIP is to create moderator-approved public communities around a
 
 # New Post Request
 
-Any Nostr event can be a post request. Clients should simply add the community's `a` tag to be presented for the moderator's approval.
+Any Nostr event can be a post request. Clients MUST add the community's `a` tag to the new post event in order to be presented for the moderator's approval.
 
 ```js
 {
@@ -55,11 +55,11 @@ Any Nostr event can be a post request. Clients should simply add the community's
 }
 ```
 
-Community management clients can filter all mentions to a given `kind:34550` event and request moderators to approve each submission. The same moderator can delete his/her approval of the post at any time using event deletions (See [NIP-09](09.md)).
+Community management clients MAY filter all mentions to a given `kind:34550` event and request moderators to approve each submission. Moderators MAY delete his/her approval of a post at any time using event deletions (See [NIP-09](09.md)).
 
 # Post Approval by moderators
 
-The post-approval event includes a stringified `new post request` event inside the `.content` of the approval ([NIP-18-style](18.md)).
+The post-approval event SHOULD include a stringified `new post request` event inside the `.content` of the approval ([NIP-18-style](18.md)).
 
 ```js
 {
@@ -81,14 +81,16 @@ It's recommended that multiple moderators approve posts to avoid deleting them f
 
 # Displaying
 
-Community clients can display posts that have been approved by at least 1 moderator or by the community owner.
+Community clients SHOULD display posts that have been approved by at least 1 moderator or by the community owner.
 
 The following filter displays the approved posts.
 
 ```js
 {
-  "authors": ["<Author>", "<Moderator1>", "<Moderator2>", "<Moderator3>", ...],
+  "authors": ["<Author pubkey>", "<Moderator1 pubkey>", "<Moderator2 pubkey>", "<Moderator3 pubkey>", ...],
   "kinds": [4550],
   "#a": ["34550:<Community event author pubkey>:<d-identifier of the community>"],
 }
 ```
+
+Clients MAY hide approvals by blocked moderators at the user's request. 
--- a/172.md
+++ b/172.md
@@ -79,6 +79,10 @@ The post-approval event MUST include `a` tags of the communities the moderator i
 
 It's recommended that multiple moderators approve posts to avoid deleting them from the community when a moderator is removed from the owner's list. In case the full list of moderators must be rotated, the new moderator set must sign new approvals for posts in the past or the community will restart. The owner can also periodically copy and re-sign of each moderator's approval events to make sure posts don't dissapear with moderators. 
 
+Replaceable events can be submitted for approval in two ways: (i) by their `e` tag if moderators want to approve each individual change to the repleceable event or (ii) by the `a` tag if the moderator trusts the original post author to not modify the event beyond community rules. 
+
+Clients SHOULD evaluate any non-`34550:*` `a` tag as posts to be included in all `34550:*` `a` tags.  
+
 # Displaying
 
 Community clients SHOULD display posts that have been approved by at least 1 moderator or by the community owner.
--- a/172.md
+++ b/172.md
@@ -10,7 +10,7 @@ The goal of this NIP is to create moderator-approved public communities around a
 
 # Community Definition
 
-`Kind:34550` SHOULD include any field that helps define the community and the set of moderators.
+`Kind:34550` SHOULD include any field that helps define the community and the set of moderators. `relay` tags MAY be used to describe the preferred relay to download requests and approvals. 
 
 ```js
 {
@@ -30,9 +30,10 @@ The goal of this NIP is to create moderator-approved public communities around a
     ["p", "<32-bytes hex of a pubkey2>", "<optional recommended relay URL>", "moderator"],
     ["p", "<32-bytes hex of a pubkey3>", "<optional recommended relay URL>", "moderator"],
 	
-    // relays used by the community 
+    // relays used by the community (w/optional marker)
     ["relay", "<relay hosting author kind 0>", "author"],
-    ["relay", "<relay where to post requests to and fetch approvals from>"], 
+    ["relay", "<relay where to send and receive requests>", "requests"], 
+    ["relay", "<relay where to send and receive approvals>", "approvals"],
     ["relay", "<relay where to post requests to and fetch approvals from>"]
   ]
 }
--- a/172.md
+++ b/172.md
@@ -77,7 +77,7 @@ The post-approval event includes a stringified `new post request` event inside t
 }
 ```
 
-It's recommended that multiple moderators approve posts to avoid unapproving them when a given moderator is removed from the owner's list. In case the full list of moderators must be rotated, the new moderator set must sign new approvals for posts in the past or the community will restart. The owner can also periodically copies of each moderator's approval events. 
+It's recommended that multiple moderators approve posts to avoid deleting them from the community when a moderator is removed from the owner's list. In case the full list of moderators must be rotated, the new moderator set must sign new approvals for posts in the past or the community will restart. The owner can also periodically copy and re-sign of each moderator's approval events to make sure posts don't dissapear with moderators. 
 
 # Displaying
 
@@ -88,7 +88,7 @@ The following filter displays the approved posts.
 ```js
 {
   "authors": ["<author>", "moderator1", "moderator2", "moderator3", ...],
-  "kinds": ["4550"],
+  "kinds": [4550],
   "#a": ["34550:<community event author pubkey>:<d-identifier of the community>"],
 }
 ```
--- a/172.md
+++ b/172.md
@@ -47,7 +47,7 @@ Any Nostr event can be a post request. Clients should simply add the community's
   "id": "<32-bytes lowercase hex-encoded SHA-256 of the the serialized event data>",
   "pubkey": "<32-bytes lowercase hex-encoded public key of the event creator>",
   "created_at": "<Unix timestamp in seconds>",
-  "kind": "1",
+  "kind": 1,
   "tags": [
     ["a", "34550:<community event author pubkey>:<d-identifier of the community>", "<optional relay url>"],
   ], 
--- a/172.md
+++ b/172.md
@@ -30,7 +30,7 @@ Kind 34550 should include any field that helps define the community and the set 
     ["p", "<32-bytes hex of a pubkey>", "<optional recommended relay URL>", "moderator"],
     ["p", "<32-bytes hex of a pubkey>", "<optional recommended relay URL>", "moderator"],
 	
-	// relays used by the community 
+    // relays used by the community 
     ["relay", "<relay hosting author kind 0>", "author"],
     ["relay", "<relay where to post requests to and fetch approvals from>"], 
     ["relay", "<relay where to post requests to and fetch approvals from>"]
--- a/172.md
+++ b/172.md
@@ -55,7 +55,7 @@ Any Nostr event can be a post request. Clients should simply add the community's
 }
 ```
 
-Community management clients can filter all mentions of the `kind:34550` event and request moderators to approve each submission. The same moderator can remove his/her approval of the post at any time.
+Community management clients can filter all mentions to a given `kind:34550` event and request moderators to approve each submission. The same moderator can delete his/her approval of the post at any time using event deletions (See [NIP-09](09.md)).
 
 # Post Approval by moderators
 
--- a/172.md
+++ b/172.md
@@ -4,7 +4,7 @@ NIP-172
 Moderated Communities (Reddit Style)
 ------------------------------------
 
-`draft` `optional` `author:vitorpamplona`
+`draft` `optional` `author:vitorpamplona` `author:arthurfranca`
 
 The goal of this NIP is to create moderator-approved public communities around a topic. It defines the replaceable event `34550` to define the community and the current list of moderators/administrators. Users that want to post into the community, simply tag any Nostr event with an `a` tag. Moderators issue an approval event `4550` that links the community with the new post.
 
--- a/172.md
+++ b/172.md
@@ -17,7 +17,7 @@ Kind 34550 should include any field that helps define the community and the set 
   "id": "<32-bytes lowercase hex-encoded SHA-256 of the the serialized event data>",
   "pubkey": "<32-bytes lowercase hex-encoded public key of the event creator>",
   "created_at": "<Unix timestamp in seconds>",
-  "kind": "34550",
+  "kind": 34550,
   "tags": [
     ["d", "<community_name>"],
     ["description", "<community_description>"],
--- a/172.md
+++ b/172.md
@@ -71,6 +71,7 @@ The post-approval event includes a stringified `new post request` event inside t
     ["a", "34550:<community event author pubkey>:<d-identifier of the community>", "<optional relay url>"],
     ["e", "<Post Request ID>", "<optional relay url>"],
     ["p", "<Post Request Author ID>", "<optional relay url>"],
+    ["k", "<New Post Request kind>"],
   ], 
   "content": "{ <New Post Request JSON> }"
 }
--- a/172.md
+++ b/172.md
@@ -6,7 +6,7 @@ Moderated Communities (Reddit Style)
 
 `draft` `optional` `author:vitorpamplona`
 
-The goal of this NIP is to create moderator-approved public communities around a topic. It defines the replaceable event `34550` to define the community and the current list of moderators/administrators. Users that want to post into the community, simply tag any Nostr event with an `a` tag. Moderators issue an approval event `34551` that links the community with the new post.
+The goal of this NIP is to create moderator-approved public communities around a topic. It defines the replaceable event `34550` to define the community and the current list of moderators/administrators. Users that want to post into the community, simply tag any Nostr event with an `a` tag. Moderators issue an approval event `4550` that links the community with the new post.
 
 # Community definition
 
@@ -66,7 +66,7 @@ The post-approval event includes a stringified `new post request` event inside t
   "id": "<32-bytes lowercase hex-encoded SHA-256 of the the serialized event data>",
   "pubkey": "<32-bytes lowercase hex-encoded public key of the event creator>",
   "created_at": "<Unix timestamp in seconds>",
-  "kind": "34551",
+  "kind": "4550",
   "tags": [
     ["a", "34550:<community event author pubkey>:<d-identifier of the community>", "<optional relay url>"],
     ["e", "<Post Request ID>", "<optional relay url>"],
@@ -77,7 +77,7 @@ The post-approval event includes a stringified `new post request` event inside t
 }
 ```
 
-It's recommended that multiple moderators approve posts to avoid unapproving them when a given moderator is removed from the owner's list. In case the full list of moderators must be rotated, the new moderator set must sign post-approvals for posts in the past or the community will restart.
+It's recommended that multiple moderators approve posts to avoid unapproving them when a given moderator is removed from the owner's list. In case the full list of moderators must be rotated, the new moderator set must sign new approvals for posts in the past or the community will restart. The owner can also periodically copies of each moderator's approval events. 
 
 # Displaying
 
@@ -88,7 +88,7 @@ The following filter displays the approved posts.
 ```js
 {
   "authors": ["<author>", "moderator1", "moderator2", "moderator3", ...],
-  "kinds": 34551,
+  "kinds": ["4550"],
   "#a": ["34550:<community event author pubkey>:<d-identifier of the community>"],
 }
 ```
--- a/172.md
+++ b/172.md
@@ -6,11 +6,11 @@ Moderated Communities (Reddit Style)
 
 `draft` `optional` `author:vitorpamplona`
 
-The goal of this NIP is to create moderator-approved public communities arount a topic. It defines the replaceable event `34550` to define the community and the current list of moderators/administrators. Users that want to post into to the community, simply tag any Nostr event with an `a` tag. Moderators the issue an approval event `34551` that links the community with the new post. 
+The goal of this NIP is to create moderator-approved public communities around a topic. It defines the replaceable event `34550` to define the community and the current list of moderators/administrators. Users that want to post into the community, simply tag any Nostr event with an `a` tag. Moderators issue an approval event `34551` that links the community with the new post.
 
 # Community definition
 
-Kind 34550 should include any field that helps define the community and the set of moderators. 
+Kind 34550 should include any field that helps define the community and the set of moderators.
 
 ```js
 {
@@ -32,7 +32,7 @@ Kind 34550 should include any field that helps define the community and the set 
 
 # New Post Request
 
-Any Nostr event can be a post request. Clients should simply add the community's `a` tag to be presented for the moderator's approval. 
+Any Nostr event can be a post request. Clients should simply add the community's `a` tag to be presented for the moderator's approval.
 
 ```js
 {
@@ -47,11 +47,11 @@ Any Nostr event can be a post request. Clients should simply add the community's
 }
 ```
 
-Community management clients can filter all mentions to the kind-34550 event and request moderators to approve each submission. The same moderator can remove his/her approval of the post at anytime. 
+Community management clients can filter all mentions of the kind-34550 event and request moderators to approve each submission. The same moderator can remove his/her approval of the post at any time.
 
 # Post Approval by moderators
 
-The post approval includes a stringified `new post request` event inside the content's of the approval (NIP-18-style). 
+The post-approval event includes a stringified `new post request` event inside the content of the approval (NIP-18-style).
 
 ```js
 {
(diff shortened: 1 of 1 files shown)


vitorpamplona:
hum... this can be used by anyone to do content curation by topic. hah.. interesting... 

A moderator can add any existing nostr event into the topic (there is no need for the post request).

olheadinternet:
> A moderator can add any existing nostr event into the topic (there is no need for the post request). 

what if this was done through nip 51 lists... so any client that has "add to list" can be a place for curating... even if it doesn't format the lists as groups?

olheadinternet:
Is/can poster submit to just <d-identifier of the community> (/bitcoin) 

or poster is separately submitting to /vitor/bitcoin & /bob/bitcoin ?

Contrary to my last comment: Or both, what if I want to avoid /bob/bitcoin by submitting just to /vitor/bitcoin. I think it's a very different thing than Reddit like behavior if it doesn't need a submit sig.

starbackr-dev:
Did you get a chance to look at NIP-29 #566 ? Similar concept but we moved some authority functions to the relay. I am working on a demo now and should be ready by end of this week. You will not be able to achieve full moderation from client side. Control on the relay side actually makes is easier like centralized systems. However, you are not stuck to one relay, the group admins can move them to a different relay anytime with history and group members or even have it on 2 relays at the same time. Best of both worlds...!

Let me know your thoughts.

vivganes:
Thanks a lot for starting this thread 🙏 I just took a quick glance at the proposal.

Not sure if someone already brought this point up.

There is value in the "communities" aspect irrespective of whether it is moderated or not. 

There may be a team managing the community but they may not actively moderate or approve posts too. 

We could think about a way to make moderation optional too 

Will look at the proposal fully and add comments today.

fiatjaf:
This is very much like NIP-28, but for general things instead of just chat.

vitorpamplona:
> This is very much like NIP-28, but for generic things instead of just chat.

Kind of. The difference is the moderator list (28 doesn't have it) and the fact that is positive (every post is reviewed) as opposed to negative (hide all by a user). So, instead of clients having to download everything and then filter out the hidden messages/users (which is a bigger set than actual posts these days for some chats), clients can just download the approved content with a direct filter.

arthurfranca:
Very good! I see the moderator list has optional recommended relays.

I imagine it would be easier for client logic (both user and moderator client) if this is just for profile metadata (kind 0) so to not rely on NIP-65 config from multiple people (author and mods) to know where to post "New Post Requests" and fetch "Post Approval by moderators".

What do you think of adding this:
```js
{
  kind": "34550",
  ...,
  tags: [
    ..., // d, description and moderators
    // "relay" tag is used on NIP-42
    ["relay", "<relay hosting author kind 0>", "author"],
    ["relay", "<relay where to post requests to and fetch approvals from>"], // not "r" so to not index
    ["relay", "<relay where to post requests to and fetch approvals from>"]
  ]
}
```

or maybe better this:

```js
{
  kind": "34550",
  ...,
  tags: [
    ..., // d, description and moderators
    ["author_relay", "<relay hosting author kind 0>"],
    // "relays" tag is used on NIP-57
    ["relays", "<relay 1 where to post requests to and fetch approvals from>",  "<relay 2 where to post requests to and fetch approvals from>"] 
  ]
}
```

vitorpamplona:
So, similar to the issue of clients not being able to request only non-reply kind-1s, right? I agree it's a pain. 

I am not sure if wrapping is a good solution for that. Otherwise, we might get to a point where everything is wrapped. 

Somebody could duplicate kind 1 (like nip-28 channel messages) if people wanted a kind that does not show up in the timeline.

172.md
```diff
@@ -0,0 +1,93 @@
...
+{
+  "id": "<32-bytes lowercase hex-encoded SHA-256 of the the serialized event data>",
+  "pubkey": "<32-bytes lowercase hex-encoded public key of the event creator>",
+  "created_at": "<Unix timestamp in seconds>",
+  "kind": "34551",
+  "tags": [
+    ["a", "34550:<community event author pubkey>:<d-identifier of the community>", "<optional relay url>"],
+    ["e", "<Post Request ID>", "<optional relay url>"],
+    ["p", "<Post Request Author ID>", "<optional relay url>"],
+  ],
```
arthurfranca, line 74:
```suggestion
    ["k", "<New Post Request kind>"],
  ], 
```
Suggestion so community clients may fetch only currently supported kinds

staab:
We shouldn't push core changes in an ad-hoc way. Adoption of that syntax may take months to get implemented (cf COUNT, AUTH). Can we solve this problem using the tools we already have?

> we might get to a point where everything is wrapped.

I'm ok with this. But we won't wrap everything, because non-wrapped notes are already the standard, and imply "public, broadcast". Wrapping makes sense for relay-agnostic sub-networks: public group post; private group post, public DM, private DM. Wrapping could even remove the need for many existing kinds, and reduce the number of primitives clients have to implement. I.e., people are already zapping and quoting DMs, what if DMs, kind 42s, and kind 1s worked exactly the same way and could be cross published between scopes (private, group, public)? Clients would only have to special case the user's ability to decrypt a given message, and you would automatically have much richer content available in non-kind 1 contexts.

Maybe this is not the place to hash out wrapping, but the more I think about it, the more it makes sense.

jinglescode:
This is awesome, thanks for starting on this, I just started this app a few days ago, and now going into `communities`. 

Screenshot 2023-06-14 at 10 32 29 AM

I could prototype this concept and lets see how the moderation process feels like.

pj8912:
`Moderated communities`? We can can create community, add mods, and others can join? these are basic features right?? 
Anybody can upload a post to any community, whether they joined or not right? 
Does every post had to apporved by the mods and then only it will be visible to the public?

nikolalukovic:
👍🏻 for the nip to create communities
👎🏻 for the classical implementation of moderation

This style of moderation is one the biggest pain points of Reddit. It gives too much power to individuals who then get to fulfil their god complexes. Not to mention it requires manual approval to post. I'd rather people reference communities (or maybe limit to one referenced community) in their notes so clients could create "subreddits" by simple grouping of notes per referenced communities.

As for moderation I truly believe in community self moderating itself. Adding actions like "relevant" or "not relevant" which would essentially be an upvote and downvote would be more than enough. Users would be able to set up in their clients on how many "not relevant" actions to automatically hide posts, etc...

☝🏻 These are just some broad thoughts about this NIP but have essential corner stones in which I firmly believe. There's plenty of room for the minutiae.

vitorpamplona:
The difference is that in Nostr, if you want, you can just copy all posts of a community into a new one. The easiness of cloning a community keeps moderators in check.

pj8912:
why my comment got deleted?

vitorpamplona:
You can just use hash tags if you don't want moderation. This is for those who do.

pj8912:
what? So we can't ask anything here?. I had a bunch of doubts and my comment simply got deleted. Here in this thread. Not sure what you are saying.

pj8912:
@nikolalukovic was my comment there when you replied?

vitorpamplona:
Not sure what happen with @pj8912 but here are the answers

> We can can create community, add mods, and others can join? these are basic features right??

Correct. But it's always a public community. There is no join besides following it. 

> Anybody can upload a post to any community, whether they joined or not right?

Correct. But mods must approve. 

> Does every post had to apporved by the mods and then only it will be visible to the public?

Yep, every post has to be approved.

staab:
> why my comment got deleted?

Your comment was unproductive, please put more effort into your questions. For general questions, the nostr telegram channel is a better place to start. Discussions here need to be focused.

vivganes:
> The goal of this NIP is to create moderator-approved public communities around a topic and provide a simple solution to Reddit clients that might want to come into Nostr.
> 
> Let the best curators win.
> 
> Read it [here](https://github.com/nostr-protocol/nips/blob/22c1af86227ea09c15fbcc6d34bf14711847e4e4/172.md)

@vitorpamplona Just a request to help future readers, can you please edit the link in the original PR description to point to https://github.com/nostr-protocol/nips/pull/602/files ? I followed the link given there and didn't realize I was looking at an out-of-date version until I went through a few comments.

vivganes:
Thanks for putting this proposal.  Here are a few questions that I got.

* Are we proposing the 'Join community' event to be a client side information with no data stored on relays?  If yes, then we may not be able to show a community's follower count anywhere.  Follower count is a very useful metric for anyone who wants to post a 'call for help' or a 'question' in a community.

*  How do we handle if two people create a community with the same name?  [May be this is not a bug, but a potential feature]  The primary challenge could be that the users will be confused when they search for a 'bitcoin' community and 100 results show up.  May be we could come up with a notation like `ncom1` for community similar to `npub1` and `nnote1`?

* Given that everything is optional in nostr, if clients dont filter out the kind 1 events with a tags, we may end up in a situation where moderation is just a namesake thing. [May be another feature, not a bug]. Challenge is that - this may not be very intuitive to a normal person, who comes from a reddit-like heavily moderated ecosystem.

* Do we need to add a 'reason' field where anyone who unapproves a note can key-in the reason? Need not be free-form but something like what's mentioned in https://github.com/nostr-protocol/nips/issues/315 . This would keep the moderators accountable in some way.

* How do we add moderators to an already existing community?  Who validates if the event for that is sent by an owner or an imposter?  May be relays? or the clients will decide that too?

Sorry that I am posting many questions rather than suggestions.  I dont know enough information to even attempt to answer some of my questions above.

vitorpamplona:
> Are we proposing the 'Join community' event to be a client side information with no data stored on relays? 

Following communities could be a NIP-51 list. But it is not addressed in this NIP. You would then count the number of lists out there with a given community in it. I think other methods (like a new event kind) should also be considered. 

> How do we handle if two people create a community with the same name? 

There is no way to create unique names in Nostr, so we will have to rely on other data to find the best community. Likely being a community one of your follows is also following. 

> if clients dont filter out the kind 1 events with a tags

Clients should be able to choose to display or not those events. Depending on the UX model of the client, it might work to display the kind 1 with a UI hint that this was a community post and a link to go over there. This also helps communities grow and it would be how Amethyst would display in the feed.

It would be similar to how clients deal with replies today. Generally, there is a distinction between new threads and replies. This would create: New Threads, replies, and all community posts by follows. 

> Do we need to add a 'reason' field where anyone who unapproves a note can key-in the reason? 

There is no "not approved" status. There is just "not addressed". I will assume the majority of the post requests of good communities will be spam (since spam is so easy to create on nostr). I don't think rejecting lots of things is a good model. Community management clients will need to help filter out the spam for moderators. 

> How do we add moderators to an already existing community? 

The owner of the community updates the original event in the same way we update blog posts today.

arthurfranca:
> Following communities could be a NIP-51 list. But it is not addressed in this NIP. You would then count the number of lists out there with a given community in it. I think other methods (like a new event kind) should also be considered.

Hehe when I read it I was already writing a NIP for this. Please take a look at [Member List PR](https://github.com/nostr-protocol/nips/pull/607). It uses user status events instead of NIP-51 lists.

172.md
```diff
@@ -0,0 +1,94 @@
...
+# Post Approval by moderators
+
+The post-approval event includes a stringified `new post request` event inside the `.content` of the approval (NIP-18-style).
+
+```js
+{
+  "id": "<32-bytes lowercase hex-encoded SHA-256 of the the serialized event data>",
+  "pubkey": "<32-bytes lowercase hex-encoded public key of the event creator>",
+  "created_at": "<Unix timestamp in seconds>",
+  "kind": "34551",
```
arthurfranca, line 69: This parameterized replaceable event is missing a "d" tag. Considering the user can be a moderator on many communities and may approve the same event id for more than one group, i think it would be this giant string: `["d", "34550:<community event author pubkey>:<d-identifier of the community>:<Post Request ID>"]`.

Maybe it is better as a regular event in the `1000 <= n < 10000` range like `4550` or  `4551` with no "d" tag. I don't know.

172.md
```diff
@@ -0,0 +1,94 @@
...
+# Displaying
+
+Community clients can display posts that have been approved by at least 1 moderator or by the community owner.
+
+The following filter displays the approved posts.
+
+```js
+{
+  "authors": ["<author>", "moderator1", "moderator2", "moderator3", ...],
+  "kinds": 34551,
```
arthurfranca, line 91: This should be an array. Maybe the event kind should be a different number (see comment above)

172.md
```diff
@@ -0,0 +1,94 @@
...
+# New Post Request
+
+Any Nostr event can be a post request. Clients should simply add the community's `a` tag to be presented for the moderator's approval.
+
+```js
+{
+  "id": "<32-bytes lowercase hex-encoded SHA-256 of the the serialized event data>",
+  "pubkey": "<32-bytes lowercase hex-encoded public key of the event creator>",
+  "created_at": "<Unix timestamp in seconds>",
+  "kind": "1",
```
arthurfranca, line 50:
```suggestion
  "kind": 1,
```

172.md
```diff
@@ -0,0 +1,94 @@
...
+# Community definition
+
+Kind 34550 should include any field that helps define the community and the set of moderators.
+
+```js
+{
+  "id": "<32-bytes lowercase hex-encoded SHA-256 of the the serialized event data>",
+  "pubkey": "<32-bytes lowercase hex-encoded public key of the event creator>",
+  "created_at": "<Unix timestamp in seconds>",
+  "kind": "34550",
```
arthurfranca, line 20:
```suggestion
  "kind": 34550,
```

172.md
```diff
@@ -0,0 +1,94 @@
...
+  "tags": [
+    ["d", "<community_name>"],
+    ["description", "<community_description>"],
+	["image",  "<communityimagen>", "WidthxHeight"]
+	
+	//.. other tags relevant to defining the community
```
arthurfranca, line 26:
```suggestion
    ["image",  "<community_image>", "WidthxHeight"],
	
    //.. other tags relevant to defining the community
```

172.md
```diff
@@ -0,0 +1,94 @@
...
+
+    // moderators
+    ["p", "<32-bytes hex of a pubkey>", "<optional recommended relay URL>", "moderator"],
+    ["p", "<32-bytes hex of a pubkey>", "<optional recommended relay URL>", "moderator"],
+    ["p", "<32-bytes hex of a pubkey>", "<optional recommended relay URL>", "moderator"],
+	
+	// relays used by the community
```
arthurfranca, line 33:
```suggestion
    // relays used by the community 
```
Just identention fix

arthurfranca (changes requested):

172.md
vitorpamplona: Good point. I think we can move it to a regular event with event deletion to revert things.

172.md
vitorpamplona: Fixed

vitorpamplona:
Folks, I moved the post approval event from a replaceable kind to a non-replaceable kind since it doesn't really need to change over time.

olheadinternet:
> You can just use hash tags if you don't want moderation. This is for those who do.

a lot of groups on many platforms work in that moderation happens after the fact. curated topic feeds doesn't equal moderated communities... curated topic feeds is not a bad idea but it's not a community. Specially if you can just pull in any kind 1 content. If i don't sign the content to be in a community, it shouldn't appear as if I'm a member of a community.

vitorpamplona:
No problem, we can rename the NIP to Curated Topics if needed. 

I used the word community because anyone can reply to topics selected by the community. The replies make the community (9Gag style). In other words, there is no need to formally enter a select group of people to create a sense of community.

olheadinternet:
I'm not suggesting there should be a gate, but there should be intentional participation in either posting to bitcoin or posting to bob/bitcoin... ideally i would want both, and im not sure why bob couldn't choose to use the previous popular bitcoin d-identifier with his pubkey, and it sets out that way, instead of retroactively cloning via the ability to just post any kind 1 content found in vitor/bitcoin. when vitor creates the d-tag for bitcoin it could be understood others can pull from this too.

vitorpamplona:
In the way this is set up, anyone can clone any community by resigning all posts of the other one. There could be manual or automated merging with new information or filtering of communities further. For instance, If there is a `/bob/bitcoin` and a `/alice/eth`, I can have a community that automatically filters all bob's bitcoin posts that are not in alice's community. Or I can have one that merges the two completely. 

Again, the best curator wins.

pj8912:
Si this is a basically a [dev.to](dev.to) follow tags feature. This just as same as dev.to tags not reddit. 
### Tags:
image
### Tag Page with moderators
 
image

vitorpamplona:
Never heard about dev.to, so I don't know what's the difference. All the successful Reddit communities are the same: public feeds that anyone can write into (but get moderated in time). The largest ones either have very active moderators taking down spam in real time (you see posts blinking and disappearing) or they approve each post before making it available to everyone else. 

That's where the inspiration for this NIP came from.

olheadinternet:
> That's cool @vitorpamplona , the pictures clearly describe the NIP. This NIP is exactly the dev.to follow tags feature without a doubt. The mods on don't approve every post, approval means you can see the tick sign which most post don't have and still they are visible to public and can interact with it . Mods do remove spam and other posts that violate the community rules quickly.

this is only
> or they approve each post before making it available to everyone else.

i think curated feeds is a clearer way to describe what this accomplishes.

172.md
```diff
@@ -0,0 +1,100 @@
...
+
+    // moderators
+    ["p", "<32-bytes hex of a pubkey1>", "<optional recommended relay URL>", "moderator"],
+    ["p", "<32-bytes hex of a pubkey2>", "<optional recommended relay URL>", "moderator"],
+    ["p", "<32-bytes hex of a pubkey3>", "<optional recommended relay URL>", "moderator"],
+	
+    // relays used by the community 
+    ["relay", "<relay hosting author kind 0>", "author"],
+    ["relay", "<relay where to post requests to and fetch approvals from>"], 
+    ["relay", "<relay where to post requests to and fetch approvals from>"]
```
arthurfranca, line 36:
```suggestion
    ["relay", "<relay where to post requests to>", "read"], 
    ["relay", "<relay where to fetch approvals from>", "write"],
    ["relay", "<relay where to post requests to and fetch approvals from>"]
```
Regarding relays to post requests to and fetch approvals from (not "author" mark), I was thinking it would be good to have optional "write" and "read" marks just like NIP-65 (and no mark meaning both read and write).

For example, it may be useful to focus reading post requests just from one "read" relay while the moderators can "write" the approvals to many relays.

vivganes:
> No problem, we can rename the NIP to Curated Topics if needed.

This is not necessary. We can leave it to the client (may be via user preferences) if unapproved notes still need to be shown (may be with a different visual treatment than the approved notes).

This way, other non-community clients can play well with the moderation aspect, no matter what.

In my view, a cleaner option would be to add an `approvedUntilRevoked`  in the community's create/modify event.  If it is true, then clients can show the posts as approved, unless there is an explicit disapproval (may be we need a new type for this).

This way, the user can choose if he uses the community as a curated topics or unmoderated IRC equivalent 😄

vivganes:
> There is no way to create unique names in Nostr, so we will have to rely on other data to find the best community. Likely being a community one of your follows is also following.

I still believe there is a need to avoid confusion like a unique id  (in addition to non-unique name) for a community.  This allows people (and clients, bots, etc) to do actions on a community. 

This would help the users handle conversations like "Can you give me a link to the community you spoke about?"

This community unique id (hex) could be generated pretty much like we create note ids now.

172.md
```diff
@@ -0,0 +1,100 @@
...
+Any Nostr event can be a post request. Clients MUST add the community's `a` tag to the new post event in order to be presented for the moderator's approval.
+
+```js
+{
+  "id": "<32-bytes lowercase hex-encoded SHA-256 of the the serialized event data>",
+  "pubkey": "<32-bytes lowercase hex-encoded public key of the event creator>",
+  "created_at": "<Unix timestamp in seconds>",
+  "kind": 1,
+  "tags": [
+    ["a", "34550:<Community event author pubkey>:<d-identifier of the community>", "<Optional relay url>"],
```
vivganes, line 52: When modifying a community, the author of that event could be different from the author of the community creation event.  

So, this kind of 'a' tag may require us to hunt for a possibly deleted original 'community creation' event to fetch all  notes in a community?  

May be I am missing something here. Sorry if this question was silly.

1player:
I like this, and I like that there is no concept of upvoting/downvoting in this spec: it should be possible to create communities without concept of "karma", and despite what many think, a karma system is not really self-moderating, but in fact creates a very narrow type of community, one where popularity and being uncontroversial is promoted, i.e. an echo-chamber.

This spec should remain neutral to allow any kind of community, the very strictly moderated ones (say, Hacker News), the loosely moderated (Reddit), or the "self-moderating ones" (using their own karma/upvote system specified elsewhere)

Like the author mentions, the beauty of this spec is that you can easily clone all the posts with a new set of moderators, if you don't like the original ones. This also allows the creation of different "views" of the same community, where the same posts can be cloned and moderated by a different set of people.

172.md
vitorpamplona: No, the author cannot be different. Replaceable events can only be replaced by the original author.

sant0s12:
> In my view, a cleaner option would be to add an `approvedUntilRevoked` in the community's create/modify event. If it is true, then clients can show the posts as approved, unless there is an explicit disapproval (may be we need a new type for this).

Yes, I also think this approach makes a lot of sense because it allows for more flexibility in moderation. Afaik some subreddits on Reddit require approval for posting while others will take down your post after the fact.

IngwiePhoenix:
I am not a native english speaker; chances are I overlooked or mis-read something.

So, this is just to clarify:

- Each Community - or, curated topic - must have new posts approved - and, ALL posts - by a moderator?
- Since Nostr has no such things as a name, just an ID, we'd use a new scheme like `ncom1` to find them (which could potentially be embedded in QR codes et cetera)?

After that, here's some things that stuck out to me while reading all the comments here.

- While "moderation god complex" is an unfortunate side effect of giving someone power (which reminds me a lot of Megadeath's "Symphony of destruction" first few verses), it also feels like a neccessary evil. Leaving a community unmoderated can often lead to chaos. Moderating it can give some people too much power. A community is built on trust and expectations. So: Moderators should be allowed to be reported. I noticed that in Snort, I can even see all reports sent from other people, referencing the very events that offended them. This method of public reporting could be used, to inform the community creator/"admin", that certain action should be taken.
- Why is every post approval-first? Many communities, like the meme-focused ones - thrive on fast posting; which becomes very difficult if each and every post has to be verified. A community, or curated list, should have an option to declare the posting schema: approve-first (each post needs approval), approve-after (if a post is deemed unfitting or otherwise to a community, it may immediately be dropped per delete event), freeform (everyone can post and those are seen as approved and are handled in post by up- or downvotes).
- And, in terms of searchability: NIP-05. I am surprised this hadn't been brought up yet. But communities could be made much more searchable by using NIP-05, indicating the authenticity of a certain community as a whole.

As for a little side-tangent: Nostr as a Reddit alternative ought to work out better than Lemmy. I tried it, even run my own instance, but the super seemless nature of Nostr is just much easier to use. Make a keypair, and off you go. Pick a few relays that sound nice - or that advertise topics you want - and start posting. No crazy instance finding and cross-instance configuration. I am all for this NIP PR! =)

olheadinternet:
After reading a number of replies, I want to double down on renaming this curated topic. There is concern about needing to report moderators... but here curators aren't moderators they don't remove stuff or people from a community.

I like how this works, i also want to see gated but then open communities, that will more effectively emulate a facebook group or subreddit... but I also think this nip and curated topics is a far better solution to topics than hashtags.

172.md
arthurfranca: @vitorpamplona what do you think?

vitorpamplona: `read/write` is very confusing because it depends on the point of view (moderator in this case). I wish we had better nomenclature to make it crystal clear. 

What about

```js
    ["relay", "<relay where to send and receive requests>", "requests"], 
    ["relay", "<relay where to send and receive approvals>", "approvals"],
    ["relay", "<relay where to post requests to and fetch approvals from>"]
```

It's not consistent with NIP65, but I think we need to start moving away from `read`/`write`

staab: Even NIP 65 isn't consistent with @mikedilger's model. He has something like 6 concepts for where things should be written to/from. This should eventually be added to nip 65.

arthurfranca: I agree `requests/approvals` is better.

Off-topic: @staab @mikedilger got me curious to know what are the new relay marks

staab: [danieletonon](https://coracle.social/npub10000003zmk89narqpczy4ff6rnuht2wu05na7kpnh3mak7z2tqzsv8vwqk) made a neat [infographic](https://coracle.social/note1emmc9zamttnf30r7mkjdenknl90llgu5hd00vyhzw7r7fgyn5pls0azas4) with sort of an "inbox model". I'm not sure how this would be formalized.

vitorpamplona: Markers added.

172.md
```diff
@@ -0,0 +1,101 @@
...
+    ["e", "<Post Request ID>", "<Optional relay url>"],
+    ["p", "<Post Request Author ID>", "<Optional relay url>"],
+    ["k", "<New Post Request kind>"],
+  ], 
+  "content": "<New Post Request JSON>"
+}
+```
+
+It's recommended that multiple moderators approve posts to avoid deleting them from the community when a moderator is removed from the owner's list. In case the full list of moderators must be rotated, the new moderator set must sign new approvals for posts in the past or the community will restart. The owner can also periodically copy and re-sign of each moderator's approval events to make sure posts don't dissapear with moderators.
```
Giszmo, line 81: How about making the moderator list append only and give each moderator a expiration date? Then you can remove somebody starting last week without affecting all his approvals. As these approvals are even bigger than whatever they are approving, duplicating them should not be overly encouraged.

172.md
```diff
@@ -0,0 +1,101 @@
...
+  "id": "<32-bytes lowercase hex-encoded SHA-256 of the the serialized event data>",
+  "pubkey": "<32-bytes lowercase hex-encoded public key of the event creator>",
+  "created_at": "<Unix timestamp in seconds>",
+  "kind": "4550",
+  "tags": [
+    ["a", "34550:<Community event author pubkey>:<d-identifier of the community>", "<Optional relay url>"],
+    ["e", "<Post Request ID>", "<Optional relay url>"],
+    ["p", "<Post Request Author ID>", "<Optional relay url>"],
+    ["k", "<New Post Request kind>"],
+  ],
```
arthurfranca, line 76:
```suggestion
    ["!", "<Community event author pubkey>"],
  ], 
```
This would use #539 to make the community owner able to delete post approvals created by others (moderators).

Currently the owner can only delete its own approvals or remove moderators (which deletes all approvals from said moderator).

172.md
vitorpamplona: I don't think it works because the expired moderator can post in the past. I don't think that should be allowed once a moderator is kicked out.

0xtrr:
It seems like this wouldn't scale very well if we get a community with several hundred or thousands of posts a day. Moderators would have to work full time on approving posts. I'm not sure what the solution to this is but I wanted to mention it anyways.

Also, on Reddit you're able to subscribe to subreddits. How would that work with nostr? Could you have a list of `a` tags referencing the communities you "subscribe" to? The feature here would be to let a NIP-172 enabled client list the communities you're actually interested in so that you wouldn't need to actually browse all the communities each time you visit that client.

pj8912:
> It seems like this wouldn't scale very well if we get a community with several hundred or thousands of posts a day. Moderators would have to work full time on approving posts. I'm not sure what the solution to this is but I wanted to mention it anyways.

I use reddit often, depending on the rules of the community the posts will be moderated/removed quickly by mods or bots. Not every post in a subreddit is approved. Removed yes, approved not necessarily. 
 
Not sure if this depends on the client(s) implementation.

Maybe the client(s) can display all the posts and, the approved posts may have green tick symbolizing the approval.

vitorpamplona:
> It seems like this wouldn't scale very well if we get a community with several hundred or thousands of posts a day.

This can be solved by a good Community Manager client. Auto approvals by known participants, for instance, can be a thing that is solved offline. With this model, we don't need to solve any of the scaling issues in the protocol. 

> subscribe to subreddits

On Amethyst, we will have subscriptions in the default Contact List as well as NIP51 Lists. The contact lists subs of both, communities and hashtags, will be displayed as a "List" in Amethyst's top bar while the NIP51 should be reserved for making collections of communities you want to display all at once.

armstrys:
> On Reddit, your topic owner is less visible but you are still fully dependent on his/her decisions. On Nostr, hopefully, we add more weight to the choice of owner. I think clearer is better. But we lose the single topic handle (for the good and for the bad)

Is there a reason that the d-tag of a community isn't `["d", "<Event author pubkey>:<Community name>"]` instead of just `["d", "<Community name>"]`. Couldn't we then just use a NIP-19 bech encoded `naddr` to link to a community, which solves the linking problem raised by @vivganes 

Also, I am unclear from the current NIP how or if long-form content would be able to interact with the moderated communities. Can the eventid be referenced for a long-form event and it just needs to be reapproved if replaced or does the current spec not work with replaceable events. Would be nice for the long-form format to be included here since many clients tend to render Markdown in long-form but not in kind 1.

vitorpamplona:
There are two ways to work with Long Form content. On the approval event:  
1. Tag the `kind:30023` event with `a` and allow the reader's client to update the post to the newest version without additional approvals by any moderator. 
2. Tag the `kind:30023` event with `e` and require every new change to be approved again or be stuck in time. 

Keep in mind that relays delete old versions of `kind:30023`, which means the `.content` of an `e` approval MUST have the specific version you want or the reader might not be able to find that version of the content anywhere.

armstrys:
Thanks, I see that in the nip now - I am not sure how I missed that when reading.

There is an interesting distinction between those two behaviors depending on whether the moderation is via approval vs removal. In the current approval model it is more conservative to approve by the `e` tag, but in the removal model that would result in a moderator having to chase the new event every time it is change.

I think this points to a similar issue with potential spam that makes me much more in favor of the current "post by approval" model. Moderated communities will not be effective if there are easy mechanisms by which bots can spam them with content that needs to be removed - just look at global on many clients to see the likelihood of this happening. The alternative of having to displace the mods by starting a forked community seems like a much preferable negative outcome.

bigOconstant:
I for one want communities that can be moderated after the fact. Having to approve each post doesn't feel like it would be very active. Feels more like a bulletin board. Not that it wouldn't be useful but I would absolutely rename this because this is less of a community and more of a private group with public post.

arthurfranca:
🚀️

bigOconstant:
> I am not a native english speaker; chances are I overlooked or mis-read something.
> 
> So, this is just to clarify:
> 
>     * Each Community - or, curated topic - must have new posts approved - and, ALL posts - by a moderator?
> 
>     * Since Nostr has no such things as a name, just an ID, we'd use a new scheme like `ncom1` to find them (which could potentially be embedded in QR codes et cetera)?
> 
> 
> After that, here's some things that stuck out to me while reading all the comments here.
> 
>     * While "moderation god complex" is an unfortunate side effect of giving someone power (which reminds me a lot of Megadeath's "Symphony of destruction" first few verses), it also feels like a neccessary evil. Leaving a community unmoderated can often lead to chaos. Moderating it can give some people too much power. A community is built on trust and expectations. So: Moderators should be allowed to be reported. I noticed that in Snort, I can even see all reports sent from other people, referencing the very events that offended them. This method of public reporting could be used, to inform the community creator/"admin", that certain action should be taken.
> 
>     * Why is every post approval-first? Many communities, like the meme-focused ones - thrive on fast posting; which becomes very difficult if each and every post has to be verified. And, someone would have to do that. One moderator can only do so much at a time... A community, or curated topic, should have an option to declare the posting schema: approve-first (each post needs approval), approve-after (if a post is deemed unfitting or otherwise to a community, it may immediately be dropped per delete event), freeform (everyone can post and those are seen as approved and are handled in post by up- or downvotes).
> 
>     * And, in terms of searchability: NIP-05. I am surprised this hadn't been brought up yet. But communities could be made much more searchable by using NIP-05, indicating the authenticity of a certain community as a whole.
> 
> 
> As for a little side-tangent: Nostr as a Reddit alternative ought to work out better than Lemmy. I tried it, even run my own instance, but the super seemless nature of Nostr is just much easier to use. Make a keypair, and off you go. Pick a few relays that sound nice - or that advertise topics you want - and start posting. No crazy instance finding and cross-instance configuration. I am all for this NIP PR! =)

Voting has little chance of success in content curation in nostr where anyone can generate a public key.

Giszmo:
On the concern of scalability with the requirement for approvals: With multiple approvers, these could also take roles such that automatic approval would not result in all getting the message. For example with levels. Group members only see level 5 approved content by default. Moderators (and users of course) can choose to see level 1 and above, with bots giving level 1 approval.

Then it's also less dangerous to grant level 2 or 3 approval permission lightly. The founder/owner could be level 10 with moderators by default being level 5 etc.

bigOconstant:
Why should we put the responsibility on bots for auto approval though? Shouldn't we just add some moderation functionality to the relays? 

If the user doesn't like the moderation of the relay they can just switch relays.

vitorpamplona:
> Why should we put the responsibility on bots for auto approval though? Shouldn't we just add some moderation functionality to the relays?

Not on bots, but on Community Management clients. Relay operators do not have the knowledge of each of their Communities' topics to moderate them correctly. And forcing each community owner to also run a Relay is unnecessary. This proposal splits the moderation responsibilities between relay operators and moderators to help decentralize the power granted by users of such activities.

bigOconstant:
> If moderators are doing a bad job, any other person can close the community and start doing a better job. Relays operators don't get additional responsibilities.

What do you mean any person can close the community? Could you describe that more?

> Not on bots, but on Community Management clients. 

so you might have a service running with a public key that is a "moderator" that is scanning for new community post. It auto approves new post? that is like a bot to me but might work .  how would banned users work?

vitorpamplona:
> What do you mean any person can close the community? Could you describe that more?

Sorry. I meant cloNe the community. 

> so you might have a service running with a public key that is a "moderator" that is scanning for new community post. It auto approves new post? that is like a bot to me but might work . how would banned users work?

You could have a client that implements a NIP51 list of people with auto-approval status into the community. Or the opposite, a list of people that the client will always reject. More importantly, the same client can then provide tools for users to review what the bot is doing. As @Giszmo said, you can have multiple layers of approvals before you get into the community.

bigOconstant:
Might I ask, how would a user discover new communities? How would the experience be if there were multiple communities with the same name? What if I the user wanted to get all the post from community "Mustang" ran by moderators:[1346432,3957820] and also community "Mustang" ran by the other moderators[096866432,09788604]  who disagreed with moderators [1346432,3957820] about something trivial?

I feel like if a community was just a name then a user could pull post from that name from as many relays as they wanted. But if you make a community like this you are going to get lots of over lap. 15 pics communities all with only a few users and their approval bots would all work differently. 

Just trying to understand the flow here.

vitorpamplona:
> Might I ask, how would a user discover new communities? 

Clients have this magic thing of filtering and/or sorting by the number of posts, authors of those posts, follows, zaps, reactions, reports, citations, bookmarks, geolocations, and other kinds. Some of us can even run emotion extraction engines to select only happy communities. :) 

> I feel like if a community was just a name then a user could pull post from that name from as many relays as they wanted.

You can definitely do that. There are no ties to any relays or any protections from copying it. You can even get all the approved content AND the non-approved ones and then approve everything yourself. I don't think a lot of people will follow you on that journey, but you can do that. 

> get lots of over lap.

We will. And the bigger the overlap we get, the better curation will work (from a user's side). You don't need to follow just one Bitcoin community. You can follow all of them. And if the same post gets in all of them, it means that it is really good and should definitely be seen by you.

bigOconstant:
I like it. Thanks

172.md
```diff
@@ -0,0 +1,101 @@
...
+    ["p", "<Post Request Author ID>", "<Optional relay url>"],
+    ["k", "<New Post Request kind>"],
+  ], 
+  "content": "<New Post Request JSON>"
+}
+```
+
+It's recommended that multiple moderators approve posts to avoid deleting them from the community when a moderator is removed from the owner's list. In case the full list of moderators must be rotated, the new moderator set must sign new approvals for posts in the past or the community will restart. The owner can also periodically copy and re-sign of each moderator's approval events to make sure posts don't dissapear with moderators. 
+
+Replaceable events can be submitted for approval in two ways: (i) by their `e` tag if moderators want to approve each individual change to the repleceable event or (ii) by the `a` tag if the moderator trusts the original post author to not modify the event beyond community rules.
```
armstrys, line 83: There is a third interesting option here to approve by both the `a` and `e` tag. This gives the benefit of ii but would also allow the client and moderation tools to easily flag replaceable events that have changed since approval without losing the connection to the community immediately. An alternative wording that could describe this:

```suggestion
Post-approval for replaceable events MAY include an `a` tag in place of or in addition to the `e` tag to ensure that the event remains connected to a moderated community after being edited. Clients SHOULD indicate that replaceable events with an approval on the `a` tag, but a conflicting `e` tag have been edited after approval.
```

armstrys: That makes sense. Something like this?

```suggestion
Post-approval for replaceable events can be submitted in three ways (i) by their `e` tag if moderators want to approve each individual change to the replaceable event, (ii) by the `a` tag if the moderator trusts the original post author to not modify the event beyond community rules, or (iii) preferably with both the `a` and `e` tag so that clients can indicate if a post has been edited since approval.
```

vitorpamplona: Just committed a change as well. See if you like it

armstrys: @vitorpamplona your latest changes LGTM!

172.md
```diff
@@ -0,0 +1,101 @@
...
+    // moderators
+    ["p", "<32-bytes hex of a pubkey1>", "<optional recommended relay URL>", "moderator"],
+    ["p", "<32-bytes hex of a pubkey2>", "<optional recommended relay URL>", "moderator"],
+    ["p", "<32-bytes hex of a pubkey3>", "<optional recommended relay URL>", "moderator"],
+	
+    // relays used by the community (w/optional marker)
+    ["relay", "<relay hosting author kind 0>", "author"],
+    ["relay", "<relay where to send and receive requests>", "requests"], 
+    ["relay", "<relay where to send and receive approvals>", "approvals"],
+    ["relay", "<relay where to post requests to and fetch approvals from>"]
```
vivganes, line 37: Here, Can we add an 'optional' tag suggestion like this 👇

```javascript
//optional information while forking a community from an existing one
["forksource", "34550:<source community event author pubkey>:<d-identifier of the source community>", "<Optional relay url of source community>"]
```

This would provide a clean mechanism for 
1. Any client to show unmoderated feeds for 'forked communities' based on the user preferences [This is not possible without this additional information].  Client can simply follow the 'forksource' tag and show the events from old community, before the timestamp of forking.
2. Ability to see how many people have forked my community.  This creates a very similar feeling of joy when someone creates a fork of my open-source repo.  This provides motivation for moderators to create a quality community.

vivganes (changes requested):

172.md
vitorpamplona: how do you know the time stamp of the forking since the created at event will change every time the community has a new moderator?

vivganes: Ahhh my bad! Good point. 

How about adding this suggestion instead:

```
["forksource", "34550:<source community event author pubkey>:<d-identifier of the source community>", "<Optional relay url of source community>","optional timestamp of the fork"]
```
This can be considered similar to creating a new git branch from a specific commit

vitorpamplona:
There is an ambiguity to solve in mentions. 

When a Kind 1 mentions the community in the text, the post is supposed to be sharing a community with the follows and not a post to be approved by the community. 

Amethyst posted with a "mention" marker, but the filter to download post approvals must remove mentions if this is the way forward. 

```json
{
  "content": "Super active art community. Follow bellow.\n\nnostr:naddr1qqp5zun5qgs0fk6jwz7ejxchh6s7d5p473w7uwffr8pfga9m4sgrgtfz836wp5qrqsqqpphkqfarxc",
  "created_at": 1689305952,
  "id": "119a56648b87c56ba9eb27aa7e571719cd977d2b749888a0e3f96a3165fac16f",
  "kind": 1,
  "pubkey": "460c25e682fda7832b52d1f22d3d22b3176d972f60dcdc3212ed8c92ef85065c",
  "sig": "4abc30e65fd58f6ea838f94f7e76779872ad5b5f3d8b4cbce3c1f47727a131d52689c6e3f263326792ed33cd5fbcc08f4a70823f253abd2cdfa6643d9e7d492a",
  "tags": [
    [
      "p",
      "f4db5270bd991b17bea1e6d035f45dee392919c29474bbac10342d223c74e0d0",
      "",
      "mention"
    ],
    [
      "a",
      "34550:f4db5270bd991b17bea1e6d035f45dee392919c29474bbac10342d223c74e0d0:Art",
      "",
      "mention"
    ]
  ]
}
```

arthurfranca:
Probably better to use another letter for "New Post Request". Similar to how quotes are going to use "q" instead of "e".

mplorentz:
I love this NIP and it's great to see it already implemented in 3 clients in just a few weeks. We are strongly considering adding support for it in Nos. I want to suggest two modifications that are important to me:
- I agree with others that curated communities with pre-approval moderation and approved-until-revoked moderation are similar enough that they should both be accounted for in this NIP. An `approvedUntilRevoked` flag in the NIP and an example of an approval-revoking event (kind 4551? 1984?) are enough to support this use case, and will lower the burden on small teams of moderators and improve the user experience for certain communities.
- I would love to specify a kind other than `1` for notes intended to be shown "in" a community and _not_ to my accounts main feed. I want to post freely in communities about my niche interests and hobbies without annoying all my followers. I think `kind: 11` would be cute as it would have the same semantics as kind: 1, it just isn't intended to not be shown in the main feed. (This kind could be reused for future types of communities that want the same feature, and could even be used to i.e. create topical feeds under your own profile using hashtags. Maybe it deserves its own NIP and could just be referenced here?)

vitorpamplona:
> pre-approval moderation 

How would one manage the list of approved people? Would you add 100s of users in the community definition event that are approved to post directly? Maybe a reference to a NIP-51 list? Maybe a separate membership list entry event to avoid creating massive lists? Looks like the filter to get posts could be massive. 

> I would love to specify a kind other than 1 for notes intended to be shown 

I am starting to think this is a must. But looks like kind 11 is already been used for something else. Any other cool number suggestions? :) 

> Probably better to use another letter for "New Post Request". Similar to how quotes are going to use `q` 

The use of the `a` in Kind 1 should represent a reply to `a`, which is actually not that wrong for community posts: a reply to the community. If the app has a tab for new threads and another for replies, the post should appear in the reply tab instead of the main new threads one. 

The issue is when the user just wants to quote it and not reply to it. 

> Ambiguity 

I am not sure what to do here. I feel like this is a bigger problem with quoting any replaceable event kind, since there is no `q` for them.  `q` is supposed to be used only for id hexes and not for `a` tag assemblies. We will need a new `q` to quote replaceable events.

arthurfranca:
> We will need a new q to quote replaceable events

Don't see much problem in setting an event address (instead of id) to `q` tag when quoting a replaceable event.

bigOconstant:
After playing with it a bunch on Amythest we definitely need approve until revoked. Just so people can get started really quickly. For more complicated moderation we can write client bots. Most of the clients that have implemented it like flycat, zappit all have show unmoderated options because the flow is just to slow without it. But that would make spam a problem.

mplorentz:
@vitorpamplona 
> How would one manage the list of approved people? Would you add 100s of users in the community definition event that are approved to post directly?

I wasn't imagining a list of approved people. Rather a pre-approval community would display all notes from all users, unless a note has been explicitly removed from the community by one of the moderators (maybe via a NIP-32 label?).

However a list of approved community members in a list is another valid style of moderation. But I feel like supporting groups with membership is starting to diverge from the reddit-style type of group and into another type (Discordish), which probably should be in a separate NIP.

No ideas yet on a replacement number for kind 11. But I don't see anything merged with kind 11 yet so maybe there is still a chance it could happen. I think I'll write up a separate NIP for this and see if it gets any traction.

vitorpamplona:
Folks, since we have so many clients already implementing this NIP with some variations, I'd like to merge this as is in order to facilitate new PRs from everyone else that wants to modify it/add new types further. 

@fiatjaf @staab

fiatjaf:
Gonna change this number.

PR #609, opened 2023-06-16
NIP-91: Bech32 URL Query

tyiu:
This specification enables a client on any platform to replace a bespoke URL that routes to a Nostr entity on different web client to instead route internally to its own presentation of that same entity in its own client if it is supported.

It is similar in intent to the [NIP-21](21.md) `nostr:` URI scheme to enable maximum interoperability and openness in the network. However, this specification is mainly for `HTTP`/`HTTPS` where shared links from bespoke web clients might not follow the `nostr:` URI scheme and because modern web browsers do not support the `nostr:` URI scheme yet.

### DIFF
--- a/91.md
+++ b/91.md
@@ -1,0 +1,56 @@
+NIP-91
+======
+
+Bech32 URL Query
+----------------
+
+`draft` `optional` `author:tyiu`
+
+This specification enables a client on any platform to replace a bespoke URL that routes to a Nostr entity on different web client to instead route internally to its own presentation of that same entity in its own client if it is supported.
+
+It is similar in intent to the [NIP-21](21.md) `nostr:` URI scheme to enable maximum interoperability and openness in the network. However, this specification is mainly for `HTTP`/`HTTPS` where shared links from bespoke web clients might not follow the `nostr:` URI scheme and because modern web browsers do not support the `nostr:` URI scheme yet.
+
+## Rationale
+
+Each web client can have its own bespoke URL to display a Nostr entity. For example:
+* https://snort.social/p/npub1yaul8k059377u9lsu67de7y637w4jtgeuwcmh5n7788l6xnlnrgs3tvjmf
+* https://snort.social/e/nevent1qqs2u9n02ftcmrgnhyenyyu0sl8t8dr7tu94t398r24dkcp0kn9taqsppemhxue69uhkummn9ekx7mp0z2jws2
+* https://coracle.social/people/npub1yaul8k059377u9lsu67de7y637w4jtgeuwcmh5n7788l6xnlnrgs3tvjmf/notes
+* https://primal.net/tyiu
+* https://primal.net/profile/npub1yaul8k059377u9lsu67de7y637w4jtgeuwcmh5n7788l6xnlnrgs3tvjmf
+* https://primal.net/thread/note14ctx75jh3kx38wfnxgfclp7wkw68uhct2hz2wx42mdszldx2h6pq8x7dwk
+
+When a different client displays a URL in this format and the user wants to navigate to it, it forces the user to leave their current client, which makes for a poor user experience. The ideal experience should be keeping the user in the same client to view the Nostr entity instead if it is supported.
+
+## URL query with Bech32-encoded entity
+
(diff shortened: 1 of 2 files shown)


staab:
NACK at first impression, I don't think we need this. I can think of two simpler solutions:

- Most clients already parse urls and bech32 entities in content, clients could pull urls from the content, then check each url for an embedded bech32 entity and if found, replace the url with that entity. That has the nice benefit in clients like of auto-embedding the linked content rather than using a link preview scraper which frequently fail.
- This NIP depends on client adoption to work, if clients want to support that they should parse links to themselves in content and replace them with bech32 entities/nostr URIs, before publishing.

tyiu:
> NACK at first impression, I don't think we need this. I can think of two simpler solutions:
> 
> * Most clients already parse urls and bech32 entities in content, clients could pull urls from the content, then check each url for an embedded bech32 entity and if found, replace the url with that entity. That has the nice benefit in clients like of auto-embedding the linked content rather than using a link preview scraper which frequently fail.

I agree what you suggested is simpler. However, it seems 1) a bit implicit in intent and 2) it requires finer-grained string parsing of the URL (which is a smidge more performance intensive and error-prone).

> * This NIP depends on client adoption to work, if clients want to support that they should parse links to themselves in content and replace them with bech32 entities/nostr URIs, before publishing.

It does depend on client adoption to work. It does seem like a good way of expressing the originating intent of either
1. I actually want you to see this content from my web client, or
2. I don't care where you view it from as long as you see it.

vitorpamplona:
This still requires apps like Damus and Amethyst to intercept deep links by every other client. Which I think it's not a polite thing to do. 

I'd love to see us agree on using shareable links with a domain name that is designed for interception. 

If your client offers `http://nostr.open/p/<bech32>` or `http://nostr.open/e/<bech32>`, it means it's ok for another client to intercept it. The `nostr.open` website then just severs as a redirector to the client of choice of the user.

jb55:
I think this makes sense, you could always make it a client option to not route externally when you click these in-app if you think it's impolite to intercept it.

tyiu:
> This still requires apps like Damus and Amethyst to intercept deep links owned by every other client. Which I think it's not a polite thing to do.
> 
> I'd love to see us agree on using shareable links with a domain name that is designed for interception.
> 
> If your client offers `http://nostr.open/<bech32>`, it means it's ok for another client to intercept it. The `nostr.open` website then just severs as a redirector to the web client of choice of the user.

@vitorpamplona That seems centralized. Also, what happens if a browser opens the URL -- where should it redirect to? Conceptually, it's clever, but I'm not convinced that's the right choice.

I don't think it's necessarily impolite -- a query parameter indicates that you welcome any client to intercept the link. I think it can be a setting if necessary.

arthurfranca:
I get that the https:// url is a a great fallback for clients that may not support some kinds of mentioned events, but if using this when writing notes than NIP-21 will probably never be adopted.
So there's a positive and a negative side and I don't know which one weighs more.

alexgleason:
Related: https://gleasonator.com/@alex/posts/AWh4ZaUEWoe9dZmBBQ

This problem affects deep-linking, but also potentially quote posts. Users want to paste a link to a post, and it becomes a quote post. This is a highly requested feature.

On the Fediverse it isn't remotely solvable. Fortunately on Nostr, nip19 is [easy to detect](https://github.com/nbd-wtf/nostr-tools/blob/0e364701da5481de4f8a227fc4ce34ed40a8fb18/nip19.ts#L12). I was planning to just use a regex on the URLs to solve it on Nostr.

I don't know about the query string. I think it will make the URLs too long, so I don't think clients will actually do it. It would be great if web clients adopted some standardized URL routes, though.

> a bit implicit in intent

Good point. For example, consider the URLs: `NostrHealth.com/healthcheck/npub1...`, `NostrCriminals.com/report/npub1...` - these websites shouldn't be deep-linked because they are specialized.

However between standard Twitter-like clients such as Iris, Snort, Coracle, etc I don't think intercepting these links (at the user's preference) is impolite at all.

vitorpamplona:
I wonder if there is any side effect to deep links. We don't want people that are navigating in Snort/Iris/Corarcle on mobile to be asked to open Amethyst when they are just navigating the website. The popup can get really annoying if it's a catch-all.

tyiu:
Closing this PR because we were unable to build consensus.

PR #610, opened 2023-06-16
Kind 16 generic reposts

fiatjaf:
Solving https://github.com/nostr-protocol/nips/pull/576.

### DIFF
--- a/18.md
+++ b/18.md
@@ -6,8 +6,8 @@ Reposts
 
 `draft` `optional` `author:jb55` `author:fiatjaf` `author:arthurfranca`
 
-A repost is a `kind 6` note that is used to signal to followers
-that another event is worth reading.
+A repost is a `kind 6` event that is used to signal to followers
+that a `kind 1` text note is worth reading.
 
 The `content` of a repost event is empty. Optionally, it MAY contain
 the stringified JSON of the reposted note event for quick look up.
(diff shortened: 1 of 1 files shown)


staab:
Do you mind if I use this for moderating public groups?

fiatjaf:
Why not make a new kind?

staab:
I'm mostly joking, I won't use this because I don't use reposts in Coracle, but I do support nevent embeds, which currently are breaking kind 1's based on your concept of composition being non-interoperable, this PR doesn't solve that.

fiatjaf:
Everybody supports nevent embeds, I don't think they are breaking. I also am not a radical about not having composition, I just think people shouldn't be radical about having full composability.

staab:
Interesting, what is the difference between a re-post and an embed, other than the added context?

fiatjaf:
It's the same. The embed is a quote-repost and vice-versa. That's how Twitter works too. The canonical content of a Twitter quote-repost is just a comment plus a link to the other tweet.

NIP-18 reposts are pure reposts, with no added comments.

mikedilger:
ACK

vitorpamplona:
Can we change the advice to add the event into `.content` for 16? 

```
It SHOULD contain the stringified JSON of the reposted note event for a quick look up.
```

vitorpamplona:
Amethyst v60+ supports this change. Kind 1s will be boosted with kind 6 and everything else with kind 16 and the k tag. All generic boosts are shown in the feed just like kind 6s are.

fiatjaf:
Great. Thanks!

fiatjaf:
> Can we change the advice to add the event into .content for 16?

I've changed this on a separate commit.

arthurfranca:
People will use kind 6 to repost all event kinds and also kind 16 with `k: ["1"]` tag ^_^

fiatjaf:
Do you think?

fiatjaf:
[habla.news](https://habla.news/) has also already stopped sending kind:6 and is sending kind:16 now: https://github.com/verbiricha/habla.news/commit/026b678faa1445a7fe994c49201a1e1dbbb0eea2

arthurfranca:
just kidding xd
I think main clients will follow spec and that's what matters most

PR #615, opened 2023-06-20
Update 02.md

dipunm:
This clarifies what field should be used to determine "newer" contact lists as there is no mention of the "created_at" field in this document at all.

If this is incorrect, then I think it proves that more clarification is needed.

### DIFF
--- a/02.md
+++ b/02.md
@@ -25,7 +25,7 @@ For example:
 }
 ```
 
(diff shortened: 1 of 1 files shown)


pablof7z:
This is implied; created_at is the only field that provides a way to "order" these events. This is also not specific to this event kind but to multiple event kinds that are replaced by kind number.

fiatjaf:
Yes, it's implied. This change is correct, but I think it is unnecessary and it would reduce the overall clarity of the text.

PR #618, opened 2023-06-21
NIP-96: Code Collaboration over Nostr

KaffinPX:
I'm proposing and developing a solution for managing code/repositories over Nostr without storing actual data on it but on an alternate network.

CLI client: https://github.com/fostr-dev/cli
Web client: https://fostr.dev -- source code: https://github.com/fostr-dev/web

My profile is npub12nfr7wjr3exmfgn2ftl6xk2lqpwafz6ck7xpdlfwksw335ymprjsw2rfe6

### DIFF
--- a/96.md
+++ b/96.md
@@ -8,7 +8,7 @@ Code Collaboration over Nostr
 
 This NIP defines a way to collaborate on coding over nostr by using it as transport and organizer layer. We could implement it for solutions like Git but their scheme is not compatible with async networks like nostr because of their commit chain, if one commit is missing we cant build latest state of files. This NIP will consider every commitment as revision which will include latest state of files and won't keep state on nostr but just an unique indicator(an integrity proof) for file.
 
-## Definition of a File Distribution Event
+## Definition of a Code Collaboration Event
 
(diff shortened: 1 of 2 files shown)


fiatjaf:
I think this spec is too vague, and doesn't seem to have any immediate use cases. People mostly use git for code collaboration, not sending files over a web client. And this relies on IPFS to store things?

KaffinPX:
> I think this spec is too vague, and doesn't seem to have any immediate use cases. People mostly use git for code collaboration, not sending files over a web client. And this relies on IPFS to store things?

Its successor of https://github.com/nostr-protocol/nips/pull/223 and Git was a centralized service and our CLI client mimicks it well, IPFS is only used by our clients, people can use torrents, tor links or smth that proves integrity on their custom clients

Also about theres no any immediate use cases, our current IPFS default node is storing more than 50 GB, we saw lots of things already uploaded like Bitcoin whitepaper. I hope people decentralizes their repos more by pinning.

image

bndw:
It seems like this could be accomplished with the existing NIP-94 event.

PR #619, opened 2023-06-23
NIP-24: Rich Text Note

AsaiToshiya:
This NIP defines kinds for a "rich text note" such as HTML and Markdown that are primarily displayed in "social" clients.

Read [here](https://github.com/AsaiToshiya/nips/blob/nip-24-rich-text-note/24.md).

Related: #589

### DIFF
--- a/24.md
+++ b/24.md
@@ -1,0 +1,56 @@
+NIP-24
+======
+
+Rich Text Note
+--------------
+
+`draft` `optional` `author:AsaiToshiya`
+
+This NIP defines kinds for a "rich text note" such as HTML and Markdown that are primarily displayed in "social" clients.
+
+Events with these kinds are basically the same as kind `1` but with different the `content`.
+
+## Event Kinds
+
+| kind   | type     |
+| ------ | -------- |
+| `1011` | HTML     |
+| `1012` | Markdown |
+
+### HTML
+
+This type of event uses kind `1011`.
+
+The `content` should be written in HTML. And it can include references to other events using [NIP-21](21.md).
+
+Example event:
+
+```json
+{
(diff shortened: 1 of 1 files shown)


fiatjaf:
This is a slippery slope and will for sure cause centralization due to users migrating to clients that support these things over others.

I very strongly ask and recommend that you don't implement this.

AsaiToshiya:
I think I understand your thinking. I would like to hear what the rest of you think.

leoperegrino:
I think HTML is not that hard to implement for webclients.
Markdown is not that trivial, but it could be left to NIP-23 that already uses markdown.
In HTML, Webclients need to do something else besides replace `script` tags?
With so many native js applications, HTML does not seem too bad, no?

mikedilger:
The problem with HTML content is that clients that are not built on web technologies like gossip would need to bundle in a web rendering library which is a gigantic thing with plenty of security implications.  For clients that are built on browser technologies, they would need to apply a lot of scrutiny to be sure there are no security holes.  I'm very much against HTML in nostr.

As for markdown, we have NIP-23.

leoperegrino:
I understand but I'm not sure why is there a NIP for markdown and not one for HTML

AsaiToshiya:
NIP-23 is a parameterized replaceable event. If you think notes are also replaceable, NIP-23 is enough as you say.

AsaiToshiya:
I close this PR in favor of https://github.com/nostr-protocol/nips/pull/619#issuecomment-1604093362 and https://github.com/nostr-protocol/nips/issues/646#issuecomment-1628888642.

If you need HTML, I think you should propose it along with the new use cases that require it.

PR #627, opened 2023-06-29
NIP-88 WIP Nostr Cash

arcbtc:
Nostr native, very simple credit/cash/cheque/token system so relays can offer minting/custodial services.

Not blinded by default, so not an ecash, but I imagine a relay could implement a blinded version.

A Nostr cash system should be versatile enough for clearnet/darknet bitcoin/shitcoin. Credits issued by "mints" will vary in quality and trust, but that's okay, users can demand certain mints, or certain payment types (such as bitcoin, where value can be redeemed immediately). Swap services will also emerge over time.


88.md
```diff
@@ -0,0 +1,195 @@
...
+
+User DMs the relay:
+
+```json
+{
+  "amount": <integer millisats>
+}
+```
+
+Relay responds with payment request, which user pays:
```
motorina0, line 91: Should use [NIP-15 payment_options](https://github.com/nostr-protocol/nips/blob/master/15.md#step-2-merchant-request-payment-event) for consistency

arcbtc: Yes! Great

fiatjaf:
I don't understand why use this instead of Cashu, for example, which is already a simple protocol that seems to be battle-tested and used in some places.

88.md
```diff
@@ -0,0 +1,197 @@
...
+
+Electronic cash is useful and relays should have the option to offer minting/custodial services. Other mints could exist as stand-alone clients, but relays are ideally placed to package in with their service offerings.
+
+Nostr Cash is simple and flexible. User gives a mint some money and the mint issues the user electromic cash/cheque/token that can be used as money in nostr.
+
+Nostr Cash is entirely custodial and trust based, and is just a common standard for users to send/receive value in nostr. Keeping mints simple means more users can run, for themselves, their family, or the whole world.
+
+Relays offering mints must have a keypair and an additional nostr keypair for its mint.
+
+Nostr Cash uses derived keys, much like [NIP26](https://github.com/nostr-protocol/nips/blob/master/26.md). Each "cash" is actually just a derived private key that can be burned or passed to another user.
```
arthurfranca, line 17:
```suggestion
Nostr Cash uses derived keys, much like [NIP06](https://github.com/nostr-protocol/nips/blob/master/06.md). Each "cash" is actually just a derived private key that can be burned or passed to another user.
```
26 -> 06

arcbtc:
> I don't understand why use this instead of Cashu, for example, which is already a simple protocol that seems to be battle-tested and used in some places.

More because a nostr native ecash/whatever needs to be generic. Nostr is NOT bitcoin, despite them working so well together. I also worry about the type of mints you get if the electronic cash is blinded, and the regulatory issues they will have

melvincarvalho:
While nostr cash is a fun play on words, it will attract negative sentiment

A distributed mint or ledger would be useful with the following functions:

- deposit
- withdraw
- transfer
- balance

blinding is good, but optionally added on

So these functions are done via DMs instead of zaps, could it be both, I wonder.

I've always done this with a ledger table and a credit (transaction) table, which goes back to the old drupal [community forge](https://communityforge.net/en) project by matslats

This formulation is not bad and I think would work.  It wasnt that intuitive what active and spent were at first.

Probably a working demo and/or use cases would help.  Maybe a bot with a battery that can be topped up via DM or zap, and that does something useful like [paint a picture](https://make-me-an-image.bumi17.repl.co/)

cbrunnkvist:
> Cashu

Link?

PR #629, opened 2023-06-30
3% Byte Reduction in a High Data Load

dfralan:
This Pull Request proposes a change in the Nostr Protocol to reduce the size of each transfer by approximately 3%. The change involves replacing the "created_at" parameter with "at" in the protocol.

The average size of a tweet in a similar protocol is estimated to be 280 bytes, which can be considered the average size of a transfer in the Nostr Protocol. Therefore, saving nearly 3% of bytes per transfer is significant.

This byte savings becomes even more crucial in a high data load scenario. As time goes by, the accumulation of transfers increases bandwidth requirements and resource usage. By implementing this change now, we can optimize the protocol's efficiency and reduce the burden on existing systems.

It is important to note that making changes to an established protocol becomes more challenging over time. As more implementations are developed and integrated with the Nostr Protocol, modifying it may require greater coordination and effort. Hence, it is urgent to address this improvement now to avoid future complications and ensure optimal protocol performance.

This Pull Request aims to initiate a discussion and thorough evaluation of the proposed change. We appreciate your feedback, suggestions, and additional contributions to enhance the efficiency and optimization of the Nostr Protocol.

Let's take action now to drive this improvement and ensure a more efficient and scalable protocol as it grows in adoption and usage.

We look forward to your review and feedback on this proposal!

Thank you for your attention and consideration.

### DIFF
--- a/01.md
+++ b/01.md
@@ -18,7 +18,7 @@ The only object type that exists is the `event`, which has the following format 
 {
   "id": <32-bytes lowercase hex-encoded sha256 of the serialized event data>,
   "pubkey": <32-bytes lowercase hex-encoded public key of the event creator>,
-  "created_at": <unix timestamp in seconds>,
+  "at": <unix timestamp in seconds>,
   "kind": <integer>,
   "tags": [
     ["e", <32-bytes hex of the id of another event>, <recommended relay URL>],
@@ -36,7 +36,7 @@ To obtain the `event.id`, we `sha256` the serialized event. The serialization is
 [
   0,
   <pubkey, as a (lowercase) hex string>,
-  <created_at, as a number>,
+  <at, as a number>,
   <kind, as a number>,
   <tags, as an array of arrays of non-null strings>,
   <content, as a string>


fiatjaf:
Thank you for your proposal, but why not apply this technique to all properties and reduce them all to a single letter? That would yield a bigger reduction.

Or maybe even put them all in a array, that would decrease the size even more.

Better than that: we could serialize everything to a dedicated binary format and eliminate the overhead of hex and of quotes around strings.

dfralan:
The suggestions you mentioned regarding further reduction of properties, using arrays, or serializing to a binary format could potentially lead to additional size reductions in the Nostr Protocol. However, it's important to consider the trade-offs and implications of such changes.

Reducing all properties to a single letter or using arrays may significantly impact the readability and maintainability of the protocol. While it could potentially yield a larger reduction in size, it could also make the protocol more complex and difficult to understand for developers and implementers. It may increase the risk of confusion, errors, and potential compatibility issues in the future.

Serializing everything to a dedicated binary format could indeed provide further size reduction by eliminating hex encoding and quotes around strings. However, it would introduce a different set of challenges. Binary formats often require additional processing and decoding steps, which can increase the computational overhead. It might also limit the interoperability of the protocol since binary formats are generally less human-readable and may not be universally supported by all systems or programming languages.

When considering protocol optimizations, it's essential to strike a balance between reducing size, maintaining readability and maintainability, ensuring compatibility, and optimizing performance.

This Pull Request strikes such a balance, addressing the need for size reduction while keeping the protocol relatively simple and widely compatible.

I mean if we are using "sig" instead of "signature", why we should use "created_at" instead of "at"?

fiatjaf:
I think you are an AI.

dfralan:
We all are xD

dfralan:
But now seriously, numbers are numbers, and scalability leads.

dfralan:
You just talk in this same thread about apply this technique and reduce them all to a single letter, or to serialize everything to a dedicated binary format, that is not viable, but now you say that is impossible.

Are you an AI?

The protocol would stay same server and clientside.

Never is too late.

scsibug:
I do not think there is any meaningful savings when websockets w/ per-message deflate is used.

dfralan:
The saving is a fact, not a discussion, you may not be thinking.

Giszmo:
Can we close this already? I regret having read the back and forth that only wasted everybody's time. More complete "improvements" were suggested some 50 times before.

dfralan:
If you went back and forth I made my point. Congratulations, your egos are even bigger than the protocol, puro humo.

PR #630, opened 2023-07-02
Minor edit: specify the unit of the filters' date attributes (seconds)

tcheeric:
Minor edit, for clarification: The time unit for all date attributes in the filters (since, until) is in seconds.


fiatjaf:
Thank you!

PR #631, opened 2023-07-02
Allow array of filters in REQs

tcheeric:
I submit a proposal to allow a JSON array of filters in REQ messages. This is more intuitive than using a comma-separated list. Relays should still accept the latter format, to ensure backward compatibility.


fiatjaf:
Sorry, but this is an unnecessary breaking change.

PR #632, opened 2023-07-02
NIP-37: Language Tag

alexgleason:
Adds a `lang` tag for text events allowing users to specify their language.

NIP text here: https://github.com/alexgleason/nips/blob/dcb38b6e2301e149a2d592d1880bd3a6a7492f03/37.md

### DIFF
--- a/37.md
+++ b/37.md
@@ -1,0 +1,29 @@
+NIP-37
+======
+
+Language Tag
+------------
+
+`draft` `optional` `author:alexgleason`
+
+A `lang` tag may be added to text events to indicate the language of the text. The value of the tag MUST be a valid [ISO 639-1](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes) language code.
+
+For example:
+
+```json
+["lang", "en"]
+```
+
+The language tag is a hint for automatic translations, as well as a tool for filtering by language. It is not a guarantee that the text is in the specified language. Some messages could contain text in multiple languages, text in a different language, or text in no language at all.
+
+### Client behavior
+
+Clients may append a `lang` tag based on the user's selection at the time of the post, or based on a user setting. If the user has not selected a language, the client should omit the `lang` tag.
+
+### Relay behavior
+
+Relays implementing this NIP should validate that the `lang` tag is a valid ISO 639-1 language code, and if so index the event so it can be requested with tag filters:
+
+```json
+["REQ", <subscription_id>, {"#lang": ["en"]}]
+```
--- a/README.md
+++ b/README.md
@@ -50,6 +50,7 @@ They exist to document what may be implemented by [Nostr](https://github.com/nos
 - [NIP-32: Labeling](32.md)
 - [NIP-33: Parameterized Replaceable Events](33.md)
 - [NIP-36: Sensitive Content](36.md)
+- [NIP-37: Language Tag](37.md)
 - [NIP-39: External Identities in Profiles](39.md)
 - [NIP-40: Expiration Timestamp](40.md)
 - [NIP-42: Authentication of clients to relays](42.md)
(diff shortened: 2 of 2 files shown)


fiatjaf:
I think users can't be trusted to specify this tag correctly and all automatic translation tools will already detect the language automatically anyway.

alexgleason:
I basically just copied what Mastodon does since I'm implementing some equivalent features. But I figured automatic detection would be the main counterpoint to this.

I think it would still be cool if relays tagged events by their language and let you filter them.

staab:
I would suggest using NIP 32 for this use case. That would have the benefit of using single letter tags so language would be searchable, and allowing people to both mark their own notes and those of others (for example you could build a bot that publishes 1985 events based on language detection).

alexgleason:
NIP-32 actually seems like it wants a `lang` tag to exist. Because then you could label an event like `["l", "en", "#lang"]`

I'm not sure it makes more semantic sense to add `["l", "en", "ISO-639-1"]` to a note, than just `["lang", "en"]`

vitorpamplona:
I don't think there is a reliable way to implement this. On Amethyst, we could get the system language, the keyboard language or try to process it with the translator before signing. However, they are all faulty solutions. Users don't generally switch keyboards to write in other languages and the in-device language identifier is not that accurate for Nostr post (short bursts of grammatically imprecise text with tons of weirdness like URLs and bech addresses). We could ask the user to specify in the new post screen, but it might be just annoying for the average user. 

On the flipside, I think this could go into a longer discussion on content tagging/rating by third parties. The same NIP that can allow an agency to mark PG-13 content, can also be used to specify the language and other attributes of the post. Users would then follow their preferred agency to get the type of content they like to see.

alexgleason:
The cool thing to do would be for relays to just automatically detect the language with [fasttext](https://fasttext.cc/docs/en/language-identification.html), and then respond to queries for a particular language.

@vitorpamplona This MR is about letting the user set the language manually the way Mastodon does it: 

image

I have mixed feelings about it. I'm creating a compatibility layer with Mastodon. At first I'm going to use this "lang" tag to decide whether to display a "translate post" button to the user. But a better solution will likely evolve.

arthurfranca:
> The cool thing to do would be for relays to just automatically detect the language with [fasttext](https://fasttext.cc/docs/en/language-identification.html), and then respond to queries for a particular language.

Relays can use that, other libs or also check NIP-32 labels if available to decide what is the language. #605 (NIP-17 Event Metadata) PR is exactly for this. Recently added [here](https://github.com/arthurfranca/nips/blob/nip-17/17.md#nip-17language) a way to filter by language.

alexgleason:
I am using a tag like this for the language:

```json
["l", "en", "ISO-639-1"]
```

alopatindev:
I found [various issues](https://github.com/nostr-protocol/nips/issues/1234#issuecomment-2106650873) about language labeling and I'm still confused: do we have any standard way to specify language for event kind 1 right now?

> I am using a tag like this for the language:
> ["l", "en", "ISO-639-1"]

Am right that [NIP-32](https://github.com/nostr-protocol/nips/blob/master/32.md) currently *allows*
```json
{
  "kind": 1,
  "tags": [
    ["L", "ISO-639-1"],
    ["l", "en", "ISO-639-1"]
  ],
  "content": "English text",
  ...
}
```
but it's not that we've decided that it's best way to do that (because some suggest `en-US`, etc.)? Can we have this example in NIP-32?

I'd like to post in multiple languages from single account and give my followers possibility to filter out messages in languages they don't understand; ideally disallow them with mute list. Maybe even give them "Follow my English posts" and "Follow all my posts" buttons in a relevant page of a multilingual website. Does the [NIP-51](https://github.com/nostr-protocol/nips/blob/master/51.md#standard-lists) mute list support the `l` tags?

alexgleason:
Yes your event looks fine. Yes we should have the example in NIP-32.

PR #633, opened 2023-07-03
Fix content field in example in NIP-56 Reporting to replace deprecated NIP-08 indexed mentions with NIP-27 bech32 mentions

tyiu:


56.md
```diff
@@ -75,7 +74,7 @@ Client behavior
 ---------------
 
 Clients can use reports from friends to make moderation decisions if they
-choose to. For instance, if 3+ of your friends report a profile as explicit,
```
tyiu, line 78: Explicit is not a listed report type, so I replaced it with one that is.

56.md
```diff
@@ -63,10 +63,9 @@ Example events
 {
   "kind": 1984,
   "tags": [
-    ["p", <impersonator pubkey>, "impersonation"],
-    ["p", <victim pubkey>]
```
tyiu, line 67: A bit dangerous putting a `p` tag here for clients that don't parse this properly. I imagine it could be easy for a client to misinterpret a victim pubkey for an impersonator pubkey. Removing it from the NIP to reduce the risk of this happening. Instead, favouring just adding it to the `content` field.

56.md
staab: This NIP was never really built to be exhaustive, see #457 and #532 for discussion of the topic. A more complete classification system could be implemented with labels, nips 36 and 56 are sort of just for flagging content for manual review, although it would be fine to continue to build them out.

fiatjaf (approved):

tyiu:
@fiatjaf can you merge this?

fiatjaf:
Thank you for pinging me.

PR #635, opened 2023-07-03
Clarify JSON serialization for the `id` field

Semisol:
Apps can write a JSON serializer themselves for this purpose if they prefer, since the complexity is low.


fiatjaf (approved):

jb55:
On Mon, Jul 03, 2023 at 08:03:32AM -0700, Semisol wrote:
>@Semisol requested your review on: nostr-protocol/nips#635 Clarify JSON serialization for the `id` field.

ACK

01.md
```diff
@@ -43,6 +43,14 @@ To obtain the `event.id`, we `sha256` the serialized event. The serialization is
 ]
 ```
 
+To prevent implementation differences from creating a different event ID for the same event, the following rules MUST be followed while serializing:
+- No whitespace, line breaks or other unnecessary formatting should be included in the output JSON.
+- No characters except the following should be escaped, and instead should be included verbatim:
+  - The line break, `0x0A`, as `\n`
+  - A double quote, `0x22`, as `\"`
+  - A backspace, `0x5C`, as `\\`
```
scsibug, line 51: Do you mean backslash?

01.md
```diff
@@ -43,6 +43,14 @@ To obtain the `event.id`, we `sha256` the serialized event. The serialization is
...
+- UTF-8 should be used for encoding.
+
```
scsibug, line 53: I think this is missing other required escape characters in JSON - backspace, form feeds, carriage returns, and tabs.

01.md
scsibug, line 51: Do you mean backslash?

scsibug (changes requested):
I think this makes things more confusing by proposing a set of escapes that differ from the JSON standard.

fiatjaf:
> I think this makes things more confusing by proposing a set of escapes that differ from the JSON standard.

How does the JSON standard deals with this? The problem with the JSON standard is (I'm guessing) that it was conceived before UTF-8 was widespread, so it accepts escape codes for unicode characters, and no one should be using that, it is very confusing.

scsibug:
JSON has a very small set of codes that must be escaped, but unicode is generally allowed, either unescaped, or escaped.  I guess this is the problem this PR is trying to address, deciding whether or not to transform an escaped unicode character into the native representation before signing?

This spec for JSON signing may be instructive: https://www.rfc-editor.org/rfc/rfc8785#section-3.2.2.2

I don't have a suggestion at the moment for how to proceed, but I think we should be very cautious here to not introduce other problems (security or interop) or make it very painful to deal with signature validation.

Are there concrete examples of implementations that preserve the unicode escapes during signing, or specific characters that cause issues?  I would have thought a reasonable parser would convert "\uXXXX" to a native UTF-8 representation during signing (with the exception of the JSON strings that must be represented with backslash escapes such as newline, tab, etc.).

scsibug:
I guess Python's JSON serialization can output escaped unicode codepoints, which would cause some havoc if that was used for signing or verification.

A "standard" serialization would use /u escapes for U+0000 through U+001F, \b, \t, \n, \f, and \r escapes for backspace/tab/newline/formfeed/carriage return, backslash escapes for double quotes/backslashes, and the native UTF-8 values for everything else.  I think that's what we want to advocate here, which might be best to simply say, "serialize JSON string data according to rfc8785."

fiatjaf:
I would tend to say that people using U+0000 through U+001F, \b, \t, \f, and \r are probably attacking Nostr and their events should be considered invalid.

mikedilger:
@scsibug has pointed to the right way to handle this.

If you do the dumb thing and don't even think about this issue, but instead use JSON.stringify() or rust's serde_json, I think you will get the escaping described in RFC 8785 (someone should test that, I may when I get more time).

This is the full list of characters that must be escaped. No other character may be escaped (NOTE the rendering of this github note is funny because it parses some of these escape sequences, so I had to use spaces in a few places)

00 \u0000
01 \u0001
02 \u0002
03 \u0003
04 \u0004
05 \u0005
06 \u0006
07 \u0007
08 \b
09 \t
0A \n
0B \u000b
0C \f
0D \r
0E \u000e
0F \u000f
10 \u0010
11 \u0011
12 \u0012
13 \u0013
14 \u0014
15 \u0015
16 \u0016
17 \u0017
18 \u0018
19 \u0019
1A \u001a
1B \u001b
1C \u001c
1D \u001d
1E \u001e
1F \u001f
22 \ "
5C \ \

mikedilger:
On the one hand we have

1) The ability to use bells (\b) and formfeed (\f) (odd as those may be in a nostr event),
2) No changes to most clients, and
3) For Javascript:
```
JSON.stringify([
    0,
    evt.pubkey,
    evt.created_at,
    evt.kind,
    evt.tags,
    evt.content
  ])
```

On the other hand we have
1) Forbidding the use of \b and \f
2) Most clients will need to change, and
3)
```
function serializeNostrSigJson(arr) {
    return "[" + arr.map(el => {
        if (typeof el === "number") return el.toString()
        if (el instanceof Array) return serializeNostrSigJson(el)
        if (typeof el === "string")
            return "\"" + el
              .replace(/\\/g, "\\\\")
              .replace(/"/g, "\\\"")
              .replace(/\n/g, "\\n")
              .replace(/\r/g, "\\r")
              .replace(/\t/g, "\\t") + "\""
        throw new Error("Unsupported value " + el)
    }).join(",") + "]"
}
```

fiatjaf:
> Most clients will need to change

Why? They won't.

01.md
```diff
@@ -49,6 +49,8 @@ To prevent implementation differences from creating a different event ID for the
   - The line break, `0x0A`, as `\n`
   - A double quote, `0x22`, as `\"`
   - A backslash, `0x5C`, as `\\`
+  - A carriage return, `0x0D`, as `\r`
+  - A tab character, `0x09`, as `\t`
```
fiatjaf, line 53: there are no valid use cases for these either.

jb55:
I believe this makes sense. In swift I had to this `.withoutEscapingSlashes` do this so it's good to be explicit.

```swift
func event_commitment(pubkey: Pubkey, created_at: UInt32, kind: UInt32, tags: [[String]], content: String) -> String {
    let encoder = JSONEncoder()
    encoder.outputFormatting = .withoutEscapingSlashes
    let str_data = try! encoder.encode(content)
    let content = String(decoding: str_data, as: UTF8.self)
    
    let tags_encoder = JSONEncoder()
    tags_encoder.outputFormatting = .withoutEscapingSlashes
    let tags_data = try! tags_encoder.encode(tags)
    let tags = String(decoding: tags_data, as: UTF8.self)

    return "[0,\"\(pubkey.hex())\",\(created_at),\(kind),\(tags),\(content)]"
}
```

jb55 (approved):

alexgleason:
Does this remain compatible with `JSON.stringify` or not?

PR #637, opened 2023-07-03
add nip-37: general methods for dealing with lost keys

fiatjaf:


mikedilger:
What is the 3 month requirement doing?

fiatjaf:
> What is the 3 month requirement doing?

An arbitrary time to prevent the attacker from publishing a new set of trusted peers and then use them to migrate to a new key under his control immediately.

DanConwayDev (commented):
Overall I think social recovery is an important feature as nostr grows.

I can imagine a future where this is largely handled by social clients with following users not having to take any active steps. This could be similar from a UX point of view, to how signal's 'safety number' changes. Perhaps a warning could be flagged up on their posts for a few weeks to indicate their pubkey has recently been updated and to lookout for unusual activity.

Perhaps the following user's own follows could be used to build trust in the new pubkey recommendation. If their follows who also follow the pubkey have recently updated it to the recommendation then this also builds confidence in the update.

Relays could potentially play a role in building confidence in the new pubkey. They may compare data they have collected related to the old pubkey and compare it to the new key. They could also do timing analysis or sentiment / theme / keyword analysis to assess authenticity.

I'm bullish on the potential of for a smooth UX for recovery from key loss.

37.md
```diff
@@ -0,0 +1,95 @@
...
+    ["e", "<event-of-kind-10520-that-this-refers-to>"],
+    ["new-key", "<new-key>"]
+  ],
+  "content": "",
+  // ...
+}
+```
+
+Very importantly, when evaluating events of kind `1521` for a given `p`, clients should ensure that their corresponding
+event `10520` is at least 3 months older **and that an OpenTimestamps proof exists for that**.
```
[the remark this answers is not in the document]
fiatjaf: OTS can be done by clients or third parties, doesn't have to be the client directly.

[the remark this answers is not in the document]
fiatjaf: Zero usage today, but I don't see why they couldn't start being used if this NIP is to be adopted.

37.md
```diff
@@ -0,0 +1,95 @@
...
+
+```js
+{
+  "kind": 10520,
+  "pubkey": "<user-pubkey-that-will-be-compromised>",
+  "tags": [
+    ["p", "<friend1-pubkey>"],
+    ["p", "<friend2-pubkey>"],
+    ["p", "<friend3-pubkey>"],
+    ["threshold", "2"]
```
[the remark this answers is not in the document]
fiatjaf: I don't know, I think this is up for clients to decide.

37.md
```diff
@@ -0,0 +1,95 @@
...
+    ["new-key", "<new-key>"]
+  ],
+  "content": "",
+  // ...
+}
+```
+
+Very importantly, when evaluating events of kind `1521` for a given `p`, clients should ensure that their corresponding
+event `10520` is at least 3 months older **and that an OpenTimestamps proof exists for that**.
+
```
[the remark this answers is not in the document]
fiatjaf: I think clients can each decide on their own UX.

37.md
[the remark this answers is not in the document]
fiatjaf: Nothing.

37.md
```diff
@@ -0,0 +1,95 @@
...
+prominent message explaining that the action is not reversible, SHOULD explain that this does not guarantee their
+posts are deleted forever, and SHOULD require special confirmation such as requiring the user to type a message.
+
+### Method B: Social Key Migration
+
+The idea here is that users could pick some friends from their contact list and a threshold number to signal to
+the rest that the old key has been lost and what is the new key -- e.g. Alice can determine that, from Bob, Carol
+and Derek, if any 2 of these 3 agree, their recommendation is authoritative.
+
+This is done by publishing an event like this (for example):
```
alexgleason, line 57: If the attacker has the private key, they would simply post a kind 10520 event pointing to other compromised pubkeys, making it pointless.

DanConwayDev:
Here is an alternative suggestion that removes reliance on events published by a potentially compromised pubkey:
```js
{
  "kind": 1521,
  "pubkey": "<pubkey>",
  "tags": [
    ["p", "<compromised-pubkey-of-contact>"],
    ["e", "<other-trusted-kind-1521-event-referring-to-the-same-pubkey>"],
    ["new-key", "<suggested-new-key>"]
  ],
  "content": "out-of-band communication" | "report from trusted contact",
  // ...
}
```

Clients MAY listen for kind 1521 events referencing the user's contacts.
Clients MAY use the user's social graph to identify the trustworthiness of each 1521 event and the aggregate trustworthiness of reports for a pubkey. 
Clients SHOULD update the user's contact lists and publish a kind 1521 when they are assured of trustworthiness. Reports must flow through webs of trusts in order for this mechanism to be effective.

37.md
fiatjaf: That's why I added the OpenTimestamps requirement. Timestamping can be done for free by some altruistic entity for everybody.

37.md
```diff
@@ -0,0 +1,95 @@
...
+  ],
+  "content": "",
+  // ...
+}
+```
+
+Very importantly, when evaluating events of kind `1521` for a given `p`, clients should ensure that their corresponding
+event `10520` is at least 3 months older **and that an OpenTimestamps proof exists for that**.
+
+### Method C: Hierarchically Pregenerated Keys
```
alexgleason, line 93: This was your best idea IMO. I thought you already wrote it out somewhere. But we should absolutely do it. It fits perfectly with NIP-46 and would make loss of keys much less painless, as long as they were generated by an app that supports it (and most apps will).

fiatjaf: https://github.com/nostr-protocol/nips/pull/158

I'm glad that you liked it. I still like it, but it's too complicated, so I left it for a future in which people will be smarter.

37.md
```diff
@@ -0,0 +1,95 @@
...
+caused by having an evil hacker take control of someone else's keys.
+
+The implementation of this NIP could be done directly in social clients or in dedicated clients that users would
+visit from time to time in order to have an automated scan be done on their contacts and then the dedicated app
+could suggest migrating from old keys to new keys, or suggest stop following compromised keys, or even perform
+these migrations automatically.
+
+### Method A: Simple Key Deletion
+
+Based on https://github.com/nostr-protocol/nips/pull/377, the idea is that someone, after noticing that their key
```
ariard, line 23: I think NIP-109 could be modify to concern any pubkey without the restriction to the subset of kind 5 (internet identifier) as ideally you would like Nostr key management standards to work in perfect autonomy from the social layer, e.g doing key operation on hardware where there is no internet access to probe the validity of the internet identifier.

37.md
```diff
@@ -0,0 +1,95 @@
...
+
+```js
+{
+  "kind": 10529,
+  "pubkey": "<compromised-pubkey>",
+  "tags": [
+    ["reason", "I typed my key on anigma"]
+
+    // this is unnecessary, but it's an extra requirement to prevent people from making a 10529 event by chance
+    ["key-compromised"],
```
ariard, line 35: I think this is can be implicit in the value of the “kind” itself and rather make it mandatory for the client to ask a “challenge” to sign a kind 10529 with the compromised pubkey. E.g a “challenge” can be the example given after of requiring a user to type a message. Other “challenge” could be a pre-registered password in the client hard state or really any existent password protocol (e.g a time-based one-time based password).

See also PGP revocation format, it’s just a dummy 0x20 signature type to announce revocation: https://www.rfc-editor.org/rfc/rfc4880#section-5.2.1

37.md
```diff
@@ -0,0 +1,95 @@
...
+### Method A: Simple Key Deletion
+
+Based on https://github.com/nostr-protocol/nips/pull/377, the idea is that someone, after noticing that their key
+has been compromised and deciding that it is not possible to use any of the other methods here described, can
+proceed to publish the following event:
+
+```js
+{
+  "kind": 10529,
+  "pubkey": "<compromised-pubkey>",
```
ariard, line 30: I think it should be mandatory to have a opentimestamp tag, or a chain height / unix time. That way the user can commit to the temporal point relative to the Bitcoin blockchain from which the key is considered as compromise.

This might matter for the relay requirement introduced after “Relays MAY also decide to also delete all events from this key” where the deletion is done in reference to the chain height. E.g past reputation is okay, future reputation not okay.

37.md
```diff
@@ -0,0 +1,95 @@
...
+  "tags": [
```
ariard, line 31: One more tag could be the list of pubkeys actually revoked where the signing pubkey is an authoritative key where the subkeys have previously signed the authoritative key to give a compromise revocation mandate to the authoritative one. Though I think this more method C suggested at the end.

37.md
```diff
@@ -0,0 +1,95 @@
...
+
+Clients may give users the option to "delete your public key" with this type of event. Clients SHOULD display a
+prominent message explaining that the action is not reversible, SHOULD explain that this does not guarantee their
+posts are deleted forever, and SHOULD require special confirmation such as requiring the user to type a message.
+
+### Method B: Social Key Migration
+
+The idea here is that users could pick some friends from their contact list and a threshold number to signal to
+the rest that the old key has been lost and what is the new key -- e.g. Alice can determine that, from Bob, Carol
+and Derek, if any 2 of these 3 agree, their recommendation is authoritative.
```
ariard, line 55: Contrary to my understanding of the introduction, I don’t think method A and method B are exclusive though rather complementary.

From my thinking:
- method A: introduce a key compromise alert system for all relays / clients to do the required local state modification
- method B: introduce a key-based reputation migration system based on pre-committed policy

I think method B can be clearer to specify a “reputation migration certificate”. This certificate should be signed by the requesting key, timestamped in the chain with ots and commit to one or more migration threshold policies (e.g 2-of-2 or 3-of-5).

As long as one relay or client over Nostr has a stored event of the certificate it can be re-announced to the whole network to have the reputation changes happening. Without necessary implying a compromise, it can be good key management practice to rotate keys on “hot” computing hosts or even enable merchant reputation selling.

37.md
fiatjaf: I am just worried because I use this https://github.com/fiatjaf/nak CLI tool to generate events and it's harmless 99.9% of the times, but if me or someone else typo or type at random the deletion kind they are gone from the Earth.

alexgleason: Yeah but it's easier to accidentally copy and paste your private key into a public chatroom. Then it's also gone forever with even less effort.

fiatjaf: So you're saying we should have this NIP specify that you should type your private key in a public chatroom?

ariard: > I am just worried because I use this https://github.com/fiatjaf/nak CLI tool to generate events and it's harmless 99.9% of the times, 

Yeah I would expect with time that type of tools to request a challenge like totp: https://www.rfc-editor.org/rfc/rfc6238 (google authenticator based on that). For sure key compromise event generation gonna stay a worrying operation for user.

PR #640, opened 2023-07-06
NIP-25 content might be NIP-30 custom emoji

mattn:
The content for reactions might be NIP-30 custom emoji.

### DIFF
--- a/25.md
+++ b/25.md
@@ -47,3 +47,25 @@ func make_like_event(pubkey: String, privkey: String, liked: NostrEvent) -> Nost
     ev.sign(privkey: privkey)
     return ev
 }
+```
+
+Custom Emoji Reaction
+---------------------
+
+The client may specify a custom emoji ([NIP-30](30.md)) `:shortcode` in the
+reaction content. The client should refer to the emoji tag and render the
+content as an emoji if shortcode is specified.
+
+```json
+{
+  "kind": 7,
+  "content": ":soapbox:",
+  "tags": [
+    ["emoji", "soapbox", "https://gleasonator.com/emoji/Gleasonator/soapbox.png"]
+  ],
+  "pubkey": "79c2cae114ea28a981e7559b4fe7854a473521a8d22a66bbab9fa248eb820ff6",
+  "created_at": 1682790000
+}
+```
+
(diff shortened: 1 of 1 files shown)


alexgleason:
Would it include the colons, eg `:blobcat:`?

mattn:
Yes, I assume it to be `:shortcode:`.

alexgleason:
I that should be clarified. Also, it means an "emoji" tag will also need to be included on the event.

Have you thought about simply including the URL in the content, without the shortcode? What are the pros and cons of that?

alexgleason:
cc @jb55, custom emoji reactions on kind 7

mattn:
I think the emoji tag has future extensibility. For example, it might be possible to put something in the 4th element of array that can be used to decorate the emoji. This has the advantage that the information can be used from the content, I think.

alexgleason:
Okay, after thinking I support this idea. Here's my reasoning:

1. Custom reactions should not be a new kind. They should be kind 7 so they get tallied with COUNT, and unsupported clients should just drop them as they already do with invalid kind 7s.
2. I support using the "emoji" tag so the emoji definition could be extended in the future.
3. It _could_ be a new NIP instead of adding onto NIP-25, but I don't see any benefit to that since relays shouldn't care about the content of kind 7 events(?), so there's no benefit to the supported_nips field.

I do think this needs a separate section in NIP-25 with a headline like "Custom emoji reactions" with a more in-depth explanation and example event.

alexgleason:
Side-note that custom emoji reactions are already supported and working on the Fediverse, so this is not exactly uncharted territory.

mattn:
Added new section Custom Emoji Reaction.

alexgleason:
Can someone else review this? @jb55 @fiatjaf @staab

jb55:
On Mon, Jul 10, 2023 at 08:43:39AM -0700, Alex Gleason wrote:
>Can someone else review this? @jb55 @fiatjaf @staab

I think this is fine.

vitorpamplona:
Yep, looks fine to me too.

fiatjaf:
I hate emojis.

Issue #624, opened 2023-06-26
Extending kind:0 - Mark bots.


IngwiePhoenix:
Hello there!

This is a very, very short proposal/request. In the past week or so, I have gotten replies by accounts that I suspect to be AIs/Bots (possibly self-hosted LLMs or OpenAI API clients) like Cyborg or NakomotoX and potentially one or two more.

Since bots are helpful but still automated, I would like to request an extension to the metadata set with `kind:0` to indicate a bot.

In [this section](https://open.spotify.com/intl-de/track/3PVOtxnl44LA8cIq9nkbxO?si=9045c50691a44c26), I would like to extend the metadata by a very simple bool.

For instance:

```json
{
  name: <username>, 
  about: <string>, 
  picture: <url, string>,
  bot: <bool>
}
```

This could then very easily be used by clients when displaying profiles to tell if an account is a bot - or AI - or not.

It is backwards compatible as the field does nothing but add information, whilst it is also voluntary.

Thoughts on this?

Kind regards,
Ingwie

alexgleason:
Seems fine to me. Mastodon lets users self-indicate whether they are a bot.

melvincarvalho:
@alexgleason how's it done in mastodon these days?  Since forever in linked data it's been `@type : Person` or `@type = Agent`, which comes from FOAF

alexgleason:
I believe a bot has actor type `Service` in ActivityPub, but in the Mastodon API it's just a boolean `bot` field.

melvincarvalho:
@alexgleason good to know!  @type is a bit more fundamental than any field.  So bot = true could work well.  And in the not so distant future we'll have online presences which may be part human, part bot.  So it could be a useful toggle, in any case.  Good to align terms with anything existing...

decentropy:
All events have tags, per NIP-01, including kind 0

I use g-tags on kind 0.
I wonder when kind 0 should use tags vs content.

In this case (a boolean), tag seems more appropriate

basantagoswami:
We probably also need to be able to indicate whether a profile is a brand profile.

SilberWitch:
I have the same issue, but for RSS-feedbots/newsbots.

PR #641, opened 2023-07-06
Add a naming convention of object keys for kind 0

akiomik:
Some clients use both of snake case and camel case keys in `kind 0`, such as `display_name` and `displayName`, causing confusion. Many clients conventionally use snake case keys, so make it explicit.

### DIFF
--- a/01.md
+++ b/01.md
@@ -98,7 +98,7 @@ This NIP defines no rules for how `NOTICE` messages should be sent or treated.
 
 ## Basic Event Kinds
 
-  - `0`: `set_metadata`: the `content` is set to a stringified JSON object `{name: <username>, about: <string>, picture: <url, string>}` describing the user who created the event. A relay may delete past `set_metadata` events once it gets a new one for the same pubkey.
+  - `0`: `set_metadata`: the `content` is set to a stringified JSON object `{name: <username>, about: <string>, picture: <url, string>}` describing the user who created the event. `snake_case` keys should be used for this object. A relay may delete past `set_metadata` events once it gets a new one for the same pubkey.
   - `1`: `text_note`: the `content` is set to the **plaintext** content of a note (anything the user wants to say). Content that must be parsed, such as Markdown and HTML, should not be used. Clients should also not parse content as those.
   - `2`: `recommend_server`: the `content` is set to the URL (e.g., `wss://somerelay.com`) of a relay the event creator wants to recommend to its followers.
 


fiatjaf:
Instead of this, I think we need another NIP describing all extra attributes that kind 0 supports. What do we have for now?

- `lud06`
- `lud16`
- `display_name`

arthurfranca:
✔️ banner, website
❌️ nip05valid, npub, followersCount, iris

fiatjaf:
Yes, ahahah, those are just bugs from Iris.

IngwiePhoenix:
I had pointed this out in an issue a while back, in a way. #624 

It'd be nice to have a list of "Should use" and "Could use" to denote attributes that should be expected, whilst there may also be attributes that could be expected - alongside whatever other clients pack in there. But a formal list of known keys would possibly be very helpful.

melvincarvalho:
My list from a large crawl:

```
following,
relay,
name,
about,
picture,
lud06,
lud16,
website,
display_name,
nip05,
banner,
username,
badges,
displayName,
reactions,
nip05valid,
identities,
webOfTrustScore
```

One that I have a need for is:

```
subkey (this will be nostr sub keys associated with an account similar to gpg master / sub keys
ed25519 (for things like git)
```

Note display name is there twice, but camel case is more common.  I think there should also be a convention for tags on dashes vs underscores.  A disadvantage of a dash is that it's a minus sign in some languages.  And created_at is already an underscore, probably best to standardize around _ then.

Anyone can add a field to kind=0.  We could document what's in use, I have a vocab here:

https://w3id.org/nostr

fiatjaf:
You should create a new kind instead of bloating kind 0 even more.

melvincarvalho:
Makes sense, but it also opens up more questions, and two ways of doing things.

So, you're suggesting two profile types: frequent and infrequent. Who determines the content of each? Some might prefer using the frequent one. 

Perhaps letting users choose and utilizing caching could address this.

Egge7:
> Instead of this, I think we need another NIP describing all extra attributes that kind 0 supports, and we mention this convention there. What do we have for now?


Agreed. We mainly use:
- lud06
- lud16
- displayName
- nip05
- about

mattn:
Is this closable by https://github.com/nostr-protocol/nips/pull/794 ?

fiatjaf:
Yes, please take a look there @Egge7, I forgot to mention you there!

PR #659, opened 2023-07-12
Editable Social Notes

### LINKED ISSUE #646: Editing posts
Parameterized replaceable events will allow us to have text notes of kind `30000`. The semantics would be mostly the same as kind 1 (although a good opportunity to clean up anything).

The client would generate a UUID for the post (and preferably encode it to a nicer format such as base62) and add it to the "d" tag. Then notes would be looked up with the author in the pathname, eg `/@:nip05/:d`.

arthurfranca:
Read [here](https://github.com/arthurfranca/nips/blob/social-notes/99.md)

Fix #646

### DIFF
--- a/99.md
+++ b/99.md
@@ -1,0 +1,104 @@
+NIP-99
+======
+
+Social Note
+-----------
+
+`draft` `optional` `author:arthurfranca`
+
+Social notes are kind `31111` events. They accept plain text as `.content` just like kind `1` but are editable.
+
+As with any parameterized replaceable event, `clients` MUST use an `a` tag instead of `e` to reference these events for reposts,
+replies and such.
+
+`Relays` MUST sort and query these events by the publication date (stored at the `d` tag) instead of by `created_at`
+when the request filter includes `since`, `until` or `limit`.
+
+`Relays` MUST validate that the publication date is less than or equal to the `created_at` value.
+
+## Motivation
+
+To make editable social notes that are easier to write than NIP-23, for not accepting content with Markdown syntax,
+while avoiding the problem of using `created_at` field when querying or sorting by date.
+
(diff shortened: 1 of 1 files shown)


arthurfranca:
It seems most people aren't interested on editable social notes, so it isn't worth the trouble.

Though the "Publication Date" section may be useful for the [parameterized replaceable event sort order problem](https://github.com/nostr-protocol/nips/issues/663).

Also the "Text Formatting" section may be useful for using ` instead of * and _ from markdown (which some clients use) because [https://stackoverflow.com/questions/tagged/php*](https://stackoverflow.com/questions/tagged/php*) is a valid url but you can't for instance do *This Italic Test https://stackoverflow.com/questions/tagged/php** correctly or even _https://example.com/.well-known/nostr.json?name=__

alexgleason:
I'm definitely interested. I'm porting all my software from Mastodon to Nostr, and if I have to sacrifice features in the move it will make Nostr less appealing to users.

fiatjaf:
This would be a major blow that would remove elegance and simplicity from Nostr and make it more centralized and censurable. Please don't do this.

alexgleason:
One step at a time. Someone will figure out a way to do this that doesn't break everything. I'm in no rush, we're just brainstorming.

jb55:
let's not fragment kind1 just to have editing.

arthurfranca:
I agree paremetrized replaceable event isn't the answer considering kind 1 is already widespread, though it was interesting seeing how it could look like. Diff events referencing kind 1s should be the way.

nickjuntilla:
Why not just have an "update" kind and then reference the original? Then clients can choose to show the most recent update and history?

PR #638, opened 2023-07-05
NIP-XXX: new `order` event based on BOLT12 payment protocol

ariard:
This NIP proposal introduces a new event `order` of kind `32_500` based on the BOLT12 protocol aiming to enhance peer-to-peer marketplaces.

The [BOLT12](https://bolt12.org/bolt12.html) payment protocol indicates a serie of steps for one or more trading counterparties to enter into a trade agreement (`offer`, `invoice_request`, `invoice`). The new invoice format is TLV-based which enables easy extensibility (e.g adding new payment method like e-cash or issuer-related conditional-payment policy).

The new `order` event is a superset of a bolt12 and the corresponding NIP01 event field. Introducing a new kind rather than adding the offer as a tag introduces the ability to have order policy enforced by a serie of tags. A default order policy is re-using the "tag" `d` to enable an offer issuer or order broker to replace a subset of its pending orders.

The "Design" section gives more examples of order routing policy based on yet-to-be-defined in future NIP tags.

While there is some redundancy with NIP-15, the current proposal not only enable the two major types of real-world marketplace OTC and exchange by allowing the Nostr relays/services to act as broker, or marketplace boards. A "stall" can be re-introduced as a set of offers all announced under the same issuer pubkey, and with a tag to announce the brokering relays to listen for future order events. The `t` tag is not re-used as a scalable and trust-minimized overlay namespace over product/good announcement is far from simple in my understanding.

BOLT12 comes with additional advantages over BOLT11, e.g the Merkle tree construction of the signature invoice format that enables privacy-preserving broker/dealer policy.

There is a WIP implementation against the CivKit node, a Rust based NIP-01 relay software:
https://github.com/civkit/civkit-node/pull/20

Opening this NIP proposal to collect feedback and looking forward on how to standardize and harmonize peer-to-peer marketplace softwares (NIP-15 based and others) across the Bitcoin/Nostr ecosystems.

This being my first NIP proposal, if I’m missing some formatting things about NIPs, or previous discussions about p2p marketplace design in the Nostr space, thanks to let me know.

PR #662, opened 2023-07-13
NIP 99: Classified Listings 

erskingardner:
This adds a new new event `kind:30402` for generic classifieds.

An implementation of a Nostr job board using this new event: https://ostrich.work/

### DIFF
--- a/402.md
+++ b/402.md
@@ -36,7 +36,7 @@ The following tags, used for structured metadata, are standardized and SHOULD be
 - `"price"`, for the price of the thing being listed. This is an array in the format `[ "price", "<number>", "<currency>", "<frequency>" ]`.
   - `"price"` is the name of the tag
   - `"<number>"` is the amount in numeric format (but included in the tag as a string)
-  - `"<currency>"` is the currency unit in 3-character ISO 4217 format.
+  - `"<currency>"` is the currency unit in 3-character ISO 4217 format or ISO 4217-like currency code (e.g. 'btc', 'eth').
   - `"<frequency>"` is optional and can be used to describe recurring payments. SHOULD be in noun format (hour, day, week, month, year, etc.)
 
 #### `price` examples
--- a/402.md
+++ b/402.md
@@ -37,13 +37,13 @@ The following tags, used for structured metadata, are standardized and SHOULD be
   - `"price"` is the name of the tag
   - `"<number>"` is the amount in numeric format (but included in the tag as a string)
   - `"<currency>"` is the currency unit in 3-character ISO 4217 format.
-  - `"<frequency>"` is optional and can be used to describe recurring payments.
+  - `"<frequency>"` is optional and can be used to describe recurring payments. SHOULD be in noun format (hour, day, week, month, year, etc.)
 
 #### `price` examples
 
-- $50 one-time payment `["price", "50", "USD", ""]`
-- €15 per month `["price", "15", "EUR", "monthly"]`
-- £50,000 per year `["price", "50000", "GBP", "yearly"]`
+- $50 one-time payment `["price", "50", "USD"]`
+- €15 per month `["price", "15", "EUR", "month"]`
+- £50,000 per year `["price", "50000", "GBP", "year"]`
 
 Other standard tags that might be useful.
 
--- a/402.md
+++ b/402.md
@@ -1,0 +1,71 @@
+# NIP-402
+
+## Classifieds
+
+`draft` `optional` `author:erskingardner`
+
+This NIP defines `kind:30402`: a parameterized replaceable event to describe classified listings that list any arbitrary product, service, or other thing for sale or offer and includes enough structured metadata to make them useful.
+
+The category of classifieds includes a very broad range of physical goods, services, work opportunities, rentals, free giveaways, personals, etc. and is distinct from the more strictly structured marketplaces defined in [NIP-15](https://github.com/nostr-protocol/nips/blob/master/15.md) that often sell many units of specific products through very specific channels.
+
+The structure of these events is very similar to [NIP-23](https://github.com/nostr-protocol/nips/blob/master/23.md) long-form content events.
+
+### Draft / Inactive Listings
+
+`kind:30403` has the same structure as `kind:30402` and is used to save draft or inactive classified listings.
+
+### Content
+
+The `.content` field should be a description of what is being offered and by whom. These events should be a string in Markdown syntax.
+
+### Author
+
+The `.pubkey` field of these events are treated as the party creating the listing.
+
+### Metadata
+
+- For "tags"/"hashtags" (i.e. categories or keywords of relevance for the listing) the `"t"` event tag should be used, as per [NIP-12](https://github.com/nostr-protocol/nips/blob/master/12.md).
+
+The following tags, used for structured metadata, are standardized and SHOULD be included. Other tags may be added as necessary.
+
+- `"title"`, a title for the listing
+- `"summary"`, for short tagline or summary for the listing
+- `"published_at"`, for the timestamp (in unix seconds – converted to string) of the first time the listing was published.
+- `"location"`, for the location.
+- `"price"`, for the price of the thing being listed. This is a free-form string. e.g. `€2,000/week` or `$22 per hour` or `$500`
+
+Other standard tags that might be useful.
+
+- `"g"`, a geohash for more precise location
+
+## Example Event
+
+```json
+{
+  "kind": 30402,
+  "created_at": 1675642635,
+  // Markdown content
+  "content": "Lorem [ipsum][nostr:nevent1qqst8cujky046negxgwwm5ynqwn53t8aqjr6afd8g59nfqwxpdhylpcpzamhxue69uhhyetvv9ujuetcv9khqmr99e3k7mg8arnc9] dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.\n\nRead more at nostr:naddr1qqzkjurnw4ksz9thwden5te0wfjkccte9ehx7um5wghx7un8qgs2d90kkcq3nk2jry62dyf50k0h36rhpdtd594my40w9pkal876jxgrqsqqqa28pccpzu.",
+  "tags": [
+    ["d", "lorem-ipsum"],
+    ["title", "Lorem Ipsum"],
+    ["published_at", "1296962229"],
+    ["t", "electronics"],
+    ["summary", "More lorem ipsum that is a little more than the title"],
+    ["location", "NYC"],
+    ["price", "$100"],
+    [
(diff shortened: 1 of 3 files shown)


vitorpamplona:
Looks good! Let's go.

utxo-one (commented):
I wonder if it would be better to have price as a float and another tag for currency, it will be hard to build interfaces around arbitrary price strings?

mutatrum:
Never us float for currencies. Use two strings, one for the amount and one for the (ISO) curreny indicator.
String amount
String currency

nikolalukovic:
It looks awesome! I think it would be better to have all prices in sats and a separate optional tag that describes wanted currency so clients can do a live exchange rate. It's easier to standardize everything on sats.

pablof7z:
I think it needs an indexable `c` category tag to be able to differentiate the different applications that might use this for a niche (e.g. job-classified vs used-goods, etc.)

pablof7z:
> 



> The issue is the payment frequency for rentals. "50 USD/week" is communicating 3 things: price, currency, and frequency.
> 
> I would love to put this in sats and just deal with a separate frequency field.

I think this is a good idea as a `frequency` tag would easily differentiate recurring/ongoing listings from one-time ones.

NACK on defaulting to sats; I don't think that's realistic, it should have a "unit"

```json
[ "price", "<number>", "<currency>", "<frequency>" ]
```

where `<frequency>` is optional (off = non-recurring) and both number and currency are mandatory.

vitorpamplona:
Can we add one of more "image" tags for pictures of the product/rental?

vitorpamplona:
+1 for `[ "price", "<number>", "<currency>", "<frequency>" ]`

erskingardner:
Thanks for the feedback everyone! 

> I think it needs an indexable c category tag to be able to differentiate the different applications that might use this for a niche (e.g. job-classified vs used-goods, etc.)

I'm using `"t"` tags/hashtags for this. Is a `"c"` tag better for some reason? 

> NACK on defaulting to sats; I don't think that's realistic, it should have a "unit"

Agree. Sats-only isn't reasonable for now IMO.

> [ "price", "<number>", "<currency>", "<frequency>" ]

I like this. I left it as a free-form field for simplicity but this would make it more useful for sure.

> Can we add one of more "image" tags for pictures of the product/rental?

The `content` field is markdown. I think leaving people to add images to the content is enough, what do you think?

vitorpamplona:
I fail to see the usefulness of separating `c` from `t`. Do we want to make categories to be in a given list of options? 

It feels that Communities can place these events into "Categories" in any way they want to define those categories.

pablof7z:
> I fail to see the usefulness of separating `c` from `t`. Do we want to make categories to be in a given list of options?
> 
> It feels that Communities can place these events into "Categories" in any way they want to define those categories.

categories would allow to make niche classified-ads instead of just craigslist-style

i.e. an etsy website might just use `#c: ["art"]`

a job site might just use `#c: ["job-request", "job-demand"]`

vitorpamplona:
why couldn't they do it with hashtags?

erskingardner:
> ACK on image tags

So would the image tags be in addition to inline images that are in the markdown content?

vitorpamplona:
Yep. The UI renders a little card with image, title, summary and price. When the user opens it, the markdown content kicks in.

402.md
```diff
@@ -0,0 +1,71 @@
...
+
+### Metadata
+
+- For "tags"/"hashtags" (i.e. categories or keywords of relevance for the listing) the `"t"` event tag should be used, as per [NIP-12](https://github.com/nostr-protocol/nips/blob/master/12.md).
+
+The following tags, used for structured metadata, are standardized and SHOULD be included. Other tags may be added as necessary.
+
+- `"title"`, a title for the listing
+- `"summary"`, for short tagline or summary for the listing
+- `"published_at"`, for the timestamp (in unix seconds – converted to string) of the first time the listing was published.
```
ariard, line 33: Redundancy with NIP01’s `created_at` field or usage is expected to be different ?

402.md
```diff
@@ -0,0 +1,71 @@
...
+- `"location"`, for the location.
```
ariard, line 34: Note this field could be optional or move into the content as “virtual goods” might not have a location (e.g a Lightning instant inbound channel). Or is that supposed to be a “social identifier” of the seller ?

402.md
```diff
@@ -0,0 +1,71 @@
...
+The `.pubkey` field of these events are treated as the party creating the listing.
+
+### Metadata
+
+- For "tags"/"hashtags" (i.e. categories or keywords of relevance for the listing) the `"t"` event tag should be used, as per [NIP-12](https://github.com/nostr-protocol/nips/blob/master/12.md).
+
+The following tags, used for structured metadata, are standardized and SHOULD be included. Other tags may be added as necessary.
+
+- `"title"`, a title for the listing
+- `"summary"`, for short tagline or summary for the listing
```
ariard, line 32: I think it’s a good idea to have an abstract to digest the information of the market item proposed for sale, however the distinction between “title” and “summary” is unclear, is the “summary” supposed to be a list of product/service attributes (quantity of items, quality, basket of heterogenous goods sell as “one”) ?

ariard (commented):
I think the idea to have goods/services to sell as NIP33 parameterized replaceable event is a good idea as a seller can replace the set of items in function of market dynamic, and emulate NIP15 “stall” concept thanks to the author field.

On the content itself, the field description of what is sell, I think it could be more flexible than Markdown syntax. Like introducing an enum saying it can be any payment protocol like BOLT11, LNURL, BOLT12 or effectively a dumb markdown syntax. I would say payment / trading protocols are still evolving in the Bitcoin / Lightning space and it can be good for now to give flexibility to marketplace application for experimentation. In the BOLT12 direction, see #638.

On the separation between metadata and content field, I think this is were the current proposal is unclear. From my understanding, there is no clear separation in the Nostr space about what is metadata and goes into “tag” and what is content and more reserved for application.

The [IPv4 packet header format](https://www.rfc-editor.org/rfc/rfc791.html#page-11) dissociates cleanly between information to be consumed by the gateway internet host (e.g version, time to live, identification, length) and the data payload for consumption by upper layers like TCP or UDP.

At the same image of internet packet structuring, it’s unclear if the “tag” should be only for client-to-relay information (e.g NIP40 `created_at` or NIP26 `delegation` ) or client-to-client information (e.g the `location` or `price` tags).

I think the distinction matters for few reasons e.g to avoid “tag” feature redundancy (where one relay have to implement all the tags introduced by a given NIP because they want _one_ feature) and privacy purpose (where metadata only mattering for client shouldn’t be exposed to the relays e.g seller location) or proper versioning of the protocol.

erskingardner:
>Speaking from experience, if this is going to cover products it will need to handle product variants. For lots of product types, it's very important to be able to have a color/size picker on the same product page as all the other variants.

@staab I think the marketplace NIP covers those more complicated uses. This is for more ad hoc sales, not someone doing lots of volume or repeats of the same items. The benefit here is ease and flexibility. 

I'll have a look at the `l` tags too.

402.md
erskingardner: It's not redundancy. NIP-23 long-form articles do the same thing. `published_at` is the date of first publish. Created at is the creation time of the event itself. This is a way of tracking the initial publish time through edits.

erskingardner:
Updated the NIP to reflect new `price` tag format and added `image` tags to the spec.

mutatrum:
Nitpicking on the price array:

Optional can be excluded from the array, I think this is in line with other Nostr arrays:

$50 one-time payment `["price", "50", "USD"]`

I would opt for the noun (month/year/day/hour), instead of adjective (yearly/etc):

€15 per month `["price", "15", "EUR", "month"]`
£50,000 per year `["price", "50000", "GBP", "year"]`

402.md
ariard: >  published_at is the date of first publish. 

If the rationale is it to be used for each event NIP33 replacement it can be added. I can see how it can be used a “beginning” temporal point by the relay to prune one of the replaced instance of the event after a while (e.g “delete my  good announcement after X, whatever the number of replacement”). Rationale could be added.

erskingardner:
@mutatrum Updated the NIP based on your feedback, thanks.

erskingardner:
> and say created_at and title tags can't be altered on edits to be valid

@arthurfranca this breaks how PREs work. You want to them to be editable. How we have this set up now, if you change the title of your listing it's going to create a new job posting.

pablof7z (approved):

arthurfranca:
@erskingardner I forgot about the fact relays use the created_at field to know which PRE to keep. I've created a [PR](https://github.com/nostr-protocol/nips/pull/671) to address it, if merged.

@pablof7z I noticed you approved the changes without considering [this comment](https://github.com/nostr-protocol/nips/pull/662#issuecomment-1638356069) about the "title" problem.

pablof7z:
> @erskingardner I forgot about the fact relays use the created_at field to know which PRE to keep. I've created a [PR](https://github.com/nostr-protocol/nips/pull/671) to address it, if merged.
> 
> 
> 
> @pablof7z I noticed you approved the changes without considering [this comment](https://github.com/nostr-protocol/nips/pull/662#issuecomment-1638356069) about the "title" problem.

The title problem is not specific to this nip (also I don't think it's a problem) so I don't think it should be discussed here nor delay the merge of this NIP

erskingardner:
> Let me explain. The ["d", "<created_at>/<title>"] would be just a convention instead of ["d", "random string"].

@arthurfranca related to the created_at PRE PR you mentioned above, I don't think this is a problem, nor would I want to have this NIP implement something specific for one kind of PRE when the rest don't have the same convention. Changing all other PREs that use titles to use this compound title wouldn't make sense to me either.

arthurfranca:
The ["d", "<created_at>/<title>"] convention would be just for this kind not for all PREs. For NIP-23, for instance, ["d", "<created_at>"] would be enough but that ship has sailed. Other PRE kinds like NIP-51 lists don't need this.

But see at https://habla.news/ how edited content goes to the top of the front-page.

I don't think client-side sorting is a solution because you would have to fetch all posts for sorting, which may be memory intensive or need ever increasing file storage.

But I guess this ship will sail too 🛳️

vitorpamplona:
> I don't think client-side sorting is a solution because you would have to fetch all posts for sorting, which may be memory intensive or need ever increasing file storage.

Some clients will sort by the initial date others will sort by the last change. There is no way for the signing client to know how the reading client will sort by. So, I don't think the `<created_at>/<title>` guidance for the `d` tag does anything.

pablof7z:
@erskingardner can you add the kind to the README.md and change the NIP number to something smaller?

arthurfranca:
Imagine if kind 1 events were PREs. How would global feed be sorted client side?

There would have many kind 1 edits that updates created_at. Some edits may also fake the published_at tag with the current date.

The feed would be full of edited events pretending to be new. This is the future of PRE use cases.

pablof7z:
there are dozens of ways of abusing naive clients; if a user or client decides that a pubkey is misbehaving it can stop listening to it.

You can publish `kind:1` events with a `created_at` one year into the future and pin your note to my feed and I can mute your npub.

arthurfranca:
@pablof7z Events in the future aren't a big problem as they can be easily filtered out (considered invalid) client side.

The said "d" tag convention would similarly be used to easily filter out misbehaving events. People will eventually stop creating events that won't show up.

erskingardner:
@pablof7z NIP number and README updated.

arthurfranca:
🛳️====------ ^^

PR #665, opened 2023-07-14
Adds hashtag, url and event follows to Contact List

vitorpamplona:
This PR updates the guidance of the contact list to account for the new additions by clients. In the last months, we have seen an expansion of the meaning of "contact" as a PubKey, hashtag, URL, event id or `a` tag the user has reviewed and wishes to either follow or simply mark as a reviewed/trusted pointer (to help identify impersonations of a person or community). These additions seem to be accepted as practice by now.

### DIFF
--- a/02.md
+++ b/02.md
@@ -6,7 +6,7 @@ Contact List and Petnames
 
 `final` `optional` `author:fiatjaf` `author:arcbtc`
 
-A special event with kind `3`, meaning "contact list" is defined as having a list of `p`, `t`, `e` or `a` tags, one for each of the profiles, hashtags and other events such as communities or chats one is following.
+A special event with kind `3`, meaning "contact list" is defined as having a list of `p`, `t`, `r`, `e` or `a` tags, one for each of the profiles, hashtags and other events such as communities or chats one is following.
 
 Each `p` tag entry should contain the key for the profile, a relay URL where events from that key can be found (can be set to an empty string if not needed), and a local name (or "petname") for that profile (can also be set to an empty string or not provided), i.e., `["p", <32-bytes hex key>, <main relay URL>, <petname>]`. The `content` can be anything and should be ignored.
 
@@ -20,6 +20,7 @@ For example:
     ["p", "14aeb..8dad4", "wss://bobrelay.com/nostr", "bob"],
     ["p", "612ae..e610f", "ws://carolrelay.com/ws", "carol"], 
 	["t", "intro"] // hashtag
+	["r", "http://mywebsite.com"] // url follows
 	["a", "34550:3f770d65d3a764a9c5cb503ae123e62ec7598ad035d836e2a810f3877a745b24:android", "ws://carolrelay.com/ws"] // Android community
 	["e", "25e5c82273a271cb1a840d0060391a0bf4965cafeb029d5ab55350b418953fbb", "ws://carolrelay.com/ws"] // Nostr Public Chat kind:40 Creation Event
   ],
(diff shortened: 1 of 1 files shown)


staab:
ACK, I think this is a good addition.

mikedilger:
ACK in principle. Maybe we rename it since it's not just contacts anymore.

arthurfranca:
I tried but can't figure out what following a `r` tag (url) does. The rest is nice.

vitorpamplona:
Every post with that url is shown in the feed. It's good for personal page, your company, etc.

arthurfranca:
Oh got it thanks for explaining that's a good idea. Guess it needs clients to add non-media urls from kind 1 `.content` to `r` tags.

vivganes:
Personally, from the perspective of a communities/hashtag client, I prefer separate lists because:

1. Non-supporting clients could erase the bookmarked hashtags when the user tries to follow some new person.   Because of this, the communities client would stop functioning correctly.  It gets very difficult to debug and also to separate noise from the real information at that point.  The probability of some client deleting the hashtags/communities is too large, at the moment.

2.  Adding non-user tags to the contact list affects many aspects of contact-list based clients like the ones below.  Clients _might_ need to change all this code very quickly, or else, nostr _may_ appear broken to new people.
  a. Followers/Following count
  b. Feed display
  c. User profile display
  d. Showing/hiding/blurring note images
  e. Suggested follows (is anyone doing this?)
  
3. The contact list can run into huge numbers potentially.  Any 'read' on contact lists is already less performant because of this large count.  Do we need to add one or more 'filter' operations (with O(n) complexity at least) on top of this?  This could potentially affect the startup/first render time of most clients.

Personally, I prefer the lists demarcated by kind.  This provides clients to pull whatever info they want and operate on top of that.

However, if we all collectively decide to move to this, I am open to that as well. I can see the advantage of all clients fetching followed hashtags/communities from a standard location rather than their own formats.

vitorpamplona:
Having them on contact lists doesn't exclude clients from doing lists. In the end, I believe all clients will need to parse both to have a good/consistent experience.

fiatjaf:
Looks good, although using a separate kind would probably be better just so these lists don't get too big.

What clients are doing this? Are there clients that allow me to follow URLs?

vitorpamplona:
Amethyst was doing `p` and `t` tags in the contact list and we just migrated our local `e` (public chats) and `a` community definition follow lists to the main contact list. I believe Damus just added `t` tags to the Contact List as well.

jb55:
> Are there clients that allow me to follow URLs

neat idea

02.md
```diff
@@ -6,19 +6,25 @@ Contact List and Petnames
...
 {
   "kind": 3,
   "tags": [
     ["p", "91cf9..4e5ca", "wss://alicerelay.com/", "alice"],
     ["p", "14aeb..8dad4", "wss://bobrelay.com/nostr", "bob"],
-    ["p", "612ae..e610f", "ws://carolrelay.com/ws", "carol"]
+    ["p", "612ae..e610f", "ws://carolrelay.com/ws", "carol"], 
+    ["t", "intro"] // hashtag
+    ["r", "http://mywebsite.com"] // url follows
+    ["a", "34550:3f770d65d3a764a9c5cb503ae123e62ec7598ad035d836e2a810f3877a745b24:android", "ws://carolrelay.com/ws"] // Android kind:34550 moderated community
```
501st-alpha1, line 26: Small issue, missing commas at end of these new lines.

fabianfabian:
I just finished implementing support for reading and publishing "t" tags in Contact List in Nostur, but after reading this I thinks it’s better to put it in a separate kind to prevent the lists getting too big.

Right now it’s still possible for clients to be fast without being forced to rely on caching servers.

staab:
> I'd love to see us breaking away from a single contact list at some point. There have been too many race conditions associated with highly-used replaceable events like these.

I second that, and despite the cost of downloading multiple events I do think something like https://github.com/nostr-protocol/nips/pull/349 would be a better model (granular add/remove events work as a CRDT, while lists don't).

vitorpamplona:
Closing this since we moved away to a NIP-51 list for each event type.

PR #670, opened 2023-07-16
Update NIP32 with language example

monlovesmango:
add language example

btw love this NIP!


monlovesmango:
would anyone be open to also adding a 'depth' label as an example?

like an original post would have a depth of 0: ["l", "0", "depth"]
first level reply would have a depth of 1: ["l", "1", "depth"]
second level reply would have a depth of 2: ["l", "2", "depth"]

would allow for querying for direct descendants rather than all descendants (but maybe this has already been solved? been gone awhile...)

monlovesmango:
damn got too excited and misunderstood this nip. closing.

mikedilger:
> been gone awhile...

It is good to see you back here

PR #671, opened 2023-07-18
Add PRE exception to NIP-22

arthurfranca:
This is considering Parameterized Replaceable Events (PREs) can choose to keep the same `created_at` value like [in this example](https://github.com/nostr-protocol/nips/pull/662#issuecomment-1638356069) to mantain the correct sort order.

Such PREs, like in the mentioned example, can commit to a convention of adding the created_at value to the "d" tag so that if values don't match, the event can be considered invalid by clients, just like an event with wrong signature.

### DIFF
--- a/22.md
+++ b/22.md
@@ -7,6 +7,7 @@ Event `created_at` Limits
 `draft` `optional` `author:jeffthibault` `author:Giszmo`
 
 Relays may define both upper and lower limits within which they will consider an event's `created_at` to be acceptable. Both the upper and lower limits MUST be unix timestamps in seconds as defined in [NIP-01](01.md).
+However, [Parameterized Replaceable Events](33.md) should be allowed to have `created_at` value far back in the past so as not to disturb its sort order.
 
 If a relay supports this NIP, the relay SHOULD send the client a [NIP-20](20.md) command result saying the event was not stored for the `created_at` timestamp not being within the permitted limits.
 
--- a/33.md
+++ b/33.md
@@ -13,7 +13,8 @@ Implementation
 The value of a tag can be any string and is defined as the first parameter of a tag after the tag name.
 
 A *parameterized replaceable event* is defined as an event with a kind `30000 <= n < 40000`.
-Upon a parameterized replaceable event with a newer timestamp than the currently known latest
+Upon a parameterized replaceable event with a newer timestamp (informed at the `updated_at` tag, if present,
+of else at the `created_at` field) than the currently known latest
 replaceable event with the same kind, author and first `d` tag value being received, the old event
 SHOULD be discarded, effectively replacing what gets returned when querying for
 `author:kind:d-tag` tuples.


arthurfranca:
Hmm forgot about that.
It could use the "updated_at" tag (which would be needed on the mentioned example).

So updated_at if present or else created_at.

arthurfranca:
@pablof7z edited 33.md to also consider updated_at tag timestamp. I'm trying to make the classifieds NIP free of the sort order problem. If not reusing the same created_at value, one can edit a classified ad just to get to the top of the classified list, if using the default relay sort order based on created_at.

vitorpamplona:
I am not sure why this is a problem. Is anyone having issues with the `created_at` representing the signing date of the current version of a replaceable event and the `published_at` representing the original time?

arthurfranca:
Check https://habla.news/ and see how edited posts go to the top of the front-page.

For start the published_at tag can't be trusted, because the value is not in the "d" tag it can be altered freely without fear of losing referencing likes/zaps/reviews/etc.

Also, I don't think client side filtering is possible when you are trying to sort many events (not just events from me and my friends). For instance, items for sale can't be limited to just items my friends are selling as I want to have access to all available items for sale. How would you sort it client side to get the "Recently Added" considering relays are sending you events sorted as "Recently Updated"?

vitorpamplona:
> For start the published_at tag can't be trusted

Neither the created_at. Most relays don't offer any protections against events in the future or the past. 

> How would you sort it client side to get the "Recently Added" considering relays are sending you events sorted as "Recently Updated"?

I would ask for the first 500 items and then only pick those where published_at is close to created_at. If I need more, I ask for a second batch. 

In the end, singing clients can always publish new `d`'s instead of changing the past if they really want to keep their items up top. There is nothing we can do to stop this.

vitorpamplona:
This is happening right now with Live Activities for instance. The `d` tag is a "channel name" and not really a channel's stream date that would be valid just for the current session. Signing clients update both `created_at` and `starts`/`ends` to match the day.

arthurfranca:
> Neither the created_at. Most relays don't offer any protections against events in the future or the past.
> In the end, singing clients can always publish new d's instead of changing the past if they really want to keep their items up top. There is nothing we can do to stop this.

If there is a convention that clients follow when creating events and when validate received events (dropping invalid by the convention rules), there is nothing earned faking edits.

In contrast, publishing new d tags just to go to the top can't be stopped BUT it would be a brand new event, without replies/zaps/etc so its "popularity" will start from scratch.

arthurfranca:
No one seems to acknowledge there is an important difference between creating a brand new event (new "d" tag) and an edited one.

New events won't have social hints like reactions and replies.
Edited events keep these things.

So to keep these social references, you can't change whatever is inside the "d" tag. So if you have the creation date there (inside the "d" tag), I can be sure it wasn't faked. While the published_at tag can be faked freely.

So currently we can consider most items on a "Recently Added" list will be fake, no matter if you sort server or client side because you are sorting with a freely editable tag (the published_at). Even with NIP-22 as a relay policy, It is just impossible to be sure an event is really new without inferring from reply/reaction count/date.

verbiricha:
> Check https://habla.news/ and see how edited posts go to the top of the front-page.

I don't see a problem with this, I want to show articles that changed at the top even if they were published long ago. For instance, when fiatjaf updates the bounties article I always check out the new grantees. Other clients might decide to only show newly created articles, or to separate "Edits" from "New" articles.

Issue #663, opened 2023-07-13
PRE sort order problem


arthurfranca:
Nostr sort things by latest `created_at`.
Editing a parameterized replaceable event (PRE) refreshes the `created_at` value.

This is alright for some use cases like "github issues", where may be desirable to place updated issues on top.
For other cases like NIP-51 lists it doesn't matter much.

However, for many use cases like NIP-23 (blog posts) and [classifieds PR](https://github.com/nostr-protocol/nips/pull/662), where the client most likely won't fetch just events from user's friends, this attribute presents a problem: PRE authors are incentivized to edit the PRE to make the event go to the top of the list/feed.

[NIP-34 - Algorithmic Filters](https://github.com/nostr-protocol/nips/pull/579) may one day help with sorting by a specifc score (which many relays may not implement). But it doesn't change the fact that the default sort order is broken for some PRE kinds. Custom sorting client side is impossible for the mentioned use cases because of the number of events, thus we need relays help.

What should we do?

I propose a new kind range 40000 <= n < 50000 that works like PREs but with extra rules **enforced by relays**. Basically these events commit to the first `created_at` value by prefixing the `d` tag with that.  More details [here](https://github.com/arthurfranca/nips/blob/social-notes/99.md#publication-date) which is part of the [Editable Social Note draft PR](https://github.com/nostr-protocol/nips/pull/659) (although there I haven't used a kind in that range, which is bad because would need relays to add custom validation for specific kinds as new NIPs get created instead of for a well-defined range of kinds).

Then maybe we move NIP-23 to that range. Thoughts?

alexgleason:
This is an issue for clients to solve, I think. Clients should keep an index of eventId to naddr. When the new event matches an existing index, update it in your own cache and potentially don't display it in a feed.

It is actually a *benefit* that event updates have a new timestamp and gets returned by filters, as it makes it more likely clients will see the update.

Also, I think you should be able to query past versions of parameterized replaceable events for things like edit history, and to make consolidating them on the client easier.

arthurfranca:
A cache to not show again an already seen event is useful to dedupe a feed and show only things you haven't read yet.

But imagine a list of PRE news articles about things happening in your city that is expected to be sorted by latest publication date. You saw some items like 20 and the client cached them like you said.

[...]

PR #674, opened 2023-07-19
Clarify sort order when specified limit

AsaiToshiya:
From reading #579, #663 and the implementation of major relays, it seems that the descending order of the `created_at` is a common understanding, so this PR clarified it.


Egge7:
Just to make it really clear I would rephrase it further:

"When `limit: n` is present it is assumed that the events returned in the initial query will be the last `n` events ordered by `created_at`"

AsaiToshiya:
Thanks for your feedback.

PR #675, opened 2023-07-20
Fix typos about inline code

erechorse:


jb55:
I would merge this but I don't seem to have access anymore

PR #676, opened 2023-07-21
Fix event kind in example in NIP 53

benthecarman:
Everywhere else in the NIP 30311 is used


benthecarman:
@vitorpamplona @v0l

vitorpamplona:
Good catch. 

@fiatjaf simple merge.

PR #677, opened 2023-07-21
Add Rust implementation of NIP-98

utxo-one:


staab:
The link appears to be dead already

PR #679, opened 2023-07-24
Custom Emoji from Relays

eznix86:
Clients can tap from NIP-11 information to get custom emojis

### DIFF
--- a/11.md
+++ b/11.md
@@ -267,6 +267,24 @@ A URL pointing to an image to be used as an icon for the relay. Recommended to b
 }
 ```
 
+
+### Emojis ###
+
+Related to [NIP-30](./30.md), a relay can provide a list of emojis.
+
+```json
+{
+...
+   "emojis": {
+      "soapbox": "https://gleasonator.com/emoji/Gleasonator/soapbox.png",
+      "ablobcatrainbow": "https://gleasonator.com/emoji/blobcat/ablobcatrainbow.png",
+      "disputed": "https://gleasonator.com/emoji/Fun/disputed.png",
+      "gleasonator": "https://gleasonator.com/emoji/Gleasonator/gleasonator.png"
+   }
+...
+}
+```
+
 ### Examples ###
 As of 2 May 2023 the following `curl` command provided these results.
 


alexgleason:
How would this work?

fiatjaf:
I don't know if a relay is the best place to get this information. We could probably have a separate standard that works with any URL, like `https://example.com/.well-known/emojis.json` or something like that. Relays can implement that if they want, but could work with any other website.

alexgleason:
Relays themselves aren't the arbiters of custom emojis anyway. It's up to clients.

@vitorpamplona did a website with custom emoji packs I forgot the URL to, maybe he can share.

vitorpamplona:
It's not mine, but here you go: https://emojis-iota.vercel.app/

vitorpamplona:
Also, I don't understand this PR. Why would a relay advertise emojis and not a pubkey inside the relay?

eznix86:
Thank you for reviewing my PR. My aim with this update is to treat a relay as a community. If you join the "community", you should inherit the emojis associated with it. I understand that maybe the data shouldn't be directly present there and could be in a .well-known instead. However, as NIP-11 is about providing relay information, it should also include the emojis it offers.

vitorpamplona:
>  If you join the "community", you should inherit the emojis associated with it. 

This is very strange. What do we do when people have multiple relays and this is one of them? Do we need to make two posts, one with the emoji for this relay and another one without it for the rest of them? 

Either way, I don't think the direct emoji URLs should be placed in the NIP-11. I would prefer if it added the Ids of the Emoji Packs available for those connecting. Then the user can link that pack to their profile and once they do it, they can use it anywhere.

alexgleason:
On Mastodon each server has its own custom emojis. But an ActivityPub server is not the same as a Nostr relay.

eznix86:
@vitorpamplona 
> This is very strange. What do we do when people have multiple relays and this is one of them? Do we need to make two posts, one with the emoji for this relay and another one without it for the rest of them?

This is what is great with NIP-30, the link is actually inside the note. 
```
["emoji", <shortcode>, <image-url>]
```

> I would prefer if it added the Ids of the Emoji Packs available for those connecting. 

do you think it would be better to have

```json
{
  ...
  "emojis": "https://example.com/.well-known/emojis.json"
  ...
}
```

@alexgleason 
> On Mastodon each server has its own custom emojis. But an ActivityPub server is not the same as a Nostr relay.

I agree it is not the same thing, but this since the NIP already has fields like:
```md
### Community Preferences ###
For public text notes at least, a relay may try to foster a
local community. This would encourage users to follow the global
feed on that relay, in addition to their usual individual follows.
To support this goal, relays MAY specify some of the following values.
```json
{
...
  "language_tags": [ "en", "en-419" ],
  "tags": [ "sfw-only", "bitcoin-only", "anime" ],
  "posting_policy": "https://example.com/posting-policy.html",
...
}
```

I think adding emojis is a bit related to tags, maybe given as mentioned earlier like a link to a emoji pack

PR #680, opened 2023-07-24
NIP-101 HTTP REST API

jacany:
This NIP outlines a REST API to be implemented by Relays. This is being standardized to ensure compatibility between various relay implementations.
This NIP is useful for example if you simply need to publish or receive events from the command line, GitHub Actions, etc where opening a websocket connection is complex.

### DIFF
--- a/101.md
+++ b/101.md
@@ -15,7 +15,7 @@ GET `/api/event`
 Tags should be passed as query paramaters
 `/api/req?ids=id1,id2,id3&authors=pk1&since=0`
 
-Clients should return the events in a JSON array, not dissimilar from how it's done over websockets
+Relays should return the events in a JSON array, not dissimilar from how it's done over websockets
 ```json
 [
     event1,
--- a/101.md
+++ b/101.md
@@ -1,0 +1,42 @@
+NIP-101
+======
+
+Standard HTTP REST API for Relays
+-------------------------------
+
+`draft` `optional` `author:jacany`
+
+This NIP defines the structure for a standard REST API to be implemented by relays. This is to ensure compatibility between different relay implementations and the clients accessing them.
+
+## Reading Events
+
+GET `/api/event`
+
+Tags should be passed as query paramaters
+`/api/req?ids=id1,id2,id3&authors=pk1&since=0`
+
+Clients should return the events in a JSON array, not dissimilar from how it's done over websockets
+```json
+[
+    event1,
+    event2,
+    event3,
+    event4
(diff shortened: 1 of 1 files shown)


arthurfranca:
Maybe you should define a default max number of returned events when there is no "limit" on GET action or make "limit" a requirement.

jacany:
> Maybe you should define a default max number of returned events when there is no "limit" on GET action or make "limit" a requirement.

Maybe 20 by default?

101.md
```diff
@@ -0,0 +1,42 @@
...
+`draft` `optional` `author:jacany`
+
+This NIP defines the structure for a standard REST API to be implemented by relays. This is to ensure compatibility between different relay implementations and the clients accessing them.
+
+## Reading Events
+
+GET `/api/event`
+
+Tags should be passed as query paramaters
+`/api/req?ids=id1,id2,id3&authors=pk1&since=0`
```
AsaiToshiya, line 16:
```suggestion
`/api/event?ids=id1,id2,id3&authors=pk1&since=0`
```

101.md
```diff
@@ -0,0 +1,42 @@
...
+
+Clients should return the events in a JSON array, not dissimilar from how it's done over websockets
```
AsaiToshiya, line 18:
```suggestion
Relays should return the events in a JSON array, not dissimilar from how it's done over websockets
```

101.md
jacany: Any specific reason for this?

AsaiToshiya: > GET `/api/event`

Isn't this a REST API endpoint?

jacany: Yes? Though req is used in the websocket for subscribing unless i'm mistaken...

Egge7:
> Maybe you should define a default max number of returned events when there is no "limit" on GET action or make "limit" a requirement.


And add some key that indicates COUNT by default. This will make sure a client knows how many events fit the filter (good for displaying likes, followers etc) and also that they might need to follow up with pagination requests if they want to get all

fiatjaf:
We should strive to have just one way of doing things and not create apps that depend on this ever. Opening WebSocket connections is not complex, creating one more implementation requirement for all relays is complex.

jacany:
> We should strive to have just one way of doing things and not create apps that depend on this ever. Opening WebSocket connections is not complex, creating one more implementation requirement for all relays is complex.

I agree, thank you for the input.

PR #685, opened 2023-07-27
NIP-47: Nostr Wallet Connect Extensions

benthecarman:
Have gotten requests for things like this from various users. This handles the receive side of nostr wallet connect

### DIFF
--- a/47.md
+++ b/47.md
@@ -36,6 +36,7 @@ The info event should be a replaceable event that is published by the **wallet s
 a plaintext string with the supported commands, space-separated, eg. `pay_invoice get_balance`. Only the `pay_invoice` command is described in this NIP, but other commands might be defined in different NIPs.
 
 Both the request and response events SHOULD contain one `p` tag, containing the public key of the **wallet service** if this is a request, and the public key of the **user** if this is a response. The response event SHOULD contain an `e` tag with the id of the request event it is responding to.
+Optionally, a request can have an `expiration` tag that has a unix timestamp in seconds. If the request is received after this timestamp, it should be ignored.
 
 The content of requests and responses is encrypted with [NIP04](https://github.com/nostr-protocol/nips/blob/master/04.md), and is a JSON-RPCish object with a semi-fixed structure:
 
@@ -108,7 +109,8 @@ Request:
 {
     "method": "pay_invoice",
     "params": {
-        "invoice": "lnbc50n1..." // bolt11 invoice
+        "invoice": "lnbc50n1...", // bolt11 invoice
+        "amount": 123, // invoice amount in msats, optional
     }
 }
 ```
@@ -117,7 +119,7 @@ Response:
 ```jsonc
 {
     "result_type": "pay_invoice",
-    "result": { 
+    "result": {
         "preimage": "0123456789abcdef..." // preimage of the payment
     }
 }
@@ -125,6 +127,278 @@ Response:
 
 Errors:
 - `PAYMENT_FAILED`: The payment failed. This may be due to a timeout, exhausting all routes, insufficient capacity or similar.
+
+### `multi_pay_invoice`
+
+Description: Requests payment of multiple invoices.
+
+Request:
+```jsonc
+{
+    "method": "multi_pay_invoice",
+    "params": {
+        "invoices": [
+          {"id":"4da52c32a1", "invoice": "lnbc1...", "amount": 123}, // bolt11 invoice and amount in msats, amount is optional
+          {"id":"3da52c32a1", "invoice": "lnbc50n1..."},
+        ],
+    }
+}
+```
+
+Response:
+
+For every invoice in the request, a separate response event is sent. To differentiate between the responses, each
+response event contains an `d` tag with the id of the invoice it is responding to, if no id was given, then the
+payment hash of the invoice should be used.
+
+```jsonc
+{
+    "result_type": "multi_pay_invoice",
+    "result": {
+        "preimage": "0123456789abcdef..." // preimage of the payment
+    }
+}
+```
+
+Errors:
+- `PAYMENT_FAILED`: The payment failed. This may be due to a timeout, exhausting all routes, insufficient capacity or similar.
+
+### `pay_keysend`
+
+Request:
+```jsonc
+{
+    "method": "pay_keysend",
+    "params": {
+        "amount": 123, // invoice amount in msats, required
+        "pubkey": "03...", // payee pubkey, required
+        "preimage": "0123456789abcdef...", // preimage of the payment, optional
+        "tlv_records: [ // tlv records, optional
+            {
+                "type": 5482373484, // tlv type
+                "value": "0123456789abcdef" // hex encoded tlv value
+            }
+        ]
+    }
+}
+```
+
+Response:
+```jsonc
+{
+    "result_type": "pay_keysend",
+    "result": {
+        "preimage": "0123456789abcdef...", // preimage of the payment
+    }
+}
+```
+
+Errors:
+- `PAYMENT_FAILED`: The payment failed. This may be due to a timeout, exhausting all routes, insufficient capacity or similar.
+
+### `multi_pay_keysend`
+
+Description: Requests multiple keysend payments.
+
+Has an array of keysends, these follow the same semantics as `pay_keysend`, just done in a batch
+
+Request:
+```jsonc
+{
+    "method": "multi_pay_keysend",
+    "params": {
+        "keysends": [
+          {"id": "4c5b24a351", pubkey": "03...", "amount": 123},
+          {"id": "3da52c32a1", "pubkey": "02...", "amount": 567, "preimage": "abc123..", "tlv_records": [{"type": 696969, "value": "77616c5f6872444873305242454d353736"}]},
+        ],
+    }
+}
+```
+
+Response:
+
+For every keysend in the request, a separate response event is sent. To differentiate between the responses, each
+response event contains an `d` tag with the id of the keysend it is responding to, if no id was given, then the
+pubkey should be used.
+
+```jsonc
+{
+    "result_type": "multi_pay_keysend",
+    "result": {
+        "preimage": "0123456789abcdef..." // preimage of the payment
+    }
+}
+```
+
+Errors:
+- `PAYMENT_FAILED`: The payment failed. This may be due to a timeout, exhausting all routes, insufficient capacity or similar.
+
+### `make_invoice`
+
+Request:
+```jsonc
+{
+    "method": "make_invoice",
+    "params": {
+        "amount": 123, // value in msats
+        "description": "string", // invoice's description, optional
+        "description_hash": "string", // invoice's description hash, optional
+        "expiry": 213 // expiry in seconds from time invoice is created, optional
+    }
+}
+```
+
+Response:
+```jsonc
+{
+    "result_type": "make_invoice",
+    "result": {
+        "type": "incoming", // "incoming" for invoices, "outgoing" for payments
+        "invoice": "string", // encoded invoice, optional
+        "description": "string", // invoice's description, optional
+        "description_hash": "string", // invoice's description hash, optional
+        "preimage": "string", // payment's preimage, optional if unpaid
+        "payment_hash": "string", // Payment hash for the payment
+        "amount": 123, // value in msats
+        "fees_paid": 123, // value in msats
+        "created_at": unixtimestamp, // invoice/payment creation time
+        "expires_at": unixtimestamp, // invoice expiration time, optional if not applicable
+        "metadata": {} // generic metadata that can be used to add things like zap/boostagram details for a payer name/comment/etc.
+    }
+}
+```
+
+### `lookup_invoice`
+
+Request:
+```jsonc
+{
+    "method": "lookup_invoice",
+    "params": {
+        "payment_hash": "31afdf1..", // payment hash of the invoice, one of payment_hash or invoice is required
+        "invoice": "lnbc50n1..." // invoice to lookup
+    }
+}
+```
+
+Response:
+```jsonc
+{
+    "result_type": "lookup_invoice",
+    "result": {
+        "type": "incoming", // "incoming" for invoices, "outgoing" for payments
+        "invoice": "string", // encoded invoice, optional
+        "description": "string", // invoice's description, optional
+        "description_hash": "string", // invoice's description hash, optional
+        "preimage": "string", // payment's preimage, optional if unpaid
+        "payment_hash": "string", // Payment hash for the payment
+        "amount": 123, // value in msats
+        "fees_paid": 123, // value in msats
+        "created_at": unixtimestamp, // invoice/payment creation time
+        "expires_at": unixtimestamp, // invoice expiration time, optional if not applicable
+        "settled_at": unixtimestamp, // invoice/payment settlement time, optional if unpaid
+        "metadata": {} // generic metadata that can be used to add things like zap/boostagram details for a payer name/comment/etc.
+    }
+}
+```
+
+Errors:
+- `NOT_FOUND`: The invoice could not be found by the given parameters.
+
+### `list_transactions`
+
+Lists invoices and payments. If `type` is not specified, both invoices and payments are returned.
+The `from` and `until` parameters are timestamps in seconds since epoch. If `from` is not specified, it defaults to 0.
+If `until` is not specified, it defaults to the current time. Transactions are returned in descending order of creation
+time.
+
+Request:
+```jsonc
+{
+    "method": "list_transactions",
+    "params": {
+        "from": 1693876973, // starting timestamp in seconds since epoch (inclusive), optional
+        "until": 1703225078, // ending timestamp in seconds since epoch (inclusive), optional
+        "limit": 10, // maximum number of invoices to return, optional
+        "offset": 0, // offset of the first invoice to return, optional
+        "unpaid": true, // include unpaid invoices, optional, default false
+        "type": "incoming", // "incoming" for invoices, "outgoing" for payments, undefined for both
+    }
+}
+```
+
+Response:
+```jsonc
+{
+    "result_type": "list_transactions",
+    "result": {
+        "transactions": [
+            {
+               "type": "incoming", // "incoming" for invoices, "outgoing" for payments
+               "invoice": "string", // encoded invoice, optional
+               "description": "string", // invoice's description, optional
+               "description_hash": "string", // invoice's description hash, optional
+               "preimage": "string", // payment's preimage, optional if unpaid
+               "payment_hash": "string", // Payment hash for the payment
+               "amount": 123, // value in msats
+               "fees_paid": 123, // value in msats
+               "created_at": unixtimestamp, // invoice/payment creation time
+               "expires_at": unixtimestamp, // invoice expiration time, optional if not applicable
+               "settled_at": unixtimestamp, // invoice/payment settlement time, optional if unpaid
+               "metadata": {} // generic metadata that can be used to add things like zap/boostagram details for a payer name/comment/etc.
+           }
+        ],
+    },
+}
+```
+
+### `get_balance`
+
+Request:
+```jsonc
+{
+    "method": "get_balance",
+    "params": {
+    }
+}
+```
+
+Response:
+```jsonc
+{
+    "result_type": "get_balance",
+    "result": {
+        "balance": 10000, // user's balance in msats
+    }
+}
+```
+
+### `get_info`
+
+Request:
+```jsonc
+{
+    "method": "get_info",
+    "params": {
+    }
+}
+```
+
+Response:
+```jsonc
+{
+    "result_type": "get_info",
+    "result": {
+            "alias": "string",
+            "color": "hex string",
+            "pubkey": "hex string",
+            "network": "string", // mainnet, testnet, signet, or regtest
+            "block_height": 1,
+            "block_hash": "hex string",
+            "methods": ["pay_invoice", "get_balance", "make_invoice", "lookup_invoice", "list_transactions", "get_info"], // list of supported methods for this connection
+    }
+}
+```
 
 ## Example pay invoice flow
 


fiatjaf:
@vitorpamplona @jb55 @bumi @v0l

vitorpamplona:
I am not sure why you would receive via Wallet Connect instead of the usual lnurlp request from the profile, but sure. 

We can also just update NIP-47 with these new instructions instead of creating a new NIP.

benthecarman:
> I am not sure why you would receive via Wallet Connect instead of the usual lnurlp request from the profile, but sure.

because that requires a web server, this does not

benthecarman:
> We can also just update NIP-47 with these new instructions instead of creating a new NIP.

yeah NIP-47 says they will be defined in other NIPs so i made another

jb55:
On Thu, Jul 27, 2023 at 07:25:53AM -0700, benthecarman wrote:
>On Thu, 27 Jul 2023 07:20:41 -0700, Vitor Pamplona write:
>> I am not sure why you would receive via Wallet Connect instead of the
>> usual lnurlp request from the profile, but sure.
>
>because that requires a web server, this does not

Yes this would be a big deal. lnurl is a centralizing force that we
should eventually phase away from in the zaps spec. Invoice fetching
over nostr would make it way easier to run lightning nodes without
requiring a web server for receiving zap request invoices and sending
zaps.

This has an advantage over bolt12 offers in the sense it can work behind
firewalls, the onion part of bolt12 could in theory do that too somehow
by proxying through nodes on the internet, but it was never clear to me
how that would work or what the state of it is in the current bolt12
spec.

> 
> https://github.com/nostr-protocol/nips/pull/685
>

vitorpamplona:
> yeah NIP-47 says they will be defined in other NIPs so i made another

I think we meant in new PRs. At the time, it looked like each PR with something new was a new NIP.

v0l:
How can we allow anybody to fetch an invoice though, NWC connection is secret, this would only allow the owner of the wallet to generate an invoice?

jb55:
On Thu, Jul 27, 2023 at 02:14:33PM -0700, Kieran wrote:
>How can we allow anybody to fetch an invoice though, NWC connection is
>secret, this would only allow the owner of the wallet to generate an
>invoice?

yes it doesn't seem like it fits exactly into the NWC model. For this
particular RPC it would have to be ok with encrypted events from any id
instead of the one in the connection string, so that it could send an
encrypted reply to that sender.

does it make sense to wrap this into the NWC nip? It seems standalone
and has different security properties.

benthecarman:
> How can we allow anybody to fetch an invoice though, NWC connection is secret, this would only allow the owner of the wallet to generate an invoice?

you can have multiple connections for different use cases, NIP-47 already defines the info event to list out all the available functions

> does it make sense to wrap this into the NWC nip? It seems standalone
and has different security properties.

I think it makes sense to keep together, no need to re-invent the wheel, all these permissions issues should be handled client side validating the requests.

rolznz:
> > How can we allow anybody to fetch an invoice though, NWC connection is secret, this would only allow the owner of the wallet to generate an invoice?
> 
> you can have multiple connections for different use cases, NIP-47 already defines the info event to list out all the available functions

So you would create a new NWC connection that only has permissions for get_invoice, then give out the private key for it to those users who need to create an invoice? This is quite interesting, as long as users make sure to set the right permissions otherwise their wallet could be drained.

For example you could add the private key of the NWC connection in your nostr profile metadata and then you could receive lightning payments without needing a lightning address? (and then some further thought is needed on how to do zaps over NWC)

benthecarman:
> For example you could add the private key of the NWC connection in your nostr profile metadata and then you could receive lightning payments without needing a lightning address? (and then some further thought is needed on how to do zaps over NWC)

Potentially, just want a way to request invoices over nostr

vitorpamplona:
Maybe this is a new nip indeed. Requesting invoices shouldn't require authorization. The need looks more like a data vending machine that replies when a key is pinged than a nwc action.

benthecarman:
> Maybe this is a new nip indeed. Requesting invoices shouldn't require authorization. The need looks more like a data vending machine that replies when a key is pinged than a nwc action.

you guys bike shedding so much more than needed. the wallet connect functionality should have the ability to receive, that is all this does

benthecarman:
Updated to just add to NIP-47

vitorpamplona:
> you guys bike shedding so much more than needed. the wallet connect functionality should have the ability to receive, that is all this does

All we want is actual use. The NWC **could** do a whole lot of things. But we later realized all people really wanted was a way to pay without leaving the app. Not even the wallet status event was actually needed. 

The NWC is a local setup. Only the logged-in user will be able to call his own wallet connect service to get an invoice from that wallet. As it stands, the change alone is not a replacement for the public lnurlp resolution to ask for an invoice from somebody else.

v0l:
A simple solution to enable everybody to request invoices is to use the public key of the secret key issued by the NWC service.

Same flow as before except you include an extra `p` tag

benthecarman:
@v0l why would you do that? That reduces privacy because it is no longer encrypted and is no longer compatible with NWC

v0l:
> @v0l why would you do that? That reduces privacy because it is no longer encrypted and is no longer compatible with NWC

I never said it wasnt encrypted

benthecarman:
what you posted is in raw json

v0l (approved):
There is no reason why this can't be merged already right?

vitorpamplona:
I still don't see this working in the way people are discussing here, but if we have an implementation that is using it, it should be merged.

jb55:
so from what I understand, this is not an lnurl replacement, as that would need a different spec. this is just for fetching invoice from your own node?

bumi:
A few thoughts on the naming: 

`get_invoice` sounds to me a bit like it is looking up an invoice (similar to `invoice_status`)
I would also love to stick to existing namings e.g. `make_invoice` (which is for example used in webln) 

so my proposal would be 
+ `make_invoice` to request a bolt11 invoice and 
+ `lookup_invoice` to get the invoice status

for some apps it would also be great to lookup an invoice based on the bolt11 invoice otherwise those apps would need to parse the bolt11 invoice and get the payment hash first (which in some environments is annoying) 

generally I am +1 for this and I am with @benthecarman. (Going forward I could even imagine adding more wallet calls here.)

benthecarman:
@bumi good suggestions, added

I have implemented this in https://github.com/benthecarman/nostr-wallet-connect-lnd

supertestnet:
> need 2 implementations

ok

looks like a second implementation is on its way, hopefully it is here soon

https://github.com/getAlby/nostr-wallet-connect/issues/112#issuecomment-1668903161

bumi:
Yes, we're working on adding it to our NWC implementation. Which will be available on the hosted version and on the self-hosted version (incl. in the Umbrel store)

47.md
```diff
@@ -63,6 +63,92 @@ Response:
...
+
+### Lookup Invoice
+
+Request:
+```jsonc
+{
+    "method": "lookup_invoice",
+    "params": {
+        "payment_hash": "31afdf1..", // payment hash of the invoice, one of payment_hash or invoice is required
+        "invoice": "lnbc50n1..." // invoice to lookup
```
rolznz, line 132: What is the reason why we can't always just use payment_hash to lookup the invoice?

benthecarman: from @bumi 

> for some apps it would also be great to lookup an invoice based on the bolt11 invoice otherwise those apps would need to parse the bolt11 invoice and get the payment hash first (which in some environments is annoying)

you are supposed to be able to use either or

rolznz: Yes, but `make_invoice` returns the payment_hash. And we would only lookup the invoices created using `make_invoice`?

benthecarman: Possible you could have an application that doesn't call `make_invoice`

rolznz:
The Alby implementation has been updated to support the new methods. Thanks @benthecarman

rolznz (approved):

benthecarman:
https://github.com/getAlby/nostr-wallet-connect now implements this, good to merge?

47.md
```diff
@@ -63,6 +63,92 @@ Response:
...
+```jsonc
+{
+    "result_type": "lookup_invoice",
+    "error": {
+        "code": "UNAUTHORIZED",
+        "message": "human readable error message"
+    },
+    "result": {
+        "invoice": "lnbc50n1...", // the encoded invoice
+        "paid": "true" // true if the invoice has been paid
```
verbiricha, line 147: nit: this should be a boolean

```suggestion
        "paid": true // true if the invoice has been paid
```

vitorpamplona (approved):

bumi:
One thing that keeps popping up: do we also want a call to receive the transactions? That would probably complete these additions. 
What are your thoughts on this?

bumi:
> For onchain?

no, to get the lightning transaction history. We have the option now to get the account balance, create invoices, get an invoice status. So one more thing would be to also get the transaction history of the connected account.

benthecarman:
Yeah I guess that makes sense. We could just send a list of invoices, filtering by date and if paid

benthecarman:
@bumi added a `list_invoices` command, lmk what you think

rolznz:
@benthecarman `list_invoices` looks good to me. I will update here when we have implemented or have any questions.

Now we can list invoices a user has received, but it would be great to also list payments a user has sent. Should we add a `list_payments` method too?

Alby already has implemented a `get_balance` RPC call which will return a simple result `{ "balance": 10000 }` (in msat). What do you think about this, could it be added to this PR?

benthecarman:
@rolznz added!

rolznz:
@benthecarman thanks!

It would be great if we can have a `list_payments` call. It seems like it could be similar or the same as `list_invoices`. Does that make sense to you?

benthecarman:
@rolznz added!

TonyGiorgio:
How about a keysend method? That would enable wallets to be the funding source for a lot of value for value podcast stuff. 

Maybe something like this?
---

### Pay Keysend

Request:
```jsonc
{
    "method": "pay_keysend",
    "params": {
        "amount": 123, // invoice amount in msats, required
        "pubkey": "03...", // payee pubkey, required
        "message": "some message", // keysend message, optional
    }
}
```

Response:
```jsonc
{
    "result_type": "pay_keysend",
    "error": {
        "code": "UNAUTHORIZED",
        "message": "human readable error message"
    },
    "result": { // result, object. null in case of error.
        "payment_hash": "31afdf1.."
    }
}
```

benthecarman:
@TonyGiorgio good suggestion added!

benthecarman:
I think this PR is getting pretty big at this point. We should have enough functionality now for any kind of wallet.

Good to merge? @bumi @rolznz

rolznz:
@benthecarman we haven't implemented the new methods yet, if possible it would be great to do this first in case we notice any changes that need to be made.

rolznz:
Regarding the keysend method, should we add some extra parameters to better support podcasting? for example, I think we should in some way support [BLIP-10](https://github.com/lightning/blips/blob/master/blip-0010.md)

benthecarman:
@rolznz good call, added `tlv_records` to the keysend so that could be done

benthecarman:
Implemented the remaining new functions into rust-nostr: https://github.com/rust-nostr/nostr/pull/171

ok300:
Is there a reason why `make_invoice` doesn't have an optional `preimage` arg?

IMO it would be useful to allow creating invoices with chosen preimage.

benthecarman:
> Is there a reason why `make_invoice` doesn't have an optional `preimage` arg?
> 
> IMO it would be useful to allow creating invoices with chosen preimage.

good idea, added

bumi:
I think we're pretty much good to go here and get this merged, implementations have been mostly working.

But I actually have some suggestions on `list_payments` and `list_invoice`. 
Imo the data that they currently return is not enough. We should try to find some simple format to describe a the payment and not only return a bolt11 invoice. Having only a bolt11 invoice there means that clients would need to decode that invoice to get any useful information which adds complexity and also would not work with keysend payments for example.

We can return more invoice payment data like: 

```
{
   memo: "string", 
   preimage: "string",
   payment_hash: "string",
   description_hash: "string",
   amount: 123, // value in msats
   settled_at: unixtimestamp,
   metadata: {} // generic metadata hash that can be used to add zap/boostagram details like a payer name/comment/etc.
}
```

The NWC app has this information anyway and if I think as a Nostr client that wants to add such payment information it ise more useful than only the bolt11 invoice. 
Additionally we could think about a NIP to publish encrypted transaction data as an event to allow subscriptions on payments through the relay - somewhere this was already discussed. 
What is your take on this?

benthecarman:
yeah that makes sense to me

> Additionally we could think about a NIP to publish encrypted transaction data as an event to allow subscriptions on payments through the relay - somewhere this was already discussed.

think this is separate from this nip. however, i dont think that method will scale at all, if i have a million subs doing all the encrypted events will take forever to sign

47.md
```diff
@@ -63,6 +63,238 @@ Response:
...
+{
+    "result_type": "list_invoices",
+    "error": {
+        "code": "UNAUTHORIZED",
+        "message": "human readable error message"
+    },
+    "result": [
+        {
+           invoice: "string", // encoded invoice
+           payee_pubkey: "string", // pubkey of the payee, used for keysend
```
rolznz, line 217: does payee_pubkey make sense here? or only in `list_payments`?

47.md
```diff
@@ -63,6 +63,238 @@ Response:
...
+           memo: "string",
```
rolznz, line 218: can this be called `description` to be consistent with `make_invoice`?

47.md
```diff
@@ -63,6 +63,238 @@ Response:
...
+
+Request:
+```jsonc
+{
+    "method": "list_payments",
+    "params": {
+        "from": 1693876973, // starting timestamp in seconds since epoch, optional
+        "until": 1693876973, // ending timestamp in seconds since epoch, optional
+        "limit": 10, // maximum number of invoices to return, optional
+        "offset": 0, // offset of the first invoice to return, optional
```
rolznz, line 241: would it make sense to have a param here to include unsettled payments, similar to `unpaid` in `list_invoices`?

rolznz:
I would like to propose one final additional method - what do you guys think?

### Get Info

Returns information about the user's node

Request:

```jsonc
{
  "method": "get_info",
  "params": {}
}
```

Response:

```jsonc
{
  "result_type": "get_info",
  "error": {
    "code": "UNAUTHORIZED",
    "message": "human readable error message"
  },
  "result": {
    "alias": "string",
    "pubkey": "string",
    "color": "string"
  }
}
```

vitorpamplona:
I really wish the wallet service provided information in the kind 13194 instead of a request/response. The Nostr way is to simply subscribe to that event and the client will receive all updates when they happen. There is no need to make a separate request for that information. 

It's similar to `list_payments` and `list_invoices`. We already have a filter API in Nostr. There is no need to recreate them as RPC calls.

47.md
benthecarman: what would an unpaid payment be? only thing that'd make sense is failed or in progress but dont think that really fits here

benthecarman:
@vitorpamplona i disagree, you shouldn't be pushing everytime you send or receive a payment happens to nostr, that is absolutely awful for privacy.

vitorpamplona:
So, you think calling this method over and over again and just watching for the changes in the reply is much better for privacy? 

Those two things are not connected. You can make both rpc and nostr event updates equally private.

47.md
benthecarman: good call

benthecarman:
Added the `get_info` event and incorporated suggestions on the list_x calls

47.md
rolznz: Yes not `unpaid` but a similar idea, as you said, failed and in-progress payments (for example HOLD invoices).

Why does it not fit here? would it make sense as a future improvement?

rolznz:
I think consistency of the interaction with the wallet is also good to consider. It is nice to have a single mechanism (RPC) for all methods as someone who is going to use NWC in their application.

I can see the advantage of subscribing and receiving instant updates rather than having to poll, although this might not always be easy depending on what wallet is behind the NWC connection.

47.md
benthecarman: lets just do that later, think this has plenty of functionality

47.md
```diff
@@ -63,6 +63,269 @@ Response:
...
+    "method": "get_info",
+    "params": {
+    }
+}
+```
+
+Response:
+```jsonc
+{
+    "result_type": "get_balance",
```
rolznz, line 311: this should be "get_info"

benthecarman: woops, fixed

pablof7z:
what's the status on this? can we push it pass the finish line? replacing lnurl with this is such an obvious low-hanging fruit to improve so many things!

benthecarman:
I think we're just waiting on the alby guys to finish their implementation

47.md
```diff
@@ -63,6 +63,269 @@ Response:
...
+
+Response:
+```jsonc
+{
+    "result_type": "get_info",
+    "error": {
+        "code": "UNAUTHORIZED",
+        "message": "human readable error message"
+    },
+    "result": [
```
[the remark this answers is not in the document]
benthecarman: fixed

47.md
```diff
@@ -63,6 +63,269 @@ Response:
...
+           "preimage": "string",
+           "payment_hash": "string",
+           "amount": 123, // value in msats
+           "fees_paid": 123, // value in msats
+           "settled_at": unixtimestamp,
+           "metadata": {} // generic metadata that can be used to add things like zap/boostagram details for a payer name/comment/etc.
+        }
+    ],
+}
+```
```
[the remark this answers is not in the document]
benthecarman: The original intention was to be able to create invoices over NWC. That way NWC can be used as a replacement for things like LNURL in the future.

The alby guys wanted to add full functionality to be able to get all other payment information so we added list payments and invoices.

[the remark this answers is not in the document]
supertestnet: > > The original intention was to be able to create invoices over NWC. That way NWC can be used as a replacement for things like LNURL in the future.
> 
> This makes no sense to me, like you would need a separate protocol anyway for zaps that fetch invoices over Nostr since NWC is meant for mostly spending.

Adding a `get_invoice()` command to NWC seems simpler than creating a whole new protocol to do the same thing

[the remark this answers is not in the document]
benthecarman: > This makes no sense to me, like you would need a separate protocol anyway for zaps that fetch invoices over Nostr since NWC is meant for mostly spending.

We discussed this at the beginning in the comments. It makes no sense to connect your wallet but not be able to receive.

benthecarman:
@rolznz any updates here, would like to get this merged

rolznz:
> @rolznz any updates here, would like to get this merged

@benthecarman sorry for the delay, we've had to prioritize some other work and were away at conferences. I'll be working on finishing the implementation in the next couple of weeks :rocket:

47.md
```diff
@@ -63,6 +63,267 @@ Response:
...
+### Keysend
+
+Request:
+```jsonc
+{
+    "method": "pay_keysend",
+    "params": {
+        "amount": 123, // invoice amount in msats, required
+        "pubkey": "03...", // payee pubkey, required
+        "message": "some message", // keysend message, optional
```
bumi, line 103: this will actually also be part of a tlv record, isn't it? 
do we need this as a special key here?

47.md
```diff
@@ -63,6 +63,267 @@ Response:
...
+```jsonc
+{
+    "result_type": "pay_keysend",
+    "error": {
+        "code": "UNAUTHORIZED",
+        "message": "human readable error message"
+    },
+    "result": { // result, object. null in case of error.
+        "preimage": "0123456789abcdef...", // preimage of the payment
+        "payment_hash": "31afdf1..", // payment hash of the payment
```
bumi, line 125: if we return the payment_hash here should `pay_invoice` also return the payment hash? 
just thinking about consistency even though it is not needed there as the user has the invoice anyway.

47.md
```diff
@@ -63,6 +63,267 @@ Response:
...
+```jsonc
+{
+    "result_type": "lookup_invoice",
+    "error": {
+        "code": "UNAUTHORIZED",
+        "message": "human readable error message"
+    },
+    "result": {
+        "invoice": "lnbc50n1...", // the encoded invoice
+        "paid": true // true if the invoice has been paid
```
bumi, line 184: should we return the same information as one entry from the `list_payments` call?
```
{
           "invoice": "string", // encoded invoice
           "payee_pubkey": "string", // pubkey of the payee, used for keysend
           "description": "string",
           "description_hash": "string",
           "preimage": "string",
           "payment_hash": "string",
           "amount": 123, // value in msats
           "fees_paid": 123, // value in msats
           "settled_at": unixtimestamp,
           "metadata": {} // generic metadata that can be used to add things like zap/boostagram details for a payer name/comment/etc.
}
```

47.md
```diff
@@ -63,6 +63,267 @@ Response:
...
+
+### List Payments
+
+Request:
+```jsonc
+{
+    "method": "list_payments",
+    "params": {
+        "from": 1693876973, // starting timestamp in seconds since epoch, optional
+        "until": 1693876973, // ending timestamp in seconds since epoch, optional
```
bumi, line 238: it could be interesting to use an invoice payment hash as a starting point for pagination (e.g. all payments since the one with payment hash x)

bumi (approved):
I have left some minor comments and thoughts that I had while looking at the implementation of this. Alby has this mostly implemented and the next release of the NWC implementation will support most of it. 

I am +1 for merging this.

47.md
```diff
@@ -63,6 +63,267 @@ Response:
...
+
+Response:
+```jsonc
+{
+    "result_type": "list_invoices",
+    "error": {
+        "code": "UNAUTHORIZED",
+        "message": "human readable error message"
+    },
+    "result": [
```
rolznz, line 214: Returning an array directly means we cannot easily extend the response in the future. Would it make sense to return an object e.g.

```json
result: {
  invoices: [],
  // ... other properties here
}
```

LND also does this: https://api.lightning.community/api/lnd/lightning/list-invoices/index.html

47.md
```diff
@@ -63,6 +63,267 @@ Response:
...
+Below are the commands that are defined in this NIP.
+
+### Pay Invoice
+
+Request:
+```jsonc
+{
+    "method": "pay_invoice",
+    "params": {
+        "invoice": "lnbc50n1..." // invoice to pay, required
```
bumi, line 75: many use cases (zaps, boosts, etc.) use split payments and I have heard this requirement quite often. 
could we optionally allow an array in in the `pay_invoice` and `pay_keysend` methods? 
This would reduce the amount of calls that have to be made in those cases. 

As result we would then also return an array

benthecarman: Yeah have heard this requested a lot

47.md
benthecarman: i just removed the `payment_hash` to be consistent, either way it is derivable from the preimage

47.md
benthecarman: did the same for list_payments too

47.md
benthecarman: One problem is if you aren't able to pay all of them, how do you handle whether or not to send an error or success message

benthecarman: I think this would be better as a separate command. Maybe leave for future work as this PR has lots of changes already

rolznz:
@benthecarman after working on list_payments and list_invoices we think it might actually be nice to have a single method that returns a merged list of payments and invoices in one go (called list_transactions).

Reasoning:

I believe one reason NWC has positive adoption by clients already because it is simple for clients to implement. Having one transaction list is simpler and more efficient for the client (no need to manage two different requests to get invoices and payments, and paginate them both). Instead, the complexity is moved to the backend which can handle merging payments and invoices into one list of transactions.

The type of the transaction (an invoice or payment) is identified by the `type` property. Otherwise the transaction just has the combined fields of an invoice/payment, where any inapplicable properties are excluded.

### List Transactions
Request:
```jsonc
{
    "method": "list_transactions",
    "params": {
        "from": 1693876973, // starting timestamp in seconds since epoch, optional
        "until": 1693876973, // ending timestamp in seconds since epoch, optional
        "limit": 10, // maximum number of transactions to return, optional
        "offset": 0, // offset of the first transaction to return, optional
        "unpaid": true // include unpaid transactions, optional, default false
    }
}
```
Response:
```jsonc
{
    "result_type": "list_transactions",
    "error": {
        "code": "UNAUTHORIZED",
        "message": "human readable error message"
    },
    "result": {
        "transactions": [
            {
               "type": "incoming", // "incoming" for invoices, "outgoing" for payments
               "invoice": "string", // encoded invoice
               "description": "string",
               "description_hash": "string",
               "preimage": "string",
               "payment_hash": "string",
               "amount": 123, // value in msats
               "fees_paid": 123, // value in msats
               "settled_at": unixtimestamp,
               "metadata": {}, // generic metadata that can be used to add things like zap/boostagram details for a payer name/comment/etc.
               "payee_pubkey": "string", // pubkey of the payee, used for keysend
           }
        ],
    },
}
```

benthecarman:
@rolznz added

47.md
bumi: ok, you mean doing it in a `multi_pay_invoice` command or something like that? 
would be OK for me.

bumi:
We have apps that rely instantly on a reply e.g with the preimage from the pay_invoice call. But some apps don't care if the payment happens directly or later. 
Some wallets might also not be able to directly execute the command/payment but will only do that when opened the next time. (is this correct?)

should we somehow reflect this in the spec here? For example should it be possible to mark a request as time critical and the wallet has to reply immediately otherwise it can be ignored?

47.md
```diff
@@ -126,6 +126,188 @@ Response:
...
+```jsonc
+{
+    "result_type": "get_info",
+    "result": {
+            "alias": "string",
+            "color": "hexstring",
+            "pubkey": "hex string",
+            "network": "string", // mainnet, testnet, signet, or regtest
+            "block_height": 1,
+            "block_hash": "hex string",
```
rolznz, line 306:
```suggestion
            "block_hash": "hex string",
            "methods": ["make_invoice", "get_balance", ...], // NIP-47 methods supported by this connection
```

benthecarman: added

benthecarman:
> We have apps that rely instantly on a reply e.g with the preimage from the pay_invoice call. But some apps don't care if the payment happens directly or later. Some wallets might also not be able to directly execute the command/payment but will only do that when opened the next time. (is this correct?)
> 
> should we somehow reflect this in the spec here? For example should it be possible to mark a request as time critical and the wallet has to reply immediately otherwise it can be ignored?

This sounds related to an issue we are running into with Mutiny. Currently we have built in custom logic for our relay for NWC. We can't have the events be ephemeral because the user isn't always online and we also need to guarantee the events are only sent to the user once. So what we do is save them and send them to the user if they are requested, we will delete a NWC event after it has gotten a reply from the correct key.

Ideally these weren't ephemeral events and had an expiration time so we could use that to handle it, however, I think it is too late for that.

47.md
```diff
@@ -126,6 +126,189 @@ Response:
...
+
+Request:
+```jsonc
+{
+    "method": "make_invoice",
+    "params": {
+        "amount": 123, // invoice amount in msats, required
+        "description": "some description", // invoice description, optional
+        "description_hash": "31afdf1..", // invoice description hash, optional
+        "preimage": "0123456789abcdef...", // preimage for the payment, optional
```
rolznz, line 169: @benthecarman @bumi I think there are two cases for making invoices:

1. You create a connection string with only make_invoice permissions and give it to someone else. Then they can pay you. (like LNURL)
2. You create a service that has access to a connection string with make_invoice permissions to receive payments on behalf of users of the service.

Is there a good usecase for being able to set a custom preimage? because I think it breaks the first case.

47.md
```diff
@@ -126,6 +126,189 @@ Response:
...
+            {
+               "type": "incoming", // "incoming" for invoices, "outgoing" for payments
+               "invoice": "string", // encoded invoice
+               "description": "string",
+               "description_hash": "string",
+               "preimage": "string",
+               "payment_hash": "string",
+               "amount": 123, // value in msats
+               "fees_paid": 123, // value in msats
+               "settled_at": unixtimestamp,
```
v0l, line 254: missing a `created_at` timestamp

47.md
```diff
@@ -126,6 +126,189 @@ Response:
...
+    "params": {
+    }
+}
+```
+
+Response:
+```jsonc
+{
+    "result_type": "get_balance",
+    "result": [
```
v0l, line 277: I assume this is not supposed to be an array?

47.md
rolznz: Maybe `expires_at` too. (although both could be decoded from the invoice, it is nice to be available without having to decode first)

v0l:
Added support for new methods on Snort/Iris https://git.v0l.io/Kieran/snort/commit/b41e8a919acd839fe45db005dc9b262ad82598d0

47.md
benthecarman: woops, fixed

47.md
benthecarman: Yeah I guess you might not anyone with a NWC connection to be able to create with custom preimages. Ideally we don't remove the functionality and this is enforced client side, however, that makes permissioning harder since normally it just around the commands. We could add another command but that isn't very pretty.

47.md
benthecarman: We have a lot of info that can already be read from decoding the invoice. Probably fine to add these too

47.md
benthecarman: Going to remove for now. Since it is an added field this can be added in the future without breaking backwards compatibility

benthecarman:
Responded to review and added comments on a bunch of the params where things could be unclear

bumi:
> This sounds related to an issue we are running into with Mutiny. Currently we have built in custom logic for our relay for NWC. We can't have the events be ephemeral because the user isn't always online and we also need to guarantee the events are only sent to the user once. So what we do is save them and send them to the user if they are requested, we will delete a NWC event after it has gotten a reply from the correct key.
Ideally these weren't ephemeral events and had an expiration time so we could use that to handle it, however, I think it is too late for that.

Yes, we also implement measures to make sure those events are delivered reliably. 
But currently I see the issue that an app sends the request and waits for a payment confirmation reply. Some payments are time critical and some payments are not. For example a zap or some subscription is probably not time critical and it does not matter if the payment is made now or a few hours later. The app might not even wait for the reply. 
But a payment to get access to something is time critical and it must be paid now, and should no longer be paid hours later. The response with the preimage might be critical here.
In this case you might want to send a notification to the user to confirm the payment now.

So I am wondering if the app should be able to indicate this in the event. 
It could be for example an expiry timestamp. 

When a specialized relay then gets that event and sees a short expiry it might want to send a notification to the user to act.

47.md
rolznz: Sounds good to me, thanks!

47.md
gudnuf: > One problem is if you aren't able to pay all of them, how do you handle whether or not to send an error or success message

There could be one payment request event for 20 invoices, but 20 response events for each individual payment (each tagged with the request event ID, just different preimages/errors in the content). 
Allowing for an array of invoices would mean zap split payments don't have to spam the nwc relays.

47.md
```diff
@@ -126,6 +126,186 @@ Response:
...
+        "expiry": 3600, // invoice expiry in seconds, optional
+    }
+}
+```
+
+Response:
+```jsonc
+{
+    "result_type": "make_invoice",
+    "result": {
```
rolznz, line 178: I think `make_invoice` should return the same response type as `lookup_invoice` (and `list_transactions`) so they are all consistent?

benthecarman: done

47.md
bumi: @gudnuf this is perfect! 
we probably then should just add the payment_hash or the invoice itself to the result? 

@benthecarman what do you think? is that good? 

split payments has been used quite a lot so I think it would be nice to have a decent support here.

benthecarman: The challenge with that is attributing which response corresponds to which invoice. We could tag the event with the payment hash i guess

benthecarman:
Added an optional `expires` tag and `multi_pay_invoice` command

47.md
```diff
@@ -126,6 +127,224 @@ Response:
...
+          "lnbc50n1..." // bolt11 invoice
+        ],
+    }
+}
+```
+
+Response:
+
+For every invoice in the request, a separate response event is sent. To differentiate between the responses, each
+response event contains an `payment_hash` tag with the payment hash of the invoice it is responding to.
```
rolznz, line 149: this mentions `payment_hash` but is not in the response

benthecarman: it is not in the response, it is in a `tag`

47.md
```diff
@@ -126,6 +127,224 @@ Response:
...
+    }
+}
+```
+
+### `make_invoice`
+
+Request:
+```jsonc
+{
+    "method": "make_invoice",
```
rolznz, line 198: @benthecarman regarding
> Yeah I guess you might not anyone with a NWC connection to be able to create with custom preimages

Are we able to remove these fields from the `make_invoice` request for now?
- `invoice`
- `preimage`
- `payment_hash`

rolznz: Update: are some of these fields incorrectly taken from the response?

rolznz:
```json
"params": {
  "amount": 123, // value in msats
   "description": "string", // invoice's description, optional
   "description_hash": "string", // invoice's description hash, optional
   "expiry": 213 // expiry in seconds from time invoice is created
}
```

47.md
rolznz: Please re-update this to match the response of `lookup_invoice`

47.md
benthecarman: Fixed

47.md
rolznz: @benthecarman are you still ok making this change?

benthecarman: I was missing `type`, was there anything else missing?

rolznz: It would be nice for the response to match that of `lookup_invoice` (without properties that aren't applicable like fees_paid, preimage and settled_at) 
e.g.

Response:
```jsonc
{
    "result_type": "make_invoice",
    "result": {
        "type": "incoming",
        "invoice": "string", // encoded invoice
        "description": "string", // invoice's description, optional
        "description_hash": "string", // invoice's description hash, optional
        "payment_hash": "string", // Payment hash for the payment
        "amount": 123, // value in msats
        "created_at": unixtimestamp, // invoice/payment creation time
        "expires_at": unixtimestamp, // invoice expiration time, optional if not applicable
        "metadata": {} // generic metadata that can be used to add things like zap/boostagram details for a payer name/comment/etc.
    }
}

Comment (benthecarman) [reply 47.md:219]: Oh sorry was looking at wrong command, fixed

File: 47.md [hunk 22]
```diff
@@ -126,6 +127,227 @@ Response:
...
+```jsonc
+{
+    "result_type": "pay_keysend",
+    "result": {
+        "preimage": "0123456789abcdef...", // preimage of the payment
+    }
+}
+```
+
+### `make_invoice`
```
rolznz, line 193: @benthecarman you copied the response JSON into the request JSON again. I included the full make_invoice section here:

### `make_invoice`

Request:
```jsonc
{
    "method": "make_invoice",
    "params": {
        "amount": 123, // value in msats
        "description": "string", // invoice's description, optional
        "description_hash": "string", // invoice's description hash, optional
        "expiry": 213 // expiry in seconds from time invoice is created, optional
      }
}
```

Response:
```jsonc
{
    "result_type": "make_invoice",
    "result": {
        "type": "incoming", // "incoming" for invoices, "outgoing" for payments
        "invoice": "string", // encoded invoice, optional
        "description": "string", // invoice's description, optional
        "description_hash": "string", // invoice's description hash, optional
        "preimage": "string", // payment's preimage, optional if unpaid
        "payment_hash": "string", // Payment hash for the payment
        "amount": 123, // value in msats
        "fees_paid": 123, // value in msats
        "created_at": unixtimestamp, // invoice/payment creation time
        "expires_at": unixtimestamp, // invoice expiration time, optional if not applicable
        "metadata": {} // generic metadata that can be used to add things like zap/boostagram details for a payer name/comment/etc.
    }
}
```

benthecarman: sorry my bad, fixed

benthecarman:
Changed the `expires` tag to `expiration` because it seems that is standardized for other applications

47.md
```diff
@@ -126,6 +127,229 @@ Response:
...
+
+Request:
+```jsonc
+{
+    "method": "make_invoice",
+    "result": {
```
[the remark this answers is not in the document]
benthecarman: good catch

47.md
```diff
@@ -126,6 +127,229 @@ Response:
...
+### `multi_pay_invoice`
+
+Description: Requests payment of multiple invoices.
+
+Request:
+```jsonc
+{
+    "method": "multi_pay_invoice",
+    "params": {
+        "invoices": [
```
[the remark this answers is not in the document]
benthecarman: Yeah i guess, eventually people could add an amount or somehting for amountless invoices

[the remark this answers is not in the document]
benthecarman: i guess can add that now

47.md
```diff
@@ -126,6 +127,229 @@ Response:
...
+```
+
+
+### `list_transactions`
+
+Request:
+```jsonc
+{
+    "method": "list_transactions",
+    "params": {
```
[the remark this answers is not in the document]
benthecarman: Added some clarifications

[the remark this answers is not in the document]
rolznz: Looks great, thanks!

Egge21M:
Maybe we can use this PR to also clear up some inconsistencies that I have encountered while working on a NWC focussed project. Those changes mostly concern the documentation and not the protocol itself.

- NIP provides an example URI without a proper authority component ("//" between host and scheme are missing), while implementations in the wild include it. 

- NIP does not have an opinion about URL-unsafe query parameters, which IMHO it should (e.g. taken from an Alby connection string "?relay=wss://..."

Let me know if this should be a separate PR or if you want to include these changes in this proposal.

benthecarman:
@Egge21M seems out of scope for this PR but I definitely think those would be good changes.

Egge21M:
@benthecarman copy that and I am aligned. I'll suggest some clarifications in another PR.

rolznz:
> `list_transactions` now, but all the timestamps are strings instead of numbers

@v0l sorry about that, I have created an issue [here](https://github.com/getAlby/nostr-wallet-connect/issues/204).

benthecarman:
How close are we on merging this? Seems like we have 2 implementations in alby and rust-nostr with no further comments

47.md
```diff
@@ -126,6 +128,235 @@ Response:
...
+{
+    "result_type": "multi_pay_invoice",
+    "result": {
+        "preimage": "0123456789abcdef..." // preimage of the payment
+    }
+}
+```
+
+Errors:
+- `PAYMENT_FAILED`: The payment failed. This may be due to a timeout, exhausting all routes, insufficient capacity or similar.
```
rolznz, line 163: This error could also apply to `pay_keysend`. Should we add it there?

rolznz:
> How close are we on merging this? Seems like we have 3 implementations in alby, snort/iris, and rust-nostr with no further comments

Alby has mostly finished the implementation except for `multi_pay_invoice` and some minor stuff like paging which we won't do now.

Regarding `multi_pay_invoice`, we also thought `multi_pay_keysend` makes sense, since many V4V podcasts have multiple splits (and having support for split payments in NWC would be great). Unfortunately the payment hash in individual responses cannot be used like in `multi_pay_invoice` (specifying preimage in the request is optional and we thought forcing the app to provide a preimage for each split is not a good solution. Returning an index for each response could be one possible way, unless there is a better idea).

Otherwise I think @bumi is happy to merge and already approved from the Alby side.

aChrisYouKnow:
Have been lurking this thread, finding a way to do multi_pay_keysend would be really nice if there was a way to do it. I guess the alternative would be the podcast apps (clients) would send multiple requests, one for each split, and the wallet service would pay each one individually?

benthecarman:
Yeah we could do mutli pay keysend but instead of tagging the payment hash, do the pubkey

rolznz:
> Yeah we could do mutli pay keysend but instead of tagging the payment hash, do the pubkey

Unfortunately this will not be enough to work for mapping responses for keysend payments to custodial wallets that use the [696969 record](https://github.com/satoshisstream/satoshis.stream/blob/main/TLV_registry.md#field-696969---lnpay)

Would it be possible to add an extra optional `696969` tag in the response for that case along with the pubkey?

We are actively trying to push self-custody (and NWC helps make this migration much more seamless) but without able to get responses for all splits, podcasting clients might not use it (and end up using one request per split with the `pay_keysend` method).

benthecarman:
The `696969` tag is kinda ugly, we could remove and just add an `id` or something to the params and put that in the tags

47.md
```diff
@@ -126,6 +128,276 @@ Response:
...
+        "payment_hash": "string", // Payment hash for the payment
+        "amount": 123, // value in msats
+        "fees_paid": 123, // value in msats
+        "created_at": unixtimestamp, // invoice/payment creation time
+        "expires_at": unixtimestamp, // invoice expiration time, optional if not applicable
+        "settled_at": unixtimestamp, // invoice/payment settlement time, optional if unpaid
+        "metadata": {} // generic metadata that can be used to add things like zap/boostagram details for a payer name/comment/etc.
+    }
+}
+```
```
rolznz, line 303: It seems we do not have any error handling or specify what to do if the invoice is not found. Some wallet providers may not return the invoice until it is paid, so even invoices made through NWC might not be returned. Would it make sense to have a specific error here?

Errors:
- `NOT_FOUND`: The invoice could not be found by the given parameters.

benthecarman: added

benthecarman:
image

TonyGiorgio:
LGTM

pablof7z:
We have two implementations of this (Mutiny.+ Alby); looks like this simply fell through the cracks so merging.

rolznz:
> I really wish the wallet service provided information in the kind 13194 instead of a request/response. The Nostr way is to simply subscribe to that event and the client will receive all updates when they happen. There is no need to make a separate request for that information.
> 
> It's similar to `list_payments` and `list_invoices`. We already have a filter API in Nostr. There is no need to recreate them as RPC calls.

I have created a PR for notifications which is based on this idea, for cases where it does not make sense or is inefficient to have to request first. https://github.com/nostr-protocol/nips/pull/1164

I would love any feedback!

PR #688, opened 2023-07-31
Wordsmith 65.md

git-sgmoore:


staab:
SHOULD is a technical term is spec writing that should be retained.

PR #691, opened 2023-08-01
NIP-101: Mailing lists

KaffinPX:


ariard:
May I humbly suggest to port POP and SMTP over nostr where you embed the mail object as nostr EVENT content and rely on nostr keys / lightning payments for authentication / spam management.

40y of internet engineering have been poured in mail standard, nice if good parts can be reused.

KaffinPX:
> May I humbly suggest to port POP and SMTP over nostr where you embed the mail object as nostr EVENT content and rely on nostr keys / lightning payments for authentication / spam management.
> 
> 40y of internet engineering have been poured in mail standard, nice if good parts can be reused.

I think that should be another NIP?

ariard:
> I think that should be another NIP?
(Actually SMTP is already decentralized, why we would need that?)

Okay what is proposed is not really dealing with SMTP mail objects, nevermind.
Architecture protocol is decentralized, not the deployment as spam has never been really solved.

PR #694, opened 2023-08-05
NIP-97 Attachments (binary content and media)

ondra-novak:
Implemented on wss://nostr-test.novacisko.cz/ 

Very simple demo client: https://nostr-test.novacisko.cz/files/index.html

Note that above relay is reset everyday at 00:00:00 UTC . Don't use it for other purposes.

FAQ
---

**Why binary messages. NOSTR is not designed to transmit binary messages**

I believe this is simply an excuse by people who have become comfortable with the status quo. That's why they are inventing special ways of encoding messages to "pass" binary content without taking full advantage of the potential of the technology that was chosen for NOSTR.

WebSocket was finalized into RFC 6455 in 2011. Now we are in 2023, 12 years later, programmers are afraid to use binary messages to transfer binary content. It's about on par with some email systems still relying on 7-bit Internet.

Introducing the transfer of binary content using binary messages solves a lot of issues. And if there are developers who use languages or libraries that do not support binary messages in websocket, please consider changing the library or even changing the language. Binary messages are supported by javascript in every browser.

**The advantage of NIP-95 is that it does not need a special relay implementation. This NIP requires massive development on both the relay and client side**

That's not true. If you read NIP-95 carefully, there are several requirements for how the relay should handle binary content - which is stored as base64. These are not trivial requirements. For example, the actual binary content is transmitted as an event, but it must not be indexed. This means adding a bunch of exceptions to the indexes.

If I want to have some control over the binary content, as a relay developer I have to do more programming, which I have to do in this proposal anyway. Plus some things are more difficult to deal with than using binary messages. For example, I have to allow the same interface for presenting normal events and the same for presenting NIP-95 events, yet handle them diametrically differently.

In the case of NIP-95, the relay cannot control the size of the binary content. The relay can also be configured so that text messages are short, but the binary content can be large, for example for storing videos, because the relay places attachments on a separate repository. This NIP also specifies in more detail how limits are communicated to the client. If any of the attachments are too large, the client can reduce the resolution or quality of the stored media to meet the limits set

An unresolved issue relates to garbage collection. If I want to deal with garbage collecting within NIP-95, I encounter various situations where a race condition can occur, so there can be a situation where someone uploads files but they can be deleted before the actual event is published. This is because the protocol treats each event as a separate record with no linkage to other records.

**Binary content should be transferred using the HTTP protocol see NIP-96**

No, it didn't. We're mixing protocols. What if the relay is not an HTTP server? Yes this can happen, an http server can be actually an upstream proxy. The NIP-96 design is also very complex

**What if a client sends an event with the tag "attachment" but sends it via the EVENT command?**

The `FETCH` command simply fails. Attachments aren't available.

**How a client that does not support this NIP work?**

This client can't see the attachments, only the `content` of the event

**What if two clients upload the same attachment (how to resolve hash collisions)**

Same hash = same content. The file is stored only once. The relay can discard duplicate content

**In this NIP, the files are not signed, it is not possible to verify who sent them to the relay. NIP-95, on the other hand, also signs the file**

The hash is always signed, not the content. In the case of NIP-95, the hash includes the author's pubkey, whereas in this proposal only the file hash itself is signed. This allows sharing the same file without reuploading it. And if you are worried about someone "stealing" ownership of the file. That may be the case with NIP-95, no signature will prevent a copy&paste operation. The only way to protect the file is to encrypt it

**Why the tags do not have the form proposed in NIP-94**

That was originally considered. I liked the compatibility with NIP-94. It would be possible to link the file directly to the metadata above NIP-94. But this proposal would not allow to attach multiple attachments to a single event. The motivation was to have it like Twitter, where there can be up to 4 images. Or like in Reels, where you can have short videos, but they can be linked into a longer continuous video. You can also implement a form of streaming longer videos in this way

**Images in Kind:0, image placement in text, smileys, etc.**

Just as I can place anything from tags using #[index], I can reference an attachment this way

**Why does it have to be uploaded atomically?**

This is related to the garbage collection requirement. It could happen that the client uploads attachments, but before publishing the event, the garbage collector comes and because the new attachments don't have a reference, it deletes them

**Can a client initiate multiple ATTACH requests at the same time (i.e. without completing the previous one)?**

This situation is not defined. For simplicity, let us assume that the answer is no, i.e. that a new ATTACH command within the same connection invalidates the previous ATTACH command if the event of that command has not been completed and published. This can possibly be discussed.

**Why does the client have to explicitly link existing attachments to the new event (FETCH+ATTACH), wouldn't it be enough for the relay to do this automatically when it sees that some attachments are already present in the relay?**

It would be possible, but it would significantly complicate the protocol flow. The client knows how many attachments it has to send. If the relay would add some attachments automatically, it would have to communicate this fact to the client. Presumably via an ATTACH response.

However, this still does not address the situation where there may be a race condition, where the same attachment appears on a relay before the request is completed. Therefore, it is better if the client handles the attachment management and the relay behaves passively in this case.

### DIFF
--- a/97.md
+++ b/97.md
@@ -1,0 +1,298 @@
+NIP-97
+=======
+
+Attachments
+-----------
+
+`draft` `optional` `author:ondra-novak`
+
+This NIP solves the problem of sharing and distributing binary content over NOSTR networks such as media - images, short videos music, or in general any files up to a certain size
+
+Key Takeaways
+-------------
+
+* Defines how binary content (images, media, files) is managed by NOSTR network
+* Allows sending private binary content in direct messages (encrypted)
+* Defines new commands at the protocol level ("ATTACH","FETCH")
+* Defines a new tag "attachment"
+* Uses the binary messages in the websocket connection
+
+
+New tag "attachment"
+-------------
+The proposal introduces a new tag  "attachment" with the following format
+
+```
+["attachment","hash_hex","size_in_bytes","mime_type","features1", "feature2", ....]
+```
+
+* **hash_hex** - hexadecimal representation of SHA256 hash of the attached binary file. This string also serves as **attachment-id**
+* **size_in_bytes** - size of the attachment in bytes
+* **mime_type** - mime type of the attachment 
+* **features...** - optional fields with various features in format `<field>=<value>`
+    * **blurhash=** - contains NIP-94 blurhash
+    * **purpose=** - purpose of the attachment in context of current note. This can help client
+to select optimal rendering. Suggested purposes
+        * **thumbnail** - thubnail image of a following video
+        * **story** - all attachments marked by this purpose are played as a story
+        * **inline** - attachment is rendered in text, referenced by `#[<index>]`. This purpose hides this attachment from the list of attachments in note's detail.
+
+```
+["attachment","12887...","64000","image/jpeg","blurhash=qwdjq3...", "purpose=thumbnail"]
+["attachment","874788...","2258147","video/mpeg4", "purpose=story"]
+["attachment","ae3c758...","2158799","video/mpeg4", "purpose=story"]
+["attachment","788eq78...","254123","image/jpeg", "purpose=story"]
+```
+        
+
+* One event can have more than one of these tags, then it carries more attachments.
+* Each relay defines the maximum attachment size and also the maximum number of attachments within an event
+* Attachment can be attached to any "kind". Attachments for `kind:4` are always encrypted
+
+The presence of the `attachment` attribute only informs the client that the relay should also have its own attachment content available in binary form. Therefore, the client should prepare for the visualization of the post, taking into account that attachments are available.
+
+
+Upload
+------
+Events must be published simultaneously with attachments. It is necessary to prevent the client from publishing the event but not uploading the appropriate attachments. Similary, it shouldn't be possible to upload attachments first and then publish the event that refers to them. The operation must be atomic
+
+For this reason, a new command is introduced: "ATTACH"
+
+**Protocol flow**
+
+```
+client: ["ATTACH", <event>]
+relay:  ["OK","<event-id">,true,"continue"]
+client: <binary message for first attachment>
+relay:  ["ATTACH",<attachment-id>, true, "continue"]
+client: <binary message for second attachment>
+relay:  ["ATTACH",<attachment-id>, true, "continue"]
+...
+...
+client: <binary message for last attachment>
+relay:  ["ATTACH",<attachment-id>, true, "complete"]
+     //event is published
+```
+
+The protocol flow is designed in such a way that it is possible to post other commands between individual phases. Possible implementation on a relay:
+
+```
+- on ATTACH command
+   - validate the event
+   - store the event in a temporary storage associated with current connection
+
+- on binary message
+   - calculate SHA256 hash of the binary message
+   - find matching attachment in list of attachment of events stored in the temporary storage
+   - store the binary message as new attachment in the temporary storage
+   - check if all attachments have been uploaded
+        - is so, publish the stored event and attachments and delete them from temporary storage
+
+ - on connection closed before completion
+   - destroy temporary storage (no publish)
+   
+```
+
+
+**NOTE** If the relay does not support this NIP, it will not be able to respond to the new command, or it will respond with "NOTICE". If the relay does not respond to the command within a certain time (e.g. 10 seconds), then the client should verify that the relay responds to other commands (for example "REQ") and evaluate the situation as if the relay does not support the functionality. If the relay responds to the command with a "NOTICE" response, the client must assume that the relay does not support the function and try another method
+
+Response - ATTACH
+--------------------------------------
+
+Response for ATTACH is similar as response to EVENT. It uses "OK" response.
+
+The fourth argument may contain more detailed information about the operation. 
+* `"continue"` - accepted, continue sending attachments
+* `"duplicate"` - event is duplicate, you can still send attachments.
+
+Errors - ATTACH
+-----------------------------
+When an error occurs after an ATTACH command, the client must not send any binary messages in this case.
+
+Error description is specified as fouth argument
+
+```
+relay:  ["OK","<event-id">,false,"error"]
+```
+
+* `"max_attachment_size: <number>"` - one of the attachments exceeded the maximum size allowed, the relay sends this limit as a number in the error description section
+* `"max_attachment_count: <number>"` - the number of attachments has exceeded the allowed limit. Again, the relay sends this limit as a number in the error description section
+* `"<event-id">,false,"invalid: <description>"` - event is malformed, missing mandatory fields, or event doesn't have attachments
+
+
+Response - binary messages
+--------------------------------------
+
+Relay must respond to each binary message with a ["ATTACH"] response. The response has the same format as "OK"
+
+The fourth argument may contain more detailed information about the operation. 
+
+```
+relay:  ["ATTACH","<attachment-id">,true,"description"]
+```
+
+* `"continue"` - accepted, continue sending attachments
+* `"complete"` - event has been posted
+* `"duplicate"` - event has been posted, but was duplicated (so it was discarded)
+
+
+Errors - binary messages
+--------------------------------------
+```
+relay:  ["ATTACH","<attachment-id">,false,"description"]
+```
+
+Error description is specified as fouth argument
+
+* `"invalid: mismatch"` - Binary message doesn't match any expected attachment. For example, the hash or size doesn't match.
+
+
+
+Encryption - kind:4
+-------------------
+
+Attachments are sent encrypted for "kind:4". The same encryption is used as in the encryption of the message itself. An initialization vector (16 bytes) is stored at the beginning of the binary message and then the binary content itself.
+
+```
+[IV 16 bytes][ encrypted content ]
+```
+
+Download 
+--------
+
+The "FETCH" command is used to download the attachment from the relay
+
+**Protocol flow**
+
+```
+client: ["FETCH","<attachment-id>"]
+relay:  ["FETCH","<attachment-id>", true, "mime/type"]
+relay:  <binary message>
+```
+
+If the attachment does not exist, the response looks like this
+
+
+```
+client: ["FETCH","<attachment-id>"]
+relay:  ["FETCH","<attachment-id>", false, "missing: not found"]
+```
+In this case, the relay must not generate a binary message
+
+Encrypted attachments must be decrypted at client side.
+
+
+Reuse existing attachment
+-------------------------
+
+To avoid reupload of the same file, you can use special form of FETCH command to 
+attach existing attachment to a new event. This can be useful for long articles in case that user just posts an update of the article.
+
+```
+client: ["FETCH","<attachment-id>","ATTACH"]
+relay: ["ATTACH","<attachment-id>",true,"continue/complete/duplicate"]
+```
+
+in case that attachment doesn't exists
+
+```
+client: ["FETCH","<attachment-id>","ATTACH"]
+relay: ["ATTACH","<attachment-id>",false,"missing: not found"]
+```
+
+in case that attachment doesn't match to any expected attachment
+
+```
+client: ["FETCH","<attachment-id>","ATTACH"]
+relay: ["ATTACH","<attachment-id>",false,"invalid: mismatch"]
+```
+
+This command also "locks" the attachment in order to prevent to be scarped by a garbage collector (see below)
+
+
+Changes in relay information document (NIP-11)
+-----------------------------------------------
+
+New items are added in the "limitation" section
+
+```
+ "limitation": {
+    "max_attachment_count": 4,
+    "max_attachment_size": 262144,
+  }
+```
+
+* **max_attachment_count** - maximum count of attachments per event
+* **max_attachment_size** -- maximum size of single attachment in bytes
+
+
+Garbage collecting
+------------------
+
+Relay should perform garbage collecting of attachments without any reference on events. 
+
+
+FAQ
+---
+
+**Why binary messages. NOSTR is not designed to transmit binary messages**
+
+I believe this is simply an excuse by people who have become comfortable with the status quo. That's why they are inventing special ways of encoding messages to "pass" binary content without taking full advantage of the potential of the technology that was chosen for NOSTR.
+
+WebSocket was finalized into RFC 6455 in 2011. Now we are in 2023, 12 years later, programmers are afraid to use binary messages to transfer binary content. It's about on par with some email systems still relying on 7-bit Internet.
+
+Introducing the transfer of binary content using binary messages solves a lot of issues. And if there are developers who use languages or libraries that do not support binary messages in websocket, please consider changing the library or even changing the language. Binary messages are supported by javascript in every browser.
+
+**The advantage of NIP-95 is that it does not need a special relay implementation. This NIP requires massive development on both the relay and client side**
+
+That's not true. If you read NIP-95 carefully, there are several requirements for how the relay should handle binary content - which is stored as base64. These are not trivial requirements. For example, the actual binary content is transmitted as an event, but it must not be indexed. This means adding a bunch of exceptions to the indexes.
+
+If I want to have some control over the binary content, as a relay developer I have to do more programming, which I have to do in this proposal anyway. Plus some things are more difficult to deal with than using binary messages. For example, I have to allow the same interface for presenting normal events and the same for presenting NIP-95 events, yet handle them diametrically differently.
+
+Relay also cannot dictate the size of the binary content. There is a limit on the size of each kind, but clients don't examine this much and most of them react to the error message by simply not allowing the operation. For example, a client uploading a large image that the relay rejects might reduce the resolution or compression quality of the image to fit to the relay's limits. This is the motivation behind this proposal.
+
+An unresolved issue relates to garbage collection. If a reference to a shared binary content disappears, does the relay have the right to delete this content? If I take an ordinary relay that does not implement NIP-95, then these events stay there - forever. What is the correct behavior?
+
+**Binary content should be transferred using the HTTP protocol see NIP-96**
+
+No, it didn't. We're mixing protocols. What if the relay is not an HTTP server? Yes this can happen, an http server can be actually an upstream proxy. The NIP-96 design is also very complex
+
+**What if a client sends an event with the tag "attachment" but sends it via the EVENT command?**
+
+Every client should be prepared for a situation where the attachment is not stored on the relay. The `FETCH` command will simply fail. That will happen in this case as well. So the event will be published but the attachments will not be visible
+
+**How a client that does not support this NIP work?**
+
+This client can't see the attachments, only the `content` of the event
+
+**What if two clients upload the same attachment (how to resolve hash collisions)**
+
+Same hash = same content. The file is stored only once. The relay can discard duplicate content
+
+**In this NIP, the files are not signed, it is not possible to verify who sent them to the relay. NIP-95, on the other hand, also signs the file**
+
+The hash is always signed, not the content. In the case of NIP-95, the hash includes the author's pubkey, whereas in this proposal only the file hash itself is signed. This allows sharing the same file without reuploading it. And if you are worried about someone "stealing" ownership of the file. That may be the case with NIP-95, no signature will prevent a copy&paste operation. The only way to protect the file is to encrypt it
+
+**Why the tags do not have the form proposed in NIP-94**
+
+That was originally considered. I liked the compatibility with NIP-94. It would be possible to link the file directly to the metadata above NIP-94. But this proposal would not allow to attach multiple attachments to a single event. The motivation was to have it like Twitter, where there can be up to 4 images. Or like in Reels, where you can have short videos, but they can be linked into a longer continuous video. You can also implement a form of streaming longer videos in this way
+
+**Images in Kind:0, image placement in text, smileys, etc.**
+
+Just as I can place anything from tags using #[index], I can reference an attachment this way
+
+**Why does it have to be uploaded atomically?**
+
+This is related to the garbage collection requirement. It could happen that the client uploads attachments, but before publishing the event, the garbage collector comes and because the new attachments don't have a reference, it deletes them
+
+**Can a client initiate multiple ATTACH requests at the same time (i.e. without completing the previous one)?**
+
+This situation is not defined. For simplicity, let us assume that the answer is no, i.e. that a new ATTACH command within the same connection invalidates the previous ATTACH command if the event of that command has not been completed and published. This can possibly be discussed.
+
+**Why does the client have to explicitly link existing attachments to the new event (FETCH+ATTACH), wouldn't it be enough for the relay to do this automatically when it sees that some attachments are already present in the relay?**
+
+It would be possible, but it would significantly complicate the protocol flow. The client knows how many attachments it has to send. If the relay would add some attachments automatically, it would have to communicate this fact to the client. Presumably via an ATTACH response.
+
+However, this still does not address the situation where there may be a race condition, where the same attachment appears on a relay before the request is completed. Therefore, it is better if the client handles the attachment management and the relay behaves passively in this case.
+
+
--- a/97.md
+++ b/97.md
@@ -4,7 +4,7 @@ NIP-97
 Attachments
 -----------
 
-`draft` `optional` `author:ondra-novak`
+`final` `optional` `author:ondra-novak`
 
 This NIP solves the problem of sharing and distributing binary content over NOSTR networks such as media - images, short videos music, or in general any files up to a certain size
 
--- a/97.md
+++ b/97.md
@@ -49,12 +49,12 @@ to select optimal rendering. Suggested purposes
 * Each relay defines the maximum attachment size and also the maximum number of attachments within an event
 * Attachment can be attached to any "kind". Attachments for `kind:4` are always encrypted
 
-The presence of the `attachment` attribute only informs the client that the relay should also have its own attachment content available in binary form. Therefore, the client should prepare for the visualization of the post, taking into account that attachments are available.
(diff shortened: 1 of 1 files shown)


vitorpamplona:
I think we could just define a binary protocol. We don't need to mix text and binary and create a stateful mess to manage between the two.  It could be a TLV with length of hash, hash as byte array, length of content, content. Clients would send a TLV as binary.

ondra-novak:
You forget that everything revolves around the event and it is, was and will always be text-based. The text channel also controls the protocol communication and it doesn't make sense to me to create a secondary, binary channel with the same purpose

I also remind you that the attachment metadata are part of the event which is signed. So they have a double role. Firstly, you are announcing in advance how many attachments you are going upload, how big they are going to be and what type, but later also verifying who uploaded them and that hasn't been changed along the way (because the signature is in the event).

So in the end you have to send this metadata in text form anyway.

And what you're left with is just the binary content. The entire binary message is used for that without any extra processing (no TLV is needed). The whole binary message is the content.

I also completely disagree with the "stateful mess" criticism. The whole NOSTR is stateful. The REQ statement defines a certain state. Each relay must manage a state of the connected client. In my relay implementation, the class  "Peer" ([peer.h](https://github.com/ondra-novak/nostr_server/blob/master/src/nostr_server/peer.h)) is responsible for this. You can judge the complexity of the attachment state implementation by yourself ([attachments.h](https://github.com/ondra-novak/nostr_server/blob/master/src/nostr_server/attachments.h), [attachments.cpp](https://github.com/ondra-novak/nostr_server/blob/master/src/nostr_server/attachments.cpp))

vitorpamplona:
My point was that you don't need to "announce in advance" that you are going to send something. The server can just be listening to new attachments in the binary channel. We just need to move the hash in there so that the server can validate the payload.

ondra-novak:
"Announce in advance" is the key point!

It is part of a secure and controlled data transfer. The client has the possibility to find out that his attachment cannot be accepted before it is sent, for example because the relay has a limit defined. I can't allow clients to send a gigabyte binary message only to have the relay tell them at the end that their message was too big. I can also force regulation on content type, for example having a relay that only allows images (mime: image/*)

Announcing large transfers in advance is a technique that is commonly used, for example in the HTTP protocol itself (Except: 100-Continue). 

And then there is the purely abstract view. We can look at an event as a text part and attachments. The whole event needs to be sent at once, and since part of it is text and part of it is binary, the whole transfer needs to be split into a text transfer and a binary transfer. It doesn't matter how it's implemented on the relay then, if the attachments are together with the event or separated.

> And once you do that, you can also have a protocol in place to transfer any nostr event or filter over the binary protocol.
>The text is just the easiest serialization of the abstract data model. Nothing forbids Nostr from implementing other serialization structures

But that's for another proposal. 

What I'm a little sad about is that in the time it's taken to create a lot of relays and clients, no one has come up with a good proposal for how to transfer binary content over NOSTR. No one. 

Who is this network for? Users prefer pictures, videos, instagram, tiktok.

Send your own NIP, we can discuss

vitorpamplona:
If the goal is just to check if the relay can receive it, you should use [NIP-11](https://github.com/nostr-protocol/nips/blob/master/11.md) and advertise that limits there together with all the other limits. There is no need to check each individual upload.

ondra-novak:
> If the goal is just to check if the relay can receive it, you should use [NIP-11](https://github.com/nostr-protocol/nips/blob/master/11.md) and advertise binary limits there, together with all the other limits. There is no need to check each individual upload.

This one of goal. Not only one, see my first reply. Are you author of Amethyst? What about your application want to know in advance, how big attachment will process. As I remember, some large allocations on android can crash the application. 

I still don't understand the arguments. What exactly are you arguing? If the whole message is about "we have to do it differently", then please write how to do it differently. Please write it as a proposal with a detailed description. I'm not going to do your homework.

vitorpamplona:
> What about your application want to know in advance, how big attachment will process. 

There is no need for it. First, the attachment tag already has size and the app will always have the nostr event before it downloads the attachment. Second, neither relays, not apps, can trust the other party to send the right size, or hash, or mime-type. You have to design for conflict. This means the app or relay would just cut you off if the amount of data is too large. It doesn't matter what you claim the size is. 

> I still don't understand the arguments. What exactly are you arguing?

That the protocol can be a lot simpler and easier to implement without an additional state. 

> Please write it as a proposal with a detailed description. I'm not going to do your homework.

Relax man. I am giving you feedback form a client developer. If you don't want it, why are you even publishing this here...

ondra-novak:
I cannot accept the feedback because I do not find it relevant. Based on our discussion, I have decided not to change anything in the proposal for now.

97.md
```diff
@@ -0,0 +1,251 @@
...
+* Defines a new tag "attachment"
+* Uses the binary messages in the websocket connection
+
+
+New tag "attachment"
+-------------
+The proposal introduces a new tag  "attachment" with the following format
+
+```
+["attachment","hash_hex","size_in_bytes","mime_type","features1", "feature2", ....]
```
vitorpamplona, line 26: You could use a FileMetadata Event instead of this tag. https://github.com/nostr-protocol/nips/blob/master/94.md  You can just use the already common `x` tag to be the hash of the attachment. There is no need to redefine all possible metadata tags here again. 

Additionally, in that way, we don't run the risk of breaking relays out there that don't support infinite tag columns.

97.md
```diff
@@ -0,0 +1,251 @@
...
+
+```
+["attachment","12887...","64000","image/jpeg","blurhash=qwdjq3...", "purpose=thumbnail"]
+["attachment","874788...","2258147","video/mpeg4", "purpose=story"]
+["attachment","ae3c758...","2158799","video/mpeg4", "purpose=story"]
+["attachment","788eq78...","254123","image/jpeg", "purpose=story"]
+```
+        
+
+* One event can have more than one of these tags, then it carries more attachments.
```
vitorpamplona, line 48: Using NIP-94 you would have just one event per attachment, which enables more interesting features like defining owners, liking, zapping, and reporting each attachment individually.

97.md
```diff
@@ -0,0 +1,251 @@
...
+* Each relay defines the maximum attachment size and also the maximum number of attachments within an event
+* Attachment can be attached to any "kind". Attachments for `kind:4` are always encrypted
+
+The presence of the `attachment` attribute only informs the client that the relay should also have its own attachment content available in binary form. The client must download the attachments from the same relay to display them. See below
```
vitorpamplona, line 52: SHOULD or MUST? If relays have the event, they MUST have the attachment as well. Otherwise, it gets a little messy on where to find it. 

Relays that don't support this NIP will all have the event, but not the attachment. Thus, I don't think there is a good solution for MUST. Clients must always assume the relays might not have it and figure out a way to discover which relay has the attachment.

97.md
```diff
@@ -0,0 +1,251 @@
...
+For this reason, a new command is introduced: "ATTACH"
+
+**Protocol flow**
+
+```
+client: ["ATTACH", <event>]
+relay:  ["OK","<event-id">,true,"continue"]
+client: <binary message for first attachment>
+relay:  ["ATTACH",<attachment-id>, true, "continue"]
+client: <binary message for second attachment>
```
vitorpamplona, line 68: Add the attachment-id to the binary content to avoid race conditions in multi-threaded, multi-connection environments. Nostr doesn't have a notion of atomic transaction for a sequence of messages and there is no need to create one for this.

97.md
```diff
@@ -0,0 +1,251 @@
...
+Relay must respond to each binary message with a ["ATTACH"] response. The response has the same format as "OK"
+
+The fourth argument may contain more detailed information about the operation. 
+
+```
+relay:  ["ATTACH","<attachment-id">,true,"description"]
+```
+
+* `"continue"` - accepted, continue sending attachments
+* `"complete"` - event has been posted
```
vitorpamplona, line 139: If a client uploads an event with 3 attachments but uploads only 2, and posts it, what happens? And what happens if the client tries to send the last attachment a few seconds later? Will the relay have deleted the event and attachments already? Will it allow a new attachment in? Will it just remain incomplete forever?

vitorpamplona (commented):
Additional comments

97.md
ondra-novak: This is replied in FAQ section

97.md
ondra-novak: So in summary, you suggest one binary content to one event. You will need to define kind for such event. Similar to NIP-95
This is also addressed in FAQ section.

This will need to two indirections which can complicate some things. How do you implement stories (very popular facebook stories). How do you manage replaceable events. I want to have nice icon and background image in "Kind: 0" Who is responsible to erase suich unused events with  attachments? Will your client implement a gallery or a file manager?

ondra-novak: NIP-95 can't be used for Kind:4

97.md
ondra-novak: SHOULD as your client must be fault tolerant. Similarly, the image you point to via the URL may be missing

If the relay does not have an attachment, the client is under no obligation to look for it elsewhere. He just doesn't have it, it's not available. Broken image icon.

97.md
```diff
@@ -0,0 +1,251 @@
...
+* One event can have more than one of these tags, then it carries more attachments.
+* Each relay defines the maximum attachment size and also the maximum number of attachments within an event
+* Attachment can be attached to any "kind". Attachments for `kind:4` are always encrypted
+
+The presence of the `attachment` attribute only informs the client that the relay should also have its own attachment content available in binary form. The client must download the attachments from the same relay to display them. See below
+
+
+Upload
+------
+Single event must be published simultaneously with attachments. It is necessary to prevent the client from publishing the event but not uploading the relevant attachments. Similarly, it should not be possible to upload attachments first and then publish the event that references them. The operation must be atomic
```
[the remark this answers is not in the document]
ondra-novak: It is not rule, it is an explanation. 
You ran into the problem yourself below

97.md
ondra-novak: No one here is talking about a multithreading environment. 
The word atomic is meant in the sense of database transactions.
It means that in one state the relay does not have an event and attachments and in the other state it has the event and attachments at the same time. There should not be a situation where the relay will have an event without attachments or attachments without the event (in the latter case, it is more of a problem with the GC)

97.md
vitorpamplona: > you suggest one binary content to one event. You will need to define kind for such event.

NIP-94 IS that event. It already exists, it's being used and can easily do what you need here. There is no need to define a new one. 

> This will need to two indirections which can complicate some things.

Zaps have 1-2 indirections. The New DMs will have 4. Some group chats will have 5. NIP-94 also has redirections. It's really not a big deal. 

> Will your client implement a gallery or a file manager?

We already do. There is an online client as well: https://filestr.vercel.app/ 

> Who is responsible to erase suich unused events with attachments? 

Garbage collectors already work for Zaps that don't have an event, or likes that don't have an event. This shouldn't be a big problem. Orphan events are EXTREMELY common in Nostr. 

> NIP-95 can't be used for Kind:4

They allow encryption, so I already used them like that. Same for NIP-94. I don't know why you would say it can't be used.

97.md
```diff
@@ -0,0 +1,251 @@
...
+client: <binary message for second attachment>
+relay:  ["ATTACH",<attachment-id>, true, "continue"]
+...
+...
+client: <binary message for last attachment>
+relay:  ["ATTACH",<attachment-id>, true, "complete"]
+     //event is published
+```
+
+The protocol flow is designed in such a way that it is possible to post other commands between individual phases. Possible implementation on a relay:
```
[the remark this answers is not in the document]
ondra-novak: This rule is important because it allows you to send e.g. [REQ] commands at the same time as the upload. At the same time, you cannot influence whether the relay itself sends you an event in the meantime as a response to your REQ. It's just important to mention that the requests don't have to follow each other. That's why each answer contains the identifier of what it answers

97.md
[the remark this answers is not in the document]
vitorpamplona: So you are saying that "Single event MUST be published simultaneously with attachments" is not a MUST? Can you change it?

97.md
```diff
@@ -0,0 +1,251 @@
...
+Download 
+--------
+
+The "FETCH" command is used to download the attachment from the relay
+
+**Protocol flow**
+
+```
+client: ["FETCH","<attachment-id>"]
+relay:  ["FETCH","<attachment-id>", true, "mime/type"]
```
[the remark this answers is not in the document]
ondra-novak: I agree, this is not well defined. 

The original purpose was to make work easier for the demo, where (see the link above) where you enter the ID of the attachment for fetch and it is not clear how it should be displayed. And also previously there was a LINK command that allowed you to get the URL to the attachment (see demo), and where the relay must know the content-type to correctly build HTTP response. However, here I might rewrite the parameters of the FETCH command like this

```
["FETCH",<event-id>,<tag-index>]
```
This further separates the presentation from the implementation. Attachment is visually part of the event, the fact that it is addressed internally by a hash is implementation issue

In this case, you will get mime from the attachment tag

97.md
ondra-novak: This is why there is a requirement for it to have transactional behavior (atomic)

By issuing the command ATTACH you open an transaction on the relay
while the transaction is active, garbage collector cannot delete attachments.
Transaction is automatically committed by posting the last attachment (complete status)
Transaction is automatically rollbacked when you close connection or when you issue another ATTACH command.

In my implementation, I store all attachments in the main storage immediately, however the transaction holds a special object `AttachmentLock`, which serves as reference to an attachment (so GC can see that reference). This lock lives on the transaction until it is committed or rollbacked.

97.md
ondra-novak: OK I can accept this, you suggest binary version of NIP-95 (in summary, thank for feedback)

97.md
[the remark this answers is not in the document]
ondra-novak: Don't understant. Cílem je, aby události a přílohy byly zaslány ve stejné transakci a vypadalo to jako by byly publikovány současně. There is nothing to force here, this is simply an explanation of why it is necessary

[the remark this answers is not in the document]
ondra-novak: There is nothing to force here, this is simply an explanation of why it is necessary. I don't understant

97.md
```diff
@@ -0,0 +1,251 @@
...
+
+
+Changes in relay information document (NIP-11)
+-----------------------------------------------
+
+New items are added in the "limitation" section
+
+```
+ "limitation": {
+    "max_attachment_count": 4,
```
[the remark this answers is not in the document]
ondra-novak: This rule is purely to protect the relay against bad users who can send an endless stream of attachments in order to carry out a DoS attack. Assume that the resulting transaction can have some maximum size. However, the limit is not mandatory, this rule only defines the existence of the given error message (and option). Relay may not require any limit. Alternatively, we can define a limit on the total size of all attachments in the event

97.md
ondra-novak: >Garbage collectors already work for Zaps that don't have an event, or likes that don't have an event. This shouldn't be a big problem. Orphan events are EXTREMELY common in Nostr and some relays already deal with them.

Which NIP defines garbage collection of orphan of events?

97.md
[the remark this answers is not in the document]
vitorpamplona: > protect the relay against bad users who can send an endless stream of attachments in order to carry out a DoS attack. 

But that rule doesn't protect against it. Remember, creating new users and new events has cost zero in Nostr. One can create millions of users while uploading the first attachment :)

97.md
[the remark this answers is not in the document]
vitorpamplona: I don't think the mime type is needed here. The app already has it from the event and tag.

97.md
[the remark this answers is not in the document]
ondra-novak: Yeah, but i speaking about transaction size

ondra-novak:
To summarize. You would like to see the binary implementation as a binary version of NIP-95 and an extension of NIP-94. 

Why haven't you written something like that already.Overall, it bothers me guys that NOSTR is in a state where we all know how not to do it but few are able to put together something meaningful. Please solve binary content, better than base64 sent over text. And better than using third party services.

I'll withdraw my proposal when an alternative proposal appears. I will not post another proposal

97.md
vitorpamplona: I would just not have the transaction scope at all and use a timeout for new files. GC can only delete after 5 minutes since posted. It's the same as a lock but you don't need to expect clients to behave (release the lock)

97.md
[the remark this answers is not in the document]
vitorpamplona: Agree, my point is that concurrent filters and sends are the default behavior of Nostr. There is no need to specify it. 

You only have the need to specify because you are creating a temporary atomic transaction substate in each submission, which doesn't exist anywhere else in the Nostr protocol.

97.md
ondra-novak: I have always problem with magic constants in a code.

Where I write anything about releasing lock?

vitorpamplona:
> You would like to see the binary implementation as a binary version of NIP-95 and an extension of NIP-94.

I'd like to see you building on top of things that already exist instead of recreating everything. 

> Why haven't you written something like that already

Mostly because I think NIP95 is not that bad. But you presented something new (thank you for that) and I am evaluating if we code it or not. In the end, every client will do the same evaluation I am doing. 

Your proposal is not bad. It's also not that far out from NIP-94/95/98. I think it's smart but could use some help to fit in better with the others. That was my feedback.

97.md
vitorpamplona: > There should not be a situation where the relay will have an event without attachments or attachments without the event. 

But there are.... In 2 particular ways: 

1. All relays that do not implement this NIP will have events without attachments and there is nothing you can do about it. You might not care about it, but I HAVE to. The client always has to deal with multiple relays out there. Even the crazy ones. 

2. Deleting events happen all the time these days. Deleting attachment tags from a Replaceable events is also common. Either way, the GC is going to take its time. Meanwhile, for clients, your relay is offering attachments without events. As a client, I must ALWAYS expect that the relay is broken, because the GC has run yet. 

Because of 1 and 2, your atomic lock doesn't work. 

There is no need to make guarantees you cannot keep as a relay. If you can't absolutely guarantee them, there is no need to have them in the first place. 

In other words: It's ok if you have orphan attachments. You don't need to make a solution for that.

97.md
vitorpamplona: I don't think you can avoid the magic constant. You will need a time-out because the client might never finish uploading everything (it might/will crash)

As a client, the specific timeout a relay uses doesn't matter. I have to assume a broken database nevertheless.

97.md
vitorpamplona: > Which NIP defines garbage collection of orphan of events?

I don't think there is an agreed NIP for that and we might actually not need an agreement on how GCs are run. For clients, GC doesn't really matter. It's purely a relay software/operator choice. I assume most of them just do it to reduce costs (memory and database sizes). 

Paid relays usually have tons of orphans because a paid user can cite an event that was signed by an unpaid user. Since the "reply" event author has paid, they don't delete those events. Thus, orphans are everywhere.

97.md
vitorpamplona: I think all relays that use a relational database have a limit on the number of columns. We just don't know what those are :(

97.md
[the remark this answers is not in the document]
vitorpamplona: Then just replace MUST with SHOULD. It's just a better word.

mikedilger:
I like this NIP better than NIP-95 which is wasteful using base64.
Thank you for not trying to stuff everything into the Event structure.

> programmers are afraid to use binary messages to transfer binary content

I've wanted to use binary since day-1, but nostr was already in use with a JSON object.

I have no problem using the binary websocket channel.  We use Text, Ping, Pong and Close.  It's not a stretch to use Binary.   My code currently reads:
````
 WsMessage::Binary(_) => tracing::warn!("{}, Unexpected binary message", &self.url),
````

All my prior warnings about binary content making nostr more of a target are still of a concern (relay operators we will have to deal with CSAM once binary uploads are easy). But I'm not going to stand in the way of progress. It's just a sad prediction.

I haven't dug into the details of this proposal, but I wanted to make my position clear that this is IMHO on the right path for dealing with binary content.

ondra-novak:
> > You would like to see the binary implementation as a binary version of NIP-95 and an extension of NIP-94.
> 
> I'd like to see you building on top of things that already exist instead of recreating everything.
> 
Believe it or not the original intent was to build it over the NIP-94. ([link](https://github.com/ondra-novak/nostr_server/blob/8c7919be58cb11e59aa43056bdcbf50fa386081a/docs/media.md))

However, I wanted to address the current inability to add media to direct messages (kind 4). I wasn't sure if simply linking a file event to a direct message event was a sufficient solution, and how the encryption between the two would be handled.  If I could see that encrypted file passing via direct messages works over, for instance, NIP-94 or NIP-95, I would probably have no reason to suggest different approach. But I haven't found any client that implements it that way

The second reason was the idea that messages with attachments are easier to understand because it is a common feature of email clients. If I built nostr on top of the email protocol, everything would be email and it could have attachments. Actually wait a minute, that already exists, it's called [Network News Transfer Protocol.](https://en.wikipedia.org/wiki/Network_News_Transfer_Protocol)

Okay, so I'll summarize the original idea of expanding NIP-94 in the following points:
- allocate a new kind 
- use tags from NIP-94 for metadata
- one event is associated with only one binary file

However, the following issues still remain to be resolved

How to optimally pass binary content to relay. I still get the best out of my proposed transactional way of creating an event, but now only one binary message is passed. I'm looking at this from the perspective of the demo client creator, as I'm not a javascript expert, but from that perspective I found it much easier to minimize the necessary work with binary content, which is limited to just collecting the data, generating the metadata (calculating the SHA256 hash, using a library of course) and passing the binary data directly to the websocket. I have no idea how complicated it would be for me to construct some binary message using TLV notation

I still believe that we need to define new commands at the protocol level, because that way we can easily detect relays that don't support the new feature. 

This also applies to download. NIP-95 relies on obtaining this special (base64 file) event via REQ and ID. Searching the index on my relay is not a trivial task given all the combinations the filter structure offers. At some level, I'm not even able to identify what event the the search found because it only works with internal IDs. So I would rather not mix event retrieval and binary content download. That's why I remain convinced that FETCH (or something similar) has a place here

ondra-novak:
> (relay operators we will have to deal with CSAM once binary uploads are easy)

This already exists in the current implementation, but the burden of responsibility is shifted to the repository operators. 

As a relay operator, I am aware of this responsibility and I am going to implement a feature that deletes any such content based on, for example, user reports. Therefore, the client must also be prepared for the eventuality that the referenced attachment/binary content is suddenly unavailable

vitorpamplona:
> I wanted to address the current inability to add media to direct messages (kind 4). 

I see. That is indeed not well defined in NIP-94 (or anywhere else). But the idea was to use a new tag to represent a nip04 encryption, similar to that `aes-256-gcm` but for the iv + recipient pub key of the NIP04 procedure. An important part was to separate the iv from the content and store only the content with external servers just to make it more difficult to decrypt. 

> How to optimally pass binary content to relay.

Your approach seems to be the best indeed. But keep in mind that most Websocket implementations allow devs to send and listen to text and byte arrays independently. We could use that in our favor to modularize things. The relay would listed for text as usual and simply add a binary content receiver. That receiver receives the hash and content and stores it in disk, independent if there is an event for it or not. Later on the GC cleans things up. Once the hash and the content is up, it doesn't really matter how other events use it. 

> I still believe that we need to define new commands at the protocol level. 

I agree. The question is: are these commands the right ones? And do they block any future use of the binary channel? 

> So I would rather not mix event retrieval and binary content download. That's why I remain convinced that FETCH (or something similar) has a place here

Very interesting. I agree. 

> As a relay operator, I am aware of this responsibility and I am going to implement a feature that deletes any such content based on, for example, user reports. 

If this is the case, then making a NIP-94 event for each attachment is better because it's super easy for Clients to implement reports on each NIP-94 attachment. You can also use per-attachment content labeling from [NIP-31](https://github.com/nostr-protocol/nips/blob/master/32.md) as well as reactions/zaps for the attachments themselves. If attachments are individual events, they can also specify licensing restrictions and a zap-split that can be re-used when the same attachment NIP-94 event is cited somewhere else. 

Now, NIP-94 is specifically designed for external content on other servers. Even though we are talking about that event kind here, I suggest we copy-paste the kind and make it an attachment header kind. Clients that receive this particular new event kind would then use the FETCH protocol to get the content.

ondra-novak:
>  They just need to understand the hash+content byte format.
If the hash is just SHA256 of the content, you don't need to transfer the hash.

vitorpamplona:
> > They just need to understand the hash+content byte format.
> 
> If the hash is just SHA256 of the content, you don't need to transfer the hash.

Sure, as long as there are no issues with partial/broken contents or other attack vectors, it should be fine. I assumed some large files (videos) will need to be transferred in multiple binary messages, but I don't have information on how big byte buffers in websockets are.

ondra-novak:
A very simple approach can be to allow the client to upload anything as binary content, store it, and after a period of time when no related event is sent, collect it as garbage. The related event should contain valid hash and valid size

But there's a catch. What if someone wants to use the NOSTR protocol at the binary level in the future, then it may be a problem to recognize whether the client is only sending binary content to be stored or whether this binary message is to be interpreted. That's why I went the " announce first, send later" approach. Personally, I think that any switching of the NOSTR protocol to the binary alternative will be handled differently, for example in the appropriate HTTP header in the initial handshake of the websocket protocol

ondra-novak:
> Sure, as long as there are no issues with partial/broken contents or other attack vectors,

Broken content should not happen in websocket realm. This is a fairly secure protocol. Even if you use fragmented message (continuation frame). The question of the buffers is more complicated. For example, my implementation concatenates all fragments of the message before sending it to the caller. There is, I think, a limit on the maximum size of the message that causes an exception that results to disconnect (with discarding the transmitted content)

vitorpamplona:
The simpler you make them, the easier to get traction. 

STORE/RETRIEVE ? 

We use `EVENT` to send JSON text. We could use BLOB to send binary. 

Or if you find a name for this, you can use it directly, like "EVENT".

ondra-novak:
Closing for #719

PR #696, opened 2023-08-06
Add HTTP-only NIP-05 IDs

arthurfranca:
This proposes `~@domain.com` as HTTP-only NIP-05 ID.

The immediate use case is being able to use zap tags such as `[ "zap", "~@service.example.com", "", "1" ]`
to config zap splits for NIP-96 HTTP file storage servers, because ideally such off-band services shouldn't be required to
have nostr accounts. See more on this [here](https://github.com/nostr-protocol/nips/pull/547#discussion_r1268797361).

If this gets merged, the next step would be adding the above zap tag example to NIP-57 and updating NIP-96 PR.

[Read here](https://github.com/arthurfranca/nips/blob/nip-05-http-only/05.md#http-only-ids)

@staab


arthurfranca:
Lol was hoping for a more involved feedback

fiatjaf:
My response was obviously tongue-in-cheek, sorry if it was perceived as disrespectful. I think you should have left this open to see feedback from others.

PR #700, opened 2023-08-07
Simplifies NIP-65

vitorpamplona:
This is how I would write this, @mikedilger 

If this is wrong, ignore it. 

Read [here](https://github.com/vitorpamplona/nips/blob/nip-65-simplified/65.md).

### DIFF
--- a/65.md
+++ b/65.md
@@ -58,4 +58,4 @@ This NIP allows Clients to connect directly with the most up-to-date relay set f
 
 4. DMs SHOULD only be broadcasted to the author's WRITE relays and to the receiver's READ relays to keep maximum privacy. 
 
-5. Relays that support this NIP MUST accept and serve kind 10002 events for everybody whether they are paying customers or not.
+5. If a relay signals support for this NIP in their [NIP-11](11.md) document that means they're willing to accept kind 10002 events from a broad range of users, not only their paying customers or whitelisted group.
--- a/65.md
+++ b/65.md
@@ -29,9 +29,9 @@ This NIP doesn't fully replace relay lists in `kind:3`. Instead, `Kind:3` relays
 
 ## When to Use Read and Write
 
-When seeking events **from** an author, Clients SHOULD use the WRITE relays on the author's `kind:10002`
+When seeking events **from** a user, Clients SHOULD use the WRITE relays of the user's `kind:10002`
 
-When seeking events **about** a user, where the user was tagged, Clients SHOULD use the READ relays on the user's `kind:10002` 
+When seeking events **about** a user, where the user was tagged, Clients SHOULD use the READ relays of the user's `kind:10002` 
 
 When broadcasting an event, Clients SHOULD:
 
@@ -46,7 +46,7 @@ The old model of using a fixed relay list per user centralizes in large relay op
   - Many users are pulling events from a large number of relays in order to get more data at the expense of duplication
   - Events are being copied between relays, oftentimes to many different relays
   
-This NIP allows Clients to connect directly with the most up-to-date relay set from each individual user, eliminating the need to use fixed lists and large operators. 
+This NIP allows Clients to connect directly with the most up-to-date relay set from each individual user, eliminating the need of broadcasting events to popular relays. 
 
 ## Final Considerations
 
--- a/65.md
+++ b/65.md
@@ -54,7 +54,7 @@ This NIP allows Clients to connect directly with the most up-to-date relay set f
 
 2. Clients SHOULD spread an author's `kind:10002` events to as many relays as viable. 
 
-3. `kind:10002` events MUST only be used to advertise the user's preferred relays to others. The Client's configuration SHOULD stay in other lists where other factors can be considered, such as specific relays for a given UI feature of the Client.
+3. `kind:10002` events should primarily be used to advertise the user's preferred relays to others. A user's own client may use other heuristics for selecting relays for fetching data.
 
 4. DMs SHOULD only be broadcasted to the author's WRITE relays and to the receiver's READ relays to keep maximum privacy. 
 
--- a/65.md
+++ b/65.md
@@ -42,11 +42,11 @@ When broadcasting an event, Clients SHOULD:
 
 The old model of using a fixed relay list per user centralizes in large relay operators: 
 
-  - Most people are sending their posts to the same most popular relays in order to be more widely seen
-  - Many people are pulling from a large number of relays (including many duplicate events) in order to get more data
+  - Most users submit their posts to the same highly popular relays, aiming to achieve greater visibility among a broader audience.
+  - Many users are pulling events from a large number of relays in order to get more data at the expense of duplication
   - Events are being copied between relays, oftentimes to many different relays
   
-This NIP allows apps to connect directly with the most up-to-date relay set from each individual user, eliminating the need to use large operators. 
+This NIP allows Clients to connect directly with the most up-to-date relay set from each individual user, eliminating the need to use fixed lists and large operators. 
 
 ## Final Considerations
 
--- a/65.md
+++ b/65.md
@@ -54,6 +54,6 @@ This NIP allows Clients to connect directly with the most up-to-date relay set f
 
 2. Clients should spread an author's `kind:10002` events to as many relays as viable, all of them if possible. 
 
-3. Kind `10002` should only be used to advertise relays. The user's configuration when using Clients should stay in Kind 3.
+3. `kind:10002` events should only be used to advertise relays. The user's configuration when using Clients should stay in Kind 3.
 
 4. DMs should be broadcasted to the author's WRITE relays and to the receiver's READ relays to keep maximum privacy. 
--- a/65.md
+++ b/65.md
@@ -39,7 +39,7 @@ When sending an event:
 
 ### Motivation
 
-The old model of using a fixed relay list per user centralizes in large relay operators. 
+The old model of using a fixed relay list per user centralizes in large relay operators: 
 
   - Most people are sending their posts to the same most popular relays in order to be more widely seen
   - Many people are pulling from a large number of relays (including many duplicate events) in order to get more data
--- a/65.md
+++ b/65.md
@@ -25,6 +25,8 @@ The `content` is not used and SHOULD be an empty string.
   ...other fields
 ```
 
+Clients must download all `kind:10002`s from their user's contact list and assemble a relay pool from there.
+
 ## How to Use Read and Write
 
 When looking for a given event: 
@@ -45,7 +47,7 @@ The old model of using a fixed relay list per user centralizes in large relay op
   - Many people are pulling from a large number of relays (including many duplicate events) in order to get more data
   - Events are being copied between relays, oftentimes to many different relays
   
-This NIP allows apps to connect directly with the most up-to-date relay set from each individual user. In order to do so, apps must download their contact list's `kind:10002` and assemble a relay pool from there. In this design, it doesn't matter if the user is using a personal relay or one of the largest ones. 
+This NIP allows apps to connect directly with the most up-to-date relay set from each individual user, eliminating the need to use large operators. 
 
 ### Final Considerations
 
--- a/65.md
+++ b/65.md
@@ -4,13 +4,13 @@ NIP-65
 Relay List Metadata
 -------------------
 
-`draft` `optional` `author:mikedilger`
+`draft` `optional` `author:mikedilger` `author:vitorpamplona`
 
-Defines a replaceable event using `kind:10002` to advertise a user's preferred relays to others. 
+Defines a replaceable event using `kind:10002` to advertise preferred relays for discovering a user's content and receiving fresh content from others.
 
-The event MUST include a list of `r` tags with relay URIs and a `read` or `write` marker. If the marker is omitted, it means the author uses the relay for both purposes.
+The event MUST include a list of `r` tags with relay URIs and a `read` or `write` marker. If the marker is omitted, the relay is used for both purposes.
 
-The `content` is not used.
+The `.content` is not used.
 
 ```json
 {
@@ -25,19 +25,18 @@ The `content` is not used.
   ...other fields
 ```
 
-Clients must download all `kind:10002`s from their user's contact list and assemble a relay pool from there.
+This NIP doesn't fully replace relay lists in `kind:3`. `Kind:3` relays SHOULD be used as a fallback when users use clients that don't support this NIP.
 
-## How to Use Read and Write
+## When to Use Read and Write
 
-When looking for a given event: 
+When seeking events **from** an author, Clients SHOULD use the WRITE relays on the author's `kind:10002`
 
-- Download the author's `kind:10002` and use the WRITE relays
-- Download a tagged user's `kind:10002` and use the READ relays 
+When seeking events **about** a user, where the user was tagged, Clients SHOULD use the READ relays on the user's `kind:10002` 
 
-When sending an event: 
+When broadcasting an event, Clients SHOULD:
 
-- Download the author's `kind:10002` and Broadcast the new event to the WRITE relays
-- Download all tagged user's `kind:10002`s and Broadcast the new event all READ relays 
+- Broadcast the event to the WRITE relays of the author
+- Broadcast the event all READ relays of each tagged user. 
 
 ### Motivation
 
@@ -51,8 +50,10 @@ This NIP allows apps to connect directly with the most up-to-date relay set from
 
 ### Final Considerations
 
-1. Clients should spread an author's kind `10002` events to as many relays as viable, all of them if possible. 
+1. Clients should guide users to keep this list small. 
 
-2. Kind `10002` should only be used to advertise relays. The user's configuration when using Clients should stay in Kind 3.
+2. Clients should spread an author's kind `10002` events to as many relays as viable, all of them if possible. 
 
-3. DMs should only be broadcasted to each author's WRITE relays to keep maximum privacy. 
+3. Kind `10002` should only be used to advertise relays. The user's configuration when using Clients should stay in Kind 3.
+
+4. DMs should be broadcasted to the author's WRITE relays and to the receiver's READ relays to keep maximum privacy. 
--- a/65.md
+++ b/65.md
@@ -25,7 +25,7 @@ The `.content` is not used.
   ...other fields
 ```
 
-This NIP doesn't fully replace relay lists in `kind:3`. Instead, `Kind:3` relays SHOULD serve as a fallback option for users that don't have a `kind:10002` event. 
+This NIP doesn't fully replace relay lists that are designed to configure a client's usage of relays (such as `kind:3` style relay lists). Clients MAY use other relay lists in situations where a `kind:10002` relay list cannot be found.
 
 ## When to Use Read and Write
 
@@ -52,8 +52,8 @@ This NIP allows Clients to connect directly with the most up-to-date relay set f
 
 1. Clients should guide users to keep `kind:10002` lists small (2-4 relays). 
 
-2. Clients should spread an author's `kind:10002` events to as many relays as viable, all of them if possible. 
+2. Clients should spread an author's `kind:10002` events to as many relays as viable. 
 
-3. `kind:10002` events should only be used to advertise relays. The user's configuration when using Clients should stay in Kind 3.
+3. `kind:10002` events should only be used to advertise the user's preferred relays. The Client's configuration should stay in other lists where other factors can be considered, such as specific relays for a given UI feature of the Client.
 
 4. DMs should be broadcasted to the author's WRITE relays and to the receiver's READ relays to keep maximum privacy. 
--- a/65.md
+++ b/65.md
@@ -6,61 +6,11 @@ Relay List Metadata
 
 `draft` `optional` `author:mikedilger`
 
-A special replaceable event meaning "Relay List Metadata" is defined as an event with kind `10002` having a list of `r` tags, one for each relay the author uses to either read or write to.
+Defines a replaceable event using `kind:10002` to advertise a user's preferred relays to others. 
 
-The primary purpose of this relay list is to advertise to others, not for configuring one's client.
+The event MUST includes a list of `r` tags with relay URIs and a `read` or `write` marker. If the marker is omitted, it means the author uses the relay for both purposes.
 
 The `content` is not used and SHOULD be an empty string.
-
-The `r` tags can have a second parameter as either `read` or `write`. If it is omitted, it means the author uses the relay for both purposes.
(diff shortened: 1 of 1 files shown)


mikedilger:
There is a lot I like about this simplification. Let's work off of this one. I'll propose some edits shortly.

vitorpamplona:
sorry, I just moved a few things around. I won't change it anymore.

65.md
```diff
@@ -6,71 +6,53 @@ Relay List Metadata
...
+    ["r", "wss://brando-relay.com"],
+    ["r", "wss://expensive-relay.example2.com", "write"],
+    ["r", "wss://nostr-relay.example.com", "read"],
+  ],
+  "content": "",
+  ...other fields
+```
 
-Write relays are for events that are intended for anybody (e.g. your followers). Read relays are for events that address a particular person.
+Clients must download all `kind:10002`s from their user's contact list and assemble a relay pool from there.
```
mikedilger, line 28: I would say something more like:

Clients SHOULD seek the `kind:10002` event for people their user's contact list in order to know where to find their events and where to copy them on events tagging them.

But I'm also ok with removing this paragraph entirely since it is implied.

65.md
```diff
@@ -6,71 +6,53 @@ Relay List Metadata
...
 
-Clients SHOULD write feed-related events created by their user to their user's write relays.
+## How to Use Read and Write
 
-Clients SHOULD read feed-related events created by another from at least some of that other person's write relays. Explicitly, they SHOULD NOT expect them to be available at their user's read relays. It SHOULD NOT be presumed that the user's read relays coincide with the write relays of the people the user follows.
+When looking for a given event:
```
mikedilger, line 32: When seeking the events of a given user:

65.md
```diff
@@ -6,71 +6,53 @@ Relay List Metadata
...
+Clients must download all `kind:10002`s from their user's contact list and assemble a relay pool from there.
 
-Clients SHOULD write feed-related events created by their user to their user's write relays.
+## How to Use Read and Write
 
-Clients SHOULD read feed-related events created by another from at least some of that other person's write relays. Explicitly, they SHOULD NOT expect them to be available at their user's read relays. It SHOULD NOT be presumed that the user's read relays coincide with the write relays of the people the user follows.
+When looking for a given event: 
 
-Clients SHOULD read events that tag their user from their user's read relays.
+- Download the author's `kind:10002` and use the WRITE relays
```
mikedilger, line 34: Seek them on one or more of that user's WRITE relays

65.md
```diff
@@ -6,71 +6,53 @@ Relay List Metadata
...
+- Download a tagged user's `kind:10002` and use the READ relays
```
mikedilger, line 35: Scratch this.  You never read from somebody's read relays.  You only write there.

65.md
```diff
@@ -6,71 +6,53 @@ Relay List Metadata
...
 
-Clients SHOULD read feed-related events created by another from at least some of that other person's write relays. Explicitly, they SHOULD NOT expect them to be available at their user's read relays. It SHOULD NOT be presumed that the user's read relays coincide with the write relays of the people the user follows.
+When looking for a given event: 
 
-Clients SHOULD read events that tag their user from their user's read relays.
+- Download the author's `kind:10002` and use the WRITE relays
+- Download a tagged user's `kind:10002` and use the READ relays 
 
-Clients SHOULD write events that tag a person to at least some of that person's read relays. Explicitly, they SHOULD NOT expect that person will pick them up from their user's write relays. It SHOULD NOT be presumed that the user's write relays coincide with the read relays of the person being tagged.
+When sending an event:
```
mikedilger, line 37: When posting an event:

Post to all of your own user's WRITE relays.

65.md
```diff
@@ -6,71 +6,53 @@ Relay List Metadata
...
-Clients SHOULD read events that tag their user from their user's read relays.
+- Download the author's `kind:10002` and use the WRITE relays
+- Download a tagged user's `kind:10002` and use the READ relays 
 
-Clients SHOULD write events that tag a person to at least some of that person's read relays. Explicitly, they SHOULD NOT expect that person will pick them up from their user's write relays. It SHOULD NOT be presumed that the user's write relays coincide with the read relays of the person being tagged.
+When sending an event: 
 
-Clients SHOULD presume that if their user has a pubkey in their ContactList (kind 3) that it is because they wish to see that author's feed-related events. But clients MAY presume otherwise.
+- Download the author's `kind:10002` and Broadcast the new event to the WRITE relays
+- Download all tagged user's `kind:10002`s and Broadcast the new event all READ relays
```
mikedilger, line 40: Nah, don't do this.  Put this:

When posting an event that tags a user:

Also post it to one or more of that user's READ relays.

staab:
This is a big improvement, although I don't know what "The user's configuration when using Clients should stay in Kind 3." means, since NIP 02 doesn't say anything about relays. Putting relays in the content was an ad-hoc thing from before NIP 65, my understanding was 65 was supposed to replace that.

65.md
```diff
@@ -6,71 +6,53 @@ Relay List Metadata
...
-  "tags": [
-    ["r", "wss://alicerelay.example.com"],
-    ["r", "wss://brando-relay.com"],
-    ["r", "wss://expensive-relay.example2.com", "write"],
-    ["r", "wss://nostr-relay.example.com", "read"],
-  ],
-  "content": "",
-  ...other fields
-```
+3. DMs should only be broadcasted to each author's WRITE relays to keep maximum privacy.
```
mikedilger, line 58: This should be READ relays.  Send to other people at their READ relays, where *they* read from.   Confusing I know and sorry we are stuck with that.

mikedilger (commented):
Initial feedback

mikedilger:
> This is a big improvement, although I don't know what "The user's configuration when using Clients should stay in Kind 3." means, since NIP 02 doesn't say anything about relays. Putting relays in the content was an ad-hoc thing from before NIP 65, my understanding was 65 was supposed to replace that.

NIP-65 doesn't replace relay lists in kind 3, they were meant to configure clients that used a fixed set of relays, and yes it was never NIPped.

I would take this line out, but perhaps mention that clients will likely need additional relay configuration data.

mikedilger:
Add yourself to the authors @vitorpamplona

65.md
vitorpamplona: Don't you read people's read relays when you are visiting their profile? Like to count the zaps they received?

65.md
vitorpamplona: Isn't that the same thing?

65.md
vitorpamplona: Fixed

vitorpamplona:
See if it got better

https://github.com/vitorpamplona/nips/blob/nip-65-simplified/65.md

65.md
vitorpamplona: nice, that does make sense for the use in profile pages.

65.md
```diff
@@ -4,73 +4,56 @@ NIP-65
 Relay List Metadata
 -------------------
 
-`draft` `optional` `author:mikedilger`
+`draft` `optional` `author:mikedilger` `author:vitorpamplona`
 
-A special replaceable event meaning "Relay List Metadata" is defined as an event with kind `10002` having a list of `r` tags, one for each relay the author uses to either read or write to.
+Defines a replaceable event using `kind:10002` to advertise preferred relays for discovering a user's content and receiving fresh content from others.
```
mikedilger, line 9: Those two reasons sound the same to me.  How about "for finding a user's content and for ensuring content that references them can be found by them."

65.md
```diff
@@ -4,73 +4,56 @@ NIP-65
...
+    ["r", "wss://brando-relay.com"],
+    ["r", "wss://expensive-relay.example2.com", "write"],
+    ["r", "wss://nostr-relay.example.com", "read"],
+  ],
+  "content": "",
+  ...other fields
+```
 
-Clients SHOULD write feed-related events created by their user to their user's write relays.
+This NIP doesn't fully replace relay lists in `kind:3`. Instead, `Kind:3` relays SHOULD serve as a fallback option for users that don't have a `kind:10002` event.
```
mikedilger, line 28: I don't think it's appropriate to reference kind:3 relay lists as if they are a thing. There is no NIP for them. I would reference it, but amongst a more general "other relay lists".  Otherwise you are putting out a SHOULD that cannot be properly coded against since nobody knows what kind:3 relay lists are (no NIP for them).

So maybe this:  "This NIP doesn't fully replace relay lists that are designed to configure a client's usage of relays (such as kind:3 style relay lists).  Clients MAY substitute other relay lists in situations where a kind:10002 relay list cannot be found."

65.md
```diff
@@ -4,73 +4,56 @@ NIP-65
...
 
-### Purposes
-
-The purpose of this NIP is to help clients find the events of the people they follow, to help tagged events get to the people tagged, and to help nostr scale better.
-
-### Suggestions
+## Final Considerations
 
-It is suggested that people spread their kind `10002` events to many relays, but write their normal feed-related events to a much smaller number of relays (between 2 to 6 relays). It is suggested that clients offer a way for users to spread their kind `10002` events to many more relays than they normally post to.
+1. Clients should guide users to keep `kind:10002` lists small (2-4 relays).
```
mikedilger, line 53: Not so sure about 2-4.  I think the low number should be 2, to imply redundancy.  Ideally the high number would be 4 as you say, but in the current environment it's safer to post to more relays so that clients that don't support NIP-65 are more likely to hear you.  I use 8 at the moment.

65.md
```diff
@@ -4,73 +4,56 @@ NIP-65
...
-The purpose of this NIP is to help clients find the events of the people they follow, to help tagged events get to the people tagged, and to help nostr scale better.
-
-### Suggestions
+## Final Considerations
 
-It is suggested that people spread their kind `10002` events to many relays, but write their normal feed-related events to a much smaller number of relays (between 2 to 6 relays). It is suggested that clients offer a way for users to spread their kind `10002` events to many more relays than they normally post to.
+1. Clients should guide users to keep `kind:10002` lists small (2-4 relays). 
 
-Authors may post events outside of the feed that they wish their followers to follow by posting them to relays outside of those listed in their "Relay List Metadata".  For example, an author may want to reply to someone without all of their followers watching.
+2. Clients should spread an author's `kind:10002` events to as many relays as viable, all of them if possible.
```
mikedilger, line 55: Not sure about "all of them."  I use 54 relays that seem to actually operate properly, among 475 entries in my list (some of which are bugs)

65.md
```diff
@@ -4,73 +4,56 @@ NIP-65
...
+## Final Considerations
 
-It is suggested that people spread their kind `10002` events to many relays, but write their normal feed-related events to a much smaller number of relays (between 2 to 6 relays). It is suggested that clients offer a way for users to spread their kind `10002` events to many more relays than they normally post to.
+1. Clients should guide users to keep `kind:10002` lists small (2-4 relays). 
 
-Authors may post events outside of the feed that they wish their followers to follow by posting them to relays outside of those listed in their "Relay List Metadata".  For example, an author may want to reply to someone without all of their followers watching.
+2. Clients should spread an author's `kind:10002` events to as many relays as viable, all of them if possible. 
 
-It is suggested that relays allow any user to write their own kind `10002` event (optionally with AUTH to verify it is their own) even if they are not otherwise subscribed to the relay because
+3. `kind:10002` events should only be used to advertise relays. The user's configuration when using Clients should stay in Kind 3.
```
mikedilger, line 57: Again the kind 3 reference.

65.md
```diff
@@ -4,73 +4,56 @@ NIP-65
...
-  "tags": [
-    ["r", "wss://alicerelay.example.com"],
-    ["r", "wss://brando-relay.com"],
-    ["r", "wss://expensive-relay.example2.com", "write"],
-    ["r", "wss://nostr-relay.example.com", "read"],
-  ],
-  "content": "",
-  ...other fields
-```
+4. DMs should be broadcasted to the author's WRITE relays and to the receiver's READ relays to keep maximum privacy.
```
mikedilger, line 59: This is fine for now.  I think I'll do a separate PR that adds a "dm" marker which is like "write" but is specific to DMs.  This PR can just be about clarification.

mikedilger (commented):
More feedback

65.md
mikedilger: Hang on I didn't read it right.  Why should DMs go to the authors WRITE relays?  So the author can read them back? Maybe.  In either case, this doesn't do anything that improves privacy, it is puttng the DMs everywhere you might put anything. So I wouldn't say "to keep maximum privacy."

65.md
vitorpamplona: Not only by them, right? By anybody. 

The idea I was trying to convey in the latter part, is that this NIP tells other users where to broadcast events that involve the author, which is a novelty of this NIP. We don't have that logic anywhere else in Nostr. It's not about them finding it. It's about how other users should behave if they implement this NIP. The finding is a consequence of that behavior.

65.md
vitorpamplona: Changed

65.md
vitorpamplona: Removed

65.md
vitorpamplona: Fixed

65.md
vitorpamplona: Is there any way we can generalize this to private events (like a private zap) and not only DMs?

mikedilger:
This NIP is now more prescriptive and less declarative, but it doesn't quite cover what the more declarative NIP did.  In particular I think it should be said that:

```
In general, when writing an event of any kind and for any purpose, clients MUST write the event to at least all of the user's write relays.

In general, when reading an event of any kind and for any purpose, clients MUST read the event from at least some of the user's read relays.
```

...and then go on with what else to do in each situation, such as when tagging a user write to their read relays.

The reason I prefer this is that I'm not confident that I can think through all of the situations, and new situations might arise we didn't think of.  So making these more general statements is safer than giving 3 or 4 specific cookbook recipes.

vitorpamplona:
Sorry, what do you mean by "reading" and event in this sentence? 

> In general, when reading an event of any kind and for any purpose, clients MUST read the event from at least some of the user's read relays.

I don't get this one. The "Read" relays are not supposed to have events from the author, because they write to the WRITE relays. 

If the app doesn't know who the author is and has not other information but the event ID, it should read from all relays the app has access to, from all authors. If it does know either the author or the p tags, it should use the WRITE relays for author and the READ relays for pTags. 

Or am I missing something?

vitorpamplona:
> In general, when writing an event of any kind and for any purpose, clients MUST write the event to all of the user's write relays. 

This is exactly the same to me as what's written there: 

> When broadcasting an event, Clients SHOULD:
>  - Broadcast the event to the WRITE relays of the author

But it misses a huge component which is the p tags. So, the next line is also there: 
> - Broadcast the event all READ relays of each tagged user.

mikedilger:
Yes you are right.  However, did we say that when you want to find events that tag you for your inbox, you should find them on your read relays?   I'm just trying to cover all the bases.  Sorry for the duplication.

mikedilger:
Oh, and relays that support this nip MUST accept and serve kind 10002 events for everybody whether they are paying customers or not.

vitorpamplona:
> tag you for your inbox, you should find them on your read relays?

I think so. That's the second line here: 

> When seeking events **from** a user, Clients SHOULD use the WRITE relays of the user's kind:10002
> When seeking events **about** a user, where the user was tagged, Clients SHOULD use the READ relays of the user's kind:10002

mikedilger:
I think @staab and @pablof7z might want to take a look at this. They've been the most vocal about this NIP in the past.

jb55:
This is much better

65.md
```diff
@@ -4,73 +4,58 @@ NIP-65
...
+2. Clients SHOULD spread an author's `kind:10002` events to as many relays as viable. 
 
-Authors may post events outside of the feed that they wish their followers to follow by posting them to relays outside of those listed in their "Relay List Metadata".  For example, an author may want to reply to someone without all of their followers watching.
+3. `kind:10002` events MUST only be used to advertise the user's preferred relays to others. The Client's configuration SHOULD stay in other lists where other factors can be considered, such as specific relays for a given UI feature of the Client.
 
-It is suggested that relays allow any user to write their own kind `10002` event (optionally with AUTH to verify it is their own) even if they are not otherwise subscribed to the relay because
+4. DMs SHOULD only be broadcasted to the author's WRITE relays and to the receiver's READ relays to keep maximum privacy.
```
fiatjaf, line 59: I think we need some other special rule for DMs. I think ideally DMs would only go to a subset of the relays defined by the receiver, not the sender, so that would be just to _the receiver's READ relays_.

However, even that is not enough, because I may be perfectly willing to read public notes on some relays, but not DMs. For DMs specifically I may want to use one specific relay that I trust more with my metadata, for example.

Maybe add a different marker `"dm-read"` and fallback to the `"read"`?

65.md
```diff
@@ -4,73 +4,58 @@ NIP-65
...
-  "tags": [
-    ["r", "wss://alicerelay.example.com"],
-    ["r", "wss://brando-relay.com"],
-    ["r", "wss://expensive-relay.example2.com", "write"],
-    ["r", "wss://nostr-relay.example.com", "read"],
-  ],
-  "content": "",
-  ...other fields
-```
+5. Relays that support this NIP MUST accept and serve kind 10002 events for everybody whether they are paying customers or not.
```
fiatjaf, line 61: Is this line necessary? I think relay owners shouldn't be forced to accept a load of spam just because it has kind 10002 and they probably won't accept it, so better not have this line here as it may cause more confusion than help.

65.md
vitorpamplona: @mikedilger was adding DMs on the other PR. Should I add them here as well? 

Maybe DM is the wrong word. I can imagine a special relay that third parties can post to it but only the author can download from it.

65.md
vitorpamplona: I was fine with it because of the IF. IF the relay declares support for NIP-65, it MUST accept it. Relays can choose to not declare support.

mikedilger: This nip used to explain to relay operators why this isn't too big of a deal, but that disappeared from Vitor's rewrite:

- finding where someone posts is rather important
- these events do not have content that needs management
- relays only need to store one replaceable event per pubkey to offer this service

If someone tries to store a kind-10002 that has any content at all, relays are free to (and I suggest that they do) reject such events.

65.md
mikedilger: This is a bit trickier because you have to read back your own DMs from somewhere to see the full conversation.

vitorpamplona: Your own DMs from WRITE + replies from PROTECTED_READ_RELAYS

vitorpamplona: Maybe protected DM relays are the subject of a new PR when we better understand them.

mikedilger: Agreed, let's punt on the special DM stuff for now, take one step at a time.

fiatjaf: Agreed.

65.md
```diff
@@ -4,73 +4,58 @@ NIP-65
...
+    ["r", "wss://brando-relay.com"],
+    ["r", "wss://expensive-relay.example2.com", "write"],
+    ["r", "wss://nostr-relay.example.com", "read"],
+  ],
+  "content": "",
+  ...other fields
+```
 
-Clients SHOULD write feed-related events created by their user to their user's write relays.
+This NIP doesn't fully replace relay lists that are designed to configure a client's usage of relays (such as `kind:3` style relay lists). Clients MAY use other relay lists in situations where a `kind:10002` relay list cannot be found.
```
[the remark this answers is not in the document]
staab: Yes, but why mention kind 3? Just say "clients can try their best if the data isn't available".

65.md
```diff
@@ -4,73 +4,58 @@ NIP-65
...
+## Final Considerations
 
-### Suggestions
+1. Clients SHOULD guide users to keep `kind:10002` lists small (2-4 relays). 
 
-It is suggested that people spread their kind `10002` events to many relays, but write their normal feed-related events to a much smaller number of relays (between 2 to 6 relays). It is suggested that clients offer a way for users to spread their kind `10002` events to many more relays than they normally post to.
+2. Clients SHOULD spread an author's `kind:10002` events to as many relays as viable. 
 
-Authors may post events outside of the feed that they wish their followers to follow by posting them to relays outside of those listed in their "Relay List Metadata".  For example, an author may want to reply to someone without all of their followers watching.
+3. `kind:10002` events MUST only be used to advertise the user's preferred relays to others. The Client's configuration SHOULD stay in other lists where other factors can be considered, such as specific relays for a given UI feature of the Client.
```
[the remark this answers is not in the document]
staab: Fair enough, I take a different approach, but yours is valid. Where do you store these settings? Just in the app?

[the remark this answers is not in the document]
vitorpamplona: Yep, now we just store it locally. I didn't think it would be useful for anyone else

65.md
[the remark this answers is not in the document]
vitorpamplona: I like being specific. People know kind3 even if it is not specified as a NIP. Abstracting it just raises more questions (e.g. is there another Kind3-like type I don't know about?). We just need to help them wrap their minds around this NIP65 idea and how it integrates with their existing way of doing business.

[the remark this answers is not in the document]
vitorpamplona: But I can remove kind 3 if people want.

65.md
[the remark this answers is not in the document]
staab: Ok, then what about something like this:

```suggestion
3. `kind:10002` events should primarily be used to advertise the user's preferred relays to others. A user's own client may use other heuristics for selecting relays for fetching data.
```

vitorpamplona:
I think this is final. Let's merge if no one has any additional questions.

65.md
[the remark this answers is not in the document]
staab: Votes? one for, one against so far @mikedilger be our tie breaker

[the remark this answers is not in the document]
mikedilger: I like the consistency of not mentioning something not NIPped else some people will be confused. But I also like the specificity of mentioning it so people understand the main de-facto alternative we are thinking about.  So I'm not going to be a good tie-breaker here.

mikedilger:
+1 for merging once the kind-3 issue is resolved (one way or the other, I'm not too fussed)

65.md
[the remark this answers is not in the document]
staab: lol ok @fiatjaf tell us what to do

fiatjaf:
I think we can remove the kind 3 mention after this NIP has seen more adoption, for now we can leave it to help people that are using kind 3 understand.

PR #703, opened 2023-08-08
merge nips 12, 16, 20 and 33 into nip 01

fiatjaf:
also get rid of the tag prefix matching since no one is using it and it introduces unnecessary complication and slowness for relay implementations and basically no benefit.

https://github.com/nostr-protocol/nips/blob/nip01-refactor/01.md


mattn:
Yes, agree with @Semisol . NIP-11 Relay Information includes the NIP numbers. If that will be reused, the numbers might have wrong meanings.

README.md
```diff
@@ -85,7 +81,7 @@ They exist to document what may be implemented by [Nostr](https://github.com/nos
 | `6`     | Repost                     | [18](18.md) |
 | `7`     | Reaction                   | [25](25.md) |
 | `8`     | Badge Award                | [58](58.md) |
-| `16`    | Generic Repost             | [18](18.md) |
+| `18`    | Generic Repost             | [18](18.md) |
```
vitorpamplona, line 84: Are we changing the event kind of Generic Repost to 18? Looks like a simple error in the PR.

01.md
```diff
@@ -76,40 +122,36 @@ Upon receiving a `REQ` message, the relay SHOULD query its internal database and
...
 
-The `ids` and `authors` lists contain lowercase hexadecimal strings, which may either be an exact 64-character match, or a prefix of the event value.  A prefix match is when the filter string is an exact string prefix of the event value.  The use of prefixes allows for more compact filters where a large number of values are queried, and can provide some privacy for clients that may not want to disclose the exact authors or events they are searching for.
+The `ids` and `authors` lists contain lowercase hexadecimal strings, which must be an exact 64-character match, or a prefix of the event value.  A prefix match is when the filter string is an exact string prefix of the event value.  The use of prefixes allows for more compact filters where a large number of values are queried, and can provide some privacy for clients that may not want to disclose the exact authors or events they are searching for.
```
vitorpamplona, line 125: Should MUST be in uppercase?

vitorpamplona:
Looks good. My only suggestion is to generally avoid text adjustments in "re-organizing" PRs. Recognize first and then do a new PR for the "may" - > "must" changes.  It's just easier to review. 

But the changes here look good.

README.md
fiatjaf: Oops, I got confused thinking this was the NIP number, not the kind number, will fix.

AsaiToshiya (commented):
Is it intentional that `2`: `recommend_server` was removed?

fiatjaf:
Yes. No one is using it as far as I know, and it wasn't a good idea anyway. Removed to reduce confusion.

ariard:
I’m not sure this is best practice of technical standard effort to merge all the documents into a single piece of information.

IETF RFC (IP, TCP, BGP, etc) are issuing new documents with reference to “obsolete” previous versions or “updates” and “erratas”. Same with BIP, once they’re final we don’t really touch to them beyond errata, and new extensions or deprecations are announced with new document (cf. the history of BIP37 bloom filters), I can’t remember a BIP which has been updated in place.

About the Lightning BOLTs, there is an aim to maintain them more as a “living document” where they the major ones are updated when it’s minor or medium modifications, or issue “extension-BOLT” for new sub-protocol (e.g Taproot channels). Though here there is a lot and back and forth on the BOLT-side, if some part should be optional (e.g BLIPs), safety and security recommendations, part of the core protocol though not mandatory.

About the current changes, I do have few questions:
- does it make nip 12, 16, 20 and 33 mandatory part of “the” nip protocol ?
- if yes, has this change been discussed in public on a community communication channel and if yes which one ?
- have you survey the relays and clients before to make the move, or is there is some inter-compatibility threshold (e.g a NIP becomes mandatory when more than X% of relays and clients with Y% of network volume supporting it ?)
- have you considered a time point from which the new NIP part of the mandatory set are “required” and therefore clients or relays can disconnect the faultive software ?

On the changes themselves, NIP-33, NIP-16, NIP-20 were already on my list of foundational NIP to implement. NIP-12 makes sense to me as generic tag query sounds to underpin whatever is your use-case.

Personally, I would advocate to update the all NIPs (12, 16, 20 and 33) and add the mention “mandatory” and update the front README.md with the list of mandatory NIPs.

README.md
AsaiToshiya, line 1: kind `2` need to be removed from the list.

fiatjaf:
@ariard this is a public channel as much as any other. There are no processes or best practices, I'm just trying to make this thing simpler to understand and easier for new implementors. Having everything in NIP-01 is better for that reason.

From my informal research all the new "required" things only need to be required by relays, not clients, and all the relay implementations I know about already implement it and they are assumed to be implemented everywhere by all clients anyway, so nothing should be changing.

There is no need to disconnect from faultive software, since that is in no one's interest since you can still benefit from connecting with others even if they have a bug in their implementation -- also other people having bugs in most cases won't cause you any harm. Yet all clients and relays already have to watch and guard against peers that may flooding them or returning invalid data etc, so I think the spec shouldn't talk about that.

Ultimately this repository has no power over implementations, it is just a discussion board for people to agree on standards, but if people create standards from the outside this repository will follow them, not the opposite. So we don't have to be super bureaucratic.

> On the changes themselves, NIP-33, NIP-16, NIP-20 were already on my list of foundational NIP to implement. NIP-12 makes sense to me as generic tag query sounds to underpin whatever is your use-case.

Exactly, and that's what most people think too. If these things are in NIP-01 that saves people from having to read all the NIPs which in most cases have nothing to do with what they want to implement. Thus this change.

AsaiToshiya:
Just FYI:
https://gist.github.com/AsaiToshiya/f0d985eea5835184be302ed61faec134
https://nostr.watch/relays/statistics

ariard:
> There are no processes or best practices, I'm just trying to make this thing simpler to understand and easier for new implementors.

Might I suggest in the future the nostr-protocol@googlegroups.com as documented in the README.md could be a good announcement location.

> From my informal research all the new "required" things only need to be required by relays, not clients, and all the relay implementations I know about already implement it and they are assumed to be implemented everywhere by all clients anyway, so nothing should be changing.

Yes I think this is sufficient informal research at this stage of maturity of the Nostr ecosystem (though in the future the OTS commitment of the notes themselves and the number of followers can be good means to judge of popularity of a software). 

One good practice for relays can be to document the implemented NIP, and if any the divergence like Bitcoin Core is doing for the BIPs: https://github.com/bitcoin/bitcoin/blob/master/doc/bips.md

> There is no need to disconnect from faultive software, since that is in no one's interest since you can still benefit from
> connecting with others even if they have a bug in their implementation -- also other people having bugs in most cases
> won't cause you any harm.

In most cases they won’t cause you any harm - yes unless it’s a source of DoS concerns or scarce inbound slots. If I remember correctly, Bitcoin Core is disconnecting / discouraging faultive peers because of a) the network inputs are DoSy (see `Misbehaving()` in `net_processing.cpp`) so risk for the full-node and b) the implementation is non-behaving well and inbound slots is a scarce resource better allocated to effective peers (e.g the ones announcing compact blocks well).

All proportions conserved as the p2p stack of Bitcoin Core is far more mature. Though I think one good practice could be to version the nostr protocol itself with a future NIP, therefore easing future DoS mitigations or deprecation.

> Ultimately this repository has no power over implementations, it is just a discussion board for people to agree on standards, but if people create standards from the outside this repository will follow them, not the opposite. So we don't have to be super bureaucratic.

Yeah all good to avoid being super bureaucratic. IMHO the BOLT process is bit too much bureaucratic. The BIP process is just pure chaos, which is somehow a good thing. “Rough consensus and running code”, especially running code.

> Exactly, and that's what most people think too. If these things are in NIP-01 that saves people from having to read all the NIPs which in most cases have nothing to do with what they want to implement. Thus this change.

I’ll look on implement the new mandatory NIPs (12, 16, 20, 33), I just hope the documentation will age well for people implementing new relays and clients from scratch years from now. ACK on the changes.

01.md
```diff
@@ -43,9 +42,57 @@ To obtain the `event.id`, we `sha256` the serialized event. The serialization is
...
+- for kind `n` such that `10000 <= n < 20000`, events are **replaceable**, which means that, for each combination of `pubkey` and `kind`, only the latest event is expected to be stored by relays, older versions are expected to be discarded.
+- for kind `n` such that `20000 <= n < 30000`, events are **ephemeral**, which means they are not expected to be stored by relays.
+- for kind `n` such that `30000 <= n < 40000`, events are **parameterized replaceable**, which means that, for each combination of `pubkey`, `kind` and the `d` tag, only the latest event is expected to be stored by relays, older versions are expected to be discarded.
+
+These are just conventions and relay implementations may differ.
+
 ## Communication between clients and relays
 
-Relays expose a websocket endpoint to which clients can connect.
+Relays expose a websocket endpoint to which clients can connect. Clients should not open more than one websocket to each relay. One channel can support an unlimited number of subscriptions.
```
jb55, line 91: The wording on this is a bit confusing. What does it mean that `one channel can support an unlimited number of subscription`. People might think this means they can open an unlimited amount of subscriptions without any problems?

01.md
```diff
@@ -43,9 +42,57 @@ To obtain the `event.id`, we `sha256` the serialized event. The serialization is
...
+- `1`: **text note**: the `content` is set to the **plaintext** content of a note (anything the user wants to say). Content that must be parsed, such as Markdown and HTML, should not be used. Clients should also not parse content as those.
+
+And also a convention for kind ranges that allow for easier experimentation and flexibility of relay implementation:
+
+- for kind `n` such that `1000 <= n < 10000`, events are **regular**, which means they're all expected to be stored by relays.
+- for kind `n` such that `10000 <= n < 20000`, events are **replaceable**, which means that, for each combination of `pubkey` and `kind`, only the latest event is expected to be stored by relays, older versions are expected to be discarded.
+- for kind `n` such that `20000 <= n < 30000`, events are **ephemeral**, which means they are not expected to be stored by relays.
+- for kind `n` such that `30000 <= n < 40000`, events are **parameterized replaceable**, which means that, for each combination of `pubkey`, `kind` and the `d` tag, only the latest event is expected to be stored by relays, older versions are expected to be discarded.
```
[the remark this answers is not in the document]
alexgleason: I want to know what @arthurfranca thinks about this one.

01.md
fiatjaf: This text was already on the NIP, I just moved it here from below, but you're right. Let me try to improve it.

01.md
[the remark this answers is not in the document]
fiatjaf: I don't think there is a better place for them. The ranges are pretty much part of the core protocol now, pretty simple to understand and very important to know.

01.md
```diff
@@ -76,40 +122,36 @@ Upon receiving a `REQ` message, the relay SHOULD query its internal database and
...
 
-The `ids` and `authors` lists contain lowercase hexadecimal strings, which may either be an exact 64-character match, or a prefix of the event value.  A prefix match is when the filter string is an exact string prefix of the event value.  The use of prefixes allows for more compact filters where a large number of values are queried, and can provide some privacy for clients that may not want to disclose the exact authors or events they are searching for.
+The `ids` and `authors` lists contain lowercase hexadecimal strings, which MUST be an exact 64-character match, or a prefix of the event value.  A prefix match is when the filter string is an exact string prefix of the event value.  The use of prefixes allows for more compact filters where a large number of values are queried, and can provide some privacy for clients that may not want to disclose the exact authors or events they are searching for.
```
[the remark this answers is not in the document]
fiatjaf: I'm stupid.

01.md
```diff
@@ -32,7 +31,7 @@ The only object type that exists is the `event`, which has the following format
 
 To obtain the `event.id`, we `sha256` the serialized event. The serialization is done over the UTF-8 JSON-serialized string (with no white space or line breaks) of the following structure:
 
-```json
+```jsonc
```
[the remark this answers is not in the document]
fiatjaf: Just changed to `json`. Let's love the red background.

01.md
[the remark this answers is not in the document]
viktorvsk: Well, if we are to reorganize kinds, I'd like to suggest what bothers me from day 1 :)
What about making default kinds ranges limited to `32767` ? It's `40000` so not a big difference. And it's easily extendable with future NIPs. But at this point it would allow us to have 2 bytes to store kind instead of 4

Maybe I'm missing something implementation-wise here but I wasn't able to find an easy workaround to store currently existing kinds in 2-byte-way in postgres

viktorvsk:
I personally like the idea to merge NIPs even if it causes some inconvenience for existing implementations because it makes things simpler for future and it seems like now "we can afford it still", deprecations and obsoletes should become a thing once its all more mature.

However, I'd like to second that it would be great to separate NIPs reorganization and text changes (maybe not for this time but for future) — so that nothing would accidentally slip

20.md
```diff
@@ -1,93 +1,9 @@
...
-
-The `message` SHOULD start with `invalid:` if the event is invalid or doesn't meet some specific criteria (created_at is too far off, id is wrong, signature is wrong, etc)
-
-The `message` SHOULD start with `pow:` if the event doesn't meet some proof-of-work difficulty. The client MAY consult the relay metadata at this point to retrieve the required posting difficulty.
-
-The `message` SHOULD start with `rate-limited:` if the event was rejected due to rate limiting techniques.
-
-The `message` SHOULD start with `error:` if the event failed to save due to a server issue.
-
-Ephemeral events are not acknowledged with OK responses, unless there is a failure.
```
viktorvsk, line 32: Is this not a thing anymore? It would be rather good because it seems to be good to actually acknowledge NIP-42 AUTH event

fiatjaf: I didn't remember this, or maybe never realized it was there. What was the reasoning?

viktorvsk: I don't know but I would expect something like avoiding traffic generated by events like typing... indicator. Good if we remove this line

13.md
```diff
@@ -110,7 +110,7 @@ function countLeadingZeroes(hex) {
 Querying relays for PoW notes
 -----------------------------
 
-Since relays allow searching on prefixes, you can use this as a way to filter notes of a certain difficulty:
+If relays allow searching on prefixes, you can use this as a way to filter notes of a certain difficulty:
```
viktorvsk, line 113: If we are to get rid of prefix search completely I would suggest to remove this line.
Instead, to filter notes by difficulty relays may offer NIP-50 style i.e `{search: "min_difficulty:10"}`

fiatjaf: Good idea. I tried to do POW-based prefix matching on ZBD Social for filtering high quality replies, but it didn't make any difference since barely anyone is publishing POW notes and was probably just abusing relays for no reason. This would allow relays that really want to be hubs of POW content like powrelay.xyz to implement proper support for it instead of generic prefix matching.

01.md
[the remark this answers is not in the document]
arthurfranca: @alexgleason [my idea](https://github.com/nostr-protocol/nips/pull/683#issuecomment-1652239180) was to mark some NIPs as "required" exactly how @ariard said above: "Personally, I would advocate to update all the integrated NIPs (12, 16, 20 and 33) and add the mention “mandatory” and update the front README.md with the list of mandatory NIPs." But you can check on the previous link what the response was, at least it was longer than [this one](https://github.com/nostr-protocol/nips/pull/696#issuecomment-1667879114).

Now I keep my opinions to myself until something like [nips over nostr](https://github.com/nostr-protocol/nips/pull/704) ever becomes a reality.

jb55:
On Fri, Aug 11, 2023 at 04:39:47AM -0700, Viktor Vsk wrote:
>@viktorvsk commented on this pull request.
>-Relays MUST return `false` when the event was rejected and not saved.
>-
>-The `message` SHOULD provide additional information as to why the command succeeded or failed.
>-
>-The `message` SHOULD start with `blocked:` if the pubkey or network address has been blocked, banned, or is not on a whitelist.
>-
>-The `message` SHOULD start with `invalid:` if the event is invalid or doesn't meet some specific criteria (created_at is too far off, id is wrong, signature is wrong, etc)
>-
>-The `message` SHOULD start with `pow:` if the event doesn't meet some proof-of-work difficulty. The client MAY consult the relay metadata at this point to retrieve the required posting difficulty.
>-
>-The `message` SHOULD start with `rate-limited:` if the event was rejected due to rate limiting techniques.
>-
>-The `message` SHOULD start with `error:` if the event failed to save due to a server issue.
>-
>-Ephemeral events are not acknowledged with OK responses, unless there is a failure.
>
>I don't know but I would expect something like avoiding traffic generated by events like typing... indicator. Good if we remove this line

because ephemeral events aren't written. Command results were
originally conceived for determining if the command was written to the
database or not. I guess it might be useful to know if an ephemeral
event was relayed?

01.md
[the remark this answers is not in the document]
alexgleason: Also, isn't kind 3 a replaceable event? And kind 0.

PR #704, opened 2023-08-08
nips over nostr

Semisol:


staab:
NACK, I don't think we should switch away from git, at least for the time being. There needs to be some form of canonicity, even if it's centrally managed. Imagine if all 133 of our current PRs were NIPs?

I would only start to consider this after:

- It accounted for votes from the community, or some form of consensus
- It allowed the owners of NIP 89 clients to claim they've implemented a NIP
- There was a reliable client for drafting, editing, browsing, and voting on NIPs

gazhayes:
> Maybe nostrocket can achieve this.

Yes, Nostrocket can solve this in a few different ways and it's something I've been thinking about for a while now. 

But I don't know if voting per se is a good idea here, I think an evolution of the fork-pull-maintainer model is more appropriate.

1. Any pubkey MAY propose a new NIP, as a solution to an observed problem

2. Any Maintainer MAY merge a new NIP if it complies with the rules and standards defined in the NIPPLE (Nostr Improvement Possibility Protocol for Lightweight Experimentation)

3. Any pubkey MAY propose changes to a proposed or existing NIP by producing a Patch, in response to an observed problem.

4. Any Maintainer MAY merge a new Patch to an existing or a proposed NIP if it does not violate the NIPPLE.

5. Maintainers exist in a graph. Any existing Maintainer can add a new Maintainer. Whoever invites someone to the Maintainer tree can also remove them or anyone below them if they turn out to be a bad actor, or private key becomes compromised etc. 

There are currently explicit and implicit rules/standards regarding NIPs. This is what should be formalized in the NIPPLE. 

What happens at the moment is that as we encounter new NIPs, we are discovering what the minimum and simplest set of these rules and standards need to be such that the NIP repo doesn't become a clusterfuck. 

This currently happens ad-hoc (e.g. the first rule fiatjaf created was NIPs "must be implemented somewhere", which then became "must be implemented in two places", etc).

If we are going to formalize things into a process, we need a NIPPLE that:
1. defines who comes up with the rules/standards (and how), and
2. explicitly defines the current state of these rules and standards 

I suspect that this NIPPLE can simply be a NIP that is created and updated by following its own rules (the same process as a "normal" NIP).

> Such unorganizations will converge on one mindset and anything that disagrees with it will get rejected.

I want to understand why you think this. Can you elaborate either here or OOB?

fiatjaf:
I don't get this versioning scheme. Instead of referencing a version number shouldn't we reference event ids?

vitorpamplona:
+1 for moving the NIP repo away from GitHub. 
-1 for using this particular design. 

I don't think we need an event kind specific to NIPs. We should generalize it a bit to allow any versioned document. NIPs are just our own use case of it. Then we can build other event kinds to allow collaboration over the document writing process (drafts, suggestions, reviews, discussions, reactions, zaps, etc). 

Then we need a "Publisher". A pubkey that collects all "approved versions" of those documents together in a single "website". We can have multiple publishers as well. They are just collections of versions of documents anyway. 

Or we flip it around and each client is a "NIP Publisher" that "merges" specific versions of NIPs from any author out there into the Publisher's book of NIPs supported by that client. 

The most important first step is to have versioned events, which can be exactly the same as replaceable events, but relays don't delete past events when a new one arrives. We could use a new range for this design.

Giszmo:
NACK. The problem with migrating nips from GitHub to nostr is not with the format how it's stored. We can use long form posts like habla.news does, with tags. The problem is to communicate what is a community supported set of nips, the discussion around each proposal and discoverability.

If you are so inclined, you can propose new nips on nostr already. What would happen if you solved a problem and others with the same problem followed your recipe outside of the canonical nips? It would just work and others would consider it canon and probably add it to GitHub, too.

A new format wouldn't help with that. In fact it would make it worse.

So if you want to move nips to nostr, just do. Write your next nip on habla.news.

vitorpamplona:
> it doesn't seem that all worth it after all the feedback

I think it is worth it. It just requires a lot of development. But somebody has to start it :) 

The goal shouldn't be to simply replace GitHub. It should be to be 10x better at collaborating on NIPs than GitHub. Then the migration becomes a no brainer.

fiatjaf:
I agree with @mikedilger that the metadata and the ability for people to signal their adherence and support to NIPs is very important for this to not become useless chaos.

I always imagined decentralized NIPs as a way for people to follow a set of standard curators, but at the same time make those flexible.

There was also a proposal from @giszmo a while ago that had some interesting lessons to be learned. I can't find it anymore.

Giszmo:
@fiatjaf thanks for the nod. I took it as "could you please guess what I was thinking of ..."

How about [How to build a decentralized Wikipedia?](https://habla.news/u/leo@nostr.info/howToConsolidateLongFormEn)? It's also suggesting @Semisol 's versioned parameterized replaceable events and introduces endorsements to come to a very dynamic and subjective notion of a "standard". In short, fiatjaf.nip1 might be different to semisol.nip1 and semisol talking about "nip1" without specificator would make clients try semisol.nip1 first etc.

alexgleason:
I actually think this proposal is pretty good. It should coexist with the NIPs repo on GitHub. NIPs on GitHub would be seen as more essential, while NIPs on Nostr are more optional. The difference between the two is intrinsically clear.

I also think making it a separate kind is correct, since its ID has a specific meaning in that context, and they can be easily queried to display a registry of NIPs.

alexgleason:
Btw, the way people signal that a NIP is good is by implementing it. It should be added to the supported_nips field of relays, and clients should display a page of supported NIPs. You would simply implement NIPs that clients you care about also implement.

PR #686, opened 2023-07-28
NIP-17 (old 24) Sealed Gift-Wrapped Messages for Private DMs and Small Group Chats

vitorpamplona:
This PR creates a new private messaging scheme from the learnings of many past proposals, including [NIP-44](https://github.com/nostr-protocol/nips/pull/574), [NIP-59](https://github.com/nostr-protocol/nips/pull/468), [NIP-24](https://github.com/nostr-protocol/nips/pull/56), [NIP-103](https://github.com/nostr-protocol/nips/pull/499), [NIP-76](https://github.com/d-krause/nostr-nips/blob/nip76-draft-2/76.md). 

It brings their best ideas together in a slightly different approach that seems to be better for privacy.  

Read here: [NIP-17](https://github.com/vitorpamplona/nips/blob/sealed-dms/17.md)

PR #706, opened 2023-08-09
Add NIP-87 private groups

staab:
This is yet another encrypted groups draft. It follows #686's modification of #468 and splits it out into a separate NIP that can be used by both group implementations. #686 attempts to solve small groups with no moderation or key rotation, this NIP attempts to solve large groups with moderation and key rotation.

This NIP is inspired by (and very similar to) @earonesty's #580 with the following differences:

- Instead of identifying a group by event id, a group is identified instead by a pubkey.
  - This allows the admin to stay anonymous by not signing anything with his key.
  - This allows the group to publish public events on its own behalf.
  - This also allows for less stateful key sharing and validation.
  - This eliminates kinds 405 and 406, which require clients to stitch multiple groups together in a chain.
  - It ties authorized group-admin actions to event validation, which is simpler than the stateful validation used [here](https://github.com/nostr-protocol/nips/blob/88ecba85852bd91876e74dc70f37123bf1484238/112.md#wrapped-kind-402-change-channel-metadata).
- It combines 400 and 402 which are both kind 0's for the group.
- It has a broader set of supported use cases - it's not tied explicitly to chat.
- It allows members to request entry/removal, and notifies current users of a key rotation.
- It allows for multiple compatible types of moderation with no elevated permissions for moderators, based on end user preferences.
- It uses @vitorpamplona's version of wrapping, reducing the risk of leaked events.

I also have to credit @water783 for his work on [alias exchange](https://github.com/water783/nips/blob/nip101/101.md) and another [draft](https://github.com/water783/nips/blob/group-chats/xxx.md) similar to this one. The main difference is that this NIP doesn't require a monolithic member list for key sharing, instead membership is determined by possession of a key.

### DIFF
--- a/87.md
+++ b/87.md
@@ -13,12 +13,12 @@ This NIP introduces several new event kinds, and a two-tier shared key system wh
 - No admin, moderator, or member metadata leakage outside the group
 - Optionally public group metadata
 - Users may request admission or removal
-- Users may select whitelist, blacklist, or no moderation
+- Members may select whitelist, blacklist, or no moderation
 - Members and moderators may request the removal of another member
 - Shared keys can be rotated with weak forward secrecy
 - Use of wrapping allows any nostr event to be posted to the group
 
-This NIP relies on terms defined in [59.md](NIP 59) for wrapping. See that NIP for a description of the notation used below for describing wrapped events.
+This NIP relies on terms defined in [59.md](NIP 59) for wrapping.
 
 # Wrapping Notation
 
@@ -93,7 +93,7 @@ If the admin chooses to admit the new member, they may choose one of the followi
 
 ## Key Rotation Request
 
-Any group member may request a key rotation: `26:author ?->[admin_key]`. The rumor's `content` MAY include a reason for the request, and `tags` MAY include any number of `p` tags to exclude from the key rotation. These may or may not be honored by the admin, depending on whether the author's own `p` tag is included (leaving a group), or if the author is a moderator.
+Any group member may request a key rotation: `26:author shared_key->[shared_key]`. The rumor's `content` MAY include a reason for the request, and `tags` MAY include any number of `p` tags to exclude from the key rotation. These may or may not be honored by the admin, depending on whether the author's own `p` tag is included (leaving a group), or if the author is a moderator.
 
 ```json
 {
@@ -102,6 +102,8 @@ Any group member may request a key rotation: `26:author ?->[admin_key]`. The rum
   "tags": [["p", "<pubkey>"]],
 }
 ```
+
+This event is sent to the group rather than the admin so that group members (including the sender) can optimistically update their member lists. This SHOULD be limited only to removing the author's pubkey from the group.
 
 ## Members Added Notifications
 
--- a/87.md
+++ b/87.md
@@ -20,6 +20,21 @@ This NIP introduces several new event kinds, and a two-tier shared key system wh
 
 This NIP relies on terms defined in [59.md](NIP 59) for wrapping. See that NIP for a description of the notation used below for describing wrapped events.
 
+# Wrapping Notation
+
+To make descriptions of wrapped events clearer below when referring to a gift wrapped event the rumor kind, the author of the seal, the author of the  gift wrap, and the recipients are specified using this notation:
+
+`kind:author wrapper->[recipient1,recipient2]`
+
+- If using an ephemeral key, the symbol `?` should be used.
+- If using a derived key, the notation `{a,b}` should be used.
+- If using an unknown kind, use `any`.
+- If no recipients are specified, use `[]`.
+
+For example, if a kind 1 is sent by `a` to a key derived from `a`'s private key and `b`'s public key and wrapped using an ephemeral key:
+
+`1:a ?->[{a,b}]`
+
 # Protocol Description
 
 ## Admin key
--- a/87.md
+++ b/87.md
@@ -51,13 +51,18 @@ When posting to the group as a member, events MUST follow this scheme: `any:auth
 
 When posting to the group as an admin, events MUST follow this scheme: `any:admin_key shared_key->[shared_key]`.
 
-New members are invited using this scheme: `24:admin_key ?->[recipient]`. Multiple events are sent, each addressed to a single invitee directly. The rumor's `tags` MUST contain a single `privkey` tag containing the shared key and SHOULD contain one or more `r` tags to help clients find group messages.
+New members are invited using this scheme: `24:admin_key ?->[recipient]`. Multiple events are sent, each addressed to a single invitee directly. The rumor's `tags` MUST contain a single `privkey` tag containing the shared key and SHOULD contain one or more `r` tags to help clients find group messages. It MAY also include any number of `p` tags to help clients build member lists.
 
 ```json
 {
   "kind": 24,
   "content": "Just a regular key rotation, nothing to worry about, we definitely didn't leak a key",
-  "tags": [["privkey", "<new shared key>"], ["r", "wss://some-relay.com"]],
+  "tags": [
+    ["privkey", "<new shared key>"],
+    ["r", "wss://some-relay.com"],
+    ["p", "<pubkey>"],
+    ["p", "<pubkey>"]
+  ],
 }
 ```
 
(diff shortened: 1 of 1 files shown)


vitorpamplona:
NIP numbers in draft proposals are getting ridiculous :)

vitorpamplona:
Maybe the right name for this is Trusted Groups. Since any member can add any stranger, and any stranger can make the entire group visible, privacy is directly dependent on your trust in the behavior of other members. And maybe privacy is not an actual expectation of the group, which simplifies requirements a lot. 

For instance, the Facebook group with 8M people might be closed/private, but everyone knows the admin will not be able to review people when joining. So, I can't say bad things about.. say my family... in the group. They might join in the future. And that will be bad.  

On the other side, smaller Facebook groups do have a new member evaluation procedure when joining. The admin puts up some rules and evaluates people on them. It's also not super private. If the group is 3000 like-minded people and the admin is kicking out everyone that misbehaves (or doesn't share the same mindset), I might say things the "other side" might not like in peace. Even if "the other side" people join, I don't care about their opinions anyway. 

Those two groups require different implementations in my mind.

staab:
Re-organized the description, added some examples, and a notation for wrapped events that makes this kind of thing easier to read.

> Those two groups require different implementations in my mind.

Why? It seems like this NIP handles both group types fine, what do you have in mind. #686 doesn't prevent doxxing the entire group either, like @earonesty says, that's always a risk, the best you can do is forward secrecy so that different sub groups across time can't leak each other. To be clear, I think #686 is a good group chat implementation, and should co-exist with something like this.

vitorpamplona:
There is no way to dox the group in a verifiable manner on https://github.com/nostr-protocol/nips/pull/686 without leaking people's main keys. There is an incentive to protect the main key and thus the group, because the only way to leak is to leak ALL your conversations. Of course, people can always screenshot the group. But no one can publically verify the screenshot is true. 

Here, any group member can leak the chat's secret, and anyone can verify those messages are valid. There is no incentive to protect the secret.

vitorpamplona:
>> Those two groups require different implementations in my mind.
> Why? 

Mindset. People behave differently in a group where the admin is checking everyone before allowing them in. The level of that check defines the privacy requirements of the group. 

But, more importantly, users have a trusted relationship with the group admin, not with every other member. Here, the trust is with every other member. Which I think is a use case, but it doesn't cover the other one.

staab:
> There is no way to dox the group in a verifiable manner on https://github.com/nostr-protocol/nips/pull/686

Got it, yes. I looked into sharding the shared key for this PR, but any level or shared key makes the key disposable without threatening the security of the member's own private key. So it's extra complexity for no benefit.

How would you see "admin-moderated" groups working? Members submit via DM to admin, admin strips the signature to create a rumor and re-publishes to the group? That would mean 100% trust in the admin not to omit/falsify anything, so it seems like a bad design.

Keeping the name "encrypted groups" since it doesn't imply "perfectly private", and "trusted" seems to imply that it's protected by a relay.

vitorpamplona:
> I looked into sharding the shared key for this PR, but any level or shared key makes the key disposable without threatening the security of the member's own private key. So it's extra complexity for no benefit.

I don't know if it's even needed. As long as people know anyone can add and remove anyone from the group, I think the privacy mindset is established: it's private until somebody fucks up and invite the wrong people. 

> How would you see "admin-moderated" groups working? 

Also known as "it's private until admins fuck up and invite the wrong people". I think this one is very close to Moderated Communities. There is one Owner. Owner adds moderators. And Moderators and add or remove people from the group. @water783`s proposal https://github.com/water783/nips/blob/group-chats/xxx.md maybe able to require "moderation" role to rotate the key. 

> "trusted" seems to imply that it's protected by a relay.

does it? I never heard something like that before.

staab:
> > "trusted" seems to imply that it's protected by a relay.
> 
> does it? I never heard something like that before.

I'm thinking here of something like NIP 38

> As long as people know anyone can add and remove anyone from the group

Apart from #686's approach, how would you avoid that? A decryption key will always exist, and can always be leaked.

vitorpamplona:
> Apart from https://github.com/nostr-protocol/nips/pull/686's approach, how would you avoid that? A decryption key will always exist, and can always be leaked.

Maybe a Shamir's secret sharing where you can do M out of N decryption? 

We need cryptographers for that.

staab:
> Maybe a Shamir's secret sharing where you can do M out of N decryption?
>
> We need cryptographers for that.

Yes indeed 😅

I'm going to go ahead and hide our conversation to keep things focused here if that's ok with you. I'll leave a quick summary too.

staab:
I hid some comments to keep things focused. The summary is there are three kinds of groups:

- Small groups (#686)
- Large groups where members trust each other not to reveal their membership (or where membership is not sensitive information)
- Large groups where membership is protected by admins requiring Shamir's secret sharing where you can do M out of N decryption (?)

This PR targets use case #2.

staab:
@paulmillr I've ported over the essentials from #574, could you take a look to make sure I got the spirit of the thing right? Right now we have one foot in json world and one foot in csv world as far as the encoding. I'm trying to convince @vitorpamplona to use the CSV version.

@earonesty would appreciate your feedback as well.

paulmillr (approved):
We're getting closer to finalizing this.

Now we can as well ask:

- Is `JSON.stringify({ciphertext: base64.encode(ciphertext), nonce: base64.encode(nonce), v})` optimal?
- Should `base64.encode(JSON.stringify({ciphertext, nonce, v}))` be used instead?

staab:
Do we need to base64 encode at all if we're using json?

earonesty:
> We're getting closer to finalizing this.
> 
> Now we can as well ask:
> 
> Is `JSON.stringify({ciphertext: base64.encode(ciphertext), nonce: base64.encode(nonce), v})` optimal? Should `base64.encode(JSON.stringify({ciphertext, nonce, v}))` be used instead?
> 
> That's one less `base64.encode` call. Version is not immediately accessible, however, neither it's in prior version: there you need to json-parse first.

cant put bytes in json

manimejia:
> I hid some comments to keep things focused. The summary is there are three kinds of groups:
> 
> * Small groups ([NIP-24 Sealed Gift-Wrapped Messages for Private DMs and Small Group Chats #686](https://github.com/nostr-protocol/nips/pull/686))
> * Large groups where members trust each other not to reveal their membership (or where membership is not sensitive information)
> * Large groups where membership is protected by admins requiring Shamir's secret sharing where you can do M out of N decryption (?)
> 
> This PR targets use case number two.

Maybe I don’t get it … but it looks like the end user is gonna have to choose from (now) three not-obviously-different types of group, BEFORE a group is created (whose size or moderation needs might change any time in the future)? That alone is a recipe for disaster IMHO. To boot… am I right in thinking there is gonna be a secret key that group members need to remember or “carry” with them to get in … and one that will change periodically as people are booted from the group?

Sorry if I’m being dense… (or seemingly pessimistic, cause I really am not) but the UX scenario I’ve described above will not fly well IRL.

Please tell me I’m mistaken. 
Help a poor noob to understand.

staab:
The client developer will be the one to choose the implementation based on the user experience they're shooting for. Coracle will support both small and large groups, but with different interfaces. The third option in that list is theoretical, hopefully someone smart will make it a reality.

The shared key management is a relevant question, but it's not any more difficult than securing your regular key, and will only need to be done by people administering groups, so I don't think it's a big problem.

manimejia:
Thanks. How can key usage be more transparent to the end user, who may have 5 or 10 (or 110) groups that she belongs to? Would this require her to “manage” separate keys for each group … (with like a nos2x kind of experience) If so… maybe we’re forcing dependence on a bunker style key management tool across the board for users? 

(Cause private groups WILL be big and used in multiplicity by everyone.)

manimejia:
What I’d LIKE to see (that is IF group access is NOT provided by a user’s own npub, which IMHO would be best for all cases) … if users need to have separate keys for each group, their application should be “automagic” and behind the scenes. This is what (I imagine) a key bunker could do. Yes?

staab:
Group access being related to the user's own npub has the same metadata leakage problems NIP 04 has, so it's a non-started. Bunkers are one solution, another one is just a good backup solution for group keys — they could be encrypted by the user's own key and stored in nostr, backed up to a key manager, Google Drive, whatever. That's not a hard problem to solve, and can exist in the application layer.

vitorpamplona:
How about a simplification of this NIP? What's the absolute minimum a Client can do to create private groups? 

"Voting on a post" seems unnecessary at first glance.

staab:
I was thinking about that yesterday. I think you're right, existing NIP 72 facilities should be re-implemented inside groups. A group could also have an optional moderated community attached making it a private view of the community.

On the other hand, it's hard to remove because groups do need moderation outside of the NIP 72 case, so unless groups are tightly coupled to a community, it's hard to re-use NIP 72's moderation.

In my ideal world, moderation would be a separate concept that could be applied to any event. Something like this:

- I publish a moderator list with an `e` or `a` tag pointing to an event I created (could be a note, community, group, NIP 28 chat room, whatever)
- People can ignore it, or listen to moderator opinions about that event (hide/show replies, community posts, group notes, chat messages, whatever)

NIP 72 is much too rigid in my opinion, because it couples the moderator list to the community creation event. Should I write this idea up as a separate NIP?

vitorpamplona:
Well, I was kinda hoping for the opposite: a very simple moderated group chat. 

My use case is short-term projects: Billy, Chris, and I have to coordinate the party tonight. I create a private chat. Add them to it. We do our thing and that's it. The group will just fade away from my contact list.

staab:
Isn't that use case supported by NIP 24?

vitorpamplona:
> Isn't that use case supported by NIP 24?

Somewhat. Have you used Slack before? It's the difference between DMing 3 friends directly and making a little private group to discuss the "project". Direct DMing is loose, the second has a purpose/finality. 

I imagine the same tool can be used for class projects in school.

manimejia:
I'm a nobody, but I don't see a problem with any of this. 

Let NIP-72 moderated communities BE the defacto standard. ADD opt in functionality for group creators (and users) via additional NIPs. Clients and relays can support incrementally (and hopefully transparently). 

Want possibility for "default" mod list to become "opt in" list for group user to pick (one or min number of) mods? It's a NIP 

Want possibility for group users to "suggest" new mods by adding their name to a (linked by mod) NIP-28 list? It's a NIP.

Want possibility to make group metadata and content private by "gift wrap" encrypting for each group user, using their own npub? It's a NIP. (Cause gift wrapping events in general should be its own NIP)

Want possibility to scale your group to thousands of users? "Gift bag" (it's back) group messages for (assigned or elected) mods to then "gift wrap" for the rest of the group. Make sure your clients and relays implement this NIP.

staab:
> Somewhat. Have you used Slack before? It's the difference between DMing 3 friends directly and making a little private group to discuss the "project". Direct DMing is loose, the second has a purpose/finality. It's a different vibe.

Yep, I get that. Sounds like the proposed (and ditched) "subject" tag would have been enough? You could potentially hack that together by adding another recipient too.

But yeah, this is definitely intended for bigger things.

staab:
I am going to close this for now in favor of #875.

PR #711, opened 2023-08-10
Nip15 shipping cost per product

ktecho:
After a little discussion with the NostrMarket people, I've added per-product extra costs that would be added to the stall base cost.


motorina0:
LGTM

Just TLDR; of this PR:
 - some products might have extra shipping cost due to their size, weight, fragility, etc
 - the merchant can specify for each individual product this extra cost
 - on the **Data Structure** level a `shipping` field will be added to event kind `30018` (product)

<hr/>

https://github.com/ktecho/nips/blob/nip15_shipping_cost_per_product/15.md#event-30018-create-or-update-a-product

ibz:
Perfect! I'll implement it in [Plebeian Market](https://github.com/PlebeianTech/plebeian-market)!

arcbtc:
Great! LGTM

PR #713, opened 2023-08-10
NIP-25: Handle empty string as +

benthecarman:
Seems like a bunch of clients assume this so it should be documented


jb55:
The original spec was like this, not sure why we removed it...

alexgleason:
Why heart and not `+`?

benthecarman:
> Why heart and not `+`?

I am not sure, this is what other clients have implemented so I just want to document it

fiatjaf:
I thought this was the spec already. I mean handling like `+`.

benthecarman:
Changed to `+`

vitorpamplona:
hah... I didn't know that... living and learning.

jb55 (approved):

PR #715, opened 2023-08-11
Introduce NIP-44 encryption standard

staab:
A distilled version of #574 focused on the cryptography standard, and omitting DM-specific stuff. @paulmillr please feel free to incorporate this diff into the original PR and I can close this one, or close the original and we can continue from here, it's up to you.


44.md
```diff
@@ -0,0 +1,170 @@
...
+  const key = getSharedSecret(privkey, pubkey)
+  const nonce = randomBytes(24)
+  const plaintext = utf8Encoder.encode(text)
+  const ciphertext = xchacha20(key, nonce, plaintext)
+
+  return JSON.stringify({
+    ciphertext: base64.encode(ciphertext),
+    nonce: base64.encode(nonce),
+    v,
+  })
```
jb55, line 75: I'd prefer this so we can include it in at tag:

```suggestion
  return [base64.encode(ciphertext), base64.encode(nonce), v]
```

but if its too much bike shedding at this point than so be it.

44.md
```diff
@@ -0,0 +1,170 @@
...
+An encrypted payload MUST be encoded as a JSON object. Different versions may have different parameters. Every format has a `v` field specifying its version.
+
+Currently defined encryption algorithms:
+
+- `0x00` - Reserved
+- `0x01` - XChaCha with same key `sha256(ecdh)` per conversation
+
+# Version 0
+
+Version 0 is not defined, however implementations depending on this NIP MAY choose to support the payload described in NIP 04 in the same places a NIP 44 payload would otherwise be expected. This is intended to allow a smooth transition while clients and signing software adopt the new standard.
```
vitorpamplona, line 22: I don't like this ambiguity. We should say v0 is the NIP04-styled json. @v0l what you think?

44.md
```diff
@@ -0,0 +1,170 @@
+NIP-44
+======
+
+Encrypted Direct Message (Versioned)
+------------------------------------
+
+`optional` `author:paulmillr` `author:staab`
+
+The NIP introduces versioned encryption, allowing multiple algorithm choices to exist simultaneously.
```
vitorpamplona, line 9: The NIP introduces a versioned encryption data model, allowing multiple algorithm choices to exist simultaneously.

44.md
```diff
@@ -0,0 +1,170 @@
...
+
+The algorithm described in NIP4 is potentially vulnerable to [padding oracle attacks](https://en.wikipedia.org/wiki/Padding_oracle_attack) and uses keys which are not indistinguishable from random.
```
vitorpamplona, line 11: Delete

44.md
```diff
@@ -0,0 +1,170 @@
+NIP-44
+======
+
+Encrypted Direct Message (Versioned)
```
vitorpamplona, line 4: Encryption Payload Definition

44.md
paulmillr: nip4 is broken cryptography that will expose your private keys. It should never be used.

jb55:
>+Version 0 is not defined, however implementations depending on this NIP MAY choose to support the payload described in NIP 04 in the same places a NIP 44 payload would otherwise be expected. This is intended to allow a smooth transition while clients and signing software adopt the new standard.
>
>nip4 is broken cryptography that will expose your private keys. It should never be used.

do you have a demo of this?

jb55:
because if this is true then there should be an advisory and every client should remove it?

44.md
staab: I think this is helpful context for now.

44.md
staab: I'll resolve this conversation since it's a duplicate of https://github.com/nostr-protocol/nips/pull/716#discussion_r1291530129

staab:
> nip4 is broken cryptography that will expose your private keys. It should never be used.

This is an important point. Signing applications don't support the new version, so compatibility with NIP 04 could ease adoption (or may it more complicated, I'm not sure). @paulmillr can you confirm how bad it would be to enshrine NIP 04 as a deprecated version 0?

staab:
Looping in a comment from https://github.com/nostr-protocol/nips/pull/706#pullrequestreview-1573440314 we have a few different proposals for the payload:

- CSV — compact, simple parsing, might be easier on mobile for @jb55. Downside is positional and @vitorpamplona thinks it's complicated
- `JSON.stringify({ciphertext: base64.encode(ciphertext), nonce: base64.encode(nonce), v})` — currently implemented in multiple places.
- `base64.encode(JSON.stringify({ciphertext, nonce, v}))` — one fewer base64 call. I'm not sure base64 is even necessary if we're encoding to json either
- Tags — would make @jb55's life easier, but probably make interfacing with signers harder.

I'm going to leave what we currently have unless anyone has strong opinions so we can put this to rest.

paulmillr:
base64 is useful, because the result data is binary, and representing it in hex instead of base64 would increase the size by 1.5x in my benchmarks. Saving a traffic is cool.

vitorpamplona:
Just for completeness, the most performant option is a custom TLV, which we also already implement in Nostr. CSV is fast, but not as fast/simple as TLV because of sequential bytearray copy instructions. We can even do the inner gossip event as a TLV as well. 

I would still go for JSON.

staab:
Thank you, that's exactly the kind of answer I was looking for. I'm so glad you're here to help us devs get things right. I will remove the recommendation for NIP 04 and will not include backwards compatibility in my implementation.

Tangentially, does this mean that anyone who has sent a substantial number of NIP 04 messages is more likely to have their key brute-forced?

staab:
Some follow-ups to this PR:

- Deprecate use of NIP 04 in NIP 51, 15, and 47

paulmillr:
>anyone who has sent a substantial number of NIP 04 messages is more likely to have their key brute-forced?

Yes.

That's from my understanding, but I think any auditor of the protocol would come to  similar conclusions.

staab:
Well that sucks haha

BlowaterNostr:
Why is custom TLV fast?

44.md
```diff
@@ -0,0 +1,169 @@
...
+======
+
+Encrypted Payloads (Versioned)
+------------------------------
+
+`optional` `author:paulmillr` `author:staab`
+
+The NIP introduces a versioned encryption data model, allowing multiple algorithm choices to exist simultaneously.
+
+The algorithm described in NIP-04 is potentially vulnerable to [padding oracle attacks](https://en.wikipedia.org/wiki/Padding_oracle_attack) and uses keys which are not indistinguishable from random. For more information, see [here](https://github.com/nostr-protocol/nips/pull/715#issuecomment-1675301250-).
```
alexgleason, line 11: This link has an accidental `-` at the end.

```suggestion
The algorithm described in NIP-04 is potentially vulnerable to [padding oracle attacks](https://en.wikipedia.org/wiki/Padding_oracle_attack) and uses keys which are not indistinguishable from random. For more information, see [here](https://github.com/nostr-protocol/nips/pull/715#issuecomment-1675301250).
```

44.md
```diff
@@ -0,0 +1,169 @@
+NIP-44
+======
+
+Encrypted Payloads (Versioned)
+------------------------------
+
+`optional` `author:paulmillr` `author:staab`
+
+The NIP introduces a versioned encryption data model, allowing multiple algorithm choices to exist simultaneously.
```
alexgleason, line 9: I would suggest introducing the topic a little better, and then providing some clarity about where and how this should be used.

```suggestion
Nostr can use private key to encrypt data for a particular user (pubkey). Then, that user can decrypt the data with their own private key. This NIP introduces a versioned JSON format for symmetrically encrypted data, allowing multiple algorithm choices to exist simultaneously.

The resulting JSON data could be used in the content of an event, or in a tag. This NIP defines only the data model and supported algorithms and not the events themselves.
```

44.md
```diff
@@ -8,7 +8,7 @@ Encrypted Payloads (Versioned)
 
 The NIP introduces a versioned encryption data model, allowing multiple algorithm choices to exist simultaneously.
 
-The algorithm described in NIP-04 is potentially vulnerable to [padding oracle attacks](https://en.wikipedia.org/wiki/Padding_oracle_attack) and uses keys which are not indistinguishable from random. For more information, see [here](https://github.com/nostr-protocol/nips/pull/715#issuecomment-1675301250-).
+The algorithm described in NIP-04 is potentially vulnerable to [padding oracle attacks](https://en.wikipedia.org/wiki/Padding_oracle_attack) and uses keys which are not indistinguishable from random. For more information, see [here](https://github.com/nostr-protocol/nips/pull/715#issuecomment-1675301250).
```
paulmillr, line 11: Don't think it's good idea to link to the comment. Can we simplify it?

> The algorithm described in NIP-04 is not secure. It used bad cryptographic building blocks and must not be used.

staab: Your comment was so helpful, but yeah we have it in this PR for reference. I don't think most people realize that NIP 04 compromises your private key.

44.md
staab: I don't think we need to go quite back to basics, but I've tweaked the wording based on your suggestions. Also, this is generally for asymmetric rather than symmetric encryption.

paulmillr:
@fiatjaf what do you think? It's just encryption here, that can be used in any part of nostr. No DMs.

fiatjaf:
Yes, I like it.
I don't like the JSON inside the content.

fiatjaf:
Doesn't the nonce have a fixed size?

You can concat everything (as bytes) then base64-encode it: `base64(version (1 byte) + nonce (24 bytes) + payload (whatever bytes))`, that's faster than TLV and easier to do.

If the version ever changes, which I hope it doesn't, then you can still change this format if necessary, the parser will just have to read the first byte then interpret the rest accordingly.

paulmillr:
So what? Nested JSON is still terrible. Nested JSON is simpler for frontend devs, but much more involved for low-level languages. The performance and having nested JSON errors is also pretty bad.

paulmillr:
The NIP should provide 5 valid test vectors and 5 invalid ones with errors. It would be mandatory for all implementations to pass/fail them. I will think of something.

staab:
Thanks @fiatjaf and @paulmillr for weighing in. I think the consensus at this point is to use a custom TLV. I will update the NIP etc. @paulmillr could you put together those test vectors?

fiatjaf:
I agree we should be mindful of frontend devs not knowing how to deal with bytes, but in this case if they're already dealing with confusing encryption stuff that operates in bytes so they should be ready to do this simple parsing code.

In practice they will use a library.

staab:
Ok, just updated the PR to use a custom TLV instead of JSON, you can see the code/tests here: https://github.com/nbd-wtf/nostr-tools/pull/278

Let me know if it looks right, I'm one of the aforementioned web developers and haven't used TLV's before.

vitorpamplona:
> just updated the PR to use a custom TLV instead of JSON

Is it worth it? 

What are people's thoughts on zlib compressing the JSON before making the base64? It's slower, but it might save significant space if GiftWraps get used for real.

fiatjaf:
What about https://lz4.org/ instead?

fiatjaf:
I also think we should remove this Version 0 notice. It is confusing and misleading and I don't think people can or should do this "smooth transition" thing.

fiatjaf:
I don't know how to use GitHub so I did this: https://github.com/staab/nips/pull/2 @staab

staab:
Thanks, I'll test and update my PR with those changes

vitorpamplona:
Ok, I spent the day getting some performance numbers for us. 

This is the time to encrypt and decrypt a single message between two people (2 Gift Wraps): 
1. Create Msg Event
2. Create Seal (Msg.toJson, Encrypt / Encode)
3. Create 1st GiftWrap (Seal.toJson, Encrypt / Encode)
4. Create 2nd GiftWrap (Seal.toJson, Encrypt / Encode)
5. Decode 1st GiftWrap (Decode, Decrypt, Msg.fromJson)
6. Decode 1st Seal (Decode, Decrypt, Msg.fromJson)
7. Decode 2nd GiftWrap (Decode, Decrypt, Msg.fromJson)
8. Decode 2nd Seal (Decode, Decrypt, Msg.fromJson)

The table below shows the time in nanoseconds for a full passthrough, averaged over 50,000 passthroughs, for @fiatjaf's encoding, a base CSV, and 3 JSON parsers: GSON, Jackson, Kotlin serialization. All running on a regular Pixel 6, for 3 message sizes: 

- Long: My queen, you are nothing short of royalty to me. You possess more beauty in the nail of your pinkie toe than everything else in this world combined. I am astounded by your grace, generosity, and graciousness. I am so lucky to know you.
- Regular: Hi, honey, can you drop by the market and get some bread?
- Short: Hola, que tal?

Time to run each cycle. 

  | FiatJaf's | GSON | Jackson | Kotlin | CSV
-- | -- | -- | -- | -- | -- 
Long | 662,906 ns | 730,534 ns | 709,389 ns | 687,229 ns | 679,352 ns
Regular | 651,122 ns | 696,014 ns | 672,737 ns | 709,497 ns | 652,474 ns
Short | 624,790 ns | 678,260 ns | 645,922 ns | 658,035 ns | 632,484 ns

Positive % means worse than FiatJaf's

  | FiatJaf's | GSON | Jackson | Kotlin | CSV
-- | -- | -- | -- | -- | --
Long | 0.00% | 9.26% | 6.55% | 3.54% | 2.42%
Regular | 0.00% | 6.45% | 3.21% | 8.23% | 0.21%
Short | 0.00% | 7.88% | 3.27% | 5.05% | 1.22%

In Bytes, each serialized GiftWrap event takes: 

Bytes | FiatJaf's | JSON | CSV | JSON Worse By | CSV Worse By
-- | -- | -- | -- | -- | --
Long | 3738 | 3938 | 3752 | 5.08% | 0.37%
Regular | 3098 | 3298 | 3112 | 6.06% | 0.45%
Short | 2946 | 3154 | 2960 | 6.59% | 0.47%

Interesting learnings: 
1. JSON parsing and native object mapping of small objects (3 fields like this one) is actually quite fast.  
2. Size overhead is large for small messages. 
3. The encoding/decoding of the encryption parameters is just a small part of the entire processing.

staab:
Ironically, I prefer the TLV approach because it's cleaner and more elegant than the JSON approach. The performance gain is just a side benefit.

vitorpamplona:
Compression-before-encrypting tradeoffs: 

If we use FiatJaf's or Json's, the processing time increases by ~30%

Time | FiatJaf's As Is | w/ Compression | Time Worse by
-- | -- | -- | --
Long | 662,906 ns | 878,471 ns | 32.52%
Regular | 651,122 ns | 845,188 ns | 29.80%
Short | 624,790 ns | 816,035 ns | 30.61%

  | Kotlin JSON | w/ Compression | Time Worse by
-- | -- | -- | --
Long | 687,229 ns | 912,143 ns | 32.73%
Regular | 709,497 ns | 886,434 ns | 24.94%
Short | 658,035 ns | 866,524 ns | 31.68%

But the size of the payload reduces between 26 - 37% 🤯

Bytes | FiatJaf's As Is | FiatJaf's w/ Compression | Space Gains
-- | -- | -- | --
Long | 3738 | 2722 | 37.33%
Regular | 3098 | 2406 | 28.76%
Short | 2946 | 2318 | 27.09%

Bytes | JSON's As Is | JSON w/ Compression | Space Gains
-- | -- | -- | --
Long | 3938 | 2894 | 36.07%
Regular | 3298 | 2570 | 28.33%
Short | 3154 | 2494 | 26.46%

fiatjaf:
lz4 is so much faster though. Or at least that's what the charts tell me.

vitorpamplona:
> LZ4 is so much faster though. Or at least that's what the charts tell me.

Yes, I am just having a hard time finding a usable android implementation

paulmillr:
@vitorpamplona you are missing two important points:

1. base64 is much slower than JSON, order of magnitudes
2. schnorr verification is order of magnitudes slower than base64

I'm talking about JS clients.

Any arriving message will need to do both 1 and 2.

vitorpamplona:
Those are included in the time measured.

paulmillr:
I don't see where schnorr verification is included. You are not verifying signatures.

vitorpamplona:
My bad. It's part of the decode (I am simulating the JSON coming in from the Relay as STRING). So, the GiftWrap "decode" is the process that parses it into an Event object first and that includes the verification of the hash and the signature.

I just measured them individually. 

- Schnoor takes ~25,000 ns each event (* 4: 2 GiftWraps and 2 inner seals). 
- The ID Hash Check takes 13,000 ns each event (* 4: 2 GiftWraps and 2 inner seals). 
- Base64 goes for 5,000 ns each event (* 4: 2 GiftWraps and 2 inner seals).

I am still using an unoptimized code that triggers a bunch of garbage collection. These 3 actions could be way better.

paulmillr:
>Schnorr takes ~25,000 ns each event (* 4: 2 GiftWraps and 2 inner seals).

Please take a look at my emphasis:

>I'm talking about **JS clients**.

It's not 25k ns - it's 1.1ms on M2, and much more on slow android phones.

vitorpamplona:
> It's 1.1ms per verification on M2

On M2 but in JavaScript or native? 

Maybe we can run a similar benchmark in JavaScript. I ddn't realized it was that slow.

paulmillr:
That's JS. Wasm is not an option in many contexts. We need to take into account all main platforms and cases.

Native byte shuffling is easier than JSON, in JS it's slightly harder. Native zlib is slightly simpler than in JS, where it requires a custom, who-knows-how-big library.

JS is shit, but we're stuck with it.

vitorpamplona:
Flipside is that parsing JSON in JavaScript tends to be 30% faster than native Kotlin. So.. there are some advantages :)

paulmillr:
fiatjaf's stuff aka base64(concatbytes()), pure encoding / decoding, without key verification or signing:

nip44 short x 90,309 ops/sec @ 11μs/op
nip44 regular x 52,353 ops/sec @ 19μs/op
nip44 long x 18,564 ops/sec @ 53μs/op

same, but with signature:

nip44 short x 636 ops/sec @ 1ms/op
nip44 regular x 637 ops/sec @ 1ms/op
nip44 long x 631 ops/sec @ 1ms/op

JSON or not JSON doesn't matter in perf terms in js, even with long messages the impact is 1.5%

staab:
Yeah, nack on the compression stuff. I've just updated the PR to use @fiatjaf's TLV code with @paulmillr's fix.

vitorpamplona:
For the sake of the discussion, @fiatjaf's byte array concat is not a TLV. :)

staab:
Ah, fair enough, I stand correct (and clearly don't even know what TLV stands for)

44.md
```diff
@@ -44,52 +44,16 @@ This encryption scheme replaces the one described in NIP-04, which is not secure
...
+import {sha256} from '@noble/hashes/sha256'
+import {xchacha20} from '@noble/ciphers/chacha'
 
-  return concatBytes(...entries)
-}
+import {utf8Decoder, utf8Encoder} from './utils.ts'
 
 export const getSharedSecret = (privkey: string, pubkey: string): Uint8Array =>
-  sha256(secp256k1.getSharedSecret(privkey, "02" + pubkey).subarray(1, 33))
+  sha256(secp256k1.getSharedSecret(privkey, '02' + pubkey).subarray(1, 33))
```
paulmillr, line 56: My only concern with regards to the algorithms in the NIP is the fact we're always prepending `02` parity byte. Not sure about potential consequences of that. E.g. bias towards one, or another kind of shared keys.

paulmillr:
My only concern with regards to the algorithms in the NIP is the fact we're always prepending `02` prefix byte. Not sure about potential consequences of that. E.g. bias towards one, or another kind of shared keys.

Associativity tests mention that it works fine:

```js
let secp = require('@noble/curves/secp256k1');
let utils = require('@noble/curves/abstract/utils')
let rand = secp.secp256k1.utils.randomPrivateKey;
let prefix = new Uint8Array([0x02])
let pub = secp.schnorr.getPublicKey;
let ecdh = (a, b) => utils.bytesToHex(secp.secp256k1.getSharedSecret(a, utils.concatBytes(prefix, b)).subarray(1));

// associativity test
for (let i = 0; i < 10000; i++) {
  if (i % 500 === 0) console.log(i);
  let privA = rand();
  let privB = rand();
  if (ecdh(privA, pub(privB)) !== ecdh(privB, pub(privA))) {
    console.log({privA, privB});
    throw new Error('ecdh is not equal');
  }
}
```

44.md
```diff
@@ -0,0 +1,166 @@
...
+import {secp256k1} from '@noble/curves/secp256k1'
+import {sha256} from '@noble/hashes/sha256'
+import {xchacha20} from '@noble/ciphers/chacha'
+
+import {utf8Decoder, utf8Encoder} from './utils.ts'
+
+export const getSharedSecret = (privkey: string, pubkey: string): Uint8Array =>
+  sha256(secp256k1.getSharedSecret(privkey, '02' + pubkey).subarray(1, 33))
+
+export function encrypt(privkey: string, pubkey: string, text: string, v = 1) {
```
paulmillr, line 58: Should it be `privkey: string, pubkey: string`, or rather `key: Uint8Array`?

Caching an encryption key somewhere is definitely a goal. Using `privkey, pubkey` signature makes it impossible.

paulmillr: No, user can't re-derive encryption key for 50 messages. That would consume all the performance. ECDH is slow, like, 500 per sec on M2. They need to derive key once and use the same key for encryption / decryption of all conversation messages, and there can be 1000s.

paulmillr:
```js
const msgs = await fetchMessages(100);

// original
let key = getSharedSecret(privA, pubB);
msgs.map(msg => decrypt(key, msg)); // pure decryption routine: 100K/sec

// vs your proposal
msgs.map(msg => decrypt(privA, pubB, msg)); // decryption + getSharedSecret: 200x slower
```

This is just one convo. There can be like 20.

staab: This sounds like an optimization, couldn't you just lru_cache the `getSharedSecret` function?

paulmillr: One frame is best-case 16.6ms, worst-case 8.3ms. Decrypting messages in main thread (when you can't use separate one) would just make UI stuck. Even with a separate thread, decrypting 500 conversation messages could take 2-5 seconds on a mobile phone with your proposal. Forcing users to wait is terrible UX.

staab: Got it, yes the base version should just take a key, I'll update that. For stuff like browser extensions, there's no need to expose the key-only version (if the application has a key they can import nostr-tools or whatever), so that interface will look a little different.

44.md
```diff
@@ -0,0 +1,165 @@
...
+======
+
+Encrypted Payloads (Versioned)
+------------------------------
+
+`optional` `author:paulmillr` `author:staab`
+
+The NIP introduces a new data format for keypair-based encryption. This NIP is versioned to allow multiple algorithm choices to exist simultaneously.
+
+An encrypted payload MUST be encoded as a JSON object. Different versions may have different parameters. Every format has a `v` field specifying its version.
```
fiatjaf, line 11:
```suggestion
The encrypted payload is encoded as base64. The first byte of the payload represents the version.
```

44.md
```diff
@@ -0,0 +1,165 @@
...
+- `0x00` - Reserved
+- `0x01` - XChaCha with same key `sha256(ecdh)` per conversation
+
+# Version 1
+
+Params:
+
+1. `nonce`: base64-encoded xchacha nonce
+2. `ciphertext`: base64-encoded xchacha ciphertext, created from (key, nonce) against `plaintext`.
+
```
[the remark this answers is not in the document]
fiatjaf: Something like this?

staab:
Is everyone ok with merging this?

alexgleason (approved):

paulmillr:
Let's add vectors, i've been working on them, and will finish soon. Also let's ensure the code is the same we have in nostr-tools (I will update PR).

44.md
```diff
@@ -0,0 +1,168 @@
...
+- `0x00` - Reserved
+- `0x01` - XChaCha with same key `sha256(ecdh)` per conversation
+
+# Version 1
+
+Encoding:
+
+- `1` - 1 byte, version
+- `nonce` - 24 bytes
```
paulmillr, line 23: it's important to note somewhere, perhaps here, that nonce should be fetched from CSPRNG (cryptographically secure random source). If we leave it up to users, they will use garbage and leak their messages through nonce reuse.

staab:
Code samples should be in sync, see: https://github.com/nbd-wtf/nostr-tools/pull/278

vitorpamplona:
For some reason I can't seem to be able to offer changes directly. Here it goes: 

```kotlin
## Kotlin

```kotlin
// implementation 'fr.acinq.secp256k1:secp256k1-kmp-jni-android:0.10.1'
// implementation "com.goterl:lazysodium-android:5.1.0@aar"
// implementation "net.java.dev.jna:jna:5.12.1@aar"

fun getSharedSecretNIP44(privKey: ByteArray, pubKey: ByteArray): ByteArray =
  MessageDigest.getInstance("SHA-256").digest(
    Secp256k1.get().pubKeyTweakMul(
      Hex.decode("02") + pubKey,
      privKey
    ).copyOfRange(1, 33)
  )

fun encryptNIP44(msg: String, privKey: ByteArray, pubKey: ByteArray): String {
  val nonce = ByteArray(24).apply {
    SecureRandom.getInstanceStrong().nextBytes(this)
  }

  val cipher = streamXChaCha20Xor(
    message = msg.toByteArray(),
    nonce = nonce,
    key = getSharedSecretNIP44(privKey, pubKey)
  )

  return Base64.getEncoder().encodeToString(
    byteArrayOf(Nip24Version.XChaCha20.code) + nonce + cipher
  )
}

fun decryptNIP44(base64: String, privKey: ByteArray, pubKey: ByteArray): String? {
  val byteArray = Base64.getDecoder().decode(base64)

  require(byteArray[0].toInt() == Nip24Version.XChaCha20.code) { "NIP44: unknown encryption version" }

  return streamXChaCha20Xor(
    message = byteArray.copyOfRange(25, byteArray.size),
    nonce = byteArray.copyOfRange(1, 25),
    key = getSharedSecretNIP44(privKey, pubKey)
  )?.decodeToString()
}

// This method is not exposed in AndroidSodium yet, but it will be in the next version.
fun streamXChaCha20Xor(message: ByteArray, nonce: ByteArray, key: ByteArray): ByteArray? {
  return with (SodiumAndroid()) {
    val resultCipher = ByteArray(message.size)

    val isSuccessful = crypto_stream_chacha20_xor_ic(
      resultCipher,
      message,
      message.size.toLong(),
      nonce.drop(16).toByteArray(), // chacha nonce is just the last 8 bytes.
      0,
      ByteArray(32).apply {
      crypto_core_hchacha20(this, nonce, key, null)
      }
    ) == 0

    if (isSuccessful) resultCipher else null
  }
}

enum class Nip24Version(val code: Int) {
  Reserved(0),
  XChaCha20(1)
}
```

44.md
```diff
@@ -0,0 +1,167 @@
...
+// This method is not exposed in AndroidSodium yet, but it will be in the next version.
+fun streamXChaCha20Xor(message: ByteArray, nonce: ByteArray, key: ByteArray): ByteArray? {
+  return with (SodiumAndroid()) {
+    val resultCipher = ByteArray(message.size)
+
+    val isSuccessful = crypto_stream_chacha20_xor_ic(
+      resultCipher,
+      message,
+      message.size.toLong(),
+      nonce.drop(16).toByteArray(), // chacha nonce is just the last 8 bytes.
```
paulmillr, line 152: How does this work?

vitorpamplona: Not on libsodium. That code is from them.

vitorpamplona: https://github.com/jedisct1/libsodium/blob/4dc02ce8414e70b8960bea77abf4fe46efb99e93/src/libsodium/crypto_stream/xchacha20/stream_xchacha20.c#L45

fiatjaf (approved):

paulmillr:
@vitorpamplona caching is ok. But the strategy must not be forced onto all users and instead must be defined on per-user basis. It can be fine in one situation and not fine in other situation.

I don't like having `decrypt(privA, pubB, ciphertext)` - instead, it should always be `decrypt(key, ciphertext)`. This way you can cache the key if you'd like, or not do it.

Having `decrypt(privA, pubB, ciphertext)` with lru_cache is fragile: every implementation in every language would need to define the lru_cache by itself, decide proper place to store it.

44.md
paulmillr: yeah, it's ok, same as in js then.

paulmillr:
I will have test vectors ready in a few hours.

vitorpamplona:
For the benchmark curious, here is the time broken down for each state of **receiving** a single GiftWrap. 

The previous benchmark included 2 of these + signing (+/- 350,000 ns)

Time Each (ns) |  
-- | --
5,104 | parseEventFromRelayString
5,080 | checkWrapIDHash
23,765 | checkWrapSignature
50 | assembleSharedKey(me.priv, me.pub) // cached
2,404 | decodeBase64WrappedEvent
4,602 | decryptWrappedEvent
3,005 | parseWrappedEvent
5,080 | checkSealIDHash
23,765 | checkSealSignature
22,983 | assembleSharedKey(me.priv, author.pub)
1,027 | decodeBase64SealedEvent
2,597 | decryptSealedEvent
3,503 | parseSealedEvent
102,965 | **Total** (ns)

paulmillr:
Test vectors are here: https://github.com/nbd-wtf/nostr-tools/pull/221/files

They should be placed into the NIP44.

1. What should we do to base64 padding? How does nostr handle it already? Is nostr malleable? Does padding matter? I think this may be a problem in some contexts.
2. The NIP should specify that the encryption does not provide integrity. It means the encryption should always be used inside of a signed message. Otherwise, the ciphertext is decryptable under any key!
3. The NIP should specify that both private and public key inputs must be checked for validity before key is emitted.

The pull request at nostr-tools has code that could be copied into here.

mikedilger:
Add NIP-44 to the README.md in this PR.

44.md
```diff
@@ -0,0 +1,167 @@
...
+
+Example:
+
+- Alice's private key: `5c0c523f52a5b6fad39ed2403092df8cebc36318b39383bca6c00808626fab3a`
+- Bob's private key: `4b22aa260e4acb7021e32f38a6cdf4b673c6a277755bfce287e370c924dc936d`
+
+Encrypting the message `hello` from Alice to Bob results in the following base-64 encoded payload:
+
+```
+Ac4wiXzHQ20IrriCl8ZNzPWEGre97jKP6THD4ACR
```
mikedilger, line 36: This cannot be verified unless you tell us what you used for a random 24-byte nonce.

Also, I think there ought to be a test vector going in the other direction too.

44.md
```diff
@@ -0,0 +1,167 @@
...
+======
+
+Encrypted Payloads (Versioned)
+------------------------------
+
+`optional` `author:paulmillr` `author:staab`
+
+The NIP introduces a new data format for keypair-based encryption. This NIP is versioned to allow multiple algorithm choices to exist simultaneously.
+
+The encrypted payload is encoded as base64. The first byte of the payload represents the version. The rest depends on the version.
```
mikedilger, line 11: base64 without padding?

44.md
```diff
@@ -0,0 +1,167 @@
...
+
+Currently defined encryption algorithms:
+
+- `0x00` - Reserved
+- `0x01` - XChaCha with same key `sha256(ecdh)` per conversation
```
mikedilger, line 16: XChaCha20 (add the '20' to the end)

Also what is this "per conversation"?

44.md
```diff
@@ -0,0 +1,167 @@
...
+
+# Version 1
+
+Encoding:
+
+- `1` - 1 byte, version
+- `nonce` - 24 bytes fetched from a crytographically secure random source
+- `ciphertext` - variable number of bytes until the end
+
+`ciphertext` is created using `xchacha(key, nonce, plaintext)`
```
mikedilger, line 26: xchacha20(...)   (again the 20 should be there)

mikedilger: Also at this point you should explain what 'key' is, rather than putting that in a note later, because this is where the thing is being defined.

mikedilger (changes requested):
Some comments I came up with while writing code for this.

44.md
fiatjaf: That's true, but I think it's fine just as a soft-example of what it looks like.

44.md
paulmillr: It's common to call it just ChaCha, instead of Salsa20 - which is usually suffixed with 20. So it's cool I think.

mikedilger: But there are 8 and 12 round versions. Crypto people may know that 20 is implied, but nostr devs may stumble upon the other ones and not know which one to use. See this lib for example: https://docs.rs/chacha20/latest/chacha20/

44.md
mikedilger: I recant my statement. The nonce is in the output.

mikedilger:
These stream ciphers don't have any padding.  If I knew somebody was going to reply with "yes" or "no" and I saw the reply was 3 characters long, I'd know they said "yes".

Can we add some kind of scheme to pad the input?

mikedilger:
I'm not getting the example to work. I get:

```
shared key = dbeece7b5ec47058b2c0926efc760db6d20af3b1b42e2d332cd65d11a719b359
nonce = ce30897cc7436d08aeb88297c64dccf5841ab7bdee328fe9
encrypted output = Ac4wiXzHQ20IrriCl8ZNzPWEGre97jKP6dJnghV4
```

If I decrypt the stated `Ac4wiXzHQ20IrriCl8ZNzPWEGre97jKP6THD4ACR` I get `[139, 193, 14, 121, 134]` not `[104, 101, 108, 108, 111]`

staab:
Test vectors are added, and most of Mike's questions are addressed. The example seems to have been wrong, I've updated this PR ~20 times so it was probably just an error on my part.

> What should we do to base64 padding? How does nostr handle it already? Is nostr malleable? Does padding matter? I think this may be a problem in some contexts.
> The NIP should specify that the encryption does not provide integrity. It means the encryption should always be used inside of a signed message. Otherwise, the ciphertext is decryptable under any key!

Are you saying that XChaCha20 is malleable, and if the event isn't signed there's a possibility the payload has been tampered with? And that, if nostr's schnorr signature implementation is also malleable, the payload can still be tampered with, leading to decryption?

> The NIP should specify that both private and public key inputs must be checked for validity before key is emitted.

Is that just checking key length, non-zero keys, etc? Should this be added to the reference implementation?

paulmillr:
> Are you saying that XChaCha20 is malleable, and if the event isn't signed there's a possibility the payload has been tampered with

Of course. Obviously. Chacha is just a stream cipher: it takes (key, nonce) and produces ~infinite stream that's XOR-ed with plaintext. If you have different key, it would produce different stream, but XOR-ing would still work and the encrypted ciphertext would decrypt to garbage.

To get integrity, you'll need xchacha20-poly1305 instead. However, poly1305 isn't cheap and isn't without its own issues. Schnorr Signatures provide stronger integrity than poly1305, so it's redundant. I see no reason to add poly1305: for example, some schemes aren't compatible with polynomial authenticators and require something like HMAC-SHA2 instead.

@mikedilger 

>These stream ciphers don't have any padding. If I knew somebody was going to reply with "yes" or "no" and I saw the reply was 3 characters long, I'd know they said "yes". Can we add some kind of scheme to pad the input?

That sounds like a good idea. What kind of padding scheme do you have in mind? How should it be handled?

mikedilger:
My attempt at padding:

Option 1: easy but doesn't support files ending in null bytes:
This encryption scheme exposes the length of the message. If that is important to obfuscate, the input MAY be padded with any number of NUL bytes. On decryption, trailing NUL bytes MUST be stripped.

Option 2: binary-safe for all messages, defined herein so people can't forget to do it:
Before encryption, the message should be assembled as LENGTH + MESSAGE + PADDING where LENGTH is a 4-byte unsigned integer, and padding is any number of NUL bytes.

Option 3: punt to NIPs that use this:
This encryption scheme exposes the length of the message. Usage of this NIP must bear this in mind and offer some means of padding the input.

I prefer option 2.

staab:
> Of course. Obviously. Chacha is just a stream cipher

Wasn't obvious to me, I was today years old when I learned the term "malleable".

> I just think it should be mentioned.

Added that to the NIP. Making edits is no problem, if you see problems in the PR feel free to suggest changes in github and I'll merge them.

mikedilger:
libsecp256k1_ecdh() already does a SHA256.  The Javascript example then does *another* SHA256.

And this affects all those test vectors.

mikedilger:
I share my testing code here:  https://github.com/mikedilger/nip44tests

paulmillr:
@mikedilger

1. noble-curves return 33-byte output for ecdh, the same output as in ecdsa-getPublicKey: parity byte AND x coordinate. The output is not hashed for flexibility.
2. libsecp256k1_ecdh does this: https://github.com/bitcoin-core/secp256k1/blob/6b9507adf684f088f665354608685255d0bd162f/src/modules/ecdh/main_impl.h#L13-L24. It seems like it matches noble-curves behavior, with sha256() added on top?
3. In this PR I initially proposed stripping parity byte because I thought libsecp256k1 strips is, but it doesn't seem to do so.

There is no single ECDH standard, unfortunately. For example, [this rust library](https://docs.rs/secp256k1/latest/secp256k1/ecdh/fn.shared_secret_point.html) returns 64-byte unhashed point. What's needed: investigating more libraries to see how they behave and if they can be simply adjusted to one way or another.

With regards to padding, I support your decision number 2, it's simple and no-bs.

@staab 

>Wasn't obvious to me, I was today years old when I learned the term "malleable".

We just need to understand all the use-cases of the NIP and decide if the result message would always be signed with schnorr. If they won't, perhaps it's worth switching to poly1305.

mikedilger:
Ok @paulmillr I have gotten my code to match and I get it now.   The rust lib `secp256k1` has a `SharedSecret` which effectively does a SHA256(parity + X) which gives the `b1c99...` output.  The TypeScript/noble example in the NIP effectively does a SHA256(X) without the parity which gives the `0135d...` output.  I don't care, I was just trying to figure out why my code didn't match the test vectors.

staab:
> We just need to understand all the use-cases of the NIP and decide if the result message would always be signed with schnorr. If they won't, perhaps it's worth switching to poly1305.

I think that's a safe assumption for now, we can introduce a new version if we need it

44.md
fiatjaf: By padding you mean the `===` stuff at the end?

staab: I was wondering the same thing. From what I read, base64 padding is to ensure the right number of bytes, extra padding seems unusual.

fiatjaf:
Implemented this on https://github.com/nbd-wtf/go-nostr/blob/cd86ee25147ac3ec26326bbac52f90ebe2fb3710/nip44/nip44.go with the test vectors.

The test vectors don't check the type of the base64 encoding though.

44.md
mikedilger: I meant the == stuff. Some base64 impls use it, some don't. Looks like we are using it. Good, that's more normal.

This is very different from padding the plaintext which I talk about elsewhere.

paulmillr:
>Stupid question, but isn't the event id/hash also serving the same function the poly1305 thing would?

hash can be re-calculated if the underlying data is changed.

poly1305 is like hmac. they both receive `key, data` instead of just `data`, like hash. 

so, poly1305(key, changedData) would fail. hash(changedData) would simply produce new, correct hash, which an attacker could place into the hash / id field.

Again, signatures resolve all the issues here. It's not possible to forge a new message that's valid for the same signature. Unless, of course, you're using shitty malleable signature scheme. Bonus point: tweetnacl is malleable, so was secp256k1 ecdsa some years ago.

@fiatjaf I want to pursue padding addition, so this will need another day or so. Above, the good point was made with regards to it: `yes` would always produce 1+24+3 bytes, which can totally be used by an attacker to deduce the result. I will update test vectors ASAP. The padding is almost free, but is important. Secure messengers use padding.

44.md
```diff
@@ -0,0 +1,255 @@
...
+
+- `0x00` - Reserved
+- `0x01` - XChaCha20 with same key `sha256(ecdh)` per conversation
+
+# Version 1
+
+Encoding:
+
+- `1` - 1 byte, version
+- `nonce` - 24 bytes fetched from a crytographically secure random source
```
ekzyis, line 23:
```suggestion
- `nonce` - 24 bytes fetched from a cryptographically secure random source
```

_insert "it's not much but it's honest work" meme_

07.md
```diff
@@ -18,8 +18,10 @@ async window.nostr.signEvent(event: Event): Event // takes an event object, adds
 Aside from these two basic above, the following functions can also be implemented optionally:
 ```
-async window.nostr.nip04.encrypt(pubkey, plaintext): string // returns ciphertext and iv as specified in nip-04
-async window.nostr.nip04.decrypt(pubkey, ciphertext): string // takes ciphertext and iv as specified in nip-04
+async window.nostr.nip04.encrypt(pubkey, plaintext): string // returns ciphertext and iv as specified in nip-04 (deprecated)
+async window.nostr.nip04.decrypt(pubkey, ciphertext): string // takes ciphertext and iv as specified in nip-04 (deprecated)
+async window.nostr.nip44.encrypt(pubkey, plaintext, v): string // returns encrypted payload as specified in nip-44
```
ekzyis, line 23:
```suggestion
async window.nostr.nip44.encrypt(pubkey, plaintext, version): string // returns encrypted payload as specified in nip-44
```

was confusing to me until I checked the code examples and realized this is to specify the version. Why not just write it out here like everything else?

46.md
```diff
@@ -82,12 +82,18 @@ These are mandatory methods the remote signer app MUST implement:
...
-- **nip04_encrypt**
+- **nip04_encrypt** (deprecated)
   - params [`pubkey`, `plaintext`]
   - result `nip4 ciphertext`
-- **nip04_decrypt**
+- **nip04_decrypt** (deprecated)
   - params [`pubkey`, `nip4 ciphertext`]
   - result [`plaintext`]
+- **nip44_encrypt**
+  - params [`pubkey`, `plaintext`, `v`]
```
ekzyis, line 92: Depending on the rename above

```suggestion
  - params [`pubkey`, `plaintext`, `version`]
```

mikedilger:
I think this PR needs:

1) Padding (I like my option2 but open to others)
2) Switch to sha256(parity + X) (33 byte input) since that is what all the libraries are doing and we are working around the library implementations currently (unless we insist on the bitcoin tradition of doing things our own way).
3) Specify base64 uses the RFC4648 alphabet and padding.
4) Paul's new test vectors under those changes

staab:
Check on 2 and 3, just waiting for test vectors and padding. On padding, we cover that in #686, but a lower-level version would probably be better.

44.md
```diff
@@ -0,0 +1,255 @@
...
+`optional` `author:paulmillr` `author:staab`
+
+The NIP introduces a new data format for keypair-based encryption. This NIP is versioned to allow multiple algorithm choices to exist simultaneously.
+
+The encrypted payload is encoded as base64 using the character set defined in RFC4648. The first byte of the payload represents the version. The rest depends on the version.
+
+Currently defined encryption algorithms:
+
+- `0x00` - Reserved
+- `0x01` - XChaCha20 with same key `sha256(parity + ecdh)` per conversation
```
mikedilger, line 16: Actually the parity byte is part of the output of ecdh for most libraries, so this would put it in twice.

I would put something like `ecdh(private_key, public_key) where the ecdh() function computes the shared point, and then returns the sha256 hash of the x-only component prefixed by the parity byte ('02').`

fiatjaf:
I don't even know how to query what RFC4648 is. Is that what the default `base64.encode()` function does in most languages?

paulmillr:
For padding, i've did a deep dive a few days ago and padme seems to be good:

https://crypto.stackexchange.com/questions/102711/stream-cipher-padding
https://github.com/jedisct1/go-padme-padding/blob/master/go.mod
https://github.com/jedisct1/rust-padme-padding/blob/master/src/lib.rs

Need to write it in js

44.md
paulmillr: except that the libraries implement ECDSA and not schnorr. There is no parity byte in schnorr.

paulmillr:
>Switch to sha256(parity + X) (33 byte input) since that is what all the libraries are doing and we are working around the library implementations currently (unless we insist on the bitcoin tradition of doing things our own way).

To reiterate:

1. ANSI X9.63 hashes X coordinate. This is ECDH standard used in NIST, P256, etc
2. secp256k1 bitcoin hashes (y-parity + X coordinate).

Both schemes do not consider Schnorr signatures which work over X coordinates without parity bytes like ECDSA. On the other hand, there seem to be a separate development for X-only-ECDH-over-secp256k1.

We need to decide if `parity + X` is better than `X`.

staab:
Are you guys able to make suggestions on the PR? This is beyond me.

44.md
mikedilger: I only understand a little. I thought the parity byte represented the Y coordinate and there were two options since the curve loops back around on itself. And I know that with Schnorr signatures it is always even (2) for some reason.  And I thought DH just did successive multiplications to come up to the same point  g^a * g^b = g^b * g^a and I don't understand where either signature algorithm (DSA or Schnorr) comes in when determining the DH shared point (and does ECDSA even have a shared point?).

All I want is for the NIP to do the default thing that libraries do, rather than a custom thing that requires lower-level operations.

mikedilger:
> I don't even know how to query what RFC4648 is. Is that what the default base64.encode() function does in most languages?

Section 4 of https://www.rfc-editor.org/rfc/rfc4648 specifies the alphabet and pad character. Some variants use different symbols for character 62 and 63 to be "url safe" for example. Some variants don't use padding characters (they are implied by the length). But that RFC is the original canonical **base64** and if  library doesn't give you options to pass into their base64() function, it is very probably the right one.

mikedilger:
> Besides everything, we need to analyze if the scheme is CCA-secure.

The lack of randomness in the choice of key, and then using XOR, leads me to think it is not.  If you take an unknown message, XOR it with your chosen plaintext, do you get back the secret stream? That would be a serious flaw.

mikedilger:
I like PADMÉ it is elegant.  But it presumes that if you want to send a 3-character message, using 6 characters would be a huge waste of space, and that there are probably lots of 3-character messages so no padding is needed.  But my "yes"/"no" example still stands.  In our case, given the overhead of every event, I think every message should be padded out to a minimum of 32 bytes.  So I'm proposing MAX(padme(len), 32)

paulmillr:
32 bytes min seems fine. Any objections? Maybe 24?

paulmillr:
NIP44 is not robust against [chosen-ciphertext attack](https://en.wikipedia.org/wiki/Chosen-ciphertext_attack). See [code that proves it](https://gist.github.com/paulmillr/38a983741dd929a6b6af7c05ceafbdaa).

1. Choose arbitrary $m_0 \neq m_1$
2. Receive challenge ciphertext $C = (c, \sigma)$
3. Generate a new keypair, sign $c$ again, ask challenger to decrypt $C' = (c, \sigma')$
4. $\sigma'$ is valid and CCA security is broken

We need to change everything here. I don't think poly1305 would help, since it can be forged.

paulmillr:
idea 1

```
encrypt(shared_conversation_key: bytes[32], nonce: bytes[32])

  kdf = hkdf_sha256(shared_conversation_key, salt: nonce, info: 'nip44', length: 56)
  ciphertext = xchacha(shared_conversation_key, nonce: kdf.slice(0, 24))
  mac = hmac_sha256(key: kdf.slice(24, 56), data: ciphertext)
  return 1 || nonce || ciphertext || mac
```

idea 2

```
encrypt(shared_conversation_key: bytes[32], nonce: bytes[32])

  kdf = hkdf_sha256(shared_conversation_key, salt: nonce, info: 'nip44', length: 74)
  ciphertext = chacha(kdf.slice(0, 32), nonce: kdf.slice(32, 44))
  mac = hmac_sha256(key: kdf.slice(44, 76), data: ciphertext)
  return 1 || nonce || ciphertext || mac
```

paulmillr:
the idea

```
encrypt(privA: bytes[32], pubB: bytes[32], nonce: bytes[32], plaintext: bytes)
  keys = hkdf_sha256(ecdh(privA, pubB), salt: nonce[0..24], info: 'nip44', length: 64)
  enc_key = keys[0..32]
  auth_key = keys[32..64]
  padded = plaintext.length || plaintext || pad_with_zeros(plaintext)
  ciphertext = chacha(enc_key, nonce: nonce[24..32], padded) // ietf rfc chacha
  mac = hmac_sha256(key: auth_key, data: ciphertext)
  return 1 || nonce || ciphertext || mac

decrypt(privA: bytes[32], pubB: bytes[32], ciphertext: bytes)
  mac_start_length = ciphertext.length - 24
  version = ciphertext[0..1]
  nonce = ciphertext[1..33]
  padded = ciphertext[33..mac_start_length]
  mac = ciphertext[mac_start_length..] 

  keys = hkdf_sha256(ecdh(privA, pubB), salt: nonce[0..24], info: 'nip44', length: 64)
  enc_key = keys[0..32]
  auth_key = keys[32..64]

  mac_ = hmac_sha256(key: kdf.slice(44, 76), data: ciphertext)
  if mac != mac_: throw Error()

  padded = chacha(enc_key, nonce: nonce[24..32], ciphertext)
  unpadded_length = padded[0]
  plaintext = padded[1..plaintext_length]
  zeros_length = pad_with_zeros(plaintext).length
  if (1 + unpadded_length + zeros_length) != padded.length: throw Error()
  return plaintext
```

paulmillr:
@staab I will continue this in #746, there are quite a few adjustments that need to be made. Can we close this one?

mikedilger:
I don't get it. So what if Mallory signs encrypted content that she doesn't understand.  Bob will get the message, verify the valid signature, and it will decrypt into garbage.

paulmillr:
> it will decrypt into garbage

Not certain, it may decrypt into a valid text sometimes.

> So what

It's definition of CCA. It could leak the key in extreme cases.

staab:
Closed in favor of #746

PR #716, opened 2023-08-11
Introduce NIP-59 gift wrap

staab:
A modified version of #468 focused on the wrapping standard, and omitting DM-specific stuff. @v0l please feel free to incorporate this diff into the original PR and I can close this one, or close the original and we can continue from here, it's up to you.

### DIFF
--- a/59.md
+++ b/59.md
@@ -1,0 +1,250 @@
+NIP-59
+======
+
+Gift Wrap
+---------
+
+`optional`
+
+This NIP defines a protocol for encapsulating any nostr event. This makes it possible to obscure most metadata
+for a given event, perform collaborative signing, and more.
+
+This NIP relies on [NIP-44](./44.md)'s versioned encryption algorithms.
+
+# Overview
+
+This protocol uses three main concepts to protect the transmission of a target event: `rumor`s, `seal`s, and `gift wrap`s.
+
+- A `rumor` is a regular nostr event, but is **not signed**. This means that if it is leaked, it cannot be verified.
+- A `rumor` is serialized to JSON, encrypted, and placed in the `content` field of a `seal`. The `seal` is then
+  signed by the author of the note. The only information publicly available on a `seal` is who signed it, but not what was said.
+- A `seal` is serialized to JSON, encrypted, and placed in the `content` field of a `gift wrap`.
+
+This allows the isolation of concerns across layers:
+
+- A rumor carries the content but is unsigned, which means if leaked it will be rejected by relays and clients,
+  and can't be authenticated. This provides a measure of deniability.
+- A seal identifies the author without revealing the content or the recipient.
+- A gift wrap can add metadata (recipient, tags, a different author) without revealing the true author.
+
+# Protocol Description
+
+## 1. The Rumor Event Kind
+
+A `rumor` is the same thing as an unsigned event. Any event kind can be made a `rumor` by removing the signature.
+
+## 2. The Seal Event Kind
+
+A `seal` is a `kind:13` event that wraps a `rumor` with the sender's regular key. The `seal` is **always** encrypted
+to a receiver's pubkey but there is no `p` tag pointing to the receiver. There is no way to know who the rumor is for
+without the receiver's or the sender's private key. The only public information in this event is who is signing it.
+
+```js
+{
+  "id": "<id>",
+  "pubkey": "<real author's pubkey>",
+  "content": "<encrypted rumor>",
+  "kind": 13,
+  "created_at": 1686840217,
+  "tags": [],
+  "sig": "<real author's pubkey signature>"
+}
+```
+
+Tags MUST must always be empty in a `kind:13`. The inner event MUST always be unsigned.
+
+## 3. Gift Wrap Event Kind
+
+A `gift wrap` event is a `kind:1059` event that wraps any other event. `tags` MUST include a single `p` tag
+containing the recipient's public key.
+
+The goal is to hide the sender's information, the metadata, and the content of the original event from the public.
+The only public information is the receiver's public key.
+
+```js
+{
+  "id": "<id>",
+  "pubkey": "<random, one-time-use pubkey>",
+  "content": "<encrypted kind 13>",
+  "kind": 1059,
+  "created_at": 1686840217,
+  "tags": [["p", "<Receiver>"]],
+  "sig": "<random, one-time-use pubkey signature>"
+}
+```
+
+# Encrypting Payloads
+
+Encryption is done following NIP-44 on the JSON-encoded event. Place the the encryption payload in the `.content`
+of the wrapper event (either a `seal` or a `gift wrap`).
+
+# Other Considerations
+
+If a `rumor` is intended for more than one party, or if the author wants to retain an encrypted copy, a single
+`rumor` may be wrapped and addressed for each recipient individually.
+
+The canonical `created_at` time belongs to the `rumor`. All other timestamps SHOULD be tweaked to thwart
+time-analysis attacks. Note that some relays don't serve events dated in the future, so all timestamps
+SHOULD be in the past.
+
+Relays may choose not to store gift wrapped events due to them not being publicly useful. Clients MAY choose
+to attach a certain amount of proof-of-work to the wrapper event per NIP-13 in a bid to demonstrate that
+the event is not spam or a denial-of-service attack.
+
+To protect recipient metadata, relays SHOULD guard access to kind 1059 events based on user AUTH. When
+possible, clients should only send wrapped events to relays that offer this protection.
+
+To protect recipient metadata, relays SHOULD only serve kind 1059 events intended for the marked recipient.
+When possible, clients should only send wrapped events to `read` relays for the recipient that implement
+AUTH, and refuse to serve wrapped events to non-recipients.
+
+# An Example
+
+Let's send a wrapped `kind 1` message between two parties asking "Are you going to the party tonight?"
+
+- Author private key: `0beebd062ec8735f4243466049d7747ef5d6594ee838de147f8aab842b15e273`
+- Recipient private key: `e108399bd8424357a710b606ae0c13166d853d327e47a6e5e038197346bdbf45`
+- Ephemeral wrapper key: `4f02eac59266002db5801adc5270700ca69d5b8f761d8732fab2fbf233c90cbd`
+
+## 1. Create an event
+
+Create a `kind 1` event with the message, the receivers, and any other tags you want, signed by the author.
+Do not sign the event.
+
+```json
+{
+  "created_at": 1691518405,
+  "content": "Are you going to the party tonight?",
+  "tags": [],
+  "kind": 1,
+  "pubkey": "611df01bfcf85c26ae65453b772d8f1dfd25c264621c0277e1fc1518686faef9",
+  "id": "9dd003c6d3b73b74a85a9ab099469ce251653a7af76f523671ab828acd2a0ef9"
+}
+```
+
+## 2. Seal the rumor
+
+Encrypt the JSON-encoded `rumor` with a conversation key derived using the author's private key and
+the recipient's public key. Place the result in the `content` field of a `kind 13` `seal` event. Sign
+it with the author's key.
+
+```json
+{
+  "content": "AqBCdwoS7/tPK+QGkPCadJTn8FxGkd24iApo3BR9/M0uw6n4RFAFSPAKKMgkzVMoRyR3ZS/aqATDFvoZJOkE9cPG/TAzmyZvr/WUIS8kLmuI1dCA+itFF6+ULZqbkWS0YcVU0j6UDvMBvVlGTzHz+UHzWYJLUq2LnlynJtFap5k8560+tBGtxi9Gx2NIycKgbOUv0gEqhfVzAwvg1IhTltfSwOeZXvDvd40rozONRxwq8hjKy+4DbfrO0iRtlT7G/eVEO9aJJnqagomFSkqCscttf/o6VeT2+A9JhcSxLmjcKFG3FEK3Try/WkarJa1jM3lMRQqVOZrzHAaLFW/5sXano6DqqC5ERD6CcVVsrny0tYN4iHHB8BHJ9zvjff0NjLGG/v5Wsy31+BwZA8cUlfAZ0f5EYRo9/vKSd8TV0wRb9DQ=",
+  "kind": 13,
+  "created_at": 1703015180,
+  "pubkey": "611df01bfcf85c26ae65453b772d8f1dfd25c264621c0277e1fc1518686faef9",
+  "tags": [],
+  "id": "28a87d7c074d94a58e9e89bb3e9e4e813e2189f285d797b1c56069d36f59eaa7",
+  "sig": "02fc3facf6621196c32912b1ef53bac8f8bfe9db51c0e7102c073103586b0d29c3f39bdaa1e62856c20e90b6c7cc5dc34ca8bb6a528872cf6e65e6284519ad73"
+}
+```
+
+## 3. Wrap the seal
+
+Encrypt the JSON-encoded `kind 13` event with your ephemeral, single-use random key. Place the result
+in the `content` field of a `kind 1059`. Add a single `p` tag containing the recipient's public key.
+Sign the `gift wrap` using the random key generated in the previous step.
+
+```json
+{
+  "content": "AhC3Qj/QsKJFWuf6xroiYip+2yK95qPwJjVvFujhzSguJWb/6TlPpBW0CGFwfufCs2Zyb0JeuLmZhNlnqecAAalC4ZCugB+I9ViA5pxLyFfQjs1lcE6KdX3euCHBLAnE9GL/+IzdV9vZnfJH6atVjvBkNPNzxU+OLCHO/DAPmzmMVx0SR63frRTCz6Cuth40D+VzluKu1/Fg2Q1LSst65DE7o2efTtZ4Z9j15rQAOZfE9jwMCQZt27rBBK3yVwqVEriFpg2mHXc1DDwHhDADO8eiyOTWF1ghDds/DxhMcjkIi/o+FS3gG1dG7gJHu3KkGK5UXpmgyFKt+421m5o++RMD/BylS3iazS1S93IzTLeGfMCk+7IKxuSCO06k1+DaasJJe8RE4/rmismUvwrHu/HDutZWkvOAhd4z4khZo7bJLtiCzZCZ74lZcjOB4CYtuAX2ZGpc4I1iOKkvwTuQy9BWYpkzGg3ZoSWRD6ty7U+KN+fTTmIS4CelhBTT15QVqD02JxfLF7nA6sg3UlYgtiGw61oH68lSbx16P3vwSeQQpEB5JbhofW7t9TLZIbIW/ODnI4hpwj8didtk7IMBI3Ra3uUP7ya6vptkd9TwQkd/7cOFaSJmU+BIsLpOXbirJACMn+URoDXhuEtiO6xirNtrPN8jYqpwvMUm5lMMVzGT3kMMVNBqgbj8Ln8VmqouK0DR+gRyNb8fHT0BFPwsHxDskFk5yhe5c/2VUUoKCGe0kfCcX/EsHbJLUUtlHXmTqaOJpmQnW1tZ/siPwKRl6oEsIJWTUYxPQmrM2fUpYZCuAo/29lTLHiHMlTbarFOd6J/ybIbICy2gRRH/LFSryty3Cnf6aae+A9uizFBUdCwTwffc3vCBae802+R92OL78bbqHKPbSZOXNC+6ybqziezwG+OPWHx1Qk39RYaF0aFsM4uZWrFic97WwVrH5i+/Nsf/OtwWiuH0gV/SqvN1hnkxCTF/+XNn/laWKmS3e7wFzBsG8+qwqwmO9aVbDVMhOmeUXRMkxcj4QreQkHxLkCx97euZpC7xhvYnCHarHTDeD6nVK+xzbPNtzeGzNpYoiMqxZ9bBJwMaHnEoI944Vxoodf51cMIIwpTmmRvAzI1QgrfnOLOUS7uUjQ/IZ1Qa3lY08Nqm9MAGxZ2Ou6R0/Z5z30ha/Q71q6meAs3uHQcpSuRaQeV29IASmye2A2Nif+lmbhV7w8hjFYoaLCRsdchiVyNjOEM4VmxUhX4VEvw6KoCAZ/XvO2eBF/SyNU3Of4SO",
+  "kind": 1059,
+  "created_at": 1703021488,
+  "pubkey": "18b1a75918f1f2c90c23da616bce317d36e348bcf5f7ba55e75949319210c87c",
+  "id": "5c005f3ccf01950aa8d131203248544fb1e41a0d698e846bd419cec3890903ac",
+  "sig": "35fabdae4634eb630880a1896a886e40fd6ea8a60958e30b89b33a93e6235df750097b04f9e13053764251b8bc5dd7e8e0794a3426a90b6bcc7e5ff660f54259"
+  "tags": [["p", "166bf3765ebd1fc55decfe395beff2ea3b2a4e0a8946e7eb578512b555737c99"]],
+}
+```
+
+## 4. Broadcast Selectively
+
+Broadcast the `kind 1059` event to the recipient's relays only. Delete all the other events.
+
+# Code Samples
+
+## JavaScript
+
+```javascript
+import {bytesToHex} from "@noble/hashes/utils"
+import type {EventTemplate, UnsignedEvent, Event} from "nostr-tools"
+import {getPublicKey, getEventHash, nip19, nip44, finalizeEvent, generateSecretKey} from "nostr-tools"
+
+type Rumor = UnsignedEvent & {id: string}
+
+const TWO_DAYS = 2 * 24 * 60 * 60
+
+const now = () => Math.round(Date.now() / 1000)
+const randomNow = () => Math.round(now() - (Math.random() * TWO_DAYS))
+
+const nip44ConversationKey = (privateKey: Uint8Array, publicKey: string) =>
+  nip44.v2.utils.getConversationKey(bytesToHex(privateKey), publicKey)
+
+const nip44Encrypt = (data: EventTemplate, privateKey: Uint8Array, publicKey: string) =>
+  nip44.v2.encrypt(JSON.stringify(data), nip44ConversationKey(privateKey, publicKey))
+
+const nip44Decrypt = (data: Event, privateKey: Uint8Array) =>
+  JSON.parse(nip44.v2.decrypt(data.content, nip44ConversationKey(privateKey, data.pubkey)))
+
+const createRumor = (event: Partial<UnsignedEvent>, privateKey: Uint8Array) => {
+  const rumor = {
+    created_at: now(),
+    content: "",
+    tags: [],
+    ...event,
+    pubkey: getPublicKey(privateKey),
+  } as any
+
+  rumor.id = getEventHash(rumor)
+
+  return rumor as Rumor
+}
+
+const createSeal = (rumor: Rumor, privateKey: Uint8Array, recipientPublicKey: string) => {
+  return finalizeEvent(
+    {
+      kind: 13,
+      content: nip44Encrypt(rumor, privateKey, recipientPublicKey),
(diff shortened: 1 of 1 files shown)


59.md
```diff
@@ -0,0 +1,292 @@
...
+}
+```
+
+## 2. Seal the rumor
+
+Encrypt the JSON-encoded `rumor` with a conversation key derived using the **Sender**'s private key and the recipient's public key. JSON-encode the resulting object and place it in the `content` field of a `kind 13` `seal` event. Sign it with the **Sender**'s key.
+
+```json
+{
+  "content": "{\"ciphertext\":\"f5Lfq6a9laLW9PWuYatTTnrGOSHcGJUPnb5OQ/gEEVjtDnrOzDB9VFKdUkoImEMiO6Ddu+HYn0R3hUtqgTRppOjs4zNFTLNAP/3hkN3hAaKjvUzBVYVmU1xrBS749kUhCt8v/8l/tBeen9BYB4s1luAUqdS10h8jwM+NqE8aYrv9sEU9AaLuErbI1hwq0ZUEGl5unfxHpv+QjINbrPYs1x6bGUU7D416Lp0gV6fr/wD/JLkceb/quAE2UkgzVRS4gauWL8/PyaPz3332h+vH/tMWOQd2no4sDG3ftXPO8y0zXs5p/NvcPa2hIXsOsNOPzOfCFvBk7Rfv9JjwCN0K/LztDHZOzgryELAjcm+mDWPeYBw+PYWfwayFkDbbLCsQJ1KLYRBDS7BEMBOLdd1v1vF4J196tGaN\",\"nonce\":\"s6DBwU6gq9h14+9hRwnts1qZ6BdyPgHn\",\"v\":1}",
```
jb55, line 146: is there a reason these fields are as json in the content instead of just tags?

vitorpamplona:
1. If we want to break this down into multiple NIPs, the text here should be much smaller. 

Just GiftWrap + Seals of unsigned objects. 

No whys. Only hows.  

2. I don't understand the need for the `kind:author wrapper->[recipient1,recipient2]` notation.

59.md
staab: The payload is defined by NIP 44 (#715), so it can't be split across tags. @vitorpamplona wanted it to be json. I personally prefer @paulmillr's original version that is CSV oriented, but Paul has also suggested doing `base64.encode(JSON.stringify({cyphertext, nonce, v}))`. It's just bikeshedding at this point.

jb55: @vitorpamplona what is the problem with large tags?

jb55: as someone writing a relay and database, all I can say is anytime I have to parse json it's a pita. tags are in a decoded form already.

vitorpamplona: > @vitorpamplona what is the problem with large tags?

Indexing of large tags is bad. A Relational Implementation of Nostr will not have a choice to remove certain tags before indexing. I would generally avoid breaking their implementation if we can

jb55: relays only index single-char tags

vitorpamplona:
> Any group member may post to the group using any:author shared_key->[].

Sorry, I want to like this, but this is not better than the prose in the same example. There is a lot of hidden information in that notation.

59.md
staab: You shouldn't have to parse this payload on a relay unless you're decrypting the payload, it's intended to be opaque.

jb55:
>+  "content": "{\"ciphertext\":\"f5Lfq6a9laLW9PWuYatTTnrGOSHcGJUPnb5OQ/gEEVjtDnrOzDB9VFKdUkoImEMiO6Ddu+HYn0R3hUtqgTRppOjs4zNFTLNAP/3hkN3hAaKjvUzBVYVmU1xrBS749kUhCt8v/8l/tBeen9BYB4s1luAUqdS10h8jwM+NqE8aYrv9sEU9AaLuErbI1hwq0ZUEGl5unfxHpv+QjINbrPYs1x6bGUU7D416Lp0gV6fr/wD/JLkceb/quAE2UkgzVRS4gauWL8/PyaPz3332h+vH/tMWOQd2no4sDG3ftXPO8y0zXs5p/NvcPa2hIXsOsNOPzOfCFvBk7Rfv9JjwCN0K/LztDHZOzgryELAjcm+mDWPeYBw+PYWfwayFkDbbLCsQJ1KLYRBDS7BEMBOLdd1v1vF4J196tGaN\",\"nonce\":\"s6DBwU6gq9h14+9hRwnts1qZ6BdyPgHn\",\"v\":1}",
>
>You shouldn't have to parse this payload on a relay unless you're
>decrypting the payload, it's intended to be opaque.

If you say so. The database library I'm building is meant for clients
and will be processing notes on behalf of client UIs, and you can query
it like a relay. It seems like you guys have already made up your minds
on this point so I will not waste my time further.

staab:
> The database library I'm building is meant for clients

No, that's a good point. I don't think we can put this in tags because it needs to be passed to `decrypt` functions in signer applications, but would CSV or something else make your life easier compared to JSON?

staab:
> > Any group member may post to the group using any:author shared_key->[].
> 
> Sorry, I want to like this, but this is not better than the prose in the same example. There is a lot of hidden information in that notation. It will just complicate things for new NIP readers.

I'll move it into NIP 87 and we can re-introduce it at some point if it seems helpful.

jb55:
On Fri, Aug 11, 2023 at 09:21:59AM -0700, Jon Staab wrote:
>> The database library I'm building is meant for clients
>
>No, that's a good point. I don't think we can put this in tags because
>it needs to be passed to `decrypt` functions in signer applications,
>but would CSV or something else make your life easier compared to JSON?

It's not a big deal, I just think the trend of putting JSON in the
content makes everything a lot more annoying on the processing side.

Parsing JSON is *slow* and blocks UIs, so you almost always need to put
it in a worker thread somewhere. This is why in my database library I'm
doing all the heavy lifting and processing for client UIs so they don't
have to worry about it: profiles are automatically decoded into
flatbuffers and stored in the DB.

In these cases, JSON actually makes sense, because there are some many
fields that can change. For things as simple as 3 standard fields, tags
just makes way more sense and are much simpler.

We have tags, why not use them!?

Again, if ya'll already have it implemented as JSON and don't want to
change it, then fine, I will deal with it. I just wanted to voice my
distaste for JSON in content yet again xD

59.md
```diff
@@ -0,0 +1,258 @@
+NIP-59
+======
+
+Gift Wrap
+---------
+
+`draft` `optional` `authorPrivateKey:vitorpamplona` `author:kieran` `author:paulmillr` `author:staab`
```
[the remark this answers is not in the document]
staab: heh global find and replace

alexgleason (approved):
This is a very cool and interesting MR. It took some effort to understand why it has to go through 3 rounds of events. It's excessive. But I think it's a good solution.

mikedilger (approved):
I just read this NIP and I fully understand it.  Great design guys @v0l @vitorpamplona.  Well written @staab.  I also like separating this from particular DM implementations which can be NIPs on top of this.

v0l:
We should encourage gift wraps to have PoW, problems for relays accepting gift wraps when the only reliable info is the recipient

staab:
Agreed, that's in there:

> Relays may choose not to store gift wrapped events due to them not being publicly useful. Clients MAY choose to attach a certain amount of proof-of-work to the wrapper event per NIP-13 in a bid to demonstrate that the event is not spam or a denial-of-service attack.

vitorpamplona:
Dont we need some type of ring signature scheme for Relays? To the best of my knowledge, a ring signature would allow a user to sign a message that proves he/she is part of a set of white-listed keys without revealing which one. The GiftWrap would then have an additional tag that contains a ring signature to prove the sender is either a paying account or a white-listed user. I am not sure what's needed to create the "ring" tough. 

We can also build a delete-by-tag for GiftWraps only. If I receive a GiftWrap and ask to delete it, the Relay could oblige.

paulmillr:
Ring sigs have questionable privacy. There have been countless attacks on related currencies. Not sure if it's even worth pursuing. Also, how would one mix it, interactively, or not, would the keys be real, or decoys

staab:
I like this idea, and while I don't think @arthurfranca's anonymity set idea is the best we can do (it's been proposed many times, several of them by me), using something other than a pubkey could be cool if we can avoid tracking more state. For example, if there were a way to create a shared secret with a known salt that the recipient could recognize. So for example if the sender could do `getSharedId(myPrivateKey, recipientPublicKey, 'challenge')` and the recipient could do `getSharedId(myPrivateKey, senderPublicKey, 'challenge')` and get the same result. The sender could then tag that id and the recipient could search for it. The `challenge` string would be a set of ~100 known strings defined in an NIP somewhere so every user has 100 state-free aliases per other user. @paulmillr would something like this be possible?

That said, I think using `p` tags is probably the best idea, because the anonymity set is all pubkeys + all aliases, and you can't tell by which tag the sender uses whether a pubkey or alias is being used.

paulmillr:
How does the whole flow look like?

1. Sender Alice/recipient Bob do `id = getSharedId()`
2. They query a relay for the `id` tag in `p`, or in any other way
3. Relay obviously keeps all the logs to itself and sees that both Alice and Bob queried for `id` from IPs where they queried their own messages with other ppl

Meh?

vitorpamplona:
> Does the negotiated aliases approach not have this problem?

Aliases are not negotiated, they are just advertised. Clients can just listen in to changes in them via the usual GiftWrap filter. Relays won't know which package is an Alias, which is a message. 

Now because aliases will be part of a filter that comes from the same IP and the main PubKey filter, whoever sees that filter can see the alias group. So, a GiftWrap relay is a trusted one to know all your aliases. 

Aliases don't protect against the relay you use to receive GiftWraps. They protect against the public.

staab:
> Aliases are not negotiated, they are just advertised

I sort of see it as the same thing, another question I had is how you envision these being shared? Can someone request an alias? Or are they just sent after the first message to hide subsequent messages? IOW you'd still have at least 1 note addressed to the recipient's pubkey.

vitorpamplona:
> how you envision these being shared? 

I think Amethyst would send a set of 3 random keys out of two sets of 25 keys (25 for known contacts, 25 for random people) the first time the logged-in user replies to a message. So, from that moment forward, I would expect the conversation to flow through different channels. 

There is some risk of people probing for keys, so I am not completely sure about this approach. 

Then upon receiving Wraps Amethyst would unwrap and re-wrap them with one of the DM Backup keys. So that all messages can be recovered with those keys and the aliases can be fully deleted/rotated at random times without concern for the past. 

There are lots of (dangerous) flexibility in this approach.

paulmillr:
>Aliases don't protect against the relay you use to receive GiftWraps. They protect against the public.

How so? Could you describe the flow with and without aliases, what's the exact benefit?

vitorpamplona:
> Could you describe the flow with and without aliases, what's the exact benefit?

The idea is to avoid the public listening to all GiftWraps with `p` to me from all relays. The first message between two users would generally go through the main public key but from that moment forward, both clients can use secondary keys to avoid message counting.

paulmillr:
Do you think this is massively better than limitting chatting to only relays user specified? I think limits can also be fine against the public

vitorpamplona:
> Do you think this is massively better than limiting chatting to only relays user specified?

I would do both. 

People know which relays I use for GiftWraps (because they need to send them there) so they can be listening into those as well. You can opt for a private relay that doesn't allow the public to download events with p-tag = you, but that doesn't exist today.

vitorpamplona:
The interesting part is that if you are doing PushNotifications for GiftWraps as I do, you would transfer who can see the alias set from Relays to Google and Amethyst's Push Server. That's because the Push server doesn't listen for keys. It listens for every `kind:1059` from all relays. So, relays won't know a user's alias set from the filter. 

I can probably encrypt events to Google's device ID and decrypt them when the app receives them. In that way, Google or any other Push platform, can't see the p-tag in the GiftWrap and associate them all with the same device ID. 

Amethyst Push Notification server then would be the only weak point in this.

paulmillr:
> People know which relays I use for GiftWraps (because they need to send their messages there) so they can be listening into those as well

Yeah and they won't see your messages because only you can fetch your messages.

mikedilger:
NIP-44 now already encodes (version, nonce, ciphertext) into a single base64 string. We no longer need the content of a seal or a gift wrap to be JSON. It should just be the base64 output of NIP44.

vitorpamplona:
FYI Amethyst is now using the base64 byte array concat for both the seal and the giftwrap on NIP-24 messages. The new version was just updated on the PlayStore.

arthurfranca:
> I think Amethyst would send a set of 3 random keys out of two sets of 25 keys (25 for known contacts, 25 for random people)

So one needs to 1) somehow convince you to add him as a "known contact" (follow?) and 2) convince you to reply to a DM of a random pubkey. Now he can keep track of all your current used aliases.

Unless you make it a parameterized replaceable event, so a set of aliases to each contact. But the complexity rises.

vitorpamplona:
> So one needs to 1) somehow convince you to add him as a "known contact" (follow?) and 2) convince you to reply to a DM of a random pubkey. Now he can keep track of all your current used aliases.

Yeah, lots of social attacks can be made. :( 

Maybe we just do 3 keys per contact and call it a day.

staab:
Just added that. Deletion is an interesting wrinkle. In any implementation which uses shared keys (#706), this makes it possible for one someone to delete someone else's event.

vitorpamplona:
> In any implementation which uses shared keys (https://github.com/nostr-protocol/nips/pull/706), this makes it possible for one someone to delete someone else's event.

Agree, that means the shared key should be either controlled by the admin or not controlled at all. If the shared key is the sum of all keys in the group then it would require a sum of all receiver's private keys to delete it.

arthurfranca:
> [Specify that gift wrap may be deleted by the recipient as well as by the author](https://github.com/nostr-protocol/nips/pull/716/commits/fd914c600fd590e6188534cd7555116da7855503) 

@staab @vitorpamplona you want a new behavior for the gift wrap event kind which needs relays to update their code. Next time another event kind wants the same behavior, relays would need to update their code again. For example a [kind 30078 - app-specific data](https://github.com/nostr-protocol/nips/blob/master/78.md) is so multi-purpose that it may need the same behavior.

With [this PR](https://github.com/nostr-protocol/nips/pull/539), we make the said behavior generic, usable by any kind. The gift wrap, or any other event, would then just have to use `!` tag instead of `p`.

59.md
```diff
@@ -0,0 +1,266 @@
...
+
+This NIP defines a protocol for encapsulating any nostr event. This makes it possible to obscure almost all metadata for a given event, perform collaborative signing, and more.
+
+This NIP relies on [NIP-44](44.md)'s versioned encryption algorithms.
+
+# Overview
+
+This protocol uses three main concepts to protect the transmission of a target event: `rumor`s, `seal`s, and `gift wrap`s.
+
+- A `rumor` is a regular nostr event, but is **not signed**. This means that if it is leaked, it cannot be verified.
```
alexgleason, line 17: I'm skeptical about calling this a rumor. Even though the name perfectly describes it, it doesn't fit the paradigm of `EventTemplate` -> `UnsignedEvent` -> `Event`.

https://github.com/nbd-wtf/nostr-tools/blob/master/event.ts#L45

I think it is most close to `UnsignedEvent`, except that in UnsignedEvent the `id` is optional.

`rumor` is also shorter and has a specific connotation that `UnsignedEvent` doesn't. So I have mixed feelings.

mikedilger: My library nostr-types takes a `PreEvent` into an `Event` by computing an `Id` and adding a `Signature`.   A `Rumor` is half-way between those two concepts.  But I just did the pragmatic thing and made a new type, and the code that turns a PreEvent to an Event doesn't pass it through the Rumor stage.  Does it look like it could be refactored. Sure. But I'm not sure.

staab: >  it doesn't fit the paradigm of EventTemplate -> UnsignedEvent -> Event.

Seems to me a good argument for a new term. Although arguably according to the above paradigm it should be something more like `EventTemplate` -> `RawEvent` -> `UnsignedEvent` -> `Event`.

staab:
Updated with the following changes:

- Updated formatting
- Re-worded some "other considerations" with shared keys in mind
- New examples based on my [NIP 44 PR to nostr-tools](https://github.com/nbd-wtf/nostr-tools/pull/326)
- Added a new deletion strategy. Instead of `!` or special casing kind-1059 to be deletable by recipients (which breaks shared keys), clients can just include the wrapper private key on the seal's tags if they want the recipient to be able to delete the wrapper. Any problems with that?

alexgleason (approved):
There are trade-offs to this approach. It works and it solves all of the stated problems. The tradeoff I'm most concerned with is people flooding other people's inboxes with gift-wrapped events as a form of DoS. If the network becomes hostile it will become a problem if we rely too heavily on it. But I think it should exist as a NIP.

alexgleason:
Also, it's Christmas. Now is the best time to merge gift wrap. :christmas_tree:

staab:
> The tradeoff I'm most concerned with is people flooding other people's inboxes with gift-wrapped events as a form of DoS.

It could definitely happen. Lots of possible solutions, including auth'd write, pay to relay, etc.

vitorpamplona:
This PR is written in a very DM-based structure for GiftWraps and Seals. 

Do we think all uses of GiftWraps require Seals? I don't think so. 

For instance, If we do onion routing with GiftWraps to DVMs, the package can be 
`wrap(dvm3, wrap(dvm2,wrap(dvm1,wrap(receiver,seal(msg)))))`

We can probably also delete the code part from this since the complicated part is really in the NIP-44 file. 

Maybe event the example is too specific for this generic approach of GiftWraps.

staab:
> Do we think all uses of GiftWraps require Seals

Private information often benefits from deniability. The seals part is useful to NIP 87, at least. I think it will be desired more often than not. Onion routing and other use cases can be specified separately.

> We can probably also delete the code part from this since the complicated part is really in the NIP-44 file.

I found this really annoying/confusing to implement. It's helpful to have a test case to work off of.

vitorpamplona:
I think my point is that we could improve the writing to be less linear and provide some information to when to use Seals and when to just stick to basic wraps.

59.md
```diff
@@ -0,0 +1,253 @@
...
+```
+
+# Encrypting Payloads
+
+Encryption is done following NIP-44 on the JSON-encoded event. Place the the encryption payload in the `.content`
+of the wrapper event (either a `seal` or a `gift wrap`).
+
+# Other Considerations
+
+Because ephemeral keys are often used to sign `gift wrap`s, a `seal` MAY include a `wrap_privkey` tag that
```
vitorpamplona, line 80: This `wrap_privkey` links sender and receiver in a verifiable manner and can be revealed by the receiver at any time without any damage to the receiver's data. Adding it requires trusting the receiver and thus, I believe it requires some privacy risk warning.

staab: I don't understand, if the key is ephemeral what information is being shared?

vitorpamplona: If the key to the gift wrap is added to the seal, you are cryptographically linking them. The point of separating them was to unlink it. :)

staab: I see what you mean. Would putting it on the rumor be any better?

vitorpamplona: I think the best way would be to do https://github.com/nostr-protocol/nips/pull/539 with the receiver's key. Or simply make it so any GiftWrap can be deleted by the `p`-tag.

staab: Neither of those works for groups, since the shared key is the recipient (neither does this I guess). So anyone in the group would be able to delete the message. In that scenario the actual publisher would also have to hold on to the generated key in order to delete the event. But I guess if clients implement deletes, it would be enough to delete the rumor's id instead of the wrap's.

staab: I've removed the paragraph, so resolved, but still unsolved.

vitorpamplona:
@staab I was playing around and I think this defines better responsibilities between functions for the example. 

```js
import { bytesToHex } from "@noble/hashes/utils"
import type {EventTemplate, UnsignedEvent, Event} from "nostr-tools"
import {getPublicKey, getEventHash, nip19, nip44, finalizeEvent, generateSecretKey} from "nostr-tools"

type Rumor = UnsignedEvent & {id: string}

const TWO_DAYS = 2 * 24 * 60 * 60

const now = () => Math.round(Date.now() / 1000)
const randomNow = () => Math.round(now() - (Math.random() * TWO_DAYS))

const nip44ConversationKey = (privateKey: Uint8Array, publicKeyHex: string) => 
  nip44.v2.utils.getConversationKey(bytesToHex(privateKey), publicKeyHex)

const nip44Encrypt = (data: EventTemplate, privateKey: Uint8Array, publicKeyHex: string) => 
  nip44.v2.encrypt(JSON.stringify(data), nip44ConversationKey(privateKey, publicKeyHex))

const nip44Decrypt = (data: Event, privateKey: Uint8Array) => 
  JSON.parse(nip44.v2.decrypt(data.content, nip44ConversationKey(privateKey, data.pubkey)))

const createRumor = (event: Partial<UnsignedEvent>, privateKey: Uint8Array) => {
  const rumor = {
    created_at: now(),
    content: "",
    tags: [],
    ...event,
    pubkey: getPublicKey(privateKey),
  } as any

  rumor.id = getEventHash(rumor)

  return rumor as Rumor
}

const createSeal = (rumor: Rumor, privateKey: Uint8Array, toPublicKeyHex: string) => {
  return finalizeEvent(
    {
      kind: 13,
      content: nip44Encrypt(rumor, privateKey, toPublicKeyHex),
      created_at: randomNow(),
      tags: [], 
    }, 
    privateKey
  ) as Event
}

const createWrap = (rumor: Rumor, toPublicKeyHex: string) => {
  const randomKey = generateSecretKey()

  return finalizeEvent(
    {
      kind: 1059,
      content: nip44Encrypt(rumor, randomKey, toPublicKeyHex),
      created_at: randomNow(),
      tags: [["p", toPublicKeyHex]], 
    }, 
    randomKey
  ) as Event
}

// Test case using the above example
const senderPrivateKey = nip19.decode(`nsec1p0ht6p3wepe47sjrgesyn4m50m6avk2waqudu9rl324cg2c4ufesyp6rdg`).data
const recipientPrivateKey = nip19.decode(`nsec1uyyrnx7cgfp40fcskcr2urqnzekc20fj0er6de0q8qvhx34ahazsvs9p36`).data
const recipientPublicKeyHex = getPublicKey(recipientPrivateKey)

const rumor = createRumor(
  {
    kind: 1,
    content: "Are you going to the party tonight?",
  }, 
  senderPrivateKey
)

const seal = createSeal(rumor, senderPrivateKey, recipientPublicKeyHex)
const wrap = createWrap(seal, recipientPublicKeyHex)

// Receiver unwraps with his/her private key.

const unwrappedSeal = nip44Decrypt(wrap, recipientPrivateKey)
const unsealedRumor = nip44Decrypt(unwrappedSeal, recipientPrivateKey)

console.log(unsealedRumor)
```

vitorpamplona:
BTW, I believe all these utility functions should be provided by nostr-tools.

staab:
> I think this defines better responsibilities between functions for the example.

What's the difference aside from order of arguments?

> BTW, I believe all these utility functions should be provided by nostr-tools.

Maybe we can add them in a new nip 59 file.

vitorpamplona:
> What's the difference aside from order of arguments?

Small things like forcing tags to be empty in the Seal, fixed p-tag in the Wrap, new random key inside the wrap, forced pubkey in the rumor as well as updates it to use the finalizeEvent method

59.md
```diff
@@ -0,0 +1,242 @@
...
+# Protocol Description
+
+## 1. The Rumor Event Kind
+
+A `rumor` is the same thing as an unsigned event. Any event kind can be made a `rumor` by removing the signature.
+
+## 2. The Seal Event Kind
+
+A `seal` is a `kind:13` event that wraps a `rumor` with the sender's regular key. The `seal` is **always** encrypted
+to a receiver's pubkey but there is no `p` tag pointing to the receiver. There is no way to know who the rumor is for
```
monlovesmango, line 38:
```suggestion
with a conversation key derived from the author's private key and the recipient's public key and there is no `p` tag pointing to the receiver. There is no way to know who the rumor is for
```

monlovesmango:
overall this is a neat idea.

I do have one dumb question, how does the recipient obtain the ephemeral wrapper key to unwrap the seal? Is there another NIP I need to read up on to understand how this works?

vitorpamplona:
> I do have one dumb question, how does the recipient obtain the ephemeral wrapper key to unwrap the seal? Is there another NIP I need to read up on to understand how this works?

You unwrap with the receiver's private key. That's the only key you need. 

It works because the wrap is encrypted to the recepient's pub key directly, just like nip04 did. In other words, the conversation key of the random private key + the recipient pub key is equal to the conversation key of the random pubkey + recipient private key.

monlovesmango:
ahh ok its making a conversation key with the random key and recipients public key, very slick. thank you!

59.md
```diff
@@ -0,0 +1,242 @@
...
+  "tags": [],
+  "sig": "<real author's pubkey signature>"
+}
+```
+
+Tags MUST must always be empty in a `kind:13`. The inner event MUST always be unsigned.
+
+## 3. Gift Wrap Event Kind
+
+A `gift wrap` event is a `kind:1059` event that wraps any other event. `tags` MUST include a single `p` tag containing the recipient's public key.
```
monlovesmango, line 57:
```suggestion
A `gift wrap` event is a `kind:1059` event that wraps any other event. `tags` MUST include a single `p` tag containing the recipient's public key. The `gift wrap` content is **always** encrypted
with a conversation key derived from an ephemeral, random private key and the recipient's public key
```

59.md
```diff
@@ -0,0 +1,242 @@
...
+  "pubkey": "611df01bfcf85c26ae65453b772d8f1dfd25c264621c0277e1fc1518686faef9",
+  "tags": [],
+  "id": "28a87d7c074d94a58e9e89bb3e9e4e813e2189f285d797b1c56069d36f59eaa7",
+  "sig": "02fc3facf6621196c32912b1ef53bac8f8bfe9db51c0e7102c073103586b0d29c3f39bdaa1e62856c20e90b6c7cc5dc34ca8bb6a528872cf6e65e6284519ad73"
+}
+```
+
+## 3. Wrap the seal
+
+Encrypt the JSON-encoded `kind 13` event with your ephemeral, single-use random key. Place the result in the `content` field of a `kind 1059`. Add a single `p` tag containing the recipient's public key. Sign the `gift wrap` using the random key generated in the previous step.
```
monlovesmango, line 139:
```suggestion
Encrypt the JSON-encoded `kind 13` event with a conversation key derived using the ephemeral, single-use random key and the recipient's public key. Place the result in the `content` field of a `kind 1059`. Add a single `p` tag containing the recipient's public key. Sign the `gift wrap` using the random key generated in the previous step.
```

mikedilger:
Weren't we going to make it such that GiftWrap events could be deleted by the 'p' tagged person rather than the author?  I don't see it in there.  Was there some problem with that idea?  #945  Shared Key DM proposal sends a private key for this purpose which I like less.

vitorpamplona:
I think half didn't like the fact that the p-tag deletion was a rule made just for GiftWraps (but I do think it is a great rule) and the other what was waiting for the resolution to the #539 idea. Both can work for GiftWraps.

staab:
p tag for deletion can't work if there is a shared key (e.g. with my nip 87 groups proposal).

vitorpamplona:
Other than the deletion debate, I think we are ready to merge this.

staab:
Squashed and rebased, ready to merge.

59.md
```diff
@@ -0,0 +1,250 @@
...
+  "tags": [],
+  "sig": "<real author's pubkey signature>"
+}
+```
+
+Tags MUST must always be empty in a `kind:13`. The inner event MUST always be unsigned.
+
+## 3. Gift Wrap Event Kind
+
+A `gift wrap` event is a `kind:1059` event that wraps any other event. `tags` MUST include a single `p` tag
```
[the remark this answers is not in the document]
vitorpamplona: Or, since there are privacy implications, we may just need to define a separate wrap kind where tags are always empty and the pubkey is not random. 

The random pubkey kind is like UDP (just push events) and what you want is a TCP-like kind (ack-first-then-push). 

Either wrap can carry the same DM kinds. And you can define your authorization-to-receive-scheme in the wrap kind itself.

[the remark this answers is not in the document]
staab: Removed this part and added some clarifications that NIP 59 shouldn't be used on its own.

mikedilger:
I've already approved, but I like the changes and reiterate my +1

staab:
@fiatjaf ok to merge? I don't want to merge my own PR

vitorpamplona (approved):

vitorpamplona:
@fiatjaf can we merge?

PR #358, opened 2023-03-13
NIP-96: File Distribution over Nostr

KaffinPX:
I'm proposing and developing a solution for managing files/repositories over Nostr without storing actual data on it but on an alternate network.

I will release our CLI client tomorrow(Currently im testing&improving code quality) and our web client coming in some days(Its a working prototype rn).

Will release screenshots if comments ;)

PR #719, opened 2023-08-13
nip-97 version 2

ondra-novak:
Second attempt (previous attempt #694)

Working demo client and relay 
https://nostr-test.novacisko.cz/files/index.html

The relay is reset every midnight UTC, so use it for testing.

This proposal is built on NIP-94.

A new feature is ability to retrieve URL to a binary content through field added by relay (as the relay know, where the file is stored).

### DIFF
--- a/97.md
+++ b/97.md
@@ -1,0 +1,217 @@
+NIP-97: Files hosted on relay
+==============================
+
+`final` `optional` `author:ondra-novak`
+
+This NIP solves the problem of sharing and distributing binary content over NOSTR networks such as media - images, short videos, music, or in general any files up to a certain size
+
+Key Takeaways
+-------------
+
+* Builds on NIP-94
+* Defines how binary content (images, media, files) is managed by NOSTR network
+* Allows sending private binary content in direct messages (encrypted)
+* Defines new commands at the protocol level ("FILE","RETRIEVE")
+* Uses the binary messages in the websocket connection
+
+
+Changes in File header (NIP-94)
+--------------------------------
+
+```json
+{
+  "id": <32-bytes lowercase hex-encoded sha256 of the the serialized event data>,
+  "pubkey": <32-bytes lowercase hex-encoded public key of the event creator>,
+  "created_at": <unix timestamp in seconds>,
+  "kind": 1063,
+  "tags": [
+    ["f","file"],  
+    ["m", <MIME type>],
+    ["x",<Hash SHA-256>],
+    ["size", <size of file in bytes>],
+    ["url",<optional - string with URI of file>],
+    ["aes-256-gcm",<key>, <iv>],
+    ["dim", <size of file in pixels>],
+    ["blurhash", <value>]
+  ],
+  "content": <description>,
+  "sig": <64-bytes hex of the signature of the sha256 hash of the serialized event data, which is the same as the "id" field>
+}
+```
+
+* **url is optional** - This item can be still filled with a link to an external server. This may help during the transition period when the client is uploading the file to third party servers for relays that don't support this NIP. 
+
+
+* new tag **f** - it means *indexed flag*. Currently NOSTR protocol doesn't have a way to assign arbitrary flags to  events. In this case, the flag has the value `file`. This field is mandatory and it's purpose is to indicate presence of a binary content associated with this event.
+
+
+**NOTE** - when copying events from one relay to another, the service must check for the presence of a tag `["f":"file"]`. If such a tag exists on the event, it must also ensure that the binary content is copied as well
+
+**Compatibility note** - Since there are probably many relays and proxies that do not support this NIP, it may happen that the event will be copied without the binary content. If a client discovers, that the relay does not have that content, or it doesn't even support this NIP, the client is not required to look for that content elsewhere, instead it can indicate to the user the unavailability of that content.
+
+
+Upload
+------
+
+A new protocol level command is introduced for upload: `FILE`
+
+### Protocol flow
+
+```
+client: ["FILE", <file-header-event>]
+relay:  ["OK","<event-id">,true,"continue"]
+client: <binary message - content of the file>
+relay:  ["OK",<event-id>, true, ""]
+```
+
+The client MUST notify the relay that it is about to send a binary file. This is just implemented by the new 'FILE' command. The command "allocates" the binary channel of the websocket connection for file transfer. This will allow this binary channel to be used for other purposes in the future. 
+
+
+The command parameter is a modified `file header event`(see above) which must contain the tags `x`,`m`,`f` and `size`.
+
+* **f** - MUST be presented with value `file` 
+* **x** - MUST contain SHA256 of the binary content stored as HEX
+* **m** - MUST contain content type as mime specification. If the content type is unknown, use `application/octet-stream`
+* **size** - MUST contain size of the binary content.
+
+### Binary message
+
+The standard binary message is used as defined in the websocket specification. The message can be fragmented (continuation frames can be used)
+
+### Responses 
+
+The relay MUST always respond with a NIP-20 response to both the command and the binary message.
+
+Response to the command `FILE` is `OK`, with status `true` - which signals that relay is ready to accept a binary message as binary content of the file. The text part of the response is just informational for debugging purpose.
+
+If the relay respons with status `false`, or the response is not `OK` message, the client MUST NOT send any binary message. The text part of the response contains explanation of the error (similar to NIP-20)
+
+The relay must also respond once it receives the binary message if it was correctly announced using the `FILE` command. The response is also `OK`, with status `true` for success or `false` for failure.
+
+This proposal doesn't define response for binary messages sent without the announcement
+
+### Special error messages
+
+* `max_size: <size in bytes>` - this error message informs that file is too big.
+* `invalid: file mismatch` - received binary content doesn't match to announced metadata in the associated event
+
+### Rules for state control
+
+1. The relay SHOULD NOT accept any binary message without prior anouncement.
+2. The relay SHOULD publish event after successful binary transfer.
+3. The relay MUST NOT publish an event with incomplette binary transfer.
+4. If a `FILE` command is sent after another `FILE` without sending a binary message between them, the previous operation is cancelled and the associated event MUST NOT be published.
+5. If a client sends `FILE` command and then closes the connection without sending a  binary message, the operation is canceled and event MUST NOT be published.
+6. If the connection is closed during reading a binary message (continuation frames), the operation is canceled and event MUST NOT be published.
+7. If the received binary message is corrupted (hash or size doesn't match), the relay MUST NOT publish the event and signal this to the client using appropriate response.
+8. Relay is free to choose not to accept binary content for any reason. For example, Relay may test submitted images for explicit or copyrighted content and reject such content.
+
+### Possible implementation - clients
+
+ 1. calculates SHA256 hash of the binary content
+ 2. prepares the file header event, fills mandatory tags (hash, size, mime type, and flag `s:relay`)
+ 3. signs the file header event
+ 4. sends the command FILE with the event to the relay
+ 5. reads a response received from the relay
+ 6. if the response indicates rejection (status: false), stops the operation, handles error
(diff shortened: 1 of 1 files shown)


vitorpamplona:
Super nice! 

Since the `f` tag won't be useful in any other event type, it should not be an indexable, single-letter tag. First, because there is no need to index a tag that has the same value everywhere and second because we only got 26 letters to "index" things. On NIP-94 we have `url`, `magnet` URIs and, more recently, a `streaming` tag for m3u8 files. Those tags define how to download the data. This PR could use another name that represents the replay protocol proposed here. Maybe `nip97` could be a good tag name. 

What if we could make this backward compatible? If `f` is not a requirement, any user can now download their past NIP-94 content from the centralized website it's currently stored in and send that to the relay that also has the NIP-94 event. If the NIP-94 event has `x`, `m` and `size` and the hash matches the relay could store it. 

This would be a huge win to decentralize past NIP-94 content that is now completely stuck with centralized services. 

Then in the same way you have the `Retrieve the content by URL` section, you could have a simple tag "nip97":"true" to tell clients the content is available in the `RETRIEVE` method.

ondra-novak:
Issue with tag 'f' - in reference to my question in a discussion: [link](https://github.com/nostr-protocol/nips/discussions/708)

I wanted it to be indexed, for possibility to find all files hosted on the relay. The original idea was to assign it a different kind, then the flag wouldn't have to be there. The introduction of the indexed flag could of course be used elsewhere, but a new NIP would have to be created for it. Meaningfully, it could be an indexed valueless property. It is possible to include more 'f' tags for indexing under more such properties. In relation to my question, it is not currently possible to query for the existence of tags, and at the same time the proposed solution I was redirected to requires an even bigger change at the protocol level than just introducing such a tag

About single letter tags. In NIP-94 itself, I have a minor objection to the `x` tag - I don't see any use there at all, why files should be indexed according to their hash. Likewise the `m` tag, although it might already make sense to look for files of a certain type there

ondra-novak:
> Then in the same way you have the `Retrieve the content by URL` section, you could have a simple tag "nip97":"true" to tell clients the content is available in the `RETRIEVE` method as well as in the URL

Only the client can set tags and then signs them with his key. The client cannot know if the relay exposes the uploaded files to any url, nor to which url they will be exposed in the end. In the time before the event is sent, even the relay itself doesn't need to know this, if it has a repository in a distributed form (cloud)

The final URL must be set by the relay. The client also must assume, that URL is dynamic, so it is not wrong to give to different client a different url for the same file!

If the relay doesn't have a http interface (other than one-purposed for ability to accept a websocket connection), the value "file_url" wouldn't be there to indicate, that the file is not available through the http protocol.

There is also question in the forum: [link](https://github.com/nostr-protocol/nips/discussions/712). 

It looks like it's ok to put a new item in the event where the relay puts the url where the file can be retrieved. I chose "file_url" for this purpose. The client does not verify this new item for signature validation, nor does it change the post ID

It would be nice to have it designed so that there is no difference between the url for NIP-94 and the "file_url" but unfortunately it is not possible with this arrangement. Anyway, if it is still possible to insert the url tag into the event, if the file is available on a centralized network, upload it to the relay at the same time. Such an event will have both `url` and `file_url` tags.

The main purpose of being able to download a file via the http interface is to relieve the websocket connection when retrieving binary content, since the http interface is more efficient for this purpose, for example, files can be downloaded in parallel, which can be advantageous when downloading timeline posts. The files can then be cached on a local CDN.

Of course, HTTP access will simplify the design of web clients. However, I do not recommend placing the link directly in the HTML code, the client should verify that the link leads to the file he requested, i.e. he should, for example, calculate and compare the hash.

vitorpamplona:
> Only the client can set tags and then signs them with his key.

Sure, but that's not the idea. The idea is to use a new field, outside the event, set by the relay, and thus not participating in the signature verification hash. It's fully controlled by the relay and dynamic in nature. If the relay has the content, it sets it to true, if not, it doesn't even need to exist. If the client sees the flag, it can then opt for the HTTP interface OR the Websockets interface. 

Something like this: 

```json
{
   "id":"....",
   "kind": 1063,
   "pubkey":"....",
   ...
   "file_url":"https://cdn.myrelay.example.com/file/abc123acfa7e1256a.jpg", // Can download via HTTP
   "nip97": true // Can download via RETRIEVE. 
}
```

Neither the `file_url` nor the `nip97` participate in the hash to verify the event. 

> About single letter tags. In NIP-94 itself, I have a minor objection to the x tag - I don't see any use there at all, why files should be indexed according to their hash. Likewise the m tag, although it might already make sense to look for files of a certain type there

I agree, those didn't need to be single letters either. 

> I wanted it to be indexed, for the possibility to find all files hosted on the relay. 

I am not sure this is a good idea. The majority of the file servers don't expose a directory of all files on purpose to avoid constant crawling by bots out there.

ondra-novak:
OK, I understand that, so it's just a matter of the relay itself adding the information that the event also contains a binary content, so the client doesn't have to indicate this in advance with any flag. It's a shame that it makes it somewhat impossible to do the opposite check, i.e. if the client publishes this event without binary content (command EVENT), the relay that supports this NIP will not stop him. But this is not big deal.

vitorpamplona:
> if the client publishes this event without binary content

I actually think there is a use case for that. Not all relays need to have all the content. In fact, it seems more viable that binary relays would be separate from regular event relays. The Client just needs to figure out where the content is in the same way Clients figure out where events are.

We could even have a `nip97`-specific relay type on NIP-65's kind 10002. Each user would point to their "home" content provider and clients would use that information to know where to call RETRIEVE on somebody that user's content.

ondra-novak:
> We could even have a `nip97`-specific relay type on NIP-65's kind 10002. Each user would point to their "home" content provider and clients would use that information to know where to call RETRIEVE on that user's content.

I can't decide that. 

In general, I thought that the client would honestly check supported_nips and if publishing to a relay that supports NIP-97, use the FILE command, and if publishing to another relay, then use the old way.

It will be interesting if someone has a hybrid list of relays to publish to. Then he will probably have to stick with the old way of publishing.

I hope the transition period won't last long

Any idea how I could insert an icon into my profile via NIP-97. Use note: link? Is there a link to the event that would also include a relay (as I do not recommend using that url). I only know nprofile: but this is only for users

vitorpamplona:
> In general, I thought that the client would honestly check supported_nips and if publishing to a relay that supports NIP-97, use the FILE command, and if publishing to another relay, then use the old way.

In the short term, we will likely ask the user if he/she wants to upload to a server OR to the relay, either in the Settings or a per upload toggle. Automating that decision just confuses everyone because most clients won't support this protocol out of the gate and some clients will never support it. 

> I hope the transition period won't last long

Don't assume people will only use your NIP. We almost never have full consensus in Nostr. So, we always have to deal with clients not implementing NIPs they don't want to. The best strategy is to offer good value when people decide to implement your NIP and just hope for the best. 

> Any idea how I could insert an icon into my profile via NIP-97.

It's not well-adopted for images yet, but you could use an `nevent1` to point to the NIP-94 record. Then when a client sees it, it can download the image from there.

ondra-novak:
I have updated the proposal. The current demo relay doesn't reflect changes yet (I plan to incorporate changes and deploy the new version today evening CEST)

staab:
I apologize for the drive-by comment, I don't have time to get into this issue, but why are we not still pursuing #358 for this? Consensus seemed much stronger over there. From a web developer's perspective, using HTTP for file retrieval is much better, since the browser does a lot of the heavy lifting for you (image tags, caching, CSP, etc). You're throwing all that away with websockets.

vitorpamplona:
> From a web developer's perspective, using HTTP for file retrieval is much better, since the browser does a lot of the heavy lifting for you (image tags, caching, CSP, etc).

This PR allows relays to offer both, an HTTP and a Websockets connection. Clients can choose which they prefer.

ondra-novak:
> I apologize for the drive-by comment, I don't have time to get into this issue, but why are we not still pursuing #547 for this? Consensus seemed much stronger over there. From a web developer's perspective, using HTTP for file retrieval is much better, since the browser does a lot of the heavy lifting for you (image tags, caching, CSP, etc). You're throwing all that away with websockets.

As I understand it, that proposal is not blocked in any way, it can be implemented at the same time as this proposal, because it is completely outside the NOSTR network. 

My idea was that posts with binary content would travel over the NOSTR network. This is not the same as making NOSTR-RELAY another binary repository. If the binary content is part of an event, it can be replicated, cached, forwarded, and any client is able to retrieve it, even if it was deleted from the original server. This is the foundation of decentralization and censorship resistance. Of course this topic would be for a longer discussion, but my take is that it's just an extension of the original idea of NOSTR to media.

In this proposal, the binary protocol is mainly used for the upload, where the user publishes the event with its binary content. Two methods are available for downloading the binary content: the HTTP interface and the websocket connection. The HTTP interface is an optional method (but I would hope that most relays would implent it) - for the simple, technical reason that because a full HTTP interface is not mandatory for the NOSTR network, while a websocket connection is always mandatory

ondra-novak:
> I see, missed that the first time. Why would downloading via websocket ever be preferred to http? It seems to me supporting both introduces a lot of complexity that would be better spent on helping relays monetize file storage as in #547. I do like the integration with NIP 94 here.

I prefer download over HTTP, it was always faster and allows to use CDNs to distribute content. Download over websocket is fallback option. 
I can imagine a relay without an HTTP interface.

I can also imagine a bridging service that forwards events from one side of the network to the other using a pure websocket connection, where forcing an http connection for certain content would add complexity.

ondra-novak:
demo relay and client updated.

pj8912:
My old idea for file storage(which I called nostrage[file storage on nostr]) is slighlty different where I have two types of file storage:
- Public 
- Private

Public is available for anyone to see and download, Private is encrypted on upload and needs key to decrypt which will be unique per file.  
The file encryption type depends on the client. Would be better if a client have a dashboard displaying the filename, type, relays used , encryption key as columns.

Similat to the p2p file storage systems we have except the file is not split into pieces(shards) and distributed across multiple nodes.

Would this idea work on this NIP since what I thought about is relay specific?

The relay might monitize on private file storage?

pj8912:
what is `["aes-256-gcm",<key>, <iv>],` in tags?

pj8912:
@vitorpamplona what do you think about https://github.com/nostr-protocol/nips/pull/719#issuecomment-1688012440 ?? Is it possible in this nip ?

ondra-novak:
I designed this proposal as a technical solution. However one can build on this. You can use, for example, a zap to an account that is set up to collect storage fees. 

Encryption is supposed to be done on the client side so that even the provider of the relay can't see the content.

vitorpamplona:
> Is it possible in this nip ?

I see no problem with implementing what you described through this NIP.

ondra-novak:
It is a bit difficult for relay to decide what is private and what is public content. Private content is probably encrypted, but how do you determine if something is encrypted. You can't trust the mime type, because it came from the user and may have been entered intentionally incorrect.

But you could build it on analyzing the binary content (like the `file` command does in Linux, for example) and if the analysis recognizes the file, it can assume it is not encrypted and therefore public

Translated with www.DeepL.com/Translator (free version)

pj8912:
The storage type should be explicitly specified on file upload. `file_storage_type : public / private` ,
if public retrieve URL to a binary content , 
if private :
 - to be viewed/downloaded display that the file is private and ask for the key. 
 
of course the client will check if the user has the keys stored for decryption with which the file will be displayed, else the error message requiring key.

This might work :crossed_fingers:

vitorpamplona:
> to be viewed/downloaded display that the file is private and ask for the key.

Why would the relay want the key? That is not supposed to happen. The client is supposed to download the encrypted content from the relay and then decrypt it with the key locally. In that way, the key never leaves the device.

pj8912:
ask for the key as in the key required for decryption, stored by the user. Like a pop-up. Not the relay , this is on the client-side. I'll write `key for decryption` next time :wink:  .

ondra-novak:
> The storage type should be explicitly specified on file upload. `file_storage_type : public / private` ,

You can't build relay monetization based on this "flag". It is set by the user. If you require a fee for private storage, nobody is going to set this flag.

PR #722, opened 2023-08-15
NIP-107: Wrapper Event

Ungigdu:
A wrapper event is a kind 1107 event that carries another vaild event in its content field.

### DIFF
--- a/107.md
+++ b/107.md
@@ -1,0 +1,20 @@
+NIP-107
+======
+
+Wrapper Event
+-------------------------
+
+`draft` `optional` `author:ungigdu`
+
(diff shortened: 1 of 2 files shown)


vitorpamplona:
Hum.. why not using the other GiftWrap? https://github.com/nostr-protocol/nips/pull/716 

Also, NIP04 encryption is likely to be discontinued soon.

staab:
Duplicate of #716. Looks like you have a specific use cas in mind, but it should be doable using the same scheme, and possibly without a nip.

Ungigdu:
Ok, thanks. Will check nip-44 & nip-59.
But I think there is no available lib to be imported in react-native project for nip-44.

staab:
Right, but it's very close, see my PR to nostr-tools here: https://github.com/nbd-wtf/nostr-tools/pull/278

PR #723, opened 2023-08-15
Nip 51 extended kind 30001

gzuuus:
In order to standardize the new kind `30303`used in the nostree client, the content of nip 51 is extended to include the new kind and its definition. It is not really a big change in the nip, as I have tried to follow the previous standards. I hope I have expressed it as well as possible and would appreciate any feedback. 
Cheers

### DIFF
--- a/51.md
+++ b/51.md
@@ -84,6 +84,43 @@ Then the user would create a 'Categorized People' list event like below:
 }
 ```
 
+Lets say a user wants to create a 'Categorized Bookmarks' list of `bookmarks` and has keys:
+```
+priv: fb505c65d4df950f5d28c9e4d285ee12ffaf315deef1fc24e3c7cd1e7e35f2b1
+pub: b1a5c93edcc8d586566fde53a20bdb50049a97b15483cb763854e57016e0fa3d
+```
+The user wants to publicly include these bookmarks:
+
+```json
+["e", "5c83da77af1dec6d7289834998ad7aafbd9e2191396d75ec3cc27f5a77226f36", "wss://nostr.example.com"],
+["a", "30023:f7234bd4c1394dda46d09f35bd384dd30cc552ad5541990f98844fb06676e9ca:abcd", "wss://nostr.example.com"],
+["r", "https://github.com/nostr-protocol/nostr", "Nostr repository"],
+```
+and privately include these bookmarks (below is the JSON that would be encrypted and placed in the event content):
+
+```json
+[
+    ["r", "https://my-private.bookmark", "My private bookmark"],
+    ["a", "30001:f7234bd4c1394dda46d09f35bd384dd30cc552ad5541990f98844fb06676e9ca:abcd", "wss://nostr.example.com"],
+]
+```
+
+Then the user would create a 'Categorized Bookmarks' list event like below:
+
+```json
+{
+  "kind": 30001,
+  "tags": [
+    ["d", "bookmarks"],
+    ["e", "5c83da77af1dec6d7289834998ad7aafbd9e2191396d75ec3cc27f5a77226f36", "wss://nostr.example.com"],
+    ["a", "30023:f7234bd4c1394dda46d09f35bd384dd30cc552ad5541990f98844fb06676e9ca:abcd", "wss://nostr.example.com"],
+    ["r", "https://github.com/nostr-protocol/nostr", "Nostr repository"],
+  ],
+  "content": "y3AyaLJfnmYr9x9Od9o4aYrmL9+Ynmsim5y2ONrU0urOTq+V81CyAthQ2mUOWE9xwGgrizhY7ILdQwWhy6FK0sA33GHtC0egUJw1zIdknPe7BZjznD570yk/8RXYgGyDKdexME+RMYykrnYFxq1+y/h00kmJg4u+Gpn+ZjmVhNYxl9b+TiBOAXG9UxnK/H0AmUqDpcldn6+j1/AiStwYZhD1UZ3jzDIk2qcCDy7MlGnYhSP+kNmG+2b0T/D1L0Z7?iv=PGJJfPE84gacAh7T0e6duQ==",
+  ...other fields
+}
+```
+
 ## List Event Kinds
 
 | kind   | list type               |
@@ -92,7 +129,7 @@ Then the user would create a 'Categorized People' list event like below:
 | 10001  | Pin                     |
 | 30000  | Categorized People      |
 | 30001  | Categorized Bookmarks   |
-| 30303  | Categorized Reference List   |
+
 
 ### Mute List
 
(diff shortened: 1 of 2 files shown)


51.md
```diff
@@ -108,3 +109,7 @@ An event with kind `30000` is defined as a parameterized replaceable list event
 ### Categorized Bookmarks List
 
 An event with kind `30001` is defined as a parameterized replaceable list event for categorizing bookmarks. The 'd' parameter for this event holds the category name of the list. Any standardized tag can be included in a Categorized Bookmarks List.
+
+### Categorized Reference List
+  
+An event with kind `30303` is defined as a parameterized replaceable list event for categorizing user references. The 'd' parameter of this event holds the category name of the list. The reference lists may have metadata tags such as 'title', 'image', 'summary' as defined in [NIP-23 - Long-form Content](23.md) . Items in these lists are `r` tags, and follow the schema of Kind 3 events as defined in [NIP-02 - Contact List and Petnames](02.md). Each `r` tag entry should contain the 'reference URL' and the 'description', `["r", <reference URL>, <description>]`. Any standard tag can be included in a reference list.
```
staab, line 115: "categorizing user references" sounds like `p` tags to me. But this is for link trees right? So it could be `e` tags, `p` tags, urls, etc right?

gzuuus: I used user references alluding to the 'r' tag, can you think of a better name? 'Categorized Links/URIs List'? And yes, you could use 'e', 'p', 'a' or other standardized tags to be included.

pablof7z: I think this should just be a `kind:30001` tbh, it is in fact a list of bookmarks, right? Perhaps you can make a 30001 `d`ed `"linktree"` or something like that to tag the `kind:30001`s you want to display within this use case, or just display in this client those `kind:30001`s that are composed solely of `r` tags.

I don't think this warrants a specific kind especially now that kinds are [more scarce](https://github.com/nostr-protocol/nips/blob/master/01.md).

pablof7z: I constantly add `title`, `description`, `image` to `kind:30001` -- don't let the fact that it's not explicitly named as such deter you from using/adding those.

I think the most cross-compatible, sensible way for this use case is to just create these as 30001s, and, in your client, only display those kind:30001s that have the data you want (e.g. `title`, `description`, `d` and the rest just `r` tags.

If you wanted a single query to find the users who have set this use case up you could `l` label the events with the name of the use case (e.g. linktree)

staab: `l`, maybe even with a pubkey you control is definitely a workable solution.

gzuuus: I mean, yes, it is definitely a workable solution, kind 30001 with `l` tags. However, I still think that if we consider kind 30001 as a kind for everything we can think of, it might not be so cross-compatible in the future, because its parsing or visualisation might be very varied. Of course, this is not a problem for specific uses, because everyone will know what to do with the content of their 30001s. But we already have nip78 for that, don't we? For example, currently listr, one of the most focused list creation and management clients, does not allow you to add anything that is not a valid nip19 identifier, and type 30001 does not have a standardised `["tag", <reference URL>, <description>]`' scheme. This was my reasoning behind kind 30303.

staab: The labels is a good point. I wish tags were more extensible, the positional argument thing is so brittle. But if we specified that you can always add a display value (similar to petnames) at the end of any of these target tags would that make 30001 work?

gzuuus: Yes, it would work. From my point of view there are two possibilities, either we rework the definition of kind 30001 to add this schema with petnames, the optional metadata fields, and the possibility to add more than just nip19 identifiers, or we add kind 30303 as a list of external references to nostr, more flexible in tags and content.
For me, adding kind 30303 has several advantages, such as a clearer and cleaner design, and defining more specific use cases for applications that reference external content. I'm also thinking about backwards compatibility and whether reworking kind 30001 might 'break' other clients that have been using it for a long time.
In any case, I will be happy with any solution. My purpose was to push forward the standardisation of a need that I have already seen in various clients that have created lists with random kinds (e.g. Pinstr with kind 33888, zap.stream with kind 37777 for maps), and for the purpose of my own client.
On the other hand, I don't understand the argument about the scarcity of kinds when parameterised replaceable events are 30000 <= n < 40000 and we have only standardised 19 of them, leaving 39980 possibilities to define more specific use cases.
I would like to hear more points of view here. @fiatjaf @verbiricha @sepehr-safari

fmaul: I am also in favour of a more generic list type like kind 30001. That would allow us to not only point to external urls but also to kind 1063 (NIP-94) events. I would be good to be able to mix different types of references in one list.

vitorpamplona: I would use a `30001` as well. But instead of centering the website on npubs (which forces you to deal with all the other lists), Nostree should display by naddr. The website should point to individual lists, not to all lists of an author. 

As an example: if I want to do an Amethyst list, I can do a `30001:mykey:Amethyst` and share that around. I can even get `amethyst.social` to redirect to that specific list. I am still the author, so my picture will still be there as it is today. It's just not going to show all my other lists in the home page. 

Then later we can find a way to "pin" lists for each application.

gzuuus: Well, I'll use kind30001 then, with 'r' tags and kind2 scheme for petnames. In a little while I will commit to add this to nip51. 
@vitorpamplona , yes, definitely, urls for individual lists is in the roadmap, I've slowed down the pace of development because I was waiting to clarify this kind issue better so I don't mess up the current user base too much. I will propose an easy way to migrate lists to kind30001.

gzuuus: Just pushed the commit updating kind 30001

51.md
```diff
@@ -107,4 +145,4 @@ An event with kind `30000` is defined as a parameterized replaceable list event
 
 ### Categorized Bookmarks List
 
-An event with kind `30001` is defined as a parameterized replaceable list event for categorizing bookmarks. The 'd' parameter for this event holds the category name of the list. Any standardized tag can be included in a Categorized Bookmarks List.
+An event of kind `30001` is defined as a parameterised replaceable list event for categorizing bookmarks. The `d' parameter for this event holds the category name of the list. The bookmark lists may contain metadata tags such as 'title', 'image', 'summary' as defined in [NIP-23 - Long-form Content](23.md). The tags included in these lists MUST follow the schema of event tags as defined in [NIP-01 - Basic protocol flow description](01.md). Any standardized tag can be included in a Categorized Bookmark List.
```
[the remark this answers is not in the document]
gzuuus: Typos 😅

gzuuus:
@fiatjaf Are you okay with these changes?

staab:
This looks ok to me. I would add an additional heading to separate the previous example from the one you added.

vitorpamplona:
Maybe specify that `r` tags now can have a petname somewhere. `Readme`'s `r` tag can be updated.

PR #737, opened 2023-08-24
NIP-315: User Statuses

jb55:
This NIP enables a way for users to share live statuses such as what music they are listening to, as well as what they are currently doing: work, play, out of office, etc.

[315.md](https://github.com/jb55/nips/blob/user-statuses/315.md)

### DIFF
--- a/315.md
+++ b/315.md
@@ -1,0 +1,60 @@
+
+NIP-315
+=======
+
+User Statuses
+--------------
+
+`draft` `optional` `author:jb55`
+
+## Abstract
+
+This NIP enables a way for users to share live statuses such as what music they are listening to, as well as what they are currently doing: work, play, out of office, etc.
+
+## Live Statuses
+
+A special event with `kind:30315` "User Status" is defined as an *optionally expiring* _parameterized replaceable event_, where the `d` tag represents the status type:
+
+For example:
+
+```js
+{
+  "kind": 30315,
+  "content": "Sign up for nostrasia!",
+  "tags": [
+    ["d", "general"],
+    ["r", "https://nostr.world"]
+  ],
+}
+
+{
+  "kind": 30315,
+  "content": "Intergalatic - Beastie Boys",
+  "tags": [
+    ["d", "music"],
+    ["r", "spotify:search:Intergalatic%20-%20Beastie%20Boys"],
+    ["expiration", "1692845589"]
+  ],
+}
+```
+
+Two common status types are defined: `general` and `music`. `general` represent general statuses: "Working", "Hiking", etc.
+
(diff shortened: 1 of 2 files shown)


jacany (approved):

staab (approved):

vitorpamplona (approved):

vitorpamplona:
> The status MAY include an r, p, e or a tag linking to a URL, profile, note, or parameterized replaceable event.

We might need to keep updating this NIP with any rendering expectations for these links if they aren't included as nostr:bech in the content itself. 

Otherwise, this looks good.

fiatjaf (approved):

315.md
```diff
@@ -0,0 +1,60 @@
...
+  "content": "Sign up for nostrasia!",
+  "tags": [
+    ["d", "general"],
+    ["r", "https://nostr.world"]
+  ],
+}
+
+{
+  "kind": 30315,
+  "content": "Intergalatic - Beastie Boys",
```
mattn, line 32: Does this allow for empty text, text containing line breaks, or long text?

jb55:
> > The status MAY include an r, p, e or a tag linking to a URL, profile, note, or parameterized replaceable event.
> 
> 
> 
> We might need to keep updating this NIP with any rendering expectations for these links (are they always links?) if they aren't included as nostr:bech in the content itself. 
> 
> 
> 
> Otherwise, this looks good. 

I only handle r links atm but figured to add the others for completeness.

315.md
jb55: Its allowed but would be cut off in Damus

jb55: Empty statuses are equivalent to clearing the status in Damus, could make that explicit

315.md
```diff
@@ -0,0 +1,60 @@
...
+  "tags": [
+    ["d", "music"],
+    ["r", "spotify:search:Intergalatic%20-%20Beastie%20Boys"],
+    ["expiration", "1692845589"]
+  ],
+}
+```
+
+Two common status types are defined: `general` and `music`. `general` represent general statuses: "Working", "Hiking", etc.
```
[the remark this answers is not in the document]
jb55: I originally had it this way but then i realized it wouldn't work because the expiry times are different.

0xtrr:
Implemented support for this in [nostr-tool](https://github.com/0xtrr/nostr-tool/pull/58)

0xtrr (approved):

fiatjaf:
584f653ff781fe940a190c61d57ba802e776199e

fiatjaf:
7dc1f02fa70a5b9da7b3867e3faa37f48571d5e1

PR #739, opened 2023-08-24
Private Event

arthurfranca:
Considering DMs and gift wraps, it bothers me the downgrade to DX because a client can't anymore filter by:
- kind (no "k" tag added to the gift wrap)
- author (random author)
- created_at window (implementations using random past created_at up to 1 week)
- id (real id is inside gift wrap)
- and maybe by receipient tag (alias "p" tag instead of the real one)

Here I propose trusting author/created_at/"k"or kind/"p" (and other one-letter tags) metadata to relays, introducing a new client-to-relay message.

The NIP suggests NIP-43 usage because NIP-42 slows down event fetching with awkward flow.

Read text [here](https://github.com/arthurfranca/nips/blob/private/102.md)

### DIFF
--- a/102.md
+++ b/102.md
@@ -68,10 +68,19 @@ Note that **this is not the user's private key**, but a disposable one.
 Similar to DMs, the disposable or the author private key can decrypt the
 referenced event content.
 
-Note that only the author of the event referenced in the config array
-can set the disposable private key.
+Note that only the **author** of the event referenced in the config array
+can set the disposable private key when publishing private events.
 
-### Example
+The author must have previously added the disposable private key,
+if applicable (if the referenced event is to be encrypted),
+to the referenced event (or events, in case of `u` indentifier)
+inside a NIP-44 encrypted `privkey` tag next to the encrypted `p` tags.
+
+The author will need it for the configuration array when publishing
+a private event in reply to their own previous private event.
+Other users won't need it when replying.
+
+### Configuration Array Example
 
 `["a", "<nip_51_list_address>", "<disposable_privkey>"]`
 
@@ -82,7 +91,13 @@ attach to these private events an extra `config` attribute namespaced
 in the `private` key holding the "configuration array"
 **without** the "disposable privkey" array item.
 
-This is important to build client interfaces that enable configuration array editing features
+Instead of the "disposable privkey" array item it must always add the author of
+the event referenced in the config (their pubkey). The author will use it to check if
+their referenced event holds a `privkey` tag (inside the encrypted `.content`)
+it may need to privately reply to their own previous private event.
+
+The extra field is important to build client interfaces that enable
+configuration array editing features
 and to allow replying in [Private Threads](#private-threads).
 
 Example of a returned event:
@@ -94,7 +109,7 @@ Example of a returned event:
   // ...other event fields
   sig: ...,
   private: {
-    config: ["a", "<nip_51_list_address>"]
+    config: ["a", "<nip_51_list_address>", "<nip_51_list_author_PUBKEY>"]
   }
 }
 ```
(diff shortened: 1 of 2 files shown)


fernandolguevara:
@arthurfranca what do u think if we rename it as `restricted access events` ?

staab:
In many cases (DMs and group chats), it's feasible for clients to download all relevant messages. In other cases (large groups), a better approach would be to invite a relay's own pubkey to the group so it can 1. decrypt and index messages, 2. respond to regular REQs, and 3. serve the encrypted version of the event. This would probably need its own NIP so relays can signal support.

Defaulting to full trust is less easily undone than starting with minimal trust and incrementally introducing trust as a group scales. It also leaves open the option to do this in other ways, for example with a DVM that does the decryption/indexing/authentication.

arthurfranca:
@fernandolguevara I borrowed the "private" concept from the [Simple Group Chat PR](https://github.com/nostr-protocol/nips/pull/566/files#diff-84afeca5360c8d045c91a675f7466a9d4e11cce2d7f1478aada050985a726b90R48).

> In many cases (DMs and group chats), it's feasible for clients to download all relevant messages.

@staab The trend seems to be using gift wraps **without** "k" tag for inner event kind. The downside is requiring thin clients to download all kinds of gift wrapped events it may not even support/care. Also the created_at time window filter must be atleast 1 week for fetching recent events because of created_at shuffling.

> In other cases (large groups), a better approach would be to invite a relay's own pubkey to the group so it can 1. decrypt and index messages [...]

Won't this not only entrust metatada to the relay but also content?

staab:
Speaking as a thin client developer working on large groups, so far so good. 

> Won't this not only entrust metatada to the relay but also content?

Yes, that's a good point. I realize I mis-read this NIP, I thought you were suggesting publishing the event "privately" instead of encrypting it. I see now that you're sending the encrypted event + metadata. That being the case I'm not 100% against this, but I would like to wait to see how people solve this in the wild before merging a standard. I think this problem can probably be solved outside the protocol.

arthurfranca:
> I see now that you're sending the encrypted event + metadata.

@staab Yes that's it. In fact, one can send encrypted or unencrypted events.

For example it could be an encrypted kind 14 DM event directly without gift-wrap/seal/rumor package. Or encrypted kind 1 to be read by close friends only, encrypting a copy for each one of the close friends using their pubkeys (no need to add tags: ["p", ...] so the client will fetch feed events as normal without #p filter, just try to decrypt to see if you are one of the close friends).

But it could also be used for "hiding" cleartext NIP-28 group chat messages with a member list.

arthurfranca:
> I am not against the idea, but I have not had the need to add these extra filter possibilities to GiftWraps yet.

@vitorpamplona adding some metadata to giftwraps, such as true created_at, real p tag, and a k tag, may help out clients not willing to store and sort (created_at) events client-side. They would need to fetch much less events.

Perhaps giftwraps/seal/rumor should be ditched entirely, using the real event like the kind 14 DM event. Though I know you and staab had already put lots of effort on giftwraps/seal/rumor.

staab:
I strongly believe gift wraps solve more problems than they introduce. There are lots of ways to make encrypted events indexable.

vitorpamplona:
> adding some metadata to giftwraps, such as true created_at, real p tag, and a k tag, may help out clients not willing to store and sort (created_at) events client-side

As I said, I do not see that need from any client yet. And, of course, that need comes with the huge tradeoff of much less private events when such information is being shared with relays. 

There might be a use case for it, but I just don't see it right now. I don't know why would anyone go to the trouble of gift wrapping things if they are going to disclose attributes of the event to the relay anyway. Maybe there is a super specific use case of a personal relay just for this but I don't see it yet.

Egge7:
> There might be a use case for it, but I just don't see it right now. I don't know why would anyone go to the trouble of gift wrapping things if they are going to disclose attributes of the event to the relay anyway. Maybe there is a super specific use case of a personal relay just for this but I don't see it yet. 

Push Notifications is one use case. "You received something from someone" is not that great tbh.

arthurfranca:
> I don't know why would anyone go to the trouble of gift wrapping things if they are going to disclose attributes of the event to the relay anyway.

Yeah the reality is this can be seen as a gift wrap alternative proposoal because both are methods to protect event metadata leakage.

But this one is not only for entrusting metadata to relays but mainly for setting access control lists (ACL). So the metadata protection part is more of a side-effect to ACLs.

staab:
> But this one is not only for entrusting metadata to relays but mainly for setting access control lists (ACL).

Yes, I think this makes it potentially complementary to wrapping, but it's not a replacement.

mikedilger:
I'm against this.  If this is how people want to roll, I'm back to suggesting that DMs should be done outside of nostr entirely.  Why would we unsecure what we worked so hard securing?   Either you hide metadata or you don't.  If you want to expose metadata, then don't use giftwraps.

Giftwraps give you an approximate time and a recipient. That is all that should be exposed, and that is all you really need.

arthurfranca:
@mikedilger thanks for the comment. It looks like giftwraps are already being embraced by some client authors at least for DMs.

vitorpamplona:
@mikedilger sent me a GiftWrap kind 1. So, at least 2 clients are also working on private feeds. :)

mikedilger:
My earlier [comment](https://github.com/nostr-protocol/nips/pull/739#issuecomment-1692378231) was in respect to using this for giftwrapped DMs.  This could still be used otherwise and I won't outright stand against such a thing.  But I suspect due to the costs involved few relays would implement it.

@vitorpamplona I never sent you a giftwrap! You can't prove it!  Also, I didn't appreciate kind 14 when I sent that.

arthurfranca:
I add that while #784 Unbounded Lists (made up of many events, but just one for each list member) paired with Private Events make it possible to publish just 1 private event for each message in the case of private feeds and private chat rooms.

In contrast, in an 1 million subscriber scenario, #716 Gift Wraps would have to publish one million events per each message.

arthurfranca:
Just commited changes to make replies also private.

arthurfranca:
Way too complex. Relay can't know beforehand by looking at the REQ filter if the requested events require AUTH and who should AUTH.

PR #745, opened 2023-08-27
Add a note regarding matching events in real-time using custom NIP-50 filters

viktorvsk:


fiatjaf:
I think this NIP is already overspecifying behavior. Everything is already optional, we don't have to tell people that they may not everything. The antispam comment above is also unnecessary. I think we should remove that and just say relays should send events matching the search in real-time if they can.

viktorvsk:
Yeah, my main intent is to indicate to client developers that they should expect that accuracy of events sent in real-time after EOSE could be lower. And if this will be pointed out at NIP level, then relay developers will provide better docs on their custom filters guarantees/expectations.

PR #756, opened 2023-08-30
Add NIP-22 created_at limits to "limitations"

snazzybytes:
I was talking to CloudFodder and had this idea. Just wanted to see if it is needed and rough draft it during exploration


jeremyd:
+1 on this, very useful to have this info available to clients

jeremyd:
these fields have been added to relay.tools nip11 service.

Issue #742, opened 2023-08-26
NIP-XX Decentralized Web Hosting on Nostr


studiokaiji:
# Decentralized Web Hosting on Nostr

By recording HTML, CSS, and JS on the Nostr relay, it becomes possible to create a decentralized web hosting solution that eliminates the need for centralized servers. Web servers or Nostr clients retrieve these recorded data, transform them into appropriate forms, and deliver them. 

## Reasons for Hosting on Nostr

- Tamper-resistant through public key-based integrity
- Fault tolerance through deployment on multiple relays
- Resistance to blocking due to the distribution of web servers or clients
- Faster retrieval speed compared to IPFS's DHT

## Proposed Approach

Each HTML, CSS, and JS file is assigned a `kind` for identification. 

- HTML: `kind: 5392`
- CSS: `kind: 5393`
- JS: `kind: 5394`

The "content" field contains the content of the file. However, internal links (`href`, `src`, etc.) referenced within should be replaced with event IDs. 

Example: `<link rel="stylesheet" href="066b7ca0b167f0adad5c6d619ab1177050423e3979e83b8dfa069992533bdcf5">`

### Implementation on Web Server or Client

Access events using `/e/{event_id}`. Since event IDs are specified for each internal link, opening an HTML file enables automatic retrieval of data from this endpoint.

**Implementation Example (Golang)**

```go
r.GET("/e/:idHex", func(ctx *gin.Context) {
	id := ctx.Param("idHex")

	// Fetch data from nostr pool
	ev := pool.QuerySingle(ctx, allRelays, nostr.Filter{
		Kinds: []int{consts.KindWebhostHTML, consts.KindWebhostCSS, consts.KindWebhostJS, consts.KindWebhostPicture},
		IDs:   []string{id},
	})

	if ev != nil {
		// Return data with content-type adapted to kind
		switch ev.Kind {
		case consts.KindWebhostHTML:
			ctx.Data(http.StatusOK, "text/html", []byte(ev.Content))
		case consts.KindWebhostCSS:
			ctx.Data(http.StatusOK, "text/css", []byte(ev.Content))
		case consts.KindWebhostJS:
			ctx.Data(http.StatusOK, "text/javascript", []byte(ev.Content))
		default:
			ctx.String(http.StatusNotFound, http.StatusText(http.StatusNotFound))
		}
	} else {
		ctx.String(http.StatusNotFound, http.StatusText(http.StatusNotFound))
	}

	return
})
```

[...]

PR #767, opened 2023-09-07
Clarify character case of m tag value

AsaiToshiya:
Clarifying case sensitivity is useful for filtering events.

I think this change does not negatively impact existing implementations using `m` tag.


AsaiToshiya:
FYI, `m` tag is being considered for use in #742.

PR #772, opened 2023-09-09
NIP-50 search filter should work with multiple values 

viktorvsk:
In order to use `search` filter in the same way as `kinds` and `authors` it needs to accept array of strings. Those values will be used as AND conditions so that we could say that `{"kind": 1, "search": ["without_tag:e", "followers_of:<pubkey>"]}` we want to find root notes of our followers for instance.

### DIFF
--- a/50.md
+++ b/50.md
@@ -18,10 +18,10 @@ A new `search` field is introduced for `REQ` messages from clients:
 ```json
 {
   ...
-  "search": <string>
+  "search": <a list of strings>
 }
 ```
-`search` field is a string describing a query in a human-readable form, i.e. "best nostr apps". 
+`search` field is a list of strings describing a query in a human-readable form, i.e. "best nostr apps". 
 Relays SHOULD interpret the query to the best of their ability and return events that match it. 
 Relays SHOULD perform matching against `content` event field, and MAY perform
 matching against other fields if that makes sense in the context of a specific kind. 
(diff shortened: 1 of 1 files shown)


arthurfranca:
The NIP-50 search extension feels like an after thought. This is a mess: `{ search: "extension:1 look at this: full text search query extension:2" }`

viktorvsk:
> 
> So @viktorvsk I guess it should be `{"kind": 1, "search": "without_tag:e followers_of:<pubkey>"}`

why would we limit searches to non-space strings? Are there any disadvantages to using an array?

arthurfranca:
I think they got inspired by githubs `is:open is:pr a free text example` so relay should collect and remove the extensions first. The rest is free text with spaces allowed

viktorvsk:
So there is no way to filter `{“search”:”fuzzy_search_tags:Cheshire Cat full_text_search:once upon a time”}` using this NIP at the moment?

fiatjaf:
Search engines have all kinds of quirks that can be supported using just free text queries, I think we're fine relying on `"whatever:something blabla OR blublu"`, no?

staab:
I personally have always found structured text queries to be extremely annoying because they are subject either to user-defined input (resulting in syntax errors or unexpected results) or complicated escaping. Structured data and user input should never be mixed. But maybe I'm wrong since that's contrary to established convention in full text search.

viktorvsk:
I want to implement some essential yet simple filters which are a must imo in order to access data efficiently. In my opinion they (most of them) should better be a part of NIP-1 but in order to not make it too complex I think of implementing them in context of NIP-50

Some examples are:

    - [ ] Events without specific tag
    - [ ] Events of friends (of friends) by pub key
    - [ ] Fuzzy/fulltext search by content
    - [ ] Events matching local/pet/nip5 name
    - [ ] Exclude events by author, kind etc

And I don't see any sense to implement them as a single string because I see a lot of problems and zero profit. So I have 3 options:

- customize nip1
- customize nip50
- create new nip

For now I tend use a slightly customized nip50 at least until I see an argument in favor of `search` arg to have a string value

viktorvsk:
Yeah I'm talking exactly about AND filters. If say I want to find "all root notes of my friends which have text like something like 'I love nostr and bitcoin'"

PR #773, opened 2023-09-10
NIP90 Job Result kinds

cj-chua:
Instead of using just `kind: 65001` for job result, assign a unique job-result kind to each job request. (discussed [here](https://github.com/nostr-protocol/nips/pull/682))

The proposal here uses `65001-66000` for job requests, and `66001-67000` for job results. This setup has low overhead when trying to figure out the job result kind of a corresponding request (simply add `1000` to the job request kind). 

Request `kind:65001` and result `kind:66001` is now vacant and should be assigned the next job type to reduce confusion. 

`66000` is vacant as well and reserved for future/special use

viewable [here](https://github.com/lnconsole/nips/blob/2x-kinds/90.md)

### DIFF
--- a/90.md
+++ b/90.md
@@ -16,11 +16,11 @@ This NIP reserves the range `65000-67000` for data vending machine use.
 | Kind | Description |
 | ---- | ----------- |
 | 65000 | Job feedback |
-| 66000 | Placeholder |
+| 66000 | **Placeholder* |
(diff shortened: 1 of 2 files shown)


fiatjaf (approved):

cj-chua:
No longer relevant with the latest commit d9400e1e7b08225aedb6c78b522c5ad07b5198b4

Issue #754, opened 2023-08-30
Let's get public chat (NIP-28) fixed


jiftechnify:
I think the current public chat specs (NIP-28) has some problems that should be fixed.

## Problem with Updating Channel Metadata 

Since `kind:41` (set channel metadata) events are declare to be treated as **non**-parameterized replaceable events (1 event per user), a user can't set metadata to multiple channels.

(In fact, major relay implementations treat `kind:41` as non-parameterized event: )

For example, if a user updates metadata of channel A then updates that of channel B, the `kind:41` event that updates channel A get replaced with the `kind:41` that updates channel B. As a result, metadata of channel A is rolled back.

This problem was discussed in #376, though the solution presented is somewhat complicated.

### Possible Solution
We may solve it by making "set channel metadata" events parameterized replaceable.

* Use `kind:30041` instead of `kind:41`. Deprecate the use of `kind:41`s
* `d` tag must be set to the id of the `kind:40` event you want to update metadata
  * Open question: should we also reference `kind:40` by `e` tag? 
* `.content` will be the same as `kind:41`s

### Concern about Backward Compatibility
If it was the first time to define specs for public chat, I would choose to use parameterized replaceable events to describe a chat channel (creation + update) (as NIP-53 or NIP-72). Unfortunately it's not the case. 

Since a lot of `kind:42` events refer `kind:40`s by `e` tag, migrating function of `kind:40` events to parameterized replaceable events will be breaking change. This is why I chose to migrate only `kind:41` events.

## Obscure meaning of `kind:44` events

`kind:44` (channel mute user) events are to contain only a `p` tag that refers the pubkey to be muted.
I think it's obscure what `kind:44` means (or, what should clients do when they see `kind:44` events). Maybe the user referred should be muted from entire public chat ecosystem? However, it can also be achieved by `kind:10000` (mute list).

### Proposal: muting per channel basis
Making `kind:44` events possible to have optional `e` tag that refers `kind:40` events may give them a useful meaning.

We can add following rule:
If a `kind:44` event have both `p` tag and `e` tag, the user referred by the `p` tag should be muted only in the channel defined by the event referred by the `e` tag.

arthurfranca:
Imo the kind 44 is clear: client should mute user inside chat. In practice, it should hide kind 42s from the referenced pubkey. No need to change.

[...]

PR #774, opened 2023-09-10
NIP-28: use parameterized replaceable event for channel metadata

jiftechnify:
This PR proposes to migrate public channel metadata events (kind: 41) to parameterized replaceable events, allowing a single user to update metadata of multiple channels more naturally, in a backward compatible way.

cf. "Problem with Updating Channel Metadata" section in #754

### DIFF
--- a/28.md
+++ b/28.md
@@ -7,23 +7,24 @@ Public Chat
 
 `draft` `optional` `author:ChristopherDavid` `author:fiatjaf` `author:jb55` `author:Cameri`
 
-This NIP defines new event kinds for public chat channels, channel messages, and basic client-side moderation.
+This NIP defines new event kinds for public chat channels, channel messages, and basic client-side moderation:
 
-It reserves five event kinds (40-44) for immediate use:
-
-- `40 - channel create`
-- `41 - channel metadata`
-- `42 - channel message`
-- `43 - hide message`
-- `44 - mute user`
+- `40`: Channel Creation
+- `30041`: Channel Metadata
+- `42`: Channel Message
+- `43`: Hide Nessage
+- `44`: Mute User
 
 Client-centric moderation gives client developers discretion over what types of content they want included in their apps, while imposing no additional requirements on relays.
 
-## Kind 40: Create channel
+
+## Kind 40: Channel Creation
 
 Create a public chat channel.
 
-In the channel creation `content` field, Client SHOULD include basic channel metadata (`name`, `about`, `picture` as specified in kind 41).
+The `id` of a kind 40 event is regarded as an identifier of the channel by other public chat related event kinds.
+
+In the channel creation `content` field, Client SHOULD include basic channel metadata (`name`, `about`, `picture` as specified in kind 30041).
 
 ```json
 {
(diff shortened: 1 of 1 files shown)


vitorpamplona:
If we use replaceable events, there is no need to separate channel creation event from updates. Kind 42 can simply a-tag a 30040 event directly.

jiftechnify:
> If we use replaceable events, there is no need to separate channel creation event from updates.

That's true, but how do you think about backward compatibility?  Many public channel messages already reference `kind:40` events by e-tag.

vitorpamplona:
We are already breaking it because unless implementers change, they won't see new metadata updates. So, if we are going to break it, break it all the way.

For me there would be kind 40 chats and kind 30040 chats. Both render together in the same UI.

SnowCait:
I prefer kind 30040 to kind 40 + 30041.
However, it is breaking change (and last chance to do).
It is better to create a migration tool for relays and make enough time for migration.
Needless to say, we can subscribe `{"kinds": [40, 41, 30040]}` in clients but it is a little complex because they have different tags.

arthurfranca:
@vitorpamplona Considering parameterized replaceable events (PREs) break the most recent/oldest event lists (oldest if using NIP-34asc) based on created_at, this PR can be better than using 30040.

**Maybe** future PRE use cases should copy this design.

vitorpamplona:
> @vitorpamplona Considering parameterized replaceable events (PREs) break the most recent/oldest event lists (oldest if using NIP-34asc) based on created_at, this PR can be better than using 30040.

Sorry, what's "break the most recent/oldest event lists"?  I don't think I have seen this happening. 

The 30040 proposal was similar to NIP-53 (Live Activities). That one is working very well.

arthurfranca:
I mean by what NIP-01 says, relays sort by most recent when using "limit". This could be reverted using NIP34asc ordering by oldest.

PREs break that because created_at is in fact update date.

vitorpamplona:
I don't understand what NIP-34 has to do with PREs. The order declared by NIP-34 is just for filtering and should NEVER change the order or processing for PREs. Remember, relays are not supposed to keep previous versions of PREs. So, there is no way to sort ASC in replaceable events. If there is, the relay is not compliant with those NIPs.

arthurfranca:
I mean this PR doesn't change kind 40. It is still a regular event. So a client can sort chat rooms by latest or oldest(nip34) server-side.

If we use 30040 as you want, it becomes a PRE, which breaks the previously mentioned behavior.

vitorpamplona:
> I mean this PR doesn't change kind 40. It is still a regular event. So a client can sort chat rooms by latest or oldest(nip34) server-side.

Oh, but the `created_at` in kind 40 is not used like that. It's kind of useless. We sort by the message's (kind 42) `created_at`. It's very similar to how DMs are sorted.

arthurfranca:
You sort by "recent messages".
But another client could offer to show "recently created (chat room)".

vitorpamplona:
> But another client could offer to show "recently created (chat room)".

In theory, yes. In practice, no. There is too much noise in a list of recently created rooms. Even if you take your follows recently created rooms, you will find it useless. 

I coded it. It didn't please anyone. 

There is a reason why you don't see Reddit's new communities. It's all spam, scams, tests or channels that will die before they even get a single message.

arthurfranca:
It also bothers me on PREs that the `published_at` tag (original creation date), if used, can be changed at will although it should be immutable. It isn't a problem if the idea is to reference PREs just by `e` instead of by `a` tag such as when commenting on an article (so the `created_at` can be used instead and we start treating new PRE versions as a delete followed by a create, losing all previous likes and replies).

So by using PRE, a chat room, if adding a `published_at` tag,  can be said to be created 3 days ago, then change to 1 month ago then change to now then change to a future date. Nostr becomes a clown protocol where things can be easily faked.

When Moderated Communities were proposed I wasn't yet thinking about this detail or else I would say something and try to change things to a setup similar to this PR (a regular event with an auxiliary PRE holding metadata). Although probably it wouldn't have any effect and it would've been merged as is hahah.

arthurfranca:
heheh I know timestamp is whatever the user wants. Just pointing out that when using a non-replaceable event whose id will be referenced by all chat messages (`e`tag), the chosen `created_at` can't be altered anymore. In contrast, if using a replaceable one and referencing it by `a` tag, an user can edit the replaceable event's `published_at` tag at will.

PR #777, opened 2023-09-12
Add preview and caption tags to nip94

arthurfranca:
Updated NIP [here](https://github.com/arthurfranca/nips/blob/94-tags/94.md)

`preview` for thumbnails or preview text
`caption` for a loose description (as nip94 `content` is currently being used by Amethyst as alt text for accessibility)

The `caption` makes it possible to use nip94 in a media gallery without using another kind (using a `t` tag to differentiate the event from a non-media-gallery one).

### DIFF
--- a/94.md
+++ b/94.md
@@ -15,7 +15,8 @@ This NIP specifies the use of the `1063` event type, having in `content` a descr
 * `url` the url to download the file
 * `m` a string indicating the data type of the file. The [MIME types](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types) format must be used, and they should be lowercase.
 * `"aes-256-gcm"` (optional)  key and nonce for AES-GCM encryption with tagSize always 128bits
-* `x` containing the SHA-256 hexencoded string of the file.
+* `x` containing the SHA-256 hexencoded string of the **original** file before any server transformations
+* `xx` containing the SHA-256 hexencoded string of the **transformed** file after any server transformations
 * `size` (optional) size of file in bytes
 * `dim` (optional) size of file in pixels in the form `<width>x<height>`
 * `magnet` (optional) URI to magnet file
@@ -37,6 +38,7 @@ This NIP specifies the use of the `1063` event type, having in `content` a descr
     ["aes-256-gcm",<key>, <iv>],
     ["m", <MIME type>],
     ["x",<Hash SHA-256>],
+    ["xx",<Hash SHA-256>],
     ["size", <size of file in bytes>],
     ["dim", <size of file in pixels>],
     ["magnet",<magnet URI> ],
--- a/94.md
+++ b/94.md
@@ -16,7 +16,7 @@ This NIP specifies the use of the `1063` event type, having in `content` a descr
 * `m` a string indicating the data type of the file. The [MIME types](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types) format must be used, and they should be lowercase.
 * `"aes-256-gcm"` (optional)  key and nonce for AES-GCM encryption with tagSize always 128bits
 * `x` containing the SHA-256 hexencoded string of the **transformed** file after any server transformations
-* `xx` containing the SHA-256 hexencoded string of the **original** file before any server transformations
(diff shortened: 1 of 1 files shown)


vitorpamplona:
Maybe use the `thumb` tag instead of the preview? I don't like creating new tag names for similar behaviour. 

I am ok on changing Amethyst if we want to invert `caption` and `alt`. `alt` didn't exist when nip-94 was created. So, since it's now merged, we could follow the same conversion here. It breaks the past but it's early so no biggie.

arthurfranca:
> Maybe use the thumb tag instead of the preview? I don't like creating new tag names for similar behaviour.

Maybe we should split `preview` into two tags: `preview` and `thumb`?

Paywalls (e.g: a nip94 file - url tag - that can only be downloaded with nip98 auth, signaled by some tag presence) may want to use nip94 to show, for example, a preview text of a blog post (e.g.: .txt or .json being `m` as "text/plain" or "nostr/30023") or a censored NSFW image preview. So sometimes it may be a text, not a thumbnail.

vitorpamplona:
Ohh.. yes, I would separate the two. 

Maybe `thumb` and `summary`? There is also a `description` we could use. 

We do those previews for Long-form content and Badges. We could re-use it for NIP-94

arthurfranca:
What if the "Only Fans"-like creator wants to show a full dimension censored version of the paywalled picture just as a preview for people that haven't payed? It wouldn't be a thumb because it has full dimension. Summary would be for text only. It would probably be overkill to link to another nip94 event too.

So maybe keep the preview tag (with the mime tag[2] item) instead of summary? Or use preview without tag[2] and also summary for textual preview?

vitorpamplona:
Maybe we can use the `image` tag from Long form content? That's not supposed to be a `thumb`. So, we could've both an `image` and a `thumb`

vitorpamplona:
And FYI, thumbs load much faster on mobile than a regularly sized image. If the idea is to support full resolutions in the preview image, then making thumbs could be really helpful.

vitorpamplona (approved):
Looks great.

arthurfranca:
Added "xx" tag to differentiate it from the "x" one.

The "x" is for the original file hash. "xx" is for the transformed file hash (after the file went through server transformations such as compression and metadata stripping).

On NIP-96 the file will be searched by the "x" tag value. "xx" is just to tell if the file content was altered after event publishing.

fiatjaf (approved):

vitorpamplona:
Migrated from .content to alt tags: https://github.com/vitorpamplona/amethyst/commit/909d3f4fbe233aac8b587732f2bf3bd0fe3b1da9

vitorpamplona:
Is anybody using `xx`?

arthurfranca:
If not using, they should. `x` and `xx` are equal only when there is no server transformation on upload (compression, resizing or metadata stripping).

The `x` tag will hold the original file hash. So if I upload the file to two servers that transform the file in two different ways, both servers will return the same `x` tag (so I can search the file by `x` hash at any server) but two different `xx` tags (so that I can check if the file has been altered before showing it on client)

More explanation on NIP-96 comments:

https://github.com/nostr-protocol/nips/blob/eeed28ea43e77cd3eec3b4cf4f01d3c56cd04736/96.md?plain=1#L159-L174

vitorpamplona:
I just think most NIP94 implementers hash AFTER the server has returned the image URL. 

They download the image first to show the preview get approval from the user in the New Post and use that download to hash it and create the nip-94 event. 

Basically, everyone is doing xx as x now.

arthurfranca:
So maybe we should revert `x` and `xx`? I guess there is no problem as the NIP-96 search doesn't use relays but servers, so not much need for the tag to be indexable.

NIP-96 at first said servers shouldn't transform files (just clients should do it client-side) to keep `x` the same no matter to how many servers you uploaded the same file, but server owners didn't like it (they want to save space etc). That's why for NIP-96, the original file hash is important and the transformed file hash isn't, because the original one won't change.

If NIP-96 is the way to go, the original file hash is important for searching. If it is NIP-95/97, the transformed file hash is used for searching. I will revert it.

arthurfranca:
Also renamed xx to ox (original hash)

vitorpamplona:
What if we just leave x as is for now and create the xx or the ox when 96 gets implemented? We shouldn't complicate things before knowing that nip96 is going to be the way to go.

PR #781, opened 2023-09-13
NIP-77 - Trust Clock

arthurfranca:
[Read here](https://github.com/arthurfranca/nips/blob/trust-clock/77.md)

I was just reading @melvincarvalho's thoughts at telegram group on verifiable timestamps based on relay infrastructure
and then started writing this NIP and put him as author without talking to him.

It would be an alternative to NIP-03 (OpenTimestamps Attestations for Events) without using blockchain.

Can you tell me if it makes sense or is it just impossible gibberish?

### DIFF
--- a/77.md
+++ b/77.md
@@ -28,15 +28,15 @@ relay hints in a month window.
 
 The client trusted relay list at the time in seconds,
 informed on the `at` query param, can be checked at
-"https://client.example/.well-known/nostr/nip77.json?get=relays&at=\<time-in-seconds\>":
+"https://client.example/.well-known/nostr/nip77.json?get=relays&at=<time-in-seconds\>":
 `{ relays: ["https://relay1.example", "https://relay2.example", "https://relay3.example"] }`.
 
 Clients can delegate trust by responding to
-"https://client.example/.well-known/nostr/nip77.json?get=relays&at=\<time-in-seconds\>"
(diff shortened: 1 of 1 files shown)


vitorpamplona:
Can't an attacker just collect all clocks+signatures at every minute for future use? Then 10 days from now, all I need to do is to create a new event dated 10 days in the past and use the clocks and signatures those relays sent me at that time.

arthurfranca:
Yes. The proof just tells the event was created at or after the clock time, similar to NIP-03. But you can't create future events till the clock reaches the future moment.

Relays can atleast block events sent too far in the past using NIP-22.

arthurfranca:
Don't know but maybe if NIP-03 tells an event was created prior to or at a date.
And if NIP-77 tells it was created after or at a date.

It seems using both tells the exact date.

arthurfranca:
@vitorpamplona I changed how it works. The attack you mentioned won't work anymore.

Now relays sign the `event.id` if the `event.created` is near the relay clock time. And there is a `clock` field instead of `clock` tags.

ariard:
> Can you tell me if it makes sense or is it just impossible gibberish?

If I might suggest to look on https://www.rfc-editor.org/rfc/rfc5905 and all corresponding issues. Normally, you can query NTP server status with ntpq and reconfigure your NTP client accordingly to suit your application requirements. I don’t know if we nostr wanna to re-implement NTP.

PR #783, opened 2023-09-17
NIP-100, Using WebTransport Instead of WebSocket

kehiy:


Egge7:
If this was to be implanted it should be an option, not a requirement. Also there would need to be some signaling of support.

In order to keep backwards compatibility, this could be an upgrade after the initial WS connection has been established. I don't see how this could replace WS.

kehiy:
@mattn @fiatjaf can you have a review?
should I change it to optional?

alexgleason:
Interesting. If WebTransport is better than WebSockets, we should just remove WebSockets entirely and switch to WebTransport. Having both is protocol bloat. However, my understanding is that WebTransport is still experimental and not widely used.

Can you show more examples of how to connect to a WebTransport relay? Is there any implementation that exists yet?

kehiy:
@alexgleason 
from the last news HTTP/3, QUIC, and web transport going to be supported in most websites, frameworks, and web servers...
and it can tell us they going to be widely used soon.
(in my opinion, we should go with the latest technology)

About examples and implementations
the client side is fully supported by js.

and in the server, one of the implementations is:
[go](https://github.com/quic-go/webtransport-go)
[node.js](https://socket.io/get-started/webtransport)
[rust](https://github.com/BiagioFesta/wtransport)

fiatjaf:
This is a breaking change that will only introduce protocol bloat and kill Nostr. It does seem better than WebSockets, but would have to be at least 10x better and WebSockets would have to have a fatal flaw before we considered removing WebSockets and adopting this. Supporting both is a terrible idea.

Also, this thing seems to be a draft, super hard to implement, there are no libraries, they will probably change the protocol a million times because Google will decide to change, so it is really not a good idea. Maybe consider it again in 10 years.

PR #784, opened 2023-09-17
NIP-61 - Event Sets

arthurfranca:
[Read here](https://github.com/arthurfranca/nips/blob/bunch-of-events/61.md)

Like NIP-51 lists but without list size limit.


arthurfranca:
Need better name suggestions

staab:
I like the idea of having something like this. It would be nice to have it more explicit that you might have public and encrypted data coexist in a list. I'm not sure why we need `n` and `u` tags in addition to `d` though. Didn't you start by defining a new range? It seems to me we could just pick a regular event kind (<10000), use `d` as the name, and `a` instead of `u` to refer to all events in the list.

arthurfranca:
> It would be nice to have it more explicit that you might have public and encrypted data coexist in a list.

Added "Public and encrypted data can coexist in unbounded list events similar to NIP-51 lists." Thanks

> Didn't you start by defining a new range? It seems to me we could just pick a regular event kind (<10000), use d as the name, and a instead of u to refer to all events in the list.

I realized we didn't need a new range; just reserving the `n` tag to mark an event as being part of a list would be enough. Adding/removing `n` tags from an event is needed, so most times the event needs to be replaceable.

61.md
```diff
@@ -0,0 +1,76 @@
...
+  "kind": 30382, // This is a kind for public relationship info between the signer and another user
+  "pubkey": "<me>",
+  "tags": [
+    ["d", "<another-user-pubkey>"],
+    ["n", "follow"], // A "follow" "n" tag turns this event into a kind 3 equivalent entry
+    ["n", "friend"], // He is part of may "Friends" list
+    ["n", "close-friend"], // He is also part of my smaller "Close Friends" list
+    ["n", "contact"], // My contacts are users to whom I've sent DMs or started an one-to-one audio/video call
+    // ...other metadata describing the relationship
+    ["T", "3:driver"] // This is using an unreleased spec to rate this user as a great driver
```
vitorpamplona, line 63: Remove this `T` tag. Leave it for later (another NIP)

61.md
```diff
@@ -0,0 +1,76 @@
...
+  "content": ""
+  // ...other fields
+}
+```
+
+In the above example, when removing someone from an unbound list (considering a `kind:30382` event),
+a client should only delete the event when all tags other
+than the `d` one were removed and the content is an empty string.
+
+Above example's `T` tag comes from the [Trust Rank](https://github.com/arthurfranca/nips/blob/trust-rank/64.md)
```
vitorpamplona, line 74: Remove `T` stuff

61.md
```diff
@@ -0,0 +1,76 @@
+NIP-61
+======
+
+Unbound List
```
vitorpamplona, line 4: Since the list order is not defined, should this be renamed to **Event Sets**?

I understand why `Unbound` is there, but branding lessons tell me that we shouldn't name things based on "not being something else" :)

61.md
```diff
@@ -0,0 +1,76 @@
...
+
+To remove an event from a "Close Friends" list,
+remove the "close-friend" `n` tag from the event.
+
+An unbound list itself is referenced by an `u` tag
+or by an `nlist` [NIP-19](19.md) entity using [NIP-21](21.md) URI.
+
+## Use Cases
+
+Unbound lists can be used wherever a [NIP-51](51.md) list fits.
```
vitorpamplona, line 23:
```suggestion
```

61.md
```diff
@@ -0,0 +1,76 @@
...
+
+NIP-51 lists have a limit on the number of items they can hold depending on relay's event size restrictions.
```
vitorpamplona, line 25:
```suggestion
```

61.md
```diff
@@ -0,0 +1,76 @@
...
+While unbound lists can have any number of items.
```
vitorpamplona, line 26:
```suggestion
```

61.md
```diff
@@ -0,0 +1,76 @@
...
+
+The downside of unbound lists is that migrating a list from one relay to another can be rate limited
```
vitorpamplona, line 28: Move this to a final considerations kind of Section.

61.md
vitorpamplona, line 23: There is no need to compare with NIP-51

61.md
```diff
@@ -0,0 +1,76 @@
...
+due to saving many individual events during a small time window.
+
+Public and encrypted data can coexist in unbound list events, similar to NIP-51 lists.
```
vitorpamplona, line 31: Either remove this or add a section in the end to demonstrate how encryption would work.

61.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Referencing
+
+An unbound list is referenced by an `u` tag.
+It references a set of events from an author instead of a single one:
+
+`["u", "<32-bytes lowercase hex of a pubkey>:<n tag value>", "<recommended relay URL, optional>"]`
```
vitorpamplona, line 38: I would keep the kind there `kind:pubkey:n` but allow kind to be empty for lists from multiple event kinds (this seems like a rare case).

61.md
```diff
@@ -0,0 +1,76 @@
...
+
+One may also use an `nlist` entity URI: `nostr:nlist1qqstn...794234d`.
+
+## Event Kinds
+
+The same `n` tag can be applied to any event no matter the `kind` or structure.
+However, it is better to add `n` tag(s) to parameterized replaceable events, because
```
vitorpamplona, line 45: Regular events can be deleted. So, I don't see much difference between replaceables and not.

61.md
```diff
@@ -0,0 +1,76 @@
...
+`draft` `optional`
+
+An unbound list is a group of events with the same pubkey and `n` (list "n"ame) tag.
+An event can have many `n` tags, thus being member of many unbound lists.
+
+To add an event to the "Close Friends" list,
+add a "close-friend" (prefer lowercase English words in singular form with dash separator) `n` tag to the event.
+
+To remove an event from a "Close Friends" list,
+remove the "close-friend" `n` tag from the event.
```
vitorpamplona, line 16:
```suggestion
remove the "close-friend" `n` tag from the event or send a deletion event (`kind:5`).
```

vitorpamplona (commented):
I think this is getting ready to merge. So, full review.

arthurfranca:
All "set"!

fiatjaf:
I have read this today for the first time (sorry) and I think it's a very good idea.

Are there any downsides?

If we're switching from having lists be identified by event kinds to sets that are identified by strings (`"friend"`, `"follow"` etc) then I think this NIP should include a growing table of standard set names that can be reused across clients.

61.md
```diff
@@ -0,0 +1,64 @@
...
+
+The following example is based on the unreleased
+[Relationship Status](https://github.com/vitorpamplona/nips/blob/relationship-status/81.md) spec:
+
+```js
+{
+  "kind": 30382, // This is a kind for public relationship info between the signer and another user
+  "pubkey": "<me>",
+  "tags": [
+    ["d", "<another-user-pubkey>"],
```
alexgleason, line 42: What possible values can go in the `d` tag? That's what makes this whole idea valuable, right? I feel this NIP is under-spec'd without it.

vitorpamplona: You can see more on #761. This PR is just for the idea of `n` tags to represent sets of events. It uses the status as an example, but it doesn't define it. Overthere, we predicted either the d=pubkey of the "friend" OR a private SHA256 of the key for use cases where the relationship itself is private.

vitorpamplona:
The "user role" event is closer to the `kind:24` on @staab's #875

61.md
alexgleason: Are the ideas really separable? I'm having trouble picturing what other kinds of event sets would look like, to justify the `d` tag being free-form.

vitorpamplona: I can use `n` tags to put spreadsheets into lists (or directories) :)

vitorpamplona: Or group my Classifieds into collections.

arthurfranca:
> If we're switching from having lists be identified by event kind numbers (as in NIP-51) to sets [...] this NIP should include a growing table of standard set names

We have to think about how to migrate NIP-51 lists to event sets because it is not clear yet which event kind(s) and structure to use.
Should we bend "relationship status" event to become a standard for other sets?

Should we keep using NIP-51 for small lists such as "Emoji sets" and "Search Relays"?

How NIP-51 follow sets (not kind:3) are supposed to be distinguished from the regular NIP-61 "follow" set? They shouldn't all use `n-tag=follow`, although they are from the same "category" or should they? I mean, how will clients know how to fetch all follow sets? Maybe a reserved kind range so all sets use `n-tag=follow` and the first kind from the range is the "default" follow set?

Need ideas.

61.md
```diff
@@ -0,0 +1,64 @@
...
+
+An event set itself is referenced by an `s` tag
+or by an `nset` [NIP-19](19.md) entity using [NIP-21](21.md) URI.
+
+## Referencing
+
+An event set is referenced by an `s` tag.
+It references a set of events from an author instead of a single one:
+
+`["s", "<kind, optional>:<32-bytes lowercase hex of a pubkey>:<n tag value>", "<recommended relay URL, optional>"]`
```
fiatjaf, line 26: I think we should use some other format here, just to be different from normal replaceable event references.

```suggestion
`["s", "<n-tag value>#<32-bytes lowercase hex of a pubkey>", "<recommended relay URL, optional>"]`
```

And why is the kind optional? It shouldn't be optional, that's bad idea. We should either always omit it or always include it.

Maybe we come up with a reference format for including multiple kinds here if that's a requirement.

vitorpamplona:
> how to migrate NIP-51 lists to event sets

I don't see the need to migrate. Lists and Sets solve different problems. Sets are just another tool that people can use to group events in Nostr. It works for large groups while lists work best for smaller groups.

arthurfranca:
> I don't see the need to migrate. Lists and Sets solve different problems. Sets are just another tool that people can use to group events in Nostr. **It works for large groups while lists work best for smaller groups**.

If sets are better for large groups, aren't we migrating those NIP-51 lists that can grow too much like mute, follow sets (and kind:3), bookmark and bookmark sets?

staab:
I like this PR in theory, but the migration is the difficult thing, plus forcing everyone to support lists and sets violates the "multiple ways to do something" principle. With history as a guide, this will never happen. However you're right that small lists are better served with lists, so maybe this PR could be articulated as an extension to lists, with all the same semantic kinds (incremented by 1000 or something?), for use with lists > 1000 items?

alexgleason:
BTW, while we're talking theory, check my advanced Nostr philosophy idea about _event sets_: https://nostrify.dev/store/memory#event-sets 

We are using the phrase to mean two different things. Sort of.

fiatjaf:
One disadvantage is having to verify multiple event signatures instead of just one.

If you're fetching contact lists from many people and each of these people follow hundreds then amount of events downloaded may go very large, no?

61.md
arthurfranca: The n-tag should be last cause it can contain any char, including the ones used as separator.
Empty kinds may be useful when event set may contain events from too many kinds or even from kinds that were not speced?
Not sure if it is good to introduce another separator considering that the tag name dictates the value format.
Added comma-separated kinds.

arthurfranca:
@vitorpamplona hat to use `kind:30382` template with different `d` tag values (instead of just pubkey) to be able to accommodate NIP-51 lists. Also had to change the event name... looks like #761 should be merged with this PR?

arthurfranca:
> @fiatjaf One disadvantage is having to download the bloated part of the event JSON and verify multiple event signatures instead of just one.

Yeah there are some downsides. 

NIP-51 has some too like:
- the list size limit due to relay event size restrictions. 
- some people complain about `kind:3` getting wiped. I guess with NIP-61 follow sets people would lose single follows instead of whole list
- also the slow down of indexing many one-letter tags at once when updating a list
- can't download or update just one item at a time

Though can't say for sure if NIP-61 wins the competition.

vitorpamplona:
Lol, this just got ultra-complicated. Do we really need to define all these rules?

arthurfranca:
@vitorpamplona Not much changed if you look closely.
- `kind:30061` multiple events are the equivalent of your single `kind:10008` event.
- Your `kind:30382` `d` tag is set to `p:<pubkey>` instead of just `<pubkey>`. Then we can link to things other than pubkeys.

`kind:30061` is needed to have things like NIP-51 follow sets (those lists similar to kind:3 but with user-generated label - see the "Best Friends Forever" example).

Clients adding a backtick to user generated set names when saving is a rule to avoid collision with standard list names.

61.md
```diff
@@ -0,0 +1,145 @@
...
+```
+
+**2) Add a `n` tag to a "Set Item Reference" `kind:30382` event that references the item to be added:**
+Do this if the set may include events you don't own or other things that aren't events.
+
+The `kind:30382` event may have tags other than `n` with extra metadata about the referenced item.
+Because of that, when removing some metadata during an edit, clients should only delete the event when all tags other
+than the `d` one were removed and the content is an empty string.
+
+The `d` tag is set to the item reference with the format `<tag-referrer-name>:<tag-value>`. The notable exception
```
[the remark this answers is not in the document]
arthurfranca: Great insight! Your idea will enable removing follow-hashtag, mute-event etc.

About user-generated/custom sets, they need `kind:30061` (an alternative to same #761's kind:10008 concept, but individual events instead of a list) to keep track of their names (client can then present the custom names and the standard ones on a "dropdown" when adding something to a set). There's more to `kind:30061` need: marking a custom set name as "n=follow" will let client know the list may be used as custom feed.

61.md
```diff
@@ -0,0 +1,145 @@
...
+
+### User-generated
+
+Set names must be prefixed with a "\`" (backtick) when on `n` tags to avoid name clashing.
+To keep track of these custom names, a "Custom Set" `kind:30061` event is used for each of them.
+The `d` tag is the name. Example:
+
+```js
+{
+  kind: 30061,
```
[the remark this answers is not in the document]
arthurfranca: Answered on conversation below but will paste it here:

> About user-generated/custom sets, they need kind:30061 (an alternative to same https://github.com/nostr-protocol/nips/pull/761's kind:10008 concept, but individual events instead of a list) to **keep track of their names** (client can then present the custom names and the standard ones on a "dropdown" when adding something to a set). **There's more to kind:30061 need:** marking a custom set name as "n=follow" will let client know the list may be used as custom feed.

61.md
```diff
@@ -0,0 +1,145 @@
...
+## Set types
+
+There are two types of sets: standard and user-generated.
+
+### Standard
+
+Standard sets have one of the names (the `n` tag value) listed here to guarantee interoperability.
+The convention is to prefer lowercase English words in singular form with dash separator.
+
+| n-tag | description | expected d-tag prefixes if using kind:30161 |
```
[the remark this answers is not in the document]
arthurfranca: Oops, forgot to replace it. Ended up using 30382 cause they had same structure with minor differences

61.md
```diff
@@ -0,0 +1,145 @@
...
+  tags: [
+    ["d", "<random>"],
+    ["n", "`14pqug65sgl"]
+    // other tags
+  ]
+  // other fields
+}
+```
+
+**2) Add a `n` tag to a "Set Item Reference" `kind:30382` event that references the item to be added:**
```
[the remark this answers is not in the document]
arthurfranca: I'll go with your idea of different kinds for different set items, so your relationship status thing would be for pubkeys only.

61.md
```diff
@@ -0,0 +1,145 @@
...
+| file | adds events and files (urls) to a root directory similar to "/" | "i:", "a:", "r:" (url) |
+| friend | friends' pubkeys | "p:" |
+| mute | pubkeys whose posts the user doesn't want to see in their feeds | "p" (pubkey) |
+| mute-event | muted notes | "i:" (event), "a:" (event) |
+| mute-hashtag | muted hashtags | "t:" (hashtag) |
+| mute-word | muted words | "word:" (lowercase string) |
+
+### User-generated
+
+Set names must be prefixed with a "\`" (backtick) when on `n` tags to avoid name clashing.
```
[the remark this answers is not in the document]
arthurfranca: You're right, clients can hardcode which ones are standard and prevent users from trying to use these names for custom sets.

61.md
```diff
@@ -0,0 +1,145 @@
...
+An event can have many `n` tags, thus being member of many event sets.
+To remove an event from a set, remove the corresponding `n` tag from the event or
+send a deletion event (`kind:5`).
+
+## Referencing
+
+An event set is referenced by an `s` tag. It
+references many events from an author instead of a single one:
+
+`["s", "<kind, optional>,<other kind, optional>:<32-bytes lowercase hex of a pubkey>:<n tag value>", "<recommended relay URL, optional>"]`
```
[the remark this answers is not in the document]
arthurfranca: I'll change it to 1 required kind. It makes even more sense now that say followed pubkey refs and followed hashtags refs will have different kinds.

61.md
arthurfranca: Ended up with 1 required kind to lower complexity and cause it aligns with recent changes regarding "Set Item Reference" events.

arthurfranca:
@vitorpamplona added private event sets, extracting it from your relationship status PR. Do you need the kinds to change or are they ok?

vitorpamplona:
I almost made the Private stuff a separate NIP. I am not sure Clients must implement both to declare compliance with this NIP. 

Because you have the "Table of "Set Item Reference" Event Kinds", you need to define the d tag for each of the kinds. 

I would move the section "How to add items to a set" to before "Set types" because right now you are describing Set Types that are for multiple types: `follow` and `friend` and for pubkey sets; `bookmark` are for event sets. But you only describe that this difference exists much later.

arthurfranca:
> I almost made the Private stuff a separate NIP. I am not sure Clients must implement both to declare compliance with this NIP.

Shouldn't clients just pick and use the sets they need? I mean, a client doesn't need to support all public sets and/or all private sets in this NIP. I added the private stuff cause it is the same thing with just the `d` tag twist addition.

> Because you have the "Table of "Set Item Reference" Event Kinds", you need to define the d tag for each of the kinds.

Didn't understand. Do you mean the NIP should define the `d` tag separately for public and private versions? But the values are the "same". The difference is that private version obfuscates the tag value, as mentioned later at `Private "Set Item Reference" Event:`.

> I would move the section "How to add items to a set" to before "Set types"

Done, thanks

vitorpamplona:
> Shouldn't clients just pick and use the sets they need?

Yeah, but partial support is tricky. Like do you merge the `n` tags of public and private sets to the same public key? Are those two separate sets? Questions that make things complicated. 

> Didn't understand. Do you mean the NIP should define the d tag separately for public and private versions?

Yep, otherwise it just looks like private events only work for public keys.

vitorpamplona (commented):
Additional comments on the structure of the text.

alexgleason:
Even though this solves follow lists being huge, it doesn't solve follow notifications. I wish it did.

vitorpamplona:
> Even though this solves follow lists being huge, it doesn't solve follow notifications. I wish it did.

It kinda solves it. If pubkeyA makes a `REQ` for `kind:30382` and `d:pubkeyA`, pubkeyA will receive all modifications made by everybody else's follow (and other lists) with pubkeyA. You can show a notification like this: 

pubkeyB has changed your status to include the lists: `follow`, `friend` and `idiot`.

arthurfranca:
> It kinda solves it. 

Yeah the client would need to track `n` tag changes cause although the filter `{ kinds: [30382], #d: ["<user>"], #n: ["follow"], since: <last-time-checked> }` would include events that were updated since the timestamp, the update could have been on another `n` tag.

Tracking isn't that great, cause one may have thousands/millions followers.

Ideas:
A) `kind:30382`  `n` tags could include a timestamp to use client side like `["n", "follow", "<updated_at>"]`

B) Or we add uppercase `N` tag to the `kind:30382` with just `n` tag values that were added/removed at last event update. Then you do `{ ..., #N: ["follow"] }` 

C) Or we could use an auxiliary "Set Change Notification" event just with the `n` tag changes?

vitorpamplona:
I like the local solution. If it is the first start, the notification can just be more generic (people don't care about new follows in the first start). Once the client has received some events, it can then display the differences between them on updates.

alexgleason:
Why does the target event's pubkey need to be in the `d` tag for kind 30383 events?

61.md
```diff
@@ -0,0 +1,193 @@
...
+The kind and `d` tag value will vary depending on the type of item being referenced.
+
+**Some** of the event kinds **require** a `k` tag set to the referenced item kind to help with filtering or other tags.
+
+**Table of "Set Item Reference" Event Kinds and Structure by Referenced Item Type**:
+
+| Type | Kind | d-tag | other required tags |
+|-|-|-|-|
+| **pubkey** | 30382 | `<user-pubkey>` | |
+| **event** | 30383 | `<event-pubkey>:<event-id>` | `["k", "<event-kind>"]` |
```
alexgleason, line 54:
```suggestion
| **event** | 30383 | `<event-id>` | `["k", "<event-kind>"]` |
```

Including the pubkey is a painful annoyance. It makes it harder to query, and harder to create the event, with no apparent benefits.

alexgleason: I saw in another thread, the justification was that it can be used like a relay hint. That's not a good enough reason to put it in the d-tag value. You could use `["d", <event-id>, <optional relay>]` for the same purpose, or even `["d", <event-id>], ["relay", <relay>]`

arthurfranca: Let's consider it a fact that both the referenced event author's pubkey and the regular relay hint are important when searching for the referenced event. The former is slower cause needs more steps to find the event, while the latter may suffer from link rot.

With that said, this was my jerk move to force people to add NIP-65 pubkey hint. I agree with the "annoyance" part if you meant that this shouldn't be placed in the `d-tag` cause it shouldn't be indexed by relays.

Besides the `["relay", "<relay>"]` what do you think of `["nip65pk", "<pubkey">]` for the pubkey hint?

alexgleason: In the spirit of the `k` tag, I ended up adding a `p` tag of the original author in my events. Here's an example:

```json
{
  "kind": 30383,
  "id": "00dab96e90bf45215b6c8b7e2d6c238977d5395eeeaf7b236d161875537f819f",
  "pubkey": "d4ac0aa9df4ae9c491fde58ae4846079ad20332a59d7a360a186122320ad780f",
  "created_at": 1717945869,
  "tags": [
    ["d", "b8d54511a88165d8e59d29617d66d76c4564e013bdcfec021d663bf63d3b46fe"],
    ["n", "closed"],
    ["p", "0461fcbecc4c3374439932d6b8f11269ccdb7cc973ad7a50ae362db135a474dd"],
    ["k", "1984"],
    ["P", "9989500413fb756d8437912cc32be0730dbe1bfc6b5d2eef759e1456c239f905"]
  ],
  "content": "",
  "sig": "14bc909d5ea13658833d0996ad09185ac5ffbf0bd2ab7f5a9f1a2f02df79e6bda212916258de1039c0ff695b28366f77e9b9ebd0811b26e04dbb1c7f46e58227"
}
```

arthurfranca: It works but the `k` tag is indexable cause there is a need to get just `30383` events referencing events of specific kinds.
I couldn't think of an use case where one wants to filter also by referenced event pubkey. We shouldn't add one-letter tags for no reason.

And what's the uppercase `P` for?

alexgleason: Every time a report comes in, I generate a kind 30383 event for it with `["n", "open"]`. Then instead of querying kind 1984 events, I query `{ kinds: [30383], authors: [me], '#k': [1984] }` and paginate through this list.

This allows me to mark reports as open or closed, and then query by the n tag to get reports depending on their status.

I want the "p" tag so I can also filter by the user who created the report, in conjunction with filters like "open"/"closed".

The uppercase "P" is people who the report tags, so I can similarly filter reports targeting specific users users along with "open"/"closed".

arthurfranca: > I want the "p" tag so I can...

Thx for the example, changed 30383 and 31383 to use just the event id in the d-tag. Added the `p` tag too.

alexgleason (approved):

alexgleason:
@arthurfranca Should this be unmarked as a draft?

vitorpamplona:
I am just realizing that with this, we are just swapping an event with thousands of keys (e.g. the contact list) to an event with thousands of list names (e.g. new `30382`). 

Different race conditions, but race conditions might still be a problem if lists become a common feature of Nostr clients.

Regardless, this is a good evolution.

alexgleason:
@vitorpamplona It seems very improbable that a single user would add another single user to 1000+ lists. Even power users will have _hundreds_ at most, while they will follow _thousands_ of users. So the scale is very different, and it minimizes the potential damage of race conditions to your relationship with individual users instead of your entire list.

alexgleason:
Btw, I implemented event sets in Ditto. The Ditto nsec tags pubkeys with `["n", "admin"]`, `["n", "suspended"]`, and `["n", "suggested"]` for user roles, globally suggested users, and moderation (etc.)

It also generates event sets for reports and some other things, which it tags with `["n", "open"]` and `["n", "closed"]` so the server can manage their state and filter by it.

I will soon be implementing private user notes and user-recommended follows with this. I think it's very needed, and the `n` tag is a great idea.

I replaced the custom kind 30361 event I was using in Ditto with kind 30382.

arthurfranca:
I think this PR is ready. Just would like to make it clear that although this PR has all the necessary steps to migrate from NIP-51 lists, it **does not** aim to make clients migrate from NIP-51. Though it opens the possibility to experiment migrating less "important" ones like using "mute" event sets instead of "mute" NIP-51 lists. Also, it opens the door to using events sets for new use cases where one would otherwise spec a new NIP-51 list.

staab:
IMO this not much more than a clever hack, and not a fundamental solution of the problem of follow list race conditions. I don't really see myself adopting this except to be interoperable with other features I want to support. But migrating things to it is just not going to happen, for me or many other devs.

To that point, it's a very bad idea to add some of the "standard" event sets listed. What is the semantic difference between a follow list and a follow set? Which is canonical? Is the newer one better? People are going to see this NIP and think it replaces kind 3 follow lists, and start writing only sets. Or they'll attempt to implement complicated reconciliation logic and fail. I think if people start to use any of the `bookmark`, `follow`, `mute` or `pin` sets, which are already defined by NIP 51, we're going to have fragmentation of implementations, and it will greatly damage interoperability.

arthurfranca:
One can't be against event sets while in favor of #1322 cause #1322 uses event sets. Simple logic.

In fact, from the above comment I only see you are against some of the event set names. I will remove bookmark, follow, mute and pin.

alexgleason:
Indeed. I was thinking, the only thing Event Sets competes against (as far as "do things one way") is Lists... _in concept_.

But it's the specific names/list kinds that need to not compete. The concepts should be able to coexist.

vitorpamplona:
Sets and Lists are different things. There is some intersection, but sets do not replace Lists at all times.

staab:
> One can't be against event sets while in favor of https://github.com/nostr-protocol/nips/pull/1322 cause https://github.com/nostr-protocol/nips/pull/1322 uses event sets. Simple logic.

I'm against both. I just recognize that I have no power to prevent sets being used. Yes, sets are different from lists, but not different enough IMO to complicate the protocol with them (especially when we also have labels).

arthurfranca:
Removed te part that is now available at #1322 (and also extended at #761).

The PR is just defining event sets and some basic event set names.

arthurfranca:
I'm going to stop pushing NIPs for a while after I felt that I had to deface this PR to fit whatever requirements to move this forward. Sometimes there is no guidance to improve a PR, just unconstructive reviews that show up after months.

More info at https://nostr.at/nevent1qy88wumn8ghj7mn0wvhxcmmv9uq36amnwvaz7tmwdaehgu3wvf5hgcm0d9hx2u3wwdhkx6tpdshsz9thwden5te0wfjkccte9ejxzmt4wvhxjme0qy2hwumn8ghj7un9d3shjtnwdaehgu3wvfnj7qgswaehxw309ahx7um5wghx6mmd9uq32amnwvaz7tmwdaehgu3wdau8gu3wv3jhvtcpzpmhxue69uhkumewwd68ytnrwghsz8nhwden5te0dehhxarj94c82c3wwajkcmr0wfjx2u3wdejhgtcpzemhxue69uhhyetvv9ujumt0wd68ytnsw43z7qgewaehxw309aex2mrp0yh8xmn0wf6zuum0vd5kzmp0qqspk2fwtnyr32ava9skgm27f8l5plveswxvk28dhd86tyxpjkccg9cqnvj9j

PR #785, opened 2023-09-17
NIP-100, Private Key Mnemonic for Nostr

kehiy:


fiatjaf:
Check NIP-06.

Egge7:
> should we close it?

Yes, seems like a duplicate without any added value.

fiatjaf:
Yes, we should close.

I even disagree with NIP-06, I think dealing with private keys directly is better than mnemonics in the case of Nostr, but that's irrelevant.

PR #786, opened 2023-09-18
add relay sets kind to nip-51

fiatjaf:


jb55:
On Mon, Sep 18, 2023 at 08:32:07AM -0300, fiatjaf_ wrote:
>+### Categorized Relay Set
>+
>+An event of kind `30002` is defined as a parameterized replaceable list event for categorizing bookmarks. The 'd' parameter for this event holds the category name of the list. The relays lists may contain metadata tags such as 'title', 'image', 'summary' as defined in [NIP-23 - Long-form Content](23.md). These sets can be used by clients in order to determine which relays to query in different scenarios.

should be: "is defined as a parameterized replaceable list event for categorizing *relays*"

verbiricha (approved):

pablof7z:
oh shit, nice to see this PR; I had pretty much the same thing using a different kind. LFG!

pablof7z:
I'm assuming the idea is to use `r` tags for each relay, but should we make that explicit?

PR #787, opened 2023-09-18
NIP-54: decentralized wikis

fiatjaf:
The basic idea is to have multiple articles with the same name, each created by a different pubkey, and filter between them, but have them all publicly accessible for different perspectives.

> "A change in perspective is worth 80 IQ points."
_Alan Kay_

These ideas is being implemented and experimented with in https://github.com/fiatjaf/wikistr and https://github.com/fiatjaf/nwiki, but it's still a work in progress and I'm not sure it will work, but it's promising. Please give feedback.

Although having multiple articles for the same name can appear to be complicated to deal with, I think in the end, if this is successful, things will settle down to a couple of mostly canonical articles for each topic, while controversial topics will have 3 or 4 different versions and people will have be able to see all the different takes on each topic, with spam and shitty content filtered out. It only takes a few people in the world participating in the curation process to yield a better decentralized encyclopedia for everybody.

This is all very much inspired by the amazing http://fed.wiki.org/.

Readable (draft) NIP text: https://github.com/nostr-protocol/nips/blob/wiki/54.md


fiatjaf:
I think the biggest issue, at least in my mind, is picking up the text syntax format for this. I've added an explanation to the NIP about why I chose Markdown, but if someone else has opinions on this it would be great to hear them because I'm a newbie.

vitorpamplona:
The only issue I see is that a replaceable event is not a great base for a WebOfTrust graph. Basically, I can design a page, get all the reputation I want, and then delete the text and change it to an #ad without clearing its WebOfTrust score. 

Google solves this with a User Experience component to PageRank which counts the seconds the reader takes to click and come back to the search results. If the user came back too fast, it decreases the association between the term and the page. It's going to be hard for Nostr clients to code the same scoring capability.

fiatjaf:
The reactions should tag the `e` too along with the `a`, that's how I'm making them. Then the client can say: "bob liked an older version of this article". I also expect that eventually people to design relays that will archive specific versions of replaceable events on-demand (for example, whenever I like something this special relay will go there and fetch that specific version and store it for me forever) and not allow them to be replaced. What do you think?

Well, also if someone starts making a bunch of ads that person will not be able to earn too much reputation. These things could work in the beginning but as the system matures and gains more mass it will become increasingly harder to game it.

jb55:
On Mon, Sep 18, 2023 at 06:28:00AM -0700, fiatjaf_ wrote:
>I also expect that eventually people to design relays that will archive specific versions of replaceable events on-demand (for example, whenever I like something this special relay will go there and fetch that specific version and store it for me forever) and not allow them to be replaced. What do you think?

fwiw nostrdb already works like this.

arthurfranca:
> Unlike normal Markdown links `[]()` that link to webpages, wikilinks `[[]]` link to other articles in the wiki.

Maybe this could be a NIP-19 `nwiki` similar to naddr but withought an author nor kind (assumed to be 30818). So just the d tag identifier and optional relays.

Then use regular markdown links with NIP-21 `[A link](nostr:nwiki1fnt...nqc)`

erskingardner:
> Maybe this could be a NIP-19 nwiki similar to naddr but withought an author nor kind (assumed to be 30818)

Doesn't that break the only requirement of Nostr events? They always have a kind and author. 

Why wouldn't you just included the assumed kind and the author of the page?

ok300:
Can links point to a specific version of a page?

arthurfranca:
@erskingardner the nwiki would reference not a single event, but a group of events.

I change my suggestion to something more generic that could also reference [NIP-61 - Unbounded Lists](https://github.com/nostr-protocol/nips/pull/784): `nset`

`nset` is clearer cause it references a set o events. It would encode `d-tag, kind, author(optional) and relays(optional)`. With an author, it supports NIP-61 (a set of events from an author). Without an author, it supports this wiki NIP (a set of events from any author)

Referencing by tag could use an unified tag too, I was using `u` but I will change it to `s`(set). It can be similar to an `a` tag but the order is different to allow for lack of author at the end:

Referencing a wiki page set: `["s", "<kind integer>:<d tag value>", "<recommended relay URL, optional>"]`
Referencing an unbounded list: `["s", "<kind integer>:<d tag value>:<32-bytes lowercase hex of a pubkey>", "<recommended relay URL, optional>"]`

fiatjaf:
> Can links point to a specific version of a page?

They should not. That would break the entire purpose of making this an open thing and not a centralized system.
But authors of one article must have a way to signal their preferred version of other articles and clients should prioritize or highlight these when moving from one article to other. Ultimately it's client choice though.

nostrband:
I thought that collaboration on a single article was the core wiki's feature. How can I suggest a small correction for one's article? Maybe a 30820 with my changes in Unified Format, linking to the target 30818 by e and a? Or is this out of scope for now?

fiatjaf:
I think it's in scope and a good suggestion, I must think more about it.

pablof7z:
It would be useful to consider a PR-style event where a pubkey suggests making a change to an event beyond the scope of this NIP alone. I've seen a few people interested in this functionality and we should avoid having three different ways of doing the same thing if we can avoid it.

vincent822:
This is my suggestion for how to do it: (I think it would enable good collaboration, version management and a PR scheme)

three new kinds:
1. kind 1500 (regular) - creating a new document. This event will serve as the anchor event of the document. All the versions will point to it.
1. kind 1501 (regular) - creating a new version  
1. kind 30050 (parameterized replaceable event) - endorsing a specific version 

let's give an example:

Alice wants to create a new wiki article for "Blue Whale".
She creates a 1500 kind event (new document):
```
{
  "id": 5c83..,
  "kind": 1500,
}
```
Then she creates the first version by creating event 1501:
 ```
 {
  "id": 7d01..,
  "kind": 1501,
  "tags": [
    ["document_event", 5c83.. ],
    ["previous_version", None],
    [“title”: “Blue Whale”]
],
  "content": 
"The blue whale is a marine mammal..."
}
```
Then, she endorses her version with kind 30050:
```
 {
  "id": 9b1a..,
  "kind": 30050,
  "tags": [
    ["document_event", 5c83.. ],
    ["d", 7d01..],
],
}
```
Bob notices that Alice didn't include the scientific name of the blue whale. So he creates a new version.
```
 {
  "id": 87d1..,
  "kind": 1501,
  "tags": [
    ["document_event", 5c83.. ],
    ["previous_version", 7d01],
    [“title”: “Blue Whale”]
],
  "content": 
"The blue whale (Balaenoptera musculus) is a marine mapmmall…"
}
```

Then He endorses his new version with kind 30050.

He comments on his version with kind 1, tags Alice, and asks her to endorse this new version (**you can call it some kind of PR if you want**). Alice comments back that it’s a good change but asks him to also fix a typo in the word mammal and modify the title to include the scientific name too.

So Bob creates a new kind 1501:
```
 {
  "id": 92ab..,
  "kind": 1501,
  "tags": [
    ["document_event", 5c83.. ],
    ["previous_version", 87d1],
    [“title”: “Blue Whale (Balaenoptera musculus)”]
],
  "content": 
"The blue whale (Balaenoptera musculus) is a marine mammal…"
}
```

Bob endorses his version. He asks Alice to endorse his new version. Alice endorses his new version.

Now Alice and Bob agree on the "true" version of the article. 

Of course Alice has no kind of "ownership" on the article just because she created the root 1500-kind event. Other people can edit each one of the article versions and agree on some new version. Chances are that on non-controversial topics, consensus will emerge between most of the "good" players about what is the "official" version of the document, while controversial topics will have more than one popular version, as Fiatjaf siad.

dluvian:
It would be nice if articles MUST/SHOULD include a language code tag. You generally want to search for articles of only those languages that you can read.

fiatjaf:
@pablof7z @hzrd149 do you feel this is ready to merge?

pablof7z:
> @pablof7z @hzrd149 do you feel this is ready to merge?

Yes

hzrd149:
This looks good to me, however it needs example events in the `[INSERT EVENT EXAMPLE]` sections :grin: 
Also it would be good to have an example fork event in the "Forks" section

fiatjaf:
> This looks good to me, however it needs example events in the [INSERT EVENT EXAMPLE] sections 😁

We can fix that later.

PR #791, opened 2023-09-21
NIP-99 use a tags pointing to Badge Definition events for access control

neilck:
When 'a' tags referring Badge Definition events (kind 30009) are added, a client application CAN CHOOSE to verify the user has been awarded all respective badges before granting access to the product / service defined by the Classified Listing event.

Use Cases:
Requiring a "not-a-robot" badge to access a community group.
Requiring a "supporter" badge to gain early access to client features.
Presenting a "paid membership" badge to unlock access to one or more paid relays.

Creation of classified listing event with 'a' tags, and verification of badges (determining eligibility) implemented in [nostr-access-control](https://github.com/neilck/nostr-access-control) library.

Source code of application showing how to integrate this library is available at [nac-demo-app](https://github.com/neilck/nac-demo-app).

### DIFF
--- a/99.md
+++ b/99.md
@@ -48,6 +48,9 @@ The following tags, used for structured metadata, are standardized and SHOULD be
 Other standard tags that might be useful.
 
 - `"g"`, a geohash for more precise location
+- `"a"`, one or more a tags referencing kind `30009` Badge Definition events (NIP-58)
+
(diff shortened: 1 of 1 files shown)


erskingardner:
@neilck I don't think this really needs to be added to the NIP. Since `a` tags are universally usable on events, you can of course already do this without the explicit inclusion in this NIP.

PR #794, opened 2023-09-24
add nip-24: extra metadata fields

fiatjaf:
Because Nostr won't survive if all clients have to keep checking for `displayName`, `handle`, `username` and whatever else is in there.

https://github.com/nostr-protocol/nips/blob/extra/24.md

Motivation: https://njump.me/nevent1qqs9pt0ghyzzq7p6pqajnlvjgr7n9tyhy0vhpuyhgvdrd6y5r50n90cppemhxue69uhkummn9ekx7mp0qyghwumn8ghj7mn0wd68ytnhd9hx2tcar0mnt

### DIFF
--- a/24.md
+++ b/24.md
@@ -1,0 +1,24 @@
+NIP-24
+======
+
+Extra metadata fields and tags
+------------------------------
+
+`draft` `optional` `author:fiatjaf`
+
+This NIP defines extra optional fields added to events.
+
+kind 0
+======
+
+These are extra fields not specified in NIP-01 that may be present in the stringified JSON of metadata events:
+
+  - `display_name`: a bigger name with richer characters than `name`. Implementations should fallback to `name` when this is not available.
+  - `website`: a web URL related in any way to the event author.
+
+tags
+====
+
+These tags may be present in multiple event kinds. Whenever a different meaning is not specified by some more specific NIP, they have the following meanings:
+
+  - `r`: a web URL the event is referring to in some way
--- a/24.md
+++ b/24.md
@@ -15,6 +15,14 @@ These are extra fields not specified in NIP-01 that may be present in the string
 
   - `display_name`: a bigger name with richer characters than `name`. Implementations should fallback to `name` when this is not available.
   - `website`: a web URL related in any way to the event author.
+  - `banner`: an URL to a wide (~1024x768) picture to be optionally displayed in the background of a profile screen.
+
+### Deprecated fields
+
+These are fields that should be ignored or removed when found in the wild:
+
+  - `displayName`: use `display_name` instead.
+  - `username`: use `name` instead.
 
 tags
 ====


24.md
```diff
@@ -0,0 +1,24 @@
...
+`draft` `optional` `author:fiatjaf`
+
+This NIP defines extra optional fields added to events.
+
+kind 0
+======
+
+These are extra fields not specified in NIP-01 that may be present in the stringified JSON of metadata events:
+
+  - `display_name`: a bigger name with richer characters than `name`. Implementations should fallback to `name` when this is not available.
```
alexgleason, line 16: Wait, why does `display_name` exist? This is the definition for `name`. NIP-01 is wrong.

`name` is not a username. It's a display name. Nostr inherently does not have usernames. The thing that makes a username a username is that it's globally unique. The closest thing we have to real usernames is `nip05`.

There's no reason the `name` shouldn't have special characters, and there's no reason to support `display_name` and then fall back on it.

The reason for the discrepancy is because of nostr.rocks, when the concept of NIP-05 didn't exist and people were LARPing with usernames. But that was an imaginary, pretend experiment. Now we have something resembling actual usernames with NIP-05. We only need one name field.

fiatjaf: I agree, I don't see the point of having `display_name`, but it is better to have just `display_name` than to have `display_name`, `displayName`, `username`, `userName`, `handle` and what other absurdities people are doing out there.

Many clients are treating `name` as a unique identifier, prepending it with a `@` to feel like Twitter and all that.

alexgleason: It makes way more sense to use `@{nip05}` fallback to `@{nip19}`. To use `@{name}` practically guarantees that Nostr won't be understood. It's a bad UX that shouldn't be enshrined in a NIP.

I would add a "deprecated fields" section of this document and list the ones found in the wild, with a recommendation about which fields to use instead. Eg:

## Deprecated fields

The following fields have been in use by clients in the past, but considered deprecated and should be avoided.

- `display_name`, `displayName`, `username`, `userName`, `handle`, etc. - use `name` instead. For human-readable usernames, see NIP-05.

vitorpamplona:
If you want to add: 
- `picture`: the user's profile picture (ideal: squared, 300x300 pixels)
- `banner`: the user's profile banner (ideal: wide, 1024x768 pixels)
- `about`: a description of the author (it should be rendered like `kind1.content`s do). 
- `nip05`: a single Nostr address in the format of an e-mail
- `lud06`: a lud06 address
- `lud16`: a lud16 address in the form of an e-mail.

24.md
vitorpamplona: `display_name` exists because, in the early days, many clients (including Damus/Amethyst/Snort,Astral) didn't support @ with a space in the name. 

These days we might not need it anymore. So, I would be okay with calling it deprecated and using just `name` instead.

erskingardner:
ACK on marking the various versions of display name, et al. as deprecated. A single `name` field for what gets displayed plus nip-05 should be enough.

alltheseas:
> If you want to add: 
> 
> - `picture`: the user's profile picture (ideal: squared, 300x300 pixels)
> 
> - `banner`: the user's profile banner (ideal: wide, 1024x768 pixels)
> 
> - `about`: a description of the author (it should be rendered like `kind1.content`s do). 
> 
> - `nip05`: a single Nostr address in the format of an e-mail
> 
> - `lud06`: a single lud06 address
> 
> - `lud16`: a single lud16 address in the form of an e-mail. 

What do y'all think of a single optional field titled `out of band` where user can add a e.g. SimpleX address

melvincarvalho:
No strong opinion on this, but `r` is also used for relay

dtonon:
ACK on dropping `display_name`, it is duplicate/confusing.
If someone prefer to be cited by nick but also want to show the full name, he can use `about `.

24.md
DanConwayDev: There is a strong case for having two short text fields to express identity in different ways that don't have the constraints of a nip05. Many clients have opted for this. It would be great if a consistent standard was adopted. 

type 1: how a user would like to be addressed eg Will in spite of potential identity collisions.
type 2: an identifier with a smaller chance of identity collisions such as '[firstname] [lastname]' or a username used on other platforms eg jb55.

jb55:
On Sun, Sep 24, 2023 at 01:47:34PM -0700, daniele wrote:
>ACK on dropping `display_name`, it is duplicate/confusing.
>If someone prefer to be cited by nick but also want to show the full name, he can use `about `.

In Damus display_name is an optional real name and `name` is used as a
shortform for @ mentions. For example my name field is set to jb55 which
is my internet nickname/shortname and is rendered inline for mentions,
display_name is for showing my real name at the top of my posts. I don't
agree to "just use nip05" for @ mentions because many users may not have
nip05's. I also don't agree that there should be only one field, they
are different things.

To adapt to some clients only using display_name OR name I eventually
had to change Damus to handle either field being set and treating two
field set as the original damus behavior, and either field being set as
their single name (which is used in @ mentions and display)

I don't really understand why this got so out of hand in other clients.
name has always been specified, the fact that some clients were only
setting a *custom field specific to Damus* instead of name is pretty
bonkers.

If people want to use the custom Damus field they should probably stick
to Damus' intended purpose for this field.

24.md
DanConwayDev: Here is a short extract from a draft unpublished essay I wrote many months ago on identity expression and the potential for social verification on nostr:

> #### display_name
> 
> This has come to be used to display a user's handle: A string of characters that is short, memorable and relatively unique (traditionally they had to be unique on a given platform, but not the case on nostr) used across many online platforms. It almost always contains only letters and numbers with the occasional @ prefix such as @waxwing or @yegorpetrov). 
> 
> Sometimes it incorporates the user's first or second name, aspirations, or identity such as cultural references they resonate with such as heros. It therefore is often a creative expression of identity.
> 
> It is occasionally omitted all together. This may become more commonplace through the adoption curve as users join who are not part of username driven online communities.
> 

> 
> #### name
> 
> A more personal but less unique expression of the identity than display_name.
> 
> * Often the "[firstname] [lastname]" eg "Lyn Alden"
> * Sometimes "[firstname]" eg "Will"
> * Sometimes its the user's handle with spaces and capitalisation. Sometimes its identical to the users handle.
> * Occasionally "[lastname]" eg "Caremen"
> * Occasionally omitted
> 
> Feels like a closer and more direct human connection than a handle. It is much more likely to collide with other identities, much less of a creative expression of identity and personality.

jb55:
On Sun, Sep 24, 2023 at 02:14:07PM -0700, DanConwayDev wrote:
>@DanConwayDev commented on this pull request.
>Here is a short extract from a draft unpublished essay I wrote many months ago on identity expression and the potential for social verification on nostr:
>
>> #### display_name
>>
>> This has come to be used to display a user's handle: A string of characters that is short, memorable and relatively unique (traditionally they had to be unique on a given platform, but not the case on nostr) used across many online platforms. It almost always contains only letters and numbers with the occasional @ prefix such as @waxwing or @yegorpetrov).
>>
>> Sometimes it incorporates the user's first or second name, aspirations, or identity such as cultural references they resonate with such as heros. It therefore is often a creative expression of identity.
>>
>> It is occasionally omitted all together. This may become more commonplace through the adoption curve as users join who are not part of username driven online communities.
>>
>
>>
>> #### name
>>
>> A more personal but less unique expression of the identity than display_name.
>>
>> * Often the "[firstname] [lastname]" eg "Lyn Alden"
>> * Sometimes "[firstname]" eg "Will"
>> * Sometimes its the user's handle with spaces and capitalisation. Sometimes its identical to the users handle.
>> * Occasionally "[lastname]" eg "Caremen"
>> * Occasionally omitted
>>
>> Feels like a closer and more direct human connection than a handle. It is much more likely to collide with other identities, much less of a creative expression of identity and personality.

this is so wrong. it's the opposite. please stop spreading falsehoods.

alexgleason:
It's fine to use `@{name}` when rendering mentions inside of a post body. What doesn't make sense is to use `@{name}` as a handle in the _UI_, like this:

image

In the image above, `@alex@gleasonator.com` only makes sense as a NIP-05 or a NIP-19.

image

Mentions inside of text are another story. If you have a NIP-05, it should be used. If you don't, I think it's alright to fall back on an `@{name}` where the name can be a display name. Eg I see no problem with `@Alex Gleason` (highlighted or underlined) inside of text.

image

dtonon:
> For example my name field is set to jb55 which
is my internet nickname/shortname and is rendered inline for mentions,
display_name is for showing my real name at the top of my posts.

I understand. But it is fairly uncommon online for someone to want to share both a username and real name. In this case you could set the name as `Will (jb55)` and you would have a consistent display inside the app and good interoperability.
Make sense?

DanConwayDev:
> this is so wrong. it's the opposite. please stop spreading falsehoods.

apologies @jb55. I definately got it the wrong way round for your profile. I think I was trying to shoehorn some consistency when there were none was I was drafting. I should have left that unpublished... I still feel the point stands around the case for display_name and name. Its not a hill I'm going to die on :)

jb55:
On Sun, Sep 24, 2023 at 02:26:21PM -0700, daniele wrote:
>I understand. But it is fairly uncommon online for someone to want to share both a username and real name.

ok? that's nice. I do.

> In this case you could set the name as `Will (jb55)` and you would
> have a consistent display inside the app and good interoperability.
> Make sense?

no it doesn't make sense at all. I see `name` as a default petname for
@'s. This is very different than the name users might want to display
next to their profile picture. It also has different indexing
implications when creating client-local search (you may only want to
index petnames and @ mention names).

Using `name` is perfectly interoperable and always has been. It's
perfectly fine to put a realname in `name`. The display_name is for when
users want to make distinction (like me).

I don't understand why people are talking about deprecating a *custom
Damus field* that I don't plan on removing.

melvincarvalho:
FWIW, going back as far as 2004, FOAF always had `nick` (handle) and `name` (full name) to be separate things, and most people that I can remember had both.

jb55:
On Sun, Sep 24, 2023 at 02:39:18PM -0700, William Casarin wrote:
>I don't understand why people are talking about deprecating a *custom
>Damus field* that I don't plan on removing.

I think maybe the main concern would be the possibility of Damus users
only setting their real name and setting an empty username, but I can
make sure that if a Damus user only sets their real name then only the
`name` field is updated.

dtonon:
> ok? that's nice. I do.

I do it too :)
So deprecating `display_name` would make things harder for me, but I think it simplifies a lot the UX of the majority.
For example developing njump I noted that lot of people fill both, with the same value, and often it is a long multi word name. So `name` is not used as you (and me) intended.
`display_name` would have a sense if `name` was clearly constraints like a username (a-z0-9, no spaces or other symbols).

I didn't know display_name way a Damus custom name, it's used everywhere.

alexgleason:
Clients should offer nip05 registration on a domain they control, if possible. This would help justify having only one `name` field, and would be more comprehensible for users.

vitorpamplona:
Keep in mind this NIP is just guidance on what the usual behavior should be. It's fine for clients to create new fields if they need them in their user interfaces. They just won't show up in other clients (which is fine).

melvincarvalho:
> display_name

It would have been nice if we had done this at that start, but should we standardize on one system across the whole of nostr, tags, fields etc.

Given that created_at is always going to be in NIP-01.  Perhaps new things should be with underscores, too, to be consistent.

cypherhoodlum:
Having a single "primary" `name` field and the "secondary" NIP-05 with NIP-19 as a fallback is the best specification going forward in my opinion. Parroting what others have already said, `display_name` and other obscure name field variations should be deprecated. Nothing is binding.

fiatjaf:
Can we all agree `name` is required (or at least as required as such a thing can be in Nostr), `display_name` is optional, other variants are forbidden and should be nuked out of the Earth?

fiatjaf:
Clients should try to enforce the presence of `name` -- for example they can set that automatically if the user only fills in `display_name`. But they shouldn't require people to fill in `display_name`.

Clients should not read values from nonstandard fields like `displayName`, as that just perpetuates the mistakes of other clients.

fabianfabian:
I agree the focus should be on using `name` as a display name field, usernames are supposed to be unique so that description has been confusing in NIP-01.

The only valid reason I see for another field is a short mention/nickname, ideally we would have a different field for that that is not `name` with the constraints described but its probably too late for that because clients are already using `name` this way.

mikedilger:
My only comment is that when you tag somebody and type @, you need to type their name.  If people put spaces and Unicode characters in their 'name' it makes it very hard to tag them. Gossip  can work around this with petnames, but maybe we should suggest people's names are type-able with no white-space, and then write clients to handle the real world mess including collisions both accidental and malicious.

fiatjaf:
@pablof7z @staab @mattn @mbrat1 @mmalmi @v0l

v0l:
On Snort we used to allow editing the `display_name` but it was removed recently, we still show the `display_name` over `name` though.

I think there is a good case for `display_name` and it should be included in the NIPs

frnandu:
On Yana we use in this order to display in most places (first one that is filled) ->> 1) display_name 2) name 3) npub.....
Neither name nor display_name is mandatory.
The only thing I'm not sure about is whether name should be required before being able to fill display_name.... 
What are the down sides to having a display_name filled but not a name?

mattn:
I'm collecting implementations of Nostr clients about this. The numbers are ordered numbers prefered for displaying that names.

https://docs.google.com/spreadsheets/d/1L4h1xs3gKP5OcMfQ8h-N59FMPL1OaBVnGbcL4Vouyt0/edit#gid=0

erskingardner:
> Does anyone know what clients are using the weird formats? I just checked Coracle, Snort and some other smaller clients for which I had the source code locally and they all seem to be only using only name and display_name.

NDK handles kind 0's that have either `display_name` or `displayName` by setting both values to the same field on the `NDKUserProfile` object to the `displayName` property. Given the library is typescript, camel case is the convention but we do use `created_at` on `NDKEvent` so maybe there is a case to change to `display_name` @pablof7z?

In the [ndk-svelte-components](https://github.com/nostr-dev-kit/ndk/tree/master/ndk-svelte-components) library the Name component defaults to using `display_name` -> `name` -> `nip-05` -> `npub` in that order. This is what Listr, Ostrich.work, and other apps I've worked on use.

pablof7z:
`camelCase` is a shitcoin and I wish it would die

`display_name` > `displayName`

If we have agreement on this, I'll change NDK to read both `displayName` and `display_name` and always write `display_name`

dluvian:
> Does anyone know what clients are using the weird formats? I just checked Coracle, Snort and some other smaller clients for which I had the source code locally and they all seem to be only using only `name` and `display_name`.

Nozzle uses `username` and ignores everything else. I can change it to `name`.

fiatjaf:
> If we have agreement on this, I'll change NDK to read both displayName and display_name and always write display_name

@pablof7z the idea is to stop reading `displayName` entirely to prevent other clients from relying on that. From the comments here and @mattn's spreadsheet it looks like **no one** is really writing `displayName` except for Amethyst, and even Amethyst only does it "just to be sure", so ideally you would do as @mbrat1 and stop even reading that.

And, when writing, be a nice gentleman and nuke it out of your user's metadata.

melvincarvalho:
> It's fine to use `@{name}` when rendering mentions inside of a post body. What doesn't make sense is to use `@{name}` as a handle in the _UI_, like this:
> 
> image
> 
> In the image above, `@alex@gleasonator.com` only makes sense as a NIP-05 or a NIP-19.
> 
> image
> 
> Mentions inside of text are another story. If you have a NIP-05, it should be used. If you don't, I think it's alright to fall back on an `@{name}` where the name can be a display name. Eg I see no problem with `@Alex Gleason` (highlighted or underlined) inside of text.
> 
> image

Github uses both full name and username

image

Twitter uses both fullname and username

image

melvincarvalho:
Perhaps we need 3 names:

1 full name (name)
2 username (non name-spaced)
3 nip-05 (name-spaced)

Then something like [nomen](https://nomenexplorer.com/explorer) can be used if there's a tie breaker for usernames, for things like https://primal.net/jack -- id encourage all clients to use nomen for tie-breakers, and then have consistency across nostr.

I will definitely use all 3 of these in [nosdav](https://nosdav.com/), as 2/3 will only ever give a partial UX.

pablof7z:
> > If we have agreement on this, I'll change NDK to read both displayName and display_name and always write display_name
> 
> @pablof7z the idea is to stop reading `displayName` entirely to prevent other clients from relying on that. From the comments here and @mattn's spreadsheet it looks like **no one** is really writing `displayName` except for Amethyst, and even Amethyst only does it "just to be sure", so ideally you would do as @mbrat1 and stop even reading that.
> 
> And, when writing, be a nice gentleman and nuke it out of your user's metadata.

NDK now ignores `displayName`

mattn:
Yes, as far as my short digging the code of ndk, ndk seems to be using display_name.

https://github.com/nostr-dev-kit/ndk/blob/66798975fedcd9740aa0712b0430289d680ffecd/ndk/src/user/profile.ts#L31

PR #795, opened 2023-09-27
Add relay list of kind 3 to NIP-24

AsaiToshiya:
A relay list of kind 3 is not written anywhere and is not clear.

Is the relay list unnecessary? Or is it wrong to write in NIP-24?

### DIFF
--- a/24.md
+++ b/24.md
@@ -24,6 +24,13 @@ These are fields that should be ignored or removed when found in the wild:
   - `displayName`: use `display_name` instead.
   - `username`: use `name` instead.
 
+kind 3
+======
+
+This is extra field not specified in NIP-02 that may be present in the stringified JSON of events `content`:
+
+  - `{<relay-url>: {"read": <true|false>, "write": <true|false>}, ...}`: an object of relays used by a user to read/write.
+
 tags
 ====
 
--- a/24.md
+++ b/24.md
@@ -29,7 +29,7 @@ kind 3
 
 This is extra field not specified in NIP-02 that may be present in the stringified JSON of events `content`:
 
-  - `{<relay-url>: {"read": <true|false>, "write": <true|false>}, ...}`: an object of relays used by a user to read/write.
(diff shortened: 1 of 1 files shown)


vitorpamplona (approved):
Finally!

vitorpamplona:
We can mark is as deprecated, but there is no reason to not document the reality that is out there.

AsaiToshiya:
I think it is best to document and deprecate the relay list.

staab:
I could get behind that

staab (approved):

staab (commented):
Don't let me stop anyone who wants to merge this, but I still oppose the change at all. The spec isn't the place to clutter things up with random additional information. If it's not standard, it probably shouldn't be here.

https://github.com/nostr-protocol/nips/pull/746#issuecomment-1738255388

AsaiToshiya:
> The spec isn't the place to clutter things up with random additional information. If it's not standard, it probably shouldn't be here.

If you do not merge this PR for this reason, we should remove below:

https://github.com/nostr-protocol/nips/blob/d6f4598ad6d32b46de6444f39339819faab670d5/24.md?plain=1#L20-L25

https://github.com/nostr-protocol/nips/blob/d6f4598ad6d32b46de6444f39339819faab670d5/65.md?plain=1#L28

fabianfabian:
Strongly agree with at least document, and would prefer deprecate too.

jb55:
"should be removed" seems a bit aggressive

jb55:
On Thu, Sep 28, 2023 at 01:01:09PM +0200, William Casarin wrote:
>"should be removed" seems a bit aggressive

This is like saying it's deprecated but its ok to break clients that haven't updated yet. Pretty bad...

fiatjaf:
You're right, I think the text was just copied from the deprecation above. Fixed it in 84f1915cecc10cfca6584eacfbeecb35bddc65f9.

AsaiToshiya:
I prioritized using the same text, but you're right. Thank you for fixing that.

PR #798, opened 2023-09-28
65: trim trailing slashes from relay URIs

badonyx:
This is intended to address the issue of some clients duplicating connections due to trailing slashes in the relay URI.

### DIFF
--- a/65.md
+++ b/65.md
@@ -59,3 +59,5 @@ This NIP allows Clients to connect directly with the most up-to-date relay set f
 4. DMs SHOULD only be broadcasted to the author's WRITE relays and to the receiver's READ relays to keep maximum privacy. 
 
 5. If a relay signals support for this NIP in their [NIP-11](11.md) document that means they're willing to accept kind 10002 events from a broad range of users, not only their paying customers or whitelisted group.
+
+6. Clients SHOULD trim trailing slashes (`/`) from relay URIs to canonicalize connections.


vitorpamplona (approved):

staab (approved):

mikedilger:
This is backwards.

I used to do this, trim trailing slashes.  But trailing slashes are NOT trailing, they are leading. And canonical URLs have a path.  The root path is "/" not "".  I didn't think that was right, but I was convinced otherwise after a lot of arguing and research, and I put a lot of work into changing my nostr libraries to make sure every URL has a path of at least a "/".

I still could be wrong. I'm not remembering what convinced me. But I definitely remember this was a huge upset in my codebase having to change from no-slashes to all-slashes and it was not a decision taken lightly.

mikedilger:
https://stackoverflow.com/questions/2581411/do-web-browsers-always-send-a-trailing-slash-after-a-domain-name

fabianfabian:
@mikedilger if you can remember let us know

I think there is no difference between `example.com` and `example.com/` but there is a difference between `example.com/popular` and `example.com/popular/`. 

So for the first example you can remove the trailing slash and it will be the same, but for the last example you cannot they are different paths.

badonyx:
I was hoping to have this discussion before it got merged. Should have marked as draft I guess.

mattn:
The path appended to the wss URL should be used for first request for the HTTP server. This request is upgraded by the HTTP server. i.e. first request might be:

wss://example.com/ws/

GET /ws/

Or else

wss://example.com/ws?param=/

GET /ws?param=/

jb55:
On Fri, Sep 29, 2023 at 01:43:00AM -0700, badonyx wrote:
>I was hoping to have this discussion before it got merged. Should have marked as draft I guess.

yes, weird that this was merged so quickly. There is a case to be made that trailing / should be the default and not the other way around.

alexgleason:
> There is a case to be made that trailing / should be the default and not the other way around.

Indeed. Depends if it's a bare URL or not:

```
> new URL('wss://relay.mostr.pub').toString()
"wss://relay.mostr.pub/"
> new URL('wss://gleasonator.dev/relay').toString()
"wss://gleasonator.dev/relay"
```

In the first case, the URL is normalized with a slash at the end. Second case, it isn't. The reason is because URLs have a path. If you don't provide a path, the path is normalized to `/`. If you provide a path, it uses the path.

jb55:
On Fri, Sep 29, 2023 at 08:45:25AM -0700, Alex Gleason wrote:
>> There is a case to be made that trailing / should be the default and not the other way around.
>
>Indeed. Depends if it's a bare URL or not:
>
>```
>> new URL('wss://relay.mostr.pub').toString()
>"wss://relay.mostr.pub/"
>> new URL('wss://gleasonator.dev/relay').toString()
>"wss://gleasonator.dev/relay"
>```
>
>In the first case, the URL is normalized with a slash at the end. Second case, it isn't. The reason is because URLs have a path. If you don't provide a path, the path is normalized to `/`. If you provide a path, it uses the path.

ok, we should do what standard URL normalization does, Damus is in the
wrong here which is also leading to duplicates in relay lists.

staab:
I'm sorry for merging too quickly, I thought stripping slashes was the consensus. @badonyx can you open a new PR that corrects the language based on Alex's comment?

PR #799, opened 2023-09-28
05: domain without local part

badonyx:
Based on discussion at nostr:nevent1qqs2m9697a200hfpmzsq29n5w3a4mcm67y7tspnxzjz9yz0d92chrqqh3npcx

Existing clients that do this: amethyst, primal

### DIFF
--- a/05.md
+++ b/05.md
@@ -72,7 +72,7 @@ A client can also use this to allow users to search other profiles. If a client 
 
 ### Showing just the domain as an identifier
 
-Clients may treat the identifier `_@domain` as the "root" identifier, and choose to display it as just the `<domain>`. For example, if Bob owns `bob.com`, he may not want an identifier like `bob@bob.com` as that is redundant. Instead, Bob can use the identifier `_@bob.com` and expect Nostr clients to show and treat that as just `bob.com` for all purposes.
+Clients may treat the identifier `_@domain` as the "root" identifier, and choose to display it as just the `<domain>`. Similarly, clients should look up the identifer `<domain>` without `<local-part>@` under the identifier `_@domain`. For example, if Bob owns `bob.com`, he may not want an identifier like `bob@bob.com` as that is redundant. Instead, Bob can use the identifier `_@bob.com` and expect Nostr clients to show and treat that as just `bob.com` for all purposes.
 
 ### Reasoning for the `/.well-known/nostr.json?name=<local-part>` format
 


vitorpamplona:
@fiatjaf is going to hate this, but I do think this is how apps should behave anyway. There are way too many nip05 records without `_@` already. Clients can trigger an error or can just let it be. This revision makes it visible for implementers that need to be aware of.

fiatjaf:
Who is specifying these wrong NIP-05 records? I don't get, @vitorpamplona, if you know what the NIP says why do you go there and change your own name to  a nonstandard format?

fiatjaf:
Instead of asking all clients to change to use this new spec, why not just change Amethyst and Primal to adhere to the existing spec, which means less work for everybody?

And if you want users to not commit errors when typing their NIP-05, make some client-side validation for the `nip05` field to enforce the presence of a `@` or prefix domain with a `_@` when there is no `@`.

fiatjaf:
I just did a search on the biggest relays for `nip05` fields in kind0s that are naked domains and there are very very few -- there are many things that would be invalid no matter what (like some `npub` strings or other random stuff), some domains that are clearly from people that didn't understand what they were supposed to type (like full URLs with "HTTP" prefix).

I think people who own domain names will always be very few and literate enough in the domain of computers to be able to get their NIP-05 right. It would be very bad if we had to add these extra annoying and cumbersome 5 lines of parsing code to all client implementations just to accommodate for like 20 people who didn't know how to type.

Also, it's not clear what should their `/.well-known/nostr.json` endpoints are supposed to say if this is added to the spec, it would create other kinds of confusion and lead to more "flexibility" added to the spec in a couple of months, with more 5 or 10 or 15 lines of extra code that everybody has to worry about and implement.

fiatjaf:
With all that said I humbly ask that you stop validating this stuff on Amethyst.

alexgleason:
Clients should validate that the NIP-05 is correct before creating the event. They shouldn't just let you type anything into the box.

badonyx:
I'm inclined to agree with @fiatjaf 

> Clients should validate that the NIP-05 is correct before creating the event. They shouldn't just let you type anything into the box.

I just checked a few web clients. Nostrudel is the only one actually validating this.

vitorpamplona:
> Clients should validate that the NIP-05 is correct before creating the event. They shouldn't just let you type anything into the box.

I gave up trying to enforce that with other clients months ago. Clients will make all mistakes you might imagine and it's the receiving client to try to make sense of it all. 

All I can do is to control how Amethyst saves things and require that _@ . 

But that doesn't change the fact that other clients allow anything in that field and it is my job to make whatever comes in work. 

For instance, Amethyst does accept an npub instead of hexes in the e tag. We don't write it, but if it comes in, we make it work.

fiatjaf:
> it is my job to make whatever comes in work

That is only your job if you want to bloat the protocol or that you just don't care about breaking the protocol and all you care is about giving a 0.005% better UX for your users at the expense of everybody else.

fiatjaf:
I'll just close this.

fiatjaf:
Primal has agreed to stop validating naked domain entries in `nip05` and require the leading `_@`, by the way.

vitorpamplona:
> That is only your job if you want to bloat the protocol or that you just don't care about breaking the protocol and all you care is about giving a 0.005% better UX for your users at the expense of everybody else.

Better UX wins EVERY TIME. It doesn't matter how small it is.

alexgleason:
Here is my summary of this thread.

There is an adage: "be conservative in what you send and liberal in what you receive."

fiatjaf believes: "be conservative in what you send and conservative in what you receive."

Vitor believes: "be liberal in what you send and liberal in what you receive."

Of course, it depends on the circumstances. But I think maybe the original one was maybe actually the correct one. And anyway, you can do it, but just because you do it doesn't mean it should be in a NIP.

mikedilger:
I'm going to be conservative in what I receive because frankly I'm sick of Vitor's attitude that he is going to do what he wants and fuck the damage it does to the rest of us.  Fuck him this time.  I will not support his broken nip44 giftwraps.  He juggernauted ahead against our recommendations and then has the gall to get mad at us for not supporting his mess and bumping the version number.  Like I'm going to honor his mess.  I don't think we should clean up after him.  Someone has to teach this brat a lesson.  If the version number gets bumped for him, it makes gossip look bad one for not supporting one of the version numbers.

Ok yeah that was a pretty mean way to say it, but that's how I feel at this point. I'm really fucking sick of this brat always telling us his way or the highway.

mikedilger:
I apologize for my tone. I have been drinking.

jb55:
On Thu, Sep 28, 2023 at 04:25:08PM -0700, Vitor Pamplona wrote:
>> Clients should validate that the NIP-05 is correct before creating the event. They shouldn't just let you type anything into the box.
>
>For instance, Amethyst does accept an npub instead of hexes in the e tag. We don't write it, but if it comes in, we make it work.

I decided against this in Damus because it would proliferate a scenario
where things are valid in one client but broken in another. You should
really not encourage this. Being strict spec-wise when it comes to nip05
is pretty important.

PR #800, opened 2023-09-28
update nip-57 zap receipt spec to include 'a' tag

huumn:
@verbiricha [reached out letting us know SN's zapper wasn't including `a` tags in the zap receipt.](https://snort.social/e/nevent1qqsx2gxs8e6jk9mzwvwyuw6u2y863acewy2pqg0dn5hhzqed4rxj0asu7e8e8)

Figured it should be added to the spec for future zappers.

### DIFF
--- a/57.md
+++ b/57.md
@@ -128,7 +128,7 @@ The following should be true of the `zap receipt` event:
 
 - The `content` SHOULD be empty.
 - The `created_at` date SHOULD be set to the invoice `paid_at` date for idempotency.
-- `tags` MUST include the `p` tag AND optional `e` tag from the `zap request`.
+- `tags` MUST include the `p` tag AND optional `e` tag from the `zap request` AND optional `a` tag from the `zap request`.
 - The `zap receipt` MUST have a `bolt11` tag containing the description hash bolt11 invoice.
 - The `zap receipt` MUST contain a `description` tag which is the JSON-encoded invoice description.
 - `SHA256(description)` MUST match the description hash in the bolt11 invoice.


jb55:
On Thu, Sep 28, 2023 at 02:09:36PM -0700, Vitor Pamplona wrote:
>Merged #800 into master.

Since when is Vitor merging stuff without any discussion? I think a zap
targets are a terrible idea: people editing posts can make it look like
you zapped something that you didn't.

My zapper would not accept this and I don't plan on supporting it. Now
this spec is recommending something that would be spec-breaking for
many zappers who validate only e and p tags.

vitorpamplona:
Sorry, I thought this was a simple documentation of widespread behavior. If it is not, I am happy to revert.

jb55:
On Fri, Sep 29, 2023 at 09:03:07AM -0700, Vitor Pamplona wrote:
>Sorry, I thought this was a simple documentation of widespread
>behavior. If it is not, I am happy to revert.

probably best to revert, the spec currently requires there to be either
e+p to be set or just p, this defines the note and pubkey zap targets
for a zap request.

zappers should be validating these, adding an `a` tag would break
validation and would require a bit more care, ideally in v2.

I still feel like this is wrong. I think liking and zapping a-tags in an
antipattern and can lead to many issues:

1. It makes it look like someone likes a post that they didn't like
2. It makes it look like someone zapped a post that 

Imagine a scenario of a malicious user creating a post about how much
they like puppies, since alice likes puppies she zaps 10000 sats. Now to
troll, they use a replaceable event to update the article to say that
hitler was right and that jews need to be exterminated. Sorry for the
rough example, but this can really happen and is an example of how bad
the issue could be. If we have a-tag like and zap targets, people would
be shocked to see that you support such content.

This is why I was explicit about not supporting a-tags in zap requests.

Maybe it would make sense to add it in addition, but I don't think it
should be in-place of the e-tag so that clients can show that the zap is
invalidated/targetting an older post.

vitorpamplona:
We can definitely guide implementers to adding both `e` and `a`. But that's a separate PR for me. The goal of this one is just to document the option of the `a` without any mention of how it should be used.

PR #801, opened 2023-09-29
65: Normalize relay URIs

badonyx:
Update from discussion on #798. 

Does this need more explicit clarification is this sufficient?

### DIFF
--- a/65.md
+++ b/65.md
@@ -60,4 +60,4 @@ This NIP allows Clients to connect directly with the most up-to-date relay set f
 
 5. If a relay signals support for this NIP in their [NIP-11](11.md) document that means they're willing to accept kind 10002 events from a broad range of users, not only their paying customers or whitelisted group.
 
-6. Clients SHOULD trim trailing slashes (`/`) from relay URIs to canonicalize connections.
+6. Clients SHOULD deduplicate connections by normalizing relay URIs according to RFC 2616.
--- a/65.md
+++ b/65.md
@@ -60,4 +60,4 @@ This NIP allows Clients to connect directly with the most up-to-date relay set f
 
 5. If a relay signals support for this NIP in their [NIP-11](11.md) document that means they're willing to accept kind 10002 events from a broad range of users, not only their paying customers or whitelisted group.
 
-6. Clients SHOULD deduplicate connections by normalizing relay URIs according to RFC 2616.
+6. Clients SHOULD deduplicate connections by normalizing relay URIs according to [RFC 6455](https://www.rfc-editor.org/rfc/rfc6455.html#section-3).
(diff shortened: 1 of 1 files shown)


staab (approved):

jb55:
ACK

alexgleason (approved):

vitorpamplona:
RFC 6455, Section 3 doesn't seem to require the "/" at the end of the URI. Can we clarify this in the text?

badonyx:
@vitorpamplona I think it is not really important *how* they are normalized, just that they are normalized and deduplicated by the client.

vitorpamplona:
Right, the issue is the word `can` in that sentence. If we would like all clients to include the "/", we should copy and paste this and change to MUST. 

The second part is percent-encoding. The text says only the `#` that is not a fragment MUST be percent-encoded. Which is a little weird but we can make sure to only percent encode that specific char.

jb55:
On Fri, Sep 29, 2023 at 01:15:16PM -0700, Vitor Pamplona wrote:
>Right, the issue is the word `can` in that sentence. If we would like all clients to include the "/", we should copy and paste this and change to MUST.

I'm ok with stronger wording here. not doing this is a huge source of confusion.

mikedilger:
The relevant part of section 3 is this line, which if you trace it back allows empty paths or a single '/', so it's not "normalized" either way.

````
path = <path-abempty, defined in [[RFC3986], Section 3.3](https://www.rfc-editor.org/rfc/rfc3986#section-3.3)>
````

We should specify something more directly.

**Comment by badonyx** [comment]
Open to suggestions. Seems like [RFC 3986 (section 6)](https://datatracker.ietf.org/doc/html/rfc3986#section-6) is the standard reference here and sufficiently clarifies the normalization for the base path. 

```
6.2.3.  Scheme-Based Normalization

   The syntax and semantics of URIs vary from scheme to scheme, as
   described by the defining specification for each scheme.
   Implementations may use scheme-specific rules, at further processing
   cost, to reduce the probability of false negatives.  For example,
   because the "http" scheme makes use of an authority component, has a
   default port of "80", and defines an empty path to be equivalent to
   "/", the following four URIs are equivalent:

      http://example.com
      http://example.com/
      http://example.com:/
      http://example.com:80/

   In general, a URI that uses the generic syntax for authority with an
   empty path should be normalized to a path of "/".  Likewise, an
   explicit ":port", for which the port is empty or the default for the
   scheme, is equivalent to one where the port and its ":" delimiter are
   elided and thus should be removed by scheme-based normalization.  For
   example, the second URI above is the normal form for the "http"
   scheme.
```

For subpaths there is no standard normalization, as pointed out in #798.

mikedilger:
I like that RFC reference, it is better.  I don't think the term "deduplicate connections" is broad enough... it also affects records in storage.  Not sure exactly what to say though.  No biggie either.

Otherwise +1

badonyx:
I think it should be safe to merge now, better than leaving the existing statement at least.

mattn:
Thanks fixing this. Merging.

PR #802, opened 2023-09-29
make it clear that relays can keep multiple replaceable event versions

fiatjaf:


jb55:
On Fri, Sep 29, 2023 at 11:51:12AM -0700, fiatjaf_ wrote:
>@fiatjaf requested your review on: nostr-protocol/nips#802 make it clear that relays can keep multiple replaceable event versions.

Looks good to me, good to know nostrdb is allowed to be in-spec here
since nostrdb doesn't delete replaceable events which is pretty
important for restoring contact lists and things like that.

staab (approved):

fiatjaf:
It was always allowed in my view. The spirit of the spec should be that if you're not breaking other people you should be free to do whatever you want with your data, but this is a nice clarification to make it a little more obvious.

PR #804, opened 2023-09-30
rework NIP-03 so it is actually usable

fiatjaf:


03.md
```diff
@@ -4,20 +4,22 @@ NIP-03
...
+```json
 {
-  "id": ...,
-  "kind": ...,
-  ...,
-  ...,
-  "ots": <base64-encoded OTS file data>
+  "kind": 1040
+  "tags": [
+    ["e", <relay-url>, <relay-url>],
```
pablof7z, line 15: Event id

pablof7z:
Should mention NIP-33 event attestation; maybe with an `a` tag for lookup and `e` tag of the attested version

fiatjaf:
I think an `a` tag is not good here, since that may give people the wrong impression that the latest event of an `a` series is timestamped when actually only an old one is. If you have the `e` id you can fetch the actual event that was timestamped -- and if it was erased from all relays then the timestamp is useless anyway.

pablof7z:
yeah, fair enough, I was just thinking for the lookup, but you're right that even the lookup case would be idiotic if you don't have the right `e`

staab (approved):

vitorpamplona:
This is much better than the `ots` tag.

jb55:
@fiatjaf it's not clear what exactly we're supposed to commit to in the ots file?

jb55:
nm I can't read:

> The OpenTimestamps proof MUST prove the referenced `e` event id as its digest

PR #813, opened 2023-10-10
Add "s" tag to reactions

arthurfranca:
One `s` tag is set to reactions with "+" (positive or unknown sentiment, in case of some emojis) or "-" (negative sentiment) value.
This allows for using NIP-45 to count likes and dislikes separately.

### DIFF
--- a/25.md
+++ b/25.md
@@ -30,9 +30,11 @@ reacting to. This allows users to be notified of reactions to posts they were
 mentioned in. Including the `e` tags enables clients to pull all the reactions
 associated with individual posts or all the posts in a thread.
 
-The last `e` tag MUST be the `id` of the note that is being reacted to. 
+The last `e` tag MUST be the `id` of the note that is being reacted to.
 
 The last `p` tag MUST be the `pubkey` of the event being reacted to.
+
+One `s` tag MUST be added with either '+' value for positive/unknown sentiment or '-' for negative sentiment.
 
 The reaction event MAY include a `k` tag with the stringified kind number
 of the reacted event as its value.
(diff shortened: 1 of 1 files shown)


mattn:
The client behaviour is that if there is not s-tag, the + or - contained in the content is used; if there is s-tag, the content is ignored. Right?

arthurfranca:
@mattn the idea is to make the `s` tag a requirement so, giving enough time for clients to update theit code, it would always be there.

The `.content` should be used both for backwards compatibility and for using custom emoji. The `s` tag is just for filtering by "+" or "-" because we can't filter on relays by `.content`.

vitorpamplona:
Maybe `s` should have the same content as the `.content`. In that way, it covers all other potential use cases of this and later we can deprecate the use of `.content` to store the reaction type.

arthurfranca:
@vitorpamplona I thought of that but how would you count emoji reactions if you don't know which emojis were used?

That's why i limited `s` tag to + and - and defaulted emojis to + if the client doesn't know if the emoji is a negative or positive one.

arthurfranca:
I was worried about clients that have just likes and dislikes such as reddit which need to know if the reaction is positive or negative.

staab:
Reactions are too simple a concept to handle sentiment, they're already overloaded and therefore ambiguous. There are numerous ways to handle the concept, including:

- Arbitrary reactions (kind 7)
- Booleans
- Magnitude (0 to 1 or -1 to 1)
- Points (arbitrary number, like medium's claps)

There are also other dimensions like those mentioned in NIP 32, like "certainty", which means "I vouch for the quality of this but I'm not entirely sure in my assessment".

It would make sense to add a new kind called "vote" which uses booleans, and a "rating" which uses magnitude (this already exists via nip 32). This would separate out the different concerns. Clients could query only the kinds that make sense based on their UI, or all of them and attempt to fit them into their paradigm.

arthurfranca:
@staab Imo positive and negative are enough for many use cases/UIs despite its simplicity:
- count just likes or any reactions
- count likes and dislikes separately or tally
- 1 to 5 stars using an average (like means 5, dislike means 1) a platform updates from time to time

Using the same kind for as many use cases may be good in regard to lowering the chances of an event end up with no or very few ratings.

A boolean "good" or "bad" is so cheap that users may be inclined to add a rating. It is just a tap away.

That's why I think emojis should be just a presentation layer on top of a boolean type of reaction instead of a separate event kind.

mikedilger:
Create a library that maps emoji to sentiment.  An eggplant maps to neutral.  A heart to positive.  A mad face to negative.

It's too late to degrade reactions (widely variable) into a simple sentiment (far less information) at the protocol level.

arthurfranca:
Currently my NIP-45 COUNT code is
```
const count = await getEventCount({ relayUrls, filter }).catch(() => {})
if (count !== undefined) return { '+': count, '-': 0, tally: count }
```
Can't separate likes and dislikes. If we don't agree on this I will probably just remove dislikes and call it a day

PR #814, opened 2023-10-10
WIP: 107 Nostr IoT

arcbtc:
The terms/conditions of IoT software/hardware is horrible and Orwellian. Nostr gives the simple, permissionless development environment IoT needs so it can move away from its current paradigm.  A user should be able to message a light,  heating, speaker and ask it to turn on/off.


vitorpamplona:
I would not do a JSON in a NIP-04.

What about a new event kind? 

```js
{
  "kind": 30107,
  "created_at": 1675642635,
  "content": "",
  "tags": [
    ["d", "<String, UUID of the stall to which this device belong to>"],
    ["name", "<String, device name>"],
    ["description", "<String (optional), device description>"],
    ["t", "<category 1>"],
    ["t", "<category 2>"],
    ["t", "<category 3>"],
    ["p", "<User 1 pubkey in hex>"],
    ["p", "<User 2 pubkey in hex>"],
    ["image", "<URLs>"],
  ],
  "pubkey": "...",
  "id": "..."
}
```

fiatjaf:
I think it would be nice to have a kinds for generic actions often performed by IoT devices -- which to me seem to be basically notifications of intents or information.

Would be nice to gather a list of things IoT devices often do and standardize the most common cases -- such that there is interoperability.

For example:

3400: on/off -> an event stating that a device is turning on or off
3401: temperature -> an event containing the temperature
3402: lighting -> an event containing the intensity of the lighting detected
3403: movement -> etc
3404: voice -> etc

For example, if a device gathers temperature data it can emit an event of kind 3401 with temperature 39°C, then another device can be configured to react to temperature events from a fixed list of public keys or anything that comes from a specific relay or anything that comes from a public key attested with a NIP-32 event from someone trustworthy and so on, and then turn itself on or off.

Since these events are interoperable it becomes very easy to plug in new devices to the scheme, replace devices and so on -- but it also allows for nicer ideas not yet invented, like gathering the temperature in the entire world from these public events (?) or maybe doing some silly experiments, I don't know.

isaac-asdf:
I think there are too many possible IoT type devices/commands to
encapsulate each as a separate kind number. If anything, we could have one
kind for a sensor and a separate for a controller. I think tags seems like
the place to put things like units, “kind” of input for sensors (temp vs
pressure) and geohash data

erskingardner:
I agree that this should be a new kind (or kinds), with an encrypted content field so that you can keep the commands/data private. 

ACK to @isaac-asdf idea on kinds for sensor & controller but not a new kind for every conceivable type of data or command.

107.md
```diff
@@ -0,0 +1,73 @@
...
+## Nostr IoT Clients
+
+### User
+
+Where the `user` registers 'device'(s) and its keys, then updates the `device`(s).
+
+### Device
+
+The `device` uses a client like <a href="https://github.com/lnbits/arduino-nostr">nostr-arduino</a> to receive commands from the `user` or another `device`. 
+The `devie` can also push data such as senssor readings and updates.
```
motorina0, line 30:
```suggestion
The `device` can also push data such as sensor readings and updates.
```

107.md
```diff
@@ -0,0 +1,73 @@
...
+| `4    ` | `direct_message` | Communicate between `device`(s) and `user`(s). The messages can be plain-text or JSON. | [NIP04](https://github.com/nostr-protocol/nips/blob/master/04.md) |
+
+
+### Event: Register a `device`
+
+**Event Content**:
+```json
+{
+    "id": <String, UUID generated by the user.Sequential IDs (`0`, `1`, `2`...) are discouraged>,
+    "device_id": <String, UUID of the stall to which this device belong to>,
```
motorina0, line 46:
```suggestion
    "device_id": <String, UUID of the device>,
```

107.md
```diff
@@ -0,0 +1,73 @@
...
+    "name": <String, device name>,
+    "description": <String (optional), device description>,
+    "categories":[ <String (optional), device category, such as 'boiler'>],
+    "users":[ [<String (optional), user public-key>]],
+    "images": <[String], array of image URLs, optional>
```
motorina0, line 51: - is `"images"` a leftover?
```suggestion
```

107.md
```diff
@@ -0,0 +1,73 @@
...
+Note:
+ - `users`:
+   - A list that can restrict access to public-keys, `device`(s) also has keys and can be a `user`.
+
+### Event: Update a `device`
+
+**Event Content**:
+```json
+{
+    "id": <String, UUID generated by the merchant.Sequential IDs (`0`, `1`, `2`...) are discouraged>,
```
motorina0, line 64:
```suggestion
      "id": <String, UUID generated by the user. Sequential IDs (`0`, `1`, `2`...) are discouraged>,
```
- not sure if this `"id"` will be used in any way

motorina0:
### Basic IoT usecase
I'm adding this diagram as a common ground for discussion

iot drawio

- `Thermostat` and `Controller` are `devices` that listen for and publish events
- each `device` has its own key-pair
- a `device`'s software can be flashed (installed) and **configured** via a web-installer (see https://bitcoinswitch.lnbits.com/). The configuration:
  - MUST include at least the `private key` for the device
  - SHOULD include a list of accepted `public keys` from which it accepts commands (so your neighbour cannot stop the heating during the night)
- a `device` can **publish**
  - its state at regular intervals or when a change occurs (eg: publish temperature every 5 minutes and/or when the value changes)
  - a command as part of its state changing (eg: thermostat sends on/off command)
-  a `device` can **subscribe** for events that change its config (eg: update settings, turn on/off)

**Notes**:
  - this is just one of the many possible flows. An alternative is that the `Thermostat` only publishes its temperature and the `Controller` manages the settings (what the state should be based on current date and temperature)
  - the `user` can also manually turn on/off the heater
    - in this case the controller should accept events from two public keys (user and thermostat)

motorina0:
Is there an IoT expert around that can point us to some data structure standards?
All I could find is this (which is as ugly as a giant BIP) https://docs.oasis-open.org/mqtt/mqtt/v5.0/mqtt-v5.0.html

fiatjaf:
> I think there are too many possible IoT type devices/commands to
encapsulate each as a separate kind number. If anything, we could have one
kind for a sensor and a separate for a controller. I think tags seems like
the place to put things like units, “kind” of input for sensors (temp vs
pressure) and geohash data

I don't think this is true. Most use cases can probably be covered by standardized kinds, and we can add more kinds as necessary.

If this was true, though, then there would be little point in using Nostr.

motorina0:
> I don't think this is true. Most use cases can probably be covered by standardized kinds, and we can add more kinds as necessary.

- a tag should solve this issue
- kinds are kind of scarce (since they are ints in a given interval)

fiatjaf:
Tags give the idea that we don't need interoperability and anyone can write anything there, which is not what we want (again, if everybody does that then Nostr is useless). We need standardization and interoperability, kinds are much more efficient for that (it's better for relays and indexing and uses less space and everything is faster if we have a different kind for each thing).

Given that we _want_ interoperability and standardization, 1000 kind numbers seem more than enough. If we have more than that we will not be able to coordinate the standards anyway.

motorina0:
@vitorpamplona regarding:
> I would not do a JSON in a NIP-04.
> What about a new event kind?

- I think a replaceable parametrized event (like `30107`) makes sense for changing the state of a device (settings, on/off, etc)
- but in some cases the user might want to keep the event history (which is lost for `30xxx` events). 
  - **eg**: I want to see how much my solar panels produced in a time interval (at what hour they produce more so I can start more appliances), which month of the year is best, etc.

vitorpamplona:
Sure, if you want to keep the history, regular events make sense. You can even use an unbound list to group them by streams https://github.com/nostr-protocol/nips/pull/784

blackcoffeexbt:
I like both approaches, a single kind and multiple and see arguments for both. Whilst kinds are scarceish, I do agree with @fiatjaf that a per sensor type kind is probably going to be most useful for iot devices and relay filtering. 

Being able to REQ and respond to one kind for a sensor type is very useful for low memory and bandwidth iot devices. (e.g. esp8266).

Can we block out an int range for kinds for this nip? E.g. 30107 to 31106?

fiatjaf:
Yes, let's block a thousand kinds for this and start listing and specifying the most common formats.

Charlie67j:
I'm curious what the intended usage case might be or what problem this solves that isn't already solved by MQTT and the Sparkplug B standard? Many MQTT servers are bridgeable to relay data between them and it's simple to do. There are FOSS options and encryption is quite simple to accomplish for data in flight. Is this intended as an on/off ramp to allow interacting with an IoT device through relays? If so, why not leverage MQTT and make a client bridge wrapper for it? How is this better/different than the current self-hosted and FOSS options? Not intending to be a jerk here, just curious if I'm missing the point.

fiatjaf:
@Charlie67j where are the specs for the MQTT thing?

I think Nostr adds the public key infrastructure needed to make this a really open thing, while MQTT-based stuff must necessarily be closed. Also Nostr relays are much easier to setup and they are already being used for other things, which makes integration seamless. It also makes a lot of sense.

Let me reverse the question: if the MQTT solution works for IoT, why can't we ditch Nostr entirely and just use MQTT for social networking?

fiatjaf:
I've opened this new PR incorporating @blackcoffeexbt's ideas and how I'm seeing this: https://github.com/nostr-protocol/nips/pull/817

shaibearary:
> Is there an IoT expert around that can point us to some data structure standards? All I could find is this (which is as ugly as a giant BIP) https://docs.oasis-open.org/mqtt/mqtt/v5.0/mqtt-v5.0.html

i dont think there is such thing and i create sdk one by one when there is a new type of sensor coming with different format and collect them all to same format for visualization/further analysis

shaibearary:
> I'm curious what the intended usage case might be or what problem this solves that isn't already solved by MQTT and the Sparkplug B standard? Many MQTT servers are bridgeable to relay data between them and it's simple to do. There are FOSS options and encryption is quite simple to accomplish for data in flight. Is this intended as an on/off ramp to allow interacting with an IoT device through relays? If so, why not leverage MQTT and make a client bridge wrapper for it? How is this better/different than the current self-hosted and FOSS options? Not intending to be a jerk here, just curious if I'm missing the point.
for mqtt you need a mqtt server
mqtt is a pub-sub protocol. server/client dont store data and dont accept time filter or tag filter, it filter by topic and some legacy authentation instead of sign-verify. think about it when you subscribe a data with sensor id, you cannot know if its published by sensor or just a fake data. same when you write back to sensor to adjust setting
https://mqtt.org/mqtt-specification/

Charlie67j:
> @Charlie67j where are the specs for the MQTT thing?
> 
> I think Nostr adds the public key infrastructure needed to make this a really open thing, while MQTT-based stuff must necessarily be closed. Also Nostr relays are much easier to setup and they are already being used for other things, which makes integration seamless. It also makes a lot of sense.
> 
> Let me reverse the question: if the MQTT solution works for IoT, why can't we ditch Nostr entirely and just use MQTT for social networking?



> @Charlie67j where are the specs for the MQTT thing?
> 
> I think Nostr adds the public key infrastructure needed to make this a really open thing, while MQTT-based stuff must necessarily be closed. Also Nostr relays are much easier to setup and they are already being used for other things, which makes integration seamless. It also makes a lot of sense.
> 
> Let me reverse the question: if the MQTT solution works for IoT, why can't we ditch Nostr entirely and just use MQTT for social networking?

An MQTT server/broker is analogous to a nostr relay in many ways. In fact, when I first learned of nostr, I was curious if the relay was leveraging MQTT until I learned differently.  Both are trivial to set up from a server perspective and are very light weight. It feels like the intended difference here would be the methods of identification and verification of the origin authenticity of the message.  Perhaps I'm missing the point entirely why one would want to send sensor or other Iot, IIot, data over nostr relays. In my experience this is can be allot of data, and often.  Are we talking public relays here? How is that better than MQTT and why would I want to have my sensor data clogging up nostr relays. If I own the relay, sure, but then why is nostr better than MQTT which is well developed at the device and application layer with millions of people and companies using it every day?  Again, not wanting to sound like a jerk in any way, just genuinely curious where this is going and what problems this solves that are unsolved today or in the future. I love things that solve my future problems. Maybe the use case is some distributed oracle that requires a high level of trust?

Charlie67j:
> > I'm curious what the intended usage case might be or what problem this solves that isn't already solved by MQTT and the Sparkplug B standard? Many MQTT servers are bridgeable to relay data between them and it's simple to do. There are FOSS options and encryption is quite simple to accomplish for data in flight. Is this intended as an on/off ramp to allow interacting with an IoT device through relays? If so, why not leverage MQTT and make a client bridge wrapper for it? How is this better/different than the current self-hosted and FOSS options? Not intending to be a jerk here, just curious if I'm missing the point.
> 
> for mqtt you need a mqtt server mqtt is a pub-sub protocol. server/client dont store data and dont accept time filter or tag filter, it filter by topic and some legacy authentation instead of sign-verify. think about it when you subscribe a data with sensor id, you cannot know if its published by sensor or just a fake data. same when you write back to sensor to adjust setting https://mqtt.org/mqtt-specification/

I get your point. The MQTT server/broker is not intended to store the data just the "state". There are QoS options and the Sparkplug standard introduced things like birth/death certificate and devised the last will and testament to determine when a value is stale, etc. Those are all considerations that would be nice to have in a nostr IoT solution as well. It would be nice to authenticate the source of the data, but you also need to verify that the last data is truly that latest data.

fiatjaf:
I see. I've learned a bit about MQTT now. It looks good.

Just to clarify: what is the flow for implementing MQTT support, for example, if you want to bridge some devices in your home? You will first run an MQTT server (locally, for example), then you hardcode your devices to pub or sub to that server, then you send some data in arbitrary format from one device and read it from the other device?

shaibearary:
> > > I'm curious what the intended usage case might be or what problem this solves that isn't already solved by MQTT and the Sparkplug B standard? Many MQTT servers are bridgeable to relay data between them and it's simple to do. There are FOSS options and encryption is quite simple to accomplish for data in flight. Is this intended as an on/off ramp to allow interacting with an IoT device through relays? If so, why not leverage MQTT and make a client bridge wrapper for it? How is this better/different than the current self-hosted and FOSS options? Not intending to be a jerk here, just curious if I'm missing the point.
> > 
> > 
> > for mqtt you need a mqtt server mqtt is a pub-sub protocol. server/client dont store data and dont accept time filter or tag filter, it filter by topic and some legacy authentation instead of sign-verify. think about it when you subscribe a data with sensor id, you cannot know if its published by sensor or just a fake data. same when you write back to sensor to adjust setting https://mqtt.org/mqtt-specification/
> 
> I get your point. The MQTT server/broker is not intended to store the data just the "state". There are QoS options and the Sparkplug standard introduced things like birth/death certificate and devised the last will and testament to determine when a value is stale, etc. Those are all considerations that would be nice to have in a nostr IoT solution as well. It would be nice to authenticate the source of the data, but you also need to verify that the last data is truly that latest data.

qos is not suitable for relay imo.

shaibearary:
> I see. I've learned a bit about MQTT now. It looks good.
> 
> Just to clarify: what is the flow for implementing MQTT support, for example, if you want to bridge some devices in your home? You will first run an MQTT server (locally, for example), then you hardcode your devices to pub or sub to that server, then you send some data in arbitrary format from one device and read it from the other device?

yup

Charlie67j:
> I see. I've learned a bit about MQTT now. It looks good.
> 
> Just to clarify: what is the flow for implementing MQTT support, for example, if you want to bridge some devices in your home? You will first run an MQTT server (locally, for example), then you hardcode your devices to pub or sub to that server, then you send some data in arbitrary format from one device and read it from the other device?

That's correct in general terms. I run several MQTT servers for different purposes. Some for myself, some for my company and some for customers. The kind of data that we expect from IoT devices isn't useful in the same way that a stored message on a relay is useful. IoT/IIoT while using lightweight protocols can quickly become a large amount of data. The typical usage scenario would be for an IoT device (client) to publish a value(s) to the MQTT server on change of state, this could be a tenth of a degree temperature change, which might happen several times a second for a noisy sensor. Another client will subscribe to this "topic" for display and if you choose so, to store the state in a database of some kind for later analysis. In this way the good data is saved and the  relay/broker/server stays lightweight. 

I can see relay operators being inundated with large amounts of data that they'll need to decide what to do with. Perhaps filter and drop certain types of data? I have several remote sites that produce gigabytes of data every month with a relatively few number of sensors. Since the MQTT server doesn't store this data, managing it is not something that we need to consider, rather storing the data in a useful way outside the network or not at all is the decision of the user and the subscribing client.  

For bridging specifically, you need two or more MQTT servers. One or more server is configured as a bridge which will allow it to subscribe to one, many, or all topics on another server. This is not unlike syncing between relays. Same general concept except that there is no history to sync. Bridging is a good way for edge of network devices and sites to 

I'm not saying that the ability to publish events to a relay from IoT devices might not be useful, there are many other considerations that need to be worked through. Human to Human, Human to Machine, and Machine to Machine data types have different requirements and usage scenarios.

arcbtc:
There seems to be some confusion.
The motivation of this nip and what makes it great is its an easy way to not invite a bunch of third parties into your house when using IOT. 
Sure, you can run an MQTT server, but most wont do that. 
The simplicity of nostr is what makes it powerful. I can DM a light, or my solar batteries can dm me info without running complicated infrastructure. Clients can be built for all the different use cases in a permisionless environment. 
Nostr is of course useful for public IoT, but for private it solves one of the greatest threats to humanity, us all installing backdoors into our living spaces.

arcbtc:
sensor data as event kinds doesnt make sense, it must ba an array of tags, with some shared/common values.
Many devices will send multiple values.

blackcoffeexbt:
> sensor data as event kinds doesnt make sense, it must ba an array of tags, with some shared/common values. Many devices will send multiple values.

Where an event contains data from multiple sensor types, can you give an example of how the sensor data would be represented in the content property and how the tags for the event would be listed? An unabbreviated JSON structure for the event would be useful.

blackcoffeexbt:
@arcbtc  be aware that sensor type (temp, pressure etc) in the tags property is a metadata leak. Your requirement for the nip is for private IoT comms. How should this data be communicated in order to avoid this leak?

107.md
```diff
@@ -0,0 +1,72 @@
+NIP-107
+======
+
+Nostr Internet of Things 
+-----------------------------------
+
+`draft` `optional` `author:benarc` `author:blackcoffeebtc`
+
+## Ratioanale
```
motorina0, line 9:
```suggestion
## Rationale
```

motorina0:
> @arcbtc giftwrapping could be used to protect metadata but it will remove the ability to filter on tags before decrypting the event.

I think we're working with the `social-network` mindset in mind here (and we should not). 
Filtering events by `tags` and `kinds` is not a requirement (or useful).

**Scenario**:
 - the `user` has a mobile/web app for monitoring & controlling its IoT devices
 - the app has the `pubkeys` for all of the `user`'s devices
 - the app queries for (NIP04) events based on those `pubkeys` **alone**
 - the app aggregates the data in a nice format (tables, charts).
 - the app caches the the current stated and upgrade the UI based on new received DMs

Tag like fields (eg: `categories: ["boiler"]`) can be part of the JSON which is inside the DM.

v0l:
Please add data-type for radiation as cpm

v0l:
I have started to publish my weather station data on kind 8001, you can see it here:
`nak req -k 8001 -l 10 wss://relay.snort.social`

PR #819, opened 2023-10-12
Nip 06 test vectors

mikeobank:
Add test vectors for deriving from mnemonic seed words (12 and 24). This should be triple checked before merging!

Possible changes / additions:
- Markdown layout
- mnemonic + passphrase vector
- account > 0 vector


vitorpamplona:
ACK. Thank you for the test cases.

PR #821, opened 2023-10-13
Add Simple Moderated Group Chats

water783:
The sealed DM of [Sealed-DMs](https://github.com/vitorpamplona/nips/blob/sealed-dms/24.md) @vitorpamplona @staab  provides a simple and effective group chat mechanism, but it does not offer Group metadata & Moderated functionality. I aim to create a group chat based on Sealed-DMs that can supports Moderation and allows group members to be visible to one another

Any thoughts on this?
[NIP-102: Simple Moderated Group](https://github.com/water783/nips/blob/group-chats/simple-moderated-group.md)


staab:
See #580 and #706 for other approaches. I personally think with gift wrapping it's better to re-purpose existing standards like NIP 28 by wrapping them than to re-invent everything in an encrypted form. I invite you to collaborate with me on #706.

water783:
> See #580 and #706 for other approaches. I personally think with gift wrapping it's better to re-purpose existing standards like NIP 28 by wrapping them than to re-invent everything in an encrypted form. I invite you to collaborate with me on #706.

Yes, I've read NIP-87 & NIP-112. Using a shared key for large groups makes sense. However, I tried implementing the shared key approach based on Nostr, and it's too complex, and comes with many issues.

So, I think we should first adopt the method of sending messages to each group members like in a sealed group, suitable for smaller groups. It can support groups of less than 1,000 people.

water783:
Seems like wrapping NIP-28 events is also a good idea anyway. :)

vitorpamplona:
I would try to create the simplest proposal that can work first and then evolve over time.

Making things extremely simple (maybe even too simplistic) leads to better adoption by other clients. Once there is adoption, there is learning from what users want and when that happens evolution becomes a natural process.

water783:
Yeah, so far, I believe that the sealed group seems to be the most appropriate way to implement the nostr private group. 

I think I can try combining nip28 with the sealed group approach to create a simple moderated small group chat.

PR #822, opened 2023-10-14
Update NIP-65 add prioritization (by trust) of list guidelines

frnandu:
The reasoning behind this is that by running your own personal relay, you can have the highest degree of certainty that it won't censor any of your events when other people following you request them. The list should reflect the user's level of trust in the relays regarding censorship.

Clients should allow users to change the order of the relays in the list to create a sorted list based on their personal trust in the relays. For non-personal relays, trust is determined by the free market of opinions and reviews.

If a user notices, for example, a lot of reports from their trusted contacts that a certain relay is censoring events for arbitrary reasons, they can go to their list and move that relay down the list or remove it entirely.


mikedilger:
I'm not sure. It's not very common that people have good knowledge about which relays are more reliable.  Automated determination of that will be better. The gossip client uses a lot of logic to work that out and pick the better relays.

I don't want to get in the way of this idea though.  But the problem with this suggestion in particular is that presuming the ordering has this meaning requires everybody to now go check and change the order of their relay lists.  I think it would be better to add an optional marker.

fabianfabian:
I have yet to implement NIP-65 and the outbox model, so I may not fully understand why things are the way they are now, but one thing that stopped me last time when I tried to make an attempt was something missing that defines a relay as authoritative. 

The way I wanted to do things was just use a standard relay set that many clients use today, and combine it with fetching from the authoritative relay(s) of a user. I don't need to know **all** the write relays of a user, I just need to which is the one where I have to most certainty of getting their events.

Seems like this NIP update could help it a bit by using priority, but I'm not sure if that is the best way, maybe just "read" "write" and "authoritative".

frnandu:
Thanks Mike for the feedback.
Can you explain a bit more in detail the logic your gossip client uses to pick the best relays? Is it based on anything related to trust by the owner of the list?

The fact that most people don't have knowledge about how much they trust a relay or even want to do it, shouldn't eliminate the possibility for those who know and want to do it.

Not everybody will need to sort their list, but if clients tend to respect the order of the list, then the few people that might need this and do sort the list (like Snowden or other people highly probable to be censored) will have a better chance of their events being fetched from more trusted (by them) relays without being censored.

More popular relays which might be chosen by the client for the fact that they're just more popular will lead to a more centralized network, leading to a bigger incentive for outside forces to try to force them to censor certain accounts.
 
My argument against a marker (authoritative like fabian suggested) is that you might want to have several degrees of trust, like your personal relay at the top, then some payed relay(s) you consider highly trustworthy, and only then general free and less reliable relays at the bottom.

leo-lox:
your suggestions is something like this?
where (int) higher means more trust, clients could then normalize this and rank based on the user provided trust score
```{
  "kind": 10002,
  "tags": [
    ["r", "wss://my-own.example.com", "", 3 ],
    ["r", "wss://less-trust.example.com","", 2 ],
    ["r", "wss://even-less-trust.example.com", "write", 1 ],
    ["r", "wss://nostr-relay.example.com", "read" ],
  ],
  "content": "",```

frnandu:
No I think the order in the list is enough to give clients a clue where should they go for the most trustworthy source of events for this user.

leo-lox:
the drawback would be that you cant rank them equal, or unranked like @mikedilger mentioned

frnandu:
Whatever the client does with the additional information of user's preferred trust order of relays is up to the client.
You can have clients that don't really give a s*** and just rank the relays according to popularity among contacts, or you can respect that user's preference and go for the ones it choose to be higher trusted.

mikedilger:
There is a danger that if we have a ranking, or if we mark a relay as "authorative", then a censor or attacker will target the highest ranked relay, and the clients that utilize this ranking or marking will potentially fail because they failed to use decentralized redundancy logic in the first place.  I don't want to order my relays, and if I could magically scramble the order every time someone looked up my relay list I would.

frnandu:
You already have this danger that a censor or attacker going for the most popular relays. And since you can't have an unlimited amount of relays that you connect to for getting the feed of your connections it's only natural that clients try to optimize and reducing the relays it connects to for the ones that cover most connections. This results in you having not really any way of making sure people get content from you from your prefered or personal relay (which has the least probability of being censored).
Scrambling is not realisticaly doable since it would result in an enormous amount of relays that you would need to connect for connections feed. Unless you mean for usecases where you have one or few pubkeys in the filter.
I don't see the danger of me sorting my relays resulting in this being more attackable. Actually if I put my personal relay that I control at the top results in the least censor probability.

frnandu:
> The order could be a hint of preference by the author. But I don't think we should create an expectation that the order will be followed. Each client should decide how to best use the list.

Considering an implemented outbox/gossip model where you fetch notes of your contacts (feed) from their published list of preferred relays, you kind-of need some incentive in the NIP for clients on the "other-side" (people that follow you) to respect what you published as being your most trusted relay for source of your notes (personal relay or least probability of being censored).

Semisol (changes requested):
If a relay is censoring you, remove it.

frnandu:
> If a relay is censoring you, remove it.

You might not immediately know if a relay is censoring you or not.
You followers might be fetching your notes from some relay that censors you, and they will not know or notice some missing events.
Remember that when using gossip/outbox model, clients will not go fetch notes from every possible relay in the list, they will probably only go fetch from the most popular ones among the contacts in order to reduce open sockets, data and battery usage, etc.
If one of those is attacked for censoring, it will just go by unnoticed.

And you can have some relays in your list for better distribution of your notes that today might not censor but one day in the future might start.

The order for the list is to reflect your subjective opinion on the probability that something like that might or not happen per relay.
So you would put your personal relay that you control at the top, then some that you pay for and have a big trust it will not go bad, and later free ones.

frnandu:
> The problem is you may not have those other relays in your relay list at all then.
> 
> Why should a client pick any of those lower ranked relays if the client is told to ignore lower ranked relays because they might censor? So those basically do nothing.

Each client can do whatever it wants with the information on the relay list, depending on it's own strategy, or user settings, or which already connected relays, or data saving preferences, etc.

My issue is that if a certain client DOES want to prioritize according to what the author of the nip65 prefers it just can't because it lacks that information.

frnandu:
If a client is looking to save data but gives the user a setting for max-relays-per-follower on an outbox/gossip strategy in order to not have an explosion of amount of relays it has to fetch data from for constructing the feed, then which ones from each follower nip65's relay list will it choose to use?
And here you can have few strategies:
1) use the most popular ones among all followers (which will mean connecting to the least amount of relays possible and saving most on data usage)
2) random ( which helps decentralization out of popular relays)
3) taking into account each follower's preference on the order of relays are higher on their trust/priority level

This PR is just to give a bit more information so that if clients what to somehow CUT the amount of each follower's relays from which to fetch notes from, they can have that option of honoring the follower's preference (priority).

mikedilger:
Clients would have to maintain the order and update their UI/UX to handle rearranging the order, even though few clients will bother to use that order.  To me this is too much work for too little benefit.  It would be far easier to have some other method of indicating that a relay was more reliable, like a marker, than to deal with sort/rearrange in the UI.

mikedilger:
Well @semisol I run chorus.mikedilger.com.  But I also use several relays that I don't run and I don't even know who runs them, but I use them for redundancy.  If I took your advice, I would have a very short relay list.

I don't mind having some indication of reliability even if few need to use it.  But I agree with you that eventually with negentropy and such we will get the best reliability by querying all of the user's relays.  Eventually.

PR #823, opened 2023-10-15
Update 90.md

starbackr-dev:
Added optional Encryption to params and payload if the user require the input and output to be a secret.

### DIFF
--- a/90.md
+++ b/90.md
@@ -91,7 +91,7 @@ This param data will be encrypted and added to the `content` field and `p` tag s
 "content": "BE2Y4xvS6HIY7TozIgbEl3sAHkdZoXyLRRkZv4fLPh3R7LtviLKAJM5qpkC7D6VtMbgIt4iNcMpLtpo...",
     "tags": [
         [`p`, `04f74530a6ede6b24731b976b8e78fb449ea61f40ff10e3d869a3030c4edc91f`],
-        [ `param`, `encrypted`, true]
+        [`encrypted`]
     ]
 
 
@@ -124,6 +124,23 @@ Service providers publish job results, providing the output of the job result. T
 ## Encrypted Output
 
 If the request has encrypted params, then output should be encrypted and placed in  `content` field with `p` tag. If the output is encrypted, then avoid including `i` tag with input-data as clear text. 
+Add a tag encrypted to mark the output content as `encrypted`
+```json
+{
+    "pubkey": "<service-provider pubkey>",
+    "content": "<encrypted payload>",
+    "kind": 6xxx,
+    "tags": [
+        [ "request", "<job-request>" ],
+        [ "e", "<job-request-id>", "<relay-hint>" ],
+        [ "p", "<customer's-pubkey>" ],
+        [ "amount", "requested-payment-amount", "<optional-bolt11>" ],
+        [`encrypted`]
+
+    ]
+}
+```
+ 
 
 ## Job feedback
 Service providers can give feedback about a job back to the customer.
--- a/90.md
+++ b/90.md
@@ -79,8 +79,8 @@ If the user wants to keep the input parameters a secret, they can encrypt the `i
     [ "param", "temperature", "0.5" ],
     [ "param", "top-k", "50" ],
     [ "param", "top-p", "0.7" ],
-    [ "param", "frequency_penalty", "1" ],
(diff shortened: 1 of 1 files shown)


90.md
```diff
@@ -67,6 +67,37 @@ All tags are optional.
 * `relays`: List of relays where Service Providers SHOULD publish responses to
 * `p`: Service Providers the customer is interested in. Other SPs MIGHT still choose to process the job
 
+## Encrypted Params 
+
+If the user wants to keep the input parameters a secret, they can encrypt the `i` and `param` tags with the service provider's 'p' tag and add it to the content field. The user also indicates whether the output should be encrypted or not as one of the parameters.
+
+```json
+[
+    [ "i", "what is the capital of France? ", "prompt" ],
```
cj-chua, line 76: this should be `text` instead of `prompt`?

90.md
```diff
@@ -67,6 +67,37 @@ All tags are optional.
...
+    [ "param", "temperature", "0.5" ],
+    [ "param", "top-k", "50" ],
+    [ "param", "top-p", "0.7" ],
+    [ "param", "frequency_penalty", "1" ],
+    [ "param", "output", "encrypted"]
+]
+
+```
+
+This param data will be encrypted and added to the `content` field and `p` tag should be present
```
cj-chua, line 88: How about defining a new input type in addition to the existing ones(`text`, `job`, `event`, `url`) ?

The protocol currently defines the input tag to be`[ "i", <data>, <type>, <optional relay>, <marker> ]`. Perhaps introduce an new type called `encrypted`. i.e.
```
[ "i", <data>, <type>, <optional relay>, <marker> ]
[ "i", BE2Y4xvS6HI..., "encrypted", "04f74530a6ede6b24731b976b8e78fb449ea61f40ff10e3d869a3030c4edc91f" ]

... where in this example
data = the encrypted content
type = encrypted
relay = omitted
marker = SP's pubkey
```

cj-chua (commented):
Left a comment/idea that would prevent this proposal from steering too far away from the event/job structure defined in NIP90. 

As for the proposal itself I don't know enough to say if this is the better way to introduce encrypted input/output (as compared to gift wraps). Will let others comment on that

90.md
starbackr-dev: Yes. Here's this is just an example. Actual structure will be defined in the DVM/kinds repo? 

`prompt` is used universally in the AI APIs for both text and image. But if we think DVM is more than AI, then we we can change it to text.

90.md
starbackr-dev: I've thoroughly considered this. I don't want to add the service provider's public key again. Additionally, if the parameters are encrypted, only the service provider who can decrypt them can process the request. So, having a `p` tag refers to that specific service provider. Since the `content` field is empty anyway, it's best to store the encrypted data similarly to NIP-51 private lists. Therefore, if both the `p` tag and `content` exist, other service providers can ignore them, while the owner of the 'p' tag's public key can process the request. Start by decrypting the 'content' to view all the tags and then examine other tags to process the request.

90.md
cj-chua: My understanding is that as per https://github.com/nostr-protocol/nips/blob/vending-machine/90.md, `<input-type>` must be one of `text`, `url`, `job`, `event`

pablof7z: yeah, @cj-chua is right, the third value should be `text` as that defines how the input should be interpreted (i.e. it's text, it's not a job id or an event ID that should be fetched, `text` indicates the input should be used as-is)

starbackr-dev: agree and changed.

starbackr-dev:
> Left a comment/idea that would prevent this proposal from steering too far away from the event/job structure defined in NIP90.
> 
> As for the proposal itself I don't know enough to say if this is the better way to introduce encrypted input/output (as compared to gift wraps). Will let others comment on that

I am referring to NIP-51 Lists, which include both public and private lists, making them usable by users. I have already observed this with users who opt for direct messages (DMs) to generate text and images. Without encryption option I will not be able to convert all PlebAI agents into DVM compatible.

pablof7z:
Is this PR still relevant given that we merged the other encrypted one?

pablof7z:
Closing as it seems this is already accomplished.

PR #825, opened 2023-10-16
Draft for NIP-73 meetup communities

BrightonBTC:
This is a draft proposal for Meetup style location based communities. 

For a working example this is currently implemented at https://www.inmytown.social/ (https://github.com/BrightonBTC/inmytown.social).

The idea is to have communities similar to those on Meetup .com, and introduces 3 new kinds to achieve this. One for community creation, One for community metadata, and one for community follow lists. The rest of the functionality is provided by existing NIPs. NIP-52 for location and time based meetup events, NIP-28 for community chat, and NIP-51 person lists to manage membership.

### DIFF
--- a/73.md
+++ b/73.md
@@ -80,7 +80,6 @@ In addition to the fields mentioned in NIP-52 there is also the following:
 
 - an additional "g" tag formatted as the city g tags above. ie. Alpha2:City Name. This makes it possible to search by city. 
 - an ["L" , "kind"] and ["l", "meetup", "kind"] tag so that these evnts can be differentiated from those that don'e have a community associated with them
-- an "e" tag referencing the associated community.
 - an optional "brief" tag, for a short description of the event
 
 This gives us:
--- a/73.md
+++ b/73.md
@@ -1,0 +1,166 @@
+NIP-73
+======
+
+Location Based Communities (Meetup Style)
+------------------------------------
+
+`draft` `optional` `author:brighton_btc`
+
+The purpose of this NIP is to create location based communities that allow for some light moderation, leverage NIP-52 time based calendar events for the creation of in-person meetup events, NIP-51 lists for membership and moderation, and NIP-28 for community chat.
+
+# Community creation event:
+New communities are created with a `Kind:1037` event. This event acts as an anchor point for other event kinds associated with the community
+
+```json
+{
+  "content": <Optional description>,
+  "created_at": <Unix timestamp in seconds>,
+  "id": <32-bytes lowercase hex-encoded SHA-256 of the the serialized event data>,
+  "kind": 1037,
+  "pubkey": <32-bytes lowercase hex-encoded public key of the event creator>
+}
+```
+
+# Community metadata:
+
+Community metadata can be created or updated with a `Kind:30037` event. It should use the ID of the associated `Kind:1037` as its `d` tag.
+
+```json
+{
+  "content": <description of community, markdown formatted>,
+  "created_at": <Unix timestamp in seconds>,
+  "id": <32-bytes lowercase hex-encoded SHA-256 of the the serialized event data>,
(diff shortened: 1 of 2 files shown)


BrightonBTC:
I'm aware that using list events was discussed and rejected in the NIP-72 proposal due to the large size of those kinds of communities. Meetup communities don't suffer from that same issue. On meetup larger communities tend to be only a few hundred members, with only a few that stand out and go up in to the thousands, so I'm of the opinion that it's an appropriate approach to use here.

vitorpamplona:
This feels closer to NIP53-Live Activities (which also includes chats and calendar-based schedules) than to NIP-72

BrightonBTC:
> This feels closer to NIP53-Live Activities (which also includes chats and calendar-based schedules) than to NIP-72
> 
> PS: Live Activities are not for Live Streams only, it's for any event that can be broadcasted as "happening now" in people's clients.

Oh, OK I'll check that out. I had thought that was for live streams only.

73.md
```diff
@@ -0,0 +1,165 @@
...
+
+# Time-Based Calendar Event (NIP-52):
+
+The owner of a community can create `Kind:31923` calendar events to represent location based, in person, meetup events. It's `d` tag should be the ID of the associated `Kind:1037` event.
+
+In addition to the fields mentioned in NIP-52 there is also the following:
+
+- an additional "g" tag formatted as the city g tags above. ie. Alpha2:City Name. This makes it possible to search by city. 
+- an ["L" , "kind"] and ["l", "meetup", "kind"] tag so that these evnts can be differentiated from those that don'e have a community associated with them
+- an optional "brief" tag, for a short description of the event
```
staab, line 83: `alt` might be a more standard way to do this

BrightonBTC:
> This feels closer to NIP53-Live Activities (which also includes chats and calendar-based schedules) than to NIP-72
> 
> PS: Live Activities are not for Live Streams only, it's for any event that can be broadcasted as "happening now" in people's clients.

Did you mean closer to NIP-53 than NIP-52 ? 

I'm not sure I agree. Calendar events feel more appropriate here to me. If I sign up to a meetup group on meetup.com or similar then I'm not too interested in what's "happening now", I want to see upcoming events, be able to add them to my calendar, and be able to send an RSVP.

staab (commented):
I would really like to try to avoid re-specifying all functionality in different NIPs. Community name, moderation, and privacy should all be separate things that can be combined, and there should be a way to use any event in context of a community. I like the `k` tag suggestion in #753 as a way of generically doing anything in context of a community.

AsaiToshiya:
What is the reason for separating creation and metadata? I think just `kind: 30037` is enough like NIP-53 and NIP-72. Other events can refer to it with `a` tag.

BrightonBTC:
> What is the reason for separating creation and metadata? I think just `kind: 30037` is enough like NIP-53 and NIP-72. Other events can refer to it with `a` tag.

I had considered that approach but decided against it for a few reasons, primarily because I wanted to be able to know and display the creation date of communities. Having a creation event makes that simple.

AsaiToshiya:
Thanks for your answer. It makes sense, but you can also use `published_at` tag in NIP-23 for the creation date.

BrightonBTC:
> Thanks for your answer. It makes sense, but you can also use `published_at` tag in NIP-23 for the creation date.

Yeah, you're right, it could go as a separate tag as with NIP-23. My one concern here - and I suppose it also applies to NIP-23 - is that it's more open to manipulation. ie. it would be trivial for a community organiser to back date their creation date in an attempt to boost credibility (I know that creation dates on events also can't be 100% trusted but IIRC most relays at least check they fall within some reasonable time frame).

I'm not overly concerned about any of this though and am happy to change it if there's some general consensus it's a better approach.

PR #827, opened 2023-10-17
NIP-108 Lightning Gated Content

CoachChuckFF:
This PR outlines a perspective NIP-108 for lightning-gated content.

A sample client can be found at [Nostr Playground](https://nip-108.nostrplayground.com/)

An open-source server and client implementation can be found in the [Nostr Playground Repo](https://github.com/project-excalibur/Nostr-Playground/tree/main/apps/nip108)

### DIFF
--- a/108.md
+++ b/108.md
@@ -1,0 +1,170 @@
+NIP-108
+======
+
+Lightning Gated Notes
+-------------------------------
+
+`draft` `optional` `author:coachchuckff` `author:excalibur_guild`
+
+This NIP defines three events for gating Notes behind lightning paywalls: 
+
+- Lightning-Gated Note ( `kind:55` ): This note allows you wrap any type of note behind a lightning gated paywall by encrypting the payload with a purchasable decrypt key.
+- Key Note ( `kind:56` ): This note encrypts the key for a given note, per user, using [NIP-04](https://github.com/nostr-protocol/nips/blob/master/04.md) between the corresponding gate creator's public key and the purchasers. It is linked to the gated note with the `g` tag.
+- Announcement Note ( `kind:54` ): This note acts as the announcement of the gated note, giving a short preview of the content. It is linked to the gated note with the `g` tag.
+
+A complete open-source implantation of [NIP-108 is available](https://github.com/project-excalibur/Nostr-Playground/tree/main/apps/nip108).
+
+An example client can be found at [Nostr Playground](https://nip-108.nostrplayground.com/).
+
+## Protocol flow
+### Creating Gated Notes
+1. Poster creates a note of any kind
+2. Poster `JSON.stringify`s the whole note.
+3. Poster encrypts the note string with a new nsec `secret`, and `iv` using `aes-256-cbc`.
+4. Poster creates the gated note `kind:55` with the encrypted note json string in the `content` field, while putting `iv`, `cost` (mSats), and `endpoint` as tags. The `endpoint` tag is the server endpoint you use to hold your `secret`'s and issue lightning invoices from your `lud16`.
+5. Poster then creates an announcement note `kind:54` with the `g` tag (gated note's id) to preview the gated content.
+
+### Consuming Gated Notes
+1. Client finds gated content they want to purchase by browsing `kind:54` announcement notes. 
+2. Client loads the associated gated note of `kind:55` found in the `g` tag
+3. Client then GETs the `[endpoint]/[id]`
+4. Gate server will respond with a 402 PR requesting a payment for the `cost` tag's amount in mSats
+5. Client pays the amount
+6. Client uses the `successAction` url returned in the PR to fetch the `secret` which will unlock the gated content.
+7. Client uses the `secret` and the gated note's `iv` tag to decrypt the content using `aes-256-cbc`
+8. Client then creates a key note `kind:56` with the content being the `nip-04` encrypted secret with their publicKey and the gate note's creator publicKey.
+9. Upon revisiting the gated note, the client can then decrypt the content using their key note.  
+
+## Server Functions
+NIP-108 requires an outside server to store `secret`s and issue lighting invoices to those wishing to purchase the digital content.
+
+The server should have two endpoints:
+`[endpoint]/create` - POST to create new notes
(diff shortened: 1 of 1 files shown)


108.md
```diff
@@ -0,0 +1,170 @@
...
+
+Lightning Gated Notes
+-------------------------------
+
+`draft` `optional` `author:coachchuckff` `author:excalibur_guild`
+
+This NIP defines three events for gating Notes behind lightning paywalls: 
+
+- Lightning-Gated Note ( `kind:55` ): This note allows you wrap any type of note behind a lightning gated paywall by encrypting the payload with a purchasable decrypt key.
+- Key Note ( `kind:56` ): This note encrypts the key for a given note, per user, using [NIP-04](https://github.com/nostr-protocol/nips/blob/master/04.md) between the corresponding gate creator's public key and the purchasers. It is linked to the gated note with the `g` tag.
```
AsaiToshiya, line 12: Is the key note optional?

108.md
```diff
@@ -0,0 +1,170 @@
...
+
+`.content` some preview or announcement of the content you have locked away.
+
+`.tag` MUST include the following:
+
+- `g`, the id of the gated event.
+
+### Key Note ( Kind:56 )
+
+`kind:54`
```
AsaiToshiya, line 114:
```suggestion
`kind:56`
```

Egge7:
I very much like this.
This solves the same problem as https://github.com/nostr-protocol/nips/pull/542

Both approaches have their pro and cons and come with some overhead. However I want to point out that Zap Gates are easily extensible for more complex setups like fundraisers, subscriptions etc.

108.md
```diff
@@ -0,0 +1,170 @@
...
+
+`.content` the `secret` encrypted via [NIP-04](https://github.com/nostr-protocol/nips/blob/master/04.md)'s encrypt function between the gated note's creator's pubkey and your pubkey.
```
AsaiToshiya, line 116:
```suggestion
`.content` the `secret` encrypted via [NIP-04](https://github.com/nostr-protocol/nips/blob/master/04.md)'s encrypt function between the gated note's creator's pubkey and your prikey.
```

108.md
```diff
@@ -0,0 +1,170 @@
...
+
+`.tag` MUST include the following:
+
+- `g`, the id of the gated event.
+
+### Encryption/Decryption
+
+To encrypt/decrypt `kind:56` key notes, we use [NIP-04](https://github.com/nostr-protocol/nips/blob/master/04.md)'s encrypt/decrypt functions between the gated note's creator's pubkey and your pubkey.
```
AsaiToshiya, line 124:
```suggestion
To encrypt/decrypt `kind:56` key notes, we use [NIP-04](https://github.com/nostr-protocol/nips/blob/master/04.md)'s encrypt/decrypt functions between the gated note's creator's pubkey and your prikey.
```

AsaiToshiya (commented):
Sorry if I'm wrong about NIP-04.

108.md
CoachChuckFF: Technically, yes, the key note is optional but highly recommended. You can always store your own decrypt keys, or whatever client you are using could. However, I'd try to limit the reliance on 3rd parties as much as possible. Storing the key note on Nostr makes it much more likely you always have access to the gated note.

108.md
```diff
@@ -0,0 +1,170 @@
...
+
+### Consuming Gated Notes
+1. Client finds gated content they want to purchase by browsing `kind:54` announcement notes. 
+2. Client loads the associated gated note of `kind:55` found in the `g` tag
+3. Client then GETs the `[endpoint]/[id]`
+4. Gate server will respond with a 402 PR requesting a payment for the `cost` tag's amount in mSats
+5. Client pays the amount
+6. Client uses the `successAction` url returned in the PR to fetch the `secret` which will unlock the gated content.
+7. Client uses the `secret` and the gated note's `iv` tag to decrypt the content using `aes-256-cbc`
+8. Client then creates a key note `kind:56` with the content being the `nip-04` encrypted secret with their publicKey and the gate note's creator publicKey.
```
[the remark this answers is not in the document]
CoachChuckFF: You're right here! Thank you!

108.md
AsaiToshiya: It makes sense.

CoachChuckFF:
> I very much like this. This solves the same problem as #542
> 
> Both approaches have their pro and cons and come with some overhead. However I want to point out that Zap Gates are easily extensible for more complex setups like fundraisers, subscriptions etc.

I just took a look at it! Great spec! And you're right, Zap Gates are more extensible, and probably more likely to be adopted.
 
I'd say the main benefit here is simplicity and that you only have to rely on an outside server twice. Once when creating the note and once when purchasing. Once you have the unlock key, it's yours forever since it all stays within the Nostr ecosystem. Granted, this is currently only good for text and small embedded images/files. However, it's feasible that Nostr will implement systems for larger embedded files in the future.

I'm a bit of a digital preservationist, maybe there is a way to preserve purchases within Zap Gates?

108.md
```diff
@@ -0,0 +1,170 @@
...
+Lightning Gated Notes
+-------------------------------
+
+`draft` `optional` `author:coachchuckff` `author:excalibur_guild`
+
+This NIP defines three events for gating Notes behind lightning paywalls: 
+
+- Lightning-Gated Note ( `kind:55` ): This note allows you wrap any type of note behind a lightning gated paywall by encrypting the payload with a purchasable decrypt key.
+- Key Note ( `kind:56` ): This note encrypts the key for a given note, per user, using [NIP-04](https://github.com/nostr-protocol/nips/blob/master/04.md) between the corresponding gate creator's public key and the purchasers. It is linked to the gated note with the `g` tag.
+- Announcement Note ( `kind:54` ): This note acts as the announcement of the gated note, giving a short preview of the content. It is linked to the gated note with the `g` tag.
```
AsaiToshiya, line 13: What is the reason for the `g` tag? Consider using standardized `e` tag.

arkin0x:
@staab this PR can be closed. Fanfares isn't using it nor advocating for it anymore.

Issue #820, opened 2023-10-13
name is not defined in NIP-24


melvincarvalho:
I'd like to document the fields in a profile in a schema, to be compatible with linked data, possibly

https://w3id.org/nostr

However name is not defined in NIP-24

`display_name: a bigger name with richer characters than name`

Which is somewhat vague.  Normally you have something like a real name type thing (name) and a nickname type thing (nick)

If we can get some definitions I can add them to the schema

fiatjaf:
`name`, `picture` and `about` are defined in NIP-01. Maybe we need to mention that again in NIP-24?

accursed-rose:
I would suggest that `display_name` is not *inherently* "bigger" and may not have "richer characters" either (although one or both may well be true).

I have noticed that in the realm of ActivityPub implementations, it is not unheard of for users to have a longer username than display name; for example, their chosen username on the instance they joined was already claimed, so they appended data to it and then stripped it in the display name, which does not mandate uniqueness (unlike the username). I've also seen display names set temporarily for various reasons such as specific calendar events, often in a memelike format (which may result in a display name that is in entirely distinct from the username).

I also agree that referencing NIP-01 is reasonable, for clarity.

I have provided alternative phrasing in #828, in case that is helpful.

PR #828, opened 2023-10-18
Clarify wording of NIP-24

accursed-rose:
Possibly resolves nostr-protocol/nips#820. Please feel free to suggest any further changes, or to simply reject if these changes are undesired.


AsaiToshiya:
The purpose is self-explanatory in `display_name` and also the relationship with `name` can be guessed, so I don't see the need to change `a bigger name with richer characters than name`. Other than that LGTM!

fiatjaf:
I don't think `display_name` should be preferred.

PR #831, opened 2023-10-18
rewrite NIP-46

fiatjaf:
this is an attempt to simplify the text, remove garbage and incorporate the lessons from nsecbunker implementation.

result: https://github.com/nostr-protocol/nips/blob/nostr-connect/46.md

### DIFF
--- a/46.md
+++ b/46.md
@@ -80,5 +80,5 @@ And the events the remote signer sends to the client have the following format:
   - params: [`pubkey`, `plaintext`]
   - result: `nip4 ciphertext`
 - **nip04_decrypt**
-  - params: [`pubkey`, `nip4 ciphertext`]
+  - params: [`peer-pubkey`, `nip4 ciphertext`]
   - result: [`plaintext`]
--- a/46.md
+++ b/46.md
@@ -84,3 +84,15 @@ The signer key will always be the key of the user who controls the signer device
 - **nip04_decrypt**
   - params: [`third-party-pubkey`, `nip04-ciphertext`]
   - result: `plaintext`
+- **nip44_get_key**
+  - params: [`third-party-pubkey`]
+  - result: `nip44-conversation-key`
+- **nip44_encrypt**
+  - params: [`third-party-pubkey`, `plaintext`]
+  - result: `nip44-ciphertext`
+- **nip44_decrypt**
+  - params: [`third-party-pubkey`, `nip44-ciphertext`]
+  - result: `plaintext`
+- **ping**
+  - params: []
+  - result: `"pong"`
--- a/46.md
+++ b/46.md
@@ -12,12 +12,12 @@ This NIP describes a method for 2-way communication between a **remote signer** 
 
 The client always starts by generating a random key which is used to communicate with the signer, then it one of the methods below is used to allow the client to know what is the signer public key for the session and which relays to use.
 
-### Started by the signer
+### Started by the signer (nsecBunker)
 
 The remote signer generates a connection token in the form
 
 ```
-<npub1...>#<secret>?relay=wss://...&relay=wss://...
+<npub1...>#<optional-secret>?relay=wss://...&relay=wss://...
 ```
 
 The user copies that token and pastes it in the client UI somehow. Then the client can send events of kind `24133` to the specified relays and wait for responses from the remote signer.
@@ -34,7 +34,7 @@ The signer scans the QR code and sends a `connect` message to the client in the 
 
 ## Event payloads
 
-Event payloads are [NIP-04](04.md)-encrypted JSON blobs that look like JSONRPC.
+Event payloads are [NIP-04](04.md)-encrypted JSON blobs that look like JSONRPC messages (their format is specified inside the `.content` of the event formats nelow).
 
 Events sent by the client to the remote signer have the following format:
 
@@ -45,7 +45,7 @@ Events sent by the client to the remote signer have the following format:
   "tags": [
     ["p", "<signer-key-hex>"]
   ],
-  "content": "nip04_encrypted_json({id: <random-string>, method: <see-below>, params: []})",
+  "content": "nip04_encrypted_json({id: <random-string>, method: <see-below>, params: [array_of_strings]})",
   ...
 }
 ```
@@ -58,27 +58,29 @@ And the events the remote signer sends to the client have the following format:
   "tags": [
     ["p", "<client-key-hex>"]
   ],
-  "content": "nip04_encrypted_json({id: <request-id>, result: <any>, error: <reason-string>})",
+  "content": "nip04_encrypted_json({id: <request-id>, result: <string>, error: <reason-string>})",
   ...
 ```
+
+The signer key will always be the key of the user who controls the signer device.
 
 ### Methods
 
 - **connect**
   - params: [`pubkey`, `secret`]
-  - result: `null`
+  - result: `"ack"`
 - **get_public_key**
   - params: []
-  - result: `pubkey`
+  - result: `pubkey-hex`
 - **sign_event**
   - params: [`event`]
-  - result: `event_with_pubkey_id_and_signature`
+  - result: `json_string(event_with_pubkey_id_and_signature)`
 - **get_relays**
   - params: []
-  - result: `{ [url: string]: {read: boolean, write: boolean} }`
+  - result: `json_string({[url: string]: {read: boolean, write: boolean}})`
 - **nip04_encrypt**
-  - params: [`peer-pubkey`, `plaintext`]
-  - result: `nip4 ciphertext`
+  - params: [`third-party-pubkey`, `plaintext`]
+  - result: `nip04-ciphertext`
 - **nip04_decrypt**
-  - params: [`peer-pubkey`, `nip4 ciphertext`]
-  - result: [`plaintext`]
+  - params: [`third-party-pubkey`, `nip04-ciphertext`]
+  - result: `plaintext`
--- a/46.md
+++ b/46.md
@@ -77,7 +77,7 @@ And the events the remote signer sends to the client have the following format:
   - params: []
   - result: `{ [url: string]: {read: boolean, write: boolean} }`
 - **nip04_encrypt**
-  - params: [`pubkey`, `plaintext`]
+  - params: [`peer-pubkey`, `plaintext`]
   - result: `nip4 ciphertext`
 - **nip04_decrypt**
   - params: [`peer-pubkey`, `nip4 ciphertext`]
--- a/46.md
+++ b/46.md
@@ -10,7 +10,7 @@ This NIP describes a method for 2-way communication between a **remote signer** 
 
 ## Signer Discovery
 
-The client must somehow be able to contact the signer through a common relay.
+The client always starts by generating a random key which is used to communicate with the signer, then it one of the methods below is used to allow the client to know what is the signer public key for the session and which relays to use.
 
 ### Started by the signer
 
@@ -27,7 +27,7 @@ The user copies that token and pastes it in the client UI somehow. Then the clie
 The client generates a QR code in the following form (URL-encoded):
 
 ```
-nostrconnect://<client-key-hex>?relay=wss://...&metadata={"name":"client-name"}
+nostrconnect://<client-key-hex>?relay=wss://...&metadata={"name":"...", "url": "...", "description": "..."}
 ```
 
 The signer scans the QR code and sends a `connect` message to the client in the specified relays.
--- a/46.md
+++ b/46.md
@@ -2,161 +2,83 @@ NIP-46
 ======
 
 Nostr Connect
-------------------------
+-------------
 
 `draft` `optional`
 
-## Rationale
+This NIP describes a method for 2-way communication between a **remote signer** and a normal Nostr client. The remote signer could be, for example, a hardware device dedicated to signing Nostr events, while the client is a normal Nostr client.
 
-Private keys should be exposed to as few systems - apps, operating systems, devices - as possible as each system adds to the attack surface.
+## Signer Discovery
 
-Entering private keys can also be annoying and requires exposing them to even more systems such as the operating system's clipboard that might be monitored by malicious apps.
+The client must somehow be able to contact the signer through a common relay.
 
+### Started by the signer
 
-## Terms
+The remote signer generates a connection token in the form
 
-* **App**: Nostr app on any platform that *requires* to act on behalf of a nostr account.
-* **Signer**: Nostr app that holds the private key of a nostr account and *can sign* on its behalf.
+```
+<npub1...>#<secret>?relay=wss://...&relay=wss://...
+```
 
+The user copies that token and pastes it in the client UI somehow. Then the client can send events of kind `24133` to the specified relays and wait for responses from the remote signer.
 
-## `TL;DR`
+### Started by the client
 
+The client generates a QR code in the following form (URL-encoded):
 
-**App** and **Signer** sends ephemeral encrypted messages to each other using kind `24133`, using a relay of choice. 
+```
+nostrconnect://<client-key-hex>?relay=wss://...&metadata={"name":"client-name"}
+```
 
-App prompts the Signer to do things such as fetching the public key or signing events.
+The signer scans the QR code and sends a `connect` message to the client in the specified relays.
 
-The `content` field must be an encrypted JSONRPC-ish **request** or **response**.
+## Event payloads
 
-## Signer Protocol
+Event payloads are [NIP-04](04.md)-encrypted JSON blobs that look like JSONRPC.
 
-### Messages
+Events sent by the client to the remote signer have the following format:
 
-#### Request
-
-```json
+```js
 {
-  "id": <random_string>,
-  "method": <one_of_the_methods>,
-  "params": [<anything>, <else>]
+  "pubkey": "<client-key-hex>"
+  "kind": 24133,
+  "tags": [
+    ["p", "<signer-key-hex>"]
+  ],
+  "content": "nip04_encrypted_json({id: <random-string>, method: <see-below>, params: []})",
+  ...
 }
 ```
 
-#### Response
+And the events the remote signer sends to the client have the following format:
 
-```json
-{
-  "id": <request_id>,
-  "result": <anything>,
-  "error": <reason>
-}
+```js
+  "pubkey": "<signer-key-hex>"
+  "kind": 24133,
+  "tags": [
+    ["p", "<client-key-hex>"]
+  ],
+  "content": "nip04_encrypted_json({id: <request-id>, result: <any>, error: <reason-string>})",
+  ...
 ```
 
 ### Methods
 
-
-#### Mandatory 
-
-These are mandatory methods the remote signer app MUST implement:
-
-- **describe**
-  - params []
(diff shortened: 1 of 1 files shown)


vitorpamplona:
This is better. I would add an explanation for the use of the secret.

pablof7z:
@fiatjaf, added a description of `secret`

pablof7z (approved):

Egge7:
I think this is missing an explicit example and definition to be clear. 

- What is the npub specified in the connection string
- Which key is used to encrypt the payload? Is it the secret?
- What is the "user-public-key"?

Maybe it is just me, but I wasnt able to implement this just now. Having a "step-by-step" flow and more clarity about the different keys would certainly help.

46.md
```diff
@@ -2,161 +2,67 @@ NIP-46
...
 
-
-#### Mandatory 
-
-These are mandatory methods the remote signer app MUST implement:
-
-- **describe**
-  - params []
-  - result `["describe", "get_public_key", "sign_event", "connect", "disconnect", "delegate", ...]`  
+- **connect**
```
rolznz, line 51: Is the connect event still needed given the user now pastes the key from the signer into the app?

fiatjaf: I agree, but I think implementations out there are relying on this? @pablof7z @staab

staab: I don't know, I use pablo's implementation

pablof7z: if the user is not sending a secret to apply a pre-authorized list the `connect` method is redundant; nsecBunker doesn't depend on it. Ok with me if we make it optional

fiatjaf: Can we just remove it?

mikedilger:
What determines the client-deterministically-generated-key? that wasn't specified

mikedilger:
The 'p' tag in the event sent to the signer says it uses 'user-pubkey-hex'.  The term 'user' there confuses me. It makes sense if this is the pubkey of the remote signer. Is that what this is?

46.md
```diff
@@ -2,161 +2,67 @@ NIP-46
...
-
-### Delegate
-
-1. The **App** will send a message with metadata to the **Signer** with a `delegate` request along with the **conditions** query string and the **pubkey** of the **App** to be delegated.
-2. The **Signer** will show a popup to the user to delegate the **App** to sign on his behalf
-3. The **Signer** will send back a message with the signed [NIP-26 delegation token](https://github.com/nostr-protocol/nips/blob/master/26.md) or reject it
-
-
+  - params: [`pubkey`, `nip4 ciphertext`]
+  - result: [`plaintext`]
```
monlovesmango, line 68: since we are revamping this NIP, can we change nip04_decrypt result to not be an array? its the only result that returns an array, was there a reason for that? seems inconsistent.

46.md
```diff
@@ -2,161 +2,67 @@ NIP-46
...
+  - result: `event_with_pubkey_id_and_signature`
 - **get_relays**
-  - params []
-  - result `{ [url: string]: {read: boolean, write: boolean} }` 
+  - params: []
+  - result: `{ [url: string]: {read: boolean, write: boolean} }`
 - **nip04_encrypt**
-  - params [`pubkey`, `plaintext`]
-  - result `nip4 ciphertext`
+  - params: [`pubkey`, `plaintext`]
```
monlovesmango, line 64: for the nip04_encrypt and nip04_decrypt methods below, we should update params to contain `peer pubkey` or some other description that makes it clear it is not the user's own pubkey

46.md
monlovesmango: I think connect is nice to have for future connections, just so client can verify that it can reach the signer (and if not notify the user). however I am not opposed to removing it entirely. 

If it is not removed, we should at least remove the params (since the signer should be able to identify the client from the event's pubkey already)

46.md
```diff
@@ -2,161 +2,67 @@ NIP-46
...
-#### Request
+Events sent by the client to the remote signer have the following format:
 
-```json
+```js
 {
-  "id": <random_string>,
-  "method": <one_of_the_methods>,
-  "params": [<anything>, <else>]
+  "pubkey": "<client-deterministically-generated-key-for-the-connection>"
```
monlovesmango, line 27: wait is it really just for one connection? shouldn't we allow reuse of the client's keys for multiple connections?

46.md
```diff
@@ -2,161 +2,67 @@ NIP-46
...
-{
-  "id": <request_id>,
-  "result": <anything>,
-  "error": <reason>
-}
+```js
+  "pubkey": "<user-pubkey>"
+  "kind": 24133,
+  "tags": [
+    ["p", "<signer-pubkey-hex>"]
```
monlovesmango, line 43: "<signer-pubkey-hex>" should be "<client-pubkey-hex>"

monlovesmango:
general question about intended usage of nostrConnect, do we anticipate that this should be compatible with relay authentication?

It seems like a good idea to be able to use nostrConnect over authenticated relays, however right now I don't think you can establish a nostrConnect connection over an authenticated relay.

46.md
fiatjaf: No, because it would break implementations. This rewrite doesn't try to change the functionality.

fiatjaf:
> What determines the client-deterministically-generated-key? that wasn't specified

I have no idea.

> The 'p' tag in the event sent to the signer says it uses 'user-pubkey-hex'. The term 'user' there confuses me. It makes sense if this is the pubkey of the remote signer. Is that what this is?

I think it could be either. The signer can act as if it was actually the user and send messages as the user, but it may be better to use a random key for the session or a persistent signer-key. In practice it doesn't matter, at least for now. It may matter when restricted relays start to be used, then we change the NIP.

46.md
fiatjaf: Yes, I believe we allow anything.

46.md
fiatjaf: I think this is fixed.

46.md
fiatjaf: I am so confused by this I won't update it until I actually know what I'm doing, but please send a PR if you do.

46.md
monlovesmango: ok fine. I guess nip4 will soon be deprecated in favor of nip44 soon anyway.

fiatjaf:
Please review this again. We must merge this ASAP.

And if you find a bug or disagree with a small point that should be fixed please send a suggestion I can click on and apply.

46.md
```diff
@@ -2,161 +2,83 @@ NIP-46
...
+  - result: `event_with_pubkey_id_and_signature`
 - **get_relays**
-  - params []
-  - result `{ [url: string]: {read: boolean, write: boolean} }` 
+  - params: []
+  - result: `{ [url: string]: {read: boolean, write: boolean} }`
 - **nip04_encrypt**
-  - params [`pubkey`, `plaintext`]
-  - result `nip4 ciphertext`
+  - params: [`pubkey`, `plaintext`]
```
monlovesmango, line 80:
```suggestion
  - params: [`peer-pubkey`, `plaintext`]
```

46.md
```diff
@@ -2,161 +2,83 @@ NIP-46
...
-3. The **Signer** will send back a message with the event including the `id` and the schnorr `signature` as a response to the `sign_event` request
-
-### Delegate
-
-1. The **App** will send a message with metadata to the **Signer** with a `delegate` request along with the **conditions** query string and the **pubkey** of the **App** to be delegated.
-2. The **Signer** will show a popup to the user to delegate the **App** to sign on his behalf
-3. The **Signer** will send back a message with the signed [NIP-26 delegation token](https://github.com/nostr-protocol/nips/blob/master/26.md) or reject it
-
-
+  - params: [`pubkey`, `nip4 ciphertext`]
```
monlovesmango, line 83:
```suggestion
  - params: [`peer-pubkey`, `nip4 ciphertext`]
```

46.md
```diff
@@ -2,161 +2,83 @@ NIP-46
...
-
-#### Mandatory 
-
-These are mandatory methods the remote signer app MUST implement:
-
-- **describe**
-  - params []
-  - result `["describe", "get_public_key", "sign_event", "connect", "disconnect", "delegate", ...]`  
+- **connect**
+  - params: [`pubkey`, `secret`]
```
monlovesmango, line 68:
```suggestion
  - params: []
```

monlovesmango (changes requested):
connect shouldn't need any parameters right? all the signer needs is the client's pubkey which it can get from the event.

46.md
```diff
@@ -2,161 +2,83 @@ NIP-46
...
+### Started by the signer
 
-## Terms
-
-* **App**: Nostr app on any platform that *requires* to act on behalf of a nostr account.
-* **Signer**: Nostr app that holds the private key of a nostr account and *can sign* on its behalf.
+The remote signer generates a connection token in the form
 
+```
+<npub1...>#<secret>?relay=wss://...&relay=wss://...
```
v0l, line 20: Can this also be a URL please? like `nostrconnect://<remote-pubkey>?relay=wss://example.com#<secret>`

this should work in both directions, `metadata` is not required in this direction ofc

46.md
```diff
@@ -2,161 +2,83 @@ NIP-46
...
+```
 
-## `TL;DR`
+The user copies that token and pastes it in the client UI somehow. Then the client can send events of kind `24133` to the specified relays and wait for responses from the remote signer.
```
v0l, line 23: I think the term `<remote-pubkey>` and `<local-pubkey>` are better because the local/remote are reversed depending on which side its initiated from, and ofc `<remote-pubkey>` is always who you're supposed to `p` tag which wasn't clear in this rewrite

46.md
```diff
@@ -2,161 +2,83 @@ NIP-46
...
 `draft` `optional`
 
-## Rationale
+This NIP describes a method for 2-way communication between a **remote signer** and a normal Nostr client. The remote signer could be, for example, a hardware device dedicated to signing Nostr events, while the client is a normal Nostr client.
 
-Private keys should be exposed to as few systems - apps, operating systems, devices - as possible as each system adds to the attack surface.
+## Signer Discovery
 
-Entering private keys can also be annoying and requires exposing them to even more systems such as the operating system's clipboard that might be monitored by malicious apps.
+The client must somehow be able to contact the signer through a common relay.
```
v0l, line 13: `The client always starts by generating a random key which is used to communicate with the signer`

46.md
```diff
@@ -2,161 +2,83 @@ NIP-46
...
+The user copies that token and pastes it in the client UI somehow. Then the client can send events of kind `24133` to the specified relays and wait for responses from the remote signer.
 
+### Started by the client
 
-**App** and **Signer** sends ephemeral encrypted messages to each other using kind `24133`, using a relay of choice. 
+The client generates a QR code in the following form (URL-encoded):
 
-App prompts the Signer to do things such as fetching the public key or signing events.
+```
+nostrconnect://<client-key-hex>?relay=wss://...&metadata={"name":"client-name"}
```
v0l, line 30: Missing the other metadata here? `url` / `description` / `icon`

46.md
monlovesmango: no, based on this flow the signer will create the client's random key. also it should be possible (but not required) to reuse this random key for future sessions.

46.md
monlovesmango: only the client can initiate a request and only a signer can send a response. I don't think we need to worry about reverse situations..? maybe i'm not understanding what you are getting at.

right now the `p` tag seems really clear and explicit to me, the remote signer should tag the client pubkey, and the client should tag the remote signer pubkey.

46.md
monlovesmango: I don't think there is any both directions. previously the client created the connection details for the signer. now the signer creates the connection details for the client (so its completely reversed). but maybe i'm misunderstanding...

monlovesmango:
wait why are we adding back the old method of also allowing client to establish connection details? wasn't the point of this PR to remove that? I think its more confusing if either side can create the connection details, we should only have one side doing this. I'm ambivalent on wither its the signer or the client, both have pros and cons.

also for "Started by the client" section, the signer should only be able to send a connect response (not request). either have the signer send a connect response with a defaulted ID defined in this PR or have the client send connect requests until the signer responds.

alexgleason (approved):
Much better than before.

46.md
fiatjaf: This is changing the API that is already in use out there in a bunch of places.

46.md
fiatjaf: That wasn't in the NIP before, will add.

46.md
fiatjaf: Do we even need this `secret`? Maybe we should go on the other direction and just paste an `npub1`?

But yes, I personally prefer the URL, the problem is that doing it will break the the nsecBunker flow as it is implemented in everywhere.

fiatjaf:
@monlovesmango I don't understand what "peer-pubkey" means to be honest, but changed it anyway.

> wait why are we adding back the old method of also allowing client to establish connection details? wasn't the point of this PR to remove that? I think its more confusing if either side can create the connection details, we should only have one side doing this.

No, the point was to make this NIP readable and to add the nsecBunker method. Snort has the other flow implemented and I think both could be supported because it is easier, for example, for a hardware device or dedicated mobile signer to use the second method. At least that's my understanding, but it might be worth removing it anyway, one is always better than two.

@v0l what do you think of removing the second method? Is anyone else implementing it?

46.md
rolznz: Since `secret` is being introduced can it please be explained what it is and why? @pablof7z can you help here?

I think this is from:

> if the user is not sending a secret to apply a pre-authorized list the connect method is redundant; nsecBunker doesn't depend on it. Ok with me if we make it optional

What is a pre-authorized list?

46.md
v0l: In bunker the secret is basically a token which will grant certain access to a specific key, it can expire if not claimed and so on, so it has a purpose there.

Its probably because the client is communicating from a random key and this key is not enrolled into the bunker, they claim the token and then the key is associated to those permissions.

monlovesmango:
ok makes sense. I still prefer only one method of connection but will not push back over it.

if we are keeping the client -> signer connection method, can we also keep all the other functionality that was originally spec'ed in nip46? i'll make some suggestions.

46.md
```diff
@@ -2,161 +2,83 @@ NIP-46
...
+<npub1...>#<secret>?relay=wss://...&relay=wss://...
+```
 
-## `TL;DR`
+The user copies that token and pastes it in the client UI somehow. Then the client can send events of kind `24133` to the specified relays and wait for responses from the remote signer.
 
+### Started by the client
 
-**App** and **Signer** sends ephemeral encrypted messages to each other using kind `24133`, using a relay of choice. 
+The client generates a QR code in the following form (URL-encoded):
```
monlovesmango, line 27:
```suggestion
The client generates a random key and creates an URL encoded `nostrconnect` link in the following form (all metadata fields are optional):
```

46.md
```diff
@@ -2,161 +2,83 @@ NIP-46
...
 
-App prompts the Signer to do things such as fetching the public key or signing events.
+```
+nostrconnect://<client-key-hex>?relay=wss://...&metadata={"name":"...", "url": "...", "description": "..."}
```
monlovesmango, line 30:
```suggestion
nostrconnect://<client-key-hex>?relay=wss://...&metadata={"name":"...", "url": "...", "description": "...", "icons": ["...", "..."]}
```

46.md
```diff
@@ -2,161 +2,83 @@ NIP-46
...
+```
 
-The `content` field must be an encrypted JSONRPC-ish **request** or **response**.
+The signer scans the QR code and sends a `connect` message to the client in the specified relays.
```
monlovesmango, line 33:
```suggestion
This link is input into the signer using one of the methods below:
- signer scanning a QR code
- deeplinking to the signer app
- pasting it into the signer

Once the link has been entered into the signer, the signer should send an event with payload of { id: "connect", result: `true` } using the indicated relays as acknowledgement that the signer is ready to receive requests from the client.
```
we can change the payload to whatever people think is best. Since the spec says that the payloads for events sent by the signer should look like {id: <request-id>, result: <any>, error: <reason-string>} it was confusing to say that the signer should send a 'connect' message. I figured its easier to just send a dummy initial connection payload. alternatively, we can change the spec to allow signer to send requests as well but I don't like that at all.

46.md
monlovesmango: @fiatjaf thats true. @rolznz also has a point that @pablof7z said that 'connect' is not needed for nsecbunker, so why are we adding secret here? if secret is needed here for nsecbunker, can we atleast indicate it is optional for client -> signer initializations?

updated my suggestion, also renamed pubkey to indicate that it is the client pubkey. though maybe this is intended to be the signer pubkey? I actually am not sure but its a good thing to state explicitly.

46.md
pablof7z: I think we should go into a URL, clients are the ones parsing the client, so it's not a breaking change for them.

pablof7z: How about `nostrconnect://<nip05|pubkey>?relay=...&secret=...`

46.md
nostrband: Signer should be able to send the relay list in this 'connect' message for the app->signer flow. App can't know which relays work best with this signer, app chooses some random relay and puts into nostrconnect: url, signer sends 'connect' back with a list of proper relays as payload. App should then switch and use those relays instead of the random one (or at least add those relays to the pool).

General purpose relays don't work well with nip46 - you quickly hit rate limits for EVENT messages. Plus you can build specialized optimized relays for ephemeral-only load. Plus signer might actually be bundled with a relay to avoid a second network hop. Given all this, it's the signer that should decide which relays work best with it.

monlovesmango: I actually completely agree with this (about signer sending back relay with connect response). was going to suggest something similar but thought it would add more complexity and changes so I didn't bother. thank you for throwing it out there!

nostrband: Also, Nostr Wallet Auth proposal https://github.com/nostr-protocol/nips/pull/851/ is solving a similar flow of app->wallet, they too publish a 'connect'-like event after connection is confirmed in the wallet, but they use 30k replaceable event. Their justification is:
> We use a parameterized replaceable event to allow for the wallet and app to
renegotiate the connection if needed and to allow for the wallet being able to look up the connection details later.

Since this 'connect' confirmation event is a new idea in this rewrite maybe we could use a 30k event here too?

monlovesmango: I think they use 10k replaceable event, and it doesnt seem to contain information specific to the user (just holds what methods the wallet service supports). 

i'm not sure how comfortable I would be for this nostrConnect connect event to be queryable. 

it is interesting how different these two nips are despite some of their similarities. for example nip47 uses two different kinds for request and response, here we use one.

nostrband: They changed it to 30k, and it's encrypted so it only leaks metadata that a user was connecting to 'some app', so maybe it's not that bad.

nostrband: > it doesnt seem to contain information specific to the user

Their connect event does contain a relay, which is super useful.

fiatjaf:
If no one opposes I'll merge this as it is. Please send smaller PRs fixing stuff afterwards.

fiatjaf:
Savage.

PR #835, opened 2023-10-24
find tag by name

mattn:
If `#` is given for filters, REQ returns a result containing all of the list of tag names have that follow.

motivation: current specification can not find events only that have `g` tag.


mattn:
Ah, yes. closing.

PR #846, opened 2023-10-27
Clarify NIP-42 flow

arthurfranca:
Previous discussion: #841 and #571 

Added a paragraph to NIP-42 explaining the expected flow when relay sends an AUTH in response to a client message that requires authentication. This is a important detail that wasn't guaranteed to be followed by relays. Because of that, clients couldn't rely on it.

It uses the relay "buffering" method @staab suggested.

### DIFF
--- a/42.md
+++ b/42.md
@@ -4,7 +4,7 @@ NIP-42
 Authentication of clients to relays
 -----------------------------------
 
-`draft` `optional` `author:Semisol` `author:fiatjaf`
+`draft` `optional` `author:Semisol` `author:fiatjaf` `author:arthurfranca` `author:staab`
 
 This NIP defines a way for clients to authenticate to relays by signing an ephemeral event.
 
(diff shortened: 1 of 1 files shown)


staab:
After much discussion and a break from this topic, I think there is complexity whether the relay or the client does the buffering. I still don't like eager AUTH for privacy reasons (the easiest thing to do for clients is to always authenticate), but I accept that making the relay buffer messages isn't any better than having the client re-send messages. I will revisit #841. Thanks @arthurfranca!

fiatjaf:
Has anyone tried to implement AUTH in a real-world scenario?

arthurfranca:
> If clients are only interested in events that don't require authentication they should be specifying the kind in their filters.

Very smart, its a good idea for relays to look at the `kinds` filter and if empty simply consider it as a restricted query.

> I think there is complexity whether the relay or the client does the buffering.

Yeah this one adds complexity specially to relays. #891 helps mitigate some of NIP-42 problems by letting client authenticate before requesting but just when it thinks the request should use auth.

> Has anyone tried to implement AUTH in a real-world scenario?

When receiving AUTH, my client prototype currently just does `console.log('Oh noes AUTH!')` and drops connection lol. That's because there's no way to know how relays implement it. (Is this like this PR? or does the relay wants me to retry REQ after authentication or what?)

fiatjaf:
OK, I'm convinced.

PR #847, opened 2023-10-28
Change kind of community post from 1 to 72

fiatjaf:
This is a breaking change and we will all die, but hopefully it is minor enough that implementors can switch easily. And there are few enough implementations that it should be doable.

Also implementations can keep reading 1 and 72 simultaneously while creating just 72 from now on, maybe?

The thing is that having different kinds for different purposes is generally better.

### DIFF
--- a/72.md
+++ b/72.md
@@ -48,7 +48,7 @@ Any Nostr event can be a post request. Clients MUST add the community's `a` tag 
   "id": "<32-bytes lowercase hex-encoded SHA-256 of the the serialized event data>",
   "pubkey": "<32-bytes lowercase hex-encoded public key of the event creator>",
   "created_at": <Unix timestamp in seconds>,
-  "kind": 1,
+  "kind": 72,
   "tags": [
     ["a", "34550:<Community event author pubkey>:<d-identifier of the community>", "<Optional relay url>"],
   ],


fiatjaf:
@hzrd149 @lovvtide

vitorpamplona:
Duplicated from here: https://github.com/nostr-protocol/nips/pull/753

weex:
ACK, makes more sense than 11 as well.

hzrd149:
I think it makes sense to separate community posts from the common post event. however there are a few things that would need clarification.

What kind should replies be? should they still be kind 1 and follow NIP-10?

What about other event types? the way NIP-72 is written it make no mention of restrictions on event types that can be posted to a community (it only uses kind 1 as an example)
In noStrudel I've added the feature to post kind 6 (repost) events to communities, as a way for users to share other events that are outside of the community into the community.
Maybe it make sense to create a new event kind for "sharing" other events into a community?

@vitorpamplona #753 seems to be about community exclusive posts in addition to kind 1. this is similar but I think its more about not cluttering up the "main" kind 1 feed of other apps with community posts.

vitorpamplona:
> @vitorpamplona https://github.com/nostr-protocol/nips/pull/753 seems to be about community exclusive posts in addition to kind 1.

Yep, but many clients already have a specific kind for exclusive posts. So, I would just use that kind instead of the 11. And then remove the possibility for Kind 1s to be attached to communities.

hzrd149:
> I would just use that kind instead of the 11. And then remove the possibility for Kind 1s to be attached to communities.

Sorry I'm out of the loop, but what is kind 11? I don't see it mentioned in the other PR or this one.

vitorpamplona:
> > I would just use that kind instead of the 11. And then remove the possibility for Kind 1s to be attached to communities.
> 
> Sorry I'm out of the loop, but what is kind 11? I don't see it mentioned in the other PR or this one.

Sorry, kind 72 (there was another issue that wanted to use kind 11 for exclusive posts). My point is that `kind:4549` is already been used. So, we should just go with them.

hzrd149:
Agreed, whether we use `kind:72` or `kind:4549` both achieve the same goal of an exclusive community post. 
However after reading through #753 more I think it introduces extra stuff that is not necessary in my opinion. more specifically the requirement for a "k" tag as a alternative kind for the event.
This would introduce an extra lever of complexity for clients when displaying an event. instead of just checking the `kind` they would also have to check for the "k" tag.

PR #851, opened 2023-10-29
NIP 49: Nostr Wallet Connect Connect

benthecarman:
There are a lot of technical and UX problems with NWC connection strings, this defines a better way to establish a wallet connection.

Naming and wording can definitely be improved, putting up for review now as I work on an implementation.


49.md
```diff
@@ -0,0 +1,83 @@
...
+Finally, Nostr Wallet Connect requires the wallet service to generate both key pairs and the service to use these. This
+limits the key management strategies that a service can use and often leads to suboptimal key management.
+
+This NIP proposes a new protocol that solves these problems by having the client and service generate a NWC connection
+URI together. This URI is then used to connect the client and service.
+
+## Theory of Operation
+
+1. The user wants to connect their wallet to a service. The service provides a QR code or other means of connecting to
+   the service. The QR code contains a `nip49://` URI with a `service_key` parameter that contains the service's public
```
vitorpamplona, line 28: `nip49://` or `nostr+walletauth://` (used below)? Are they the same? Or are two URIs being defined?

49.md
```diff
@@ -0,0 +1,83 @@
...
+## Specification
+
+### URI Format
+
+The **service** generates this connection URI with protocol `nip49:` and base path it's hex-encoded `pubkey` with the
+following query string parameters:
+
+- `relay` Required. URL of the relay where the **wallet service** is connected and will be listening for events. May be
+  more than one.
+- `required_commands` Required. A space-separated list of commands that the **wallet service** requires. The **client**
```
vitorpamplona, line 45: It's strange to require commands from the client. From the list of possible commands listed below, I don't see why it would require any of them. But I may not be understanding what this is about.

49.md
```diff
@@ -0,0 +1,83 @@
...
+  MUST NOT
+  connect if it does not support all of these permissions.
+- `optional_commands` Optional. A space-separated list of commands that the **wallet service** can enable to add
+  additional functionality. The **client** MAY ignore these.
+- `budget` Optional. A budget that the **wallet service** will use to limit the amount of funds that the **service** can
```
vitorpamplona, line 50: Is this budget related to a specific account the service manages or is it a max budget for all accounts? The account doesn't seem to be included in the URI, so I am confused. I am not sure if the client would care about a budget of all accounts.

49.md
```diff
@@ -0,0 +1,83 @@
...
+
+#### Example URI
+
+```sh
+nostr+walletauth://b889ff5b1513b641e2a139f661a661364979c5beee91842f8f0ef42ab558e9d4?relay=wss%3A%2F%2Frelay.damus.io&required_commands=pay_invoice%20pay_keysend%20make_invoice%20lookup_invoice&optional_commands=list_invoices%20list_payments&budget=10000%2Fday
+```
+
+### Event
+
+This NIP only uses one event kind `13193`. We use a replaceable event to allow for the wallet and service to renegotiate
```
vitorpamplona, line 62: Who writes this event? And how does this "negotiation" work? What messages go back and forth? What's the end goal of this negotiation? 

Maybe an example of all messages going back and forth can clarify this

49.md
```diff
@@ -0,0 +1,83 @@
...
+        "make_invoice",
+        "lookup_invoice",
+        "list_invoices",
+        "list_payments",
+    ],
+    "public_key": "b64c553a1f9f540f08890cc11c2a20d76fbebdc8d84a96263127526e2b136532" // hex-encoded public key from the wallet
+}
+```
+
+After the service receives this event, both the wallet and service should be able to communicate using NWC.
```
vitorpamplona, line 83: I am not sure what this is adding on top of the usual "nostr+walletconnect" URI.

vitorpamplona:
I am confused about the name Auth. The current text doesn't seem to define Authentication mechanisms for clients to follow. Is it coming at a later date or is this just about declaring and identifying compatibility of available commands?

49.md
benthecarman: woops, missed that

49.md
benthecarman: If I am linking my wallet to a service, the service is going to know which commands it is going to use. This way when the wallet creates the NWC connection it can permission it correctly to only support the necessary commands.

vitorpamplona: Ohh so, these are commands the client is requesting from the service and not the other way around. That makes more sense.

benthecarman:
@vitorpamplona auth is the best name someone has told me.

I think you're not understanding what the purpose of this is. This is for when you want to connect your wallet to a service (something like ZapplePay). Currently, pasting nwc strings causes too much friction and has a lot of downsides that I mention in the intro of the NIP.

It would be better if the user can just scan a QR from their wallet to connect it to the service instead of having proprietary buttons for every wallet or requiring the user to create the connection themself and paste the string into the service.

What we want is a user to know exactly what they are authorizing in their wallet, similar to what you get when you authorize a service with a github account, you get a list of permissions you are giving up in the connection.

image

49.md
benthecarman: this is the budget for the single connection, it is in the uri

vitorpamplona:
> I think you're not understanding what the purpose of this is.

Agree. This whole thing is very confusing. I suggest reviewing the terms and mentions to a "Wallet", the "Nostr Client", "NWC Service" and any other "Service", like ZapplePay (I suppose you are not calling ZapplePay a Nostr client); and clearly identify who does what, when. A full example of a complete setup between all these elements and where each string should be used could be very helpful. 

> This is for when you want to connect your wallet to a service (something like ZapplePay)

Ohhhhh... that should be the first line of this NIP, then. It's really not clear that this is between the Lightning Wallet and the Wallet Connect service. 

> Currently, pasting nwc strings causes too much friction and has a lot of downsides that I mention in the intro of the NIP.

See, now I am confused again. Why you are using NWC strings to connect to the wallet? Those are for the Nostr client only. 

> What we want is a user to know exactly what they are authorizing in their wallet

Is a NIP really the best place to document this if this is between the Wallet and the Service?

benthecarman:
I removed all mentions of nostr clients, I think this is what you're getting hung up on. Nostr is not limited to social media use cases and I think that is what you're getting hung up on.

benthecarman:
> See, now I am confused again. Why you are using NWC strings to connect to the wallet? Those are for the Nostr client only.

That is the problem today, is you have to generate the nwc from the wallet and take it to the app you want to connect to, it makes it very confusing to users (people often have a very hard time connecting to Amethyst from our experience)

> Is a NIP really the best place to document this if this is between the Wallet and the Service?

Yes, this should be the standard way to setup a NWC

fiatjaf:
Why not add this as an optional pre-step to NIP-47?

vitorpamplona:
It's about your terms, not the social media clients. You are mixing them together or using them interchangeably, which makes it confusing. Probably because on Mutiny they are together, but those are all separate systems for everyone else. 

> That is the problem today, is you have to generate the nwc from the wallet 

See, you did it again. It's not from the Wallet (the lightning wallet doesn't even know nostr exists), it's from the NWC service. 

If this PR is between the nostr client and the NWC service, then I go back to my original impression. There is no auth happening here. You are just making a QR to declare which commands are available in the NWC service. The client must read the QR, and somehow send the request for the commands it will use. Then the NWC displays on screen the permission request and if approved, returns another uri to be pasted in the app. 

Did I get this right?

benthecarman:
> Why not add this as an optional pre-step to NIP-47?

I guess this could be done. It seems I need to get alby to implement anything to get merged into the NIP and they've been really slow.

This flow isn't strictly required for NWC so I think it makes sense as a separate NIP.

benthecarman:
>  It's not from the Wallet (the lightning wallet doesn't even know nostr exists), it's from the NWC service.

If the wallet doesn't know nostr exists then payments can't happen. Frankly, it doesn't seem like you understand how NWC works. You keep saying "nostr client" but this term is way to broad, everyone is a nostr client, so this term means nothing.

I think I make it pretty clear, there is a wallet and a service. The wallet is someone like alby, mutiny, nostr-wallet-connect-lnd, something that receives nwc requests and makes payments. The service is someone like Amethyst, Damus, Zapple Pay, who makes requests to the wallet.

vitorpamplona:
Well, you are definitely the first one to ever call Amethyst a "service". And now I am more confused than ever. 

I don't know who does what in this PR.

benthecarman:
Renamed service to app and added a terms section.

vitorpamplona:
Much better text! Thanks! 

So, the app generates the QR and the Wallet or NWC service now needs a QR Reader. 

The issue that I see now is that the app doesn't know beforehand which relay the wallet is using. So generating a QR with the relay string in it is not possible. Or is this a separate relay that should not be used for Payment Requests in NIP-47? 

What happens after the user grants permissions in the Wallet? Does it generate the usual NIP-47 QR? If so, then the QR will contain the private relay the NWC is using and the app can connect to it. That can work.

benthecarman:
> The issue that I see now is that the app doesn't know beforehand which relay the wallet is using. So generating a QR with the relay string in it is not possible. Or is this a separate relay that should not be used to send Payment Requests in NIP-47?

The app picks the relay, so the wallet can just use that relay. This is the relay used throughout.

> What happens after the user grants permissions in the Wallet? Does it generate the usual NIP-47 QR? If so, then the QR will contain the private relay the NWC is using and the app can connect to it. That can work.

No, the whole point is to get around the NIP 47 connection string. After the wallet grants permission both sides will have the information they need to do full NIP 47 communication

vitorpamplona:
> After the wallet grants permission both sides will have the information they need to do full NIP 47 communication

How do I get the secret if the NWC generated one for me? What if the NWC only operates on their private relay? Is there a way to send that info back to the app.

vitorpamplona:
From the URI, The wallet doesn't know which PubKey is requesting access. So, how can It know which key is the app using?

benthecarman:
It is the first part of the URI??

> The **app** generates this connection URI with protocol `nostr+walletauth:` and base path it's hex-encoded `pubkey`

vitorpamplona:
Ohh, I see. I thought that would be the users main key so that the permission screen can bring the users picture. 

I would the key is from the secret. So the wallet app can't see which nostr user is requesting access.

benthecarman:
> So the wallet app can't see which nostr user is requesting access.

I guess we could, would need a signature as well then otherwise you could put any npub in there, which is going to make the QR really big. The user should know where this is coming from, they scanned it from their wallet.

vitorpamplona:
> I guess we could, would need a signature as well then otherwise you could put any npub in there, which is going to make the QR really big. The user should know where this is coming from, they scanned it from their wallet.

It builds confidence to let the user know they are doing the right thing and no one got in between. In the same way we verify the address and amount in Bitcoin invoices even though we created it ourselves, the requester's info should also be verifiable by the user in the permission's screen maximize UI goodness.

benthecarman:
> This can also be used as a QR.

The problem is this means you have to start in the wallet. So you have to go back and forth many times which is terrible UX and needless.

>  I do not see the issue with users not understanding their connection URI's confidentiality, since it is the wallet's responsibility to inform that this is sensitive information that can get their funds stolen.

You can inform users all you want but many are stupid, would be better to not have them handle these at all.

> There's an info event...

The info event is created after the connection is made. You could have the app check if the user created a correct NWC but that just leads to tedious back and forth.

> How?

For things like Zapple Pay, I have to use the keys the user generates for me and store them in a database. With this I can use an extended private key and just store a derivation path per user so I don't need to keep key material in a database.

benthecarman:
> A database is in most production scenarios not any less secure than the storage for your key material (the machine your server runs on). It is way more likely an app flaw would compromise secrets than a database one in my opinion.

For now, yeah. But eventually people could be using things like HSMs

49.md
```diff
@@ -0,0 +1,88 @@
...
+
+#### Example URI
+
+```sh
+nostr+walletauth://b889ff5b1513b641e2a139f661a661364979c5beee91842f8f0ef42ab558e9d4?relay=wss%3A%2F%2Frelay.damus.io&required_commands=pay_invoice%20pay_keysend%20make_invoice%20lookup_invoice&optional_commands=list_invoices%20list_payments&budget=10000%2Fday
+```
+
+### Event
+
+This NIP only uses one event kind `13193`. We use a replaceable event to allow for the wallet and app to renegotiate
```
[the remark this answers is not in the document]
benthecarman: > There is no requirement saying that one npub must host only one wallet.

yeah but that shouldn't be done, can add a note about that

> This should be a replaceable event.

it is one

> There is also the issue that an attacker can monitor for all events of this kind, and then create their own event to trick the client into thinking that the attacker pubkey is the wallet in a denial of service attack. This can be fixed by including a secret in the connection URL, and sending that in the encrypted payload

Good idea

kiwiidb (changes requested):
Overall, I think this is not a bad idea and it would be useful for future work regarding zap subscriptions/autozap (services implementing cronjobs for zaps with regards to Pablo's [proposal](https://nostr.band/note1myutdchnmwd5u93mf02p9747wnclqhn7y0ulkd50u48dmn4hhgwqr5358e)).

However, I do kind of agree with the [points](https://github.com/nostr-protocol/nips/pull/792#issuecomment-1742012546) raised in the other PR in that this can be a potential footgun. I feel that the traditional apps that already have access to the users private key (Damus/Amethyst) are still better of using a wallet-generated secret and receiving it back through a deeplink. These apps themselves linking to the wallets with a standardized link scheme is fine obviously. 

However like I said, I do think this is great for 3rd party services / subscription automations, if the wallet has enough measures in place to protect against phishing.

kiwiidb:
Oh and I also think this should be an optional pre-step to nip-47 like fiatjaf said. Regarding that you are waiting for us to implement something before something can be merged I don't get it. You and us are both wallet developers. So if you develop it, and you find a nostr app to develop the other side (eg. your own zapplepay), you have a working setup that doesn't require us to develop anything, right?

49.md
```diff
@@ -0,0 +1,103 @@
...
+   use NWC to communicate.
+
+## Specification
+
+### URI Format
+
+The **app** generates this connection URI with protocol `nostr+walletauth:` and base path it's hex-encoded `pubkey`
+with the following query string parameters:
+
+- `relay` Required. URL of the relay where the **wallet** is connected and will be listening for events. May be
```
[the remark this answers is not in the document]
benthecarman: good catch

benthecarman:
Changed to parameterized replaceable events so keys can be reused by a wallet.

benthecarman:
> However, I do kind of agree with the https://github.com/nostr-protocol/nips/pull/792#issuecomment-1742012546 raised in the other PR in that this can be a potential footgun. 

This doesn't really make sense and was refuted in the further comments.

> I feel that the traditional apps that already have access to the users private key (Damus/Amethyst) are still better of using a wallet-generated secret and receiving it back through a deeplink. These apps themselves linking to the wallets with a standardized link scheme is fine obviously.

While they only use it for one-tap zaps I can see the case, but as apps use more commands they will run into this renegotiate problem and nwc uris will not hold up.

benthecarman:
This could go as an optional pre-step to nip 47, but this nip is already about the same length. Imo since it is optional it makes more sense as a separate nip.

benthecarman:
The reason to use unique keys would be for privacy, so someone couldn't just count how many events are tied to your key and get a user count. However it does make the implementation much simpler and more efficient if you use a static key

nostr-wine:
Hello again. We are working on a NWC service and had a few more questions.

- The example URI uses nostr+walletauth: but I believe Mutiny currently requires nostr+walletauth:// instead. Which one should we use? 
- Why space separated lists for required_commands/optional_commands instead of commas?
- Instead of budget with max_amount/period, could we separate max_amount and period to reduce unnecessary parsing?

49.md
```diff
@@ -0,0 +1,100 @@
...
+
+The **app** generates this connection URI with protocol `nostr+walletauth:` and base path it's hex-encoded `pubkey`
+with the following query string parameters:
+
+- `relay` Required. URL of the relay where the **app** is connected and will be listening for events. May be
+  more than one.
+- `secret` Required. A random identifier that the **wallet** will use to identify the connection. Without this the
+  **app** cannot authenticate they are receiving a message from the user they want, it could be an attacker just sending
+  kind `33194` events to the **app**. This should be unique for each connection.
+- `required_commands` Required. A space-separated list of commands that the **wallet** requires. The **wallet**
```
jklein24, line 53:
```suggestion
- `required_commands` Required. A space-separated list of commands that the **app** requires from the **wallet**. The **wallet**
```

49.md
```diff
@@ -0,0 +1,100 @@
...
+  MUST NOT connect if it does not support all of these permissions.
+- `optional_commands` Optional. A space-separated list of commands that the **wallet** can enable to add additional
+  functionality. The **wallet** MAY ignore these.
+- `budget` Optional. A budget that the **wallet** will use to limit the amount of funds that the **app** can spend.
```
jklein24, line 57: I think having a budget is great, but should it really be up to the **app** to decide this budget? On one hand, it's nice for the app to be able to self-limit (I promise I'll only send 10 sats per day), but I think it might also be nice to give the wallet and even the user the ability to change this budget, like Alby allows today.

benthecarman: This is just a recommendation from the app, it is not required by the wallet and can be changed. You can see how we do this with mutiny + zapple pay if you're curious

nostr-wine:
nostr.wine now supports NWC via NIP-49 as a way to connect to customer wallets and send them invoices when their subscriptions are due. Thanks again Ben!

https://njump.nostr.wine/nevent1qqsz9w3mjvlk4dmuuaxvuqqn32ga95j4mj98upmyh7qhsn9kxhs5mrqzyq7cg2h7e40zj0egke38jvmsfglm3ns4825367g2ky0k5afdgjjz68tw0gf

Egge21M:
> - The example URI uses nostr+walletauth: but I believe Mutiny currently requires nostr+walletauth:// instead. Which one should we use? 

The same inconsistency exists in the NWC base spec, which is why I opened https://github.com/nostr-protocol/nips/pull/953.

":" is the correct way, as "//" indicates a authority as per RFC 3986. However most implementations that exists in the wild currently use "<scheme>://<pubkey>"

49.md
```diff
@@ -0,0 +1,99 @@
...
+This NIP only uses one event kind `33194`. We use a parameterized replaceable event to allow for the wallet and app to
+renegotiate the connection if needed and to allow for the wallet being able to look up the connection details later.
+
+The event SHOULD contain one `d` tag, containing the public key of the app from the URI.
+
+The content of the event is encrypted with [NIP04](04.md), and should be a JSON structure as follows:
+
+```jsonc
+{
+    "secret": "b8a30fafa48d4795b6c0eec169a383de", // string, the secret from the URI
```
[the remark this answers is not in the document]
benthecarman: This secret isn't a key and it should be unique for every connection. Added some clarification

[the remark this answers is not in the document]
vfiano: More of a pairing code than the NIP-47 secret. It might be good to use a different name to avoid confusion.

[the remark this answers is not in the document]
benthecarman: Yeah I wish I named this something else but there are already multiple clients using this in production today, it is too late to change

49.md
```diff
@@ -0,0 +1,100 @@
...
+
+As well, Nostr Wallet Connect has lots of different functions that a wallet may or may not support or an app may or may
+not need. This either runs into silent failures or requires the user to understand what functions their wallet supports
+and what functions the app needs.
+
+Finally, Nostr Wallet Connect requires the wallet to generate both key pairs and the app to use these. This limits the
+key management strategies that an app can use and often leads to suboptimal key management.
+
+This NIP proposes a new protocol that solves these problems by having the wallet and app generate a NWC connection URI
+together. This URI is then used to connect the wallet and app.
```
vfiano, line 23: This wording is a little confusing. They aren't making the exact same URI as NIP-47 since this proposal replaces symmetric encryption with public key.

benthecarman:
What is the status on getting this merged? I think there are now 4 implementations in the wild now

- mutiny
- zapple pay
- prisms
- nostr.wine

benthecarman:
Seems like 49 was taken so changed to 67

benthecarman:
image

fiatjaf:
Just help me understand: this is just an inversion of the connect flow of NIP-47?

NIP-46 defines both connect flows: from client to signer and from signer to client. After the initial connection everything is the same, which seems to be the case here.

If that is the case then I would like to read this more carefully and try to merge it into NIP-47.

benthecarman:
> If that is the case then I would like to read this more carefully and try to merge it into NIP-47.

Okay I can reformat it to do that

jklein24:
@benthecarman I really like this proposal and hope it's merged (either here or to NIP-47)! One question - do you think it might make sense for the confirmation event to include an optional budget as well? I know with Alby's NWC UI, the user can manually change the budget, etc. If the app requests a specific budget, but the wallet/user grants a different one, does the app need to know that? I realize that with NWC, the app isn't made aware of a budget restriction except for via `QUOTA_EXCEEDED` errors, but in this case if the app specifically requests a budget, it seems more relevant.

67.md
```diff
@@ -0,0 +1,101 @@
...
+  kind `33194` events to the **app**. This should be unique for each connection. This is not a key and should be
+  different for every new URI.
+- `required_commands` Required. A space-separated list of commands that the **app** requires from the **wallet**. The
+  **wallet** MUST NOT connect if it does not support all of these permissions.
+- `optional_commands` Optional. A space-separated list of commands that the **wallet** can enable to add additional
+  functionality. The **wallet** MAY ignore these.
+- `budget` Optional. A budget that the **wallet** will use to limit the amount of funds that the **app** can spend.
+  The **app** MUST NOT spend more than this amount. The budget is a string of the form `max_amount/period`.
+  The `period` is one of `daily`, `weekly`, `monthly`, `yearly`. The `max_amount` is a satoshi amount.
+- `identity` Optional. The hex-encoded `pubkey` of the *app* that the **wallet** is connecting to. This is used to
```
jklein24, line 61: Why is this needed if it's also the base path of the URI? Is that a different app pubkey?

benthecarman: this is meant to be different the base path, this is used to socially tag the connection so you don't need to use the same keys you use for nwc that you use for your profile

benthecarman:
> do you think it might make sense for the confirmation event to include an optional budget as well?

yeah that's a good idea

benthecarman:
i have some free time next week, I'll try to restructure this to be an edit to NIP-47 like @fiatjaf suggested

vitorpamplona:
Are we doing this? Can we clear all debates over this and move to either implement/merge or close it?

PR #852, opened 2023-10-30
NIP-100

kehiy:
This NIP specifies a default port for relay nodes.

> NOTE: The reason for choosing port 444 is it's open and not used, also some relay implementations use it now.

@melvincarvalho

### DIFF
--- a/100.md
+++ b/100.md
@@ -10,4 +10,5 @@ This document standardizes the default network port for Nostr relays, each relay
 
 # motivation
 
-This standard can help developers for selecting a default port. also users know a `nostr://` url probably is on port `4444` and the relay nodes don't need to provide port by default when providing IP address.
+This standard can help developers for selecting a default port. also users know a `nostr:` url probably is on port `4444` and the relay nodes don't need to provide port by default when providing IP address.
+Clients also can consider it as default port.
--- a/100.md
+++ b/100.md
@@ -1,0 +1,13 @@
+NIP-100
+======
+
+Default Relay Port Standard
+-----------------
+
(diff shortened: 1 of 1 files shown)


melvincarvalho:
LGTM:  

Minor nit:  the relay protocol is ws(s):// rather than nostr://

kehiy:
@melvincarvalho fixed, I used a better way to write the part.

melvincarvalho:
Fonstr already uses port 4444 as default.  I dont think many people will run 2 relays so having a default port could be helpful, especially on mobile devices where you might not always be able to figure out which port runs what, due to restrictions.

https://dev.to/melvincarvalho/run-a-nostr-relay-on-your-phone-with-termux-and-fonstr-4cmg

fiatjaf:
The default port is already 80 for `ws://` and 443 for `wss://`, what is the purpose of this?

derekross:
I don't like this. Why not just use 443? 444 could be blocked at the router level, making relays easily censored across the world. You can't block 443 without blocking the world wide web, making Nostr itself much more resilient.

jspeigner:
Having a specific port will make it very easy to block access on that port. 443 is not able to be block since most everything is using it.

weex:
There's a typo of `4444` in the draft, also it'd be good if someone can fix the title of this PR so it's more descriptive.

melvincarvalho:
80 and 443 are natural ports, but require root access.  Often a relay will not run as root, e.g. on a phone, and sit behind a reverse proxy for example.

PR #854, opened 2023-10-31
Add Twitter verification caveat to NIP-39.

ibz:


mattn:
Do you expect to take over the previously used verification of proof when a user changes their twitter account name?

ibz:
Not sure I fully understand the question. But the verification should (in theory) be done by clients every time the profile is displayed. In practice, the result of the verification could be cached, if successful (the downside I see with that is that somebody could steal your Twitter account at a later moment in time, but then - would they be stupid enough to delete the proof?). If the Twitter identity changed, the cached verification would of course become invalid and would have to be redone.

fiatjaf:
That is a good point. Example here: https://twitter.com/fiatjaf/status/1719400674008682686

But it's not clear how clients are even supposed to check these Twitter proofs, but this caveat is wrong because there is no _actual_ redirect. Instead the URL seems to change due to some JavaScripty thing.

ibz:
Indeed, it is very hard to check the proofs without running the Javascript on the page because the page returned by Twitter does not even contain the tweet content!

At Plebeian Market we implemented this using `arsenic`, which is something like Selenium... browser automation basically.

Of course this only works if you have a backend. If your app is all client-side it becomes impossible. But then, if your app is all Javascript running the the browser you can't even make the GET request to Twitter due to CORS.

All in all, I found this NIP the hardest to implement, although it looks trivial at the first glance. But we think it is important to show the other identities on the user's profile. So we did it...

PR #855, opened 2023-10-31
NIP-25 addition: Client may specify suggested reactions

timrmaurer:
A client may specify suggested reactions (emojis, or custom emoji shortcodes) in the tags of any event. Clients displaying the event may prompt users to use the suggested reactions to react.

### DIFF
--- a/25.md
+++ b/25.md
@@ -74,3 +74,24 @@ content as an emoji if shortcode is specified.
 ```
 
 The content can be set only one `:shortcode:`. And emoji tag should be one.
+
+Reaction Suggestions
+--------------------
+
+A client may specify suggested reactions in the tags of any event. These may be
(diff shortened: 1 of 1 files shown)


AsaiToshiya:
Out of curiosity, can you please share what is the specific use case that requires this?

vitorpamplona:
I like this. It can be used for voting. 

I just don't know what these two are referring to: 

```json
    ["reaction", ":joy:"],
    ["reaction", ":man-shrugging:"],
```

They seem to be custom emojis, but without the image url

mattn:
Yes, I also like this  since this could be used for voting.

AsaiToshiya:
Sounds good. I look forward to it being implemented.

PR #856, opened 2023-10-31
For improved SEO of web clients, add canonical tag to NIPs 23 and 52.

shawndearmond:
When displaying Long-form Content (NIP-23) and Calendar Events (NIP-52) on web clients, there is a risk that search engines will penalize these web clients (and the original publisher of the article) due to the "duplicate content penalty". 

A simple way around this is for Nostr events to include a `canonical` tag with a single url. Then web clients can easily add the `<link rel="canonical" href="URL_HERE" />` tag to their web pages when displaying a persistent page for an article or calendar event. This shouldn't be used when displaying lots of articles or calendar events in a list, of course, just when there's a page that displays the article or calendar event by itself or as the top-most parent of the thread.

The `r` tag is insufficient for this as there could be multiple `r` tags, and they're meant to display links within the article, not necessarily the canonical url of the article itself. To mix them together would be confusing.

### DIFF
--- a/23.md
+++ b/23.md
@@ -28,6 +28,7 @@ Other metadata fields can be added as tags to the event as necessary. Here we st
 - `"image"`, for a URL pointing to an image to be shown along with the title
 - `"summary"`, for the article summary
 - `"published_at"`, for the timestamp in unix seconds (stringified) of the first time the article was published
+- `"canonical"`, for the canonical url of the article. Web clients SHOULD include the `<link rel="canonical" href="URL_HERE" />` tag when returning a persistent page for the article.
 
 ### Editability
 
@@ -55,6 +56,7 @@ References to other Nostr notes, articles or profiles must be made according to 
     ["t", "placeholder"],
     ["e", "b3e392b11f5d4f28321cedd09303a748acfd0487aea5a7450b3481c60b6e4f87", "wss://relay.example.com"],
     ["a", "30023:a695f6b60119d9521934a691347d9f78e8770b56da16bb255ee286ddf9fda919:ipsum", "wss://relay.nostr.org"]
+    ["canonical", "https://example.com/preferred-url-here/"],
   ],
   "pubkey": "...",
   "id": "..."
--- a/52.md
+++ b/52.md
@@ -30,6 +30,7 @@ The list of tags are as follows:
 * `start` (required) inclusive start date in ISO 8601 format (YYYY-MM-DD). Must be less than `end`, if it exists.
 * `end` (optional) exclusive end date in ISO 8601 format (YYYY-MM-DD). If omitted, the calendar event ends on the same date as `start`.
 * `location` (optional) location of the calendar event. e.g. address, GPS coordinates, meeting room name, link to video call
+* `canonical` (optional) canonical url of the event. Web clients SHOULD include the `<link rel="canonical" href="URL_HERE" />` tag when returning a persistent page for the event.
 * `g` (optional) [geohash](https://en.wikipedia.org/wiki/Geohash) to associate calendar event with a searchable physical location
 * `p` (optional, repeated) 32-bytes hex pubkey of a participant, optional recommended relay URL, and participant's role in the meeting
 * `t` (optional, repeated) hashtag to categorize calendar event
@@ -66,6 +67,9 @@ The list of tags are as follows:
     // Reference links
     ["r", "<url>"],
     ["r", "<url>"]
+
+    // Canonical URL
+    ["canonical", "<url>"],
   ]
 }
 ```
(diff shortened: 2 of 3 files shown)


tyiu:
I'm not sure if I understand why this change is needed. Would the web client be able to add the canonical link without the `r` tag tag on the Nostr event?

imo, in general, we should try to have Nostr events to not be tied to any single webpage, because that's centralizing.

Let me know if I'm misunderstanding anything.

shawndearmond:
The web client can add the canonical tag to the page entirely separate from the `r` tags. The client can still list `r` links or unfurled urls at the end of the article or wherever they want to do.

I totally understand wanting Nostr events to not be tied to a single web page, but articles (and to a lesser extent calendar events) are indexed by search engines and page rank is affected when the same rich content is being published on multiple different web sites (clients). 

Usually, the source of the article isn't Nostr itself, but rather a news web site, and they would be more inclined to publish their articles in their entirety on Nostr if they knew that the canonical tag was in place so that the article on their web site, and their article pages on Nostr web clients, don't get hit with the duplicate content penalty.

fiatjaf:
Is it really true that if I copy some website's content Google will penalize that website?

shawndearmond:
Yes, it's to prevent web site scraping. 

If the full copy of a news article is exactly the same on multiple different web sites, and Google can't tell which page is the original (canonical) one, it can affect the search ranking of all of them. 

So, if all the Nostr web clients are showing the same NIP-23 article in its entirety, plus the original (on, say, the author's blog web site), Google may not know which one to rank higher. 

The solution is for the web site to add that `canonical` tag in the header of that page that points to the "original" article.

This only matters when there is a persistent page for an article. Obviously the home page of a Nostr web client will be an ever-changing list of new content. That doesn't need the canonical tag.

Similarly, and this is up to the client, but maybe if a NIP-23 article was written on snort.social, Snort may want to add its own page as the Canonical URL, to protect against the lower ranking when other Nostr web clients "scrape" the content and display it on their own sites.

Sebastix:
I'm not a SEO expert but does the duplicate content penalty really exist?

https://developers.google.com/search/docs/crawling-indexing/consolidate-duplicate-urls

https://developers.google.com/search/blog/2008/09/demystifying-duplicate-content-penalty (2008, but still relevant today?)

shawndearmond:
yeah, those articles you link to are still correct. 

It’s known as the Duplicate Content Penalty, but it’s semantically a little bit of a misnomer.  It’s not so much that a “penalty” (as in Google docks you points) as it is avoidable search engine confusion when the same content exists on many different web sites. With many Nostr web clients all pulling the same content, this will happen and the “correct” thing to do, according to that first link, is to use the canonical tag.

shawndearmond:
It also says in that article that without the canonical tag, “Google will identify which version of the URL is objectively the best version to show to users in Search”

Which we don’t want. The author of the Nostr event should be the one to decide the best version to have in Search.

Sebastix:
It reminds me of https://indieweb.org/original_post_link where the same problem arises with POSSE

erskingardner:
Is it possible that clients could just use `nostr:<event_id>` as the value for the canonical attribute? Search engines probably wouldn't know what to do with that but might not penalize. 

I'm with @tyiu, I don't like the idea that clients or users would have to pick a canonical URL/client for their content.

Sebastix:
I'm curious what the original story is of creating the `canonical` tag back in the days.
I can imagine that some lobby of search engines were involved. */me starts searching*
https://en.wikipedia.org/wiki/Canonical_link_element

> In February 2009, [Google](https://en.wikipedia.org/wiki/Google), [Yahoo](https://en.wikipedia.org/wiki/Yahoo) and [Microsoft](https://en.wikipedia.org/wiki/Microsoft) announced support for the canonical link element, which can be inserted into the [<head> section](https://en.wikipedia.org/wiki/HTML_element#head) of a web page, to allow webmasters to prevent these issues.[[7]](https://en.wikipedia.org/wiki/Canonical_link_element#cite_note-7) The canonical link element helps webmasters make clear to the search engines which page should be credited as the original.

With that said, I think it's up to those search engines (who broke the web partial) to find a solution of 'duplicate' content provided through the Nostr network. They will have to find out there is no single source of truth which can let to centralization. On the long term when Nostr is succeeding, other Nostr specific search clients will arise (which are not by default related to webpages on domains).

dearmond:
Okay, I agree with you. And I realized that I can achieve what I need to using normal `r` tags like so:

` ["r", "https://example.com/preferred-url-here/", "canonical"]`

That way, the link will always be there for all clients, whether they use the canonical tag or not.

shawndearmond:
Okay, I agree with you. And I realized that I can achieve what I need to using normal r tags like so:

 ["r", "https://example.com/preferred-url-here/", "canonical"]

That way, the link will always be there for all clients, whether they use the canonical tag or not.

PR #860, opened 2023-11-01
Adding encryption NIP-90 Data Vending Machine

starbackr-dev:
We are not able to implement NIP-90 without encrypting the generative AI parameters and results. Currently the users are using kind 4 event to communicate with AI agents. This is not useful since kind 4 events are not visible within the DVM eco system. The change requested follows the same logic as NIP-51 where the private lists are encrypted and stored on the content field. 

link to changes: https://github.com/starbackr-dev/nips/blob/patch-2/90.md

### DIFF
--- a/90.md
+++ b/90.md
@@ -91,7 +91,7 @@ This param data will be encrypted and added to the `content` field and `p` tag s
 "content": "BE2Y4xvS6HIY7TozIgbEl3sAHkdZoXyLRRkZv4fLPh3R7LtviLKAJM5qpkC7D6VtMbgIt4iNcMpLtpo...",
     "tags": [
         [`p`, `04f74530a6ede6b24731b976b8e78fb449ea61f40ff10e3d869a3030c4edc91f`],
-        [ `param`, `encrypted`, true]
+        [`encrypted`]
     ]
 
 
@@ -124,6 +124,23 @@ Service providers publish job results, providing the output of the job result. T
 ## Encrypted Output
 
 If the request has encrypted params, then output should be encrypted and placed in  `content` field with `p` tag. If the output is encrypted, then avoid including `i` tag with input-data as clear text. 
+Add a tag encrypted to mark the output content as `encrypted`
+```json
+{
+    "pubkey": "<service-provider pubkey>",
+    "content": "<encrypted payload>",
+    "kind": 6xxx,
+    "tags": [
+        [ "request", "<job-request>" ],
+        [ "e", "<job-request-id>", "<relay-hint>" ],
+        [ "p", "<customer's-pubkey>" ],
+        [ "amount", "requested-payment-amount", "<optional-bolt11>" ],
+        [`encrypted`]
+
+    ]
+}
+```
+ 
 
 ## Job feedback
 Service providers can give feedback about a job back to the customer.
--- a/90.md
+++ b/90.md
@@ -79,8 +79,8 @@ If the user wants to keep the input parameters a secret, they can encrypt the `i
     [ "param", "temperature", "0.5" ],
     [ "param", "top-k", "50" ],
     [ "param", "top-p", "0.7" ],
-    [ "param", "frequency_penalty", "1" ],
-    [ "param", "output", "encrypted"]
+    [ "param", "frequency_penalty", "1" ]
+    
 ]
 
 ```
@@ -90,8 +90,8 @@ This param data will be encrypted and added to the `content` field and `p` tag s
 ```
 "content": "BE2Y4xvS6HIY7TozIgbEl3sAHkdZoXyLRRkZv4fLPh3R7LtviLKAJM5qpkC7D6VtMbgIt4iNcMpLtpo...",
     "tags": [
-        [`p`, `04f74530a6ede6b24731b976b8e78fb449ea61f40ff10e3d869a3030c4edc91f`]
-
+        [`p`, `04f74530a6ede6b24731b976b8e78fb449ea61f40ff10e3d869a3030c4edc91f`],
+        [ `param`, `encrypted`, true]
     ]
 
 
@@ -120,6 +120,10 @@ Service providers publish job results, providing the output of the job result. T
 * `request`: The job request event stringified-JSON.
 * `amount`: millisats that the Service Provider is requesting to be paid. An optional third value can be a bolt11 invoice.
 * `i`: The original input(s) specified in the request.
+
+## Encrypted Output
+
+If the request has encrypted params, then output should be encrypted and placed in  `content` field with `p` tag. If the output is encrypted, then avoid including `i` tag with input-data as clear text. 
 
 ## Job feedback
 Service providers can give feedback about a job back to the customer.
--- a/90.md
+++ b/90.md
@@ -73,7 +73,7 @@ If the user wants to keep the input parameters a secret, they can encrypt the `i
 
 ```json
 [
-    [ "i", "what is the capital of France? ", "prompt" ],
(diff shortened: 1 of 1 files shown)


starbackr-dev:
@pablof7z Please review this when you get a chance.

pablof7z:
Hey, sorry, been impossible to look at this until now.

I'm unclear on what exactly gets encrypted. Is it nip04.encrypt(json_stringified([ [ "i" ...] ]))?

starbackr-dev:
> Hey, sorry, been impossible to look at this until now.

Yes. understand you were in Tokyo.. just keeping the pressure on..! 
> 
> I'm unclear on what exactly gets encrypted. Is it nip04.encrypt(json_stringified([ [ "i" ...] ]))?

We do not want to dictate what fields need to be encrypted. Currently, in all 5xxx events content field is empty. So if a p tag is present and content field is not empty, then decrypt content field similar to kind 4 which will give all the encrypted parameters. It is up to the client to encrypt any or all parameters for the input. 

Same is true in 6xxx events. try to decrypt content field to see the payload, if decryption fails, then it is not encrypted. 

Please read NIP-51,  which follows the exact same logic for public and private lists. Below is the extract. 

> A "list" event is defined as having a list of public and/or private tags. Public tags will be listed in the event tags. Private tags will be encrypted in the event content. Encryption for private tags will use [NIP-04 - Encrypted Direct Message](https://github.com/nostr-protocol/nips/blob/master/04.md) encryption, using the list author's private and public key for the shared secret.

pablof7z:
> > Hey, sorry, been impossible to look at this until now.
> 
> Yes. understand you were in Tokyo.. just keeping the pressure on..!
> 
> > I'm unclear on what exactly gets encrypted. Is it nip04.encrypt(json_stringified([ [ "i" ...] ]))?
> 
> We do not want to dictate what fields need to be encrypted. Currently, in all 5xxx events content field is empty. So if a p tag is present and content field is not empty, then decrypt content field similar to kind 4 which will give all the encrypted parameters. It is up to the client to encrypt any or all parameters for the input.
> 
> Same is true in 6xxx events. try to decrypt content field to see the payload, if decryption fails, then it is not encrypted.
> 
> Please read NIP-51, which follows the exact same logic for public and private lists. Below is the extract.
> 
> > A "list" event is defined as having a list of public and/or private tags. Public tags will be listed in the event tags. Private tags will be encrypted in the event content. Encryption for private tags will use [NIP-04 - Encrypted Direct Message](https://github.com/nostr-protocol/nips/blob/master/04.md) encryption, using the list author's private and public key for the shared secret.

oh, jesus, I must have been very tired last night when I reviewed it in the plane; it's literally [just there](https://github.com/starbackr-dev/nips/blob/patch-2/90.md#encrypted-params) I just wanted to have the stuff that needed to be encrypted explicitly on the NIP and it was there already!

I didn't see the result stuff encrypted, is it there too and I missed it again?

One thing I would change is keep the `output` untouched since its already used for other things. If the job request comes encrypted it should just be returned encrypted, or perhaps we should have a simple `[ "encrypted" ]` flag on the job request when the user wants the result encrypted.

starbackr-dev:
> I didn't see the result stuff encrypted, is it there too and I missed it again?

Yes. If the job request comes encrypted it should just be returned encrypted on the content field. I did add this to the PR to make it explicit and to avoid adding 'i' tag in the result. 

> One thing I would change is keep the output untouched since its already used for other things. If the job request comes encrypted it should just be returned encrypted, or perhaps we should have a simple [ "encrypted" ] flag on the job request when the user wants the result encrypted.

agree with not changing 'output'. So changed the tag to 'encrypted' true | false 

[ "param", "encrypted", true]

pablof7z:
We should probably not use `param` for this, as this is not a param for the job but rather *about* the job. I.e. what happens if some job request tries to use the param `encrypted` for something?

I think this should literally just be an `[ "encrypted" ]` tag, no second argument and it should always be required if the job is encrypted instead of trying to decrypt if it was p-tagged; I think making this explicit is better in both job request and job result is better.

starbackr-dev:
> We should probably not use param for this, as this is not a param for the job but rather about the job. I.e. what happens if some job request tries to use the param encrypted for something?

agree and changed the PR. 

> I think this should literally just be an [ "encrypted" ] tag, no second argument and it should always be required if the job is encrypted instead of trying to decrypt if it was p-tagged; I think making this explicit is better in both job request and job result is better.



Here's the summary

Added [`encrypted`] to both 5xxx and 6xxx events.
if `encrypted` tag present, then content field is encrypted with p tag. 
It is between service provider and user on what they encrypt.

90.md
```diff
@@ -90,6 +121,27 @@ Service providers publish job results, providing the output of the job result. T
...
+{
+    "pubkey": "<service-provider pubkey>",
+    "content": "<encrypted payload>",
+    "kind": 6xxx,
+    "tags": [
+        [ "request", "<job-request>" ],
+        [ "e", "<job-request-id>", "<relay-hint>" ],
+        [ "p", "<customer's-pubkey>" ],
+        [ "amount", "requested-payment-amount", "<optional-bolt11>" ],
+        [`encrypted`]
```
pablof7z, line 138: This should be " instead `

90.md
```diff
@@ -67,6 +67,37 @@ All tags are optional.
...
+]
+
+```
+
+This param data will be encrypted and added to the `content` field and `p` tag should be present 
+
+```
+"content": "BE2Y4xvS6HIY7TozIgbEl3sAHkdZoXyLRRkZv4fLPh3R7LtviLKAJM5qpkC7D6VtMbgIt4iNcMpLtpo...",
+    "tags": [
+        [`p`, `04f74530a6ede6b24731b976b8e78fb449ea61f40ff10e3d869a3030c4edc91f`],
```
pablof7z, line 93: Same, these should be ", not `

90.md
```diff
@@ -67,6 +67,37 @@ All tags are optional.
 * `relays`: List of relays where Service Providers SHOULD publish responses to
 * `p`: Service Providers the customer is interested in. Other SPs MIGHT still choose to process the job
 
+## Encrypted Params 
+
+If the user wants to keep the input parameters a secret, they can encrypt the `i` and `param` tags with the service provider's 'p' tag and add it to the content field. The user also indicates whether the output should be encrypted or not as one of the parameters.
```
pablof7z, line 72: This should mention explicitly that this is following the NIP-04 encryption scheme. Just bring in the same text that NIP-51 has:

> Encryption for private tags will use [NIP-04 - Encrypted Direct Message](https://github.com/nostr-protocol/nips/blob/master/04.md) encryption, using the list author's private and public key for the shared secret

90.md
```diff
@@ -90,6 +121,27 @@ Service providers publish job results, providing the output of the job result. T
 * `amount`: millisats that the Service Provider is requesting to be paid. An optional third value can be a bolt11 invoice.
 * `i`: The original input(s) specified in the request.
 
+## Encrypted Output
+
+If the request has encrypted params, then output should be encrypted and placed in  `content` field with `p` tag. If the output is encrypted, then avoid including `i` tag with input-data as clear text.
```
pablof7z, line 126: > with `p` tag.

What does this mean? Does it mean that the customer's pubkey should be `p` tagged? This is always the case so there's nothing special about it being encrypted or not; if that's what's meant here we should remove this since it makes it more confusing.

starbackr-dev:
ok. updated and addressed all the comments..!

pablof7z (approved):

PR #865, opened 2023-11-05
NIP-43: Bounties

ChristianChiarulli:
Demo site here: https://resolvr.io

This is a project we are working at over at resolvr.io, you can read more about how we can further improve the trust problems inherent with a bounty board here: https://github.com/Resolvr-io/Resolvr

### DIFF
--- a/117.md
+++ b/117.md
@@ -42,9 +42,13 @@ The following tags MAY be present:
 
 - A `context` tag to provide a link to a website or other resource that provides more information about the bounty.
 
-- A `p` tag of the applicant public key to track who is currently assigned to the bounty. **NOTE** This tag MUST be present when the bounty has the `assigned` status.
+The following tags MUST be present when the bounty is updated to `assigned` status:
 
-- An `application` tag with the stringified **Bounty Application Event**. **NOTE** This tag SHOULD be present when the bounty has the `assigned` status.
+- A `p` tag of the applicant public key to track who is currently assigned to the bounty.
+
+- An `e` tag to reference the **Bounty Application Event**.
+
+- An `application` tag with the stringified **Bounty Application Event**.
 
 - An arbitrary amount of `t` tags to provide a list of tags that can be used as categories to filter bounties.
 
--- a/117.md
+++ b/117.md
@@ -1,0 +1,100 @@
+NIP-117
+======
+
+Bounties
+------
+
+`draft` `optional` `author:ChristianChiarulli`
+
+## Abstract
+
+This NIP defines a standard for creating, applying to and assigning bounties.
+
+## Event Overview
+
+1. **Bounty Event:** defined as a regular event of kind 8050. This event is published by the creator of the bounty.
+
+2. **Bounty Application Event:** defined as a regular event of kind 8051. This event is published by an applicant and should reference the **Bounty Event** with a single `e` tag.
+
+3. **Bounty Assignment Event:** defined as a parameterized replaceable event of kind 30050. This event is published by the creator of the bounty after receiving at least one applicant and should reference the chosen applicant by the public key associated with the **Bounty Application Event**. This event may be updated to change the associated `p` tag to a different applicant. 
+
+## Bounty Event
+
+The `.content` of these events should be a string text in Markdown syntax similar to NIP-23.
+
+The following tags MUST be present:
+
+- A `title` tag to provide a title for the bounty.
+
+- A `reward` tag to provide a reward for the bounty. This is an array in the format `[ "reward", "<number>", "<currency>" ]`,
+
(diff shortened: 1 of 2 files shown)


117.md
```diff
@@ -0,0 +1,103 @@
...
+
+- A `timeline` tag to provide a date and time when the applicant expects to complete the bounty by.
+
+## Bounty Lifecycle
+
+The following is a description of the lifecycle of a bounty:
+
+1: **Open:** The bounty is created and published by the creator and accepting applications.
+
+2: **Assigned:** The bounty is has been assigned.
```
BitcoinBrief, line 65: typo:  remove "is"

staab:
I would add a `canceled` status, maybe with a timestamp. I also think if you're going to use a replaceable event you might as well explicitly reference the application that was granted.

117.md
```diff
@@ -0,0 +1,121 @@
...
+
+- An `application` tag with the stringified **Bounty Application Event**.
+
+## Bounty Application Event
+
+The `.content` of these events should be a simple plain text string explaining why the creator should choose the applicant.
+
+The following tags MUST be present:
+
+- A `d` tag to reference the **Bounty Event**.
```
nostrband, line 61: Should this be 'a' tag to reference the 30050 Bounty Event?

ChristianChiarulli: yea it should be, I'll update this

117.md
```diff
@@ -0,0 +1,121 @@
...
+
+We can work around this by making sure that the content and reward of the bounty event is the same as the content and reward found in the stringified **Bounty Event** in the **Bounty Event Application**. No **Bounty Event** without a matching reward across these three fields (including the zap receipt) will be considered complete. (Unless both parties agree to settle outside of the network)
+
+If the creator changes the content or reward in an effort to scam the applicant, the applicant has a record of the original content and reward in the **Bounty Event Application**. And the bounty will not count as complete which will at the very least not improve the creators reputation.
+
+If the applicant puts a fraudulent **Bounty Event** in the description of the **Bounty Event Application** the creator can choose not to pay.
+
+Without an agreed upon 3rd party there will always be a way to scam the system. This is the best way I can think of to discourage scammers.
+If the bounty was not replaceable, the creator could just delete the event anyway and skip paying after the applicant has submitted their work.
```
AsaiToshiya, line 102: That should also be possible to delete a replaceable.

ChristianChiarulli: I guess what I mean to say here is that even if the bounty wasn't replaceable the creator could just delete their bounty.

AsaiToshiya: Ah now I understand, thanks.

nostrband:
I've been thinking about bounties again bcs we plan to show github issues under repos on nostrapp.link and then allow people to pledge sats with one click to someone who would solve the issue. The difference there from the flow in this NIP:
- the 'problem' statement is not in the 'bounty event', it's external and created by users unaware of the bounty system, the 'bounty' is just a promise to pay sats with a reference to the problem
- thus naturally many pledgers might reference the same problem
- and then if someone wants to 'apply' the only reason to do that is to include the pledges inside 'apply' event to make sure those aren't deleted - you don't need permission from (multiple) pledgers bcs the problem wasn't created by them, and the solution will be accepted by maintainers, not pledgers
- the 'apply' event becomes an optional 'I'm working on this' announcement - might also serve as a signal for others to not compete with me
- the resolution of the problem happens out of band (same with this NIP), but in my case pledgers need to be notified about the need to pay, since pledgers aren't the ones accepting the solution - I'm thinking about asking maintainers to publish a 'problem solved' event referencing the original issue, referencing all pledges and their authors and containing a zap-split set up targeting the people who solved the problem
- the pledgers would have to zap the 'problem solved' event which would become the proof that they've sent the sats they promised to send

I'd be happy if I could apply this NIP to my use case of 'pledge sats to github issues', but it seem too incompatible. 

Is there a chance we could somehow get this NIP closer to being applicable? Like making a bounty event just a reference to an externally stated problem? Making it non-replaceable - remove the bounty states (no one 'assigns' the problem), add 'solved' event (replaces the 'complete' status)? 

WDYT?

arbadacarbaYK:
Cross-link: [gittr.space](https://gittr.space) implements a **different** bounty model on **NIP-34 issues** (kind `9806`, LNURL-withdraw, release on PR merge) — see implementation profile: https://github.com/arbadacarbaYK/gittr-helper-tools/blob/main/snippets/nip34-issue-bounties/README.md

Opened a separate discussion for a NIP-34 companion approach: https://github.com/nostr-protocol/nips/issues/2414 — not trying to revive NIP-43 as-is, but interested in converging on interoperable git-issue bounties where possible.

Issue #537, opened 2023-05-17
Ephemeral NIP-47 is too ephemeral


TonyGiorgio:
According to [NIP-16](https://github.com/nostr-protocol/nips/blob/master/16.md), ephemeral events MUST NOT be stored. These seem to only be useful for synchronous communication, which conflicts with the asynchronous nature of nostr. 

One of the use cases makes sense: 

> Typing indicators: A chat application may use ephemeral events as a typing indicator.

However, this example given in `NIP-16` does not: 

> Messaging: Two pubkeys can message over nostr using ephemeral events.

It is extremely unreliable to use this as a messaging protocol between two users if the events must not be stored and are discarded if one of the users even momentarily has a disconnection from a relay. However, there's another use case that is using ephemeral events when it's not strictly needed.

[NIP-47](https://github.com/nostr-protocol/nips/blob/master/47.md) seems to have been created with the assumption that nostr users should be using custodians or are running an always online lightning node at home or in the cloud. These are two opposite extremes and it does not well support the middle ground which is non-custodial mobile wallets.

I would like to start some conversation around how to best support this use case with some adaptations to either `NIP-16`, `NIP-47`, or a new NIP all together. Essentially the way a non-custodial mobile wallet could take advantage of NWC is by letting a relay queue up the `pay_invoice` requests until they come online again. This should allow for the same one-click experience that is finality-delayed until the wallet application is opened again. Once opened, all the `pay_invoice` requests could be retrieved from the relay (something that does not seem possible with ephemeral events today) and paid all at once.

---

Here's some options I've thought of:

#### Semi-ephemeral events

1. Could be an adaptation to `NIP-16` that specifies that ephemeral events behave more like semi-ephemeral events. IE, MUST NOT be stored, but SHOULD BE in memory for a temporary amount of time (10 minutes?) in order to allow some time for the recipient to come online to get the event.
2. Could be a new event range. IE, `30000 <= n < 40000`, but that seems wasteful.

#### queue_pay_invoice

A new NIP for a new NWC command type. This event could look almost identical to `pay_invoice` except it is in the "Regular Event" range so it is stored by the relays and allowed to be fetched when the wallet comes back online. Wallet services (ie, mobile wallets) could advertise this one instead of `pay_invoice` in their `NIP-47` `13194` event type. 

These events would require more relay resources than semi-ephemeral events, so I suggestion finding some balance and flexibility with ephemeral events instead.

---

[...]

PR #866, opened 2023-11-05
Recurring Subscriptions

pablof7z:
NIP to implement Patreon-style support.

Adds a way to:

* Create support tiers (`kind:7002`)
* Subscribe to support tiers (`kind:7001`)

Viewable: https://github.com/nostr-protocol/nips/blob/nip88/88.md


vitorpamplona:
Very cool! 

I'd like to see a way to tell the NWC service (or a DVM) to keep paying the 7001 event until that event gets deleted. 

It must be a service to avoid missing payments (in case the user doesn't open a client on that day) and also avoid double payments in case the zap event disappears (relay is offline or busy for a day or so).

88.md
```diff
@@ -0,0 +1,79 @@
...
+`.content` SHOULD be a description of what users who subscribe can expect.
+Tag `title` is an optional title for the tier.
+Tag `amount` MUST specify the payment required for this tier and its cadence.
+    * The first argument should be the stringified amount and the second argument the currency
+    * The third argument SHOULD be one of `daily`, `monthly`, `yearly`
+One or more `amount` tags MUST exist.
+
+#### Examples
+* `[ "amount", "100", "usd", "daily" ]`, $1.00 a day.
+* `[ "amount", "1000000", "btc", "daily" ]`, 1000000 millisats a day.
```
benthecarman, line 36: This is kinda confusing, why not make the unit msats

88.md
```diff
@@ -0,0 +1,79 @@
...
+        [ "e", "<kind-7001-event-id>" ]
+    ]
+```
+
+Clients supporting this NIP can check for zaps e-tagging the `kind:7001` event to find the pubkeys that have a valid, paid subscriptions at each different period.
+
+The same `kind:7001` is re-zapped on a regular basis per the cadence specified in the event.
+
+## Stopping a subscription
+A user who wants to stop a subscription by publishing a `kind:5` deletion request of the `kind:7001` event.
```
benthecarman, line 79: This could be a problem, most relays don't support this AFAIK. Also a subscribed user's client might just think the relay is acting up, not that the subscription is over. 

I think it'd be better if there was a revoke event, it could have an e tag that points to the new tier that replaces it.

88.md
```diff
@@ -0,0 +1,79 @@
...
+}
+```
+
+This event is generated by a pubkey who wants to allow users to support with different tiers.
+
+`.content` SHOULD be a description of what users who subscribe can expect.
+Tag `title` is an optional title for the tier.
+Tag `amount` MUST specify the payment required for this tier and its cadence.
+    * The first argument should be the stringified amount and the second argument the currency
+    * The third argument SHOULD be one of `daily`, `monthly`, `yearly`
```
benthecarman, line 31: Should use the same arguments as #851

88.md
```diff
@@ -0,0 +1,79 @@
...
+One or more `amount` tags MUST exist.
```
TonyGiorgio, line 32: How should changing the amount happen? A new `7002` ? Replace it? Reference the previous `7002` in the new `7002`? 

The question of reasoning here is to consider what might happen if denominated in sats and need to change that with huge price changes, or standard repricing of services. 

I wonder how to: 
1. historically track subscriptions previously no matter what the prices might have changed to
2. indiciate to the user that the previous pricing model has changed and they SHOULD NOT zap with the previous price without creating a new `7001` at least.

88.md
```diff
@@ -0,0 +1,79 @@
...
+    "content": "<optional-message>",
+    "tags": [
+        [ "p", "<recipient-pubkey>" ],
+        [ "e", "<supporting-tier-id>" ],
+        [ "amount", "<amount-in-base-unit>", "currency", "<cadence>" ]
+    ]
+}
+```
+
+When a user wants to subscribe to subscribe to support a user they create a `kind:7001` event.
```
TonyGiorgio, line 52: nit: typo

```suggestion
When a user wants to subscribe to support a user they create a `kind:7001` event.
```

88.md
TonyGiorgio: Echo some of the concerns, plus another one: historical references. The user _was_ a subscriber for X months, for example. That might be useful for both accounting and reputation, in addition to possible content that could still be accessed from those months if someone chooses to do a newsletter/patreon client but let users keep what they paid for during the subscription time. 

Another possible approach: 

Could we just have the client stop zapping? I can see the usefulness in having a nostr native event to trigger the start of the subscription, but since verification depends upon analyzing the zaps to check if sub'd, couldn't we just have the user stop zapping? If they gave out an NWC string, then some user action might be required anyways to make sure the service stops withdrawing anyways despite a deleted 7001.

88.md
```diff
@@ -0,0 +1,79 @@
...
+
+* `.content` is an optional message the supporter can write.
+* The `p` tag MUST tag the pubkey being supported.
+* The `e` tag is optional, and should point to a `kind:7001` support tier event. There MUST be exactly 0 or 1 `e` tags.
+* The `amount` tag specifies what the supporters is committing to pay to the supported pubkey and the candence.
```
TonyGiorgio, line 57: Should this specify that `amount` MUST BE at or above the `amount` tag in the `7002`?

88.md
```diff
@@ -0,0 +1,79 @@
...
+{
+    "kind": 9734,
+    "content": "",
+    "tags": [
+        [ "p", "<recipient-pubkey>" ],
+        [ "e", "<kind-7001-event-id>" ]
+    ]
+```
+
+Clients supporting this NIP can check for zaps e-tagging the `kind:7001` event to find the pubkeys that have a valid, paid subscriptions at each different period.
```
TonyGiorgio, line 74: It might be worth pulling this sentence out of `Paying` and start a new subsection called `Verifying`, where it can address things like checking the `cadence` the user was/is subscribed to and that `amount` matches whatever it was for that duration.

AnthonyRonning (commented):
Approach sounds great, thank you for drawing it up! Just a few edge cases and thoughts around changing prices and keeping historical references.

TonyGiorgio:
> I'd like to see a way to tell the NWC service (or a DVM) to keep paying the 7001 event until that event gets deleted.

I would love to see some NWC interactions as well, though I'm concerned about it triggering based on seeing _something_ end it. We can't guarantee the wallet is going to see deleted events or the `stop` event if we take @benthecarman's suggestion.

benthecarman:
While I agree that nwc should be used for giving invoices to the user, I don't see where it really fits into this nip outside of a recommendation section.

vitorpamplona:
Creating the 7001 event and sending to the NWC relay could mean that the NWC service is authorized by the user to get an invoice and pay that invoice in the appropriate cadence. 

Or maybe it's just an extra event that has the 7001 inside of it that tells the NWC service to pay periodically. 

However, I'd like to see a way to tag which NWC service is authorized to pay each subscription in case the user has multiple NWC services setup.

TonyGiorgio:
> Creating the 7001 event and sending to the NWC relay could mean that the NWC service is authorized by the user to get an invoice and pay that invoice in the appropriate cadence. 
> 
> Or maybe it's just an extra event that has the 7001 inside of it that tells the NWC service to pay periodically. 
> 
> However, I'd like to see a way to tag which NWC service is authorized to pay each subscription in case the user has multiple NWC services setup. 

There's not really a concept of NWC services on a nostr level yet, I feel that it would be out of scope for this NIP. And any person can forward an event to "an NWC relay" (which there also isn't an exclusive concept of) so I don't think that would work either.

vitorpamplona:
> There's not really a concept of NWC services on a nostr level yet,

The NWC wallet service replies to payment requests with signed events (kind:23195). The pubkey used for those responses could be used to identify which NWC service is supposed to be paying the 9001 subscription. We could add that key directly as a tag in the 9001.

Clients already use that key to identify the NWC wallet service. So that could be very easy to integrate.

88.md
fiatjaf: Indeed, otherwise this will be a parsing hell.

kiwiidb:
I envisioned something along the following lines, using the new NIP49 proposal:

- Next to their LUD16 a user has a Zapplepay-like service in their profile identified by pubkey/relay
- Subscriber produces a 7001 event signaling intent to subscribe
- The NWC service of the subscriber (Alby/Zapplepay) picks up the 7001 (and optionally asks the user for confirmation in some way through DM/email/notificiation) and then sets up the connetion with the service described above. The connection has a budget/period and also the 7001 event is referenced when setting up the connection.
- The Zapplepay-like service of the receiver is responsible for generating the invoices on behalf of the user in due time and using the NWC connection to pay them. It will also produce the zap event but clients should recognise that this is actually the 7001 author zapping. The subscriber should probably also get notified if an automatic payment like this happens.
- Cancelling probably be just deleting the NWC connection on the sender side. Clients need to validate the payments anyway, the 7001 event in itself doesn't mean much.

88.md
pablof7z: oh yeah, you're right

88.md
pablof7z: yeah, that's the thing, we need a way for different clients to know they shouldn't continue zapping.

Yeah, some kind of flagging would probably be better than an kind:5 which has broader implications.

TonyGiorgio:
I can see something like that working. Though ideally there's no external NWC third party necessary. I'm a bit concerned of the centralizing factor here for third parties storing user NWC strings, especially if it becomes a "one string to rule them all" scenario instead of a specific string per subscription action. It'll lead to needing bigger budgets set on the wallet side and will lead to more trust that the service provider doesn't steal as much of the user's funds as possible.

I'd like to see services with subscription capabilities bring their own NWC capabilities in house, or keep it on the nostr client side. If an NWC service is needed just to set a cron job, then I'd like that service to be as trustless and servicable by anyone as possible. I don't like that zapplepay has to be a thing, it's a hack and I don't want to see it being part of important integration points for nostr.

vitorpamplona:
Maybe we can invert the architecture here. What if the pubkey receiving subscriptions must own a server that periodically parses 9001 events and sends ln invoices for payment in the cadence it offers? This would simulate the traditional way of paying for things (you have to create invoices and send them out when they are due). 

That service can send the invoice over regular DMs OR we can create a new event kind (e.g. 7002) to carry the ln invoice in such a way that clients can create specific UIs for their payment, including automatically paying them with a Zap when they arrive. 

Since the service already created the LN Invoice, it's not likely that it will accept double payment in the same invoice.

TonyGiorgio:
> What if the pubkey receiving subscriptions must own a server that periodically parses 7001 events and sends ln invoices for payment in the cadence it offers

Yeah I think that would be similar to my "services bring NWC in house" comment. Though I think it would work with NWC as is, I don't think it needs a new event kind. 

Regarding the "must own a server part", couldn't individual users take advantage of this as well? IE, Alice set up a subscription tier for her own npub, Bob was to "subscribe" to Alice. Bob gives Alice a specific NWC meant for her, with appropriate budgets set up, and when Alice comes online in her nip88 aware client, she sends Bob an invoice via NWC. It would work better with a monthly cadence, but it still allows individual control and no server needed.

vitorpamplona:
> Bob gives Alice a specific NWC meant for her

This sounds weird. It's like giving my vendor user/password access to my bank account so my vendor can pay itself. Sure, we can set things up to make everything work within the budget, but it's still weird granting people the right to spend in my wallet. 

> Alice comes online in her nip88 aware client

If the responsibility of keeping the cadence is with the subscription receiver, then sure. If Alice is late because she didn't connect her NIP88-aware client, it's her fault. But being late will make things messy (people like exact predictability, e.g. invoices come on the 1st day of the month). I still think a better user experience should have a server for the cron job. 

> no server needed

Isn't the NWC a server? Otherwise, the 23194 payment request event, which is ephemeral, might never be seen by the NWC. Looks like this method requires both clients (Alice's sender and Bob's payer) to be online at the same time to make sure the ephemeral event is processed.

TonyGiorgio:
> weird granting people the right to spend in my wallet

At the end of the day, it's a payment request, not an explicit spend. Almost every payment app that exists allows payment requests to occur and internal controls around it. 

> Isn't the NWC a server

No, it's not. It works just fine with Mutiny mobile wallets. The requirement for server is not necessary, the user pulls these down themselves. 

> 23194 payment request event, which is ephemeral

That's why I created this awhile back: https://github.com/nostr-protocol/nips/issues/537

The Mutiny nostr relay stores these in order for users to be able to retrieve them when they are offline. Making NWC ephemeral & basically synchronous is a design flaw IMO. 

I'll bring up that there's a lot of things in flight with NWC still and as long as _some_ model can work with nip88, I don't think we absolutely have to know what that should be yet. That's just my opinion, I don't want to hijack nip88 with comments around NWC but I'll leave with those thoughts. I do think there's a lot of ways this can play nicely with NWC (or potentially NWC services if we expand upon those more) but I don't think it's an absolute requirement.

vitorpamplona:
Agree, the connection to NWC shouldn't stop this PR. 

Let's not forget a lot of people don't like using NWC, so we may need to make an option that is fully manual.

I think the cron job to send invoices should be a DVM. The receiver pays the DVM to send invoices to everyone.

88.md
benthecarman: actually you should keep yours, these are better and what alby uses for their `get_balance` nwc command. I will change #851, but can you add `weekly`?

88.md
```diff
@@ -0,0 +1,79 @@
...
+## Tier Event
+A pubkey can create "tiers". These tiers might provide certain benefits to the supporters who subscribe to these.
+
+```json
+{
+    "kind": 7002,
+    "content": "<description of the tier>",
+    "tags": [
+        [ "title", "..." ],
+        [ "amount", "<amount-in-base-unit>", "currency", "<cadence>" ]
```
erskingardner, line 20: In NIP-99 we use the same tag format but we call it `price`, I feel like we should standardize. I have no preference on which word we use but given NIP-99 is already out there and in use, it might be easier to use that. 

https://github.com/nostr-protocol/nips/blob/master/99.md#price-examples

88.md
pablof7z: I think it might just make sense to move this to NIP-33 events and the 7001 should `e` tag the version it's subscribed tro.

pablof7z:
Ok, made quite a few updates; moved the tier event to a 37001 to use NIP-33, subscribing events can keep a copy of the subscribed event.

believethehype:
NIP88 in its current form has been implemented in the NostrDVM framework and noogle client.

TheIcarusWings:
I've been building a creator platform on Nostr ([Nostreon](https://github.com/Nostreon)) and implemented this spec in production. Sharing some observations and suggestions from running it with real subscriptions.

## What works well

The core event structure (kind 37001 tiers, kind 7001 subscribe, kind 7002 unsubscribe) is solid and straightforward to implement. The `perk` tags are great for letting clients render tier benefits consistently. The separation of subscribe/unsubscribe into distinct kinds rather than relying on `kind:5` deletions is the right call.

## Feedback from my implementation

### 1. Payment verification beyond zaps

The spec currently relies on zap receipts to verify active subscriptions. In practice, I found this insufficient for a few reasons:

- Zaps require the subscriber's client to be online and their wallet connected at the right time
- Failed zaps are silent — there's no way for the creator or subscriber to know a payment didn't go through
- Verifying subscriptions by scanning for zap receipts across relays is slow and unreliable

I ended up using **BTCPay Server invoices with webhook confirmations** — the subscriber pays an invoice, BTCPay confirms it, and I store the subscription state with an explicit `expires_at` timestamp. This is closer to how traditional subscription billing works and is much more reliable.

**Suggestion:** The spec could be more explicit that the payment mechanism is flexible. The `payment-verifier-pubkey` concept is good, but the spec could define a `kind:7003` receipt more concretely — e.g., the payment verifier publishes a receipt with an `expires_at` tag so any client can check subscription validity without re-scanning zap history.

```json
{
  "kind": 7003,
  "tags": [
    ["p", "<subscriber-pubkey>"],
    ["e", "<kind:7001-subscribe-event-id>"],
    ["valid_until", "<unix-timestamp>"],
    ["amount", "<amount-paid>", "<currency>"]
  ]
}
```

### 2. Subscription lifecycle: expiry, grace periods, reminders

The spec doesn't address what happens between payments. In production I needed:

- **Explicit expiry** — each subscription has an `expires_at` derived from the cadence. Access is checked against this, not against "has a recent zap receipt"
- **Renewal reminders** — I send Nostr DMs at 7 days and 1 day before expiry. This significantly improves renewal rates
- **Grace period** — a short window after expiry where content is still accessible but the subscriber is prompted to renew

**Suggestion:** Consider adding a `valid_until` or `expires_at` tag to the `kind:7003` payment receipt. This makes subscription validity deterministic rather than dependent on client-side zap scanning logic.

### 3. Content gating: how do subscribers actually access content?

The spec defines the subscription events but doesn't specify how content access is enforced. The `r` tag pointing to a relay is a good hint, but the full flow needs spelling out.

I implemented a **dual-relay architecture**: a public relay for free content and discovery, and a gated relay that requires NIP-42 authentication + subscription verification. I've proposed this as a separate NIP ([nostr-protocol/nips#2311](https://github.com/nostr-protocol/nips/issues/2311)) and have an open-source reference implementation at [Nostreon/relay-auth](https://github.com/Nostreon/relay-auth).

This spec could reference or recommend a relay-based content gating pattern to close the loop between "user is subscribed" and "user can access content."

### 4. Annual billing / multiple cadence options

The `amount` tag supports multiple cadences, which is great. In practice, offering annual billing with a discount (e.g., 2 months free) is a strong conversion driver. The spec handles this naturally with multiple `amount` tags on a tier:

```json
["amount", "500", "usd", "monthly"],
["amount", "5000", "usd", "yearly"]
```

Just noting that this pattern works well and should probably be called out as an example.

### 5. Price change handling

Echoing @AnthonyRonning's earlier comment — price changes need clearer guidance. My approach: existing subscribers keep their current price until their subscription expires, new subscribers get the new price. The `event` tag (snapshotting the tier at subscribe time) helps here, but the spec should explicitly recommend this behavior.

## Summary

The spec is in good shape. The main gaps I hit in production were around payment verification reliability (beyond zaps) and subscription lifecycle management (expiry, reminders, grace periods). Adding a more concrete `kind:7003` receipt with `valid_until` would solve most of these.

I've open-sourced my implementation as a reference at [Nostreon/nip88-subscriptions](https://github.com/Nostreon/nip88-subscriptions) — covers event builders, subscription lifecycle, invoice-based payment verification, and renewal reminders.

Happy to share more details or help move this forward.

fiatjaf:
@Joaofcm may I suggest that you look into https://github.com/nostr-protocol/nips/pull/2156? Your opinion would be appreciated.

PR #867, opened 2023-11-08
Problem: nostrocket problem tracker event kind not included in list

gazhayes:
Mostly just letting people know that this event kind is now in use by Nostrocket.


erskingardner:
@gazhayes I think the kind number needs to part of a merged NIP before it gets included here in the readme

gazhayes:
> @gazhayes I think the kind number needs to part of a merged NIP before it gets included here in the readme

I'm pushing the Overton window towards making decentralisation great again by not having a central place NIPs. 

This repo should be a convenient index for NIPs rather than an authority over them.

vitorpamplona:
I don't dislike the idea of citing another NIP repo, but it should be separate from the NIPs managed by this repo.

fiatjaf:
I kinda like this.

gazhayes:
> I don't dislike the idea of citing another NIP repo, but it should be separate from the NIPs managed by this repo.

A separate index? That would be less useful to new people coming in.

I regularly have people asking me "can I do this", or "can I do that", with regard to making a new `Kind` or `Tag` or something. This never used to happen, people used to just experiment. This is a trend in the wrong direction. I suspect the largest contributing factor is the fact that there is a central authority over NIPs, which makes people want to ask for permission.

A second-class index of "unauthorized" NIPs is less useful in addressing this specific problem, and also less useful in general.

If you're worried about things becoming a mess, an indication of NIP lifecycle status could help though, I've defined this but haven't published yet, here's a summary:
"RAW" = not implemented, request for comments
"DRAFT" = implemented but could change
"STABLE" = implemented, and WE DO NOT BREAK USERSPACE
"DEPRECATED" = use this other NIP instead
"RETIRED" = no one is using this anymore

weex:
The use of the prefix NIP for a different process is confusing. Is it possible to use a different acronym/prefix for documents coming out of the [nostrocket/NIPS](https://github.com/nostrocket/NIPS) process?

gazhayes:
> The use of the prefix NIP for a different process is confusing. Is it possible to use a different acronym/prefix for documents coming out of the [nostrocket/NIPS](https://github.com/nostrocket/NIPS) process?

What specifically is confusing to you?

fiatjaf:
I think we could merge this, but only because Nostrocket is a separate protocol managed elsewhere and the NIP here mentioned doesn't have any intentions of being adopted by other Nostr apps, so it makes sense for us to care very much about it and the reference here would be just for parking the event kind and pointing people to the actual spec.

weex:
> > The use of the prefix NIP for a different process is confusing. Is it possible to use a different acronym/prefix for documents coming out of the [nostrocket/NIPS](https://github.com/nostrocket/NIPS) process?
> 
> I don't understand the concern. What specifically is confusing to you?

The nostrocket/NIPS repo defines states which aren't seen in nostr-protocol/nips. Wait, what does NIPS stand for? That might be the source of confusion.

gazhayes:
> Wait, what does NIPS stand for? That might be the source of confusion.

This is what I suspected. NIPs are quite different to what most people are used to (e.g. BIPs). They are really just documentation for something already seen in the wild. That was the original intention anyway.

pablof7z:
I like seeing this a lot.

pablof7z (approved):

staab (approved):

Issue #720, opened 2023-08-14
Signer Interface Spec


vitorpamplona:
It would be nice to standardize the signing spec among NIP-07, NIP-46 and [NIP-100](https://github.com/nostr-protocol/nips/pull/868). @greenart7c3 is developing a native Android Signer called [Amber](https://github.com/greenart7c3/Amber) that should offer similar methods to 07 and 46 but via Android Intents (the way Android apps call one another, offline). There may be a similar app for iOS in the future with yet another protocol for inter-app communication. Nevertheless, all of these different protocols should implement the same functions, with the same inputs and outputs, namely: 

- sign_event
- get_public_key
- nip04_encrypt
- nip04_decrypt
- nip44_encrypt
- nip44_decrypt
- delegate

The format of the input/output and the calling method might change, but in the end they are all the same. 

With this definition, we can potentially merge NIP-07 and NIP-46, while opening space for new PRs on the Android Intent protocol and the iOS Intents and avoid 4 NIPs that would be doing the same work in just different formats.

alexgleason:
Yes. But it should use camelCase names to match nip07 since that one is already the most widely implemented. The world won't suffer if nip46 implementations break.

fiatjaf:
Let's remove `nip04_*` and `delegate` from the spec.

PR #868, opened 2023-11-08
Android Signer Application

greenart7c3:
Nostr event signer for Android. It allows users to keep their nsec segregated in a single, dedicated app without requiring any extensions or web servers.

### DIFF
--- a/100.md
+++ b/100.md
@@ -432,63 +432,49 @@ Android intents and browsers url has limitations, so if you are using the return
   - params:
 
     ```js
-    const intent = `intent:#Intent;scheme=nostrsigner;S.compressionType=none;S.returnType=signature;S.type=get_public_key;S.callbackUrl=https://example.com/?event=;end`;
-
-    window.href = intent;
+    window.href = `nostrsigner:?compressionType=none;returnType=signature;type=get_public_key;callbackUrl=https://example.com/?event=`;
     ```
 
 - **sign_event**
   - params:
 
     ```js
-    const intent = `intent:${eventJson}#Intent;scheme=nostrsigner;S.compressionType=none;S.returnType=signature;S.type=sign_event;S.callbackUrl=https://example.com/?event=;end`;
-
-    window.href = intent;
+    window.href = `nostrsigner:${eventJson}?compressionType=none;returnType=signature;type=sign_event;callbackUrl=https://example.com/?event=`;
     ``` 
 
 - **nip04_encrypt**
   - params:
 
     ```js
-    const intent = `intent:${plainText}#Intent;scheme=nostrsigner;S.pubKey=${hex_pub_key};S.compressionType=none;S.returnType=signature;S.type=nip04_encrypt;S.callbackUrl=https://example.com/?event=;end`;
-
-    window.href = intent;
+    window.href = `nostrsigner:${plainText}?pubKey=${hex_pub_key};compressionType=none;returnType=signature;type=nip04_encrypt;callbackUrl=https://example.com/?event=`;
     ``` 
 
 - **nip44_encrypt**
   - params:
 
     ```js
-    const intent = `intent:${plainText}#Intent;scheme=nostrsigner;S.pubKey=${hex_pub_key};S.compressionType=none;S.returnType=signature;S.type=nip44_encrypt;S.callbackUrl=https://example.com/?event=;end`;
-
-    window.href = intent;
+    window.href = `nostrsigner:${plainText}?pubKey=${hex_pub_key};compressionType=none;returnType=signature;type=nip44_encrypt;callbackUrl=https://example.com/?event=`;
     ```   
 
 - **nip04_decrypt**
   - params:
 
     ```js
-    const intent = `intent:${encryptedText}#Intent;scheme=nostrsigner;S.pubKey=${hex_pub_key};S.compressionType=none;S.returnType=signature;S.type=nip04_decrypt;S.callbackUrl=https://example.com/?event=;end`;
-
-    window.href = intent;
+    window.href = `nostrsigner:${encryptedText}?pubKey=${hex_pub_key};compressionType=none;returnType=signature;type=nip04_decrypt;callbackUrl=https://example.com/?event=`;
     ```     
 
 - **nip44_decrypt**
   - params:
 
     ```js
-    const intent = `intent:${encryptedText}#Intent;scheme=nostrsigner;S.pubKey=${hex_pub_key};S.compressionType=none;S.returnType=signature;S.type=nip44_decrypt;S.callbackUrl=https://example.com/?event=;end`;
-
-    window.href = intent;
+    window.href = `nostrsigner:${encryptedText}?pubKey=${hex_pub_key};compressionType=none;returnType=signature;type=nip44_decrypt;callbackUrl=https://example.com/?event=`;
     ```     
 
 - **decrypt_zap_event**
   - params:
 
     ```js
-    const intent = `intent:${eventJson}#Intent;scheme=nostrsigner;S.compressionType=none;S.returnType=signature;S.type=decrypt_zap_event;S.callbackUrl=https://example.com/?event=;end`;
-
-    window.href = intent;
+    window.href = `nostrsigner:${eventJson}?compressionType=none;returnType=signature;type=decrypt_zap_event;callbackUrl=https://example.com/?event=`;
     ```          
 
 ## Example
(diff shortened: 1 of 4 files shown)


greenart7c3:
I think it would also be better to have a Signer interface spec instead of separated nips like @vitorpamplona suggests in #720

fiatjaf:
This looks good.

vitorpamplona:
Looks good. 

Is there a way to identify all signers installed in the phone? 

For instance: Amethyst's login screen would have a "Login with another app" button that creates a URI intent that triggers Android's app chooser. The chosen Signer app receives the request to connect and replies with `package name` and any other details to assemble follow up calls.

greenart7c3:
> Looks good.
> 
> Is there a way to identify all signers installed in the phone?
> 
> For instance: Amethyst's login screen would have a "Login with another app" button that creates a URI intent that triggers Android's app chooser. The chosen Signer app receives the request to connect and replies with `package name` and any other details to assemble follow up calls.

A quick test I did you can use this code but you need the QUERY_ALL_PACKAGES permission:

```kotlin
val intent = Intent(Intent.ACTION_VIEW, Uri.parse("nostrsigner:"))
for (info in this.packageManager.queryIntentActivities(intent, 0)) {
    val packageName = info.activityInfo.packageName
    Log.d("Packages", "package = $packageName")
}
```

I havent found a way to get the package name from the android app chooser

vitorpamplona:
I am starting to think we should deprecate the Private Zap implementation used by Damus and Amethyst (event kind 9733)

The main issue is that it requires a special decrypt_zap function and a special handling of the sign function here, in the browser extension, and in NIP-46. And it requires it because the private key used in private events is a `sha256(privkey + (event_id for posts or pubkey for users) + createdAt.toString())`

Since apps with external signers (extension + NIP-46 + this one) cannot access the user's private key, they cannot generate the new key by themselves. The signer application must then have a special method to receive a created_at + event_id or a pubkey to assemble the key and sign it :(

greenart7c3:
I am also getting some feedback from the #nostrdesign and I think I'll add some permissions to the get public key so the users can approve or reject them when logging. So we can avoid showing a lot of pop ups.

https://cln.sh/y4MWbVYD

And in the content provider I need to add something to tell the app the user rejected the event kind/type forever

vitorpamplona:
@greenart7c3 can you check if this up-to-date with your implementation?

greenart7c3:
It is. Should I remove the nip 04 methods from here?

100.md
```diff
@@ -0,0 +1,528 @@
...
+      data = Uri.parse("nostrsigner:")
+    }
+  val infos = context.packageManager.queryIntentActivities(intent, 0)
+  return infos.size > 0
+}
+```
+
+## Using Intents
+
+To get the result back from the Signer Appication you should use registerForActivityResult or rememberLauncherForActivityResult in Kotlin. If you are using another framework check the documentation of your framework or a third party library to get the result.
```
dluvian, line 41:
```suggestion
To get the result back from the Signer Application you should use `registerForActivityResult` or `rememberLauncherForActivityResult` in Kotlin. If you are using another framework check the documentation of your framework or a third party library to get the result.
```

100.md
```diff
@@ -0,0 +1,528 @@
+# NIP-100
+
+## Android Signer Application
+
+`draft` `optional`
+
+This NIP describes a method for 2-way communication between a android signer and any Nostr client on Android. The Android signer is an Android Application and the Client can be a Web Client or an Android Application.
```
dluvian, line 7:
```suggestion
This NIP describes a method for 2-way communication between an Android signer and any Nostr client on Android. The Android signer is an Android Application and the client can be a web client or an Android application.
```

100.md
```diff
@@ -0,0 +1,528 @@
...
+
+# Usage for Android applications
+
+The Android signer uses Intents and Content Resolvers to communicate between applications.
+
+To be able to use The Android signer in your application you should this to your AndroidManifest.xml:
```
dluvian, line 13:
```suggestion
To be able to use the Android signer in your application you should add this to your AndroidManifest.xml:
```

vitorpamplona:
We should merge this. @fiatjaf

100.md
```diff
@@ -0,0 +1,528 @@
...
+      data = Uri.parse("nostrsigner:")
+    }
+  val infos = context.packageManager.queryIntentActivities(intent, 0)
+  return infos.size > 0
+}
+```
+
+## Using Intents
+
+To get the result back from the Signer Application you should use `registerForActivityResult` or `rememberLauncherForActivityResult` in Kotlin. If you are using another framework check the documentation of your framework or a third party library to get the result.
```
dluvian, line 41: I personally would appreciate a simple example for using `rememberLauncherForActivityResult`. It would have made things easier for me :D


```kotlin
val launcher = rememberLauncherForActivityResult(contract = ActivityResultContracts) { activityResult ->
  val signature = activityResult.data?.getStringExtra("signature")
  // Do something with signature ...
}

launcher.launch(input = intent, options = ActivityOptionsCompat.makeBasic())
```

100.md
```diff
@@ -0,0 +1,528 @@
...
+      val eventJson = intent.data?.getStringExtra("signature")
+      // the id you sent
+      val id = intent.data?.getStringExtra("id")
+      ```         
+
+## Using Content Resolver
+
+To get the result back from Signer Application you should use contentResolver.query in Kotlin. If you are using another framework check the documentation of your framework or a third party library to get the result.
+
+If the user did not check the remember my choice option, the npub is not in Signer Application or the signer type is not recognized the contentResolver will return null
```
dluvian, line 236:
```suggestion
If the user did not check the "remember my choice" option, the npub is not in Signer Application or the signer type is not recognized the `contentResolver` will return null
```

100.md
```diff
@@ -0,0 +1,528 @@
...
+
+For the SIGN_EVENT type Signer Application returns two columns "signature" and "event". The column event is the signed event json
+
+For the other types Signer Application returns the column "signature"
+
+If the user chose to always reject the event signer application will return the column "rejected" and you should not open signer application
```
dluvian, line 242:
```suggestion
If the user chose to always reject the event, signer application will return the column "rejected" and you should not open signer application
```

100.md
```diff
@@ -0,0 +1,528 @@
...
+
+        if (result.moveToFirst()) {
+            val index = it.getColumnIndex("signature")
+            val eventJson = it.getString(index)
+        }
+      ```    
+
+# Usage for Web Applications
+
+Since web applications can't receive a result from the intent you should add a modal to paste the signature or the event json or create a callback url.
```
dluvian, line 419:
```suggestion
Since web applications can't receive a result from the intent, you should add a modal to paste the signature or the event json or create a callback url.
```

100.md
```diff
@@ -0,0 +1,528 @@
...
+
+If you send the callback url parameter Signer Application will send the result to the url.
```
dluvian, line 421:
```suggestion
If you send the callback url parameter, Signer Application will send the result to the url.
```

100.md
```diff
@@ -0,0 +1,528 @@
...
+
+If you don't send a callback url Signer Application will copy the result to the clipboard.
```
dluvian, line 423:
```suggestion
If you don't send a callback url, Signer Application will copy the result to the clipboard.
```

100.md
```diff
@@ -0,0 +1,528 @@
...
+
+You can configure the returnType to be **signature** or **event**.
```
dluvian, line 425:
```suggestion
You can configure the `returnType` to be **signature** or **event**.
```

100.md
```diff
@@ -0,0 +1,528 @@
...
+
+Android intents and browsers url has limitations, so if you are using the returnType of **event** consider using the parameter **compressionType=gzip** that will return "Signer1" + Base 64 gzip encoded event json
```
dluvian, line 427:
```suggestion
Android intents and browser urls have limitations, so if you are using the `returnType` of **event** consider using the parameter **compressionType=gzip** that will return "Signer1" + Base64 gzip encoded event json
```

dluvian (commented):
Some nits.

I don't implement `Using Content Resolver` and `Usage for Web Applications` sections, so I can't comment on them but the rest above looks good.

100.md
```diff
@@ -0,0 +1,552 @@
...
+                var param1 = params.get("event");
+                if (param1) alert(param1)
+            }
+            let json = {
+                kind: 1,
+                content: "test"
+            }
+            let encodedJson = encodeURIComponent(JSON.stringify(json))
+            var newAnchor = document.createElement("a");
+            newAnchor.href = `intent:${encodedJson}#Intent;scheme=nostrsigner;S.compressionType=none;S.returnType=signature;S.type=sign_event;S.callbackUrl=https://example.com/?event=;end`;
```
tyiu, line 545: This is notably an Android-centric URI. For example, iOS doesn't have intents. Does this mean that web applications need to start being platform aware? If so, I'm not sure if this is a good choice. Can we achieve the same result in a platform-agnostic way?

I've been thinking through how to build an iOS signer application, and I feel like we can shape this spec to be platform agnostic (while still allowing flexibility for platform-specific first-class features be used to make the UX better).

Something like
```js
newAnchor.href = `nostrsigner:${encodedJson}?compressionType=none&S.returnType=signature;S.type=sign_event&callbackUrl=https://example.com/?event=;end
```

Let me know if I've misunderstood anything.

greenart7c3: We can change to do the way you did so it is platform-agnostic.

vitorpamplona:
Can we get this merged, please? Multiple android clients are using Amber with this API.

PR #873, opened 2023-11-10
Add DVM kinds to list

AsaiToshiya:


fiatjaf:
I can't see the changes because of the reformatting.
Can we add just the DVM range?

AsaiToshiya:
@fiatjaf Sure, updated the commit.

fiatjaf:
Oh, I see it now. Thank you.
I was actually just asking you to say what had changed, no need to revert the table formatting.
So it was indeed just the ranges.

weex:
@AsaiToshiya To keep patches minimal, it's better not to change column widths in the source file. Column widths as a display property are better handled by the renderer. If this has been discussed before please disregard.

AsaiToshiya:
@weex Thanks for your advice. I know, but I don't know if that has been discussed before.

PR #874, opened 2023-11-10
Clarify NIP 32

staab:
This is a follow up from #872. It doesn't change anything, it only attempts to clarify the NIP text since there is a lot of confusion about it.

One common source of confusion is the existence of values in labels. I wouldn't mind seeing RDF triples being supported by nostr at some point, but they have a number of problems:

- Labels are completely different semantically from triples. Combining both in a single NIP is probably not a good idea.
- There is no way to search arbitrary key/value pairs, since nostr only indexes single-letter tags, and only their first argument.

RDF triples weren't the goal of this NIP, so it should be left to another PR (either to this NIP or a new one) to support them.

@s3x-jay @sroertgen @rodant

### DIFF
--- a/32.md
+++ b/32.md
@@ -6,7 +6,30 @@ Labeling
 
 `draft` `optional` `author:staab` `author:gruruya` `author:s3x-jay`
 
-A label is a `kind 1985` event that is used to label other entities. This supports a number of use cases, including distributed moderation, collection management, license assignment, and content classification.
+A label is a `kind 1985` event that is used to label other entities. This supports a number of use cases,
+including distributed moderation, collection management, license assignment, and content classification.
+
+This NIP introduces two new tags:
+
+- `L` denotes a label namespace
+- `l` denotes a label
+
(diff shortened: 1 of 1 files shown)


pablof7z:
One PR at a time and NIP-32 will become BitVM; this was always @staab 's masterplan.

rodant:
@staab Thanks for the quick reaction and for doing this new PR :+1: From my point of view now the scope of this NIP is well limited to just labels. @sroertgen and I have the RDF triple use case in mind too and maybe we'll elaborate on a new NIP for that, but after doing some prototyping. I think it would be great to be able to implement this use case on Nostr too.

fiatjaf (approved):

sroertgen:
I agree with @rodant.
The NIP is now well defined and much clearer. Thanks a lot for quick reaction and understanding 🙏

sroertgen (approved):

vitorpamplona:
Nice! The new text is way better.

PR #875, opened 2023-11-10
Add closed communities

staab:
This is an attempt at superseding https://github.com/nostr-protocol/nips/pull/706 and incorporating existing NIPs for community definitions and member lists. The design is worse in many ways, but will be more familiar to devs, and less heavy to implement. I've attempted to compose existing pieces as much as possible rather than introduce entirely new concepts. This proposal competes with #566. This PR depends on #876, if you'd like to comment on the key sharing part please do it there.


staab:
@Semisol yes, I know, we fundamentally agree on the viability of this use case. I've been thinking it over for three months at this point, and have gotten input from many other devs about what "good enough" means. In the implementation I also intend to support defense-in depth using relay AUTH but didn't want to specify it before I know what it looks like. Would #566's signature stripping be preferable?

nourspace:
I find it curiois that members need admin's approval to leave a closed community.

staab:
> I find it curiois that members need admin's approval to leave a closed community.

Originally I had them posting their exit request to the group so people could update member lists eagerly. But membership is essentially equivalent to access to the shared key. Once you have it, you can't get provably get rid of it, so in theory you're a member until the key is rotated. UX things like eagerly removing the group from your joined list are another matter, and can easily be done by sending a copy of the exit request to your own pubkey.

72.md
```diff
@@ -22,6 +24,7 @@ The goal of this NIP is to create moderator-approved public communities around a
     ["d", "<Community name>"],
     ["description", "<Community description>"],
     ["image", "<Community image url>", "<Width>x<Height>"],
+    ["access", "open|closed|hybrid"],
```
vitorpamplona, line 28: Shouldn't the current sharedkey's public key be listed as a reference to decrypt the group? Maybe an `e` tag to the most recent `kind:24` for this group? 

Clients then look at the public key/e-tag and check if they have a private key in a `kind:24` that matches this public key.

86.md
```diff
@@ -0,0 +1,51 @@
...
+## Key Sharing
+
+Keys are shared via a `kind 24` rumor sealed by the `admin` key, wrapped using the shared key, and addressed to each recipient individually. The rumor MUST have a single `shared_key` tag containing the shared private key.
+
+```json
+{
+  "kind": 24,
+  "content": "Just a regular key rotation",
+  "tags": [["shared_key", "<new shared key>"]],
+}
```
vitorpamplona, line 22: Shouldn't `kind: 24` tag which community is the shared secret for?

86.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Key Rotation
+
+To rotate the shared key, the `admin` must publish a new `kind 24` rumor as described above. A `grace_period` tag MAY be included on the rumor indicating how many seconds after the new event's `created_at` timestamp previous keys are valid. This invalidates previous keys and replaces them with a new shared key.
+
+An `expiration` tag (defined by NIP-40) MAY be included on the wrapper to support a weak form of forward secrecy (weak, since it relies on relays to delete the event). This can reduce the impact of a member's private key being leaked, which could otherwise expose old shared keys and messages addressed to those keys.
+
+## Using the Shared Key
+
+Anyone holding the shared key may privately post events of any kind as a rumor sealed by their own key, and signed by the shared key.
```
vitorpamplona, line 33: What's "signed by the shared key"? The GiftWrap?

87.md
```diff
@@ -0,0 +1,83 @@
...
+Because the admin key is used for managing key rotation, it's recommended that a dedicated key be used for community administration instead of the community admin's personal key.
+
+## Key Sharing
+
+Key sharing and rotation is accomplished as described in [85.md](NIP 85).
+
+## Member Lists
+
+Admins MAY publish member lists using a kind `30000` event with the community's address as the `d` tag, as described in NIP 51. These events MAY be wrapped using NIP 59 to support private member lists.
+
```
vitorpamplona, line 38: Are these Closed Communities for large groups? If so, consider an unbounded list (https://github.com/nostr-protocol/nips/pull/784) instead of a fixed list. Otherwise, this event can become very large and updating all the time.

72.md
staab: Just pushed an update to nip 87 that specifies that the key share event should include an `a` tag referencing the group. Anyone who has the key will know about it.

86.md
staab: Yep, just added that.

86.md
staab: Pushed a clarification

87.md
staab: I strongly dislike nip 61. My preference would be for large groups to publish "users added" and "users removed" events instead. This same approach would fix the race conditions associated with lists (although not likely a problem in this case). But in the interest of a simple first draft I've used regular lists since they should still support several thousand members.

87.md
```diff
@@ -0,0 +1,84 @@
...
+  "tags": [["a", "34550:<admin pubkey>:<group name>"]]
+}
+```
+
+## Messages
+
+Any community member may post an event to the community either publicly following NIP 72, or privately. Private events MUST be sealed by the user's own key, and their wrapper signed by
+the shared key. Anyone with the shared key may decrypt these messages.
+
+All messages MUST include an `a` tag pointing to the community definition event's address. Events of any kind MAY be published to the community either publicly or privately.
```
vitorpamplona, line 74: Do you mean the GiftWraps should have an `a` tag?

87.md
```diff
@@ -0,0 +1,84 @@
...
+  "kind": 26,
+  "content": "I'm outta here",
+  "tags": [["a", "34550:<admin pubkey>:<group name>"]]
+}
+```
+
+## Messages
+
+Any community member may post an event to the community either publicly following NIP 72, or privately. Private events MUST be sealed by the user's own key, and their wrapper signed by
+the shared key. Anyone with the shared key may decrypt these messages.
```
vitorpamplona, line 72: Ok, if I need to download all private posts of community `C` what I have to do is: 
1. Download all `kind:24`s with a `a` tag to `C`.
2. Get all public keys of the private keys shared in the `kind:24`. 
3. Create a filter for all GiftWraps that have been authored by all keys from 2. 
4. Use each key to decrypt the GiftWrap. 
5. Check if the sealed event has an `a` tag to `C`. 
6. Display all of 5s events as `C`. 

Is that right? 

That means that we need a way to discard past keys that might have been compromised and were used so fill the community with spam. 

Or should clients only use the last key for each `a`?

87.md
staab: To the rumor, added a clarification

87.md
staab: That's basically right, except step 1 would be kind 24s addressed to the current user (keys are shared with members individually).

When a key is rotated, it invalidates the previous one with a grace period. So there may be multiple keys valid simultaneously, but only for a short period of time. Clients should track that and stop accepting events signed by revoked keys (timestamp spoofing could be done, but that would bury stuff in the timeline, and clients could mitigate that too if needed).

Also, step 4 doesn't require trying all keys, since you know the key the event was encrypted with.

87.md
```diff
@@ -0,0 +1,120 @@
...
+{
+  "kind": 26,
+  "content": "I'm outta here",
+  "tags": [["a", "35834:<admin pubkey>:<group name>"]]
+}
+```
+
+## Messages
+
+Any group member MAY post an event to the group either publicly following NIP 72, or privately. Private events MUST be sealed by the user's own key, wrapped by the shared key, and addressed to the shared key. Anyone with the shared key may decrypt these messages.
```
arthurfranca, line 89:
```suggestion
Any group member MAY post an event to the group either publicly following NIP 72, or privately. Private events MUST be sealed by the user's own key, wrapped by the shared key, and addressed to nobody. Anyone with the shared key may decrypt these messages.
```
If the wrapper's pubkey is the shared one, members can already filter by author instead of by #p.

arthurfranca:
It would be great for privacy if every member of the group would always post with an alias pubkey (although the member list would include the real one). An alias they wouldn't change (atleast not very often).

Then an user can choose to reveal theirself (with a rumor kind for that) to other specific members manually or auto if to their follows (by looking at the member list of real pubkeys).

This would reduce the problem of some suspicious member handling the shared key to the public, cause although outsiders would be able to decrypt messages, they wouldn't know who is talking. Yet we keep a good UX by allowing friends to recognize each other.

mattn:
I love this since this can be great for privacy even though someone do broadcast the note.

staab:
@arthurfranca interesting idea, I'll play around with that.

fiatjaf:
Can we put the shared key mechanism in the same NIP for now? If in the future it becomes apparent that it will be used in other situations then it can be moved to its own NIP. For now it just creates unnecessary complexity and confusion.

staab:
Sure, just pushed an update

87.md
```diff
@@ -0,0 +1,151 @@
+NIP-87
+=======
+
+Closed Groups
+------------------
+
+Some online communities may want a measure of privacy. While nostr's architecture isn't conducive to high levels of privacy due to metadata leakage and the difficulty of implementing ratchets for key rotation, it can provide basic encrypted channels appropriate for use with large groups with dynamic member lists.
+
+This NIP relies on [./86.md](NIP 86) for key sharing.
```
abhay-raizada, line 9: can someone point me to nip-86? couldn't find it

AsaiToshiya: https://github.com/nostr-protocol/nips/pull/876

staab: This was a leftover, key sharing is now included in 87.

87.md
```diff
@@ -0,0 +1,149 @@
...
+
+## Admin Key
+
+The key used to publish the group's `kind 35834` definition event MUST be used for performing privileged actions, such as publishing key rotations. Non-administrative events MAY be published to the group by the admin.
+
+Because the admin key is used for managing key rotation, it's recommended that a dedicated key be used for group administration instead of the group creator's personal key.
+
+## Key Sharing
+
+Keys are shared via a `kind 24` rumor sealed by the `admin` key, wrapped using an ephemeral key, and addressed to each group member individually, as well as to the admin's pubkey.
```
abhay-raizada, line 39: how do we wrap keys with ephemeral keys? What is meant by an ephemeral key? I'm new to cryptography, anything I could read up on?  might be good to expand upon this on the nip?

abhay-raizada: but wouldn't the user need the corresponding pubkey to decrypt the `privkey` tag? is it shared out of band? also how does the user get notified about being added on the group? is the `kind:24` event shared out of band as well?

staab: It's all wrapped inside a 1059, which is p-tagged to the recipient. So clients have to decrypt those to find out what is inside. They can decrypt them, because the payload is encrypted for the recipient's private key.

vitorpamplona: I think @staab intended to have two kind 24 events, one with `admin` role with the editing key and one with the `member` role (terrible name) for the view-only key.

abhay-raizada: okay I think I finally got it, hadn't gone through `https://github.com/nostr-protocol/nips/blob/master/59.md` It'd be nice to mention nip-59 on this doc. Was not aware of it and was very confused :P

staab:
Putting this PR to rest. Relay-based groups are where I'm putting my effort for now, and I'm encouraged about Jeff's work on MLS.

PR #876, opened 2023-11-10
Add nip for key sharing

staab:
This PR adds a simple scheme for sharing keys, including simple rotation that supports weak forward secrecy. This is a prerequisite to #875.

### DIFF
--- a/86.md
+++ b/86.md
@@ -1,0 +1,51 @@
+NIP-86
+=======
+
+Shared Keys
+-----------
+
+`draft` `optional` `author:staab` `author:earonesty` `author:vitorpamplona` `author:water783`
+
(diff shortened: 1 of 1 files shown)


fiatjaf:
I think this should be inside NIP-87.

staab:
I didn't include it because I think it has other use cases, for example something like water783's [alias exchange](https://github.com/water783/nips/blob/nip101/101.md).

fiatjaf:
That's premature modularization.

Once NIP-87 exists the alias exchange NIP can just say "we share keys like in NIP-87" and then proceed to describe the added differences.

staab:
Yeah, that's fair. I do think it's clearer to separate the concerns anyway, but I have no real objection to merging the nips.

PR #673, opened 2023-07-18
Specifies a hide keyword tag into Mute Lists

vitorpamplona:
Users can add keywords to their Mute list and Clients can then filter posts that contain those keywords out of the view.

PR #880, opened 2023-11-15
Make NIP-51 useful again

fiatjaf:
https://github.com/nostr-protocol/nips/blob/save-nip-51/51.md

This NIP was a mess and wasn't being followed and there were many wild standards not documented. We should try to agree and actually follow some standards.

Also I've simplified the format to display some clearly defined simple tables so it's easier for people to add more stuff here.

We should stick to the 1xxxx range going forward for lists with standard meaning for which users are only expected to have one, 30001 for random lists and new 3xxxx lists for lists with standard meaning for which users are expected to have more than one.

---

**Update:**

This actually turned out to be a big movement involving tons of client developers in rehabilitating NIP-51, writing down old standards, creating new ones and cleaning up some mess from the past.

### DIFF
--- a/51.md
+++ b/51.md
@@ -30,12 +30,12 @@ rely on them to augment the user profile or browsing experience.
 For example, _mute lists_ can contain the public keys of spammers and bad actors users don't want to see in
 their feeds or receive annoying notifications from.
 
-| name             | kind  | "d" tag         | description                                               | expected tag items                                |
-| ---              | ---   | ---             | ---                                                       | ---                                               |
-| Mute list        | 10000 |                 | things the user doesn't want to see in their feeds        | `"p"` (pubkeys), `"t"` (hashtags)                 |
-| Bookmarks list   | 30001 | `"bookmark"`    | things the user intends to save for the future            | `"e"` (kind:1 notes), `"a"` (kind:30023 articles) |
-| Pin list         | 30001 | `"pin"`         | events the user intends to showcase in their profile page | `"e"` (kind:1 notes)                              |
-| Communities list | 30001 | `"communities"` | [NIP-72](72.md) communities the user belongs to           | `"a"` (kind:34550 community definitions)          |
+| name             | kind  | description                                               | expected tag items                                |
+| ---              | ---   | ---                                                       | ---                                               |
+| Mute list        | 10000 | things the user doesn't want to see in their feeds        | `"p"` (pubkeys), `"t"` (hashtags)                 |
+| Pin list         | 10001 | events the user intends to showcase in their profile page | `"e"` (kind:1 notes)                              |
+| Bookmarks list   | 10003 | things the user intends to save for the future            | `"e"` (kind:1 notes), `"a"` (kind:30023 articles) |
+| Communities list | 10004 | [NIP-72](72.md) communities the user belongs to           | `"a"` (kind:34550 community definitions)          |
 
 ## Sets
 
@@ -59,11 +59,12 @@ Aside from their main identifier, the `"d"` tag, sets can optionally have a `"ti
 
 Some clients have used these lists in the past, but they should work on transitioning to the [standard formats](#standard-lists) above:
 
-- kind:30000 and `d` "mute"
-
-| kind  | "d" tag  | use instead            |
-| ---   | ---      | ---                    |
-| 30000 | `"mute"` | kind 10000 _mute list_ |
+| kind  | "d" tag         | use instead                   |
+| ---   | ---             | ---                           |
+| 30000 | `"mute"`        | kind 10000 _mute list_        |
+| 30001 | `"pin"`         | kind 10001 _pin list_         |
+| 30001 | `"bookmark"`    | kind 10003 _bookmarks list_   |
+| 30001 | `"communities"` | kind 10004 _communities list_ |
 
 ## Examples
 
--- a/51.md
+++ b/51.md
@@ -16,9 +16,41 @@ The kinds `30000` and `30001` have been reserved for generic lists. These must b
 
 ## Standard lists
 
-| name           | kind  | "d" tag        | description                                                                                   | expected tag items                                |
-| ---            | ---   | ---            | ---                                                                                           | ---                                               |
-| Muted list     | 30000 | `"mute"`       | things the user doesn't want to see in their feeds                                            | `"p"` (pubkeys), `"t"` (hashtags)                 |
-| Bookmarks list | 30001 | `"bookmark"`   | things the user intends to save for the future                                                | `"e"` (kind:1 notes), `"a"` (kind:30023 articles) |
-| Pin list       | 30001 | `"pin"`        | events the user intends to showcase in their profile page                                     | `"e"` (kind:1 notes)                              |
-| Relay sets     | 30002 | (user-defined) | user-defined relay groups the user can easily pick and choose from during variadic operations | `"relay"` (relay URLs)                            |
+| name             | kind  | "d" tag         | description                                                                                   | expected tag items                                |
+| ---              | ---   | ---             | ---                                                                                           | ---                                               |
+| Mute list       | 30000 | `"mute"`        | things the user doesn't want to see in their feeds                                            | `"p"` (pubkeys), `"t"` (hashtags)                 |
+| Bookmarks list   | 30001 | `"bookmark"`    | things the user intends to save for the future                                                | `"e"` (kind:1 notes), `"a"` (kind:30023 articles) |
+| Pin list         | 30001 | `"pin"`         | events the user intends to showcase in their profile page                                     | `"e"` (kind:1 notes)                              |
+| Communities list | 30001 | `"communities"` | [NIP-72](72.md) communities the user belongs to                                               | `"a"` (kind:34550 community definitions)          |
+| Relay sets       | 30002 | (user-defined)  | user-defined relay groups the user can easily pick and choose from during variadic operations | `"relay"` (relay URLs)                            |
+
+## Examples
+
+### A _Mute list_ with some public items and some encrypted items
+
+```json
+{
+  "id": "a92a316b75e44cfdc19986c634049158d4206fcc0b7b9c7ccbcdabe28beebcd0",
+  "pubkey": "854043ae8f1f97430ca8c1f1a090bdde6488bd5115c7a45307a2a212750ae4cb",
+  "created_at": 1699597889,
+  "kind": 30000,
+  "tags": [
+    ["d", "mute"],
+    ["p", "07caba282f76441955b695551c3c5c742e5b9202a3784780f8086fdcdc1da3a9"],
+    ["p", "a55c15f5e41d5aebd236eca5e0142789c5385703f1a7485aa4b38d94fd18dcc4"]
+  ],
+  "content": "TJob1dQrf2ndsmdbeGU+05HT5GMnBSx3fx8QdDY/g3NvCa7klfzgaQCmRZuo1d3WQjHDOjzSY1+MgTK5WjewFFumCcOZniWtOMSga9tJk1ky00tLoUUzyLnb1v9x95h/iT/KpkICJyAwUZ+LoJBUzLrK52wNTMt8M5jSLvCkRx8C0BmEwA/00pjOp4eRndy19H4WUUehhjfV2/VV/k4hMAjJ7Bb5Hp9xdmzmCLX9+64+MyeIQQjQAHPj8dkSsRahP7KS3MgMpjaF8nL48Bg5suZMxJayXGVp3BLtgRZx5z5nOk9xyrYk+71e2tnP9IDvSMkiSe76BcMct+m7kGVrRcavDI4n62goNNh25IpghT+a1OjjkpXt9me5wmaL7fxffV1pchdm+A7KJKIUU3kLC7QbUifF22EucRA9xiEyxETusNludBXN24O3llTbOy4vYFsq35BeZl4v1Cse7n2htZicVkItMz3wjzj1q1I1VqbnorNXFgllkRZn4/YXfTG/RMnoK/bDogRapOV+XToZ+IvsN0BqwKSUDx+ydKpci6htDRF2WDRkU+VQMqwM0CoLzy2H6A2cqyMMMD9SLRRzBg==?iv=S3rFeFr1gsYqmQA7bNnNTQ==",
+  "sig": "1173822c53261f8cffe7efbf43ba4a97a9198b3e402c2a1df130f42a8985a2d0d3430f4de350db184141e45ca844ab4e5364ea80f11d720e36357e1853dba6ca"
+}
+```
+
+## Encryption process pseudocode
+
+```scala
+val private_items = [
+  ["p", "07caba282f76441955b695551c3c5c742e5b9202a3784780f8086fdcdc1da3a9"],
+  ["a", "a55c15f5e41d5aebd236eca5e0142789c5385703f1a7485aa4b38d94fd18dcc4"],
+]
+val base64blob = nip04.encrypt(json.encode_to_string(private_items))
+event.content = base64blob
+```
--- a/51.md
+++ b/51.md
@@ -52,9 +52,9 @@ Aside from their main identifier, the `"d"` tag, sets can optionally have a `"ti
 
 | name          | kind  | description                                                                                   | expected tag items                                |
 | ---           | ---   | ---                                                                                           | ---                                               |
+| Follow sets   | 30000 | categorized groups of users a client may choose to check out in different circumstances       | `"p"` (pubkeys)                                   |
 | Relay sets    | 30002 | user-defined relay groups the user can easily pick and choose from during variadic operations | `"relay"` (relay URLs)                            |
 | Curation sets | 30004 | groups of articles picked by users as interesting and/or belonging to the same category       | `"a"` (kind:30023 articles), `"e"` (kind:1 notes) |
-| Follow sets   | 30005 | categorized groups of users a client may choose to check out in different circumstances       | `"p"` (pubkeys)                                   |
 
 ## Deprecated standard lists
 
--- a/51.md
+++ b/51.md
@@ -50,10 +50,11 @@ can be used by apps to showcase curations made by others tagged to different top
 Aside from their main identifier, the `"d"` tag, sets can optionally have a `"title"`, an `"image"` and a
 `"description"` tag that can be used to enhance their UI.
 
-| name                  | kind  | description                                                                                   | expected tag items                                |
-| ---                   | ---   | ---                                                                                           | ---                                               |
-| Relay sets            | 30002 | user-defined relay groups the user can easily pick and choose from during variadic operations | `"relay"` (relay URLs)                            |
-| Article Curation sets | 30004 | groups of articles picked by users as interesting and/or belonging to the same category       | `"a"` (kind:30023 articles), `"e"` (kind:1 notes) |
+| name          | kind  | description                                                                                   | expected tag items                                |
+| ---           | ---   | ---                                                                                           | ---                                               |
+| Relay sets    | 30002 | user-defined relay groups the user can easily pick and choose from during variadic operations | `"relay"` (relay URLs)                            |
+| Curation sets | 30004 | groups of articles picked by users as interesting and/or belonging to the same category       | `"a"` (kind:30023 articles), `"e"` (kind:1 notes) |
+| Follow sets   | 30005 | categorized groups of users a client may choose to check out in different circumstances       | `"p"` (pubkeys)                                   |
 
 ## Deprecated standard lists
 
--- a/51.md
+++ b/51.md
@@ -47,8 +47,7 @@ For example, _relay sets_ can be displayed in a dropdown UI to give users the op
 relays they will publish an event or from which relays they will read the replies to an event; _curation sets_
 can be used by apps to showcase curations made by others tagged to different topics.
 
-Aside from their main identifier, the `"d"` tag, sets can optionally have a `"name"`, a `"picture"` and an
-`"about"` tag that can be used to enhance their UI.
+Aside from their main identifier, the `"d"` tag, sets can optionally have a `"title"`, an `"image"` and a `"description"` tags that can be used to enhance their UI.
 
 | name          | kind  | description                                                                                  | expected tag items                                                                |
 | ---           | ---   | ---                                                                                          | ---                                                                               |
--- a/51.md
+++ b/51.md
@@ -18,9 +18,6 @@ stored in the `.content`.
 
 ### Generic lists
 
-The kind `30001` has been reserved for generic lists. These must be accompanied by a `d` tag identifying the
-list, but these have no standard meaning and are generally client-specific (except in the standard cases
-specified below).
 
 ## Standard lists
 
--- a/51.md
+++ b/51.md
@@ -34,6 +34,7 @@ their feeds or receive annoying notifications from.
 | Public chats   | 10005 | [NIP-28](28.md) chat channels the user is in                | `"e"` (kind:40 community definitions)                                             |
 | Blocked relays | 10006 | relays clients should never connect to                      | `"relay"` (relay URLs)                                                            |
 | Search relays  | 10007 | relays clients should use when performing search queries    | `"relay"` (relay URLs)                                                            |
+| Interests      | 10015 | topics a user may be interested in and pointers             | `"t"` (hashtags) and `"a" (kind:30015 interest set)`                              |
 | Emojis         | 10030 | user preferred emojis and pointers to emoji sets            | `"emoji"` (see [NIP-30](30.md)) and `"a"` (kind:30030 emoji set)                  |
 
 ## Sets
@@ -55,6 +56,7 @@ Aside from their main identifier, the `"d"` tag, sets can optionally have a `"na
 | Relay sets    | 30002 | user-defined relay groups the user can easily pick and choose from during various operations | `"relay"` (relay URLs)                                                            |
 | Bookmark sets | 30003 | user-defined bookmarks categories , for when bookmarks must be in labeled separate groups    | `"e"` (kind:1 notes), `"a"` (kind:30023 articles), `"t"` (hashtags), `"r" (URLs)` |
 | Curation sets | 30004 | groups of articles picked by users as interesting and/or belonging to the same category      | `"a"` (kind:30023 articles), `"e"` (kind:1 notes)                                 |
+| Interest sets | 30015 | interest topics represented by a bunch of "hashtags"                                         | `"t"` (hashtags)                                                                  |
 | Emoji sets    | 30030 | categorized emoji groups                                                                     | `"emoji"` (see [NIP-30](30.md))                                                   |
 
 ## Deprecated standard lists
--- a/51.md
+++ b/51.md
@@ -6,148 +6,19 @@ Lists
 
 `draft` `optional` `author:fiatjaf` `author:arcbtc` `author:monlovesmango` `author:eskema` `author:gzuuus`
 
-A "list" event is defined as having a list of public and/or private tags. Public tags will be listed in the event `tags`. Private tags will be encrypted in the event `content`. Encryption for private tags will use [NIP-04 - Encrypted Direct Message](04.md) encryption, using the list author's private and public key for the shared secret. A distinct event kind should be used for each list type created.
+This NIP defines lists of things that users can create. Lists can contain references to anything, and these references can be **public** or **private**.
 
-If a list should only be defined once per user (like the "mute" list) the list is declared as a _replaceable event_. These lists may be referred to as "replaceable lists". Otherwise, the list is a _parameterized replaceable event_ and the list name will be used as the `d` tag. These lists may be referred to as "parameterized replaceable lists".
+Public items in a list are specified in the event `tags` array, while private items are specified in a JSON array that mimics the structure of the event `tags` array, but stringified and encrypted using the same scheme from [NIP-04](04.md) (the shared key is computed using the author's public and private key) and stored in the `.content`.
 
-## Replaceable List Event Example
+## Generic lists
 
-Lets say a user wants to create a 'Mute' list and has keys:
-```
-priv: fb505c65d4df950f5d28c9e4d285ee12ffaf315deef1fc24e3c7cd1e7e35f2b1
-pub: b1a5c93edcc8d586566fde53a20bdb50049a97b15483cb763854e57016e0fa3d
-```
-The user wants to publicly include these users:
+The kinds `30000` and `30001` have been reserved for generic lists. These must be accompanied by a `d` tag identifying the list, but these are generally client-specific, except in the standard cases specified below.
 
-```json
-["p", "3bf0c63fcb93463407af97a5e5ee64fa883d107ef9e558472c4eb9aaaefa459d"],
-["p", "32e1827635450ebb3c5a7d12c1f8e7b2b514439ac10a67eef3d9fd9c5c68e245"]
-```
-and privately include these users (below is the JSON that would be encrypted and placed in the event content):
+## Standard lists
 
-```json
-[
-    ["p", "9ec7a778167afb1d30c4833de9322da0c08ba71a69e1911d5578d3144bb56437"],
-    ["p", "8c0da4862130283ff9e67d889df264177a508974e2feb96de139804ea66d6168"]
-]
-```
-
-Then the user would create a 'Mute' list event like below:
-
-```json
-{
-  "kind": 10000,
-  "tags": [
-    ["p", "3bf0c63fcb93463407af97a5e5ee64fa883d107ef9e558472c4eb9aaaefa459d"],
-    ["p", "32e1827635450ebb3c5a7d12c1f8e7b2b514439ac10a67eef3d9fd9c5c68e245"],
-  ],
-  "content": "VezuSvWak++ASjFMRqBPWS3mK5pZ0vRLL325iuIL4S+r8n9z+DuMau5vMElz1tGC/UqCDmbzE2kwplafaFo/FnIZMdEj4pdxgptyBV1ifZpH3TEF6OMjEtqbYRRqnxgIXsuOSXaerWgpi0pm+raHQPseoELQI/SZ1cvtFqEUCXdXpa5AYaSd+quEuthAEw7V1jP+5TDRCEC8jiLosBVhCtaPpLcrm8HydMYJ2XB6Ixs=?iv=/rtV49RFm0XyFEwG62Eo9A==",
-  ...other fields
-}
-```
-
-
-## Parameterized Replaceable List Event Example
-
-Lets say a user wants to create a 'Categorized People' list of `nostr` people and has keys:
-```
-priv: fb505c65d4df950f5d28c9e4d285ee12ffaf315deef1fc24e3c7cd1e7e35f2b1
-pub: b1a5c93edcc8d586566fde53a20bdb50049a97b15483cb763854e57016e0fa3d
-```
-The user wants to publicly include these users:
-
-```json
-["p", "3bf0c63fcb93463407af97a5e5ee64fa883d107ef9e558472c4eb9aaaefa459d"],
-["p", "32e1827635450ebb3c5a7d12c1f8e7b2b514439ac10a67eef3d9fd9c5c68e245"]
-```
-and privately include these users (below is the JSON that would be encrypted and placed in the event content):
-
-```json
-[
-    ["p", "9ec7a778167afb1d30c4833de9322da0c08ba71a69e1911d5578d3144bb56437"],
-    ["p", "8c0da4862130283ff9e67d889df264177a508974e2feb96de139804ea66d6168"]
-]
-```
-
-Then the user would create a 'Categorized People' list event like below:
-
-```json
-{
-  "kind": 30000,
-  "tags": [
-    ["d", "nostr"],
-    ["p", "3bf0c63fcb93463407af97a5e5ee64fa883d107ef9e558472c4eb9aaaefa459d"],
-    ["p", "32e1827635450ebb3c5a7d12c1f8e7b2b514439ac10a67eef3d9fd9c5c68e245"],
-  ],
-  "content": "VezuSvWak++ASjFMRqBPWS3mK5pZ0vRLL325iuIL4S+r8n9z+DuMau5vMElz1tGC/UqCDmbzE2kwplafaFo/FnIZMdEj4pdxgptyBV1ifZpH3TEF6OMjEtqbYRRqnxgIXsuOSXaerWgpi0pm+raHQPseoELQI/SZ1cvtFqEUCXdXpa5AYaSd+quEuthAEw7V1jP+5TDRCEC8jiLosBVhCtaPpLcrm8HydMYJ2XB6Ixs=?iv=/rtV49RFm0XyFEwG62Eo9A==",
-  ...other fields
-}
-```
-
-Lets say a user wants to create a 'Categorized Bookmarks' list of `bookmarks` and has keys:
-```
-priv: fb505c65d4df950f5d28c9e4d285ee12ffaf315deef1fc24e3c7cd1e7e35f2b1
-pub: b1a5c93edcc8d586566fde53a20bdb50049a97b15483cb763854e57016e0fa3d
-```
-The user wants to publicly include these bookmarks:
-
-```json
-["e", "5c83da77af1dec6d7289834998ad7aafbd9e2191396d75ec3cc27f5a77226f36", "wss://nostr.example.com"],
-["a", "30023:f7234bd4c1394dda46d09f35bd384dd30cc552ad5541990f98844fb06676e9ca:abcd", "wss://nostr.example.com"],
-["r", "https://github.com/nostr-protocol/nostr", "Nostr repository"],
-```
-and privately include these bookmarks (below is the JSON that would be encrypted and placed in the event content):
-
-```json
-[
-    ["r", "https://my-private.bookmark", "My private bookmark"],
-    ["a", "30001:f7234bd4c1394dda46d09f35bd384dd30cc552ad5541990f98844fb06676e9ca:abcd", "wss://nostr.example.com"],
-]
-```
-
-Then the user would create a 'Categorized Bookmarks' list event like below:
-
-```json
-{
-  "kind": 30001,
-  "tags": [
-    ["d", "bookmarks"],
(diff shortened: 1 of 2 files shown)


mattn:
Sounds good to me.

mattn (approved):

51.md
```diff
@@ -6,148 +6,51 @@ Lists
...
-    ["p", "9ec7a778167afb1d30c4833de9322da0c08ba71a69e1911d5578d3144bb56437"],
-    ["p", "8c0da4862130283ff9e67d889df264177a508974e2feb96de139804ea66d6168"]
-]
-```
+## Standard lists
 
-Then the user would create a 'Mute' list event like below:
+| name             | kind  | "d" tag         | description                                                                                   | expected tag items                                |
+| ---              | ---   | ---             | ---                                                                                           | ---                                               |
+| Mute list       | 30000 | `"mute"`        | things the user doesn't want to see in their feeds                                            | `"p"` (pubkeys), `"t"` (hashtags)                 |
```
erskingardner, line 21: This makes me sad. I know it's super common so we should document it but this feels very superfluous given we have 10000 mute lists.

mostafa-khaldi:
We're gonna close [this](https://github.com/nostr-protocol/nips/pull/492) PR due to the new kind:30004 and we're gonna comply with its new standards, please confirm the new changes so we can close this PR

mostafa-khaldi:
@fiatjaf We're gonna close [this](https://github.com/nostr-protocol/nips/pull/492) PR due to the new kind:30004 and we're gonna comply with its new standards, please confirm the new changes so we can close [this](https://github.com/nostr-protocol/nips/pull/492) PR

51.md
fiatjaf: OK, I just learned that Snort used kind 30000/"mute" but has now changed to kind 10000, so tides are in our favor. I changed the NIP and let's try to get this right this time.

fiatjaf:
Who is using kind 30000 with `d` set to `"mute"`? Can we work on a standardization plan here to move everybody to kind 10000? I know zap.stream, snort.social, coracle.social and nostrudel.ninja are using kind 10000.

fiatjaf:
It would also be great if people could agree to move from kind 30001 with `d` tags "pin", "bookmark" and "communities" to dedicated kinds 10001, 10003 and 10004 respectively -- but I won't push for that. Who is using these lists, in any case?

51.md
```diff
@@ -6,148 +6,153 @@ Lists
...
-[
-    ["p", "9ec7a778167afb1d30c4833de9322da0c08ba71a69e1911d5578d3144bb56437"],
-    ["p", "8c0da4862130283ff9e67d889df264177a508974e2feb96de139804ea66d6168"]
-]
-```
+## Standard lists
 
-Then the user would create a 'Mute' list event like below:
+Users are expected to have a single list of each of these types. They have special meaning and clients may
+rely on them to augment the user profile or browsing experience.
```
staab, line 28: `d` should not be both client-defined *and* user-defined on the same kind, because that results in an overloaded namespace. What if a user creates a list called "communities"? I suppose `d` could be a uuid or something instead. But even then, clients that only want to show user-defined lists have to whitelist the standard `d` tags. Standard lists should be different kinds: 10001 for bookmarks, 10002 for pins, 10003 for communities. This is how the old spec was written, although those kinds weren't specified and people ended up using 30001 for them:

> ### Categorized Bookmarks List
> 
> An event of kind `30001` is defined as a parameterized replaceable list event for categorizing bookmarks. The 'd' parameter for this event holds the category name of the list.

"The 'd' parameter for this event holds the category name of the list." This is how Coracle uses kind 30001 — by setting the `d` tag to the `title` of the list (I didn't actually know about title or description, I don't know when they were introduced). 

Can we not break my correct implementation because of widespread incorrect ones?

staab (changes requested):

staab:
>It would also be great if people could agree to move from kind 30001 with d tags "pin", "bookmark" and "communities" to dedicated kinds 10001, 10003 and 10004 respectively -- but I won't push for that.

I will, I think it's important.

jb55:
On Wed, Nov 15, 2023 at 05:38:29AM -0800, Vitor Pamplona wrote:
>Looks good. If this gets approval, I will switch Amethyst from using
>kind 30000's with d="mute" for mute lists to the 10000.

sure I will do the same.

jb55:
> > It would also be great if people could agree to move from kind 30001 with d tags "pin", "bookmark" and "communities" to dedicated kinds 10001, 10003 and 10004 respectively -- but I won't push for that.
> 
> I will, I think it's important.

curious as to why? I haven't followed these discussions.

staab:
@jb55 see my explanation in [this comment](https://github.com/nostr-protocol/nips/pull/880#discussion_r1394505239).

Since early on, user-defined lists in Coracle have been polluted with client-generated lists from other clients which don't make sense as user-generated lists (like "communities"). This is confusing to users because they never made those lists, and since clients can use any `d` tag they feel like, there's no way to really keep those things from showing up where they shouldn't.

jb55:
@staab not sure what you mean, do you query 30000 without a #d tag?

mikedilger:
Gossip has implemented categorized people list at 30000 with user-chosen 'd' tags.  But the UI still hasn't exposed this implementation so nobody is using it yet, meaning we could still change things.

The only list that is widely used in gossip is the Mute list 10000.

fiatjaf:
@staab although I agree with you that it sucks that some apps are creating a million garbage lists for people, I don't think you were following the spec either because the spec was unfollowable. https://coracle.social/ uses kind 30001 for bookmarks, but puts "people" in them, and not on the kind 30000 which was supposed to be for people. Anyway, it's a big mess.

I think the best solution is to leave kind 30001 as the chaotic free-for-all land and create other kinds for much more well-defined and smaller in scope lists.

"Bookmark" is probably too generic already. On Chrome and in Twitter you can bookmark tweets and they just get thrown in the same list. Coracle is using bookmarks to mean different sets of people that the user might be interested in checking out from time to time, is that it? This is actually a good idea that had never occurred to me and I probably others.

I think we could create kind 30005: "Follow sets" and define it to be strictly about the way Coracle is doing.

staab:
> @staab not sure what you mean, do you query 30000 without a #d tag?

Coracle uses kind 30001, but yes, that's how I implemented my lists feature (see below). In the screenshot you can see an example of the problem I'm describing. I used some client to back up my follows list, and now it's showing up as a custom list. That's a specific use case not defined by a user, that client should have used a new `100xx` kind to represent it.

Screenshot 2023-11-15 at 10 04 14 AM

staab:
> I don't think you were following the spec either because the spec was unfollowable.

That's fair, I actually asked for my use case (heterogenous lists) when we were drafting the spec but no one wanted to add it. Maybe 30001 should be client-defined lists, and new kinds should be added for user-defined heterogenous lists and other stuff.

monlovesmango:
> It would also be great if people could agree to move from kind 30001 with `d` tags "pin", "bookmark" and "communities" to dedicated kinds 10001, 10003 and 10004 respectively -- but I won't push for that. What apps are using these lists, in any case?

I agree with this, but will push for it. 

30001 should be more of a free for all kind that clients can use for whatever unique usecase they want to implement.

any list that has a shared use case across clients, has "expected tag items" types, and only should have one list per user (like "pin", "bookmark", "communities") should have a dedicated 1000X kind.

if we are taking the time to "fix" nip51 and actually start following some standards, now seems like the perfect time to push for "pin", "bookmark", and "communities" kinds.

staab:
> Coracle is using bookmarks to mean different sets of people that the user might be interested in checking out from time to time, is that it?

Coracle is using "bookmarks" in the broadest sense possible, i.e. "something". So `p`, `e`, `t`, `r` are all admissible, and I interpret them as custom feeds. That's certainly weird and I want to revisit custom feeds at some point (pablo told me about `nreq` the other day, seems promising). But the point I'm making here is the difference between client-defined and user-defined lists.

mikedilger:
In general if a user is naming a list, it should be in the 3xxxx range, but if a list has a well-known meaning it should be 1xxxx, so I think we should keep mute=10000, pin=10001 (as they already are), and also leave categorizedpeople=30000, categorizedbookmarks=30001 as they are as well.

So I don't understand why this change.  I don't know why people put "pin" and "communities" under 30001 that seems very wrong to me (I agree with Staab's comments regarding this).

Who is putting 'communities' under 30001?  Who is putting 'pin' under 30001?  Those clients should be fixed.

monlovesmango:
> In general if a user is naming a list, it should be in the 3xxxx range, but if a list has a well-known meaning it should be 1xxxx, so I think we should keep mute=10000, pin=10001 (as they already are), and also leave categorizedpeople=30000, categorizedbookmarks=30001 as they are as well.
> 
> So I don't understand why this change. I don't know why people put "pin" and "communities" under 30001 that seems very wrong to me (I agree with Staab's comments regarding this).
> 
> Who is putting 'communities' under 30001? Who is putting 'pin' under 30001? Those clients should be fixed.

agree. and categorized bookmarks seems like the perfect kind to allow clients to implement whatever new list types they want.

staab:
> agree. and categorized bookmarks seems like the perfect kind to allow clients to implement whatever new list types they want.

I don't think so, because clients would be the ones naming these new list types, resulting in some lists being user-defined and others client-defined as we see above.

monlovesmango:
> @staab although I agree with you that it sucks that some apps are creating a million garbage lists for people, I don't think you were following the spec either because the spec was unfollowable. https://coracle.social/ uses kind 30001 for bookmarks, but puts "people" in them, and not on the kind 30000 which was supposed to be for people. Anyway, it's a big mess.
> 
> I think the best solution is to leave kind 30001 as the chaotic free-for-all land and create other kinds for much more well-defined and smaller in scope lists.
> 
> "Bookmark" is probably too generic already. On Chrome and in Twitter you can bookmark tweets and they just get thrown in the same list. Coracle is using bookmarks to mean different sets of people that the user might be interested in checking out from time to time, is that it? This is actually a good idea that had never occurred to me and I probably others.
> 
> I think we could create kind 30005: "Follow sets" and define it to be strictly about the way Coracle is doing.

kind 3000 "Categorized People" was created to be used as "follow sets"

fiatjaf:
About the mute lists, we have so far:

- using kind 10000:
  - Coracle
  - Snort
  - Nostrudel
  - Gossip
  - zap.stream
- using kind 30000 + "mute" tag:
  - Damus
  - Amethyst

From looking at the source code of nos.social, Nostur, yana.do it seems they don't have this implemented. There are other clients that I know don't have these features either. Please tag client developers and let's see if everybody agrees.

@mbrat1 @frnandu @mplorentz @syusui-s @lovvtide 

Please answer about the 30001 "pin", "bookmark" and "communities" lists too. I'll change the PR to reflect the new kinds.

monlovesmango:
> I don't think so, because clients would be the ones naming these new list types, resulting in some lists being user-defined and others client-defined as we see above.

hmm true. maybe we need dedicated 3000X kind for client defined lists, and all other lists should be strictly user defined? is that what you've been trying to suggest?

fiatjaf:
@mikedilger 

> so I think we should keep mute=10000, pin=10001 (as they already are), and also leave categorizedpeople=30000, categorizedbookmarks=30001 as they are as well.
> So I don't understand why this change. I don't know why people put "pin" and "communities" under 30001 that seems very wrong to me (I agree with Staab's comments regarding this).

There was never a pin=10001 in this NIP as far as I know. "categorized people" and "categorized bookmarks" are too broad in names and because of that I think clients developers feel like they can put anything in these -- that combined with a general impression that kinds are scarce so they should reuse kinds and never add new kinds to the NIP. We must change that going forward.

monlovesmango:
> There was never a pin=10001 in this NIP as far as I know.

I'm seeing it in the current NIP 51 spec... am I missing something...?
https://github.com/nostr-protocol/nips/blob/master/51.md

image

fiatjaf:
> @fiatjaf I'm seeing it in the current NIP 51 spec... am I missing something...?
master/51.md

No, I'm just crazy.

staab:
> maybe we need dedicated 3000X kind for client defined lists, and all other lists should be strictly user defined? is that what you've been trying to suggest?

Some dedicated kind for client defined lists, yeah. I would make it a `100xx` though probably

monlovesmango:
> Some dedicated kind for client defined lists, yeah. I would make it a `100xx` though probably

but then clients could only create one list per user max, you think this will suffice?

staab:
> but then clients could only create one list per user max, you think this will suffice?

No, you'd use `d` tags, just client-defined instead of user-defined.

@fiatjaf I like your solution of completely deprecating kind 30001.

fiatjaf:
Clients should use NIP-78 for their client-specific ad-hoc lists. Looking at you @gzuuus.

monlovesmango:
@fiatjaf make "follow sets" kind 30000, it aligns with the "categorized people" kind from current spec. agree "follow sets" is a better name.

staab:
> Clients should use NIP-78 for their client-specific ad-hoc lists. Looking at you @gzuuus.

I actually disagree, I think it's good to put those in lists so other clients can add support if desired. They can do that anyway with NIP 78, but this at least indicates that that's an acceptable thing to do.

monlovesmango:
> No, you'd use `d` tags, just client-defined instead of user-defined.

I don't think you can use `d` tags on 100XX kind events..? am I missing something? did you mean to say 300XX?

fiatjaf:
> I actually disagree, I think it's good to put those in lists so other clients can add support if desired. They can do that anyway with NIP 78, but this at least indicates that that's an acceptable thing to do.

Try adding support to this:

2023-11-14-191448_374x1054_scrot

monlovesmango (approved):
love it

staab:
> I don't think you can use d tags on 100XX kind events..? am I missing something? did you mean to say 300XX?

Oh, that's a good point, I forgot `1xxxx`s were not parameterized. Kind 3xxxx makes sense then. Or NIP 78, I don't have a strong refutation of @fiatjaf's screenshot.

hzrd149:
I am completely onboard with this, right now noStrudel uses kind:30001 "communities" but ill start transitioning it to use kind:10004

51.md
```diff
@@ -6,148 +6,128 @@ Lists
...
+    [
+      "d",
+      "jvdy9i4"
+    ],
+    [
+      "title",
+      "Yaks"
+    ],
+    [
+      "summary",
```
hzrd149, line 107: is it supposed to be a "summary" or "description" tag? because earlier on line 50 it says "title", "image", and "description" tags

frnandu:
I'm starting to implement relay sets (using kind 30002 - as suggested in the merged recently PR) in yana for blocked & search relay sets. Will give option to either use with defined  d names "blocked" & "search" or be able to choose other relay sets that the user might have of kind 30002 with custom names. Is this ok or should we standardize d names for such common usecases?

staab:
> I'm starting to implement relay sets (using kind 30002 - as suggested in the merged recently PR) in yana for blocked & search relay sets. Will give option to either use with defined d names "blocked" & "search" or be able to choose other relay sets that the user might have of kind 30002 with custom names. Is this ok or should we standardize d names for such common usecases?

This is sort of interesting. I think it would be better to standardize this as a 1xxxx kind, but it could make sense to add a `purpose` tag to 30002 instead to indicate what the relay set is for.

vitorpamplona:
Can we add a Chat List as `kind:10005` pointing to `kind:40`s? 

Can we add the `word` tag in mute lists from https://github.com/nostr-protocol/nips/pull/673 ? 

Amethyst already uses the `word` tag in those lists.

monlovesmango:
agree with @staab . anytime we're thinking of standardizing "d" tags is an indicator that it probably should be spec'ed as a new 100XX kind.

generally "d" tags should be user defined, not client defined.

51.md
```diff
@@ -6,148 +6,128 @@ Lists
...
+### Generic lists
 
-```json
-["p", "3bf0c63fcb93463407af97a5e5ee64fa883d107ef9e558472c4eb9aaaefa459d"],
-["p", "32e1827635450ebb3c5a7d12c1f8e7b2b514439ac10a67eef3d9fd9c5c68e245"]
-```
-and privately include these users (below is the JSON that would be encrypted and placed in the event content):
+The kind `30001` has been reserved for generic lists. These must be accompanied by a `d` tag identifying the
+list, but these have no standard meaning and are generally client-specific (except in the standard cases
+specified below).
```
staab, line 23:
```suggestion
```

51.md
```diff
@@ -6,148 +6,128 @@ Lists
...
-[
-    ["p", "9ec7a778167afb1d30c4833de9322da0c08ba71a69e1911d5578d3144bb56437"],
-    ["p", "8c0da4862130283ff9e67d889df264177a508974e2feb96de139804ea66d6168"]
-]
-```
+## Standard lists
 
-Then the user would create a 'Mute' list event like below:
+Users are expected to have a single list of each of these types. They have special meaning and clients may
+rely on them to augment the user profile or browsing experience.
```
staab, line 28:
```suggestion
Standard lists use non-parameterized replaceable events, meaning users may only have a single list
of each kind. They have special meaning and clients may rely on them to augment a user's profile or
browsing experience.
```

SnowCait:
I vote `kind:10005` and `word` tag in mute lists.
In addition, can we add the `e` tag pointing to notes and other kind stuff in mute lists? I'd like to mute hell thread.
[nostter](https://github.com/SnowCait/nostter) uses `kind:10000` including `word` and `e` tags.

>Bookmarks list (`kind:10003`)

Bookmarks list will be large and we won't be able to find notes from one large list.
I think it is better to use `kind:3000x` with user-defined `d` tags.

>Standard lists

Relays have limit of tags or event.
It would be better every standard lists (and `kind:3`) can contain `a` tag which points to generic lists or sets.

Examples: (I'm not familiar with relay implementation, sorry if wrong)
- strfry has [events__maxNumTags](https://github.com/hoytech/strfry/blob/719654779acef41b54e826d765139a055fb254fb/golpe.yaml#L140) (default is 2,000)
- nostr-rs-relay has [max_event_bytes](https://github.com/scsibug/nostr-rs-relay/blob/8a67770206e6f8ea281ed4e75b2a421ec8764afc/src/config.rs#L69)

51.md
```diff
@@ -6,148 +6,128 @@ Lists
...
-  "content": "VezuSvWak++ASjFMRqBPWS3mK5pZ0vRLL325iuIL4S+r8n9z+DuMau5vMElz1tGC/UqCDmbzE2kwplafaFo/FnIZMdEj4pdxgptyBV1ifZpH3TEF6OMjEtqbYRRqnxgIXsuOSXaerWgpi0pm+raHQPseoELQI/SZ1cvtFqEUCXdXpa5AYaSd+quEuthAEw7V1jP+5TDRCEC8jiLosBVhCtaPpLcrm8HydMYJ2XB6Ixs=?iv=/rtV49RFm0XyFEwG62Eo9A==",
-  ...other fields
-}
-```
+For example, _mute lists_ can contain the public keys of spammers and bad actors users don't want to see in
+their feeds or receive annoying notifications from.
 
+| name             | kind  | description                                               | expected tag items                                |
+| ---              | ---   | ---                                                       | ---                                               |
+| Mute list        | 10000 | things the user doesn't want to see in their feeds        | `"p"` (pubkeys), `"t"` (hashtags)                 |
```
staab, line 35: Coracle currently supports `e` tag muting, and Amethyst supports `word` muting.

51.md
```diff
@@ -6,148 +6,128 @@ Lists
...
-]
-```
+Aside from their main identifier, the `"d"` tag, sets can optionally have a `"title"`, an `"image"` and a
+`"description"` tag that can be used to enhance their UI.
+
+| name          | kind  | description                                                                                   | expected tag items                                |
+| ---           | ---   | ---                                                                                           | ---                                               |
+| Follow sets   | 30000 | categorized groups of users a client may choose to check out in different circumstances       | `"p"` (pubkeys)                                   |
+| Relay sets    | 30002 | user-defined relay groups the user can easily pick and choose from during variadic operations | `"relay"` (relay URLs)                            |
+| Curation sets | 30004 | groups of articles picked by users as interesting and/or belonging to the same category       | `"a"` (kind:30023 articles), `"e"` (kind:1 notes) |
```
staab, line 57: Suggestion (probably for another day)

```
| Filter sets | 30005 | a list of REQ filters defining custom feeds | `"req"` (a JSON-encoded REQ filter) |
```

staab (changes requested):

fiatjaf:
@SnowCait wanna include the emoji list thing?

SnowCait:
@fiatjaf Yes!

51.md
fiatjaf: Good question. What should it be?

fiatjaf: I think it's better to go with `name`, `picture` and `about`, following kind 0 nomenclature.

staab:
Thanks for doing this @fiatjaf, I didn't realize how much technical debt would be solved by this PR.

reyamir:
it sounds good to me

51.md
```diff
@@ -6,148 +6,116 @@ Lists
...
-```
-and privately include these bookmarks (below is the JSON that would be encrypted and placed in the event content):
+### A _curation set_ of articles and notes about yaks
 
-```json
-[
-    ["r", "https://my-private.bookmark", "My private bookmark"],
-    ["a", "30001:f7234bd4c1394dda46d09f35bd384dd30cc552ad5541990f98844fb06676e9ca:abcd", "wss://nostr.example.com"],
-]
 ```
```
proudmuslim-dev, line 91: Nitpick: Change this to ` ```json `

proudmuslim-dev:
> > Bookmarks list (`kind:10003`)
> 
> Bookmarks list will be large and we won't be able to find notes from one large list. I think it is better to use `kind:3000x` with user-defined `d` tags.

Agreed with @SnowCait. Bookmarks make much more sense in the 3xxxx range

syusui-s:
reply to https://github.com/nostr-protocol/nips/pull/880#issuecomment-1813035499

Rabbit doesn't support these lists as for now. I will follow the specification.

51.md
Darecjo: We're doing something similar in YakiHonne related to the user preferences for content customizations via `kind-30100`. I think its better to include such specifications once for all because soon or later feeds customization will be faced (at least thats what we experienced).

content_settings
topics

```
{
  "id": "ed4291ff8ba4343ca444d03624afcee98257aba4af17410a73144ffdc369c467",
  "kind": 30100,
  "pubkey": "1cb0ebecfbe8e291d46c4e082df1e6a183bb321245ef29df5e2703f84b1d86e1",
  "created_at": 1700102857,
  "content": "",
  "tags": [
    [
      "d",
      "MyFavoriteTopicsInYakihonne"
    ],
    [
      "t",
      "YakiHonne"
    ],
    [
      "t",
      "Technology"
    ],
    [
      "t",
      "Crypto"
    ],
    [
      "t",
      "Travel"
    ],
    [
      "t",
      "Social"
    ],
    [
      "t",
      "Nostr"
    ],
    [
      "t",
      "Economy"
    ],
    [
      "t",
      "Finance"
    ],
    [
      "t",
      "Entrepreneurship"
    ],
    [
      "t",
      "Self improvement"
    ],
    [
      "t",
      "Food"
    ],
    [
      "t",
      "Marketing"
    ],
    [
      "t",
      "Art"
    ],
    [
      "t",
      "Photography"
    ],
    [
      "t",
      "Entertainment"
    ],
    [
      "t",
      "Music"
    ]
  ],
  "sig": "4fdc6a423bc7320b432e2edc1e794895468894b92ad7bb9e119bd16c1a6c1d0098bbd8a33ac25856acbae222fda565d7407c17d2b09f6b326de3762ad30a8a1c"
}
```

lovvtide:
> It would also be great if people could agree to move from kind 30001 with `d` tags "pin", "bookmark" and "communities" to dedicated kinds 10001, 10003 and 10004 respectively -- but I won't push for that. What apps are using these lists, in any case?

Yep, Satellite currently uses kind 30001 with `d` tag "communities". I'll change it to 10004.

51.md
hzrd149: I believe listr.lol is already using "description" but I'm fine with any of them as long as there is a set standard
"about" or "summary" might be slightly better working since they imply a single line of text and not a full paragraph

TsukemonoGit:
> > > Bookmarks list (`kind:10003`)
> > 
> > 
> > Bookmarks list will be large and we won't be able to find notes from one large list. I think it is better to use `kind:3000x` with user-defined `d` tags.
> 
> Agreed with @SnowCait. Bookmarks make much more sense in the 3xxxx range

I think so too.
I already have a lot of bookmarks and without the d-tag it is not enough.
([https://nostr-bookamrk-viewer5.vercel.app/npub1sjcvg64knxkrt6ev52rywzu9uzqakgy8ehhk8yezxmpewsthst6sw3jqcw](https://nostr-bookamrk-viewer5.vercel.app/npub1sjcvg64knxkrt6ev52rywzu9uzqakgy8ehhk8yezxmpewsthst6sw3jqcw))

Darecjo:
> It would also be great if people could agree to move from kind 30001 with `d` tags "pin", "bookmark" and "communities" to dedicated kinds 10001, 10003 and 10004 respectively -- but I won't push for that. What apps are using these lists, in any case?

@fiatjaf What if the users have the ability to save bookmarks under different names (for example: A user wants to save a list of bookmarks specified for `Food` only and another one for `Sport` only. Are custom tags allowed in such case?

Screenshot 2023-11-16 at 3 03 47 PM

51.md
erskingardner: I'm using `title` and `description` tags on lists in Listr. I would love to leave those be, as they seem more applicable than `name`, `about`. I would also be in favor of using `image` instead of `picture`, keeping more in line with NIP-23 than kind 0. 

FWIW, the reason for the `title` tag is that in order to give users the ability to edit the name of a list without creating a brand new list, you have to have to keep the `d` tag constant, so I use a uuid value for the `d` tag in Listr and then show the `title` tag in the UI instead of the `d` tag. This pattern works really well for user defined lists and means that other clients can also allow users to rename their lists, as long as they respect and maintain the `d` tag.

erskingardner:
Love how much attention this change is getting... long overdue IMO.

I mentioned above in a comment but would love to get more input from other devs on this. I think that `d` tags shouldn't be human readable but instead should be a `uuid` value of some sort. The big benefit this provides is that it makes it possible to edit the user visible name of the list using the `title` tag value. If we're moving to a world where kind `1xxxx` lists have specific known uses based on the kind number and kind `3xxxx` lists are user-created, I think we need to make sure that kind `3xxxx` lists are structured in a way that allows for great UX.

51.md
Darecjo: > I'm using `title` and `description` tags on lists in Listr. I would love to leave those be, as they seem more applicable than `name`, `about`. I would also be in favor of using `image` instead of `picture`, keeping more in line with NIP-23 than kind 0.
> 
> FWIW, the reason for the `title` tag is that in order to give users the ability to edit the name of a list without creating a brand new list, you have to have to keep the `d` tag constant, so I use a uuid value for the `d` tag in Listr and then show the `title` tag in the UI instead of the `d` tag. This pattern works really well for user defined lists and means that other clients can also allow users to rename their lists, as long as they respect and maintain the `d` tag.

We're using the `d` tag for the same exact reason allowing editing while maintaining the same list.

51.md
```diff
@@ -6,148 +6,116 @@ Lists
...
-```
-and privately include these users (below is the JSON that would be encrypted and placed in the event content):
+| name          | kind  | description                                                                                   | expected tag items                                |
+| ---           | ---   | ---                                                                                           | ---                                               |
+| Follow sets   | 30000 | categorized groups of users a client may choose to check out in different circumstances       | `"p"` (pubkeys)                                   |
+| Relay sets    | 30002 | user-defined relay groups the user can easily pick and choose from during variadic operations | `"relay"` (relay URLs)                            |
+| Curation sets | 30004 | groups of articles picked by users as interesting and/or belonging to the same category       | `"a"` (kind:30023 articles), `"e"` (kind:1 notes) |
+| Emoji sets    | 30030 | categorized emoji groups                                                                      | `"emoji"` (see [NIP-30](30.md))                   |
```
erskingardner, line 57: Suggestion to add: `Interest Set` which is kind `30015`, and takes a set of `t` tags.

erskingardner (changes requested):

51.md
erskingardner: @pablof7z is it worth also mentioning Hightlight lists?

Darecjo: > We're doing something similar in YakiHonne related to the user preferences for content customizations via `kind-30100`. I think its better to include such specifications once for all because soon or later feeds customization will be faced (at least thats what we experienced).
> 
> content_settings topics
> 
> ```
> {
>   "id": "ed4291ff8ba4343ca444d03624afcee98257aba4af17410a73144ffdc369c467",
>   "kind": 30100,
>   "pubkey": "1cb0ebecfbe8e291d46c4e082df1e6a183bb321245ef29df5e2703f84b1d86e1",
>   "created_at": 1700102857,
>   "content": "",
>   "tags": [
>     [
>       "d",
>       "MyFavoriteTopicsInYakihonne"
>     ],
>     [
>       "t",
>       "YakiHonne"
>     ],
>     [
>       "t",
>       "Technology"
>     ],
>     [
>       "t",
>       "Crypto"
>     ],
>     [
>       "t",
>       "Travel"
>     ],
>     [
>       "t",
>       "Social"
>     ],
>     [
>       "t",
>       "Nostr"
>     ],
>     [
>       "t",
>       "Economy"
>     ],
>     [
>       "t",
>       "Finance"
>     ],
>     [
>       "t",
>       "Entrepreneurship"
>     ],
>     [
>       "t",
>       "Self improvement"
>     ],
>     [
>       "t",
>       "Food"
>     ],
>     [
>       "t",
>       "Marketing"
>     ],
>     [
>       "t",
>       "Art"
>     ],
>     [
>       "t",
>       "Photography"
>     ],
>     [
>       "t",
>       "Entertainment"
>     ],
>     [
>       "t",
>       "Music"
>     ]
>   ],
>   "sig": "4fdc6a423bc7320b432e2edc1e794895468894b92ad7bb9e119bd16c1a6c1d0098bbd8a33ac25856acbae222fda565d7407c17d2b09f6b326de3762ad30a8a1c"
> }
> ```

How about this from above it complies in your side and some of what @staab wants to achieve I Think his proposition is based on the advance search he has on coracle.social

fabianfabian:
> From looking at the source code of nos.social, Nostur, yana.do it seems they don't have this implemented.

Yes Nostur lists are just local to the users devices right now, I will eventually make them available for relays after the dust has settled.

erskingardner:
Anyone know what Primal is doing with lists? @pritk @nikolalukovic

gzuuus:
> Clients should use NIP-78 for their client-specific ad-hoc lists. Looking at you @gzuuus.

@fiatjaf Why not? Don't you think that external references or links are appropriate for a bookmark list? Also, some time ago I debated here about which kind to use for nostree, and it seemed that kind30001 was the most appropriate, everyone agrees #723.

On the other hand, I like the pr and think it brings improvements and clarifies the list structure, as well as reducing the technical debt for us in the future, but as @erskingardner said, I think the `d` tags should be set by the client using some form of uuid, and `title` should be used to render the title in the UI, for several reasons. 
- uuid is better at handling list identifiers and possible versions of them.
- If the value of `d` is set by the user, this can lead to problems with long strings, strings containing special characters, or formatting that makes it inconvenient to form `a` tags.
- Bad UX, because users would not be able to change the title of their lists without creating a new one.

These were the reasons that led me to the current design of nostree, they are also used like this in listr, however I think there is room for improvement as this pr reflects, so from my point of view `d` is good to be set by the client and represent a unique identifier, and the title of the list is `title` which seems more appropriate for it.

Regarding kind, I think we could keep kind30001 for user-defined categorized bookmark list, as mentioned above, as users may want to define multiple lists or sort their bookmarks in a categorized and ordered way according to their criteria, not just one flat list for everything. This doesn't make kind30001 a disaster drawer, as it is bound by the conventions established in this nip, and clients can expect standard formatting, as a disaster drawer we already have kind30078.

verbiricha:
Thanks for doing this fiatjaf, important and necessary housekeeping work.

> What if the users have the ability to save bookmarks under different names (for example: A user wants to save a list of bookmarks specified for Food only and another one for Sport only.

I agree with others that have shared this concern, I think it'd be desirable to have multiple bookmark lists instead of a single list. Other than that looks great.

mutatrum:
NIP-02 is looking over here jealously.

staab:
> Agreed with @SnowCait. Bookmarks make much more sense in the 3xxxx range

I think there are two kinds of bookmarks — unnamed bookmarks (10003), and named bookmark lists (listr's version of 30001). These should probably be able to contain anything, not just `e` and `a` tags. The latter should have a user-defined name, either in the `d` tag or as the `title` attribute with `d` being a uuid (uuid is better for the reasons @gzuuus gave).

Here's what I propose for sorting this out:

- Expand the scope of 10003 to include `p`, `r`, `t` etc tags
- Deprecate 30001 because it already has lots of user data in it
- Add 30003 as `named bookmarks` with the same rules as 10003 but with d = uuid and a user-defined `title` tag.

vitorpamplona (approved):

hzrd149:
> Should we stop using `"e"`, `"p"`, `"r"` etc in these lists and use bigger tagnames instead? That would be easier on relays.

I don't know about the "e" tags, but one use for using "p" tags in lists is you can show lists in the reverse direction. i.e you can show all the lists a user has been put into, and maybe use that as a way of recommending similar users.

vitorpamplona:
> Should we stop using "e", "p", "r" etc in these lists and use bigger tagnames instead? That would be easier on relays.

I was going to start counting bookmarks and show those in the main feed as well. Bookmarks might be more relevant to determine the quality of the post than likes and replies. However, downloading large lists per post on the client-side is quite memory-heavy.

fiatjaf:
OK, let's keep these things then.

> However, downloading large lists per post on the client-side is quite memory-heavy.

Another use case for `COUNT`?

51.md
fiatjaf: @Darecjo if your list is automatically named `MyFavoriteTopicsInYakihonne` I think that is an indication that you should be using NIP-78. What do you think?

@erskingardner would these things be a single standard list or multiple? It's hard to imagine someone having _interests_, which is such a generic concept, categorized in different groups. I feel like each interest is already a "group" on its own.

On the other hand one interest is translated into potentially huge number of tags -- i.e. if I'm interested in "fruits" I may have the hashtags "banana", "mango", "passionfruit" etc. Or if I'm interested in "banana" I may have "banana", "cavendish", "bananas", "plaintain", "banana-recipes", "cooking-bananas" whatever.

So what do we do? Add both a canonical standard interests list and interest sets?

fiatjaf:
Sorry, @gzuuus, I didn't mean to tag you in a bad way, it was mostly a joke. I think everybody was using 30001 lists in a quite ad-hoc way and that we didn't know better -- but I hope we are agreeing on something.

vitorpamplona:
> Another use case for `COUNT`?

COUNT would only work if we figure out a way to merge COUNT all the relays the client is downloading things from for a particular post.

verbiricha (approved):

frnandu:
I'm now trying to use **private** relay set, kind 30002. 
And since these are parameterized replaceable events, do I understand correctly that the **d** tag must be NOT PRIVATE in order for the replace ability to work? 
First I assumed that **d** name of the set would also be encrypted in the content, and thus private, but now I'm not so sure anymore.

vitorpamplona:
The only d-tag that matters is the public one. You can add one to private tag, but no relay will see it.

Lists can have parts that are public and parts that are private at the same time. It's the same list (same public d-tag)

51.md
erskingardner: Good question. @pablof7z added `30015` to NDK a while back and we've both been using it as a PRE. I think that there is a good reason to keep separate lists, e.g. I might have a music interests lists with music related stuff, and also an outdoors list that is all about climbing and skiing. I might want to use one or the other to view curated topic feeds.

Maybe that means there is a case to do both `10015` and `30015`, in similar style to emoji lists. Where your `10015` list is the full canonical list and can include either `t` tags or `a` tags pointing to other `30015` lists.

mbrat1:
All Primal clients already use kind 10000 for mute lists. We don't currently support pin, bookmark or communities lists, but once we do, we will go with the new 1000x kinds.

monlovesmango:
> Maybe too cryptic, but the contacts list is also a list. It would be nice to have private contacts as well. On face value this looks like a good match. It accommodates pure followers lists, but also tags, or other things and you can even have names contact lists.

@mutatrum the "follows sets" can accommodate private contact lists. Is this what you were looking for?

gzuuus:
> Sorry, @gzuuus, I didn't mean to tag you in a bad way, it was mostly a joke. I think everybody was using 30001 lists in a quite ad-hoc way and that we didn't know better -- but I hope we are agreeing on something.

No drama :) I like the proposal. ACK
However, in this effort to tidy up this nip I think it is worth mentioning some use cases that are not covered here and that are used in other clients, for example:
- Kind33888 and 33889 for boards and pins in pinstr
- Kind37777 for user cards in zap.stream

What should we do with these examples?

fiatjaf:
> However, in this effort to tidy up this nip I think it is worth mentioning some use cases that are not covered here and that are used in other clients, for example:
> 
> Kind33888 and 33889 for boards and pins in pinstr
> Kind37777 for user cards in zap.stream
> 
> What should we do with these examples?

I think these do not fit into NIP-51, or do they? The Pinstr thing is more like a "spreadsheet table" NIP, and the the other one I don't get, where can I see it?

fiatjaf:
What are Highlight lists @erskingardner? Was that a joke or do they exist?

gzuuus:
> > However, in this effort to tidy up this nip I think it is worth mentioning some use cases that are not covered here and that are used in other clients, for example:
> > Kind33888 and 33889 for boards and pins in pinstr
> > Kind37777 for user cards in zap.stream
> > What should we do with these examples?
> 
> I think these do not fit into NIP-51, or do they? The Pinstr thing is more like a "spreadsheet table" NIP, and the the other one I don't get

Yes maybe pinstr ones do not fit in nip51 due to its totally different convention to structure the data, but zap.stream ones are more like kind30023 since they are editable and use markdown for formatting.

> where can I see it?

https://github.com/v0l/zap.stream/blob/01eaf9996cfed9904906362c02597c65385102d1/src/const.ts#L9
https://njump.me/nevent1qqsgf58pds9la64vwcnxjwaat38f7dctk5x5gd0kquy0ug730fmw23g5ytm4f

v0l:
> * Kind37777 for user cards in zap.stream

The card itself is an event `37_777`, the content can be updated, and the list of cards on a profile is the `17_777`

fiatjaf:
Any last additions or objections before we can merge this?

vitorpamplona:
Can we get devs to declare their current compliance level to this new lists? 

2+ clients and we merge. 

I am running behind with Amethyst. :(

hzrd149:
@vitorpamplona noStrudel is using:
 - k:10000 for muting pubkeys
 - k:10030 for users favorite emoji lists
 - k:10004 and k:30001 "communities" for users communities (migrating to k:10004)

staab:
Coracle's status:

- 10000 for muting pubkeys and events
- 30003 for bookmarks (I'll keep read-only support for 30001 for a while)
- 10004 for communities coming soon (probably December)

Kind of hard to really meet the threshold of "2 clients" for this one because there are so many different list kinds being merged. But the PR has a record amount of client dev buy-in so it's probably fine.

51.md
Darecjo: > @Darecjo if your list is automatically named `MyFavoriteTopicsInYakihonne` I think that is an indication that you should be using NIP-78. What do you think?

We are going to check the interests list and sets since they are more suitable for our case it is basically the same thing.

51.md
Darecjo: > I'm using `title` and `description` tags on lists in Listr. I would love to leave those be, as they seem more applicable than `name`, `about`. I would also be in favor of using `image` instead of `picture`, keeping more in line with NIP-23 than kind 0.

I think it is better to keep the generic tags for `kind-30004` to `title`, `description`, and `image` or `thumbnail` since it is widely used same as @erskingardner suggested.

gzuuus:
I will implement on nostree k30003 asap, but will continue to support k30001 read-only for a period of time.

51.md
fiatjaf: changed to `title`, `image` and `description`.

51.md
```diff
@@ -6,148 +6,116 @@ Lists
...
+`"about"` tag that can be used to enhance their UI.
 
-```json
-["p", "3bf0c63fcb93463407af97a5e5ee64fa883d107ef9e558472c4eb9aaaefa459d"],
-["p", "32e1827635450ebb3c5a7d12c1f8e7b2b514439ac10a67eef3d9fd9c5c68e245"]
-```
-and privately include these users (below is the JSON that would be encrypted and placed in the event content):
+| name          | kind  | description                                                                                   | expected tag items                                |
+| ---           | ---   | ---                                                                                           | ---                                               |
+| Follow sets   | 30000 | categorized groups of users a client may choose to check out in different circumstances       | `"p"` (pubkeys)                                   |
```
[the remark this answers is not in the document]
fiatjaf: I agree, but that is the old name and we must change something since that wasn't working.

SnowCait:
How about adding NIP-65 kind:10002 relays to standard lists table as reference?

Public chats's kind:40 description seems to be wrong.

fiatjaf:
What is wrong with the description?

SnowCait:
Kind:40 is channel definitions, isn't it? (at [L27](https://github.com/nostr-protocol/nips/pull/880/files#diff-0d42225c4bd69ef69ef24d2de380ff90ca763019d04692558a1b78adeda25d50R27))

>`"a"` (kind:34550 community definitions)
>`"e"` (kind:40 community definitions) 

Seems to be copied from kind:34550 :)

fiatjaf:
Oh, that took me a while to spot.

KoalaSat (approved):

erskingardner:
> What are Highlight lists @erskingardner? Was that a joke or do they exist?

Sorry, didn't see this until now. They're not a joke, @pablof7z uses them in Highlighter I believe (unless he's changed that).

erskingardner:
> I will implement on nostree k30003 asap, but will continue to support k30001 read-only for a period of time.

Agreed. Going to implement these changes in Listr this week (including making 30001 lists read-only). I'll also be adding a merge tool to help people migrate faster/easier.

pablof7z:
Ah yeah, I forgot I added those; 39802 (highlight is 9802 thus this number)

PR #881, opened 2023-11-15
Documentation Enhancement: Link NIP-24 to Kind 0 Metadata Event & Reference "R" Tags

WilliamConnatser:
After reviewing NIP-1, I assumed it covered all specifications for Kind 0 Metadata events. However, analyzing metadata events across various relays revealed ambiguities in handling `name`, `display_name`, `displayName`, and `username`. Including NIP-24 in the main events list could clarify these distinctions and reduce confusion.

Regarding NIP-24's reference to deprecated fields for Kind 3 events, I recommend against linking it to these events to avoid misinterpretation.

Although NIP-24 isn't a complete spec for the reference "r" tag, it's useful for understanding the reference "r" tag and its usage in non-kind 10002 events. I've added a reference to NIP-24 and revised its wording for clarity, based on my understanding and a prior explanation of reference tags from @fiatjaf via another communication channel. I'm currently unaware of any other more specific usage of "r" tags other than kind 10002 events so this change to NIP-24 clarifies the distinctions.


fiatjaf:
I think this would be good but I can't really spot the changes because of many irrelevant markdown changes, I think they might be caused by your editor doing some kind of autoformatting?

WilliamConnatser:
My bad @fiatjaf - I removed the auto formatting so the changes are more easily discernible.

AsaiToshiya:
> Regarding NIP-24's reference to deprecated fields for Kind 3 events, I recommend against linking it to these events to avoid misinterpretation.

I agree. LGTM.

PR #882, opened 2023-11-16
NIP-19: add `nreq`

fiatjaf:
https://github.com/nostr-protocol/nips/blob/nreq/19.md

For discussion.

@v0l @pablof7z @staab

### DIFF
--- a/19.md
+++ b/19.md
@@ -4,7 +4,7 @@ NIP-19
 bech32-encoded entities
 -----------------------
 
-`draft` `optional` `author:jb55` `author:fiatjaf` `author:Semisol`
+`draft` `optional` `author:jb55` `author:fiatjaf` `author:Semisol` `author:kieran`
 
 This NIP standardizes bech32-formatted strings that can be used to display keys, ids and other information in clients. These formats are not meant to be used anywhere in the core protocol, they are only meant for displaying to users, copy-pasting, sharing, rendering QR codes and inputting data.
 
@@ -33,9 +33,10 @@ For these events, the contents are a binary-encoded list of `TLV` (type-length-v
 These are the possible bech32 prefixes with `TLV`:
 
   - `nprofile`: a nostr profile
-  - `nevent`: a nostr event
+  - `nevent`: a nostr event pointer
   - `nrelay`: a nostr relay
   - `naddr`: a nostr _replaceable event_ coordinate
+  - `nreq`: a nostr filter
 
 These possible standardized `TLV` types are indicated here:
 
@@ -46,14 +47,28 @@ These possible standardized `TLV` types are indicated here:
     - for `nrelay`, this is the relay URL
     - for `naddr`, it is the identifier (the `"d"` tag) of the event being referenced. For non-parameterized replaceable events, use an empty string.
 - `1`: `relay`
-  - for `nprofile`, `nevent` and `naddr`, _optionally_, a relay in which the entity (profile or event) is more likely to be found, encoded as ascii
+  - for `nprofile`, `nevent`, `naddr`, `nreq`, _optionally_, a relay in which the entity (profile or event) is more likely to be found, encoded as ascii
   - this may be included multiple times
 - `2`: `author`
   - for `naddr`, the 32 bytes of the pubkey of the event
   - for `nevent`, _optionally_, the 32 bytes of the pubkey of the event
+  - for `nreq`, _optionally_, 32 bytes of one entry in the `"authors"` field, can be specified multiple times
 - `3`: `kind`
   - for `naddr`, the 32-bit unsigned integer of the kind, big-endian
   - for `nevent`, _optionally_, the 32-bit unsigned integer of the kind, big-endian
+  - for `nreq`, _optionally_, 32-bit unsigned integer representing one entry in the `"kinds"` field, can be specified multiple times
+- `4`: `id`
+  - for `nreq`, _optionally_, 32 bytes of one entry in the `"authors"` field, can be specified multiple times
+- `5`: `tag`
+  - for `nreq`, _optionally_, the UTF-8 encoded single-letter of the tag to be searched (the `*` in a `"#*"` field) followed by one entry in this `#*` field, can be specified multiple times
+- `6`: `limit`
+  - for `nreq`, _optionally_, 32-bit unsigned integer representing the `"limit"` field
+- `7`: `since`
+  - for `nreq`, _optionally_, 32-bit unsigned integer representing the `"since"` field
+- `8`: `until`
+  - for `nreq`, _optionally_, 32-bit unsigned integer representing the `"until"` field
+- `9`: `search`
+  - for `nreq`, _optionally_, the UTF-8 encoded string of the `"search"` field
 
 ## Examples
 


staab:
seems like a good idea to me

@Semisol

staab:
Ok, so one thing @Semisol pointed out is that you can't reference a list of authors (i.e. follow list or whatever) in this. In coracle I have a `DynamicList` type where authors can be an array of pubkeys, or `follows | network`, and then I compile that based on whatever the user's current network looks like. So it makes custom feeds static, which seems highly limiting.

fiatjaf:
So you want unix pipes?

staab:
yes

staab:
Can we make a list that's turing complete?

staab:
`nlua`

fabianfabian:
I like the idea of nreq, but this one seems a bit limited? It could be powerful if it is just any filter+relays+title so you could share almost any kind of feed.

v0l:
As you know this isnt how i implemented it, i feel like putting TLV entries for search/limit/until into NIP-19 just doesn't make any sense because they dont apply to the other prefixes.

I simply put the JSON req filter array into type 0, then i just pass this value directly to query system, its very simple and flexible. 

Here is the example: https://main.snort-social.pages.dev/nreq1qp54k7ezdd5kuernyga9kv2a9s3xc6tdd96zyw33xqkzyct4w35x7unnyga9kg3nvfnrqcekxdnxxc3exv6rvve5xqmkze3exasn2ef4v4jnvdrxvyursvmyxycrwetx89jn2dfcxsmnyce5v43rjctpv9jkvcf5x5ukggja04wsz9rhwden5te0wfjkccte9ejxzmt4wvhxjmc09xpux 

A link to fiatjafs 10 latest posts: `[{"kinds":[1],"limit":10,"authors":["3bf0c63fcb93463407af97a5e5ee64fa883d107ef9e558472c4eb9aaaefa459d"]}]`

fabianfabian:
> As you know this isnt how i implemented it, i feel like putting TLV entries for search/limit/until into NIP-19 just doesn't make any sense because they dont apply to the other prefixes.
> 
> I simply put the JSON req filter array into type 0, then i just pass this value directly to query system, its very simple and flexible.
> 
> Here is the example: https://main.snort-social.pages.dev/nreq1qp54k7ezdd5kuernyga9kv2a9s3xc6tdd96zyw33xqkzyct4w35x7unnyga9kg3nvfnrqcekxdnxxc3exv6rvve5xqmkze3exasn2ef4v4jnvdrxvyursvmyxycrwetx89jn2dfcxsmnyce5v43rjctpv9jkvcf5x5ukggja04wsz9rhwden5te0wfjkccte9ejxzmt4wvhxjmc09xpux
> 
> A link to fiatjafs 10 latest posts: `[{"kinds":[1],"limit":10,"authors":["3bf0c63fcb93463407af97a5e5ee64fa883d107ef9e558472c4eb9aaaefa459d"]}]`

This looks good, I think the addition of a title or description field would improve it because the request can be anything. Right now you said "A link to fiatjafs 10 latest posts" but it could have been just part of the nreq and a client would just render a link titled "fiatjafs 10 latest posts" and after clicking it it would open the feed.

staab:
I like `title` and `filters` as the keys. It should definitely support multiple filters.

gzuuus:
> This looks good, I think the addition of a title or description field would improve it because the request can be anything. Right now you said "A link to fiatjafs 10 latest posts" but it could have been just part of the nreq and a client would just render a link titled "fiatjafs 10 latest posts" and after clicking it it would open the feed.

I think that description within the entity may be superfluous since the filter is self-explanatory, if you want a description it could be formed by parsing the filter and verbosing.

vitorpamplona:
I am not sure if this is feasible. Sure we can share filters (I like @v0l's idea of a single text field) but it's really hard for the Client to figure out how to display the results. 

What if it is a filter for chat messages? You would have to somehow parse the filter, see that this is returning only chat messages, and direct the user to that View type. Obviously, most filters will return a set of different event kinds. What happens if the filter is just anything by users X, Y and Z at the same time? It's not a profile view. Reactions are mixed with DMs, which are mixed with ephemeral NWC events, etc.. What do you do when you open this in your client?  

We don't have a super generic way to build a view for that.

staab:
> We don't have a super generic way to build a view for that.

Coracle's feeds can mostly handle this (other than DMs which is a degenerate case). If a client dev doesn't want to build generic feeds, they can also open the link in a client which does support it (maybe njump could have an nreq page).

vitorpamplona:
> Coracle's feeds can mostly handle this (other than DMs which is a degenerate case).

Sure, feed-based clients can just dump the whole thing in front of the user without any additional context. I don't know how to display a Zap Request or an Event Deletion but if it is in the filter it should be displayed somehow. 

This, of course, assumes this link is to be displayed. It could be used just to build a local storage of events. 

Either way, the client parsing the request doesn't know if this is a display thing or a storage thing. Or maybe something to send to relays just to confuse their user tracking tools. I don't know... It feels too broad.

staab:
> Is this a solution in search for a problem?

Maybe

> Maybe we should actually really come up with a `nscript` thing for querying.

lol

The use case I had imagined for this was custom feed lists, where each entry was a filter. But that could be done independent of `nreq`.

fiatjaf:
The `nscript` could be a WASM script like @jb55 was doing -- although I don't know how his thing works at all.

What I am thinking is that this would specify various steps in a query process and yield events in the process. Here's a stupid example that would use some constant values assumed to be defined and then perform a three step querying process to yield a custom feed of _recent posts from people of whom I have bookmarked content but I don't follow_.

```scala
val me;
val myOutboxRelays;
val myInboxRelays;
val myFollowList;

for tag in querySync({kinds: [10003], authors: [me]}, myOutboxRelays).tags:
  if (tag.0 == "e" or tag.0 == "a") and tag.2:
    val bookmark = querySync({ids: [tag.1]}, [tag.2])
    if bookmark.pubkey not in myFollowList:
      val events = querySync({kinds: [1], authors: [bookmark.pubkey]}, [tag.2])
      yield events
```

This is not a fleshed out idea in any sense, by the way, and may also be a solution in search for a problem just like `nreq` (in this case a much more complex solution).

Can you give a concrete example of how you would do custom feeds, @staab?

vitorpamplona:
Read-only query languages can be quite effective. 

For instance, I have a use case for contextual Amber alerts based on Nostr data. It would be a simple new Event kind with the message and a boolean filter. The client runs the filter in the local DB and if it passes, it displays the message. 

Then you could do alerts on things like: "If you have pneumonia and you are in New York, today's wildfire-affected air-quality is likely to delay your recovery rate by days. Make sure to use a good air-filter and stay indoors to avoid any side effects. Click here if you have questions or need additional recommendations".

staab:
> Can you give a concrete example of how you would do custom feeds, @staab?

Thinking through this more has been on my to-do list for a while. But off the top of my head, a custom feed would be an OR of multiple dynamic filters, where:

- `authors`/`#p` is a list of pubkeys, a list id, or the keywords "follows", "network"
- `kinds` is an optional list of kinds (but probably omitted and set to some reasonable default by the client)
- `relays` is optional list of urls or a list id of relays that replace the client's default routing behavior
- `topics` is an optional list of topics or a list id of topics
- `#l` is an optional list of labels or a list id of labels. Clients should probably interpret this as the targets of the labels, not the labels themselves
- `search` is an optional string
- Zero or more `dvm` tags that define a job pipeline to run the filter or results through. Really, all the special logic above could probably be handled by a dvm instead of the client.

Here's what Coracle currently does:

Screenshot 2023-11-16 at 8 33 37 AM

vitorpamplona:
Neil has been pushing for Custom feeds as well: 

https://image.nostr.build/b495a4fc9021551f6b8e69ab4b8739e55abf4facf8676e7412ea678aebbffa3c.png

fiatjaf:
It's looking like `nreq`s directly won't be as powerful as anyone really wants, and having ad-hoc special cases added on top of the `nreq` would break it formidably, such that the `nscript` approach is looking better.

In the simplest case the `nscript` could be just a request filter that is sent to a relay and the results are returned, so exactly like the simplest `nreq` case.

What is your approach for taking input into the nscript, @jb55? Can a user be prompted to select some options -- for example, a user loads a script and the script wants the user to pick from one of the users' relay sets or something like that before continuing?

v0l:
Personally i don't have any use case for `nreq` (right now) and would much rather see `nscript` be standardized, an `nscript` to load a specific DVM requested algorithmic feed would be very cool

fiatjaf:
@Semisol I think they definitely fit in a NIP, but yes, I would like to see what @jb55 has done how we can proceed to get a working standard from there too.

fiatjaf:
I'll close this for now since we all agree `nreq`s are not very useful for now.

AsaiToshiya:
I feel like `nreq` is optimal for dynamic references like wikilinks.
https://asaitoshiya.github.io/nostr-toybox/wiki-article/dist/#nreq1qvzxy7qqqqz3qertdaex2ctw94kxzmn8w4skwegkhym52

PR #884, opened 2023-11-16
Add client tag to nip 89

staab:
As a result of this discussion: nostr:nevent1qqstw8cw0nw50le9uugahpze78q9y0j90swhgan636ac08acmkn6v0gpz3mhxue69uhhyetvv9ukzcnvv5hx7un89uqs6amnwvaz7tmzwf3zu6t09uq3wamnwvaz7tmjv4kxz7fwwvehstnnda3kjctv9uqsuamnwvaz7tmwdaejumr0dshsz9mhwden5te0wfjkccte9ehx7um5wghxyctwvshsz9mhwden5te0dehhxarj9eex2mrp09jhytnnv5hszyrhwden5te0v5hxummn9ekx7mp0qyfhwumn8ghj7ur4wfcxcetsv9njuetn9uq3xamnwvaz7tm0venxx6rpd9hzuur4vghszxthwden5te0wfjkccte9eekummjwsh8xmmrd9skctcku97e5

### DIFF
--- a/89.md
+++ b/89.md
@@ -74,6 +74,19 @@ Multiple tags might be registered by the app, following NIP-19 nomenclature as t
 
 A tag without a second value in the array SHOULD be considered a generic handler for any NIP-19 entity that is not handled by a different tag.
 
+# Client tag
+When publishing events, clients MAY include a `client` tag in the same format as the recommendation event's `a` tags. This has privacy implications for users, so clients SHOULD allow users to opt-out of using this tag.
+
+```json
+{
+  "kind": 1,
+  "tags": [
+    ["client", "31990:app1-pubkey:<d-identifier>", "wss://relay1", "ios"]
+  ]
+  ...
+}
+```
+
 # User flow
 A user A who uses a non-`kind:1`-centric nostr app could choose to announce/recommend a certain kind-handler application.
 


vitorpamplona:
It's nice to see people referencing other Nostr events instead of using DNS addresses or hard-coded codes everywhere.

fiatjaf:
This will make NIP-89 great again.

pablof7z (approved):
Very cool idea

fiatjaf (approved):

staab:
We should slow down on merging new stuff, the rule is 2 clients, I don't think anyone does this yet. I was thinking about this recently actually, it would be nice to have a "pending merge" issue or something where people can subscribe and get notified just before something goes through so they have one last chance before NIPs go in.

vitorpamplona:
> We should slow down on merging new stuff, the rule is 2 clients, I don't think anyone does this yet. I was thinking about this recently actually, it would be nice to have a "pending merge" issue or something where people can subscribe and get notified just before something goes through so they have one last chance before NIPs go in.

Agree. The 2+ clients requirement is a good thing.

fiatjaf:
I don't know, I think if we have a bunch of people agreeing and with intention to implement we can probably merge, especially if it's such a simple and straightforward thing. That makes everybody in a more comfortable spot for implementing without risking having to rewrite later.

For more complex NIPs it is probably wise to wait for implementations as the implementations may shed more light into the design.

But in all cases we can always edit the NIPs if after a couple of implementations we find out that the design was bad.

vitorpamplona:
I know it looks straightforward, but this is the first `a-Tag` that is not named `a`. To be it is a big deal. 

We need a name for the `a`-tag format: `kind:pubkey:dtag` so we can use in other tag names.

staab:
> We need a name for the a-tag format: kind:pubkey:dtag so we can use in other tag names.

It's called an `address` isn't it? Short for "replaceable event address".

vitorpamplona:
Isn't it too close to the `naddr` format? We don't want people writing naddr in tags by mistake.

staab:
> Isn't it too close to the naddr format? We don't want people writing naddr in tags by mistake.

I definitely made that mistake when I was greener, fixed one of those today in fact. But a naddr encodes an address, so I think it does make sense.

fabianfabian:
I have been doing `["client","Nostur"]` because I saw in my local cache:
````
["client","coracle"]
["client","agora"]
["client","gossip"]
["client","stemstr.app"]
["client","zapthreads"]
["client","more-speech - 2023-07-13T12:37"]
["client","yakihonne.com"]
["client","blogstack.io"]
````

and was planning to change it to ```["client", "Nostur", "31990:9be0be0fc079548233231614e4e1efc9f28b0db398011efeecf05fe570e5dd33:1685868693432"]``` in the next update to do it in the least breaking way.

This way we can have a simple implementation of showing a simple "posted via ..." for clients that don't do full NIP-89, while also allowing those who do to show rich information.

fiatjaf:
@fabianfabian's idea is much better, let's change the NIP to do it.

pablof7z:
Yeah, this is a good idea

vitorpamplona:
> @fabianfabian's idea is much better, let's change the NIP to do it. @pablof7z @staab @vitorpamplona

Sure. I believe it is also backwards compatible to what some clients do today (app name as the second value), right?

pablof7z:
> > @fabianfabian's idea is much better, let's change the NIP to do it. @pablof7z @staab @vitorpamplona
> 
> 
> 
> Sure. I believe it is also backwards compatible to what some clients do today (app name as the second value), right? 
> 

Yup. I'm adding this version to NDK now.

staab:
Good idea @fabianfabian, added it to https://github.com/nostr-protocol/nips/pull/900

Issue #769, opened 2023-09-08
About visivility of NIP-36 (content-warning) specified content


ryogrid:
# Problem
- Users specify content-waring to own post for making not visible it to people who don't approbed that it is sensitive content
- But current NIP-36 specification can't realize above because not NIP-36 compliant clients such as Damus shows all kind-1 event's content without noticing the content-warning marking
- Lack of consideration described above left users with a psychological barrier that makes it difficult to post sensitive content

# Solution
- Modify NIP-36 specification to use kind number for content-warning specified event instead of 1
- Or, define NIP-36 extended new NIP like above

AsaiToshiya:
I agree with need for this proposal. Probably we need to think about some kinds not only kind 1.

AsaiToshiya:
What about the solution to wrap the event?

For example:

```json
{
    "pubkey": "<pub-key>",
    "created_at": 1000000000,
    "kind": 30036,
    "tags": [
      ["d", "<id of the wrapped event>"],
      ["k", "1"],
      ["L", "content-warning"],
      ["l", "reason", "content-warning"],
      ["L", "social.nos.ontology"],
      ["l", "NS-nud", "social.nos.ontology"],
      ["content-warning", "reason"] /* reason is optional */
    ],
    "content": "<stringified event, e.g. kind 1>",
    "id": "<event-id>"
}
```

ryogrid:
@AsaiToshiya 
Thanks for comment and suggestion :)

> What about the solution to wrap the event?

Your idea looks nice.
I understund this as technique like NIP of repost (NIP-18) .

I have a question.
- Your suggested idea assumes that wrapped event is not sent to relay in not wrapped form?
  - (If answer of above is Yes, I think that suggested idea can solve the iessue)

AsaiToshiya:
> Your idea looks nice. I understund this as technique like NIP of repost (NIP-18) .

Ah yes right.

> I have a question.
> 
> * Your suggested idea assumes that wrapped event is not sent to relay in not wrapped form?
>   
>   * (If answer of above is Yes, I think that suggested idea can solve the iessue)

Yes, the wrapped event itself is not to be sent to relays.

But, my solution have a problem that cannot filter by d tag value of the wrapped event.

SnowCait:
How about copying tags which are expected to be filtered (= `#<single-letter>`) to kind 30036?

AsaiToshiya:
Yes, it is the only way.

[...]

PR #885, opened 2023-11-17
Addition to NIP-36: Specifying kind for Content with content-warning

imksoo:
In the context of NIP-36 regarding the content-warning tag, it is advised that when posting content that requires a content warning, the kind attribute of the event should be set to 18 instead of 1. This adjustment ensures that the nature of the content is appropriately categorized and handled by clients adhering to the Nostr protocol specifications. The use of kind 18 specifically designates the content as requiring a warning, which aligns with the intent of NIP-36 to offer users control over the visibility of sensitive content.

### DIFF
--- a/36.md
+++ b/36.md
@@ -9,6 +9,11 @@ Sensitive Content / Content Warning
 The `content-warning` tag enables users to specify if the event's content needs to be approved by readers to be shown.
 Clients can hide the content until the user acts on it.
 
+For content utilizing the `content-warning` tag, it is recommended to use `kind 18` to appropriately categorize such events.
+This recommendation is grounded in the fact that using the usual `kind 1` could lead to inappropriate display of sensitive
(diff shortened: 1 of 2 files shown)


AsaiToshiya:
Please read the related issue: #769

imksoo:
I'd like to clarify my stance on the association of my Pull Request with #769. Here are my main points:

1. #769 is Ideational, Not Solution-Oriented: The issue primarily discusses a concept rather than providing a tangible solution.

2. Distinction in Kind Numbers: The proposed kind 30036 in Issue #769 is meant for Parameterized Replaceable Events, which is different in nature from kind 1. This distinction is essential for understanding their respective usages.

3. Advocating for Kind 18 for Simplicity: Introducing kind 18 offers a simpler and more understandable categorization for content that should not be shown by default in non-compliant clients, while still being akin to kind 1 content. This approach would help maintain clarity and ease of implementation within the Nostr protocol.

PR #889, opened 2023-11-18
NIP-11: add `"restricted_writes"` limitation

fiatjaf:
Because relays may have other kinds of arbitrary restrictions that are not `payment_required`.


v0l (approved):

fiatjaf:
Implemented this in https://github.com/nbd-wtf/go-nostr and in https://github.com/github-tijlxyz/khatru-invite.

v0l (approved):
LGTM

PR #893, opened 2023-11-19
Nip 211 -  info triple

unostr:
Here's a NIP for the Info Triple 

The NIP can be read [here](https://github.com/unostr/nips/blob/nip-211---info-triple/211.md)


vitorpamplona:
How do we know what the triplet is for? Or what is it trying to say?

unostr:
Well, in one of three ways. If our app doesn't know them to begin with, we'll have to look them up and see if there are known triples describing it, or if a descriptor note has been made with some human readable info.

And thank you for the questions it made me realize I haven't mentioned any details on lookup services.

chrisacheson:
> How do we know what the triplet is for? Or what is it trying to say?

It sounds like some of the identifiers will be documented with the "descriptor notes" in #892. The meaning of the other identifiers would be inferred through the structure of the triples. For example, you might have these identifiers documented:

- 0xfe7: An image
- 0x246: A person
- 0x97d: An apple
- 0xa8c: Action, eating
- 0x985: Action, depicting

If you had an image of a person eating an apple and you wanted to tag it so that others could find it, you might just use simple hashtags, like "#person #eating #apple". However, that wouldn't differentiate it from say, a short story about having a picnic under an apple tree.

Using this system, you could tag the image in a much more detailed fashion like this:

- 0x32a 0x246 0xa8c (a person <-> action, eating)
- 0xae9 0x32a 0x97d (a person eating <-> an apple)
- 0xd82 0xfe7 0x985 (an image <-> action, depicting)
- 0xe5d 0xd82 0xae9 (an image depicting <-> a person eating an apple)
- 0x3a8 0xe5d [sha256 of the actual image] (this specific image depicts a person eating an apple)

I find the bi-directionality of it kind of confusing, but it seems workable.

vitorpamplona:
It would be nice to have some information on how to decode these things. The idea for Nostr is to build interoperable systems. So, any app that uses the Triple system will need to interoperate with any other app that also uses it. 

They not only will need to define boundaries for each of these codes, so that they don't interfere with one another, but they most likely will need to agree on a common coding structure for ALL users of these triplet design.

chrisacheson:
@unostr How would these be queried from a relay? Are you thinking of having specialized relays to store and index triples? As far as I know, current relays don't index content, just single-letter tags. Seems like the indices for these could get pretty large, too.

As an aside, I find this kind of semantic web/knowledge engineering stuff fascinating, but also kind of impenetrable. Do you have any good beginner to intermediate resources for learning more on the topic? NIP-32 and NIP-36 are capable of using ontologies as namespaces, but I have no idea how to start looking for specific vocabularies to use.

unostr:
> > How do we know what the triplet is for? Or what is it trying to say?
> 
> It sounds like some of the identifiers will be documented with the "descriptor notes" in #892. The meaning of the other identifiers would be inferred through the structure of the triples. For example, you might have these identifiers documented:
> 
>     * 0xfe7: An image
> 
>     * 0x246: A person
> 
>     * 0x97d: An apple
> 
>     * 0xa8c: Action, eating
> 
>     * 0x985: Action, depicting
> 
> 
> If you had an image of a person eating an apple and you wanted to tag it so that others could find it, you might just use simple hashtags, like "#person #eating #apple". However, that wouldn't differentiate it from say, a short story about having a picnic under an apple tree.
> 
> Using this system, you could tag the image in a much more detailed fashion like this:
> 
>     * 0x32a 0x246 0xa8c (a person <-> action, eating)
> 
>     * 0xae9 0x32a 0x97d (a person eating <-> an apple)
> 
>     * 0xd82 0xfe7 0x985 (an image <-> action, depicting)
> 
>     * 0xe5d 0xd82 0xae9 (an image depicting <-> a person eating an apple)
> 
>     * 0x3a8 0xe5d [sha256 of the actual image] (this specific image depicts a person eating an apple)
> 
> 
> I find the bi-directionality of it kind of confusing, but it seems workable.

Yes your example looks alright. To get over the bi-directionality you'll need to tag more elements.
Like this:

  0x135 description of a picture
  0x57a hash of a picture file
  0x2d1 description and hash of picture
  
  0x4b3 0x3a8 0x2d1 (your description and hash <-> a description and a hash of a picture tagged on protocol level)
  0x94f [sha256 of the actual image] 0x57a  (this specific hash <-> hash of a picture file)
  0x550 0xe5d 0x135 (your description <-> description of a picture)
 
  With that our app can analyze 0x4b3 and figure out both picture and description.

unostr:
> It would be nice to have some information on how to decode these things. The idea for Nostr is to build interoperable systems. So, any app that uses the Triple system will need to interoperate with any other app that also uses it.
> 
> They not only will need to define boundaries for each of these codes, so that they don't interfere with one another, but they most likely will need to agree on a common coding structure for ALL users of these triplet design.

The encoding could be expressed in as NIP-32 tags, but a pre-decided mandatory encoding format would be better.

It's a trivial process to change encoding in that you just redo all IDs in another encoding format, however there may be a non-trivial amount of IDs to re-encode, so yes you are right about that.

unostr:
> @unostr How would these be queried from a relay? Are you thinking of having specialized relays to store and index triples? As far as I know, current relays don't index content, just single-letter tags. Seems like the indices for these could get pretty large, too.
> 
> As an aside, I find this kind of semantic web/knowledge engineering stuff fascinating, but also kind of impenetrable. Do you have any good beginner to intermediate resources for learning more on the topic? NIP-32 and NIP-36 are capable of using ontologies as namespaces, but I have no idea how to start looking for specific vocabularies to use.

yes definitely use indexes.

unostr:
Updated the specs. This Nip is in an ongoing draft-mode which is fine as it is to be perceived as such. Reference implementation is on its way. Using Nostr tags now - so have coming around on that one...

PR #894, opened 2023-11-20
NIP-46: Add target POW difficulty when signing events

alexgleason:
When using Nostr Connect, it makes the most sense to negotiate proof-of-work requirements for signed events within the NIP-46 request itself.

This MR adds a `options` object to the `sign_event` method. So far it supports only one option, `pow` which is the target proof-of-work difficulty for the signed event.

(Marked as a draft because I have not actually implemented this yet.)

### DIFF
--- a/46.md
+++ b/46.md
@@ -66,7 +66,9 @@ These are mandatory methods the remote signer app MUST implement:
   - params []
   - result `pubkey` 
 - **sign_event**
-  - params [`event`]
+  - params [`event`, `options`]
+  - options
+    - `pow` minimum proof-of-work difficulty
   - result `event_with_signature` 
 
 #### optional
--- a/46.md
+++ b/46.md
@@ -67,8 +67,8 @@ These are mandatory methods the remote signer app MUST implement:
   - result `pubkey` 
 - **sign_event**
   - params [`event`, `options`]
-  - options
-    - `pow` minimum proof-of-work difficulty
+    - `options` (object, optional)
+      - `pow` target proof-of-work difficulty for the signed event
   - result `event_with_signature` 
 
 #### optional


fiatjaf:
You can generate the POW locally and just submit the event to be signed afterwards.

alexgleason:
Yes, this is so the client can know the target POW of the server before mining. NIP-46 is for dealing with a traditional webserver, and that server may impose a POW requirement to prevent spam.

alexgleason:
I have now implemented this between [soapbox!2864](https://gitlab.com/soapbox-pub/soapbox/-/merge_requests/2864) and [ditto!58](https://gitlab.com/soapbox-pub/ditto/-/merge_requests/58), and it works.

I believe it's necessary to make NIP-13 work with NIP-46, in the circumstance where the server cares.

monlovesmango:
> Yes, this is so the client can know the target POW of the server before mining.

can you expand on this? why doesn't the client talk to the server directly? why does nostrConnect device need to be involved?

alexgleason:
@monlovesmango The server can adjust the pow difficulty depending on the event itself and history of the user. Eg to register an account, you need a high pow, but to make a post you need less. Additionally, if you set off triggers that you're a spammer, we can gradually increase the base pow required for all actions.

monlovesmango:
why can't client handle those changes in POW? why does this need to involve the nostrConnect device?

alexgleason:
Because the client needs to know what difficulty the server will accept. Feel like we're talking in circles.

monlovesmango:
same. wait are you saying the nostrConnect device would actually be a server? I was assuming it is a personal device.

alexgleason:
There is an app and a signer. The signer is a client, and the app is a server/relay. That's the whole point of Nostr Connect.

monlovesmango:
maybe we are using different terms but this is not at all how I understood nostrConnect.

there is an app, a signer, and a relay. the signer is a personal device, the app is a client, and the relay is a server. the app talks to the relay and the relay dictates what level of POW is needed. the app also talks to the signer, via a relay (either the same relay or separate relay) in order to get events signed. the app will execute the required level of POW before sending an event to the signer to be signed. no where in there does the signer need to be aware of the relay's required level of POW.

the point of Nostr Connect is to behave as an alternative to NIP 7 browser extension signing.

this is my understanding. maybe I could understand your use case better if you can tell me where you disagree with me?

alexgleason:
> the app talks to the relay and the relay dictates what level of POW is needed. 

How does it do that?

monlovesmango:
app (aka client) talks to relay through websockets. how the relay communicates the level of POW needed can either be NIP 11 or some other out of band method that is not defined in a NIP.

alexgleason:
Re-reading this, I see the disconnect. My signer is not a backend server like nsecBunker. My situation is reversed. The signer is a web client (or other dedicated signer app), and the "app" is a server. The server cannot afford to mine pow, so it needs to ask the signer to do it.

NIP-46 is about remote signing. It defines a signer and an app, but it does not prescribe a specific architecture. There is more than one way to NIP-46.

monlovesmango:
ok I understand what you are trying to solve now. however, even if the web client or app cannot afford to mine POW on their server, its pretty trivial for the client/app to perform that POW on the user device instead right? this seems simpler than adding POW variable to `sign_event` method, and either way it will push the POW compute to the user's device.

but maybe there are additional scenarios I am not taking into account that won't be able to perform the POW on the user's device instead?

alexgleason:
Yes. But this signer still needs to know the POW amount that the server would accept, which is why it still needs to be passed in.

BTW I think my perception of NIP-46 aligns with how it was originally written. The signer would be like a dedicated Android app, not a cloud server like nsecbunker.

Also, I care less whether this specific PR actually makes it to the NIPs, and more that somebody doesn't add a new positional argument to sign_event later that breaks compatibility with my options object. For sure after the first few truly _required_ params, an options object should be used.

monlovesmango:
I have the same perception of NIP46, still trying to wrap my brain around the nsecbunker structure tbh.

However I am still confused by this:
>But this signer still needs to know the POW amount that the server would accept, which is why it still needs to be passed in.

while I agree the signer needs to know the POW amount in order to sign the event, I don't agree that the signer itself needs to be responsible for the mining. the event template sent in `signEvent` should already have the proper POW tag attached. let me lay out how I would anticipate this to work, and you can tell me where I am incorrect or misunderstanding.

1. web client/app wants to create/publish a nostr event with a certain amount of proof of work
2. web client/app creates event template and mines the event using the user's device resources
3. once event is mined, web client/app takes the unsigned event (with POW tag) and sends it in a `signEvent` request to the signer
4. signer sends back signed event (which has POW)

this is the flow that was used for NIP07 and I would expect NIP46 to have similar flow.

mikedilger:
I think we should not assume that the event generator has the power to mine.  It might be that the signer has much more power.  In which case the signer could do the PoW mining.  I which case the option is reasonable.

Or we could just be happy with DVMs.  Have the event generator to first use a DVM to get the PoW done, and then use the signer to sign.

monlovesmango:
by DVM do you mean Data Vending Machine? I think that could work.

I think my main gripe with having signers do POW is that it makes it either more edge cases that implementers of POW apps have to handle (to account for different ways to execute the POW) or if the app only implements POW via signer then it limits which signers an end user can use with the app.

and overall I don't really see much of a pro with adding ability to do POW via signer. either way, the POW will have to be performed by the user (either on their device or via paid service) its just a question of orchestration. 

>I think we should not assume that the event generator has the power to mine.
either we assume this, or we have to assume the signer has the power to mine. both are equally questionable imo. one questionable assumption is easier to account for than two, and the assumption that signer has power to mine is harder to handle of the two.

alexgleason:
Here is the one thing I want people to take away from this thread most: why does **NIP-47 accept an options object with named keys**, while **NIP-46 accepts an array of parameters by index**? 

The way NIP-47 does it is infinitely better, and it's my biggest gripe with NIP-46. Also a point of contention in #976 

I don't care whether it's a good idea to make the signer do POW. I need it for my use-case, in _some_ circumstances where it doesn't matter if doesn't play nicely with other implementations. You can ignore the `pow` property, *I* just want to add it. But because of the precious way arrays are used in NIP-46, I am forced to claim a stake on having an options object in the second index. It's a bad design, and I'm very glad @Semisol went for an options object in NIP-47. But I feel like it should not be an expectation and not a privilege to expect good design from a NIP.

alexgleason:
I actually think the data format of NIP-46 is still so bad, it's worth changing and completely breaking all existing implementations because of how bad it is.

You know the other thing NIP-47 gets right? It has a DIFFERENT KIND for request and response! NIP-46 uses the same kind for both, which can be _worked around_, but is not _good_.

NIP-47 uses an options object with named parameters. NIP-46 uses precious indexes of an array. This is a lot worse. But it's not even the only problem!

These design choices matter because they continue to impact developers far into the future. It's still soon enough to change it.

monlovesmango:
agree with everything you said. I am 100% down to redo NIP-46 to align with NIP47. I find it really terrible how differently they are spec'ed and prefer NIP-46 structure as well.

monlovesmango:
maybe we can slowly get buy in by opening a PR that has new recommended NIP-46 format and moves current format to legacy support? happy to open a PR for this if others agree.

signers will have to support both for a bit, but doesn't seem too big of a hassle if everyone agrees its better for long term support.

alexgleason:
I was wrong about this.

PR #897, opened 2023-11-22
delete NIP-22

fiatjaf:
because it is useless and no one cares about it.


vitorpamplona:
I didn't even know those things existed.

AsaiToshiya:
https://github.com/nostr-protocol/nips/blob/7822a8b12670312aff104ddc03066425882f739d/11.md?plain=1#L134-L136

These may also need to be delete.

weex:
I'm not sure if we want to encourage people to bring old content to nostr but created_at would help decide between relays if you were looking to [archive old social media posts](https://blog.lopp.net/how-to-exit-x-to-social-network-sovereignty/) and didn't want to have run your own relay.

fiatjaf:
There are already fields on NIP-11 describing what relay accept or not -- although I think this could and should be decided mostly manually anyway.

pablof7z (approved):

AsaiToshiya:
Is NIP-22 a reusable number? Won't it be a problem like NIP-15 in the future?

PR #899, opened 2023-11-24
Add nfile

staab:
This adds a new entity that points to a static file hosted at one or more urls.

### DIFF
--- a/19.md
+++ b/19.md
@@ -36,6 +36,7 @@ These are the possible bech32 prefixes with `TLV`:
   - `nevent`: a nostr event
   - `nrelay`: a nostr relay
   - `naddr`: a nostr _replaceable event_ coordinate
+  - `nfile`: a file
 
 These possible standardized `TLV` types are indicated here:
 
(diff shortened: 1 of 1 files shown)


19.md
```diff
@@ -45,8 +46,10 @@ These possible standardized `TLV` types are indicated here:
     - for `nevent` it will be the 32 bytes of the event id
     - for `nrelay`, this is the relay URL
     - for `naddr`, it is the identifier (the `"d"` tag) of the event being referenced. For non-parameterized replaceable events, use an empty string.
+    - for `nfile`, it is the "original hash" (`ox` from NIP 95) of the file being referenced.
```
arthurfranca, line 49: from NIP 94

arthurfranca:
NIP-96 has `<server-url>` that leads to a config file at `<server-url>/.well-known/nostr/nip96.json`.
The config has the `<download-url>` that is equal to `config.download_url || config.api_url`.
Then to download the file one can do `<download-url>/<ox>`.

**So which url of these two below do you think should be inside the `nfile`?**
1) `<server-url>`: useful to get other routes from the config like the upload one and to kind of advertise NIP-96 compatible servers existence. Although one could rely just on https://github.com/nostr-protocol/nips/blob/d67c4c202a0ed914db27105a1f0967f63ef0750e/96.md?plain=1#L304 to discover new servers
2) `<download-url>`: is faster because you just need to append `ox` to this. downside is you can't rely on it to extract the `<server-url>`

For the NIP-94 `ox` tag I picked the `<server-url>` but could change it to the `<download-url>`. I'm unsure

staab:
The way NIPs 94/96 are written, it's hard to point to a "host" without coupling to it. What I'm proposing is that `urls` are fully-resolved download urls. `ox` would be used to search NIP 94 events, which would in turn point to the hosting provider. However this is not currently possible because `ox` is not a single-letter tag — why was this not supported?

staab:
> If you already have a bunch of fully-resolved download urls there isn't much need for the ox hash.

This would allow you to still dereference the hash and find the image if all alt urls in the nfile are dead, and the image has been replicated to another host.

I am starting to feel like nips 94/95/96 need to be combined into a single NIP that operates based on the same mental model. I feel like there are many conflicting visions here.

PR #901, opened 2023-11-27
NIP-88: Pay request spec for relays

vitorpamplona:
Paid relays often run into friction when renewing subscriptions or when upselling users to the next plan they have available. This PR allows relays to request payments on an as-needed basis. 

This has been implemented on inbox.nostr.wine and Amethyst.

### DIFF
--- a/88.md
+++ b/88.md
@@ -1,4 +1,4 @@
-NIP-XX
+NIP-88
 ======
 
 PAY Request
--- a/88.md
+++ b/88.md
@@ -6,21 +6,21 @@ PAY Request
 
 `draft` `optional`
 
-At any point, the relay can send a `PAY` request to a client with a BOLT 11 invoice, a description of what the user is purchasing by paying the invoice and a url for other options or more information. The Lightning invoice is the id of the request. 
+At any point, the relay can send a `PAY` request to a client with a BOLT 11 invoice, a description of what the user is purchasing by paying the invoice, and a URL for other options or more information. The Lightning invoice is the ID of the request.
 
 ```js
 [
   "PAY",
   "<BOLT 11 Invoice>",
-  "<Description>"	
+  "<Description>"
   "<Url for other plan options>"
 ]
 ```
 
-Supporting clients SHOULD display a popup or notification to describe the action needed and collect immediate response from the user: `pay` and keep using the relay, `dismiss` and stop using the relay or `more information` by navigating the user to the url. Clients SHOULD expect that the same request might be sent multiple times, if the user dismisses or pays the amount, the client SHOULD ignore following requests with the same invoice. 
+Supporting clients SHOULD display a popup or notification to describe the action needed and collect immediate response from the user: `pay` and keep using the relay, `dismiss` and stop using the relay, or `more information` by navigating the user to the URL. Clients SHOULD expect that the same request might be sent multiple times, if the user dismisses or pays the amount, the client SHOULD ignore following requests with the same invoice.
 
-Supporting relays SHOULD make sure the lightning invoice is the same for similar payment requests and minimize the number of times this request is sent to avoid annoying users with poorly-implemented clients. 
+Supporting relays SHOULD make sure the lightning invoice is the same for similar payment requests and minimize the number of times this request is sent to avoid annoying users with multiple popups.
 
 ### Motivation
 
-Paid relays often run into friction when renewing subscriptions or when upselling users to the next plan they have available. The API described here allows clie
+Paid relays often run into friction when renewing subscriptions or when upselling users to the next plan they have 
--- a/88.md
+++ b/88.md
@@ -1,0 +1,26 @@
+NIP-XX
+======
+
+PAY Request
+-----------
+
+`draft` `optional`
+
+At any point, the relay can send a `PAY` request to a client with a BOLT 11 invoice, a description of what the user is purchasing by paying the invoice and a url for other options or more information. The Lightning invoice is the id of the request. 
+
+```js
+[
+  "PAY",
+  "<BOLT 11 Invoice>",
+  "<Description>"	
+  "<Url for other plan options>"
+]
+```
+
+Supporting clients SHOULD display a popup or notification to describe the action needed and collect immediate response from the user: `pay` and keep using the relay, `dismiss` and stop using the relay or `more information` by navigating the user to the url. Clients SHOULD expect that the same request might be sent multiple times, if the user dismisses or pays the amount, the client SHOULD ignore following requests with the same invoice. 
+
+Supporting relays SHOULD make sure the lightning invoice is the same for similar payment requests and minimize the number of times this request is sent to avoid annoying users with poorly-implemented clients. 
+
+### Motivation
+
+Paid relays often run into friction when renewing subscriptions or when upselling users to the next plan they have available. The API described here allows clie
--- a/88.md
+++ b/88.md
@@ -23,4 +23,4 @@ Supporting relays SHOULD make sure the lightning invoice is the same for similar
 
 ### Motivation
 
-Paid relays often run into friction when renewing subscriptions or when upselling users to the next plan they have 
+Paid relays often run into friction when renewing subscriptions or when upselling users to the next plan they have available. The API described here allows clients to ease the subscription payment process on an as-needed basis. 


fiatjaf:
I don't like this idea, I think it is too short-sighted. Ties Nostr too much into Bitcoin and Lightning. It's cool for its limited use case, but I don't see it gaining much adoption, even if I liked it.

Another concern I have is: will you throw this popup up for every relay that sends a `PAY` message? Or is it just for relays that you have added to your "inbox" list? Sounds too limited at first glance.

I feel like a `NOTICE` with a clickable link that leads the user into a webpage for paying (or even a clickable `lightning:...` invoice inside the NOTICE text) would work better and be simpler and more flexible and easier for everybody.

Or, if you want a very custom experience, do a normal `REQ` with the relay's own pubkey only for the specified relays you know you're paying and whenever you get an event from that relay pubkey you display it in a custom way?

vitorpamplona:
> Another concern I have is: will you throw this popup up for every relay that sends a PAY message?

Every **different** PAY message, yes. This works for all types of relays, even public ones that just want donations. Any payment request any relay wants to send will be displayed. If they annoy the user, it's their customer to loose. 

> Or, if you want a very custom experience, do a normal REQ with the relay's own pubkey only for the specified relays you know you're paying and whenever you get an event from that relay pubkey you display it in a custom way?

We don't know what types of access relays will be requesting payments for. So, it's better for them to send us a note when we ask for something that is behind a paywall. 

> I feel like a NOTICE with a clickable link that leads the user into a webpage for paying (or even a clickable lightning:... invoice inside the NOTICE text) would work better and be simpler and more flexible and easier for everybody.

We could use NOTICE as well, but we will need to know which ones are DEBUG messages and which NOTICES are to be displayed to the user immediately. Today, the only use of NOTICE is for debugging. A popup for each NOTICE is too much.

88.md
```diff
@@ -0,0 +1,26 @@
...
+  "PAY",
+  "<BOLT 11 Invoice>",
+  "<Description>"
+  "<Url for other plan options>"
+]
+```
+
+Supporting clients SHOULD display a popup or notification to describe the action needed and collect immediate response from the user: `pay` and keep using the relay, `dismiss` and stop using the relay, or `more information` by navigating the user to the URL. Clients SHOULD expect that the same request might be sent multiple times, if the user dismisses or pays the amount, the client SHOULD ignore following requests with the same invoice.
+
+Supporting relays SHOULD make sure the lightning invoice is the same for similar payment requests and minimize the number of times this request is sent to avoid annoying users with multiple popups.
```
KoalaSat, line 22: Maybe an additional suggestion would be to set the invoice with a high expiry date.

88.md
```diff
@@ -0,0 +1,26 @@
...
+```js
+[
+  "PAY",
+  "<BOLT 11 Invoice>",
+  "<Description>"
+  "<Url for other plan options>"
+]
+```
+
+Supporting clients SHOULD display a popup or notification to describe the action needed and collect immediate response from the user: `pay` and keep using the relay, `dismiss` and stop using the relay, or `more information` by navigating the user to the URL. Clients SHOULD expect that the same request might be sent multiple times, if the user dismisses or pays the amount, the client SHOULD ignore following requests with the same invoice.
```
staab, line 20: I would leave out the first sentence here. Clients SHOULD handle this in a way that makes sense. A pop-up might not be it.

staab (changes requested):
I've been wanting to see this for a long time, but now that it comes to it I think I agree with fiatjaf that we should start with a URL rather than an invoice. That way we're not tying ourselves to lightning, but we can also handle other use cases like subscriptions, terms of service, and anything else a relay cares to pack into the page its url points to. The key pain point this solves is flagging within clients that a rely requires admission, which doesn't work at all currently, not facilitating payment, since invoices have to be relayed to wallet software anyway.

vitorpamplona:
> I think I agree with fiatjaf that we should start with a URL rather than an invoice (although NOTICE seems like a bad idea). That way we're not tying ourselves to lightning, but we can also handle other use cases like subscriptions, terms of service, and anything else a relay cares to pack into the page its url points to. 

URL alone does add friction though. My goal was to get going as soon as possible. On-click, NWC payment if possible. Every extra click matters. 

Also, the URL is there, but it is the exception: "pay this to continue now or look at this URL for terms + other options".

staab:
What if clients just "open" the url and custom protocol handlers can take over from there? So https: would go to a browser, nostr: would go to the user's nostr app, invoice: (or whatever that prefix would be) would open a wallet. Clients can then special case those custom protocols if they want to handle them explicitly.

vitorpamplona:
> What if clients just "open" the url and custom protocol handlers can take over from there?

We can do lightning:lnbc...  I am not sure if the generalization is useful. 

Maybe we can the invoice optional and the URL the message's identifier?

staab:
> Maybe we can make the invoice optional and the URL the message's identifier?

I would be fine with this. To justify the generalization though, it would allow relays to control the UX. If all they want is a payment and want to support a minimal-clicks UX, they can send an invoice and clients can use NWC to pay it (or redirect to a wallet if NWC isn't supported). If they want to offer other payment methods, ask users to create an account or accept a ToS, or provide additional tools for relay configuration or whatever, they can send a vanilla url to a page they own. They could even redirect users to a note using a `nostr:` prefix that explains how to gain access to the relay. This makes this proposal simpler, and allows for relays to have pretty unlimited flexibility and control over the UX.

vitorpamplona:
Well, if that is the case, then we can just have a URL and a Kind1-renderable description. So, If there is a lightning invoice in it, it will be shown. If there is a cashu token as a refund, it will be shown as well. Any other links would be clickable. Images can also be present. 

Then the description should contain the entire text, including what to do if you decide to not continue to pay for the subscription (e.g. removing the relay from the relay list). 

Then relays can also use it to show a... post of the day, for instance. It is not a PAY-only instruction anymore. 

We can do a `USER_MESSAGE` spec instead of `PAY` only

staab:
I can't tell if you're being sarcastic, but yes haha that sounds fine. Although I would again not call it interrupt_user, since it's not necessary to make it an interruption. Clients could aggregate these in the background and create a digest, automatically handle them, whatever.

vitorpamplona:
Sorry. No, I wasn't sarcastic :) 

The issue is that the message is related to whatever the user is trying to use at that moment. For instance, if the user is trying to access Chats and this relay is paid for Chats, the relay won't return anything until the user Pays and the screen will be empty. So, it makes sense to have an immediate interruption of the flow in order to continue the UX the user wants. 

Putting my sales hat, an immediate response is a big deal. If these messages just go to notifications then we don't need a spec for it. The relay can just send a Nostr event that will show up in Notifications anyway.

staab:
Interesting point. Why do we need to add this to the spec then? A relay could send a NIP 24 DM from the relay's pubkey specifying payment details (or we could just create a new kind that represents a general-purpose payment request). This is how it's already done in fact. All the UX stuff you're talking about could be supported by clients pretty easily by identifying pubkeys that belong to relays. Latency isn't a problem, because the relay can send the DM using the connection that's already open.

vitorpamplona:
Sure, as long as the new event kind has the current filter/subscription in it so the app knows what request (or post) triggered it, we can do it as an event kind as well. 

I don't think we have an official pubkey per relay, do we? We have the owner's key, but not a key that the relay could use to send new notifications anywhere. 

Clients will have to add a new filter to listen to those events from special keys for each relay being used, but sure. 

I think this is just simpler.

vitorpamplona:
Ok, I just changed this to a more generic messaging spec and removed the direct dependency on bitcoin invoices. 

Let me know what you all think.

nostr-wine:
We've been working with Vitor on this and we are happy with the latest version of using NOTIFY instead of PAY and letting the relay decide what messaging to use.

I think it is important that clients always show the ws url that sent the NOTIFY when it is displayed, particularly if it includes an invoice.

jb55:
On Tue, Nov 28, 2023 at 09:48:21AM -0800, nostr-wine wrote:
>We've been working with Vitor on this and we are happy with the latest
>version of using NOTIFY instead of PAY and letting the relay decide
>what messaging to use.
>
>I think it is important that clients always show the ws url that sent
>the NOTIFY when it is displayed, particularly if it includes an
>invoice.

I think that makes the most sense.

nostr-wine:
inbox.nostr.wine has been updated with the new spec for testing. Here is an example:

```
Connected to wss://inbox.nostr.wine.
> ["REQ","cn",{"limit":2,"kinds":[4], "#p": ["7088717b932504f20b337bf3f9aea0b76516e4f3e1e504eadb8bd2dfa962b103"]}]
< ["AUTH", "e677f109-a033-4880-8bf9-291d86c456f7"]
> ["AUTH",{"id": "ca0b6b7c68b5d86409dbf897fb874e3fe1c63e0690d5ff355ea05677d88d5150","pubkey": "7088717b932504f20b337bf3f9aea0b76516e4f3e1e504eadb8bd2dfa962b103","created_at": 1701195326,"kind": 22242,"tags": [["relay","wss://inbox.nostr.wine"],["challenge","e677f109-a033-4880-8bf9-291d86c456f7"]],"content": "","sig": "e3a5bc45c322a773621bf7f3c43eca85e21b6f76e8ba4af0017c01e37b38a0ef334033982c564a83449904db2397c2740df917737e40bcb78f1dcb0971ba8156"}]
< ["OK", "ca0b6b7c68b5d86409dbf897fb874e3fe1c63e0690d5ff355ea05677d88d5150", false, "error: user unauthorized"]
< ["NOTIFY", "This relay requires an active subscription. You can purchase access for 30 days by paying the invoice below.\nlnbc50u1pjkvtj8sp570z6jc6zpdn522s4lmjw7s3clapad70ds6cswewf6v8ha2xsws5qpp5gz4pjr8h2fh2uyyzyfg7xslfzh6aa56zg85ws9h6z84k27qy7j2qdpdd9hxymmc9ehx7um5wgh8w6twv5sy6mmww35xc7fqgejk2xqy6thlcqpjrzjqf9l4ux2helcwn7n86lhcm6w2wzew879qnhn7jfyxt578mrhux6u7rzgqyqqtwqqqqqqqqqqqqqqtkcqjq9qxpqysgqqzdjhk3ft2pa0pxg87lu7dqk3pxfrrgtnqlpu7tqamyayvhz3dunn6reersjg9u08w2hqj8vqlxaapat0tdecyzjm24mdzlfvr86ymcpma3762\nBy paying this invoice, you are accepting our terms and conditions.\nLearn more and see additional subscription options on https://inbox.nostr.wine/\nIf you do not intend to use this relay anymore, please remove it from your relay list."]
```

vitorpamplona:
New NOTIFY request is now available on Amethyst and both nostr.wine and inbox.nostr.wine

vitorpamplona:
So many ideas with this design :)

Yonle:
> So many ideas with this design :)

Including being "Message of the day".
Screenshot_20231201-005028

vitorpamplona:
@Semisol feel free to propose an alternative that is actually implemented in your relays. I am happy to code and test if it's better than this. 

As you know, I have been asking for something like this for 5 months already. You ended up never proposing anything. @nostr-wine not only proposed, but implemented, shipped and it's working really well.

vitorpamplona:
Clients can block Notifications from relays that are not on the user's relay list. It's up to them. But if people have a paid relay in their list and are not paying, I think it's fair to blast them with payment requests. They can choose to pay or to remove the relay from their main relay list. 

If clients are pinging different relays than those the user expects them to ping, they don't need to show such notifications if they don't want to. The NOTIFY works best for relays the user knows and is choosing to use. 

> but generally we shouldn't create new message types that could largely be ignored.

Relays do that all the time, though. Most of the "NOTICE" messages are completely ignored. 

> create a new ephemeral event kind for relay notifications, and clients...

It's an alternative. But you have to create the infrastructure to make it work. Clients will need to know what is the pubkey used for those messages for each relay and discard everything else. Right now, relays don't have a pubkey for them to message clients. Clients will also need to make sure they don't notify from an old event that was stored in a separate relay and it just happens that that relay decided to send that event. 

In other words, competing relays can keep old versions of these notifications, and as long as they are sent before a newer event, the user will think it is real. Basically, relays can send a popup hell chain by collecting and rebroadcasting events from other relays. 

Anyway, we went there and it looked very messy. I am not opposed to it, but I don't know if I mapped out all the threats in that model.

monlovesmango:
>Most of the "NOTICE" messages are completely ignored.

agree, but its the only message type that is ignored, and it is needed bc we need at least one way to communicate why connection failed, which shouldn't generally be user facing. this wasn't designed to be a user facing message, NOTIFY is.

>But you have to create the infrastructure to make it work. Clients will need to know what is the pubkey used for those messages for each relay and discard everything else. Right now, relays don't have a pubkey for them to message clients.

no that is not what I am proposing. the pubkey of the ephemeral event could be anything and the client doesn't need to know what it is. the client would just send a req with filter { kinds: [relayNotificationKind], #p: [userPubkey] } and the relay would know to create an ephemeral event (with any key) to send to the client. this should require zero extra infrastructure above what is proposed in the NOTIFY workflow.

basically a relay notification event should be associated with the relay it comes from, not the pubkey on the event. so the "competing relays can keep old versions of these notifications" doesn't really apply.

vitorpamplona:
Why do you need to filter, then? The relay can just send it at any point in the regular subscription. 

It's weird that it is a random key. The relay will need to make sure to not accept these events from anyone else because there is no way for the client to check if this was created at the relay or if it was broadcasted by somebody else. 

Sounds dangerous. Like relays that don't implement this NIP wouldn't know they have to block a given event kind and will pass the notification down.

vitorpamplona:
> seems like an unrealistic attack

After I saw people trying to dupe users with fake replies on the Nostr Wallet Connect, I think everything is possible. But in that case, the client knows which key is supposed to be used and can discard everything else. 

> takes a lot of effort for very low gain

I am not really sure if it is very low. Somebody can claim that their subscription to nostr.wine is up and include an invoice to the attacker's wallet. Just create payloads for each user in Nostr and send them all to unsupporting relays. This could be a profitable attack. 

But if somebody wants to create this variation, I can do a test code as well.

monlovesmango:
>Just create payloads for each user in Nostr and send them all to unsupporting relays. This could be a profitable attack.

they would have continuously be sending these for thousands of users to multiple relays (while avoiding being filtered by relay firewall) in the hope that for one specific user it overlaps with the 1-2 sec it takes the client to query for relay notifications (as the sub really shouldn't need to be kept open). sounds like lot of needed resources with low probability of success and low payout.

and again, could recommend only querying relays who support this nip.

mikedilger:
This doesn't seem like a well thought out solution:

> Clients SHOULD expect the same request to be received multiple times. If the user dismisses it once, the client SHOULD ignore new requests with the same message.

Are you saying relays must watermark their messages to make sure they never repeat?

vitorpamplona:
> Clients SHOULD expect the same request to be received multiple times. If the user dismisses it once, the client SHOULD ignore new requests with the same message.

This was more of a protection against poorly implemented relays than anything else. We see relays sending 20 AUTH messages a second, so it's not hard to imagine that some implementations will send similar amounts of NOTIFY. 

For Amethyst, we only allow one new message per session (go to the background and come back). Repeated messages can play when the user resumes the app unless the user marks to always dismiss this message, then they won't see it anymore.

Yonle:
This function worked almost as similiar as javascript `alert()`. So we are neither using the current `NOTIFY` or the `ALERT`

quentintaranpino (approved):
I was not aware of this pull request, it makes sense, I see it as easy to implement, simple, and it helps mitigate (in my opinion) a problem of Nostr, the long-term sustainability of the relays.

jeremyd:
we're gonna need a new NIP# on this badboy and support for lnurlp :+1: this is now implemented in relay.tools and I think it's great.  thank you clients that implement this :pray:

nostr-wine:
We've been using NOTIFY for 18 months now and it has been a huge success for us (and Amethyst). It drastically reduces the friction necessary for a new user signing up to a paid relay or an existing customer renewing their subscription.

I'm sure NOTIFY has other use cases outside of payments too but just wanted to once again voice our support for this PR after using it for a while.

fiatjaf:
Reading this again isn't it exactly the same as NOTICE?

mikedilger:
As a data point:  Gossip logs NOTICE to the console, but shows NOTIFY to the user in the GUI. These two messages only get about 8 lines of code in gossip (outside of their enum definition in nostr-types):

````rust
RelayMessage::Notice(msg) => {
    tracing::warn!("{}: NOTICE: {}", &self.url, msg);
}
RelayMessage::Notify(msg) => {
    let message = format!("{}: NOTIFY: {}", &self.url, msg);
    tracing::warn!(message);
    GLOBALS.status_queue.write().write(message);
}
````

vitorpamplona:
> Reading this again isn't it exactly the same as NOTICE?

Notice is for debugging. 

Notify is designed to show directly to the user. 

The relay can choose which to use.

vitorpamplona:
You can add as many invoices as you want in the message. The examples are just examples. 

Since these messages are short-lived and supposed to be presented immediately to the user (so that the query can continue), you can delete unpaid invoices in a few minutes.

kehiy:
seems interesting. is there any potential that relay spam users client?

vitorpamplona:
> seems interesting. is there any potential that relay spam users client? 

Yep. They are free to do so and piss off their own users.

PR #902, opened 2023-11-27
`CLOSED` messages for relays that want to reject REQs and NIP-42 `AUTH` integration

fiatjaf:
I will quote my own message from https://github.com/nostr-protocol/nips/pull/841#issuecomment-1828345933 to explain this:

> Basically we have two distinct problems:
> 
> 1. knowing why a relay is asking for `AUTH`
> 2. knowing why a subscription has failed
> 
> adding a third element to `AUTH` only solves them in the occasion that the two happen together.
> 
> creating a new message solves 2 and making `AUTH` be subservient to that (and also to `OK`) solves 1.

More relevant messages were posted to that other thread but should probably have been posted to this.

### DIFF
--- a/42.md
+++ b/42.md
@@ -66,16 +66,15 @@ subscriptions by clients):
 
 ## Protocol flow
 
-At any moment the relay may send an `AUTH` message to the client containing a challenge. The challenge is to be valid for the duration of
-the connection or until a next challenge is sent by the relay. The client MAY decide to send its `AUTH` event at any point and the
-authenticated session is valid afterwards for the duration of the connection.
+At any moment the relay may send an `AUTH` message to the client containing a challenge. The challenge is valid for the duration of
+the connection or until another challenge is sent by the relay. The client MAY decide to send its `AUTH` event at any point and the
 
 ### `auth-required` in response to a `REQ` message
 
 Given that a relay is likely to require clients to perform authentication only for certain jobs, like answering a `REQ` or accepting an
 `EVENT` write, these are some expected common flows:
 
-```json
+```
 relay: ["AUTH", "<challenge>"]
 client: ["REQ", "sub_1", {"kinds": [4]}]
 relay: ["CLOSED", "sub_1", "auth-required: we can't serve DMs to unauthenticated users"]
@@ -98,7 +97,7 @@ response to the `auth-required` `CLOSED` message.
 The same flow is valid for when a client wants to write an `EVENT` to the relay, except now the relay sends back an `OK` message instead of
 a `CLOSED` message:
 
-```json
+```
 relay: ["AUTH", "<challenge>"]
 client: ["EVENT", {"id": "012345...", ...}]
 relay: ["OK", "012345...", false, "auth-required: we only accept events from registered users"]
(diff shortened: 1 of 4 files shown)


42.md
```diff
@@ -50,27 +56,56 @@ Relays MUST exclude `kind: 22242` events from being broadcasted to any client.
 }
 ```
 
+### `OK` and `CLOSED` machine-readable prefixes
+
+This NIP defines two new prefixes that can be used in `OK` (in response to event writes by clients) and `CLOSED` (in response to rejected
+subscriptions by clients):
+
+- `"auth-required: "` - for when a client has not performed `AUTH` and the relay requires that to fulfill the query or write the event.
+- `"restricted: "` - for when a client has already performed `AUTH` but the key used to perform it is still not allowed by the relay.
```
monlovesmango, line 65:
```suggestion
- `"restricted: "` - for when a client has already performed `AUTH` but the key used to perform it is still not allowed by the relay or is exceeding it's authorization.
```

monlovesmango:
some minor changes, but I approve.

personally think we should just use "CLOSE" rather than "CLOSED" to mirror the client side "CLOSE" (in the same way that "EVENT" is the same on client and relay sides and mean the same thing). seems way more consistent to me.

monlovesmango (approved):

arthurfranca:
ACK!! The flow is very clear now with examples and all.

Personally I would just tell relays to always send AUTH right after connection is established. It leaves less room for errors (like sending CLOSED/OK before AUTH).

staab (approved):

monlovesmango:
should this also apply to nip 45?

mikedilger:
+1

42.md
```diff
@@ -50,27 +49,49 @@ Relays MUST exclude `kind: 22242` events from being broadcasted to any client.
...
+
+- `"auth-required: "` - for when a client has not performed `AUTH` and the relay requires that to fulfill the query or write the event.
+- `"restricted: "` - for when a client has already performed `AUTH` but the key used to perform it is still not allowed by the relay or is exceeding its authorization.
+
 ## Protocol flow
 
-At any moment the relay may send an `AUTH` message to the client containing a challenge. After receiving that the client may decide to
-authenticate itself or not. The challenge is expected to be valid for the duration of the connection or until a next challenge is sent by
-the relay.
+At any moment the relay may send an `AUTH` message to the client containing a challenge. The challenge is valid for the duration of the connection or until another challenge is sent by the relay. The client MAY decide to send its `AUTH` event at any point and the authenticated session is valid afterwards for the duration of the connection.
```
proudmuslim-dev, line 61:
```suggestion
At any moment the relay may send an `AUTH` message to the client containing a challenge. The challenge is valid for the duration of the connection or until another challenge is sent by the relay. The client MAY decide to send its `AUTH` event at any point and the authenticated session is then valid for the remainder of the connection.
```

proudmuslim-dev (changes requested):

fiatjaf:
Implemented in https://nostr.wine

nostr-wine:
Ha - you beat me to it. We implemented CLOSED with the new flow for DM REQs on all nostr.wine regional relays and will work on the rest of our relays today.

PR #904, opened 2023-11-27
Add imeta tag

staab:
This is adapted from Damus' DIPS repo: https://github.com/damus-io/dips/blob/master/01.md. I don't love the format, but I think there are probably a good number of events in the wild already using it. This PR is backwards-compatible except for renaming sha256 to x following NIP 94.

@jb55 correct me if I'm wrong and there is an opportunity to revisit the format. My preference would be to make the `url` the first argument of the tag rather than the value of the `url` key to make it more conventional. It would also be nice to use a query string or json or something instead of space delimited data.

### DIFF
--- a/29.md
+++ b/29.md
@@ -1,0 +1,43 @@
+NIP029
+======
+
+imeta
+--------------
+
+`imeta` is a tag for adding media attachments to events. `imeta` tags MUST match URLs
+in the event content. Clients may replace imeta URLs with rich previews. `imeta` tags
+contain extra information about the media attachment, which clients can use to provide
+a better experience when loading images.
+
+The `imeta` tag is variadic, and each entry is a space-delimited key/value pair.
+Each `imeta` tag MUST have a `url`, and at least one other field. `imeta` may include
+any field specified by [NIP 94](./94.md). There SHOULD be only one `imeta` tag per url.
+
+## Example
+
+```json
+{
+  "content": "More image metadata tests don’t mind me https://nostr.build/i/my-image.jpg",
+  "kind": 1,
+  "tags": [
+    [
+      "imeta",
+      "url https://nostr.build/i/my-image.jpg",
+      "blurhash eVF$^OI:${M{o#*0-nNFxakD-?xVM}WEWB%iNKxvR-oetmo#R-aen$",
+      "dim 3024x4032",
+      "alt A scenic photo overlooking the coast of Costa Rica",
+      "x <sha256 hash as specified in NIP 94>",
+      "fallback https://nostrcheck.me/alt1.jpg"
+      "fallback https://void.cat/alt1.jpg"
+    ]
+  ]
+}
+```
+
+## Recommended client behavior
+
+When uploading images during a new post, clients MAY include this metadata
+after the image is uploaded and included in the post.
+
+When pasting urls during post composition, the client MAY download the image
+and add this metadata before the post is sent.
--- a/94.md
+++ b/94.md
@@ -25,6 +25,7 @@ This NIP specifies the use of the `1063` event type, having in `content` a descr
 * `image` (optional) url of preview image with same dimensions
 * `summary` (optional) text excerpt
 * `alt` (optional) description for accessibility
+* `fallback` (optional) zero or more fallback file sources in case `url` fails
 
 ```json
 {


v0l:
snort is pushing some `imeta` tags when using nostr.build, happy to see this added to nips:
https://git.v0l.io/Kieran/snort/src/commit/509228d53287b62cc79c9f4d28fdfe1ac63dd029/packages/app/src/Element/Event/Create/NoteCreator.tsx#L198-L205

fabianfabian:
Nostur is already reading imeta tags, and I am now in the process of adding them to new posts. I don't have strong opinion against changing the spec but changing it now would mean I would have to support 2 versions because there are already many imeta posts out there.

fiatjaf:
I think there is no way to change this. Let's merge it.

vitorpamplona:
I continue my position that static content should never be referenced with URLs inside Nostr due to: 
1. the centralizing nature of domain name registrars, 
2. the full state control of the DNS resolution stack, 
3. the centralizing nature of the domain name resolution to a single provider
4. the immutability break of today's extremely frequent editing of such images by the image host without the author's approval or review
5. the potential for rug-pulling a massive amount of users at once by a hostile takeover of an image host provider to replace the image and video files for profit (ads, porn, SEO-hacking, etc) or criminal activities (scams, impersonation, etc). 
 
The dynamic nature of the content resolution makes it impossible for a Nostr or Web crawler to correctly weigh the content of such URLs since the crawler cannot verify if the image host has changed the contents without the author's consent or not, making Nostr search way more difficult to correctly index. 

But I won't block adding this to the repo if people are using it. 2+ clients rule always win. 

I just think it is a huge mistake.

arthurfranca:
I don't get why not use [NIP-54](https://github.com/nostr-protocol/nips/pull/521).
NIP-54 reuses and supports all NIP-94 tag names and works both inline (in `.content`) or inside a tag.

For fallback urls it would be like this:
`https://abc.com/img.jpg#url=<fallback-url-1>&url=<fallback-url-2>` and it would work inside `.content` for both `kind:1` and `kind:30023` markdown.

It is like query params but using `#` instead of `?` so that the params don't get sent to the server (in the example an `<a>` html tag with the src as `https://abc.com/img.jpg#url=<fallback-url-1>&url=<fallback-url-2>` would request just `https://abc.com/img.jpg` and could use `onerror` handler to swap src to the fallback url)

jb55:
On Tue, Nov 28, 2023 at 05:08:57AM -0800, Vitor Pamplona wrote:
>I continue my position that static content should never be referenced with URLs inside Nostr due to:
>1. the centralizing nature of domain name registrars,
>2. the full state control of the DNS resolution stack,
>3. the centralizing nature of the domain name resolution to a single provider
>4. the immutability break of today's extremely frequent editing of such images by the image host without the author's approval or review
>5. the potential for rug-pulling a massive amount of users at once by a hostile takeover of an image host provider to replace the image and video files for profit (ads, porn, SEO-hacking, etc) or criminal activities (scams, impersonation, etc).
>
>The dynamic nature of the content resolution makes it impossible for a Nostr or Web crawler to correctly weigh the content of such URLs since the crawler cannot verify if the image host has changed the contents without the author's consent or not, making Nostr search way more difficult to correctly index.
>
>But I won't block adding this to the repo if people are using it. 2+ clients rule always win.
>
>I just think it is a huge mistake.

Your tirade against URLs is bizarre. URLs are here to stay unless there
are some fundamental changes to the internet itself. nostr is an
internet protocol, it will naturally interact with other aspects of the
internet and the web.

jb55:
On Mon, Nov 27, 2023 at 02:50:53PM -0800, hodlbod wrote:
> It would also be nice to use a query string or json or something instead of space delimited data.

I maintain that splitting on the first space is way simpler than query
string or json encoding. This scheme allows a variable number of
arguments without fixed positioning downsizes, and allows us to not
duplicate the url string into many tags.

jb55:
On Tue, Nov 28, 2023 at 06:26:24AM -0800, arthurfranca wrote:
>I don't get why not use
>[NIP-54](https://github.com/nostr-protocol/nips/pull/521). NIP-54
>reuses and supports all NIP-94 tag names and works both inline (in
>`.content`) or inside a tag.
>
>For fallback urls it would be like this:
>`https://abc.com/img.jpg#url=<fallback-url-1>&url=<fallback-url-2>` and
>it would work inside `.content` for both `kind:1` and `kind:30023`
>markdown.
>
>It is like query params but using `#` instead of `?` so that the params don't get sent to the server (in the example an `<a>` html tag with the src as `https://abc.com/img.jpg#url=<fallback-url-1>&url=<fallback-url-2>` would request just `https://abc.com/img.jpg` and could use `onerror` handler to swap src to the fallback url)

I would use this for DMs, but imeta for public stuff allows you to
quickly pull out the blurhash before parsing the contents. Pulling out
the blurhash as quick as possible so that you can generate an image is
pretty important if you want to load posts as quickly as possible
without image popping and before the image has been downloaded.

jb55 (approved):

mattn:
I don't strong opinion, but is better to be two elements tags? since this can pass empty value like `["alt": ""]`
```
{
  "content": "More image metadata tests don’t mind me https://nostr.build/i/my-image.jpg",
  "kind": 1,
  "tags": [
    [
      "imeta",
      "url https://nostr.build/i/my-image.jpg",
      "blurhash eVF$^OI:${M{o#*0-nNFxakD-?xVM}WEWB%iNKxvR-oetmo#R-aen$",
      "dim 3024x4032",
      "alt A scenic photo overlooking the coast of Costa Rica",
      "x <sha256 hash as specified in NIP 94>",
      "fallback https://nostrcheck.me/alt1.jpg"
      "fallback https://void.cat/alt1.jpg"
    ]
  ]
}
```

jb55:
On Tue, Nov 28, 2023 at 08:46:05AM -0800, mattn wrote:
>I don't strong opinion, but is better to be two elements tags? since
>this can pass empty value like `["alt": ""]`

If you want an empty value you just don't include the tag element? If
there was more than one image what would you do? The point of imeta is
to group metadata for each image. The querystring thing is another
approach but this modifies the contents and makes the urls huge.

>```
>{
>  "content": "More image metadata tests don’t mind me https://nostr.build/i/my-image.jpg",
>  "kind": 1,
>  "tags": [
>    [
>      "imeta",
>      "url https://nostr.build/i/my-image.jpg",
>      "blurhash eVF$^OI:${M{o#*0-nNFxakD-?xVM}WEWB%iNKxvR-oetmo#R-aen$",
>      "dim 3024x4032",
>      "alt A scenic photo overlooking the coast of Costa Rica",
>      "x <sha256 hash as specified in NIP 94>",
>      "fallback https://nostrcheck.me/alt1.jpg"
>      "fallback https://void.cat/alt1.jpg"
>    ]
>  ]
>}

AsaiToshiya:
> #751

I really like this idea.

jb55:
On Tue, Nov 28, 2023 at 08:53:19AM -0800, Alex Gleason wrote:
>There really needs to be a way to represent media attachments that are
>not part of the text content. This is very common on platforms such as
>Twitter, Facebook, Instagram, email, Mastodon, and practically every
>other social platform users are accustomed to that isn't Medium, and it
>fills me with existential dread that we cannot do it on Nostr.

What does this gain you over urls and metadata?

vitorpamplona:
> Your tirade against URLs is bizarre. URLs are here to stay unless there are some fundamental changes to the internet itself. nostr is an internet protocol, it will naturally interact with other aspects of the internet and the web.

They can stay. I just find it dumb to have a decentralized protocol based on a centralized and even state-controlled infrastructure. 

But people never pay attention to these things until they get rug pulled. I was rug-pulled once over DNS when doing signed payloads between two countries at war with one another. I learned my lesson.

jb55:
I think there are room for both approaches but not everyone is going to have the same threat model.

vitorpamplona:
> I think there are room for both approaches but not everyone is going to have the same threat model.

People always say that until somebody buys nostr.build out and start advertising fake Damus coins in the images of your own signed posts for everyone but your own location. People will be scammed by your own posts and you won't even see until it's too late. 

That's the threat we should fix as soon as possible.

alexgleason:
> What does this gain you over urls and metadata?

By replacing URLs with inline attachments in posts (which most clients do), we have inadvertently created an undocumented custom data format for Nostr similar to Markdown or HTML.

I just want to add `imeta` tags to my events, and clients will display them as attachments even if they don't match any URL in the post content.

jb55:
On Tue, Nov 28, 2023 at 09:26:11AM -0800, Vitor Pamplona wrote:
>> I think there are room for both approaches but not everyone is going to have the same threat model.
>
>People always say that until somebody buys nostr.build out and start
>advertising fake Damus coins in the images of your own signed posts for
>everyone but your own location. People will be scammed by your own
>posts and you won't even see until it's too late.

>That's the threat we should fix as soon as possible.

Then I would just add sha256 to imeta? I would download the image after
upload and then hash it before posting.

vitorpamplona:
> Then I would just add sha256 to imeta? I would download the image after upload and then hash it before posting.

Not just add but verify as well. Then that does solve one of the issues there. 

Then you have the rest: 
- User tracking by the image host. 
- Posts might be broken when such a host disappears. 
- Image hosts might fully censor your past or just decrease the bandwidth/increase the latency for disliked persons. 
- Image hosts might request payment from downloaders before sending the image/video. 

So many things can happen after the post is signed...

fabianfabian:
For me the image dimensions and hash are the most important, this solves jumpy image loading and content rug replacement. Where this info is seems less relevant as long as its somewhere in the event, NIP-54 and imeta seems to solve it good enough, not sure why one would be better than the other, but imeta is already in use.

I don't think DMs is a good enough reason to use NIP-54, because DMs shouldn't leak any tags at all, this needs to be fixed on the DM level so you can add any tag probably in a way similar to NIP-51 lists. Otherwise for every spec that requires tags you need to find a special way to use it in DMs also.

vitorpamplona:
> I don't think DMs is a good enough reason to use NIP-54, because DMs shouldn't leak any tags at all

Agree, DMs should never, ever have any tag but the receiver's pubkey. 

And static content (pictures, videos, pdfs, etc) transferred through DMs must always be encrypted to the receiver in the same way the content of the event is. We tested that expectation a few months back with our users. Neither a hidden but public link nor an image encrypted with a shareable secret is enough. People expect significant privacy on DM attachments.

jb55:
On Tue, Nov 28, 2023 at 09:37:34AM -0800, Vitor Pamplona wrote:
>> Then I would just add sha256 to imeta? I would download the image
>> after upload and then hash it before posting.
>
>Not just add but verify as well. Then that does solve one of the issues there.
>
>Then you have the rest:
>- User tracking by the image host.

Somewhat ironic, don't you track every view on every post on your app
using images? VPN or image proxies solve this. Image proxies could
modify the image, so you would need to fallback to the original link and
check the source if the hash check failed, an action the user could do
manually if they were paranoid enough.

>- Posts might be broken when such a host disappears.

This is what the fallback url is for

>- Image hosts might fully censor your past or just decrease the
>bandwidth/increase the latency for disliked persons.

Based on what, their IP? Use a VPN.

>- Image hosts might request payment from downloaders before sending the
>image/video.

That's a different spec.

jb55:
On Tue, Nov 28, 2023 at 09:39:29AM -0800, Fabian wrote:
>I don't think DMs is a good enough reason to use NIP-54, because DMs
>shouldn't leak any tags at all.

I was only referring to the querystring mechanism for metadata, which
wouldn't leak anything.

staab:
I didn't read Vitor and Will's entire lovers' quarrel, but it sounds like we can merge this? @alexgleason I agree attachments are a good thing to support, is there anything here that doesn't cover that use case? We could definitely clarify that using them for attachments is ok.

29.md
```diff
@@ -0,0 +1,41 @@
+NIP029
+======
+
+Image Metadata
+--------------
+
+Image metadata is additional information associated with urls in notes that
```
alexgleason, line 7:
```suggestion
`imeta` is a tag for adding media attachments to events. Typically `imeta` tags match URLs in the event content, but `imeta` tags with no matching URLs SHOULD still be displayed with the event (such as underneath the post). Clients may replace imeta URLs with rich previews. `imeta` tags contain extra information about the media attachment, which
```

alexgleason (changes requested):
@staab Here's what I think. I want first-class media attachments on Nostr. I think the imeta format is good as it is, and given that it's widely used, it can stay exactly the same with no changes. What I want to change is the framing of this NIP. `imeta` is a tag that adds media attachments. This introduces a new behavior, which is that clients SHOULD render `imeta` previews even if the they don't match any URLs in the content.

alexgleason:
@staab I added a commit here: 8f3dce2fbcc68701139d1bf7370e2d6d0e9b6133 

Do you agree with these changes @jb55 @fiatjaf ? Reasoning: https://github.com/nostr-protocol/nips/pull/904#pullrequestreview-1753907086

jb55:
On Tue, Nov 28, 2023 at 01:16:14PM -0800, Alex Gleason wrote:
>@staab I added a commit here: 8f3dce2fbcc68701139d1bf7370e2d6d0e9b6133
>
>Do you agree with these changes @jb55 @fiatjaf ? Reasoning: https://github.com/nostr-protocol/nips/pull/904#pullrequestreview-1753907086

well it would break images in most clients but at least there's an
upgrade path that isn't too difficult.

I still like urls for plaintext clients that could care less about an
image metadata spec.

29.md
vitorpamplona: Is the idea to include videos, gifs, and PDFs in this?

alexgleason:
The idea of rendering previews for URLs _without_ some kind of metadata like `imeta` is fundamentally flawed. How do you render a preview for this URL? https://media.gleasonator.dev/ipfs/QmVLHE4WC6fbArqm2ndqqYcEsdPmZEHz3jTJGtup4QV5gB

Relying on the file extension is flawed. Some URLs (like above) do not have a file extension. Some file extensions are ambiguous, like `.ts`, which is both a TypeScript file and a video format. We need an explicit media type at the very least. Just because people are doing it currently doesn't mean it's good.

mikedilger:
Clients that do not (yet) recognize imeta will ignore the tag and display the post with no preview and no link either.

fiatjaf:
Just append the URL to the text so people can click on it.

arthurfranca:
> How do you render a preview for this URL? https://media.gleasonator.dev/ipfs/QmVLHE4WC6fbArqm2ndqqYcEsdPmZEHz3jTJGtup4QV5gB [...] Relying on the file extension is flawed.

When extension is absent or ambiguous the client can append the mime type using nip54: `#m=image%2Fpng`. It solves inline metadata.

alexgleason:
imeta and hash queries are virtually the same solution. The point I'm trying to make is that the initial idea - that we should treat URLs in the content as media attachments - was not a good idea in the first place.

Furthermore, what if I WANT to include a URL like gleasonator.com/pikachu.png in the post text, and I specifically don't want it to be rendered as an image? There's no way to escape it, which is an oversight. For example, I may want to share the URL itself because we are having a technical discussion about the format of URLs, and it inadvertently gets replaced with an image.

It only makes sense so treat URLs as attachments IF there is a matching imeta tag. @jb55 you should be in support of this because it will encourage clients to actually add imeta tags giving us the information we need to render media properly in Damus etc.

NIPs should be used to canonicalize what already exists, but they are ultimately just a recommendation. And since people don't have to follow NIPs anyway, we should at least make our recommendations be good.

fiatjaf:
> what if I WANT to include a URL like gleasonator.com/pikachu.png in the post text, and I specifically don't want it to be rendered as an image? There's no way to escape it, which is an oversight. For example, I may want to share the URL itself because we are having a technical discussion about the format of URLs, and it inadvertently gets replaced with an image.

I agree about this and it bothers me very much that everything gets turned into media previews. But honestly this problem exists in Twitter too and in other places.

I think it can be fully solved by some kind of opt-in "do-not-render-previews-on-this" indicator, maybe even in an `imeta` tag.

jb55:
On Wed, Nov 29, 2023 at 07:45:03AM -0800, fiatjaf_ wrote:
>> what if I WANT to include a URL like gleasonator.com/pikachu.png in
>> the post text, and I specifically don't want it to be rendered as an
>> image? There's no way to escape it, which is an oversight. For
>> example, I may want to share the URL itself because we are having a
>> technical discussion about the format of URLs, and it inadvertently
>> gets replaced with an image.
>
>I agree about this and it bothers me very much that everything gets
>turned into media previews. But honestly this problem exists in Twitter
>too and in other places.
>
>I think it can be fully solved by some kind of opt-in
>"do-not-render-previews-on-this" indicator, maybe even in an `imeta`
>tag.

In damus you can copy the URL from the preview or turn on an option that
disables previews, so these issues are really non-issues.

staab:
The worst case scenario with attachments is that clients that don't support them will render a note that says "see attached" and users will have no idea what that person is talking about. But this is a problem with attachments, not with the implementation. I don't think there's a way to add attachments in a way that avoids this, other than creating a new kind that supports attachments. But since many kinds may have an attachment, this will result in `n * m` total kinds. We can't do that every time we add a feature. So I think the UX problem is unavoidable and attachments will become a required part of the protocol if we add them. I support adding them, but it is awkward.

staab:
Ok, I can't think of any real problems with that.

fiatjaf:
Can you please decide on this versus the inline metadata? Otherwise we will end up everybody having to implement both in the long run.

staab:
> Can you please decide on this versus the inline metadata? Otherwise we will end up everybody having to implement both in the long run.

Probably too late for that, Coracle reads both and only publishes imeta.

29.md
alexgleason: Yes, why not? Clients shouldn't display previews for any file type they don't want to. They can display a generic "file download" component.

alexgleason:
As a "requested-changes" person on this MR, I've conceded the point about not having URLs in the text. It's work-around-able.

We do need some kind of media tag, though. I wish it were just called `media` instead of `imeta`.

It would support images, videos, and any other type of file. The `mime` would be required, and it would determine which other fields matter and how they're used.

We get the feeling that things are set in stone and we can't change them. But Nostr is really still in its early stages. We absolutely can afford to change, and we'll be glad we did 5 years from now. So I would strongly suggest naming the tag `media` instead of `imeta`, and keeping everything else the same. But if we don't do that, I won't die. Either way, using a tag is 1000% better than inline URL fragments.

fabianfabian:
Nostur also reads both but only publishes imeta.

jb55:
Damus only reads and publishes imeta

fiatjaf:
Snort also apparently reads and publishes `imeta`.

Looks like should merge this.

I think we should merge @alexgleason's code suggestions before.

staab:
@fiatjaf I thought the consensus was that appending urls to the end of text notes was the best way to handle this to avoid weird UI problems in clients that don't support attachments.

alexgleason:
When there is an `imeta` tag that doesn't match any URL in the content, do clients display the media or ignore the tag?

jb55:
ignore the tag

29.md
alexgleason: @staab I updated my suggested text here, but GitHub doesn't give me any way to commit it.

fiatjaf:
> @fiatjaf I thought the consensus was that appending urls to the end of text notes was the best way to handle this to avoid weird UI problems in clients that don't support attachments.

Yes, I definitely think that should be how it works. Isn't that what this NIP describes?

staab:
> Yes, I definitely think that should be how it works. Isn't that what this NIP describes?

Yes, Alex just had an outstanding suggestion that changed that. Resolved now. I've rebased and fixed a couple formatting things, merging now.

fiatjaf:
You have merged your own PR!

staab:
I kept waiting for you to do it

fiatjaf:
Outrageous!

(I'm kidding)

PR #905, opened 2023-11-28
Added encryption description for encrypted job feedback events

believethehype:


vitorpamplona:
I would add if the encryption follows NIP-04 or the new NIP-44 standard. 

Or if this is just for `tags` in `.content`, you can reference NIP-51 encryption as the standard to follow. If that changes, this changes too.

believethehype:
I added the information. As this already existed for job requests and job results, job feedback should use the same encryption of tags with nip04 as the other two events.

PR #911, opened 2023-12-01
Add kind Lightning.Pub's kind 21000

capitalistdog:
RPC messages for Lightning Pub


capitalistdog:
> Can you provide more info about how this kind should be used?

its just for ephemeral classification of a specific application: https://github.com/shocknet/Lightning.Pub/blob/master/proto/autogenerated/client.md

vitorpamplona:
Sure, just add the link like `kind:1971` did :)

capitalistdog:
added

fiatjaf:
What a lazy PR.

PR #914, opened 2023-12-03
Amend NIP-52 to require the content field to be provided, and decouple calendar identifier from calendar name

tyiu:


52.md
```diff
@@ -22,14 +22,14 @@ This kind of calendar event starts on a date and ends before a different date in
 
 The format uses a parameterized replaceable event kind `31922`.
 
-The `.content` of these events is optional and should be a detailed description of the calendar event.
+The `.content` of these events should be a detailed description of the calendar event. It is required but can be an empty string.
```
tyiu, line 25: I believe `content` is always expected on any kind. Please correct me if I'm wrong.

52.md
```diff
@@ -137,15 +137,23 @@ A calendar is a collection of calendar events, represented as a custom replaceab
 
 ### Format
 
+The `.content` of these events should be a detailed description of the calendar. It is required but can be an empty string.
+
 The format uses a custom replaceable list of kind `31924` with a list of tags as described below:
-* `d` (required) calendar name
```
tyiu, line 141: This was a huge miss by me in my opinion. Calendar names should be decoupled from their identifiers. If someone wants to rename their calendar, they should be able to do so without changing the entire addressable identifier.

staab (approved):

tyiu:
> Many other event kinds (lists being the most prominent ones) are using `title` instead of `name`, I'd suggest that we keep consistent here too.

@erskingardner Should that be mentioned in [NIP-24](https://github.com/nostr-protocol/nips/blob/master/24.md#tags) then? That it's a `title` tag that should typically be used to give an event kind a title / name.

erskingardner:
Yeah it probably should. Reading 24, I feel like there are a lot of other tags that could be added there.

tyiu:
> Yeah it probably should. Reading 24, I feel like there are a lot of other tags that could be added there.

@erskingardner I think my only hesitation to changing `name` to `title` is that it's sort of superfluous change and there are a few clients who have implemented NIP-52 as is already. Date-based and time-based calendar events already use `name`. What do you think?

`name` on calendar is net-new, so I could change that one to `title`.

vitorpamplona:
Just mark as deprecated. 

We can always check title first and then name if it doesnt exist. But new events then would use title.

tyiu:
I've implemented and merged this amendment in Nostr SDK for Apple Platforms:
https://github.com/nostr-sdk/nostr-sdk-ios/pull/123

tyiu:
Any chance this PR can be merged?

vitorpamplona:
LGTM

tyiu:
OK. If it looks good, whoever has merge powers, please merge.

24.md
```diff
@@ -39,3 +39,4 @@ tags
 These tags may be present in multiple event kinds. Whenever a different meaning is not specified by some more specific NIP, they have the following meanings:
 
   - `r`: a web URL the event is referring to in some way
+  - `title`: title of the event
```
AsaiToshiya, line 42: By "event" here, you mean the calendar event, right?

tyiu: Correct.

tyiu: Well, not necessarily, it could also just be the title of a non-calendar Nostr event.

AsaiToshiya: Ok, thanks!

PR #919, opened 2023-12-06
NIP-88: Discreet Log Contracts over Nostr

benthecarman:
Still pretty heavy WIP but wanted to get it out there for other people's feedback in nostr and DLC communities.

Will be implementing this in: https://github.com/MutinyWallet/mutiny-node/pull/888


bennyhodl:
Would using [NIP-44](https://github.com/nostr-protocol/nips/pull/746/files), if merged,  be a better option?

Instead of having to generate child keys for different contracts?

staab:
NIP 44 is currently being audited, it should be done pretttty soon.

erskingardner:
I'd be super curious to hear more about the use case you're planning on implementing in Mutiny (if you want to talk about it publicly). You mentioned betting and a marketplace for DLCs in the NIP draft. Is this mostly about wagers/bets with friends?

benthecarman:
> I'd be super curious to hear more about the use case you're planning on implementing in Mutiny (if you want to talk about it publicly). You mentioned betting and a marketplace for DLCs in the NIP draft. Is this mostly about wagers/bets with friends?

Yes, but can also be for financial things like creating pseudo stable coins

tvolk131:
I'm working on a team that's building a bounty adjudication system using DLCs and we were already planning on using Nostr as the communication protocol so it's awesome to see this spec proposed! For our use-case, we need a way for a potential participant in a DLC to request that an oracle create a new event announcement that it can then use. The announcement would be for a particular bounty, and if the oracle created the announcement it would be signing up to adjudicate whether the bounty is completed in the future. Would it make sense to add some method of requesting an oracle to create an announcement for an event? I would guess there are other potential use-cases for such behavior.

88.md
```diff
@@ -0,0 +1,130 @@
...
+  "kind": 88,
+  "content": "base64 contract details",
+  "tags": [
+    [
+      "relays",
+      "wss://nostr.mutinywallet.com",
+      "wss://relay.damus.io"
+    ],
+    [
+      "expires", // todo this is a good way to do this?
```
bonomat, line 70: What's the alternative you are thinking off? 

A DLC offer definitely needs an expiry.

88.md
```diff
@@ -0,0 +1,130 @@
...
+a unique opportunity to use a different transport layer for DLCs. This NIP proposes a new `kind:28_888` message that
+acts like a mailbox for DLC messages. The DLC messages are sent over nostr and available for the user the next time
+they open their wallet. This allows for a more mobile-friendly DLC experience.
+
+### `kind:28_888`
+
+// todo probably don't want an ephemeral event for this, but it's a good starting point
+
+Kind 28_888 is a simple message that contains a [NIP04](04.md) encrypted DLC message and it tagged with the recipient's
+public key with a `p` tag and if it is in reply to another event, that event should be tagged with an `e` tag.
```
bonomat, line 27: How should a client react to two competing messages following the same preceding event? 
I assume the client should take the last seen?

88.md
```diff
@@ -0,0 +1,130 @@
...
+  "pubkey": "97c70a44366a6535c145b333f973ea86dfdc2d7a99da618c40c64705ad98e322",
+  "created_at": 1679673265,
+  "id": "30efed56a035b2549fcaeec0bf2c1595f9a9b3bb4b1a38abaf8ee9041c4b7d93",
+  "sig": "f2cb581a84ed10e4dc84937bd98e27acac71ab057255f6aa8dfa561808c981fe8870f4a03c1e3666784d82a9c802d3704e174371aa13d63e2aeaf24ff5374d9d"
+}
+```
+
+## DLC Offers
+
+DLCs are done between two parties and Nostr provides a unique opportunity to create a global market for DLCs. This
```
bonomat, line 52: A DLC offer might only be _consumable_ once. To prevent that the receiver gets thousands of requests for the same offer which might have been taken ages ago,  maybe an additional message which allows to invalidate (or mark as taken) an old offer could be useful.

88.md
benthecarman: Yeah was looking to see if there was a tag already defined for expiry but seems like not, so will use this

88.md
benthecarman: Yeah a client should never send 2 messages and the receiver should only process the ones they expect.

Will add clarifications

88.md
benthecarman: Good point, changed to a parameterized replaceable event so the user could rescind it after it has been fulfilled

benthecarman:
> I'm working on a team that's building a bounty adjudication system using DLCs and we were already planning on using Nostr as the communication protocol so it's awesome to see this spec proposed! For our use-case, we need a way for a potential participant in a DLC to request that an oracle create a new event announcement that it can then use. The announcement would be for a particular bounty, and if the oracle created the announcement it would be signing up to adjudicate whether the bounty is completed in the future. Would it make sense to add some method of requesting an oracle to create an announcement for an event? I would guess there are other potential use-cases for such behavior.

Tbh this sounds out of scope and more applicable to something like DVMs

bennyhodl:
I don't think DLC messages should be ephemeral. It would cause issues of clients not being connected to completely miss messages.

I think that when an unseen event is received by a client it should request the event to be deleted with https://github.com/nostr-protocol/nips/blob/master/09.md and only be stored on the client side.

benthecarman:
> I don't think DLC messages should be ephemeral. It would cause issues of clients not being connected to completely miss messages.

Agreed, there isn't really a message type that works well for this however. What mutiny does for nwc is ideal but this isn't standardized. What we do is store the event until it gets a reply from the correct pubkey.

> I think that when an unseen event is received by a client it should request the event to be deleted with [master/09.md](https://github.com/nostr-protocol/nips/blob/master/09.md?rgh-link-date=2024-01-10T16%3A27%3A38Z) and only be stored on the client side.

Most relays don't implement deletion AFIAK and isn't really reliable because someone can just rebroadcast the event, or it might be on another relay.

benthecarman:
> I don't think DLC messages should be ephemeral. It would cause issues of clients not being connected to completely miss messages.
> 
> I think that when an unseen event is received by a client it should request the event to be deleted with [`master`/09.md](https://github.com/nostr-protocol/nips/blob/master/09.md?rgh-link-date=2024-01-10T16%3A27%3A38Z) and only be stored on the client side.

Changed to kind `8,888`, just a normal event. I think this is best, then the user can just timestamp and sync dlc messages from their last known checkpoint.

88.md
```diff
@@ -0,0 +1,146 @@
...
+
+```jsonc
+{
+  "kind": 89,
+  "content": "base64 oracle attestation",
+    "tags": [
+    [
+      "e", // the event id of the announcement
+      "30efed56a035b2549fcaeec0bf2c1595f9a9b3bb4b1a38abaf8ee9041c4b7d93",
+    ],
```
conduition, line 129: [Oracle announcements also contain an `event_id` field](https://github.com/discreetlogcontracts/dlcspecs/blob/master/Messaging.md#oracle_event), and [so do oracle attestations](https://github.com/discreetlogcontracts/dlcspecs/blob/master/Messaging.md#oracle_attestation). Announcement event IDs _must_ be distinct from the Nostr `kind:88` event IDs (due to the fact that the announcement is part of the preimage for the nostr event ID). This might cause some confusion.

Perhaps we should clarify that the `e` tag must be the _Nostr event ID,_ not the _announcement event ID,_ and that the `e` tag should be used to look up the Nostr event which embeds an oracle announcement. 

One might want to enforce on the client that the `oracle_attestation` data embeds the same _announcement event ID_ that you find by looking up the event indicated by the `e` tag. In pseudocode:

```python
def validate_attestation_event(attestation_event):
  oracle_attestation = decode_attestation(attestation_event.content)
  announcement_event = nostr_event_lookup(attestation_event.tags.e)
  oracle_announcement = decode_announcement(announcement_event.content)
  assert oracle_announcement.event_id == oracle_attestation.event_id
```

conduition (commented):
The draft looks great! I'm personally looking forward to moving my oracle to Nostr. This would make DLCs far more easily consumable by clients.

This NIP combined with [the ECash DLCs NUT](https://github.com/cashubtc/nuts/pull/128) could be extremely powerful. 

@benthecarman Are you still working on this? If not I'd be happy to continue the effort.

conduition:
PR to address the above concerns: https://github.com/benthecarman/nips/pull/1

conduition:
@1440000bytes This NIP doesn't try to tell clients how they should settle DLCs. Clients are free to use on-chain, off-chain, side-chain, or even cross-chain settlement layers to buy into and enforce DLCs. The purpose of this NIP is simply to standardize certain Nostr event kinds so that relays can be used as a discovery layer for DLC oracles, and as marketplaces for DLC offers, regardless of the settlement layer they wish to use.

conduition:
This reminds me: @benthecarman what do you think of adding a `settlement` tag to the `kind:30088` event to allow the offerer to specify which settlement layer implementations they support?

```json
{
  "kind": 30088,
  "content": "base64 contract_info",
  "tags": [
    // ...
    [
      "settlement",
      "btc",
      "dlctix",
      "cashu:https://mint.minibits.cash/Bitcoin"
    ]
  ],
  // ...
}
```

conduition:
@benthecarman I submitted [a PR which adds `title` and `description` tags to the announcement event type](https://github.com/benthecarman/nips/pull/2)

conduition:
I'm thinking about this PR and I'm wondering if it'd be smarter to reduce the scope of this NIP to include only the oracle announcement and attestation messages (kind 88, 89). 

There is more work to be done upstream in the DLC specs, improving the efficiency and privacy of DLCs. Right now, for instance, they use ECDSA adaptor signatures instead of Schnorr, and no musig aggregation. We should give upstream specs authors more time.

But [there is a demand for Nostr as a pricing data source](https://github.com/nostr-protocol/nips/pull/1658), and DLC oracles are the most expressive way to communicate that data in a way clients can use for financial transactions. So how about we separate the DLC oracle announcement/attestation event kinds into their own NIP, and leave the offer and encrypted-message event kinds for another spec later? 

This way, people can start building NIP88-compliant oracles, and applications which use NIP88 announcements/attestations for things like [ecash DLCs](https://github.com/cashubtc/nutshell/pull/576), [ticketed DLCs](https://github.com/conduition/dlctix), or [bitcoin-backed loans](https://www.lava.xyz/).

bennyhodl:
> There is more work to be done upstream in the DLC specs - @conduition 
The spec for V1 (ECDSA DLCs) is pretty well defined. And I have a working implementation of nostr DLC messages in [`dlcdevkit`](https://github.com/bennyhodl/dlcdevkit/tree/master/ddk/src/transport/nostr). Maybe add a tag for a V1 ECDSA DLC and then as upstream is updated we can add a V2 tag for Schnorr and musig aggregation.

I started a quick [prototype of Taproot DLC](https://github.com/bennyhodl/taproot-dlc) to eventually move to upstream. Still just use script spend and no key agg.

conduition:
I have a new PR with reduced scope and other improvements here: #1681

PR #922, opened 2023-12-07
Nip07 - add nip44

monlovesmango:
add nip44 calls to NIP07 spec

https://github.com/monlovesmango/nips/blob/NIP07-Add-nip44/07.md

### DIFF
--- a/07.md
+++ b/07.md
@@ -20,9 +20,9 @@ Aside from these two basic above, the following functions can also be implemente
 async window.nostr.getRelays(): { [url: string]: {read: boolean, write: boolean} } // returns a basic map of relay urls to relay policies
 async window.nostr.nip04.encrypt(pubkey, plaintext): string // returns ciphertext and iv as specified in nip-04
 async window.nostr.nip04.decrypt(pubkey, ciphertext): string // takes ciphertext and iv as specified in nip-04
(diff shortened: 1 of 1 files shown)


staab:
I'm personally against it, I don't think it accomplishes anything that a well-optimized nip 07 extension can't.

vitorpamplona:
Get key should be available on the Android Signer and the NIP-46 Signers. I think we should add to nip-07 just for consistency in case somebody is making a single app that uses any of the 3 signers.

pablof7z:
+1

I'm with @vitorpamplona ; consistency is good and you can drive a truck through "well-optimized nip07 extension" 😅

monlovesmango:
I have reopened this in a new PR (github wasn't letting me reopen this one) https://github.com/nostr-protocol/nips/pull/940

NIP07 updates were removed from NIP44 PR so opening this again to come to agreement on how nip07 should be updated

staab:
I haven't been able to track all the changes to this file and NIP 46 that have been proposed, what's the argument for `getKey`? The downside is that an evil client can exfiltrate conversation keys for all DMs, and since we have no forward secrecy or post-compromise security that gives them access to snoop forever.

PR #924, opened 2023-12-08
Fix NIP-46 wording error

s3-odara:
According to the [NIP-26](https://github.com/nostr-protocol/nips/blob/master/26.md) definition, delegation token is a signature of the SHA256 hash of the delegation string, so it is considered unnatural to say that "signed" delegation token.


AsaiToshiya:
Sorry, I'm going to close this due to outdated.

PR #925, opened 2023-12-08
Change NIP-46 App to Client

s3-odara:
The word "App" reminds us of a smartphone application, and **Signer** is also an application, making it difficult to understand the difference between the two.
I suggest that **App** be changed to **Client**, as a similar position is considered a client in NIP-47.

Furthermore, change the word app to application to avoid the word app being associated with a smartphone application.

### DIFF
--- a/46.md
+++ b/46.md
@@ -15,14 +15,14 @@ Entering private keys can also be annoying and requires exposing them to even mo
 
 ## Terms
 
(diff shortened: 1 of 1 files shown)


AsaiToshiya:
Sorry, I'm going to close this due to outdated.

PR #927, opened 2023-12-08
NIP-XX Places

arkin0x:
This PR represents an evolution of the current implementation of Places on https://go.yondar.me. The goal of this PR is to bridge the gap between existing mapping standards (OpenStreetMaps, Schema, etc.) and nostr, enable crowdsourcing for Places, and give a foothold to the usage of geospatial data on nostr.

I'd like to get feedback on this PR and implement that feedback into Yondar; then we will have an implementation justifying a merge.

### DIFF
--- a/34.md
+++ b/34.md
@@ -37,6 +37,27 @@ The `r` tag annotated with the `"euc"` marker should be the commit ID of the ear
 
 Except `d`, all tags are optional.
 
+### Nostr Clone URL format
+
+A `nostr://` URL can be used to reference a repository announcement in a way that is compatible with `git clone` when a [git-remote-nostr](https://git-scm.com/docs/gitremote-helpers) helper is installed:
+
+```
+nostr://<naddr>
+nostr://<npub|nip05>/<identifier>
+nostr://<npub|nip05>/<relay-hint>/<identifier>
+```
+
+Both `<relay-hint>` and `<identifier>` MUST be percent-encoded per [RFC 3986 §2.1](https://datatracker.ietf.org/doc/html/rfc3986#section-2.1). `<relay-hint>` is a relay URL whose `wss://` scheme may be omitted for brevity. `<identifier>` is the `d` tag value of the kind `30617` event.
+
+Examples:
+
+```
+nostr://npub15qydau2hjma6ngxkl2cyar74wzyjshvl65za5k5rl69264ar2exs5cyejr/relay.ngit.dev/ngit
+nostr://npub15qydau2hjma6ngxkl2cyar74wzyjshvl65za5k5rl69264ar2exs5cyejr/my%20%F0%9F%9A%80%20repo
+nostr://danconwaydev.com/ws%3A%2F%2Flocalhost%3A7334/my-local-only-repo
+nostr://danconwaydev.com/relay.ngit.dev/ngit
+```
+
 ## Repository state announcements
 
 An optional source of truth for the state of branches and tags in a repository.
@@ -239,7 +260,7 @@ The event SHOULD include a list of `g` tags with grasp service websocket URLs in
   "content": "",
   "tags": [
     ["g", "<grasp-service-websocket-url>"], // zero or more grasp sever urls
-  ]
+  ],
 }
 ```
 


erskingardner:
I didn't see any mention of dealing with place squatters, is that something that you foresee with events like these or is it possible to create two places events for the same physical location and allow the market to sort of coalesce around the "right" one?

sroertgen:
The NIP is well written and the examples are actually useful.

One suggestion you could further clarify (at least it was not apparent to me): How to handle osm and schema.org values that are objects?

Also you should maybe elaborate (maybe here is enough don't know if it needs to be in the NIP) Why do you want to integrate vocabularies like OSM and schema.org/Place and namespaces you don't even know yet in the NIP and why can't the appropriate attributes be all defined in the NIP?
Should there be a list of known/used namespaces included?

vitorpamplona:
Looks good. The only weird part is this `prop` thing. I wonder if you couldn't just dump all osm properties as Nostr tags directly. 

I'd be careful with the references to pubkeys "owning" places. Many people can own the same place. I don't know how to identify the actual owner of the place.

arkin0x:
> I didn't see any mention of dealing with place squatters, is that something that you foresee with events like these or is it possible to create two places events for the same physical location and allow the market to sort of coalesce around the "right" one?

I imagined that the solution to this is about the same as the solution to multiple pubkeys claiming to be the same person. You need to discern the legit pubkey, and NIP-05 can help. If you follow the real pubkey, you'll see the real place on a map of your contacts/friends. In cases where there isn't a single owner, or the owner isn't on nostr, this discernment can be aided by reviews, proof-of-work, social graph reputation, likes, or any number of ways. So, yes coalesce, but using mechanisms of the nostr protocol.

However, I'm open to the idea that a kind 1754 event could have a tag that says "this Place is a duplicate of this other Place". These kinds of reports could be made and clients could interpret this. Some clients could have a "related Place" section in the Place details and list the *real* Place for others to find it easier. Interpreting these reports could be up to the client.

arkin0x:
> The NIP is well written and the examples are actually useful.
> 
> One suggestion: You could further clarify (at least it was not apparent to me): How to handle osm and schema.org values that are objects?

I've addressed this with 3404f89

> Also you should maybe elaborate (maybe here is enough don't know if it needs to be in the NIP) Why do you want to integrate vocabularies like OSM and schema.org/Place and namespaces you don't even know yet in the NIP and why can't the appropriate attributes be all defined in the NIP? Should there be a list of known/used namespaces included?

I've received a lot of feedback privately and in the Yondar telegram group that OpenStreetMaps ought to be compatible with whatever geospatial solution we create for nostr.  Indeed, OSM is a very large, very active, and very high quality data set that can be leveraged to bring high quality Place data to nostr, and perhaps even bring nostr data to OSM when appropriate.

OSM uses a "folksonomy" approach to applying attributes to geospatial data. The result is something akin to "the wild west". There are conventions that everyone tends to follow with surprising consistency, but anything goes.

From https://taginfo.openstreetmap.org
image

If you take a moment to digest the wide variety of terms in use here, it's easy to see how things you didn't even think about could suddenly become much more important than you anticipated. Being that Places are so new, I am certainly not confident that I can anticipate how people will use them in the future.

However, I'm not opposed to providing some basic defaults that people could utilize so that we have the fundamentals covered, ex. how do you give your place a name, how do you give it an address, how do you give it a phone number, etc. That would be useful.

arkin0x:
> Looks good. The only weird part is this `prop` thing. I wonder if you couldn't just dump all osm properties as Nostr tags directly.

If you did this, it would be hard to get a list of tags from an event that were props and not-props. At least in this way the props are namespaced and don't pollute. I figured if each tag should have a purpose, then prop's purpose is to corral key=value properties.

> I'd be careful with the references to pubkeys "owning" places. Many people can own the same place. I don't know how to identify the actual owner of the place.

I'd be happy to clarify in the NIP that I'm referring to ownership of the nostr event, not necessarily the physical place.

Yes, identifying the owner of a real place based on a kind 37515 is at least as difficult as determining if the owner of Michael Saylor's pubkey is actually Michael Saylor. But as I mentioned [above](https://github.com/nostr-protocol/nips/pull/927#issuecomment-1848101117), there are ways to help with this and tame the global feed of places, and some of these techniques have yet to be tried.

sroertgen:
This NIP seems to have parts that are accepted (`content`, `d`, `admin`, `g`) and parts that lead to discussion and/or refusal, the `prop` part.
It raises some of the same issues we saw in recent [[NIP-32 disccusion](https://github.com/nostr-protocol/nips/pull/878)](https://github.com/nostr-protocol/nips/pull/878). The arguments are more or less the same: NIPs should have well defined attributes, so clients can know what to expect and handle appropriately. The arguments of people wanting the openness are also quite the same: We don't know yet, what people will use and want to be open for things we can't anticipate.  
  
What if we would separate concerns here: NIP-44 really just does the geo coordinate part, no ones seems to argue about, but the annotation part is removed.  
  
But we want annotation of places. So what about a way to synthesize the want for openness and the need for structure? We make another event kind that: 

- references an event
- provides a validation schema
- and gives the data for annotation

sth like

  ```js
  {
    kind: 424242,
    content: "stringified data be it json, xml (🤢) , whatever",
    tags: [
      ["a", "1234-referenced-event-id..."],
      ["s", "json-schema/xml-schema/whateverelse-schema", "https://uri-to-the-schema"]
    ],
   ...
  }
  ```

This enables to annotate arbitrary events with structured data. Clients could query for schemas they understand.

The place to discuss about appropriate attributes would no longer be part of a NIP discussion for all kinds of "Other Stuff" people want to build on Nostr. This would reduce the burden of discussion from Nostr CEOs and put it outside. At the same time it would be evident for everyone, what kind of schema one is using for the data and the schema is at best well annotated or accompanied by a spec. Better schemas will find more reuse.

So Yondar would go forward and define a schema, or reuse an existing one (don't know if OSM provides sth like that). The NIP could hold references to clients using schemas for specific objects types (e.g. Places).

fiatjaf:
Why not always use the same schema and then everybody can read all events?

paulmillr:
I have naming concern: NIP44, which is currently popular in E2EE DM context, has been going through the audit, and some packages for programming languages have also been published for this purpose. See #746. There are a bunch of NIP44 events on relays already.

fiatjaf:
Yes, this NIP must pick a different name, please.

arkin0x:
> What if we would separate concerns here: NIP-44 really just does the geo coordinate part, no ones seems to argue about, but the annotation part is removed.
> 
> But we want annotation of places. So what about a way to synthesize the want for openness and the need for structure?

@sroertgen I appreciate you cutting through the noise to the heart of the issue here. I think you're right.

Funny enough, this NIP was originally a proposal for _only_ annotation with key=value pairs (nothing to do with Places), and it was shot down immediately when I sent it to some nostr dads. It was shorter, simpler, but I was told that we shouldn't add a `prop` tag for every event kind. So, I rewrote the whole thing to be a Places-specific thing.

We could store all the key=value properties in the `properties` attribute in the stringified GeoJSON (this is what Yondar currently does) but if we want crowdsourcing/annotation, which I think is the ultimate goal, then we shouldn't stuff properties into the GeoJSON and do them "the nostr way".

This NIP does specify the kind `1754` event as an annotation event specifically for Places. Does that fulfill your suggestion for an annotation event?

---

By the way, your 🤮 emoji made me laugh... everyone says we don't like stringified JSON/XML in the content, so putting stuff in tags is the alternative. JSON is key=value, so I am surprised that the non-indexed key=value tag I've proposed has been met with such resistance. We either stringify key=value and put it in the content, or we put it into a non-indexed tag. I vote for the tag approach.

---

> Why not always use the same schema and then everybody can read all events?

@fiatjaf There are several different schemas used by mapping people, and I've been petitioned to support them. This allows conventions to come across to nostr, as well as the importing of select data from other sources. I thought that NIP-32's usage of namespaces was a model I could follow here to enable that flexibility. I think it would be a good idea to add some basic keys like `name` and `phone` etc. for people to get started, but some have suggested we simply define a new schema for nostr and I feel like that's not possible, because the needs cannot be anticipated. I like to reference this image from OpenStreetMaps to convey how intractable this is -- these are the popular keys on OpenStreetMaps and many of them I would have never even thought of. It just shows how varied the needs can be for these schemas: 
image

Of course, if someone wants to put forth a schema I'm certainly open to it, but trying to manage that could be its own job and I don't think nostr needs to take on that work.

Maybe the answer is to just define a basic set of properties that require client support, and allow for clients to support additional properties/schemas.

---

@paulmillr Thank you Paul for the details. I don't care about the number, so I can change this to XX until we get to the merge stage.

arkin0x:
The kitchen schema example has been confusing so I'm going to remove it and replace it with something more applicable to Places. The point was to demonstrate how to translate a schema to nostr tags when the schema isn't flat.

staab:
> nostr dads

😆

The "generic triples" thing was discussed [here](https://github.com/nostr-protocol/nips/pull/878). The argument against was summarized well by @vitorpamplona [here](https://github.com/nostr-protocol/nips/pull/878#issuecomment-1810597033) — generic key/value data is a One Ring to confuse them all, and will tempt developers to think they're supporting interoperability by putting whatever garbage they want in triples.

I think [my point](https://github.com/nostr-protocol/nips/pull/878#issuecomment-1810548840) still stands that we need a way to invoke external nomenclatures. Maybe it could be done by specifying some special tag name per use case. So instead of  `prop`, maybe use `schema/Place` as the tag name. This way you don't have to re-define the entire vocabulary in the NIP, but only specific use cases are defined.

fiatjaf:
> Maybe the answer is to just define a basic set of properties that require client support, and allow for clients to support additional properties/schemas.

Why not just take OpenStreetMaps's schema and use that as the Nostr schema then?

If you say "we support all these 4 different schemas" that doesn't really mean we support all these 4 different schemas, it means some people implement some, others implement others, some implement both and most people get confused as hell and implement nothing and the idea dies.

If you are expecting many people to write clients that implement more than one schema and you are ok with that why wouldn't it be more reasonable to force the people who are publishing the data to conform to one schema?

It's much easier for the person publishing the data to crunch their data to a specific format then to condemn everybody else for saecula saeculorum to navigate a sea of mess.

staab:
> Why not just take OpenStreetMaps's schema and use that as the Nostr schema then?

Because I assume the reason there are multiple schemas that do the same thing is that there are trade-offs between them. But maybe this is one of those content-type issues, where we should specify which schema should be used, and if someone wants a different schema, they can create a new kind.

arkin0x:
I think the points raised by @fiatjaf and @staab (and Vitor's explanation) make a strong case for having a well defined schema specifically for nostr. I am willing to give this a shot. I may get slapped around by the map people but we'll see how it goes!

arkin0x:
> Why not just take OpenStreetMaps's schema and use that as the Nostr schema then?

@fiatjaf the OpenStreetMaps schema is "any string as a key = any string as a value", so that won't really work 😁

arkin0x:
I think the points raised by @fiatjaf and @staab (and Vitor's explanation) make a strong case for having a well defined schema specifically for nostr. I am willing to give this a shot. I may get slapped around by the map people but we'll see how it goes!

arkin0x:
Help us define the Place properties that nostr should use: 

https://docs.google.com/forms/d/e/1FAIpQLScF2LYr9RIYi4-YFKNFaaYH3YYTdUo09j6RRGVjf992wNC1EQ/viewform

Results:

https://docs.google.com/spreadsheets/d/1r7mYftFUgED1VTvTaIv790a2JKXnfVDt9T0_INxzgd8/edit?usp=sharing

vinaylalco:
I look forward to having a consistent NIP for location based data on Nostr. I built mapstr.xyz and would start to use this if merged. There are multiple Nostr mapping apps around atm and it would be good to converge on a common NIP and build up Nostr geographic related data. 

I thought there was a great deal of thought that went into the structure of the NIP and dealing with geographic based data. Well done!

arkin0x:
@vinaylalco Thank you!

If I am understanding your question correctly, you are wondering if you could discern which client produced a kind 37515 event. This NIP does not specify anything about that.

I don't know if any clients try to leave their mark on the events produced by their users. I could see this potentially being undesirable to users.

If you were to leave a distinguishing mark on a nostr event to indicate which client produced it, you would probably have to use a tag. I'm not aware of any standard tags to do this.

arkin0x:
@vinaylalco I would invite you to submit your recommended properties to the form linked in https://github.com/nostr-protocol/nips/pull/927#issuecomment-1852610873 so that this NIP can support the properties you desire for Places to have. Thank you, and nice work on mapstr.xyz! 👾💜

AsaiToshiya:
> If you were to leave a distinguishing mark on a nostr event to indicate which client produced it, you would probably have to use a tag. I'm not aware of any standard tags to do this.

It is probably a `client` tag: https://github.com/nostr-protocol/nips/blob/master/89.md#client-tag

vinaylalco:
> @vinaylalco Thank you!
> 
> If I am understanding your question correctly, you are wondering if you could discern which client produced a kind 37515 event. This NIP does not specify anything about that.
> 
> I don't know if any clients try to leave their mark on the events produced by their users. I could see this potentially being undesirable to users.
> 
> If you were to leave a distinguishing mark on a nostr event to indicate which client produced it, you would probably have to use a tag. I'm not aware of any standard tags to do this.

A tag can probably handle it yeah and as you point out there is a tag for this already. My thinking on this one was that maps often have layers of data and it may be desirable to have a a layer for each client app so the user could toggle data from diff location related Nostr apps. It was one use case I'm considering over at mapstr anyhow but yeah a tag will be fine to handle it as you can filter on this basis. It could be an optional tag the user could add/remove as desired to alleviate any privacy user concerns.

fiatjaf:
> @fiatjaf the OpenStreetMaps schema is "any string as a key = any string as a value", so that won't really work 😁

Well, Nostr also allows any key, but I bet OSM has a set of core properties that they actually use in the UI, no?

dadofsambonzuki:
Sorry, late to the party here.

I think an _optional_ OSM ID tag would be of use. e.g.

`"osm_id","node|way:5843771398"`

Note both 'node' and 'way' are possible. As is 'relation', but I'm not sure we need this although it might be better to include for completeness. More info about [OSM Elements here](https://wiki.openstreetmap.org/wiki/Elements).

This would enable nostr clients to additionally poll OSM for more data or to refresh the data, which is under constant maintenance by the OSM community. This additional information could then be published by `kind 1754` as discussed or as a replacement `kind 37515`. There could in theory be bridges back to OSM, but that is beyond the scope of this discussion.

We need to decide what happens if any tags in a `kind 1754` conflict with the tags in the base `kind 37515`. I suggest latest wins as long it is from an authorized npub.

As discussed, there is no OSM schema as such, just [consensus on commonly used tags](https://wiki.openstreetmap.org/wiki/Tags). These evolve over time, sometimes have overlapping meaning but tend towards consensus. I suggest we embrace the chaos (it's actually pretty stable) and follow the same conventions.

I think we should consider the inclusion of *any* of the OSM tags (which is any key-value pair in theory, but following the OSM consensus in practice) as `Optional Properties` in the `kind 37515` or as Place labels in `kind 1754`.  No need to re-invent the wheel here. If this is adopted then we should obviously use the OSM naming conventions e.g. `contact:website` and not `website`.

dadofsambonzuki:
> I didn't see any mention of dealing with place squatters, is that something that you foresee with events like these or is it possible to create two places events for the same physical location and allow the market to sort of coalesce around the "right" one?

I think this comes down to WoT as @arkin0x says. With the OSM link discussed above, this would allow a key-pair to 'claim' an OSM location such as a merchant. Like Google Places, but not permissioned.

I have been working on an attestation concept, whereby other key-pairs (including oracles) can signal something to be true, perhaps by gift wrapping a note. More on that soon.

Clearly it's down to each user to weigh those attestations accordingly.

dadofsambonzuki:
How do we get a NIP number 'assigned' this so we cab actually refer to it?

dadofsambonzuki:
It might be worth considering [NIP-73 (External Content IDs)](https://github.com/nostr-protocol/nips/blob/744bce8fcae0aca07b936b6662db635c8b4253dd/73.md) for the references out to OSM (and perhaps other geographic data sources such as Google Maps and [Wikidata.org](https://www.wikidata.org).)

DestBro:
What is the status of this ?

dadofsambonzuki:
Languishing and undiscoverable really. A few new apps have started rolling there own NIPs for very custom usecases (like listing skateparks) that should really be leveraging this and adding to it.

I spoke with @arkin0x on this a while back and agreed I'd publish it on NostrHub for better visibility. I'll get onto that ...

DestBro:
I have a use case to map an OSM POI to Nostr (Google Maps for Nostr). Basically, an OpenStreetMap POI → Nostr bridge. That way, later I’d be able to mention, chat about, or put an OSM POI into lists.

So far, this merge request aims to solve this problem, right?

I know there are apps like Yondar, BTCMap, Satlantis, etc. that have tried to solve this problem. Now I’m here as well, and I’m not sure whether I should develop something custom too or not

DestBro:
@dadofsambonzuki Are you on Telegram? I’d like to connect - looking for someone with OSM + Nostr experience for occasional chats. Are you a member of Nostr chat group https://t.me/nostr_offtopic ? I discussed with Constant this problem last months, did a brainstorming in the group

DestBro:
For example would make sense to have a feature to query "osm:way:12345678" over Nostr, and get:
- opening hours
- photos with the place, published by nostr users
- comments, published by nostr users
- etc

dadofsambonzuki:
> For example would make sense to have a feature to query "osm:way:12345678" over Nostr, and get:
> 
> * opening hours
> * photos with the place, published by nostr users
> * comments, published by nostr users
> * etc

That would be a good DVM or API.

I am about to start working on an Nostr-OSM bridge.

leonacostaok:
Hey @dadofsambonzuki, can we share thoughts on your Nostr-OSM bridge?

I've made reviews+ratings on [Mapping Bitcoin](https://mappingbitcoin.com), but I don't like my solution. So, I designed something new, and I'd like some feedback:

# Dead Anchor + Platform Attestation
## Nostr ↔ OSM Bridge

---

## Concept

A **dead anchor** is a Nostr keypair generated once, used to publish a single event pointing to an external canonical identifier (OSM node, ISBN, etc.), then the nsec is discarded. The resulting npub becomes a permanent, ownerless public wall — anyone can post *to* it, no one can post *as* it.

The **platform** (MappingBitcoin) acts as the trust anchor by signing attestation events. No shared keys. No signer infrastructure.

---

## Architecture

```
OSM Node (canonical data)
        │
        ▼
Dead Anchor npub  ←──── generated once, nsec trashed
        │
        ├── Comments / Reviews (tagged by anyone's npub)
        ├── Venue owner posts (their own npub, tagged to OSM ID)
        └── Platform attestation (MappingBitcoin npub signs verification)
```

---

## Event Structure

**1. Anchor Event** (published once, then nsec trashed)
```json
{
  "kind": 30000,
  "pubkey": "<dead-anchor-npub>",
  "tags": [
    ["d", "osm:node:1234567890"],
    ["r", "https://www.openstreetmap.org/node/1234567890"]
  ],
  "content": ""
}
```

**2. Platform Attestation Event** (signed by MappingBitcoin npub)
```json
{
  "kind": 30382,
  "pubkey": "<mappingbitcoin-npub>",
  "tags": [
    ["d", "osm:node:1234567890"],
    ["p", "<venue-owner-npub>"],
    ["method", "domain"],
    ["status", "verified"]
  ],
  "content": ""
}
```

**3. Review / Comment Event** (signed by any user's npub)
```json
{
  "kind": 38381,
  "pubkey": "<reviewer-npub>",
  "tags": [
    ["d", "osm:node:1234567890"],
    ["rating", "4"]
  ],
  "content": "Great place, solid espresso."
}
```

---

## Role Table

| Actor | Key | Role | Posts as |
|---|---|---|---|
| OSM | — | Canonical data source | — |
| Dead anchor | npub only (nsec trashed) | Permanent identifier | Nobody |
| MappingBitcoin | Live npub | Trust authority / attestor | Itself |
| Venue owner | Their own npub | Responds, updates | Themselves |
| Reviewer | Their own npub | Comments, ratings | Themselves |

---

## Tradeoff Table

| Dimension | Dead Anchor + Attestation | NIP-46 (Remote Signer) | NIP-26 (Delegation) |
|---|---|---|---|
| **Complexity** | Low | High | Medium |
| **Decentralization** | High | Medium | Medium |
| **Cost** | Near zero | Server infra | Near zero |
| **Time to build** | Days | Weeks | Days–week |
| **Censorship resistance** | High | Low (signer dependency) | Medium |
| **Key custody risk** | None | High | Low |
| **Ecosystem support** | Native | Active | Largely abandoned |
| **Multi-owner** | Attestation whitelist | Signer auth logic | Token per owner |

---

## Verification Flow

```
1. Venue owner contacts MappingBitcoin (email / domain / phone)
2. MappingBitcoin verifies off-chain
3. MappingBitcoin publishes attestation event (kind 30382) linking owner npub → OSM ID
4. Anyone can verify: "did MappingBitcoin vouch for this npub?"
5. Venue owner posts responses/updates using their own npub, tagged to OSM ID
```

No nsec shared. No signer infra. Platform is the CA, not the custodian.

---

## Why This Works

- **OSM owns the data** — addresses, coordinates, hours live there, maintained by the community
- **Nostr owns the social layer** — reviews, verifications, responses are sovereign and portable
- **Dead anchor** = immutable bridge between the two, costs nothing to maintain
- **Platform attestation** = trust without custody

> The venue doesn't need an identity. The social graph around it does.

---

Let me know your thoughts!

raucao:
> @arkin0x [force-pushed](https://github.com/nostr-protocol/nips/compare/4da44f8a5acc40cd6fc73abf22da3698cc50e79d..0a25dd524389d642ec153d9009a16fb183eb23ce) the master branch from [4da44f8](https://github.com/nostr-protocol/nips/commit/4da44f8a5acc40cd6fc73abf22da3698cc50e79d) to [0a25dd5](https://github.com/nostr-protocol/nips/commit/0a25dd524389d642ec153d9009a16fb183eb23ce) [2 hours ago](https://github.com/nostr-protocol/nips/pull/927#event-24590895741)

Is GitHub hallucinating or did someone actually force-push to master?

PR #933, opened 2023-12-16
Add read status NIP

staab:
This is simple, and granular. It uses `e` tags, not `a` tags to allow clients to differentiate versions of replaceable events. It allows for expiration and tagging multiple events at once to reduce storage requirements. It's not a replaceable list, so there won't be conflicts between writes. It allows for public and private read receipts, so it covers DM/group chat read receipts as well as user notification read status.

### DIFF
--- a/97.md
+++ b/97.md
@@ -1,0 +1,27 @@
+NIP-97
+======
+
+Read Status
+-----------
+
+`draft` `optional`
+
+Clients may wish to track whether certain events have been seen by a user, including mentions, replies, likes, messages, community posts, and more.
+
+A kind `15` event may be used to mark one or more `e` tags as read.
+
+```json
+{
+  "kind": "15",
+  "tags": [
+    ["expiration", "<one year in the future>"],
+    ["e", "<event1_id>"],
+    ["e", "<event1_id>"]
+  ]
+}
+```
+
+Events MAY be wrapped as described in [NIP 59](./59.md) in order to create private read status events. These can be useful for preserving user privacy, or to selectively reveal read status to other users, for example in a group chat or message thread.
+
+Read status events SHOULD expire after a sufficient amount of time to save storage space, since very old events can be assumed to have been seen. Wrapped events SHOULD put the `expiration` tag on the wrapper event. See [NIP 40](./40.md) for more information expiration.
+
--- a/README.md
+++ b/README.md
@@ -85,6 +85,7 @@ They exist to document what may be implemented by [Nostr](https://github.com/nos
 - [NIP-92: Media Attachments](92.md)
 - [NIP-94: File Metadata](94.md)
 - [NIP-96: HTTP File Storage Integration](96.md)
+- [NIP-97: Read Status](97.md)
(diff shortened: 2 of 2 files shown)


jb55:
nice!

basantagoswami:
We can also use a similar scheme to save drafts of all kinds of events can't we?

staab:
Coracle now supports this for marking notifications as read. I currently use NIP 59 to wrap the events. If the user's signer doesn't support NIP 44 I fall back to encrypting with NIP 04 for now.

97.md
```diff
@@ -0,0 +1,27 @@
...
+
+`draft` `optional`
+
+Clients may wish to track whether certain events have been seen by a user, including mentions, replies, likes, messages, community posts, and more.
+
+A kind `15` event may be used to mark one or more `e` tags as read.
+
+```json
+{
+  "kind": "15",
```
fiatjaf, line 15:
```suggestion
  "kind": 15,
```

fiatjaf:
Is this implemented anywhere?

fiatjaf:
Do you think this could be used for the "counting views" flow like what is described in https://github.com/nostr-protocol/nips/pull/923 @staab?

staab:
> Is this implemented anywhere?

Coracle uses it for keeping track of notifications that have been seen. I'm still trying to decide if it's a good idea, because small bugs can result in loops that create many copies of the same event. These all have to be decrypted (since notification read status are private), which is a lot of load on clients.

> Do you think this could be used for the "counting views" flow like what is described in https://github.com/nostr-protocol/nips/pull/923 @staab?

Maybe, it depends on what the semantics are. But probably not, because it could end up confusing one use case with another (I saw that I was mentioned in this video, vs I watched some arbitrary threshhold of the video).

vitorpamplona:
I tested this in batches of 100 events. It works well. Two issues: 
- It uses a lot of memory to represent a boolean. The phone has to get rid of them as fast as possible. 
- Private events should not have these. 

In the end, I reverted it to a route-based system that simply stores the `.created_at` of the newest post visible on the screen or the phone's local time of the user is at the top of each feed. It's not as granular, but it is quite private and compact, and solves most of the need.

staab:
> simply stores the .created_at of the newest post

This is what I used to do, and it might be the better option. We should specify where to store this information in an interoperable way, which was the point of this NIP. Clients can handle the out-of-order edge case if desired by storing event ids locally.

vitorpamplona:
I think we all represent routes slightly differently (like I use Java's hash integer of the event_ids in a group DM to Route to the group), but it could work.

staab:
Yeah, we could also break it down pretty granularly with canonical names, e.g. `mentions_last_checked` but there would probably always be views that departed from our expectations 🤔

staab:
After having this implemented for about 6 months, I'm abandoning it because it creates a whole lot of 1059-wrapped events that I don't want to download when asking for DMs. I created another attempt based on replaceable events, and only encrypting the content here: #1405.

PR #939, opened 2023-12-20
Clean up NIP 44

staab:
- Clarify separation of concerns (encryption vs messaging)
- Improve formatting
- Clarify encryption/decryption steps
- Re-order sections to be more welcoming to less technical readers
- Clarify that this is not a drop in replacement for NIP 04

One thing to note is I took out the language about putting an encrypted payload into `content` specifically, since an encrypted payload could also be placed in `tags` (probably a bad idea in practice, but up to other NIPs to specify). The important thing is that the payload is part of the event's hash and authenticated by its signature.

### DIFF
--- a/44.md
+++ b/44.md
@@ -10,6 +10,10 @@ The NIP introduces a new data format for keypair-based encryption. This NIP is v
 to allow multiple algorithm choices to exist simultaneously. This format may be used for
 many things, but MUST be used in the context of a signed event as described in NIP 01.
 
+*Note*: this format DOES NOT define any `kind`s related to a new direct messaging standard,
+only the encryption required to define one. It SHOULD NOT be used as a drop-in replacement
+for NIP 04 payloads.
+
 ## Versions
 
 Currently defined encryption algorithms:
@@ -30,7 +34,7 @@ event. When applying this NIP to any use case, it's important to keep in mind yo
 model and this NIP's limitations. For high-risk situations, users should chat in specialized E2EE
 messaging software and limit use of nostr to exchanging contacts.
 
-On its own, messages sent using this scheme has a number of important shortcomings:
+On its own, messages sent using this scheme have a number of important shortcomings:
 
 - No deniability: it is possible to prove an event was signed by a particular key
 - No forward secrecy: when a key is compromised, it is possible to decrypt all previous conversations
--- a/44.md
+++ b/44.md
@@ -1,42 +1,14 @@
-# NIP-44
+NIP-44
+=====
 
-## Encrypted Payloads (Versioned)
+Encrypted Payloads (Versioned)
+------------------------------
 
 `optional`
 
 The NIP introduces a new data format for keypair-based encryption. This NIP is versioned
-to allow multiple algorithm choices to exist simultaneously.
-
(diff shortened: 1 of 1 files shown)


fiatjaf:
LGDR: Looks Good, Didn't Read

vitorpamplona:
OMG!!! @fiatjaf is just approving stuff without reading it! Looks like a politician 🤣

vitorpamplona:
I would move Limitations and the first section of V2 to near the end of the file as if it were an appendix. 

Give them the goods right away.

paulmillr:
@vitorpamplona perhaps, but those are limitations of v2 and not of some future v3 - which may be quantum-resistant and forward-secure.

paulmillr:
lgtm

staab:
90% of people who read this are not going to care about the details, just that it exists and they can use it. Limitations is going to be much more relevant for them, since pretty much all an application developer can do with this spec is misuse it or not misuse it. I also think it's helpful to have the context in mind when getting into the weeds to orient yourself.

To @paulmillr's point I put the limitations above the v2 heading because they currently apply to all versions, and many will always apply because they're limitations of nostr's architecture, not the encryption scheme. If we end up with a quantum-resistant v3, we can add a note under that limitation, or re-organize.

vitorpamplona:
We should merge this and keep going.

44.md
```diff
@@ -1,74 +1,150 @@
...
 
 The NIP introduces a new data format for keypair-based encryption. This NIP is versioned
-to allow multiple algorithm choices to exist simultaneously.
+to allow multiple algorithm choices to exist simultaneously. This format may be used for
+many things, but MUST be used in the context of a signed event as described in NIP 01.
 
-Nostr is a key directory. Every nostr user has their own public key, which solves key
-distribution problems present in other solutions. The goal of this NIP is to have a
-simple way to send messages between nostr accounts that cannot be read by everyone.
+*Note*: this format DOES NOT define any `kind`s related to a new direct messaging standard,
```
jklein24, line 13: @staab apologies for commenting on a very old PR like this, but I'm curious if you could provide guidance on *why* NIP-44 shouldn't be used as a drop-in replacement for NIP-04.

I'm investigating the best way to replace the NIP-04 usage in NIP-47 (NWC), and I would love to understand the tradeoffs here (should messages be gift wrapped, etc).

vitorpamplona: There is ambiguity between NIP-04, the encryption scheme, and NIP-04, the DM kind. NIP-44 is a replacement for the encryption scheme, not the DM scheme. 

It should not be replaced in `kind:4` DMs because of all the metadata leaks that are incompatible with the DM use case. On NIP-47, you can just use NIP-44 like you have been using NIP-04 encryptions.

staab: Right, nip-04-type dms with nip44 encryption would also be incompatible with all other messaging specs. Using nip 44's encryption for NWC/NC is encouraged, with the caveat that decryption support should happen before people start encrypting messages with nip 44 to avoid interoperability problems.

jklein24: That makes sense, thank you! I'm also thinking through how a migration path might work here.

PR #940, opened 2023-12-20
NIP07 - add nip44 calls

monlovesmango:
add nip44 calls to NIP07 spec

https://github.com/monlovesmango/nips/blob/NIP07-Add-nip44/07.md

@paulmillr does the version need to be added to these calls?

### DIFF
--- a/07.md
+++ b/07.md
@@ -20,6 +20,9 @@ Aside from these two basic above, the following functions can also be implemente
 async window.nostr.getRelays(): { [url: string]: {read: boolean, write: boolean} } // returns a basic map of relay urls to relay policies
 async window.nostr.nip04.encrypt(pubkey, plaintext): string // returns ciphertext and iv as specified in nip-04 (deprecated)
 async window.nostr.nip04.decrypt(pubkey, ciphertext): string // takes ciphertext and iv as specified in nip-04 (deprecated)
+async window.nostr.nip44.getKey(pubkey): string // takes third party pubkey and returns shared secret key as specified in nip-44
+async window.nostr.nip44.encrypt(pubkey, plaintext): string // takes third party pubkey, plain text and returns ciphertext as specified in nip-44
+async window.nostr.nip44.decrypt(pubkey, ciphertext): string // takes third party pubkey, ciphertext and returns plain text as specified in nip-44
 ```
 
 ### Implementation
--- a/07.md
+++ b/07.md
@@ -20,9 +20,10 @@ Aside from these two basic above, the following functions can also be implemente
 async window.nostr.getRelays(): { [url: string]: {read: boolean, write: boolean} } // returns a basic map of relay urls to relay policies
 async window.nostr.nip04.encrypt(pubkey, plaintext): string // returns ciphertext and iv as specified in nip-04 (deprecated)
 async window.nostr.nip04.decrypt(pubkey, ciphertext): string // takes ciphertext and iv as specified in nip-04 (deprecated)
-async window.nostr.nip44.getKey(pubkey): string // takes third party pubkey and returns shared secret key as specified in nip-44
-async window.nostr.nip44.encrypt(pubkey, plaintext): string // takes third party pubkey, plain text and returns ciphertext as specified in nip-44
-async window.nostr.nip44.decrypt(pubkey, ciphertext): string // takes third party pubkey, ciphertext and returns plain text as specified in nip-44
+async window.nostr.nip44.v2.encrypt([[pubkey1, plaintext1], [pubkey2, plaintext2], ...]): string[] // takes array of [pubkey, plaintext] pairs, returns array of ciphertexts as specified in nip-44
+async window.nostr.nip44.v2.decrypt([[pubkey1, ciphertext1], [pubkey2, ciphertext2], ...]): string[] // takes array of [pubkey, ciphertext] pairs, returns array plaintexts as specified in nip-44
+async window.nostr.nip44.v2.getConversationPublicKey([pubkey1, pubkey2, ...]): string[] // takes array of pubkeys, returns array of conversation public keys
+async window.nostr.nip44.v2.signEventWithConversationKey(pubkey, event: { created_at: number, kind: number, tags: string[][], content: string }): Event // takes a pubkey and event object, adds `id`, `pubkey` and `sig` using the conversation key as the private key and returns it
 ```
 
 ### Implementation
--- a/07.md
+++ b/07.md
@@ -20,7 +20,7 @@ Aside from these two basic above, the following functions can also be implemente
 async window.nostr.getRelays(): { [url: string]: {read: boolean, write: boolean} } // returns a basic map of relay urls to relay policies
 async window.nostr.nip04.encrypt(pubkey, plaintext): string // returns ciphertext and iv as specified in NIP-04 (deprecated)
 async window.nostr.nip04.decrypt(pubkey, ciphertext): string // takes ciphertext and iv as specified in NIP-04 (deprecated)
-async window.nostr.nip44.getConversationKey(pubkey): string // takes pubkey, returns shared conversation key as specified in NIP-44
+async window.nostr.nip44.getConversationKeys([[pubkey1], [pubkey2], ...]): string[] // takes array of [pubkey] tuple, returns array of shared conversation keys as specified in NIP-44
 async window.nostr.nip44.encrypt([[pubkey1, plaintext1], [pubkey2, plaintext2], ...]): string[] // takes array of [pubkey, plaintext] tuples, returns array of ciphertexts as specified in NIP-44
 async window.nostr.nip44.decrypt([[pubkey1, ciphertext1], [pubkey2, ciphertext2], ...]): string[] // takes array of [pubkey, ciphertext] tuples, returns array of plaintexts as specified in NIP-44
 ```
--- a/07.md
+++ b/07.md
@@ -22,8 +22,6 @@ async window.nostr.nip04.encrypt(pubkey, plaintext): string // returns ciphertex
 async window.nostr.nip04.decrypt(pubkey, ciphertext): string // takes ciphertext and iv as specified in nip-04 (deprecated)
 async window.nostr.nip44.encrypt([[pubkey1, plaintext1], [pubkey2, plaintext2], ...]): string[] // takes array of [pubkey, plaintext] tuples, returns array of ciphertexts as specified in nip-44
 async window.nostr.nip44.decrypt([[pubkey1, ciphertext1], [pubkey2, ciphertext2], ...]): string[] // takes array of [pubkey, ciphertext] tuples, returns array of plaintexts as specified in nip-44
-async window.nostr.nip44.getConversationPublicKeys([pubkey1, pubkey2, ...]): string[] // takes array of pubkeys, returns array of conversation public keys
-async window.nostr.nip44.signEventWithConversationKey(pubkey, event: { created_at: number, kind: number, tags: string[][], content: string }): Event // takes a pubkey and event object, adds `id`, `pubkey` and `sig` using the conversation key as the private key and returns it
 ```
 
 ### Implementation
--- a/07.md
+++ b/07.md
@@ -18,10 +18,11 @@ async window.nostr.signEvent(event: { created_at: number, kind: number, tags: st
 Aside from these two basic above, the following functions can also be implemented optionally:
 ```
 async window.nostr.getRelays(): { [url: string]: {read: boolean, write: boolean} } // returns a basic map of relay urls to relay policies
-async window.nostr.nip04.encrypt(pubkey, plaintext): string // returns ciphertext and iv as specified in nip-04 (deprecated)
-async window.nostr.nip04.decrypt(pubkey, ciphertext): string // takes ciphertext and iv as specified in nip-04 (deprecated)
-async window.nostr.nip44.encrypt([[pubkey1, plaintext1], [pubkey2, plaintext2], ...]): string[] // takes array of [pubkey, plaintext] tuples, returns array of ciphertexts as specified in nip-44
-async window.nostr.nip44.decrypt([[pubkey1, ciphertext1], [pubkey2, ciphertext2], ...]): string[] // takes array of [pubkey, ciphertext] tuples, returns array of plaintexts as specified in nip-44
+async window.nostr.nip04.encrypt(pubkey, plaintext): string // returns ciphertext and iv as specified in NIP-04 (deprecated)
+async window.nostr.nip04.decrypt(pubkey, ciphertext): string // takes ciphertext and iv as specified in NIP-04 (deprecated)
+async window.nostr.nip44.getConversationKey(pubkey): string // takes pubkey, returns shared conversation key as specified in NIP-44
+async window.nostr.nip44.encrypt([[pubkey1, plaintext1], [pubkey2, plaintext2], ...]): string[] // takes array of [pubkey, plaintext] tuples, returns array of ciphertexts as specified in NIP-44
+async window.nostr.nip44.decrypt([[pubkey1, ciphertext1], [pubkey2, ciphertext2], ...]): string[] // takes array of [pubkey, ciphertext] tuples, returns array of plaintexts as specified in NIP-44
 ```
 
 ### Implementation
--- a/07.md
+++ b/07.md
@@ -20,10 +20,10 @@ Aside from these two basic above, the following functions can also be implemente
 async window.nostr.getRelays(): { [url: string]: {read: boolean, write: boolean} } // returns a basic map of relay urls to relay policies
 async window.nostr.nip04.encrypt(pubkey, plaintext): string // returns ciphertext and iv as specified in nip-04 (deprecated)
 async window.nostr.nip04.decrypt(pubkey, ciphertext): string // takes ciphertext and iv as specified in nip-04 (deprecated)
-async window.nostr.nip44.v2.encrypt([[pubkey1, plaintext1], [pubkey2, plaintext2], ...]): string[] // takes array of [pubkey, plaintext] pairs, returns array of ciphertexts as specified in nip-44
-async window.nostr.nip44.v2.decrypt([[pubkey1, ciphertext1], [pubkey2, ciphertext2], ...]): string[] // takes array of [pubkey, ciphertext] pairs, returns array plaintexts as specified in nip-44
-async window.nostr.nip44.v2.getConversationPublicKey([pubkey1, pubkey2, ...]): string[] // takes array of pubkeys, returns array of conversation public keys
-async window.nostr.nip44.v2.signEventWithConversationKey(pubkey, event: { created_at: number, kind: number, tags: string[][], content: string }): Event // takes a pubkey and event object, adds `id`, `pubkey` and `sig` using the conversation key as the private key and returns it
+async window.nostr.nip44.encrypt([[pubkey1, plaintext1], [pubkey2, plaintext2], ...]): string[] // takes array of [pubkey, plaintext] tuples, returns array of ciphertexts as specified in nip-44
+async window.nostr.nip44.decrypt([[pubkey1, ciphertext1], [pubkey2, ciphertext2], ...]): string[] // takes array of [pubkey, ciphertext] tuples, returns array of plaintexts as specified in nip-44
+async window.nostr.nip44.getConversationPublicKeys([pubkey1, pubkey2, ...]): string[] // takes array of pubkeys, returns array of conversation public keys
+async window.nostr.nip44.signEventWithConversationKey(pubkey, event: { created_at: number, kind: number, tags: string[][], content: string }): Event // takes a pubkey and event object, adds `id`, `pubkey` and `sig` using the conversation key as the private key and returns it
 ```
 
 ### Implementation


paulmillr:
I think it's only:

```
async window.nostr.nip44.v2.getConversationKey(pubkey): string // takes third party pubkey and returns shared secret key as specified in nip-44
```

Because encrypt and decrypt take `conversationKey` directly - not priv/pubKeys.

monlovesmango:
I think it would be better to include the version as an argument so we aren't having to update the specification with each new version. it would probably also be much easier on clients to have a single call rather than a unique call for each version. what do others think? @staab @vitorpamplona @pablof7z 

encrypt and decrypt only make sense if they take pubkey. if they take conversation key there is no point of these functions (because the app can do the encryption/decryption itself). also it is good to give apps the option to NOT store the conversation keys (which isn't possible if encrypt/decrypt only take conversation key).

as for naming, I would prefer getSharedKey over getConversationKey (because it will have use cases outside of just messaging). but whichever we choose we should keep it consistent with NIP46 (which is just nip44_get_key right now).

vitorpamplona:
on the version, here's what I am doing: 

`conversationKey`: version is optional. It should return the newest version, but to decrypt locally, the caller will need specific versions based on what's on the ciphertext. 

`encrypt` should encrypt with the newest version all the time. 

`decrypt` has the version in the ciphertext and must decode accordingly (if the extension supports it).

Both decrypt and encrypt should only be called with the client cannot perform these operations by itself.

paulmillr:
> also it is good to give apps the option to NOT store the conversation keys

No. It's **great** to store conversation keys and it's **not great** to store master private key, which is used to sign all events and generate all conversation keys.

That was my thinking when making nip44. getConversationKey was explicitly made caching-friendly. It is supposed to be cached and stored, for performance reasons. It is really slow.

This also allows building trezor-like external signers. They would just sign the events themselves and won't encrypt individual messages. They would be able to generate conversation keys, which the frontend would store / cache in itself.

paulmillr:
As for the version stuff: I don't know how v3 could look like. Will it use method similar to getConversationKey at all? I'm not sure - maybe not. We are talking about mixing post-quantum-cryptography keys, maybe ephemeral secrets, or something. So it's not certain. v2.getConverastionKey is descriptive and local.

>I would prefer getSharedKey over getConversationKey

You should have voiced this concern in nip44. Conversation key term is used all over it, that includes test vectors and implmenentations.

monlovesmango:
@vitorpamplona this makes sense. when you call encrypt/decrypt are you using the pubkey or the conversation key?

@paulmillr agree its good to store conversation keys. in my mind not storing conversation keys doesn't mean you have to store master private key either. however, I hear you and understand that the conversation key should be cached locally. from the convo in nip44 it seemed that this would be optional, not required. 

as for convsersationKey verses sharedKey verses getKey, I guess I don't care how we refer to it in the technical aspects but we should give thought to the naming of the public facing interface. getConversationKey wasn't in the nip44 spec and I had tried to add it in [here](https://github.com/nostr-protocol/nips/pull/746/files#r1419650667). I only chose getKey because that was what had been added to nip46 and was trying to stay consistent.

vitorpamplona:
> when you call encrypt/decrypt are you using the pubkey or the conversation key?

Yes. Then I expect the signer to cache the conversation key to avoid recreating all the time.

monlovesmango:
per @staab from prior closed pr https://github.com/nostr-protocol/nips/pull/922 (want to consolidate the conversation):
>I haven't been able to track all the changes to this file and NIP 46 that have been proposed, what's the argument for getKey? The downside is that an evil client can exfiltrate conversation keys for all DMs, and since we have no forward secrecy or post-compromise security that gives them access to snoop forever.

@paulmillr is saying we should primarily rely on getKey only and have apps/clients cache it. however I agree with @staab that we should be cautious of having apps/clients cache the shared key by default. I lean towards the browser extension (or signing device in the case of NIP46) to do the caching instead and not force the use of getKey by apps/clients.

staab:
Right, the whole point of signers is to minimize trust, so I feel that we should... minimize trust. What clients choose to do is irrelevant, if clients have access to getConversationKey it can be abused.

vitorpamplona:
> Right, the whole point of signers is to minimize trust, so I feel that we should... minimize trust.

The signer can have an option where the user authorizes the client to decrypt all content. In that case, the client should have access to the Conversation key directly. If the GetConversation Key fails, then the client can encrypt/decrypt directly with the extension.

staab:
> The signer can have an option where the user authorizes the client to decrypt all content.

But why? How is that different from "remember my choice and don't ask again"?

vitorpamplona:
The reason is performance. Especially for NIP-46.

staab:
Because of shared key generation? Caching of conversation keys can be done in the extension/provider the same way it can be done on the client.

vitorpamplona:
Yep, but calling decrypt for every chat message and calling just once to get the key and decrypting them all locally have very different performance impacts. It's definitely true on NIP-46 and on the android signer, maybe less on the extension.

monlovesmango:
the only difference in performance would be network latency, right?

@staab I tend to agree with you (latency is worth not having shared keys cached in every app) however I don't think its worth fighting against getKey/conversationKey since the browser extension or signing app can always restrict this on their side. so ultimately the user can prevent clients/apps from caching the shared secrets very easily.

does anyone opinions on the naming of the getKey call? should it be consistent with nip46 (or is it just me and my OCD that cares about consistency)? @vitorpamplona I know you already started implementing something

I prefer getConversationKey over getKey (I just used getKey bc it was in nip46). but I think getSharedKey is better than both.

vitorpamplona:
@paulmillr why did you choose "conversation" key instead "shared" key? 

I tend to agree with @monlovesmango that shared key is a better name for a secret shared between two keys. `Conversation key` only works on one-on-one chats. Group chats using GiftWraps will have conversation keys for each pair of users inside that group. So, it's not really a "conversation" key.

paulmillr:
I picked it because there was no discussion or differing opinions on the naming. That's mostly it. It's possible to still change it, if that's the final decision.

My reasoning was mostly that there's a key which should only be used for conversations. It's separate from all other keys. encrypt and decrypt only use this key and not some "private" or "public" key. If there was a public key in encrypt/decrypt, the question is: whose public key it is? etc

fiatjaf:
I like the name "conversation key".

staab:
> Yep, but calling decrypt for every chat message and calling just once to get the key and decrypting them all locally have very different performance impacts. It's definitely true on NIP-46 and on the android signer, maybe less on the extension.

So we're talking network latency here? I don't think that would be substantial in NIP 07. And for NIP 46, the correct solution would be to add `encrypt_bulk` type methods. That way you only experience the network latency once per chunk.

No one has addressed the security issue I mentioned, is it not an issue for anyone that clients can exfiltrate conversation keys? It seems to defeat most of the purpose of a signer.

monlovesmango:
>No one has addressed the security issue I mentioned, is it not an issue for anyone that clients can exfiltrate conversation keys? It seems to defeat most of the purpose of a signer.

signer can easily limit use of or disallow conversation key requests (and only respond to encrypt/decrypt)

I do think bulk encrypt/decrypt methods might be nice

vitorpamplona:
> No one has addressed the security issue I mentioned

GiftWraps come from random keys. So, there is a getKey for every single message. The getKey of the Seal (which comes from the real author) will provide performance improvements. 

That means, even if the client shares the conversation key with a "snooper" client, it still has to share the conversation key of each individual wrap. There is no way to just listen in to the conversation by sharing the inner key. 

Also, it's impossible to sign a valid event with just the conversation key. You need the master private key to forge any encrypted message.

staab:
> signer can easily limit use of or disallow conversation key requests (and only respond to encrypt/decrypt)

Then why have it? This will also break clients that assune a signer implements the full interface.

> GiftWraps come from random keys.

NIP 44 doesn't assume single-use keys. The issue is mitigated with gift wraps, but still exists because as you say, the seal's key can be re-used. But for non-gift wrap applications of NIP 44 this doesn't apply.

vitorpamplona:
Well, I don't have a good view of non-giftwrapped uses of NIP-44. Maybe there are cases where the metadata leak is ok, but I don't know. 

I will leave the decision to you guys. 

Maybe next time we can create a Conversation key that changes weekly (it must match the year+week of the event). In that way if the conversation key leaks, it can only be used to decrypt events in that week.

vitorpamplona:
> Maybe next time we can create a Conversation key that changes weekly (it must match the year+week of the event). In that way if the conversation key leaks, it can only be used to decrypt events in that week.

Ok, I am starting to think this is a great idea. For v3 (sorry for bringing a new version this soon), what if we used the first 7 digits of the `.created_at` as the conversation key salt that today is fixed as `nip44-v2`? That seems to create new keys at every 11 days. So, those can be used to create shared secrets that automatically rotate based on date. We can even use 8 or 9 digits and make it rotate every day or 3 hours, respectively. Maybe event kinds can choose their own salt rotation based on the security needs they have. And signers can approve the client with an expiration date. 

Then conversation keys can be shared with clients without much fear of clients going rogue with them. 

And if they ever leak, there will always be a time-based protection into them.

paulmillr:
I am out of loop on the wrapping. Can someone give a 2-sentence overview of what’s the goal with the key sharing here?

vitorpamplona:
@paulmillr sorry for the long explanation. 

The goal of the key sharing between applications (signer app and the nostr app) is to avoid too many calls between those two apps. Usually, the signer app has the user's private key and the nostr app calls decrypt in every chat message before displaying. If there is network latency in each call, things get slow very quickly. Sharing the conversation key with the nostr app then allows the app to just call getKey once and decrypt all messages locally. But that comes at a risk that the client might be ill-designed and leak the conversation key. With the control of the conversation key, an attacker could then decrypt all past and future messages between the two people. 

Now, the wrapping design offers some protections because each message is encrypted and signed twice. The outer event that gets transferred through Nostr relays (the wrap) always uses a random private key to encrypt and sign and thus the conversation key is unique to each message. However, the inner event (the seal) is signed by the sender's main private key to the receiver's main pub key. Sharing the inner key between apps can provide significant performance improvements and if it leaks the attacker still must decrypt all the outer events with their individual keys. Thus, when using GiftWraps, it's ok to share the conversation key. 

The issue happens when Giftwraps are not being used AND the conversation key is being shared. Take NIP-51 Lists for instance. There is a private part to lists where we encrypt the information to my own user. The conversation key is between the user's private key and the user's pubkey. If that conversation key leaks, an attacker can then watch the user's private payload forever. 

Which leads to the idea of using the salt in the conversation key design. If we allow that to be flexible, each NIP can have a way to determine the salt based on information in the signed event. We would just add a third parameter to the conversation key that defaults to `nip44-v2` if nothing is provided. By using the first `X` chars of the `.createdat` timestamp, each event kind can choose a time-based rotating salt to reduce the impacts of a conversation key leakage if it ever happens. 

The List NIP for instance, could specify the use of the first 7 digits of the `.created_at` as salt for the encryption. This means that the conversation key will be different at every ~11 days and if it leaks only the messages inside that 11 day block will be decrypted by an attacker. We could reduce that further to a day by specifying the use of the first 8 digits as salt. 

This seems beneficial and it would only slightly change the NIP44 design. We would just offer the salt as a parameter in the `get_conversation_key` function with the default value if NIPs don't care about specifying one. 

From:

```
def get_conversation_key(private_key_a, public_key_b):
  shared_x = secp256k1_ecdh(private_key_a, public_key_b)
  return hkdf_extract(IKM=shared_x, salt=utf8_encode('nip44-v2'))
```

To: 

```
def get_conversation_key(private_key_a, public_key_b, salt=utf8_encode('nip44-v2')):
  shared_x = secp256k1_ecdh(private_key_a, public_key_b)
  return hkdf_extract(IKM=shared_x, salt)
```

paulmillr:
I like it. Good idea. Make sure you've got integer encoding rules correct: amount of bytes and endianness (BE in other parts of spec).

And we should keep the version prefix for future proofness. So, instead of `salt=created_at`, it would be `salt=concat(encode('nip44-v2'), created_at)`.

vitorpamplona:
Is there a size limit to this salt?

paulmillr:
it should be 32 bytes at most, also `hash(nip44-v2 + created_at)` is not better than `concat`

paulmillr:
Also in this case we may need to add created_at to `aad` of hmac

vitorpamplona:
My concern is that if we allow any string or byte array in the get conversation key API (concatenated with the nip44-v2), then some NIPs might decide to put the event id or a longer tag in there. So, we either advise them to not do it or add a 32-byte hash of whatever comes in.

paulmillr:
i prefer the len check. no need to do additional hashing

vitorpamplona:
@staab @mikedilger What do you guys think? Is this additional complexity worth pursuing? I have Amethyst ready for a release for the NIP-17 with v2 of NIP-44, but I can hold off if we want to quickly add this to v2. I don't think we have any public version of v2 yet. 

If we add this salt, I think NIP-17 should adopt weekly hashes for the conversation key.

staab:
If we can add the salt in a backwards compatible way, I think it would be fine. NIP44 is already merged into nostr-tools and so is in the wild.

paulmillr:
> Clients SHOULD randomize created_at in up to a week in the past

how does this work with new salt proposal? if epoch begins on monday and the msg is created on tuesday, there's a chance "up to one week in the past" would mean saturday, a different epoch

paulmillr:
So, with gift wraps, there are two decryptions, meaning two different getConversationKey. Correct? How does it play with remote signers, exactly?

vitorpamplona:
>> Clients SHOULD randomize created_at in up to a week in the past

> how does this work with new salt proposal?

Clients should use whatever is declared in the `created_at` field. It doesn't matter if the date is accurate or not. If two messages end up yielding 2 conversation keys because they are two separate weeks, it's fine. Any cache of the key in the client side must be by pubkey + salt (first 7 digits of the created_at in this case). 

So, with gift wraps, there are two decryptions, meaning two different getConversationKey. Correct?

For the same message, correct. The conversation key of the inner event might be the same between messages. 

> How does it play with remote signers, exactly?

In this case, the client will ask for a different key for each GiftWrap (the external event) but it would reuse keys to decrypt the internal event.

vitorpamplona:
> If we can add the salt in a backwards compatible way, I think it would be fine. NIP44 is already merged into nostr-tools and so is in the wild.

If we add this now, NIP-44 is still backwards compatible. But folks that have implemented NIP-17 with the v2 of NIP-44 would be (which I don't think anyone did already).

staab:
Solutions that go from idea to standard in two days are not likely to be very good. I'm sure better ideas will come up if we give it time to simmer. I don't mind changing the function signature though, seems like a useful affordance.

staab:
All this does is reduce the impact of a leaked conversation key right? That seems like not the most fundamental problem with nostr messages (leaked primary key), and wouldn't be a problem if we didn't expose getKey. Let's just keep it simple and deal with performance issues with more conventional tools.

mikedilger:
I haven't released anything yet, and I like using some created_at bits as part of the salt a lot.  I don't see why not use even more bits of created at and have conversation keys roll over rapidly.  As long as the same value is used for decryption (and it would be as it is written into 'created_at' in the event) it should be fine.

arthurfranca:
#945 proposes the use of the conversationKey as a private key that both participants use to sign DM gift-wraps [here](https://github.com/nostr-protocol/nips/pull/945/files#diff-58e2c23e46271957d48ea789043bb41d6c096df68176c1e46573bb5711498d7dR113-R116). Then both can fetch DMs from that specific chat window by filtering with `{ authors: ["<conversation-key-AB-derived-pubkey>"] }`

This would demand from NIP-07 the following:
1) not to reveal conversationKey to clients
2) window.nostr.nip44.signEvent <- sign with conversationKey
3) window.nostr.nip44.getPublicKey <- the pubkey derived from conversationKey to be used on filter

monlovesmango:
starting to agree with @staab, we shouldn't expose getConversationKey. any performance concerns with encryption/decryption are solely network latency based and can easily be resolved by allowing bulk operations (encryption/decryption methods that take array instead of single param). I don't think getConversationKey should be exposed in the name of efficiency if there is another easy way to be efficient. also the scheme @arthurfranca proposed above seems like it is worth having.

I also really like the idea of adding part of the created_at into the salt, however my opinion in this subject area is pretty meaningless haha.

so I am thinking about the following changes:

- remove 'getConversationKey' (aka 'getKey')
- encrypt should take an event template and return an encrypted and signed event (or can be for multiple events but doesn't seem necessary to me for encryption..?)
- decrypt should take an array of events and return an array of plaintext strings

this would be very different than the established methods for nip04, but perhaps this is a better path forward. my thought is encrypt should take an event template so that it will be easy to incorporate other event attributes to the salt and so the signed event can be returned at the same time (I could not think of an instance where you would want to encrypt something but it wouldn't want it to be in a signed event, so it seems logical to put these together however I could be missing some use case). for decrypt it should also have all the attributes of the event incase they are used for salt in future versions, and we should be able to decrypt multiple events at the same time. 

what do folks thing about this?

@arthurfranca I think your 2 and 3 above should be dedicated method calls. would love to hear everyone's thought on these as well and am happy to add these too.

staab:
> I could not think of an instance where you would want to encrypt something but it wouldn't want it to be in a signed event

In fact nip 44 requires that payloads be within a signed event. This would enforce that, which is good. However, your design assumes `content` is what is encrypted. I can imagine use cases for encrypted tag values coming up though, we probably shouldn't lock people out of building those.

vitorpamplona:
I think encrypt should be on any string, not on an event. 

If decrypt takes an array, I think encrypt should also take one. Clients that are sending several messages at the same time to different receivers could then use that call. 

The client then reassembles the resulting data as it wishes.

monlovesmango:
>However, your design assumes content is what is encrypted. I can imagine use cases for encrypted tag values coming up though, we probably shouldn't lock people out of building those.

good point, will ponder on what might be a good way to specify this

>I think encrypt should be on any string, not on an event.

whats your reasoning here? are there scenarios you can think of where you will be encrypting but not putting resulting ciphertext in a signed event? also if we make it a string it will be necessary to include all elements contributing to the salt as additional params (which could get especially inelegant if we change the salt with subsequent versions). if we just send the event template we'll likely cover future salt schemes in a single, simple object.

>If decrypt takes an array, I think encrypt should also take one. Clients that are sending several seals and wraps at the same time to different receivers could then use that call.

fair enough, agree lets just make encrypt an array as well.

vitorpamplona:
> whats your reasoning here?

Tags can be encrypted separately from the content. And tags are just strings. That being said, it shouldn't add anything else. The encrypt method should be an array of text and pubkey pairs. 

You can check https://github.com/nostr-protocol/nips/pull/716 for an example on the encrypt call.

arthurfranca:
The methods could be `window.nostr.nip44.signEventWithConversationKey` and `window.nostr.nip44.getPublicKeyFromConversationKey` for easier understanding. Maybe they could accept a version argument that would default to the latest nip44 one.

Imo using the conversationKey as a privateKey for a 1-1 DM was a brilliant idea (from @vitorpamplona) cause:
- it hides who the 2 participants are
- even if you could figure out who they are it hides which one of them signed an specific message/was online at its created_at timestamp (because both participants use the same key to sign their gift wraps)
- as an alias it doesn't need to be exchanged/stored because both already know it, just need to poke the other party so they can start filtering by `{ authors: ["<pubkey-from-getPublicKeyFromConversationKey>"] }`

monlovesmango:
what if we create a separate method for each encryption use case? so `encryptContentAndSign` would be one method, `sealWrapAndSign` could be another, when people create a use case for encrypting tags another method can be created for this.

maybe people will hate this idea but I actually think it works rather well and is simple to use (yes I am still trying to push for sending event templates only into the encryption method @vitorpamplona hehe)

pros:

- more defined error handling of unsupported signer methods
- much easier implementation on client side (encrypt and sign in one step, only need single event template param)
- encryption use cases will be concretely and narrowly defined for signers
- doesn't need additional params to be scoped out if we start using event attributes as part of salt in new nip44 versions (like using `created_at` as @vitorpamplona suggested) bc the event template should contain these attributes.
- should eliminate the need for `signEventWithConversationKey` that @arthurfranca mentioned (since the signer can handle the sealing and wrapping and just return the gift wrap event), though a `getConversationPublicKey` method will still be needed

cons:

- more methods that potentially need to be supported (though implementation of these should be pretty straightforward)
- ??? (anyone have any cons to add?)

thoughts? anything I am missing?

vitorpamplona:
It's similar to what Amber had to do to sign Private Zaps. It's kind of a mess. Because any NIP can create their own way of using the NIP44 encryption and if they do you will have to create another method in the Signer to support that NIP. Right now it's just the GiftWraps and Seals but I don't think it will end there. 

NIP-82, medical data, for instance, will have to encrypt a single secret for each receiver and place it in a separate tag in the event and then encrypt the contents of the event with that secret. You might have 5-10 encryptions into a single event.

monlovesmango:
agree it won't end with gift wraps and seals.

however for the NIP-82 example it sounds more trivial to implement another method in the signer to support the nip than asking each client to make 5-10 encryption calls to put together the event. especially if adoption and latency mitigation is the goal, it seems like having one complete call implemented at the signer level is a superior option to forcing clients to orchestrate multiple encryptions themselves.

staab:
I don't think we should increase the surface area of the api to cover every use case. Just bulk/single sign/encrypt should be enough to cover everything, and simple enough for devs to use.

monlovesmango:
ok I have updated the pr to just implement simple encrypt/decrypt but with an array of params and removed `getConversationKey`. I used nip44.v2 format because I agree with @paulmillr that we don't know what v3 will look like and what additional params it may need. 

I also added methods needed for nip59. let me know if anyone thinks the `signEventWithConversationKey` method should take an array.

arthurfranca:
> The List NIP for instance, could specify the use of the first 7 digits of the .created_at as salt for the encryption. This means that the conversation key will be different at every ~11 days

@mikedilger @vitorpamplona +1 to the idea of adding the salt to NIP-44 current v2. If using the conversation key as a private key (an already known alias that doesn't need to be exchanged - like on #945) paired with the 7 digits salt idea, the alias would be auto-rotated at the said 11 days frequency. Already-Known-Auto-Rotatable-Aliases™ with no need of an alias exchange step. 

Although with added client complexity when fetching events from such alias pubkeys, it would help with security.

arthurfranca:
> The List NIP for instance, could specify the use of the first 7 digits of the .created_at as salt

In fact, it should be the first 4 digits from the left for a ~12 days window.
3 digits would be ~4 months or 5 digits for ~1 day window or 6 digits for ~3h and so on.

I think for alias rotation use case, the 4 digits would be best balance because more than that would mean too many `REQ` filter time range change frequency when syncing new events.

staab:
WRT to the conversation key salt thing, would this addition invalidate messages currently being encrypted? IOW, would it be backwards compatible? I'm not opposed to the idea itself, but if it's a completely incompatible implementation, I think the ship has sailed. Most people do not assiduously follow this repo, and if messages suddenly stop being interoperable, most people will not know why. We shouldn't be making up random standards for other people to follow in a vacuum like this. Interoperability is already at risk with the migration to nip 04 to 44, let's not add more implicit versions than we need.

arthurfranca:
https://github.com/nbd-wtf/nostr-tools/blob/348d118ce40b712aa2ed2bdd1766f467ed20649e/nip44.ts#L22

@staab the above function would change to accept a custom salt instead of the fixed `'nip44-v2'` string.
So we would just have to add a third arg to the NIP-44 getConversationKey function that would default to `'nip44-v2'`  but can be
set to whatever the NIP-XX requiring NIP-44 decides, like the stringified first 4 digits of an event's created_at.

> would this addition invalidate messages currently being encrypted?

If I'm not mistaken, with the default salt being `'nip44-v2'`, nothing would change for people unaware of this addition.

staab:
Yes, but it would change NIP 17, which isn't "finalized", but is currently in use, which is more important in nostr world.

monlovesmango:
if the methods in nip07 and nip46 are adopted these would also need to be updated for users. since we are no longer exposing `getConversationKey` maybe the addition of the created_at in the salt isn't as necessary immediately? or maybe we want to hold off on nip44 until we can get v3 with salt sorted?

also, should we replicate the structure of the NIP-07 method calls in NIP-46 as well? (ie remove `nip44_get_key`, change `nip44_encrypt` and `nip44_decrypt` to have v2 in name and take arrays, add conversation key methods)

arthurfranca:
@staab if NIP-17 defines a salt to use than yes it would be a breaking change to NIP-17. However, just adding the optional salt arg to NIP-44 wouldn't break anything.

@monlovesmango no need to add a getConversationKey function to NIP-07/46. The nip44.encrypt, nip44.decrypt, nip44.getConversationPublicKey, nip44.signEventWithConversationKey functions would need to accept an optional salt argument, though.

vitorpamplona:
It's easy to make the conversation salt backward compatible: Just define a new event tag (e.g., `key_salt`) into NIP-44. If the tag is present, it's used as salt of the conversation key. If not, the default value is used as: encode('nip44-v2'+key_salt)

All current uses do not have the tag, so it is backward compatible.

staab:
> All current uses do not have the tag, so it is backward compatible.

But not forward compatible. Existing naive implementations will fail to decrypt messages. That's a huge UX problem, which I think needs to be more in view in these discussions. We can't just introduce new formats and expect everyone to pick them up.

vitorpamplona:
> But not forward compatible. Existing naive implementations will fail to decrypt messages. 

That's for the NIPs using the salt tag to decide. Forward compatibility is not broken by the addition of this parameter. For instance, all NIP-17 implementers will work because the salt is not used in the current NIP-17.

monlovesmango:
I don't think its right to use a single method for all future nip44 versions and make salt optional bc:

- it will effectively **NOT** be optional for future versions that require salt
- it will be hard to enforce the right salt is being sent for encryption, and if misused can cause errors and confusion with decryption (when those errors should be caught with encryption instead)
- it will require an additional version parameter in order to accommodate multiple future versions with different salt schemes (unless we can guarantee that all future salt schemes will be simplistic and only use a single string)

this is why I was inclined to try and create purpose built methods instead that take events or event templates, as this would likely not need additional optional parameters to accommodate future unspec'ed versions of nip44 with a single method call.

vitorpamplona:
> it will be hard to enforce the right salt is being sent for encryption

The signer should not be enforcing what is the right salt. It's whatever NIPs decide. Obviously, implementers will need to pay attention to it. But that is not that different from using new keys or secrets to encrypt/decrypt. 

> we can guarantee that all future salt schemes will be simplistic and only use a single string

I don't see any other non-string version (salt must be on tags and tags must be strings).

--

Also, I think the decrypt function should decrypt all available versions instead of requiring devs to pick the method that matches the version in the event (e.g. v2.decrypt, v3.decrypt, etc.  ) Otherwise the client must implement part of the NIP-44 decoding scheme just to know which version the ciphertext was encrypted. 

Of course, functions for each version can be available as well, but we don't need to require clients to figure out which one is which. The master decrypt function should be the default use. 

Similarly, a master encrypt function should always return the ciphertext in the latest version it supports. Devs can call individual versions, but by default, they should not worry about it.

paulmillr:
Can someone show me a PoC code of the whole wrapped stuff with temporary per-epoch keys?

Just 1-2 events would be enough. Code is better than posts.

monlovesmango:
>The signer should not be enforcing what is the right salt.

then how will signer know what version to mark the encryption as?

>Also, I think the decrypt function should decrypt all available versions instead of requiring devs to pick the method that matches the version in the event (e.g. v2.decrypt, v3.decrypt, etc. ) Otherwise the client must implement part of the NIP-44 decoding scheme just to know which version the ciphertext was encrypted.

devs will already need to do this bc they will have to determine which salt to use, or whether salt is needed at all. having a single decrypt function for all versions will not make this easier, just more ambiguous.

vitorpamplona:
> wont devs already need to do this bc they will have to determine which salt to use, or whether salt is needed at all? having a single decrypt function for all versions will not make this easier, just more ambiguous. or is the salt able to be extracted from the encrypted string?

The point was independent of the salt decision. If the salt decision is added, every encrypt and decrypt method will need to take an additional param as the salt to create the conversation key.

monlovesmango:
> The encryption version is encoded inside the ciphertext.

yes I am aware, my question is around how the version is determined when encrypting (before ciphertext is created). if we only have a single encrypt method for all versions and we don't have a version parameter and we don't validate the type of salt, how does the encrypt method know which version to encode in the ciphertext? 

>The point was independent of the salt decision. If the salt decision is added, every encrypt and decrypt method will need to take an additional param as the salt to create the conversation key.

I understand your point is independent of the salt decision. I was just saying that due to the need to accommodate future salt schemes, your point about not "requiring devs to pick the method that matches the version in the event (e.g. v2.decrypt, v3.decrypt, etc. )" buys minimal convenience bc devs will already have to pick the salt that matches the version of the event to make the decrypt call. so they will already have to tailor their decrypt method call by version regardless of whether the decrypt call is versioned or not.

given that nip44 will likely form the basis of a lot of functionality in nostr I think its important that these methods are well defined and non-ambiguous.

vitorpamplona:
> how does the encrypt method know which version to encode in the ciphertext?

That was my proposal. It should use the latest version of NIP-44 that the signer has implemented. This will guarantee that most clients move to new versions automatically as new implementations come out by the signer as opposed to have NIP-X on stuck on version 2 and NIP-Y stuck on version 3. Of course, individual methods for each version should also be available as a fallback for developers who want to fix their code in specific versions. But if these are true "versions" of the same spec and not a menu of algorithms people can choose to use, it makes sense.

monlovesmango:
that seems to contradict with what you said earlier (quoted below) but I may still be misunderstanding something. earlier it seemed like you were suggesting that NIPs should independently define which version it employs, but now it seems like you are suggesting every NIP should default to new version. what am I misinterpreting?

> That's for the NIPs using the salt tag to decide. Forward compatibility is not broken by the addition of this parameter. For instance, all NIP-17 implementers have forward compatibility because the salt is not used in the current NIP-17.

and thinking about it more, seems really buggy to automatically use the latest version for encryption. lets say a client is on v2 and signer is on v3, and v2 and v3 use different salts. the client calls encrypt with v2 salt but the signer then encrypts it as v3 ciphertext. this would lead to very confusing error upon decryption. is there something I am missing here?

vitorpamplona:
> earlier it seemed like you were suggesting that NIPs should independently define which version it employs

Not which version, but which salt. Salt choice here is orthogonal axis to the version choice - there are many salt models for v2 and the same models can be used for v3. The salt model defines when the conversation key rotates and that should be a NIP decision. The NIP-44 version is the NIP-44 decision: as new versions come in, every nip should move to the new versions. 

> lets say a client is on v2 and signer is on v3, and v2 and v3 use different salts. the client calls encrypt with v2 salt but the signer then encrypts it as v3 ciphertext

Maybe `salt` is not a good name for the partial string NIPs provide to rotate a shared key calculation. Yes, the use of that parameter by v2 and v3 can be different, they will be deterministic. In the sense that, newer versions will always receive a string for a shared key salt and they can decide how to use that string in any way they see fit. Since the ciphertext has the version, any encryption of a version using the salt can be decrypted by the same version using that salt.

vitorpamplona:
Maybe we are extending this conversation too much. Let's get methods published for the current version of NIP-44 without the salt ASAP so web clients can start using NIP-44.

monlovesmango:
wow thank you for detailing that, I did not realize salt is orthogonal to version.

I removed the versioning, I think it should be ready to go now?

apologies to all for the excessive discussion.

staab:
Do we have a concrete use case for `signEventWithConversationKey` yet? If not, we should remove it until we do. If so, should we make the signature batch-friendly?

arthurfranca:
> Let's get methods published for the current version of NIP-44, without the salt 

What's blocking the salt addition? @paulmillr and @mikedilger were already in favor on previous messages.
Just think `encode('nip44-v2'+salt_suffix)` may need a separator (like  `:` string) prepended to `salt_suffix` when the suffix is present.

arthurfranca:
@monlovesmango we should delete getConversationPublicKeys and signEventWithConversationKey
because ["conversation key can’t be used as private key"](https://github.com/nostr-protocol/nips/pull/945#issuecomment-1882545014)

arthurfranca:
Regarding the salt, I've just rewritten the previous DM PR to use it to try achieving forward secrecy.

[Here](https://github.com/nostr-protocol/nips/pull/978/files#diff-58e2c23e46271957d48ea789043bb41d6c096df68176c1e46573bb5711498d7dR132-R136)

monlovesmango:
@arthurfranca ok I have removed those, thanks for heads up.

as for salt, I think that should only be added after there is a nip44 version that uses salt. has v3 of nip44 been spec'ed yet?

jb55:
> has v3 of nip44 been spec'ed yet?

just exactly when will I be able to start implementing this? is it still churning this much?

monlovesmango:
so should I add optional `salt` parameter into these methods?

staab:
So I was adding this to Alby, and there is a downside to only providing bulk methods, in that it makes the dialog much more complex for end users. I think we should have singular encrypt/decrypt methods with optional bulk ones.

monlovesmango:
if we do this, are you saying that Alby would not support the bulk methods?

monlovesmango:
even if Alby doesn't use the bulk messages, it will still have the same dialog requirements no? even if singular encrypt/decrypt is used, the number of items that are encrypted/decrypted and need a dialog will be the same. 

I guess I'd be against having separate singular/bulk calls bc it makes it more complicated for every client (you have to handle scenario where bulk method is not supported) and imo doesn't buy much efficiency on signer's side (the same number of encrypt/decrypt requests still need to be handled and presented to the user). also, it doesn't seem that hard to iterate through the array of requests to display a dialog for each individual one if Alby really only wants to display one at a time. you could also group requests for the same pubkey together so its easier for user to review in batches.

anyone else have opinions on this? I can break out the single/bulk methods if thats what people want. I just don't like adding complexity for little gain.

staab:
I think what you're saying makes sense. I'll let you know if the alby guys express an opinion.

vitorpamplona:
Are we ready to merge this? What's the status with the extensions?

staab:
I haven't heard much from the alby guys yet about the UX, see https://github.com/getAlby/lightning-browser-extension/pull/2653. Let's wait for two implementations before merging as per usual. I have an open PR to nos2x as well, but it needs to be updated to the new API.

staab:
Updated my PR to nos2x: https://github.com/fiatjaf/nos2x/pull/47

staab:
@erskingardner doesn't like positional arguments and I agree (he's working on an update to nip 46). Can we do something like:

```
async window.nostr.nip44.encrypt([{pubkey, plaintext}, ...]): string[] // takes array of {pubkey, plaintext} objects, returns array of {ciphertext} objects where ciphertext is as specified in nip-44
async window.nostr.nip44.decrypt([{pubkey, ciphertext}, ...]): string[] // takes array of {pubkey, ciphertext} objects, returns array of {plaintext} objects where plaintext is as specified in nip-44
```

monlovesmango:
yeah I'm okay with that.

your comment talks about returning an array of objects as well but I chose to keep that as a simpler string array. was there a particular reason you wanted the returned array to be of objects instead of strings?

arthurfranca:
What about the.......... [salt](https://github.com/nostr-protocol/nips/pull/1020)?

staab:
> your comment talks about returning an array of objects as well

Just for extensibility. It's always possible we would want to return more information. But it probably doesn't matter, and it occurs to me that as bad as positional arguments are, tuples are a more compact representation. Maybe we should go back to tuples (sorry).

> What about the.......... https://github.com/nostr-protocol/nips/pull/1020?

I think we agreed that that can be added later when there's a specific proposal that uses it with some buy-in. From @monlovesmango:

> as for salt, I think that should only be added after there is a nip44 version that uses salt.

fiatjaf:
Shouldn't we do the same thing NIP-46 is doing?

staab:
nip 46 should do what we're doing. The array notation doesn't actually do any good for NIP 07, it's only for consistency with what NIP 46 should do, which is accept bulk requests to reduce round trips.

monlovesmango:
yep, what @staab said. ultimately the design decision we made here was for the benefit of NIP46, so that NIP46 would work well and NIP07 can be consistent with it.

fiatjaf:
So can we merge this?

staab:
Waiting on implementations *ahem* https://github.com/fiatjaf/nos2x/pull/47

fiatjaf:
Merged.

But not deployed. I must go back to nos2x and review the other open PR and then deploy everything.

vitorpamplona:
Let's go! 🚀

alexgleason (approved):

fiatjaf:
Shouldn't we have the `getConversationKey` here too, even if as optional?

monlovesmango:
we had a lot of discussion about this. tbh I think `getConversationKey` should be removed from nip46 too, not added here as well. if people really feel strongly about it, I am ok adding it back in (since it should be easy for the signer to prevent usage of this method if users don't like conversation key exposed). but I lean towards keeping the spec simpler since there isn't much performance gain, creates another attack vector if clients maliciously use the conversation key, and adds complexity to client implementation (and no guarantee clients will handle scenario where `getConversationKey` is not available properly).

if we do expose this, we should at least have a good reason for doing so. since encrypt/decrypt are now batched method calls there isn't much efficiency gain by having getConversationKey. whats another reason to have the `getConversationKey`?

again, most of the considerations around the NIP07 method calls were really had with NIP46 in mind, so that NIP07 would align with NIP46.

I know @staab felt strongly that it shouldn't be exposed, and agree with the sentiment.

fiatjaf:
@monlovesmango what do you think of this? https://github.com/nostr-protocol/nips/pull/1047#issuecomment-1952141468

Basically just make the two ways of decrypting optional and have clients implement one but fall back to the other.

The most important thing is having a uniform interface between NIP-07 and NIP-46 to prevent too much spaghetti.

staab:
> The most important thing is having a uniform interface between NIP-07 and NIP-46 to prevent too much spaghetti.

Definitely this. I still haven't heard a compelling reason to expose the conversation key if we have bulk methods, but it would be faster and less prone to network failures. I'd prefer to leave it out still, but @fiatjaf's compromise would also be fine (although I think it will just result in get_conversation_key being implemented and used everywhere).

07.md
```diff
@@ -20,6 +20,8 @@ Aside from these two basic above, the following functions can also be implemente
 async window.nostr.getRelays(): { [url: string]: {read: boolean, write: boolean} } // returns a basic map of relay urls to relay policies
 async window.nostr.nip04.encrypt(pubkey, plaintext): string // returns ciphertext and iv as specified in nip-04 (deprecated)
 async window.nostr.nip04.decrypt(pubkey, ciphertext): string // takes ciphertext and iv as specified in nip-04 (deprecated)
+async window.nostr.nip44.encrypt([[pubkey1, plaintext1], [pubkey2, plaintext2], ...]): string[] // takes array of [pubkey, plaintext] tuples, returns array of ciphertexts as specified in nip-44
+async window.nostr.nip44.decrypt([[pubkey1, ciphertext1], [pubkey2, ciphertext2], ...]): string[] // takes array of [pubkey, ciphertext] tuples, returns array of plaintexts as specified in nip-44
```
alexgleason, line 24: I will also just add real quick, that letting you encrypt/decrypt multiple at once leads to the possibility that the signer will return an array with a different length than the one you passed in. It's not really a big deal since clients need to validate the response anyway.

erskingardner:
Does the proposal to add a salt actually fix the issue with unconstrained conversation key reuse? (I haven’t looked at it at all yet). If it does, then it might make sense to just add the conversation key method once that’s released. On Feb 19, 2024, at 17:06, Alex Gleason ***@***.***> wrote:﻿
@alexgleason commented on this pull request.



In 07.md:
> @@ -20,6 +20,8 @@ Aside from these two basic above, the following functions can also be implemente
 async window.nostr.nip04.encrypt(pubkey, plaintext): string // returns ciphertext and iv as specified in nip-04 (deprecated)
 async window.nostr.nip04.decrypt(pubkey, ciphertext): string // takes ciphertext and iv as specified in nip-04 (deprecated)
+async window.nostr.nip44.encrypt([[pubkey1, plaintext1], [pubkey2, plaintext2], ...]): string[] // takes array of [pubkey, plaintext] tuples, returns array of ciphertexts as specified in nip-44
+async window.nostr.nip44.decrypt([[pubkey1, ciphertext1], [pubkey2, ciphertext2], ...]): string[] // takes array of [pubkey, ciphertext] tuples, returns array of plaintexts as specified in nip-44

I will also just add real quick, that letting you encrypt/decrypt multiple at once leads to the possibility that the signer will return an array with a different length than the one you passed in. It's not really a big deal since clients need to validate the response anyway.

staab:
> Does the proposal to add a salt actually fix the issue with unconstrained conversation key reuse?

No, it only increases the number of keys an attacker would have to generate.

07.md
monlovesmango: if theres an error, I would think that returned array should retain the same number of values and just put a blank value for the one that err'ed. would probably be good to explicitly say that here.

do you think thats a valid way of handling it?

alexgleason: If any individual message has an error, the whole function call should throw.

monlovesmango:
>The most important thing is having a uniform interface between NIP-07 and NIP-46 to prevent too much spaghetti.

agree.

added the `getConversationKey` method back in. would be nice to have a strong rational for exposing this, but really don't feel strongly about it one way or another. right now the method only takes a single pubkey, do people think it should be changed to take an array as well? or fine as is?

07.md
alexgleason: The more I think about it, the more I think it's actually a problem. The client has no way to validate messages before sending them to the signer, right? So any invalid text can cause the signer to throw (or return a rejected promise).

First I would question if we really need to support decrypting multiple messages. Is it just because Alby prompts every time?

If we really need to do it, it should have a return type of `Promise<string>[]` instead of `Promise<string[]>` so individual calls can be either resolved or rejected.

staab:
An array is probably a good idea, especially if we add salt. I'm still nack on it, but I suppose if it's ever exploited signers can drop support for it.

07.md
monlovesmango: no, it is because for the NIP44 calls in NIP46 it would be extremely inefficient to fetch one decryption at a time. so it is better for NIP46 to have these calls be arrays, and we want the NIP07 calls to be consistent with the NIP46 calls. hence why the NIP44 calls have array parameters.

can you return a promise from a browser extension? returning a promise definitely wouldn't work for NIP46 either, which we are trying to stay consistent with.

monlovesmango: maybe the array returned should house objects like:
```json
{
  "result": "result string"
  "error": "error description"
}
```

don't love wrapping everything in a object though. maybe only the item that err'ed should return an object (rather than a string)? that should be pretty parseable right?

monlovesmango:
ok updated `getConversationKey` to take array instead. I went ahead and made it an array of tuples in anticipation of adding salt in a future PR, so its easy to extend the method to use salt. if it was just a simple array of pubkeys we would need a second array of just salts once salt is added. however I am ok with that if thats what people would prefer (but it would be a significantly different pattern than what we have for encrypt/decrypt.

>I'm still nack on it, but I suppose if it's ever exploited signers can drop support for it.

same :)

07.md
alexgleason: I would say consistency with NIP-46 is not a good enough reason for this to be an array. It's okay for them to be different if there's a good reason. This is a JavaScript API. No other JavaScript API has you pass in an array of multiple arguments. It's an anti-pattern. We should change it to do only one at a time. NIP-46 can do it the complicated way. Let's not turn something simple into something complex just because another thing is complex.

monlovesmango: do you have anything I can read on JavaScript API patterns?

anti-patterns with other nostr functionality (ie NIP46) was my main concern here and imo is the most important pattern to adhere to. (I don't think I'm the only one who feels this way, see https://github.com/nostr-protocol/nips/pull/940#issuecomment-1952150248, and https://github.com/nostr-protocol/nips/pull/940#issuecomment-1952770826)

besides, I do think the array argument is beneficial to NIP07 as well as it allows browser extension signers to group permission requests if trying to create a better request pop up flow for the user. otherwise browsers are forced to prompt the user for each request individually (in the scenario that the user doesn't want to grant blanket decryption/encryption approval to the browser signer).

alexgleason: Check my NIP-07 TypeScript interface: https://gitlab.com/soapbox-pub/NSpec/-/blob/main/interfaces/NostrSigner.ts

And here is an implementation, NSecSigner: https://gitlab.com/soapbox-pub/NSpec/-/blob/main/src/NSecSigner.ts

Usage:

```ts
const signer = new NSecSigner(secretKey); // We pass the secretKey when constructing the object

// Now it's API-compatible with NIP-07 and can even polyfill `window.nostr`
await signer.getPublicKey();
await signer.signEvent(event);
```

As another example, `NSeedSigner` lets you pass a mnemonic seed phrase, derives the secretKey, and then uses `NSecSigner` internally: https://gitlab.com/soapbox-pub/NSpec/-/blob/main/src/NSeedSigner.ts

Finally, `SoapboxSigner` is an application-specific signer that chooses a signer based on the user's preferences: https://gitlab.com/soapbox-pub/soapbox/-/blob/ditto-auth/src/features/nostr/SoapboxSigner.ts

Believe me, I signer. I'm very invested in the signers. I get it. And so the question is, can you implement NIP-46 using the `NostrSigner` interface?

The answer is yes, even if NIP-46 uses arrays and NIP-07 uses atomic calls. You would simply throttle the NIP-07 function in the Signer implementation, and then batch them into a single NIP-46 call. NIP-46 would then return an array of results object like you mentioned, and the Signer implementation would resolve or reject the promises.

Here is a very rough pseudocode:

```ts
export class NConnectSigner extends EventTarget implements NostrSigner {
  #nip46: NIP46;
  #queue: string[][];

  constructor(pubkey: string, relays?: string[], delay?: number) {
    this.#nip46 = new NIP46(this.#pubkey, this.#relays);

    setInterval(() => {
      this.#batchAndSendCalls();
    }, delay ?? 100);

    super();
  }

  async getPublicKey(): Promise<string> {
    return this.#nip46.getPublicKey();
  }

  async signEvent(event: Omit<NostrEvent, 'id' | 'pubkey' | 'sig'>): Promise<NostrEvent> {
    return this.#nip46.signEvent(event);
  }

  readonly nip44 = {
    encrypt: async (pubkey: string, plaintext: string): Promise<string> => {
      this.#queue.push(['encrypt', pubkey, plaintext]);

      return new Promise((resolve) => {
        this.addEventListener('encrypt', resolve);
      })
    },

    decrypt: async (pubkey: string, ciphertext: string): Promise<string> => {
      this.#queue.push(['decrypt', pubkey, ciphertext]);

      return new Promise((resolve) => {
        this.addEventListener('decrypt', resolve);
      })
    },
  };

  #batchAndSendCalls() {
    for (const [method, ...args] of this.#queue) {
      this.#nip46[method](...args).then((result) => {
        this.dispatchEvent(new CustomEvent(method, { detail: result }));
      });
    }
  }
}
```

The point is, NIP-46 is a different beast than NIP-07. NIP-07 defines a JavaScript API. We should focus on creating the ideal JavaScript API. The way you future-proof an API is by making it as simple as possible.

alexgleason: Btw, what I am suggesting here is something that could open the door to nip46 batching multiple types of different calls. So you could sign an event and decrypt stuff in the same call. This is why the JS API should be atomic.

arthurfranca:
TLDR: Exposing conversation key (CK) without salt arg is bad similar to exposing the privkey. Changing salt too often defeats the purpose (caching) of exposing the CK. Conclusion: a client should use the same random salt for a period to be able to use a cached CK for encryption/decryption of many events.

> would be nice to have a strong rational for exposing this

Fact is that If people don't change salt from time to time, exposing conversation key (CK) is bad, cause the client can leak the CK and who has it can decrypt all events that used (and will in the future use) the CK for encryption.

The only [anticipated use case](https://github.com/nostr-protocol/nips/pull/1047#discussion_r1490556297) (that wasn't [debunked](https://github.com/nostr-protocol/nips/pull/945#issuecomment-1882545014) atleast) for exposing CK (with salt arg) is for a theoretical NIP-46 signer that is shared by thousands of users that don't wan't to cache CKs (without salt) from all these users or even with the cached CKs don't want to decrypt too many events (to save compute resources).

If it was not for the above (that may or may not be reasonable, not sure), for [my DM use case](https://github.com/arthurfranca/nips/blob/priv-dm/43.md#session-channel) I would be ok with not exposing CK but just being able to pass a salt arg to nip44encrypt/decrypt functions.

So, imo if people want to expose CKs to cache them on clients, clients **should reuse** a random salt for some time (keep it locally stored and change from time to time) and add it to a public `["salt", <...>]` tag on encrypted events (to be able to later decrypt them).

Because if a client changes the salt too often, it won't be able to take advantage of the CK caching. And if it is the case, it would defeat the purpose of "safely" exposing the CK, cause the client would have to ask the signer to generate many CKs, one for each used salt.

So, imo if we want to expose the CK: without salt (in practice it uses one default salt) is bad; with one different salt per event is bad; with one salt per many events is good.

07.md
monlovesmango: I still think its more important to create the ideal nostr API than javascript API. why should we force throttling, batching and creating a bunch of promises just so nip07 signer and nip46 signer can both use the same interface? if theres a real tangible reason I'd be more amenable, but just to follow some arbitrary javascript api standards (which I would still like to read more about if you have a resource) seems like not a good enough reason to increase implementation complexity.

>The way to future-proof an API is by making it as simple as possible, NOT to make it flexible for existing systems.

we are hardly making this flexible, just allowing multiple calls but for very straightforward and single use case. plus this arbitrary desire for atomic simplicity directly leads to real, known complexity on implementation of signer integration. 

>Btw, what I am suggesting here is something that could open the door to nip46 batching multiple types of different calls. So you could sign an event and decrypt stuff in the same call. This is why the JS API should be atomic.

having atomic calls, rather than arrays, is not a prerequisite for nip46 batching calls. plus the most useful and wanted type of batching is batching encryption and decryption requests. so why implement an api in its least useful form so that maybe someday we can implement something on top that makes it more useful? 

also, it sounds like you are also advocating for atomic nip46 as well? I understand the method call structure may not be standard, but imo making the method call practical and enabling good user experience is more important.

staab: I think I mostly agree with Alex here that NIP 46 should include batching, but software interfaces should be simple with throttle/batch behind the scenes. It would make sense in theory for the NIP 07 interface to match the simple js interface, and have the extension do batching, but from looking at Alby's code I think that might require a lot more involved re-write.

Implementation aside, which is not a good argument, is there ever a time when a NIP 07 should do batching? Is there ever a time when NIP 07 would not be talking to a browser extension? Probably not.

That said, one thing we haven't talked much about is getting rid of NIP 07 entirely, and making it just an implementation of NIP 46. 46 currently relies on relays explicitly, but if you could separate the interface/api and architecture, you could provide a NIP 46 signer in an extension, over http, etc. That would sort of make this a moot point since NIP 07 would no longer exist.

alexgleason: What does Alby do about failed decrypt calls at the moment, when the failed element is in the middle of the array? Does it return `undefined` or return a rejected promise for the whole function? Because I think rejecting the whole promise is the correct idiomatic JavaScript way, but it would make the API totally useless if they did that.

alexgleason: > from looking at Alby's code I think that might require a lot more involved re-write.

That doesn't make sense. Alby doesn't have to do batching unless it's converting calls to NIP-46. It would be less code on their part to only decrypt the one. They can even update their code with a one-liner: `nip44.decrypt = (...args) => _oldDecrypt([...args])`

alexgleason: Found the Alby MR and commented here: https://github.com/getAlby/lightning-browser-extension/pull/2653#pullrequestreview-1891078523

staab: > What does Alby do about failed decrypt calls at the moment

That's a super good point, it throws.

> Alby doesn't have to do batching unless it's converting calls to NIP-46

Alby may want to do batching for UX rather than performance reasons. So instead of 300 pop ups, they should really show "Coracle wants to decrypt 300 messages. ok?".

staab: 🤔 I might agree with you. I know @fiatjaf currently wants the interface the same. Maybe he would be convinced by your argument.

monlovesmango:
fyi @alexgleason has opened alternative version of this PR with atomic calls #1063

monlovesmango:
closing in favor of #1063

PR #941, opened 2023-12-22
Mention deprecation of kind 2 in README

schmijos:
It's now the second time that I searched for what's the matter with kind `2`.

Kind `2` was [consciously](https://github.com/nostr-protocol/nips/pull/703#issuecomment-1672098794) removed in https://github.com/nostr-protocol/nips/commit/72bb8a128b2d7d3c2c654644cd68d0d0fe58a3b1#diff-39307f1617417657ee9874be314f13aabdc74401b124d0afe8217f2919c9c7d8L105.
Mentioning the fact in the event kinds table should help prevent further confusion.


vitorpamplona:
Shouldn't we mark it as deprecated or unrecommended?

PR #942, opened 2023-12-22
Emphasis on multiple filters in NIP-01

schmijos:
As mentioned further down in NIP-01

> A REQ message may contain multiple filters.

The comma makes it more clear that the message array can container more than 3 items. So it is understood easily to match

```
// correct:
["REQ", <subscription_id>, filters1, filters2, …]
```

instead of 

```
// wrong:
["REQ", <subscription_id>, [filters1, filters2, …]]
```


AsaiToshiya:
> // correct:
> ["REQ", <subscription_id>, filters1, filters2, …]
> ```

I think this one is more clear.

staab:
Me too, feel free to pr

PR #943, opened 2023-12-22
Rephrase deprecation of kind 2

schmijos:
I didn't want to touch all that white-space in #941, but here you go @vitorpamplona 😉


fiatjaf:
Why not just "deprecated"? The date doesn't matter. It was never really used.

schmijos:
I thought for reference about people encountering kind 2 events in the wild dated from before they were deprecated. But I was probably overthinking.

AsaiToshiya:
LGTM.

PR #952, opened 2023-12-27
NIP-115 Yet Another Geo Tag (ISO-3166-1/2/3)

dskvr:
This NIP sits somewhere between `g` defined in `NIP-52` and _"places"_ over in #927. It ~extends the `g` tag~ adds `G` tag]. This is the result of patiently watching from the sidelines and conversations from #763. After writing [`nostr-geotags`](https://github.com/sandwichfarm/nostr-geotags/) it was apparent that with `NIP-32`, a specification for interoperability would still be required. And after all was said and done, would still require domain knowledge of ISO standards to properly use. This NIP provides ~_almost*_~ full coverage over the well-established ISO-3166 standard (ISO-3166-1/2/3)

~_* ISO-3166-1:numeric was removed because it has possible collisions with geohashes_~

Primary issues I found with using NIP-32 for geotags:

1. Tag Bloat 
2. Lacks self-description
3. Redundancy in both directions, both publishing and filtering. 
4. Overly complex and verbose. 

Implementations:
1. **[COMPLETE]** Default output of `nostr-geotags` is in line with this NIP ~I'll add an option to `nostr-geotags` that produces tags proposed here instead of the NIP-32 tags.~


vitorpamplona:
hum... some clients may crash on `g`s that are not geohashes. But after they fix it, this looks ok to me. 

You know.. all these `L`/`l` and `G`/`g` shenanigans could have been solved by a more generic filtering scheme https://github.com/nostr-protocol/nips/issues/267#issuecomment-1457107761. Just saying... :)

arthurfranca:
`G` would be needed just if you wanted to request e.g. "All events that have filled `g` tag with any city name".

It won't really be combined with the `g` tag as expected, for example `{ #G: ["countryCode"], #g: ["NA"] }` will include both events from Namibia and all events from North America that also have country code filled with any value.

That's why all possible g values must not have ambiguous meanings (NA shouldn't mean a country or a continent).

So we shouldn't for instance represent a country with 4 standards, and instead should pick just one that won't collide with other g entities.

dskvr:
> It won't really be combined with the g tag as expected, for example { #G: ["countryCode"], #g: ["NA"] } will include both events from Namibia and all events from North America that also have country code filled with any value.

This is not an issue. You post-process the results using `gTag[2]`, which is also required by `NIP-32`

> So we shouldn't for instance represent a country with 4 standards

It's not being represented with 4 standards. ISO-3166 is one standard, that has multiple drafts, and different ways to express the country.

> and instead should pick just one that won't collide with other g entities.

Standards take years, when a better one exists, this NIP can be superceded.

fiatjaf:
Looks good.

dskvr:
`nostr-geotags` now produces _tags_ in alignment with this nip.

yagt.md
```diff
@@ -0,0 +1,91 @@
...
+    [ "g", "u2mwdd8q", "geohash"  ],
+    [ "g", "u2mwdd8", "geohash"  ],
+    [ "g", "u2mwdd", "geohash"  ],
+    [ "g", "u2mwd", "geohash"  ],
+    [ "g", "u2mw", "geohash"  ],
+    [ "g", "u2m", "geohash"  ],
+    [ "g", "u2", "geohash"  ],
+    [ "g", "u", "geohash"  ],
+    [ "G", "countryCode" ],
+    [ "g", "AI", "countryCode", "alpha-2" ],
```
fernandolguevara, line 74: can we move the component descriptor to be last element on the array `[ "g", "AI", "alpha-2", "countryCode"]` ?

dskvr: Why?

fernandolguevara: bc it is easy to parse the values ​​knowing that the last element is the component descriptor

dskvr: Other than making the specification both unwriteable and unreadable, this can already be achieved because the _label_ (component descriptor) is ALWAYS at `tag[2]`, as you have suggested it would only sometimes be at `tag[2]`. If you are requesting this because you have built a parser that makes this assumption, be aware that it will not work for the majority of tags that exist on nostr. We primarily use key positions.

But I digress, I was going to suggest removing the _format_ key altogether anyways, because there will be no collisions and the format can be assumed based on the string anyways. + removes bloat.

Keep the prior comments in mind though.

dskvr:
Removed `tag[3]` from NIP

- less bloat
- was unneeded, especially since the `g` namespace is collision free.

yagt.md
dskvr: Is now moot with recent update.

dskvr:
`nostr-geotags` source and docs updated to support omission of `tag[3]` from geotags.

yagt.md
```diff
@@ -0,0 +1,90 @@
...
+    [ "G", "gh" ],
+    [ "g", "u2mwdd8q4", "gh"  ],
+    [ "g", "u2mwdd8q", "gh"  ],
+    [ "g", "u2mwdd8", "gh"  ],
+    [ "g", "u2mwdd", "gh"  ],
+    [ "g", "u2mwd", "gh"  ],
+    [ "g", "u2mw", "gh"  ],
+    [ "g", "u2m", "gh"  ],
+    [ "g", "u2", "gh"  ],
+    [ "g", "u", "gh"  ],
```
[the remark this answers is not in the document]
dskvr: That would be ideal, but then limits which relays can accept geohashes. This is more or less what everyone has been doing for geospatial data discovery for a while now (at least a year)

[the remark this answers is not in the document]
vitorpamplona: Or just design a Nostr filter that follows a SQL `LIKE` operator.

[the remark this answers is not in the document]
dskvr: Yeah, it makes sense. Until a NIP is drafted and adopted implemented however, this is the _only_ interoperable method to make geotags usable in a geospatial way (that I am aware of)

staab:
This seems fine, pretty clean and clear, although my preference would be to make the following changes:

- Use `L` and `l`. It doesn't really matter that `G` labels are related to a particular domain, they're exactly equivalent to regular labels.
- Don't overload `g`. The current version is very much not a label, and is super useful as such. This avoids having to deal with backwards compatibility.
- Instead of referring to NIP 52 for the definition of `g`, have NIP 52 refer to this NIP for defining `g`.

dskvr:
> Use L and l. It doesn't really matter that G labels are related to a particular domain, they're exactly equivalent to regular labels.

@fiatjaf 

> Don't overload g. The current version is very much not a label, and is super useful as such. This avoids having to deal with backwards compatibility.

As it is written, `NIP-115` would be compatible with all prior `NIP-52` implementations and clients. As suggested, all prior `NIP-52` content would be broken. 

> Instead of referring to NIP 52 for the definition of g, have NIP 52 refer to this NIP for defining g.

Makes sense.

staab:
> My primary argument against this is collisions.

The only possible danger is that _only for events using multiple `L` tags_, you may fetch some false positives. But that should happen rarely enough that you can just discard invalid results by re-validating client-side using the `l` tag's second parameter.

> As suggested, all clients would have to implement two patterns.

I don't understand this. My suggestion is that you don't change `g` at all, which would only result in one pattern existing.

dskvr:
> I don't understand this. My suggestion is that you don't change g at all, which would only result in one pattern existing

`NIP-32`: If I have a client that implements `NIP-52`, I now have to support both `g` and `NIP-32`
`NIP-115`: If I have a client that implements `NIP-52`, I simply have to 
  1. Add a condition to ignore `NIP-115`, all previous and future content works.
  2. Support `NIP-115`, which is backwards compatible with `NIP-52` `g` tags. All previous content works, all new content works.
  
> ...and is super useful as such

As written this NIP is strictly additive, so this actually statement doesn't actuall change. Being nostr, I could (and am) publishing events with `g` tags that are not geohashes, and so, these clients _should_ be defensive already. 

I would argue that `g` was allocated without full consideration over geodata, was added via a very specific use case without full consideration over the domain and so limiting the `geo` tag to only geospatial data whilst relegating `ISO-3166` to a generalized pool of values is a disservice to the protocol.

dskvr:
Primary updates  

1. Moved everything to `G` tag.
2. Added `subRegionCode` and `interRegionCode`, 
3. Added `numeric` again since there is no more collision with geohashes.
4. Changed format of `regionCode` to align with spec. 
5. Removed `cityName` due to a lack of standards.

115.md
```diff
@@ -0,0 +1,100 @@
...
+
+This NIP adds a new indexable tag `G`. 
+
+## Motivation
+The core aim of this NIP is to establish a unified and efficient method for geocoded events. While attempting to implement `ISO-3166` with NIP-32, it became evident that more specific guidelines were needed for geocoding events. This NIP is designed to provide a clear and standardized format for geocoded events, promoting uniformity and effectiveness of geodata across nostr.
+
+## Rationale
+Developing a dedicated geotagging NIP offers several advantages over the general application of NIP-32 for geographical data:
+
+- **Full ISO-3166 support** NIP-115 is an objective, non-opinionated complete implementation of ISO-3166 that is compatible with nostr filters with no known collisions.
```
dskvr, line 13: unopinionated*

fernandolguevara (approved):

PR #953, opened 2023-12-29
Clarify the anatomy of Nostr Wallet Connect URIs

Egge21M:
I stumbled upon some issues while working on a NWC-centric project and decided that this NIP might need some clarification.

However clarifying might not be enough, as implementations in the wild already diverted from this NIP. Some providers generate URIs that have `pubkey` as authority. Other do not URL-encode relay addresses. Inconsistencies like this make it hard to implement.


vitorpamplona:
LGTM

Egge21M:
Every implementation I found so far places the wallet pubkey as authority instead of as path. I wonder if it makes sense to adjust the spec to reflect that defacto standard.

It also does not require mental gymnastics to think of the wallet provider as governing authority of that URI.

This would require a change from 

'nostr+walletconnect:<pubkey>' to 'nostr+walletconnect://<pubkey>'

47.md
```diff
@@ -78,11 +78,13 @@ If the command was successful, the `error` field must be null.
 - `OTHER`: Other error.
 
 ## Nostr Wallet Connect URI
-**client** discovers **wallet service** by scanning a QR code, handling a deeplink or pasting in a URI.
+**client** discovers **wallet service** by it's URI; either by scanning a QR code, handling a deeplink or pasting in the URI.
```
jklein24, line 81:
```suggestion
**client** discovers **wallet service** by its URI; either by scanning a QR code, handling a deeplink or pasting in the URI.
```

PR #954, opened 2023-12-30
index zap senders with P tag

pablof7z:
A lot of blood has been shed by the lack of an index to find zaps sent by a pubkey.

Mutiny wallet is running a primal index server just to be able to run this simple query.

CHILDREN ARE DYING!

### DIFF
--- a/57.md
+++ b/57.md
@@ -111,6 +111,7 @@ When a client sends a `zap request` event to a server's lnurl-pay callback URL, 
 5. There should be a `relays` tag with the relays to send the `zap receipt` to.
 6. If there is an `amount` tag, it MUST be equal to the `amount` query parameter.
 7. If there is an `a` tag, it MUST be a valid event coordinate
+8. There MUST be 0 or 1 `P` tags. If there is one, it MUST be equal to the `zap receipt`'s `pubkey`.
 
 The event MUST then be stored for use later, when the invoice is paid.
 
@@ -128,7 +129,7 @@ The following should be true of the `zap receipt` event:
 
 - The `content` SHOULD be empty.
 - The `created_at` date SHOULD be set to the invoice `paid_at` date for idempotency.
-- `tags` MUST include the `p` tag AND optional `e` tag from the `zap request` AND optional `a` tag from the `zap request`.
+- `tags` MUST include the `p` tag (zap recipient) AND optional `e` tag from the `zap request` AND optional `a` tag from the `zap request` AND optional `P` tag from the pubkey of the zap request (zap sender).
 - The `zap receipt` MUST have a `bolt11` tag containing the description hash bolt11 invoice.
 - The `zap receipt` MUST contain a `description` tag which is the JSON-encoded invoice description.
 - `SHA256(description)` MUST match the description hash in the bolt11 invoice.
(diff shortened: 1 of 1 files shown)


TonyGiorgio:
We use primal for zaps sent and zaps received by collection of pubkeys. 

Specifically interested in all economic activity by someone's follow list.

benthecarman:
Definite concept ack

jb55:
ack. huge oversight in the original spec.

57.md
```diff
@@ -128,7 +129,7 @@ The following should be true of the `zap receipt` event:
 
 - The `content` SHOULD be empty.
 - The `created_at` date SHOULD be set to the invoice `paid_at` date for idempotency.
-- `tags` MUST include the `p` tag AND optional `e` tag from the `zap request` AND optional `a` tag from the `zap request`.
+- `tags` MUST include the `p` tag (zap recipient) AND optional `e` tag from the `zap request` AND optional `a` tag from the `zap request` AND optional `P` tag from the pubkey of the zap request (zap sender).
```
benthecarman, line 132: the uppercase `P` can be pretty confusing, might be better as something like `sender`

jb55:
On Sat, Dec 30, 2023 at 08:38:35PM -0800, benthecarman wrote:
>the uppercase `P` can be pretty confusing, might be better as something like `sender`

Then it wouldn't be efficiently queryable. most relays only index single-char tags.

fiatjaf (approved):

v0l:
Probably also worth mentioning that zap services can add this tag to the receipt and its not needed in zap request at all?

vitorpamplona:
LGTM

It is worth mentioning that the Zapper service can `P`-tag incorrectly and/or malicious actors can use this to spam everyone. In other words, the new tag is just a hint to who the sender is and not proof that that person sent it.

v0l:
> LGTM
> 
> It is worth mentioning that the Zapper service can `P`-tag incorrectly and/or malicious actors can use this to spam everyone. In other words, the new tag is just a hint to who the sender is and not proof that that person sent it.

zaps are not proof of anything

benthecarman:
Implemented in rust-nostr: https://github.com/rust-nostr/nostr/pull/230

pablof7z:
Implemented in nostdress: https://github.com/believethehype/nostdress/pull/28

P-tagged Zap: https://njump.me/nevent1qqs2rnyrwjf3yfl6h7j0wg2zke6p0unh34c8pupnmwcll4uqascfm8qprpmhxue69uhhyetvv9ujumn0wdmksetjv5hxxmmdqgsgyq7wmt5892vd437h99ta4sd8gp5x74pmy2wdn0r3vq639dn6qls4tujt9

yukibtc:
I see the `P` tag really confusing.

What if we use the `d` tag, composed with a prefix + the pubkey? Something like: `["d", "zap-sender:<public-key>"]`?

So the filter will be: `{ "kind": 9735, "#d": "zap-sender:<pk>" }`

What do you think?

benthecarman:
For added context, if you look at my rust-nostr PR, this change is really ugly for typed languages where we already have a type defined for p, so we need to name this one UpperP

PR #955, opened 2023-12-31
Create 100.md

nextdebug:
## NOSTR Decentralized Advertising Network (NOSTR-DAN)
-------------------
`draf` `optional`

This NIP defines the `kind:20100`: a parameterized event dispatched to a listener or pool of listeners to offer space for advertising,
a listener can answer accepting the offer and contracting the space.

## REVIEW

- NOSTR-DAN needs only to members to work:
  - The client that offers spaces to the advertisers
  - The listeners (advertisers) that contract the spaces to advertise
- NOSTR-DAN community can create custom schemas for your ads, allowing the dynamic communication required by competitive industries and unpredictable market changes
- Members of the NOSTR-DAN can apply the whitelist and bannedlist to filter ads

### DIFF
--- a/100.md
+++ b/100.md
@@ -1,0 +1,183 @@
+# NIP-100
+
+
+## NOSTR Decentralized Advertising Network (NOSTR-DAN)
+-------------------
+`draf` `optional`
+
+This NIP defines the `kind:20100`: a parameterized event dispatched to a listener or pool of listeners to offer space for advertising,
+a listener can answer accepting the offer and contracting the space.
+
+## Motivation
+-------------------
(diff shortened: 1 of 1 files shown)


Egge21M:
As someone who spent most his life working on header bidding and digital advertisement, I spent many hours thinking about solutions involving Lightning and nostr.

I came to the conclusions that a Lightning-based OSS solution is required, but I don't think nostr should be involved. The amount of data that needs to be transmitted (we are talking hundred thousands if not millions of bids and calls each hour) will bloat relays. 

Additionally latency is just to important. A few hundred ms can mean difference of order of magnitude.

Issue #958, opened 2023-12-31
The order of tags in follows/blocks/mutes/bookmarks/etc


alexgleason:
Should new tags be added to the end or the beginning?

```js
// Adding the new tag at the end:
const tags = [...event.tags, ["p", "123"]];

// Adding the new tag at the beginning:
const tags = [["p", "123"], ...event.tags];
```

I want to display newest items at the top in the UI. So it raises the question about the order of tags.

I don't care whether it's top or bottom, but we should standardize one or the other to avoid inconsistent behavior in clients. What is the most common way?

vitorpamplona:
Bottom is probably people's first way of coding this without even thinking about the order, so I would go with that option.

PR #960, opened 2024-01-01
NIP-02, NIP-51: new tags should be added to the end of the list

alexgleason:
Fixes https://github.com/nostr-protocol/nips/issues/958

### DIFF
--- a/02.md
+++ b/02.md
@@ -27,6 +27,8 @@ For example:
 
 Every new following list that gets published overwrites the past ones, so it should contain all entries. Relays and clients SHOULD delete past following lists as soon as they receive a new one.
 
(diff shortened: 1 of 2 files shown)


staab (approved):

fiatjaf (approved):

monlovesmango:
agree new tags should be added to end of list. however, do we want to say all lists should strictly be chronological order? or asked another way, do we see no reason to allow users to order the lists themselves? 

personally can't think of any strong use case for allowing users to manually order lists, but maybe others would like to be able to do this for some reason?

monlovesmango (approved):

PR #963, opened 2024-01-03
Improving Nip01 Text

vitorpamplona:
I have added some missing info and reduced some of the verbosity of the text. 

Is this better? 

Read [here](https://github.com/vitorpamplona/nips/blob/improving-nip01/01.md)

### DIFF
--- a/01.md
+++ b/01.md
@@ -1,23 +1,23 @@
 NIP-01
 ======
 
-Basic protocol flow description
--------------------------------
+Nostr Protocol
+--------------
 
 `draft` `mandatory`
 
-This NIP defines the basic protocol that should be implemented by everybody. New NIPs may add new optional (or mandatory) fields and messages and features to the structures and flows described here.
+This NIP defines the complete protocol that MUST be implemented by everybody. New NIPs may add new optional (or mandatory) fields, messages and features to the structures and flows described here.
 
-## Events and signatures
+## User and Events
 
-Each user has a keypair. Signatures, public key, and encodings are done according to the [Schnorr signatures standard for the curve `secp256k1`](https://bips.xyz/340).
+Each user has a keypair and is identified by their public key. 
 
-The only object type that exists is the `event`, which has the following format on the wire:
+Event is only object type available. It is a hashed and signed payload with the following format:
 
 ```json
 {
-  "id": <32-bytes lowercase hex-encoded sha256 of the serialized event data>,
-  "pubkey": <32-bytes lowercase hex-encoded public key of the event creator>,
+  "id": <32-byte lowercase hex-encoded sha256 of the serialized event data>,
+  "pubkey": <32-byte lowercase hex-encoded public key used to sign>,
   "created_at": <unix timestamp in seconds>,
   "kind": <integer between 0 and 65535>,
   "tags": [
@@ -25,11 +25,15 @@ The only object type that exists is the `event`, which has the following format 
     ...
   ],
   "content": <arbitrary string>,
-  "sig": <64-bytes lowercase hex of the signature of the sha256 hash of the serialized event data, which is the same as the "id" field>
+  "sig": <64-byte lowercase hex of the signature of the "id" field>
 }
 ```
 
-To obtain the `event.id`, we `sha256` the serialized event. The serialization is done over the UTF-8 JSON-serialized string (which is described below) of the following structure:
+### Signature
+
+Signatures and encodings are done according to the [Schnorr signatures standard for the curve `secp256k1`](https://bips.xyz/340).
+
+To obtain the hash `.id`, we `sha256` the UTF-8 bytearray of a JSON-serialized string with the following structure:
 
 ```
 [
@@ -42,21 +46,20 @@ To obtain the `event.id`, we `sha256` the serialized event. The serialization is
 ]
 ```
 
-To prevent implementation differences from creating a different event ID for the same event, the following rules MUST be followed while serializing:
-- No whitespace, line breaks or other unnecessary formatting should be included in the output JSON.
-- No characters except the following should be escaped, and instead should be included verbatim:
-  - A line break, `0x0A`, as `\n`
-  - A double quote, `0x22`, as `\"`
-  - A backslash, `0x5C`, as `\\`
-  - A carriage return, `0x0D`, as `\r`
-  - A tab character, `0x09`, as `\t`
-  - A backspace, `0x08`, as `\b`
-  - A form feed, `0x0C`, as `\f`
-- UTF-8 should be used for encoding.
+The JSON serialization MUST follow this rules: 
+- Minified: No whitespace, line breaks or other unnecessary formatting used
+- Escape Set: ONLY the following characters MUST be escaped:
+  - line break, `0x0A`, as `\n`
+  - double quote, `0x22`, as `\"`
+  - backslash, `0x5C`, as `\\`
+  - carriage return, `0x0D`, as `\r`
+  - tab character, `0x09`, as `\t`
+  - backspace, `0x08`, as `\b`
+  - form feed, `0x0C`, as `\f`
 
 ### Tags
 
-Each tag is an array of strings of arbitrary size, with some conventions around them. Take a look at the example below:
+Each tag is an array of strings of arbitrary size. Their meaning is determined by the event `.kind` and defined in NIPs in this repository.
 
 ```json
 {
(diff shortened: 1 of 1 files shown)


arthurfranca:
It looks better.
One important thing that got deleted was the implied **descending** sort order if using `limit` filter when the original text said "**last** `n` events ordered by the `created_at`"

vitorpamplona:
Yep, it was duplicated. I kept only in the json descriptor.

01.md
```diff
@@ -42,136 +47,161 @@ To obtain the `event.id`, we `sha256` the serialized event. The serialization is
...
+Relays send 3 types of messages during a subscription:  
 
-Relays can send 4 types of messages, which must also be JSON arrays, according to the following patterns:
-
-  * `["EVENT", <subscription_id>, <event JSON as defined above>]`, used to send events requested by clients.
-  * `["OK", <event_id>, <true|false>, <message>]`, used to indicate acceptance or denial of an `EVENT` message.
+  * `["EVENT", <subscription_id>, <JSON-serialized string of the event>]`, used to send events requested by clients.
   * `["EOSE", <subscription_id>]`, used to indicate the _end of stored events_ and the beginning of events newly received in real-time.
-  * `["CLOSED", <subscription_id>, <message>]`, used to indicate that a subscription was ended on the server side.
+  * `["CLOSED", <subscription_id>, <prefix>:<message>]`, used to indicate that a subscription was ended on the server side.
```
fiatjaf, line 164: why isn't the `OK` message listed here?

fiatjaf:
I like this, let's do it.

01.md
vitorpamplona: I split them into a subscription protocol and a publishing protocol.

vitorpamplona:
@fiatjaf maybe later we can move the NOTICE part to a non-mandatory NIP. They are cool, but I think we must make NIP-01's feature set as small as possible.

01.md
```diff
@@ -1,177 +1,216 @@
...
+Nostr Protocol
+--------------
 
 `draft` `mandatory`
 
-This NIP defines the basic protocol that should be implemented by everybody. New NIPs may add new optional (or mandatory) fields and messages and features to the structures and flows described here.
+This NIP defines the mandatory part of the Nostr protocol. New NIPs may add new optional (or mandatory) fields, messages, and features to the structures and flows described here.
 
-## Events and signatures
+# Events
```
jb55, line 11: I always call these Notes instead of Events in my codebases, but maybe the ship has sailed on that one because of ["EVENT",...]

README.md
```diff
@@ -30,7 +30,7 @@ They exist to document what may be implemented by [Nostr](https://github.com/nos
 - [NIP-07: `window.nostr` capability for web browsers](07.md)
 - [NIP-08: Handling Mentions](08.md) --- **unrecommended**: deprecated in favor of [NIP-27](27.md)
 - [NIP-09: Event Deletion](09.md)
-- [NIP-10: Conventions for clients' use of `e` and `p` tags in text events](10.md)
+- [NIP-10: Text Notes](10.md)
```
jb55, line 33: NIP-10 is more about threads than Text Notes, you can have threaded events of different kinds.

01.md
fiatjaf: "event" was probably a poor name choice, but "relay" too.

vitorpamplona:
FYI, I just noticed NIP01 never defines the type of each attribute in the Relay protocol messages. I think this new hack of using the `js` syntax highlight can be quite effective to separate type and description in a <> inside the json.

fiatjaf:
What if we moved the basic protocol stuff to NIP-00 and used NIP-01 for the "basic microblogging client flow" including kind 0 and kind 1 -- then we can even include implementation recommendations like to which relays connect and how to better load threads, for example? Even conflicting alternative recommendations if that's the case.

staab:
I think if we're going to restructure things that heavily, we might as well get rid of the NIP system

vitorpamplona:
I do think it makes sense to separate the basic microblogging stuff out of the core protocol discussion but I am not sure the "basic microblogging stuff" is a well-defined set right now. For instance, I don't consider `kind:0` to be "microblogging stuff".

pablof7z:
ACK on moving nip-01 to nip-00 and extracting the microblogging flow to it's own NIP.

I agree with @vitorpamplona in that `kind:0` is not part of the microblogging flow

01.md
alexgleason: I thought kind 1 events are "notes". All notes are events but not all events are notes.

alexgleason:
> What if we moved the basic protocol stuff to NIP-00 

It will make NIP-11's `supported_nips` potentially wrong. One thing I noticed after a bunch of NIPs were combined into NIP-01 is that you can no longer tell which relays support replaceable events or parameterized events.

vitorpamplona:
> which relays support replaceable events or parameterized events

In theory, both are required to be supported these days.

pablof7z:
> It will make NIP-11's `supported_nips` potentially wrong. 

Only as far as `supported_nips` not indicating that NIP-00 is supported, but given that NIP-00 is required that can be implied (otherwise it's not nostr and you wouldn't see a `supported_nips` at all)

The only caveat is that NIP-01 must remain backward-compatible so relays don't lie when they say they support NIP-01.

vitorpamplona:
This is also ready to merge if people agree it's better than the current NIP-01 description.

01.md
```diff
@@ -1,177 +1,235 @@
...
+| Name          | Range                             | SHOULD retain                                          |
+| ------------- | --------------------------------- | ------------------------------------------------------ |
+| Regular       | ` 1000 <= kind < 10000`           | All events                                             |
+| Replaceable   | `10000 <= kind < 20000` OR `0, 3` | The newest event for each `pubkey` and `kind`          |
+| Ephemeral     | `20000 <= kind < 30000`           | Nothing                                                |
+| Parameterized | `30000 <= kind < 40000`           | The newest event for each `pubkey`, `kind` and `d`-tag |
 
-### Kinds
+Parameterized Replaceable events MUST include a `d`-tag with an identifier value to be used as a reference. Tag values that reference the newest replaceable MUST use this format: `<kind>:<32-byte lowercase hex of a pubkey>:<d-tag value>`.
```
schmijos, line 82: I like this emphasis on #d (see my confusion here: https://github.com/nostr-protocol/nips/pull/982)

01.md
```diff
@@ -1,177 +1,235 @@
...
+Each tag is an array of strings of arbitrary size. Their meaning is determined by the event `.kind`. Tags with the same name might have entirely different meanings in different kinds. 
 
-In case of replaceable events with the same timestamp, the event with the lowest id (first in lexical order) should be retained, and the other discarded.
+The first element of the tag array is referred to as the tag _name_ and the second as the tag _value_. All elements after the second do not have a conventional name.
 
-When answering to `REQ` messages for replaceable events such as `{"kinds":[0],"authors":[<hex-key>]}`, even if the relay has more than one version stored, it SHOULD return just the latest one.
+All single-letter (only English alphabet letters: a-z, A-Z) tag names are indexed by relays for faster queries. 
 
-These are just conventions and relay implementations may differ.
+This NIP defines the format of 3 standard tags: `e`, `p`, and `a`. Tags `e`, `p` can be used to reference events and pubkeys respectively, and tag `a` references the latest version of a replaceable event. Clients MAY `a`- and `e`-tag parameterized replaceables simultaneously.
```
schmijos, line 96:
```suggestion
This NIP defines the format of 4 standard tags: `e`, `p`, `a` and `d`. Tags `e`, `p` can be used to reference events and pubkeys respectively, and tag `a` references the latest version of a replaceable event. Clients MAY set `a`- and `e`-tag parameterized replaceables simultaneously.
```

01.md
```diff
@@ -1,177 +1,235 @@
...
 
-## Communication between clients and relays
+| Name | Value                                                      | Other Params            |
+| ---- | ---------------------------------------------------------- | ----------------------- |
+| `e`  | `<32-byte lowercase hex of an event id>`                   | `<relay URL, optional>` |
+| `p`  | `<32-byte lowercase hex of a pubkey>`                      | `<relay URL, optional>` |
+| `a`  | `<kind>:<32-byte lowercase hex of a pubkey>:`              | `<relay URL, optional>` |
+| `a`  | `<kind>:<32-byte lowercase hex of a pubkey>:<d-tag value>` | `<relay URL, optional>` |
```
schmijos, line 103: Where is now #d ?

01.md
vitorpamplona: Do we really need to duplicate the definition here? It was already defined in the previous section.

01.md
schmijos: Let me explain why I have difficulties with the "d" tag.
I looked at the README and saw: "Aha, there's an official #d. Let me have a look at what this means."
I then go to NIP-01 where it's supposed to be defined. I ctrl-f through the page on GitHub and don't find it.

After your PR here, when I ctrl-f, I'll still be confused because I'm lazy and I'll find two things:
* A statement saying 
  > This NIP defines the format of 3 standard tags: `e`, `p`, and `a`.
  
  Where the heck is now #d?
* A table with `e`, `p` and `a`
  Why the heck is it missing #d?

01.md
```diff
@@ -1,177 +1,235 @@
 NIP-01
 ======
 
-Basic protocol flow description
--------------------------------
+Nostr Protocol
```
AsaiToshiya, line 4:
```suggestion
Basic Protocol
```

The title "Nostr Protocol" is pretty cool, but it is meaningless, because I think all NIPs are descriptions of the Nostr protocol.

vitorpamplona: Yeah. I was trying to find a way to say that this is the only mandatory part everyone must code. Everything else is optional. "The Basic Flow" before wasn't doing it. 

The other idea was to avoid the assumption that the "Nostr Protocol" includes all NIPs, which I think makes it too big for our own good.

pablof7z: > Yeah. I was trying to find a way to say that this is the only mandatory part everyone must code. Everything else is optional. "The Basic Flow" before wasn't doing it. 
> 
> The other idea was to avoid the assumption that the "Nostr Protocol" includes all NIPs, which I think makes it too big for our own good. 
> 
> 

This.

NIP-01 is the protocol and the other NIPs are optional conventions. One could even say, _layers_, built on top of NIP-01

AsaiToshiya: Yeah, that make sense to me. Thanks.

01.md
```diff
@@ -6,11 +6,20 @@ Nostr Protocol
...
 `draft` `mandatory`
 
-This NIP defines the mandatory part of the Nostr protocol. New NIPs may add new optional (or mandatory) fields, messages, and features to the structures and flows described here.
+This NIP defines the mandatory part of the Nostr protocol. New NIPs may add new fields, message kinds, and features to the structures and flows described here.
 
-# Events
+# Signed Events
 
-Event is the only object type available. It is a hashed and signed payload in the following format:
+Events are the only data type transferred across the network. They consist in 4 main attributes:
```
mikedilger: 5 main attributes

staab: 3 m'lord, 3

schmijos: 6 if you count `id`?

mikedilger: 5 things are hashed:  pubkey, created_at, kind, tags, and content.   The other 2 are generated:  id, sig

vitorpamplona:
Closing this due to partial adoption via other PRs. Though I still think the Relay section is better described here (separated by action) than in the current nip01.

PR #970, opened 2024-01-07
Practical guidance

ionextdebug:
Add a practical tutorial about how to implement NIP-01 initial part. Important to make newcomers software developers be confortable in the baby steps or "get your feet wet" stage.


AsaiToshiya:
This seems helpful, but I think it is not to write here.

vitorpamplona:
No need for it

PR #976, opened 2024-01-08
NIP-46 `create_account` command

pablof7z:
This PR introduces a couple things:

* Flow to use a NIP-05 to connect client<>signer
* `create_account` method, or like @fiatjaf likes to call with, the "OAuth-like flow"
* An (optional) `auth_url` response that can greatly improve the flow of the user having to do something to authorized a client's signature

### DIFF
--- a/46.md
+++ b/46.md
@@ -32,6 +32,10 @@ nostrconnect://<client-key-hex>?relay=wss://...&metadata={"name":"...", "url": "
 
 The signer scans the QR code and sends a `connect` message to the client in the specified relays.
 
+### Using a NIP-05
+
+Clients can choose to connect with a backend with just a NIP-05 for the target user. The user enters a NIP-05 Nostr Address, which the client uses to resolve to a pubkey. NIP-05's nostr.json SHOULD include a `nip46` root element of the pubkey and the relays that should be used to communicate with the remote signer for this pubkey.
+
 ## Event payloads
 
 Event payloads are [NIP-04](04.md)-encrypted JSON blobs that look like JSONRPC messages (their format is specified inside the `.content` of the event formats nelow).
@@ -63,6 +67,23 @@ And the events the remote signer sends to the client have the following format:
 ```
 
 The signer key will always be the key of the user who controls the signer device.
+
+### Auth URL response
+A signer can reply with a `result: "auth_url", error: "<url>"`. This response can provide an easier way for the user to authorize this client to perform an action. Clients can redirect/popup this URL so the user can take an action directly on the signer backend.
+
+When redirecting, the client can choose to add a `callbackUrl` query string parameter to tell the signer when to send the user after an action has been performed.
+
+#### Example flow:
+
+* User goes to a new client
+* Logins by typing their nip-05 "hello@example.com"
+* Client gets example.com/.well-known/nostr.json?name=hello
+* Client finds nip46 relays wss://relay.example.com and pubkey <123>
+* Client sends `connect` with remote pubkey `<123>`
+* Signer sends back `result: 'auth_url", error: "https://auth.examople.com/ABCD"`
+* Client opens popup `https://auth.examople.com/ABCD`
+* User authorizes the client to sign on behalf of pubkey <123>
+* Client's `connect` RPC call is `ack`ed.
 
 ### Methods
 
@@ -96,4 +117,10 @@ The signer key will always be the key of the user who controls the signer device
 - **ping**
   - params: []
   - result: `"pong"`
+- **create_account**
+  - params: `["username", "domain", "email"]
+  - result: `generated-pubkey`
+
+#### Account creation
+This method creates a new key inside the backend. All params are optional. Username+domain can be specified so a NIP-05 record is created to point to this key. Email is an optional parameter that allows the backend to check if this user has already created an account in this backend. The `create_account` method should be sent to the pubkey of the bunker, which should match the root NIP-05 entry (`_domain`).
 
--- a/46.md
+++ b/46.md
@@ -71,7 +71,7 @@ The signer key will always be the key of the user who controls the signer device
 ### Auth URL response
 A signer can reply with a `result: "auth_url", error: "<url>"`. This response can provide an easier way for the user to authorize this client to perform an action. Clients can redirect/popup this URL so the user can take an action directly on the signer backend.
 
-When redirecting, the client can choose to add a `callbackUrl` query string parameter to tell the signer when to send the user after an action has been performed.
+When redirecting, the client can choose to add a `callbackUrl` query string parameter to tell the signer where to send the user after an action has been performed.
 
 #### Example flow:
 
--- a/46.md
+++ b/46.md
@@ -122,5 +122,5 @@ When redirecting, the client can choose to add a `callbackUrl` query string para
   - result: `generated-pubkey`
 
 #### Account creation
-This method creates a new key inside the backend. All params are optional. Username+domain can be specified so a NIP-05 record is created to point to this key. Email is an optional parameter that allows the backend to check if this user has already created an account in this backend. The `create_account` method should be sent to the pubkey of the bunker, which should match the root NIP-05 entry (`_domain`).
+This method creates a new key inside the backend. All params are optional. Username+domain can be specified so a NIP-05 record is created to point to this key. Email is an optional parameter that allows the backend to check if this user has already created an account in this backend. The `create_account` method should be sent to the pubkey of the bunker, which should match the root NIP-05 entry (`_@domain`).
 
--- a/46.md
+++ b/46.md
@@ -80,7 +80,7 @@ When redirecting, the client can choose to add a `callbackUrl` query string para
 * Client gets example.com/.well-known/nostr.json?name=hello
 * Client finds nip46 relays wss://relay.example.com and pubkey <123>
 * Client sends `connect` with remote pubkey `<123>`
-* Signer sends back `result: 'auth_url", error: "https://auth.example.com/some-long-secret-path"`
+* Signer sends back `result: "auth_url", error: "https://auth.example.com/some-long-secret-path"`
 * Client opens popup `https://auth.examople.com/ABCD`
 * User authorizes the client to sign on behalf of pubkey <123>
 * Client's `connect` RPC call is `ack`ed.
--- a/46.md
+++ b/46.md
@@ -118,7 +118,7 @@ When redirecting, the client can choose to add a `callbackUrl` query string para
   - params: []
   - result: `"pong"`
 - **create_account**
-  - params: `["username", "domain", "email"]
+  - params: `["name", "domain", "email"]
   - result: `generated-pubkey`
 
 #### Account creation
--- a/46.md
+++ b/46.md
@@ -80,7 +80,7 @@ When redirecting, the client can choose to add a `callbackUrl` query string para
 * Client gets example.com/.well-known/nostr.json?name=hello
 * Client finds nip46 relays wss://relay.example.com and pubkey <123>
 * Client sends `connect` with remote pubkey `<123>`
-* Signer sends back `result: 'auth_url", error: "https://auth.examople.com/ABCD"`
+* Signer sends back `result: 'auth_url", error: "https://auth.example.com/some-long-secret-path"`
 * Client opens popup `https://auth.examople.com/ABCD`
 * User authorizes the client to sign on behalf of pubkey <123>
 * Client's `connect` RPC call is `ack`ed.
--- a/46.md
+++ b/46.md
@@ -81,7 +81,7 @@ When redirecting, the client can choose to add a `callbackUrl` query string para
 * Client finds nip46 relays wss://relay.example.com and pubkey <123>
 * Client sends `connect` with remote pubkey `<123>`
 * Signer sends back `result: "auth_url", error: "https://auth.example.com/some-long-secret-path"`
-* Client opens popup `https://auth.examople.com/ABCD`
+* Client opens popup for `auth_url` (for example `https://auth.example.com/ABCD`)
 * User authorizes the client to sign on behalf of pubkey <123>
 * Client's `connect` RPC call is `ack`ed.
 


erskingardner (approved):

46.md
```diff
@@ -64,6 +68,23 @@ And the events the remote signer sends to the client have the following format:
...
+A signer can reply with a `result: "auth_url", error: "<url>"`. This response can provide an easier way for the user to authorize this client to perform an action. Clients can redirect/popup this URL so the user can take an action directly on the signer backend.
+
+When redirecting, the client can choose to add a `callbackUrl` query string parameter to tell the signer when to send the user after an action has been performed.
+
+#### Example flow:
+
+* User goes to a new client
+* Logins by typing their nip-05 "hello@example.com"
+* Client gets example.com/.well-known/nostr.json?name=hello
+* Client finds nip46 relays wss://relay.example.com and pubkey <123>
```
fiatjaf, line 81: underspecified

46.md
```diff
@@ -64,6 +68,23 @@ And the events the remote signer sends to the client have the following format:
 
 The signer key will always be the key of the user who controls the signer device.
 
+### Auth URL response
+A signer can reply with a `result: "auth_url", error: "<url>"`. This response can provide an easier way for the user to authorize this client to perform an action. Clients can redirect/popup this URL so the user can take an action directly on the signer backend.
+
+When redirecting, the client can choose to add a `callbackUrl` query string parameter to tell the signer when to send the user after an action has been performed.
```
fiatjaf, line 74:
```suggestion
When redirecting, the client can choose to add a `callbackUrl` query string parameter to tell the signer where to send the user after an action has been performed.
```

46.md
```diff
@@ -64,6 +68,23 @@ And the events the remote signer sends to the client have the following format:
 
 The signer key will always be the key of the user who controls the signer device.
 
+### Auth URL response
+A signer can reply with a `result: "auth_url", error: "<url>"`. This response can provide an easier way for the user to authorize this client to perform an action. Clients can redirect/popup this URL so the user can take an action directly on the signer backend.
```
fiatjaf, line 72: reply to what? to `connect`?

46.md
```diff
@@ -96,4 +117,10 @@ The signer key will always be the key of the user who controls the signer device
 - **ping**
   - params: []
   - result: `"pong"`
+- **create_account**
```
fiatjaf, line 120: when do you call `create_account`? before calling `connect`?

46.md
```diff
@@ -96,4 +117,10 @@ The signer key will always be the key of the user who controls the signer device
...
+  - params: `["username", "domain", "email"]
```
alexgleason, line 121: Is email so important to this flow that it's a positional argument? NIP-46 is allergic to options objects, but it would offer a lot more flexibility.

pablof7z: it's not really important and can be left empty (all params are optional); we can take it out if it feels too off.

46.md
erskingardner: > it will reply with a new pubkey <pubkey1>

How does this reply come through? Eg what type of event are we subscribing to for connect calls?

pablof7z: just the standard nip-46 protocol, a `kind:24133` p-tagging the key you are writing from (in this case the client/local key).

46.md
fiatjaf: Option objects just mean more documentation to read and more strings to write. If we have more options to add later we can add them to the end of the array and leave the email empty.

alexgleason: I like this MR. In fact I'm going to implement it. But I need to add a "reason" text field, which the user can fill in to request an account. Requests are approved manually by the admin. The admin account then creates a kind 30361 event to confirm the account of the user: https://gitlab.com/soapbox-pub/ditto/-/blob/main/docs/events.md?ref_type=heads#users 

If approval-mode is disabled, the admin account would create this event manually, and no "reason" field would be required.

alexgleason: If this MR goes as it is, I'm going to create a positional 4th argument that is an options object, including `opts.reason `

alexgleason:
There is nothing I see to stop someone from generating unlimited accounts in the system. Maybe proof-of-work, but I think signups should be more robust and spam resistant.

I have almost the exact same use-case as this proposal (to create accounts on a NIP-46 server). But to deal with spam, I want the server admin to be able to approve accounts manually at a later time. Therefore the "create_account" event needs to be a regular event, not ephemeral, so the admin can query invite requests whenever.

I've been trying to figure out if I can make it work with this proposal, or if I have to invent a different way to create accounts. I actually can't think of a way without using a regular event kind. Using a regular kind would still allow the automatic signup flow, so it's flexible to either use case, and would probably open doors later.

To me, the most important new part of the "OAuth-like flow" is the NIP-05 extension, specifically how you identify the server pubkey (the `_` user), and that a user with this name is always present in `nostr.json`. Another possible way would be to use the NIP-11 `pubkey`, but it doesn't exist at a `.well-known` path, so it would be harder to find. I like the `nostr.json` approach better.

pablof7z:
I thought about delegating the POW of sale down of account creation to a step the provider determines (hence it's not in the spec because it would be completely up to the provider) but this could be as easily as just embedding a captcha on the signup form, a small LN payment or any type of there rate limiting actions that currently exists.

46.md
alexgleason:
```suggestion
  - params: `["name", "domain", "email"]
```

Username should be `"name"` to match NIP-05.

(`"domain"` should technically be `"host"` so it includes the port, but NIP-05 calls it `"domain"` already)

Also, I wonder if instead it should be a single parameter called `nip05` to match kind 0's metadata field.

46.md
```diff
@@ -96,4 +117,10 @@ The signer key will always be the key of the user who controls the signer device
...
+  - result: `generated-pubkey`
+
+#### Account creation
+This method creates a new key inside the backend. All params are optional. Username+domain can be specified so a NIP-05 record is created to point to this key. Email is an optional parameter that allows the backend to check if this user has already created an account in this backend. The `create_account` method should be sent to the pubkey of the bunker, which should match the root NIP-05 entry (`_domain`).
```
alexgleason, line 125:
```suggestion
This method creates a new key inside the backend. All params are optional. Username+domain can be specified so a NIP-05 record is created to point to this key. Email is an optional parameter that allows the backend to check if this user has already created an account in this backend. The `create_account` method should be sent to the pubkey of the bunker, which should match the root NIP-05 entry (`_@domain`).
```

alexgleason:
The server should respond in some way. On success, a [kind 30361](https://gitlab.com/soapbox-pub/ditto/-/blob/main/docs/events.md?ref_type=heads#users) event could be published, and then embedded in an NIP-46 message back to the user. On fail, it should still a NIP-46 message describing additional measures to take (eg "confirm your email address") or denying the request for some other reason ("name unavailable")

46.md
monlovesmango, line 120: my main objection to this call is that it allows the creation of a user private key without the opportunity to force the user to back up their private key. additionally, if a user did want to back up their private key their is no clear way to direct them to do so. for small instances of nsecbunker this may not be too big a deal, but it does open the door to larger service providers owning user's keys.

also `username` and `domain` params both need validation on nsecbunker side and abstracting them into client fields feels confusing for the user when either are invalid.

how would you prevent abuse of this method? could easily be used to create numerous spam accounts and NIP05 identifiers.

generally I think key generation is important to highlight to users and it should be explicit which entities a user is delegating power over their keys to. this method would create a protocol approved way of obfuscating that, though I do agree it would likely be more convenient for the user.

monlovesmango (commented):
should NIP05 also be updated to cross post this new optional NIP46 element?

monlovesmango:
>embedding a captcha on the signup form, a small LN payment or any type of there rate limiting actions that currently exists

how would these be integrated into the `create_account` method?

46.md
```diff
@@ -64,6 +68,23 @@ And the events the remote signer sends to the client have the following format:
...
+
+#### Example flow:
+
+* User goes to a new client
+* Logins by typing their nip-05 "hello@example.com"
+* Client gets example.com/.well-known/nostr.json?name=hello
+* Client finds nip46 relays wss://relay.example.com and pubkey <123>
+* Client sends `connect` with remote pubkey `<123>`
+* Signer sends back `result: 'auth_url", error: "https://auth.example.com/some-long-secret-path"`
+* Client opens popup `https://auth.examople.com/ABCD`
```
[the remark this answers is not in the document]
pablof7z: it comes from the `auth_url` from the signer

pablof7z:
> > embedding a captcha on the signup form, a small LN payment or any type of there rate limiting actions that currently exists
> 
> how would these be integrated into the `create_account` method?

It doesn't have to be integrated in any way; the signer, interacting with the user can do whatever it wants; POW on the window, captcha, ask for a 1000 sats, whatever. That interaction doesn't need to be spec'ed out.

46.md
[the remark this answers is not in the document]
monlovesmango:
```suggestion
* Client opens popup for `auth_url` (for example `https://auth.example.com/ABCD`)
```

46.md
```diff
@@ -64,6 +68,23 @@ And the events the remote signer sends to the client have the following format:
...
+When redirecting, the client can choose to add a `callbackUrl` query string parameter to tell the signer where to send the user after an action has been performed.
+
+#### Example flow:
+
+* User goes to a new client
+* Logins by typing their nip-05 "hello@example.com"
+* Client gets example.com/.well-known/nostr.json?name=hello
+* Client finds nip46 relays wss://relay.example.com and pubkey <123>
+* Client sends `connect` with remote pubkey `<123>`
+* Signer sends back `result: 'auth_url", error: "https://auth.example.com/some-long-secret-path"`
```
monlovesmango, line 83:
```suggestion
* Signer sends back `result: "auth_url", error: "https://auth.example.com/some-long-secret-path"`
```

monlovesmango:
> It doesn't have to be integrated in any way; the signer, interacting with the user can do whatever it wants; POW on the window, captcha, ask for a 1000 sats, whatever. That interaction doesn't need to be spec'ed out.

I was more asking "how would `create_account` method be employed for these use cases?", because it seems like the `create_account` method wouldn't even be used for these. in order to use any of these spam prevention measures the signer needs to be able to gate keep the creation of new accounts, which it can't with `create_account` method.

but maybe I am not understanding the full flow? just seems like `create_account` is vulnerable to spam and I don't understand how the spam mitigation measures you mentioned would actually be able to be used with `create_account`.

staab:
I also wonder about create_account. I would just do connect, redirect, and the bunker should handle telling the user about duplicates.

erskingardner:
> but maybe I am not understanding the full flow? just seems like create_account is vulnerable to spam and I don't understand how the spam mitigation measures you mentioned would actually be able to be used with create_account.

Having just implemented this with [nostr-ignition](https://github.com/erskingardner/nostr-ignition), maybe I can shed some light. 

Bunkers, and any remote signer for that matter, receives these calls via a Nostr event. This means that they can do whatever they want to block spam, e.g. they can drop it with no response, require a certain amount of PoW for the event that is sent, or (most likely) they can send back an `auth_url` and and the page that this url points to will have captcha or some other spam measure on it. The auth_url points to a page that the bunker itself controls.

With the Oauth-like flow, it's up to the client to do some checking beforehand about the availability of usernames on the domain they are sending the `create_account` command to. This is trivial by just looking up username values in the `/.well-known/nostr.json` for the domain. 

One thing that isn't specced out well in the current doc is how to send better errors that the client can actually understand and act upon.

pablof7z:
> I also wonder about create_account. I would just do connect, redirect, and the bunker should handle telling the user about duplicates.

But to what pubkey do you send the connect; who are you trying to connect as?

pablof7z:
> > It doesn't have to be integrated in any way; the signer, interacting with the user can do whatever it wants; POW on the window, captcha, ask for a 1000 sats, whatever. That interaction doesn't need to be spec'ed out.
> 
> 
> 
> I was more asking "how would `create_account` method be employed for these use cases?", because it seems like the `create_account` method wouldn't even be used for these. in order to use any of these spam prevention measures the signer needs to be able to gate keep the creation of new accounts, which it can't with `create_account` method.
> 
> 
> 
> but maybe I am not understanding the full flow? just seems like `create_account` is vulnerable to spam and I don't understand how the spam mitigation measures you mentioned would actually be able to be used with `create_account`.

The client that wants to create an account sends create_account to the bunker it chooses
gets an auth_url back
redirects/opens that url
bunker does captcha or some spam-prevention in the popup
account created and pubkey sent back to the client

monlovesmango:
ok thanks this makes sense, I just didn't realize that auth flow could be used for the `create_account` method as well (or I guess any method right?). I like that. however I kind of think this should be a new kind or have different payloads so it is explicit that this response should be handled differently. for example, for nip04 decryption if the plaintext just happens to be a URL how is the client supposed to know whether that is an auth URL response or a real response?

so for existing bunker users who want to login on a new client, what method would the client use? are you assuming the user will have a nip05 that the client can look up and then just use the `connect` method?

>With the Oauth-like flow, it's up to the client to do some checking beforehand about the availability of usernames on the domain they are sending the create_account command to. This is trivial by just looking up username values in the /.well-known/nostr.json for the domain.

I am strongly against making it the clients responsibility to do the bunker's account validation. if we go down this route we need to define error handling between the bunker and the client in the spec imo.

pablof7z:
> ok thanks this makes sense, I just didn't realize that auth flow could be used for the `create_account` method as well (or I guess any method right?). I like that. however I kind of think this should be a new kind or have different payloads so it is explicit that this response should be handled differently. for example, for nip04 decryption if the plaintext just happens to be a URL how is the client supposed to know whether that is an auth URL response or a real response?

There's a rewrite of NIP-46 in the works where all "methods" are moved to kinds, which is much much better than this jsonrpc approach that makes things so damn complex.

> 
> so for existing bunker users who want to login on a new client, what method would the client use? are you assuming the user will have a nip05 that the client can look up and then just use the `connect` method?

`connect` always takes a pubkey, the NIP-05 *can* be used just a lookup method because a nip-05 is more memorable than an npub/hexpubkey -- regardless, that's a client-specific flow. So yeah, clients always use `connect` with hexpubkey to login and `create_account` to signup.

> 
> > With the Oauth-like flow, it's up to the client to do some checking beforehand about the availability of usernames on the domain they are sending the create_account command to. This is trivial by just looking up username values in the /.well-known/nostr.json for the domain.
> 
> I am strongly against making it the clients responsibility to do the bunker's account validation. if we go down this route we need to define error handling between the bunker and the client in the spec imo.

not really, that's also just a client thing and client's must do some validation to prevent evil bunkers announcing create_accounts for e.g. domains they don't control. They don't *have* to, but they probably should.

monlovesmango:
>So yeah, clients always use connect with hexpubkey to login and create_account to signup.

I guess I should have been more clear in my question. in the event that an existing bunker user doesn't have a nip05, what method/workflow should be used to login?

>client's must do some validation to prevent evil bunkers announcing create_accounts for e.g. domains they don't control

how would a client know that a bunker is evil? or what domains a bunker controls? the bunker address is input by the user, right? all the client gets back is a pubkey. 

is there a PR for that nip 46 rewrite? I think that would be better (to have different kinds). might also be good to align with nip47 structure.

PR #978, opened 2024-01-09
Secure DM

arthurfranca:
Read [here](https://github.com/arthurfranca/nips/blob/priv-dm/43.md)


staab:
This seems fine, but again, similar to [NIP 101](https://github.com/water783/nips/blob/nip101/101.md). I'd also like to see zaps as an alternative to PoW. That wouldn't help relays since they can't verify zaps, but recipients could fetch zaps, validate them, then fetch the envelopes that the zaps tagged. This is preferable to PoW for me because the work expended is durable rather than disposable. It also allows for senders to bid or receivers to set different thresholds for message requests.

arthurfranca:
@staab Can't open issue there at the NIP-101 repo nor there is a PR to comment on.

@water783 if you ever read this, I think NIP-101 is more complex because:
- It uses 5 event kinds per session. I think "reject" and "close" are things one doesn't want to reveal. "Update" isn't needed when one can just start a new session (in our case where we don't need the other party's alias). Here we use just 2 event kinds (kind:1044 and 1045 rumors).
- If I got it correctly, both users need to get online once before anyone can start sending DMs (because the sender doesn't know the receiver's alias pubkey to add it to the DM gift-wrap's `p` tag?). Here our DM gift wrap uses no `p` tag at all.
- It has custom expiration (and update interval). Here we simply use a convention fixed on a three weeks expiration window.
- It needs to set a relay. Here we just use NIP-65.

arthurfranca:
@staab zaps aren't an alternative to PoW cause of the relay part you mentioned. Zaps could be used in additional to PoW though.

I admit that atm I'm a bit lazy to open NIP-57 and figure out how to stuff it in here xD

arthurfranca:
> Furthermore, you could get forward secrecy if you did in fact generate keys and exchange them. Why not strive for forward secrecy? https://github.com/nostr-protocol/nips/pull/945#issuecomment-1881932884

@mikedilger could you review [these lines](https://github.com/nostr-protocol/nips/blob/49d1f7305971b57eae1cbacc269546b47bdd6276/43.md?plain=1#L123-L136)? New sessions would not be able to decrypt previous DMs after discarding previous session's privkey (it's chars are used as salt for decrypting the old DMs).

43.md
```diff
@@ -0,0 +1,188 @@
...
+### Last Received At
+
+Optionally, the `created_at` of a `kind:15` NIP-59 rumor
+informs the last time an user's client received DMs from the other participant.
+
+This event should be sent to atleast one of the recipient's NIP-65 `read` relays.
+
+### Last Read At
+
+Optionally, the `created_at` of a `kind:16` NIP-59 rumor
```
AsaiToshiya, line 185: Nit: `kind:16` is a `generic repost` defined in NIP-18.

mattn:
Just a question, if the session-key for A & B to another one, and if it also add C's pubkey to the p tag. it can become private group chat?

43.md
arthurfranca: fixed. thanks!

arthurfranca:
@mattn I'm not sure if you want to add `p` tags to `kind:1044` rumor (Chat Session Request) or to `kind:14` (DM) or elsewhere.

Either way, we would need some type of chat room identifier to add to the `kind:10043` list like in place of "\<pubkey-A\>" from the `["s", "<pubkey-A>", "<session-privkey-A>", "1704525400", "1"]` example. And maybe replace `s` with another tag name like `gs` (as in "g"roup "s"ession)

43.md
```diff
@@ -0,0 +1,191 @@
...
+`draft` `optional`
+
+This direct message (DM) scheme between two participants is end-to-end encrypted, hides
+most metadata, has forward secrecy and aims to minimize spam.
+
+This NIP currently uses [NIP-44](44.md) v2 encryption.
+
+## Flow
+
+- Before sending a DM to someone (`B`), an user (`A`) picks a random privkey to be used as chat session (`session-privkey-A`);
```
AsaiToshiya, line 16:
```suggestion
- Before sending a DM to someone (`B`), an user (`A`) picks a random privkey to be used as chat session (`session-privkey`);
```

Will there be a `session-privkey-B`? `A` to `B` and `B` to `A` are the same session, aren't they?

arthurfranca: Yeah it could confuse people. There is no session-privkey-B. They use the same privkey.
Didn't notice in time I could've applied yout commit suggestion sorry. Renamed other occurrences too

arthurfranca:
Added new device usage detection to help identify unauthorized privkey access.

In my mind the flow works but it may have a flaw I may be missing.

The idea is simple: a hacker with my privkey won't be able to read my messages nor talk to someone until it reveals his device id (called LID - any random string local to the client/device) to the other chat participant (peer).

Then the peer will broadcast the hacker's LID (using a LID Tracker Event) so that I can compare it with my own LID. If it is different than mine and I hadn't used a new device recently, someone else accessed my account.

PR #979, opened 2024-01-10
cleanup NIP-52: remove confusing parts

fiatjaf:
I don't think anyone has implemented anything with the `name` tag so it's better to not even mention it. People who know about it already know that it should be removed.

Removing the mention to the private events limitation since it isn't relevant. There are infinite things each NIP _doesn't_ do and there is no need to mention all of them, so we should stick to what it does.

### DIFF
--- a/52.md
+++ b/52.md
@@ -35,9 +35,6 @@ The list of tags are as follows:
 * `t` (optional, repeated) hashtag to categorize calendar event
 * `r` (optional, repeated) references / links to web pages, documents, video calls, recorded videos, etc.
 
-The following tags are deprecated:
-* `name` name of the calendar event. Use only if `title` is not available.
-
 ```json
 {
   "id": <32-bytes lowercase hex-encoded SHA-256 of the the serialized event data>,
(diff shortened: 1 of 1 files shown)


fiatjaf:
@tyiu

tyiu (approved):

fiatjaf:
@zmeyer44

52.md
```diff
@@ -35,20 +35,19 @@ The list of tags are as follows:
...
-* `name` name of the calendar event. Use only if `title` is not available.
-
 ```json
 {
   "id": <32-bytes lowercase hex-encoded SHA-256 of the the serialized event data>,
   "pubkey": <32-bytes lowercase hex-encoded public key of the event creator>,
   "created_at": <Unix timestamp in seconds>,
   "kind": 31922,
-  "content": "<description of calendar event>",
+  "content": "",
```
vitorpamplona, line 44: Won't description go here anymore? `""` alone might imply no description is allowed.

fiatjaf: I got confused.

52.md
```diff
@@ -49,6 +46,8 @@ The following tags are deprecated:
     ["d", "<UUID>"],
 
     ["title", "<title of calendar event>"],
+    ["image", "<image url>"],
```
tyiu, line 49: Oh, these look new. I guess @zmeyer44 added them to Flockstr. We should add documentation in the date-based and time-based calendar events on what the `image` and `banner` tags are for both kinds.

ciegovolador (approved):
LGTM

staab (approved):

PR #980, opened 2024-01-10
remove NIP-52 label cruft

fiatjaf:
This is a breaking change, but I hope everybody sees how simpler (and cheaper) things are when we remove these unnecessary label and think it's worthwhile to make the required adjustments.

If you think it's not worth the trouble, then at least let's not repeat this mistake in future standards.


fiatjaf:
@tyiu @zmeyer44

tyiu (approved):
Looks good to me. I'm fine with making it simpler like this.

The original draft of NIP-52 had these fields as `t` tags so that they would be indexable (as all single letter tags are), in which the feedback was that it'd be better to use labels then (to not confuse with actual hashtags). Yet, the labels approach felt cumbersome for this use case.

In practice, using labels may be overthinking use cases when they haven't presented themselves yet. I'm fine moving forward with the `status` and `fb` labels.

vitorpamplona:
Looks better. I think the namespace + label pair only makes sense when users are defining their own terms.

zmeyer44:
I'd down with these changes. Updating flockstr now

Sebastix:
@zmeyer44 Is flockstr updated? I still see published kind `31923` events with labels and not with the `status` tag.

PR #982, opened 2024-01-10
Remove "d" tag reference from README

schmijos:
This is about the matter of replaceable events.

Even though a lot of NIPs still reference the "d" tag, I was confused by the explicit reference to NIP-01. The "d" tag lost its specific definition in https://github.com/nostr-protocol/nips/pull/703 where NIP-33 was merged into NIP-01.

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,7 +187,7 @@ Please update these lists when proposing NIPs introducing new event kinds.
 | `e`               | event id (hex)                       | relay URL, marker    | [01](01.md), [10](10.md)              |
 | `p`               | pubkey (hex)                         | relay URL, petname   | [01](01.md), [02](02.md)              |
 | `a`               | coordinates to an event              | relay URL            | [01](01.md)                           |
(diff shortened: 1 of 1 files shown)


vitorpamplona:
This rewrite of NIP-01 makes the definition of the d-tag more clear: https://github.com/nostr-protocol/nips/pull/963

fiatjaf:
Why? The `d` tag is defined in NIP-01.

schmijos:
Where? I cannot find it.

CleanShot 2024-01-11 at 11 25 58@2x
CleanShot 2024-01-11 at 11 27 15@2x
CleanShot 2024-01-11 at 11 27 37@2x

I guess it's a bug then?
In this case @vitorpamplona's suggestion is better.
Closing.

schmijos:
I now found the reference. It's web searchable with "d tag"
CleanShot 2024-01-11 at 11 35 30@2x

schmijos:
@fiatjaf I think the main confusion was caused by this statement in NIP-01

> This NIP defines 3 standard tags

PR #985, opened 2024-01-11
Update 89.md and 90.md with Customer Feedback Data

dtdannen:
This adds an additional protocol flow step to Nip-90 where a Customer can provide feedback back to the service provider.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,7 @@ They exist to document what may be implemented by [Nostr](https://github.com/nos
 - [NIP-72: Moderated Communities](72.md)
 - [NIP-73: External Content IDs](73.md)
 - [NIP-75: Zap Goals](75.md)
+- [NIP-77: Negentropy Syncing](77.md)
 - [NIP-78: Application-specific data](78.md)
 - [NIP-7D: Threads](7D.md)
 - [NIP-84: Highlights](84.md)


dtdannen:
Actually, this should really be a kind 7 event from NIP-25: https://github.com/nostr-protocol/nips/blob/8331354947f2d577e13eb5da4a56133071cb1019/25.md#L5

dtdannen:
I updated 90.md to use kind:7 instead, like:

image

dtdannen:
Updated both NIP-89 and NIP-90 to be aligned. The overall process for DVMs to receive additional feedback from Customers is the following:

* NIP-89 applications can include an optional `pff` tag specifying preferred feedback format
* NIP-90 customer feedback events should attempt to align the data in the additional `feedback` tag to align with the instructions from the `pff` tag from NIP-89. 

Note 1: Customer feedback may not be from a human but from another DVM or application.

Note 2: These new tags for preferred feedback format `pff` and `feedback` are both placeholders for data, and no other commitments are made about their contents (other than they should be strings). It is between any application and customer to work together to ensure consistency. 

The use case for this additional feedback is for DVMs that have AI components (or human components) that may learn over time. The additional feedback could include, but is not limited, to such values like Likert scale scores 1-5, numeric or floating point scores between arbitrary ranges, qualitative descriptions of feedback in natural language, or symbolic structured representations.

believethehype:
Do i get it right that the kind of feedback is always up to the dvm?  (in their preferred format etc).

I could imagine it would be useful to have some sort of standardised feedback or rating for dvms (maybe additionally to what they ask for). I'm thinking about apps using dvms that want to filter out bad actors/dead dvms/dvms that generally don't work as expected). It could be something like "2 out of 5 stars" etc. Apps could then collect ratings, build average ratings and decide if they include them in their offers/results. This only works if the kind of feedback is somehow standardised.

dtdannen:
With this update, there would be two ways to give feedback:

1. regular reactions (kind:7) that are limited to emojis or '+' / '-'
2. feedback in the form of a string as described in this PR

The reason for introducing additional feedback (#2) is to enable DVMs to get richer feedback data that isn't possible via the reaction events, where the point of the feedback is to help the DVM which processed the request get better over time, the DVM should specify the format of the feedback it would like to receive. 

What you're talking about seems more along the lines of reputation or reviews about DVMs. That's certainly a fine idea, although the purpose of it is different than the feedback I proposed here. The feedback you're talking about would serve to help users choose which DVM to use. The main question with that is whether kind:7 reactions are good enough.

PR #991, opened 2024-01-15
update NIP-65: add DM relays, clarify when to use all vs some of a relay kind

mikedilger:
We have talked about 'DM' relays for a long time now but we never put them in NIP-65.

Three main changes of note

1. DM relays added. Fallback goes to READ relays

2. Posting a copy of DMs that you send out to your own DM relays (instead of your own WRITE relays).  Clients will have to look for them in both places due to the transition, but this is clearly more secure.

3.
    In general, when interacting with a user's own relays, clients should interact with all of them (write to all WRITE relays, read from all READ and DM relays).  When interacting with other user's relays, clients need only interact with a subset of them (read from some of their WRITE relays, write to some of their READ and DM relays).

    I'm not sure if that is best, but it is how my client has always operated, and clients should probably have a unified understanding of what is expected. Of course relays can be down so oftentimes it is inadvertently a "subset" even when you attempt to post to "all". But we should at least be trying.

### DIFF
--- a/65.md
+++ b/65.md
@@ -6,11 +6,100 @@ Relay List Metadata
 
 `draft` `optional`
 
-Defines a replaceable event using `kind:10002` to advertise preferred relays for discovering a user's content and receiving fresh content from others.
+Defines two replaceable events to advertise preferred relays for discovering a user's content and receiving fresh content from others.
 
-The event MUST include a list of `r` tags with relay URIs and a `read` or `write` marker. Relays marked as `read` / `write` are called READ / WRITE relays, respectively. If the marker is omitted, the relay is used for both purposes.
+Relay Usages
+------------
+
+There are three defined usages for a relay:
+
+- `INBOX` relays are for events that the relay list owner will be reading, specifically watching for events that tag them. These do not include DMs (except in the deprecated form of relay list).
+
+- `OUTBOX` relays are for events that the relay list owner has authored and is publishing for public consumption.
+
+- `DIRECT MESSAGE` relays are like INBOX relays, except they are for reciept of Direct Messages (DMs) as they are considered more secure for this purpose by the relay list owner.
+
+
+### When to use which
+
+When posting events (except DMs):
+
+- Clients SHOULD post events to all of their client user's OUTBOX relays,
+- For each p-tag in a posted event, clients SHOULD post the event to at least some of the INBOX relays of the tagged person.
+
+When posting DMs:
+
+- Clients SHOULD post events to at least some (preferably all) of the the p-tag recipient's DM relays.
+- Clients SHOULD post a copy to their client user's own DM relays.
+- If a user has no DM relays, INBOX relays should be used instead.
+
+When seeking events from a user:
+
+- Clients SHOULD seek them from at least some of that user's OUTBOX relays.
+
+When seeking events about a client's user (where the user was tagged):
+
+- Clients SHOULD seek them from all of their user's INBOX relays.
+
+When seeking **DM**s for a client's user:
+
+- Clients SHOULD seek them from all of their user's DM and INBOX relays.
+
+In general, when interacting with a user's own relays, clients should interact with all of them (write to all OUTBOX relays, read from all INBOX and DM relays). When interacting with other user's relays, clients need only interact with a subset of them (read from some of their OUTBOX relays, write to some of their OUTBOX and DM relays).
+
+Relay List Event
+----------------
+
+`kind:10102` is a relay list event.
+
+This event SHOULD include `relay-usage` tags of the following form:
+
+```json
+["relay-usage", "<relay-url>", "<usage-flags>"]
+```
+
+`usage-flags` includes a sequence of letters, which of which MUST be one of the following:
+
+- `i` - This indicates the relay is used as the user's INBOX
+- `o` - This indicates the relay is used as the user's OUTBOX
+- `d` - This indicates the relay is used for the user's DIRECT MESSAGES
+
+`usage-flags` SHOULD not have duplicate letters. Clients should ignore any other letters found that they do not understand.
 
 The `.content` is not used.
+
+### Example:
+
+```json
+{
+  "kind": 10102,
+  "tags": [
+    ["relay-usage", "wss://alicerelay.example.com", "io"],
+    ["relay-usage", "wss://brando-relay.com", "io"],
+    ["relay-usage", "wss://expensive-relay.example2.com", "o"],
+    ["relay-usage", "wss://nostr-relay.example.com", "i"],
+    ["relay-usage", "wss://secure.example.com", "d"],
+  ],
+  "content": "",
+  ...other fields
+}
+```
+
+Deprecated Relay List Event
+---------------------------
+
+`kind:10002` is the deprecated version of the relay list event. Clients will want to create this to support existing clients that have not upgraded to the new `kind:10102`.
+
+This event SHOULD include `r` tags of the following form:
+
+```json
+["r", "<relay-url>", "<read-or-write>"]
+```
+`read-or-write` MUST be the string `"read"` (for an INBOX) or the string `"write"` (for an OUTBOX), or it should be left out (or empty). If it is left out (or empty) the relay is both an INBOX and an OUTBOX. Clients should ignore any other strings found that they do not understand.
+
+The `.content` is not used.
+
+### Example:
 
 ```json
 {
@@ -19,46 +108,35 @@ The `.content` is not used.
     ["r", "wss://alicerelay.example.com"],
     ["r", "wss://brando-relay.com"],
     ["r", "wss://expensive-relay.example2.com", "write"],
-    ["r", "wss://nostr-relay.example.com", "read"]
+    ["r", "wss://nostr-relay.example.com", "read"],
   ],
   "content": "",
   ...other fields
 }
 ```
 
-This NIP doesn't fully replace relay lists that are designed to configure a client's usage of relays (such as `kind:3` style relay lists). Clients MAY use other relay lists in situations where a `kind:10002` relay list cannot be found.
-
-## When to Use Read and Write Relays
-
-When seeking events **from** a user, Clients SHOULD use the WRITE relays of the user's `kind:10002`.
-
-When seeking events **about** a user, where the user was tagged, Clients SHOULD use the READ relays of the user's `kind:10002`.
-
-When broadcasting an event, Clients SHOULD:
-
-- Broadcast the event to the WRITE relays of the author
-- Broadcast the event all READ relays of each tagged user
-
 ## Motivation
 
-The old model of using a fixed relay list per user centralizes in large relay operators: 
+The old model of using a fixed relay list per user centralizes in large relay operators:
 
   - Most users submit their posts to the same highly popular relays, aiming to achieve greater visibility among a broader audience
   - Many users are pulling events from a large number of relays in order to get more data at the expense of duplication
   - Events are being copied between relays, oftentimes to many different relays
-  
-This NIP allows Clients to connect directly with the most up-to-date relay set from each individual user, eliminating the need of broadcasting events to popular relays. 
+
+This NIP allows Clients to connect directly with the most up-to-date relay set from each individual user, eliminating the need of broadcasting events to popular relays.
 
 ## Final Considerations
 
-1. Clients SHOULD guide users to keep `kind:10002` lists small (2-4 relays). 
+1. Authors may publish to their own INBOX relays events that they intend to read back later. This doesn't make them OUTBOX relays.
 
-2. Clients SHOULD spread an author's `kind:10002` event to as many relays as viable. 
+2. Clients SHOULD guide users to keep relay lists small (2-6 relays).
 
-3. `kind:10002` events should primarily be used to advertise the user's preferred relays to others. A user's own client may use other heuristics for selecting relays for fetching data.
+3. Clients SHOULD advertise an author's relay list events to as many relays as viable.
 
-4. DMs SHOULD only be broadcasted to the author's WRITE relays and to the receiver's READ relays to keep maximum privacy. 
+4. Relay list events should primarily be used to advertise the user's preferred relays to others. A user's own client may use other heuristics for selecting relays for fetching data or for other purposes. Clients MAY use other relay lists (such as `kind:3` contents) in situations where a `kind:10002` relay list cannot be found.
 
-5. If a relay signals support for this NIP in their [NIP-11](11.md) document that means they're willing to accept kind 10002 events from a broad range of users, not only their paying customers or whitelisted group.
+5. DMs SHOULD only be broadasted to the author's DM relays and to the receiver's DM relays to keep maximum privacy. If either party doesn't have DM relays, INBOX relays should be used instead.
 
-6. Clients SHOULD deduplicate connections by normalizing relay URIs according to [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986#section-6).
+6. If a relay signals support for this NIP in their [NIP-11](11.md) document that means they're willing to accept kind 10002 events (and hopefully kind 10102 events) from a broad range of users, not only their paying customers or whitelisted group.
+
+7. Clients SHOULD deduplicate connections by normalizing relay URIs according to [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986#section-6).
--- a/65.md
+++ b/65.md
@@ -64,7 +64,7 @@ This event SHOULD include `relay-usage` tags of the following form:
 - `o` - This indicates the relay is used as the user's OUTBOX
 - `p` - This indicates the relay is used for the user's PRIVATE messages
 
-`usage-flags` SHOULD not have duplicate letters. Clients should ignore any other letters found that they do not understand.
+`usage-flags` SHOULD not have duplicate letters. Clients should ignore and preserve any other letters found that they do not understand.
 
 The `.content` is not used.
 
@@ -95,7 +95,7 @@ This event SHOULD include `r` tags of the following form:
 ```json
 ["r", "<relay-url>", "<read-or-write>"]
 ```
-`read-or-write` MUST be the string `"read"` (for an INBOX) or the string `"write"` (for an OUTBOX), or it should be left out (or empty). If it is left out (or empty) the relay is both an INBOX and an OUTBOX. Clients should ignore any other strings found that they do not understand.
+`read-or-write` MUST be the string `"read"` (for an INBOX) or the string `"write"` (for an OUTBOX), or it should be left out (or empty). If it is left out (or empty) the relay is both an INBOX and an OUTBOX. Clients should ignore and preserve any other strings found that they do not understand.
 
 The `.content` is not used.
 
--- a/65.md
+++ b/65.md
@@ -13,24 +13,23 @@ Relay Usages
 
 There are three defined usages for a relay:
 
-- `INBOX` relays are for events that the relay list owner will be reading, specifically watching for events that tag them. These do not include DMs (except in the deprecated form of relay list).
+- `INBOX` relays store public events that mention the relay list owner. Any user can publish and download such events from this relay. The relay list owner generally monitors these relays for events mentioning them.
 
-- `OUTBOX` relays are for events that the relay list owner has authored and is publishing for public consumption.
+- `OUTBOX` relays store events that the relay list owner has authored. Any user can publish and download such events from this relay.
 
-- `PRIVATE INBOX` relays are like INBOX relays, except they are for reciept of private messages such as perhaps direct messages (DMs), giftwraps, or the like (but this NIP does not specify what kinds should use relays with this designator).
+- `PRIVATE INBOX` relays store private events that mention the relay list owner, such as perhaps direct messages (DMs) and giftwraps. Any user can publish such events to this relay, but may not be able to download them (depending on the relay's actual behavior).
 
 ### When to use which
 
-When posting events (except private ones):
+When publishing public events:
 
-- Clients SHOULD post events to all of their client user's OUTBOX relays,
-- For each p-tag in a posted event, clients SHOULD post the event to at least some of the INBOX relays of the tagged person.
+- Clients SHOULD publish events to all of the signer's OUTBOX relays.
+- Clients SHOULD publish the event to all of the INBOX relays of each tagged key.
 
-When posting private messages:
+When publishing private messages:
 
-- Clients SHOULD post events to at least some (preferably all) of the the p-tag recipient's PRIVATE INBOX relays.
-- Clients SHOULD post a copy to their client user's own PRIVATE INBOX relays.
-- If a user has no PRIVATE INBOX relays, INBOX relays should be used instead.
+- Clients SHOULD publish events to all of the PRIVATE INBOX relays of each of the recipients.
+- Clients SHOULD publish a copy to all of their client user's own PRIVATE INBOX relays.
 
 When seeking events from a user:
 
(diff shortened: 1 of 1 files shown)


fiatjaf:
What if we rename the DM relays tag to `["dm", "wss://..."]`? No one is using it yet so it won't break anything, and we save one index entry for relays.

mikedilger:
I think we can do that.

staab:
I think this is an improvement, but I'm a little worried about the UX. Also a nitpick, what about "authenticated-read" or something to support other private read use cases?  Either `r` or a new tag name is fine with me.

mikedilger:
The 'r' tags were my mistake that became de-facto standard. So anything new should have a different tag. Could just be 'relay' with markers, but I'm not very opinionated on this aspect.

What is an authenticated-read relay? Why would you need to advertise it? If I put things on a relay that require authentication to read, I wouldn't list it in my 10002 event. Maybe you are saying "please subscribe to my paid service?"

mikedilger:
If we deprecate 'r' and switch to 'relay' (for example), there will be a transition period where new clients have to write both, and old clients may or may not preserve the new tags.  Then once everybody uses the new tags, all the clients have to change again to not bothering to write the old 'r' tags.  I'm not against doing this.

But should they be 'relay' with marker?  Or something different (perhaps without a marker)?

staab:
Marks are evil IMO.

Maybe authenticated-read is the wrong term, I'm just thinking of use cases other than DMs (like encrypted group messages) that should be treated similarly.

mikedilger:
Relays can be used for any combination of (inbox, outbox, dm) and maybe a fourth kind @staab is talking about.  So those can't be the tags unless we are adding the same relay multiple times.

Given this example:
```
example1 is both 'inbox' and 'dm'
example2 is both 'inbox' and 'outbox'
example3 is just 'outbox'
```

So here is the option of adding relays multiple times (including the lines needed for backwards compatibility):
````
[
    ["inbox", "wss://example1.com"],
    ["inbox", "wss://example2.com"],
    ["outbox", "wss://example2.com"],
    ["outbox", "wss://example3.com"],
    ["dm", "wss://example1.com"],
    ["r", "wss://example1.com", "read"],
    ["r", "wss://example2.com"],
    ["r", "wss://example3.com", "write"],
]
````

Markers make it shorter, but we would need multiple markers per relay (or an encoded marker which I'm not showing).
````
[
    ["relay", "wss://example1.com", "inbox", "dm"],
    ["relay", "wss://example2.com", "inbox", "outbox"],
    ["relay", "wss://example3.com", "outbox"],
    ["r", "wss://example1.com", "read"],
    ["r", "wss://example2.com"],
    ["r", "wss://example3.com", "write"],
]
````

**Comment by mikedilger** [comment]
I'm not against encoding the usage as a bit vector.  bit0=write, bit1=read, bit2=dm.  Then there is just a single marker. The tag could be "relayusage".

**Comment by mikedilger** [comment]
Here is how it looks with bit flags (my new preferred option)

````
[
   ["relayusage", "wss://example1.com", 6],
   ["relayusage", "wss://example2.com", 3],
   ["relayusage", "wss://example3.com", 1],
   ["r", "wss://example1.com", "read"],
   ["r", "wss://example2.com"],
   ["r", "wss://example3.com", "write"],
]
````

**Comment by jb55** [comment]
On Tue, Jan 16, 2024 at 01:41:49PM -0800, Michael Dilger wrote:
>Here is how it looks with bit flags (my new preferred option)
>
>````
>[
>   ["relayusage", "wss://example1.com", 6],
>   ["relayusage", "wss://example2.com", 3],
>   ["relayusage", "wss://example3.com", 1],
>   ["r", "wss://example1.com", "read"],
>   ["r", "wss://example2.com"],
>   ["r", "wss://example3.com", "write"],
>]
>````

To keep it simple and more descriptive, you could use single-char flags.
It could be "rwd", "rw", "d", etc.

Then you wouldn't need "relayusage"

File: 65.md [hunk 1]
```diff
@@ -30,34 +41,50 @@ This NIP doesn't fully replace relay lists that are designed to configure a clie
...
-1. Clients SHOULD guide users to keep `kind:10002` lists small (2-4 relays). 
+1. Clients SHOULD guide users to keep `kind:10002` lists small (2-6 relays).
 
-2. Clients SHOULD spread an author's `kind:10002` event to as many relays as viable. 
+2. Clients SHOULD advertise an author's `kind:10002` event to as many relays as viable.
 
 3. `kind:10002` events should primarily be used to advertise the user's preferred relays to others. A user's own client may use other heuristics for selecting relays for fetching data.
 
-4. DMs SHOULD only be broadcasted to the author's WRITE relays and to the receiver's READ relays to keep maximum privacy. 
+4. DMs SHOULD only be broadcasted to the recipient's and author's DM relays. Previously this NIP specified for them to be broadcast to the author's WRITE relays but that is less secure. Clients may have to look for them in both places, but should post them only to DM relays.
```
vitorpamplona, line 87: There is no need to add historical reasoning. I would just keep the guidance for what should happen from now on.

Actually... since this guidance seems to be duplicated from the text above, there might not be a need for this 4th point here anymore. This point was here exactly because there was no explicit pattern for DMs before.

65.md
```diff
@@ -8,7 +8,17 @@ Relay List Metadata
...
 Defines a replaceable event using `kind:10002` to advertise preferred relays for discovering a user's content and receiving fresh content from others.
 
-The event MUST include a list of `r` tags with relay URIs and a `read` or `write` marker. Relays marked as `read` / `write` are called READ / WRITE relays, respectively. If the marker is omitted, the relay is used for both purposes.
+The event MUST include a list of `r` tags with relay URIs with an optional single marker indicating what the relay is used for.
+
+The marker, if it exists, MUST be one of:
+
+- `write` - The user writes events to this relay for the public to view.
+- `read` - The user reads events from this relay if they tag the user.
+- `dm` - This is identical to `read` but also signals that encrypted direct messages should preferably be written or one of these relays.
```
vitorpamplona, line 17: The last change in this PR suggests that DM is both a READ and a WRITE relay: 

> DMs SHOULD only be broadcasted to the recipient's and author's DM relays.

Which I think it is correct, but here it says that it only works as a READ relay, which I think it is incorrect. DM relays are indeed for both write and read. 

PS: The READ base works for the current GiftWrap design (the signing key is always random and thus no WRITE is needed), but there may be other encryptions that are signed by the author.

mikedilger: Is there a case where somebody comes to your DM relay to read your DMs?  If not, it is inbox only. Just because you write into it doesn't make it an outbox.  I would leave it more flexible, except that if we make these outboxes, what was the point of special DM relays that leak your DMs?

vitorpamplona: > Is there a case where somebody comes to your DM relay to read your DMs?

Only for those gathering metadata from DMs (which does happen today already). Do we consider a shared-secret group, like what @staab has, a "DM"? They are encrypted messages but anyone with the key can access them.  

It wasn't clear to me that this PR requires DM relays to authenticate who is connecting and only provide the appropriate messages to that user. If that is the case then it makes sense to be read. Maybe this is an expected behavior of DM relays and should be added here?

65.md
mikedilger: removed.

65.md
```diff
@@ -8,57 +8,114 @@ Relay List Metadata
...
+OUTBOX relays are for events that the relay list owner has authored and is publishing for public consumption.
 
-When seeking events **about** a user, where the user was tagged, Clients SHOULD use the READ relays of the user's `kind:10002`.
+DIRECT MESSAGE relays are like INBOX relays, except they are for reciept of DMS as they are considered more secure for this purpose by the relay list owner.
 
-When broadcasting an event, Clients SHOULD:
 
-- Broadcast the event to the WRITE relays of the author
-- Broadcast the event all READ relays of each tagged user
+### When to use which
```
arthurfranca, line 70: I miss NIP-65 guidance on which relays to pick when replying. Writing just to the guy I'm replying to's inbox relays (and my own outbox ones) and then everyone else repeating the same pattern would potentially spread the thread notes through a big set of relays. It would be very hard to fetch all thread notes.

I think we should leverage the fact that all replies must add the `e` tag with `root` marker (NIP-10).

What I'm doing is I download the `root` note to get the OP's pubkey > then I fetch it's `kind:10002` event > select the first 2 inbox relays (making the order relevant) and publish to them.

So I consider the first 2 OP's inbox relays as the "thread relays".

mikedilger: IIRC there should be 'p' tags marking all the authors up the chain. For each of those 'p' tags you write to their inboxes.

arthurfranca: True! All good.

65.md
staab: Seems to me that it should act like a read relay. Auth should be recommended but optional. This works for shared key groups too, but isn't relevant to nip 65 because the group metadata defines the relays to use, not the recipient's profile.

dluvian:
I am against this.

I prefer to keep `kind:10002` small. DM relays are a very specific usecase, so we should put them in their own kind. Since not all clients handle DMs, it's best not to make them fetch and deal with it.

65.md
```diff
@@ -8,57 +8,114 @@ Relay List Metadata
 
 Defines a replaceable event using `kind:10002` to advertise preferred relays for discovering a user's content and receiving fresh content from others.
 
-The event MUST include a list of `r` tags with relay URIs and a `read` or `write` marker. Relays marked as `read` / `write` are called READ / WRITE relays, respectively. If the marker is omitted, the relay is used for both purposes.
+The event SHOULD include `relay-usage` tags, each with a relay URIs and a string indicating what the relay is used for. They should also include tags for backwards compatibility (see section below).
+
+`relay-usage` tags are in the following form:
+
+```json
+["relay-usage", "<relay-url>", "<usage-flags>"]
```
arthurfranca, line 16: Imo it's better to use `ru` instead of `relay-usage` as tag key to make the event smaller (we should strive to save space cause relays aren't supposed to delete them often). Two letters is enough for relays to skip indexing.

65.md
mikedilger: I think it depends on the specific DM spec. I was imaging a copy lands in my read relay (for me) and my partner's read relay (for them). But maybe this NIP should defer to the DM NIPs.

65.md
mikedilger: Sounds good. The Russians will love it ;-)

fiatjaf:
> I am against this.
> 
> I prefer to keep `kind:10002` small. DM relays are a very specific usecase, so we should put them in their own kind. Since not all clients handle DMs, it's best not to make them fetch and deal with it.

I agree. Let's make another kind just for the DMs.

mikedilger:
Ok I can leave it to the DM standard to do that.

I'm kind of against fixing the 'r' tag because it blow up the size with backwards-compatible entries.  I just did it to demonstrate.  I think the better solution is to just say that relays don't have to index this particular 'r' tag, even though that is a bit ugly.

staab:
Agreed, let's add a new kind for DM preferences. It will reduce contention around lists, which is always an annoying problem.

mikedilger:
Ok I made a new kind, 10102.  Thank god for nostr.

I also squashed everything and force-pushed.

staab:
I don't think we should deprecate the old format, as nice as the new one is. Also, as discussed somewhere on [this pr](https://github.com/nostr-protocol/nips/pull/686) adding a general-purpose DM relay type technically applies to nip 04 DMs, which could hurt deliverability in the short term, so it should probably be scoped to new-style DMs and any other NIPs that emerge which require special handling of relay selections.

nostr-wine:
Don't have any strong preferences on how, just want to chime in and say we definitely support some way to signal where a user wants to receive private events.

We built a relay specifically for this use case a while back (https://docs.nostr.wine/inbox/readme) but it desperately needs something like this PR. I believe the use case goes beyond just DMs and could include NWC events, private DVM results, and other private events where preventing metadata leaks is useful.

vitorpamplona:
IF `d` is only for DMs, like the text is saying, and no other event kind, then I prefer defining a new relay list kind on https://github.com/nostr-protocol/nips/pull/686

IF `d` includes any private direct **event** containing whatever is supposed to be private just for this user, then: 
1. The marker should not be called DM. 
2. I prefer to define it here.

mikedilger:
Staab: Are you suggesting we just put DMs on the new list, and leave read/write to continue to be defined by 10002 still?

Vitor: I think 'd' could be for any private direct event, including giftwraps etc. I'm fine with renaming. What do you suggest?

staab:
> Are you suggesting we just put DMs on the new list, and leave read/write to continue to be defined by 10002 still?

Yes, and we should define the DMs per nip. Because for compatibility legacy DMs should continue to be written to the same places, but NIP 17 DMs should be written to the nip-specific relay selections. Also,

> including giftwraps etc

I don't think this is a good idea, because gift wraps aren't always addressed to a particular user (like with NIP 87 groups, which are addressed to a shared key and sent to the group's relays). Another reason why relay selections should be defined per-use-case.

mikedilger:
> > Are you suggesting we just put DMs on the new list, and leave read/write to continue to be defined by 10002 still?
> 
> Yes, and we should define the DMs per nip. Because for compatibility legacy DMs should continue to be written to the same places, but NIP 17 DMs should be written to the nip-specific relay selections. Also,

All NIP-04 says is to use relays that do AUTH to restrict who can fetch your NIP-04s.  How does that conflict in any way with other DM nips? I can't even imagine why some DM systems would require different relays (other than groups which I address below).

> > including giftwraps etc
> 
> I don't think this is a good idea, because gift wraps aren't always addressed to a particular user (like with NIP 87 groups, which are addressed to a shared key and sent to the group's relays). Another reason why relay selections should be defined per-use-case.

Nothing about a relay marked "d" means that every NIP must use it to store their messages. It is the general "DM" inbox, but NIPs that do things differently can still do things differently.

staab:
> How does that conflict in any way with other DM nips?

So currently there's no recommendation. If we add one, clients may start sending messages to different relays, which could result in backwards-incompatible behavior (conversations getting dropped). The current state is dumb anyway, so this is not a big deal, I'm just attempting to be conscientious about compatibility.

> Nothing about a relay marked "d" means that every NIP must use it to store their messages. It is the general "DM" inbox, but NIPs that do things differently can still do things differently.

This is fine, but it seems like a moot point because NIP 17 is the only spec that uses encryption + relay selections right? So I'd prefer relay recommendations to be defined near where they're useful. Unless we decide it's fine to apply the same recommendation to NIP 04, in which case adding it to NIP 65 is probably better.

All fine distinctions, but I'm saying basically the same thing Vitor is.

staab:
Got you covered, Vitor added this to a "Publishing" section in NIP 17: https://github.com/nostr-protocol/nips/pull/686/files#r1515931334

mikedilger:
Ok I had not read all the convos staab/vitor/fiatjaf were having on this yesterday and the day before. My github notifications were flooded yesterday. I'm still working through reading them.

The idea of having a separate relay list for a separate DM system **signals** that the user accepts that kind of DM is not a bad idea. But that is the only positive I see from doing it that way.

I don't think there is any difference to the users. Why would I specify different relays for NIP-17 and NIP-04? I wouldn't, and that complexity just confuses people. It will confuse 90% of developers too.

I am already dreading the code bloat for managing yet another relay list event. Maybe I'll just stick to NIP-04.

staab:
> I don't think there is any difference to the users.

I agree that UX is the main sticking point. But since everything is generalizable to read/write relays, it might be pretty easy to solve by adding an "advanced" mode that people can opt into if they care about inbox privacy.

arthurfranca:
After reading your discussions, i think this NIP could say that the "io" flags could be extended with other one-char flags defined on different NIPs. Then the DM NIP can choose a char flag such as the "d" one. Then in the future another NIP can choose another new one or reuse "i", or "o", or "d".

This is the inverse of [what I previously belived was the best to do](https://github.com/nostr-protocol/nips/pull/1103#issuecomment-1981000339). Now I think relay specialization could be good. 

That's because some NIPs require special treatment of events (new ones created by the NIP) by relays that the current users' inbox/outbox relays may not honor. 

So for example a client that runs its own relay could easily add an entry to the user's `kind:10102` list with its own relay it knows will be able to support a new NIP feature using the corresponding NIP flag "z", without replacing the user's inbox/outbox ones.

65.md
```diff
@@ -6,59 +6,136 @@ Relay List Metadata
...
+
+When seeking events from a user:
+
+- Clients SHOULD seek them from at least some of that user's OUTBOX relays.
+
+When seeking events about a client's user (where the user was tagged):
+
+- Clients SHOULD seek them from all of their user's INBOX relays.
+
+When seeking private messages for a client's user:
```
vitorpamplona, line 43: Guidance here should emphasize that Clients should only connect to Private Inbox if they can AUTH the user they are requesting information for.

65.md
```diff
@@ -6,59 +6,136 @@ Relay List Metadata
...
+`usage-flags` SHOULD not have duplicate letters. Clients should ignore any other letters found that they do not understand.
 
 The `.content` is not used.
 
+### Example:
+
 ```json
 {
-  "kind": 10002,
+  "kind": 10102,
```
vitorpamplona, line 74: What if we don't switch to the new kind but keep both notations going for a while? We can mark `r` tags as deprecated and should not be changed to keep old clients working. New tags can be added without affecting the old clients.

vitorpamplona: I take it back. It seems confusing.

arthurfranca:
Really liked the usage flag mechanics.

[Here](https://github.com/arthurfranca/nips/blob/shared-account/260.md#scheduler-relays) I defined a new `s` relay usage flag to let clients configure/use user's "scheduler" relays, needed for time-releasing "delegation events".

Such relays could also be used to enable regular post scheduling feature on clients by setting two flags: `os`.

I say ship it! :rocket:

65.md
```diff
@@ -6,59 +6,138 @@ Relay List Metadata
...
-Defines a replaceable event using `kind:10002` to advertise preferred relays for discovering a user's content and receiving fresh content from others.
+Defines two replaceable events to advertise preferred relays for discovering a user's content and receiving fresh content from others.
 
-The event MUST include a list of `r` tags with relay URIs and a `read` or `write` marker. Relays marked as `read` / `write` are called READ / WRITE relays, respectively. If the marker is omitted, the relay is used for both purposes.
+Relay Usages
+------------
+
+There are three defined usages for a relay:
+
+- `INBOX` relays are for events that the relay list owner will be reading, specifically watching for events that tag them. These do not include DMs (except in the deprecated form of relay list).
```
vitorpamplona, line 16: This if this words them better: 

- `INBOX` relays store PUBLIC events that mention the relay list owner. Any user can publish and download such events from this relay.

- `OUTBOX` relays store events that the relay list owner has authored. Any user can publish and download such events from this relay

- `PRIVATE INBOX` relays store PRIVATE events that mention the relay list owner, such as perhaps direct messages (DMs) and giftwraps. Any user can publish such events to this relay, but only the owner can download them.

65.md
```diff
@@ -6,59 +6,138 @@ Relay List Metadata
...
+When posting events (except private ones):
+
+- Clients SHOULD post events to all of their client user's OUTBOX relays,
+- For each p-tag in a posted event, clients SHOULD post the event to at least some of the INBOX relays of the tagged person.
+
+When posting private messages:
+
+- Clients SHOULD post events to at least some (preferably all) of the the p-tag recipient's PRIVATE INBOX relays.
+- Clients SHOULD post a copy to their client user's own PRIVATE INBOX relays.
+- If a user has no PRIVATE INBOX relays, INBOX relays should be used instead.
```
vitorpamplona, line 33: Better wording: 

When publishing public events:

- Clients SHOULD publish events to all of the signer's OUTBOX relays,
- Clients SHOULD publish the event to at least some of the INBOX relays of each tagged key.

When sending private messages:

- Clients SHOULD publish events to at least some (preferably all) PRIVATE INBOX relays of each of the recipients.
- Clients SHOULD publish a copy to their client user's own PRIVATE INBOX relays.

65.md
```diff
@@ -6,59 +6,138 @@ Relay List Metadata
...
+
+When seeking events about a client's user (where the user was tagged):
+
+- Clients SHOULD seek them from all of their user's INBOX relays.
+
+When seeking private messages for a client's user:
+
+- Clients SHOULD seek them from all of their user's PRIVATE INBOX and INBOX relays.
+
+In general, when interacting with a user's own relays, clients should interact with all of them (write to all OUTBOX relays, read from all INBOX and PRIVATE INBOX relays). When interacting with other user's relays, clients need only interact with a subset of them (read from some of their OUTBOX relays, write to some of their INBOX and PRIVATE INBOX relays).
```
vitorpamplona, line 47: Since users should merge the results from their OUTBOX relays at the client, each relay might have a different set of records from the user. When interacting with other user's relays, wouldn't clients be missing things if they only select part of the relays?

65.md
mikedilger: My original intent was that INBOXES were "IN", not readable by anybody else.  Reading from other people's inboxes was your innovation.  I never did that or suggested that.

It has value though.  You can use this to find reactions to someone's post.

But it is bad for (for example) the chorus relay because if INBOX messages are readable, people can use the chorus relay as free disk space, and to place illegal content and the like.  Of course this is an out-of-scope issue for the nostr protocol. Maybe personal relays are a bad idea.

I'd like to hear from others on this.

65.md
mikedilger: I thought it best to write to all outbox relays, so readers only need to read from some (since they all have the same events). I think that is good for outboxes. Is there still a problem I'm missing regarding outboxes?

Conversely I thought to write to just some inbox relays since the relay user will monitor and read from all of them.

But since your innovation of people reading from other people's inboxes, that kinks it up. Since they would now have to read from all of them.

65.md
staab: > Reading from other people's inboxes was your innovation. I never did that or suggested that.

Interesting, that's how I've always read it.

vitorpamplona: Well, for me the difference between private and public inbox is that other people can download from the public inbox. Otherwise, they are the same.

65.md
vitorpamplona: To make it real: to seek replies to a note from user A, I would use all of A's INBOX servers.

65.md
mikedilger: Yeah you are right.  I've been doing it wrong.  I'm not seeking replies from the right places. I'll have to update my code.

65.md
mikedilger: Why don't we say writers write to all, readers read from some, in all cases?  That makes it easier.

vitorpamplona: I think it works. It would be nice if users could setup a mirroring service between the relays in the same category.

vitorpamplona: To make it real 2: if you know I am following you (I post my follow list in your Inbox, because you were cited in it), why don't you send your posts to each follower's INBOX relay as if they were directly cited? Then clients don't need to connect to your relay. 

A DVM, or a broadcasting service, can do that for clients.

65.md
mikedilger: I've pushed changes very close to your suggestion.

65.md
mikedilger: I've pushed changes very close to your suggestion.

mikedilger:
I have added that clients should preserve letters they don't understand.

I have already heard about other plants to add yet more letters that I will not be supporting. But I think I can preserve them.

As for clobbering (because you didn't have the latest list), that is a general issue that is already with us irrespective of this PR and can't be fixed or avoided here. Making a new kind just for PRIVATE INBOX, you still can clobber if you didn't have the latest event and you overwrite it.

vitorpamplona:
@mikedilger, I intend to use something like https://github.com/nostr-protocol/nips/pull/1148 to allow our Push Notification server access to the private inbox relays defined here. Let me know if you see any problems with this.

vitorpamplona:
FYI, yesterday's release of the Draft Events https://github.com/nostr-protocol/nips/pull/1124 led me to imagine a private **outbox** relay: a relay that only allows the signer to download their own posts. 

I never thought this would be useful, but it actually is. 

Things like App settings, Drafts, Read Statuses https://github.com/nostr-protocol/nips/pull/933 exist for the benefit of the author alone. And those can indeed be stored in a private "outbox". 

It doesn't need to be added to this PR. But it is something that should be on everybody's radar.

mikedilger:
> It doesn't need to be added to this PR. But it is something that should be on everybody's radar.

Since it is private, there is no reason to add it to NIP-65 which is about advertising to others.  But yes I agree this is a kind of relay usage.

fiatjaf:
@dluvian is 100% correct.

mikedilger:
Ok, I concede. Someone else will have to do it though.

PR #561, opened 2023-05-26
NIP-17 Code Collaboration

DanConwayDev:
This is a very early draft of a more wide ranging NIP relating to code collaboration than those that have come before it:

https://github.com/nostr-protocol/nips/pull/223
https://github.com/nostr-protocol/nips/pull/324

It builds upon the work of many others and stems from the [ngit-cli prototype](https://github.com/danConwayDev/ngit-cli) I built. It seeks to prioritise interoperability between models.

I have removed the use of the [draft Public Groups NIP](https://github.com/nostr-protocol/nips/pull/483). I am unsure that the use of parameterized replaceable event is a good trade off. It reduces auditability, particularly around changes in maintainers and the knock on effect of that.

PR #997, opened 2024-01-21
NIP-34: git stuff

fiatjaf:
The simplest possible git collaboration flow that has a chance of working.

The patch stuff is based on http://git.jb55.com/git-nostr-tools, with some changes.

https://github.com/nostr-protocol/nips/blob/git/34.md

Here's a standalone CLI tool that allows for announcing a repository, sending and downloading patches: https://github.com/fiatjaf/gitstr -- probably very rough still, but I just merged a patch from myself using it so I am happy.

### DIFF
--- a/34.md
+++ b/34.md
@@ -1,0 +1,45 @@
+NIP-34
+======
+
+`git` stuff
+-----------
+
+`draft` `optional`
+
+This NIP defines all the ways code collaboration using and adjacent to [`git`](https://git-scm.com/) can be done using Nostr.
+
+## Repository announcements
+
+Git repositories are hosted in Git-enabled servers, but their existence can be announced using Nostr events, as well as their willingness to receive patches, bug reports and comments in general.
+
+```jsonc
+{
+  "kind": 30617,
+  "content": "",
+  "tags": [
+    ["d", "<repo-id>"],
+    ["name", "<human-readable project name>"],
+    ["description", "brief human-readable project description>"],
+    ["web", "<url for browsing>"], // a webpage url, if the git server being used provides such a thing
+    ["clone", "<url for git-cloning>"], // a url to be given to `git clone` so anyone can clone it
+    ["patches", "<relay-url>"], // this signals that patches are welcome and will be searched for in this relay
+    ["issues", "<relay-url>"] // this signals that bug reports, questions, comments and so on are welcome
+  ]
+}
+```
+
+The tags `web`, `clone`, `patches` and `issues` can be specified multiple times.
+
+## Patches
+
+Patches can be sent by anyone to any repository. Patches to a specific repository SHOULD be sent to the relays specified in that repository's announcement event's `"patches"` tag. Patch events SHOULD include an `a` tag pointing to that repository's announcement address.
+
+```jsonc
+{
+  "kind": 1617,
+  "content": "<patch>", // contents of <git format-patch>
+  "tags": [
+    ["a", "30617:<base-repo-owner-pubkey>:<base-repo-id>"]
+  ]
+}
+```
--- a/34.md
+++ b/34.md
@@ -20,7 +20,6 @@ Git repositories are hosted in Git-enabled servers, but their existence can be a
     ["d", "<repo-id>"],
     ["name", "<human-readable project name>"],
     ["description", "brief human-readable project description>"],
-    ["tag", "head", "<sha1-hex>"], // the current repository commit head, as hex
     ["web", "<url for browsing>", ...], // a webpage url, if the git server being used provides such a thing
     ["clone", "<url for git-cloning>", ...], // a url to be given to `git clone` so anyone can clone it
     ["relays", "<relay-url>", ...] // relays that this repository will monitor for patches and issues
--- a/34.md
+++ b/34.md
@@ -46,10 +46,6 @@ Patches can be sent by anyone to any repository. Patches to a specific repositor
 }
 ```
 
-## Branch merge
-
-To be defined.
-
 ## Issues
 
 Issues are Markdown text that is just human-readable conversational threads related to the repository: bug reports, feature requests, questions or comments of any kind.
@@ -80,3 +76,9 @@ Replies are also Markdown text. The difference is that they MUST refer to one ro
   ]
 }
 ```
+
+## Things to be added later
+
+- branch merge kind (specifying a URL from where to fetch the branch to be merged)
+- inline file comments kind
+- review type with multiple inline comments? -- or maybe this could just be an extension of the "reply" kind above
--- a/34.md
+++ b/34.md
@@ -77,9 +77,12 @@ Replies are also Markdown text. The difference is that they MUST refer to one ro
   "kind": 1622,
   "content": "<markdown text>",
   "tags": [
-    ["a", "30617:<base-repo-owner-pubkey>:<base-repo-id>"],
-    ["e", "<root-event-id>"],
-    ["p", "<mentioned-pubkey>"]
+    ["a", "30617:<base-repo-owner-pubkey>:<base-repo-id>", "<relay-url>"],
+    ["e", "<issue-or-patch-id-hex>", "", "root"],
+
+    // other "e" and "p" tags should be applied here when necessary, following the threading rules of NIP-10
+    ["p", "<patch-author-pubkey-hex>", "", "mention"],
+    // ...
   ]
 }
 ```
--- a/34.md
+++ b/34.md
@@ -87,8 +87,9 @@ Replies are also Markdown text. The difference is that they MUST refer to one ro
 }
 ```
 
-## Things to be added later
+## Possible things to be added later
 
-- branch merge kind (specifying a URL from where to fetch the branch to be merged)
-- inline file comments kind
-- review type with multiple inline comments? -- or maybe this could just be an extension of the "reply" kind above
+- "status" kind (for letting people know a patch was merged or an issue was fixed or won't be fixed)
+- "branch merge" kind (specifying a URL from where to fetch the branch to be merged)
+- "cover letter" kind (to which multiple patches can refer and serve as a unifying layer to them)
+- inline file comments kind (we probably need one for patches and a different one for merged files)
--- a/34.md
+++ b/34.md
@@ -48,7 +48,7 @@ Patches can be sent by anyone to any repository. Patches to a specific repositor
     ["commit", "<current-commit-id>"],
     ["parent-commit", "<parent-commit-id>"],
     ["commit-pgp-sig", "-----BEGIN PGP SIGNATURE-----..."],
-    ["committer", "<name>", "<email>", "<timezone offset in minutes>"],
+    ["committer", "<name>", "<email>", "<timestamp>", "<timezone offset in minutes>"],
   ]
 }
 ```
--- a/34.md
+++ b/34.md
@@ -73,7 +73,7 @@ Issues are Markdown text that is just human-readable conversational threads rela
 
 ## Replies
 
-Replies are also Markdown text. The difference is that they MUST refer to one root thread event. Their root doesn't have to be a `kind:1621` _issue_, but also a `kind:1617` patch event.
+Replies are also Markdown text. The difference is that they MUST be issued as replies to either a `kind:1621` _issue_ or a `kind:1617` _patch_ event. The threading of replies and patches should follow NIP-10 rules.
 
 ```jsonc
 {
@@ -85,6 +85,7 @@ Replies are also Markdown text. The difference is that they MUST refer to one ro
 
     // other "e" and "p" tags should be applied here when necessary, following the threading rules of NIP-10
     ["p", "<patch-author-pubkey-hex>", "", "mention"],
+    ["e", "<previous-reply-id-hex>", "", "reply"],
     // ...
   ]
 }
--- a/34.md
+++ b/34.md
@@ -20,6 +20,7 @@ Git repositories are hosted in Git-enabled servers, but their existence can be a
     ["d", "<repo-id>"],
     ["name", "<human-readable project name>"],
     ["description", "brief human-readable project description>"],
+    ["head", "<sha1-hex>"], // the current repository commit head, as hex
     ["web", "<url for browsing>"], // a webpage url, if the git server being used provides such a thing
     ["clone", "<url for git-cloning>"], // a url to be given to `git clone` so anyone can clone it
     ["patches", "<relay-url>"], // this signals that patches are welcome and will be searched for in this relay
@@ -28,7 +29,7 @@ Git repositories are hosted in Git-enabled servers, but their existence can be a
 }
 ```
 
-The tags `web`, `clone`, `patches` and `issues` can be specified multiple times.
+The tags `web`, `clone`, `patches` and `issues` can be specified multiple times. Except `d`, all tags are optional.
 
 ## Patches
 
--- a/34.md
+++ b/34.md
@@ -44,6 +44,9 @@ Patches can be sent by anyone to any repository. Patches to a specific repositor
     ["p", "<repository-owner>"],
     ["p", "<other-user>"], // optionally send the patch to another user to bring it to their attention
 
+    // for the first patch in a thread or series
+    ["t", "root"],
+
     // optional tags for when it is desirable that the merged patch has a stable commit id
     // these fields are necessary for ensuring that the commit resulting from applying a patch
     // has the same id as it had in the proposer's machine -- all these tags can be omitted
--- a/34.md
+++ b/34.md
@@ -44,3 +44,36 @@ Patches can be sent by anyone to any repository. Patches to a specific repositor
   ]
 }
 ```
+
+## Branch merge
+
+To be defined.
+
+## Issues
+
+Issues are Markdown text that is just human-readable conversational threads related to the repository: bug reports, feature requests, questions or comments of any kind.
+
+```jsonc
+{
+  "kind": 1621,
+  "content": "<markdown text>",
+  "tags": [
+    ["a", "30617:<base-repo-owner-pubkey>:<base-repo-id>"]
+  ]
+}
+```
+
+## Replies
+
+Replies are also Markdown text. The difference is that they MUST refer to one root thread event. Their root doesn't have to be a `kind:1621` _issue_, but also a `kind:1617` patch event.
+
+```jsonc
+{
+  "kind": 1622,
+  "content": "<markdown text>",
+  "tags": [
+    ["a", "30617:<base-repo-owner-pubkey>:<base-repo-id>"],
+    ["e", "<root-event-id>"],
+  ]
+}
+```
--- a/34.md
+++ b/34.md
@@ -40,7 +40,8 @@ Patches can be sent by anyone to any repository. Patches to a specific repositor
   "kind": 1617,
   "content": "<patch>", // contents of <git format-patch>
   "tags": [
-    ["a", "30617:<base-repo-owner-pubkey>:<base-repo-id>"]
+    ["a", "30617:<base-repo-owner-pubkey>:<base-repo-id>"],
+    ["p", "<repository-owner>"]
   ]
 }
 ```
@@ -58,7 +59,8 @@ Issues are Markdown text that is just human-readable conversational threads rela
   "kind": 1621,
   "content": "<markdown text>",
   "tags": [
-    ["a", "30617:<base-repo-owner-pubkey>:<base-repo-id>"]
+    ["a", "30617:<base-repo-owner-pubkey>:<base-repo-id>"],
+    ["p", "<repository-owner>"]
   ]
 }
 ```
@@ -74,6 +76,7 @@ Replies are also Markdown text. The difference is that they MUST refer to one ro
   "tags": [
     ["a", "30617:<base-repo-owner-pubkey>:<base-repo-id>"],
     ["e", "<root-event-id>"],
+    ["p", "<mentioned-pubkey>"]
   ]
 }
 ```
--- a/34.md
+++ b/34.md
@@ -45,9 +45,12 @@ Patches can be sent by anyone to any repository. Patches to a specific repositor
     ["p", "<other-user>"], // optionally send the patch to another user to bring it to their attention
 
     // optional tags for when it is desirable that the merged patch has a stable commit id
+    // these fields are necessary for ensuring that the commit resulting from applying a patch
+    // has the same id as it had in the proposer's machine -- all these tags can be omitted
+    // if the maintainer doesn't care about these things
     ["commit", "<current-commit-id>"],
     ["parent-commit", "<parent-commit-id>"],
-    ["commit-pgp-sig", "-----BEGIN PGP SIGNATURE-----..."],
+    ["commit-pgp-sig", "-----BEGIN PGP SIGNATURE-----..."], // empty string for unsigned commit
     ["committer", "<name>", "<email>", "<timestamp>", "<timezone offset in minutes>"],
   ]
 }
--- a/34.md
+++ b/34.md
@@ -23,19 +23,18 @@ Git repositories are hosted in Git-enabled servers, but their existence can be a
     ["tag", "head", "<sha1-hex>"], // the current repository commit head, as hex
     ["web", "<url for browsing>", ...], // a webpage url, if the git server being used provides such a thing
     ["clone", "<url for git-cloning>", ...], // a url to be given to `git clone` so anyone can clone it
-    ["patches", "<relay-url>", ...], // this signals that patches are welcome and will be searched for in this relay
-    ["issues", "<relay-url>", ...] // this signals that bug reports, questions, comments and so on are welcome
+    ["relays", "<relay-url>", ...] // relays that this repository will monitor for patches and issues
   ]
 }
 ```
 
-The tags `web`, `clone`, `patches` and `issues` can have multiple values.
+The tags `web`, `clone`, `relays` can have multiple values.
 
 Except `d`, all tags are optional.
 
 ## Patches
 
-Patches can be sent by anyone to any repository. Patches to a specific repository SHOULD be sent to the relays specified in that repository's announcement event's `"patches"` tag. Patch events SHOULD include an `a` tag pointing to that repository's announcement address.
+Patches can be sent by anyone to any repository. Patches to a specific repository SHOULD be sent to the relays specified in that repository's announcement event's `"relays"` tag. Patch events SHOULD include an `a` tag pointing to that repository's announcement address.
 
 ```jsonc
 {
@@ -50,7 +49,7 @@ Patches can be sent by anyone to any repository. Patches to a specific repositor
 
 ## Issues
 
-Issues are Markdown text that is just human-readable conversational threads related to the repository: bug reports, feature requests, questions or comments of any kind.
+Issues are Markdown text that is just human-readable conversational threads related to the repository: bug reports, feature requests, questions or comments of any kind. Like patches, these SHOULD be sent to the relays specified in that repository's announcement event's `"relays"` tag.
 
 ```jsonc
 {
--- a/34.md
+++ b/34.md
@@ -42,6 +42,7 @@ Patches can be sent by anyone to any repository. Patches to a specific repositor
   "tags": [
     ["a", "30617:<base-repo-owner-pubkey>:<base-repo-id>"],
     ["p", "<repository-owner>"],
+    ["p", "<other-user>"], // optionally send the patch to another user to bring it to their attention
 
     // optional tags for when it is desirable that the merged patch has a stable commit id
     ["commit", "<current-commit-id>"],
--- a/34.md
+++ b/34.md
@@ -20,16 +20,18 @@ Git repositories are hosted in Git-enabled servers, but their existence can be a
     ["d", "<repo-id>"],
     ["name", "<human-readable project name>"],
     ["description", "brief human-readable project description>"],
-    ["head", "<sha1-hex>"], // the current repository commit head, as hex
-    ["web", "<url for browsing>"], // a webpage url, if the git server being used provides such a thing
-    ["clone", "<url for git-cloning>"], // a url to be given to `git clone` so anyone can clone it
-    ["patches", "<relay-url>"], // this signals that patches are welcome and will be searched for in this relay
-    ["issues", "<relay-url>"] // this signals that bug reports, questions, comments and so on are welcome
+    ["tag", "head", "<sha1-hex>"], // the current repository commit head, as hex
+    ["web", "<url for browsing>", ...], // a webpage url, if the git server being used provides such a thing
+    ["clone", "<url for git-cloning>", ...], // a url to be given to `git clone` so anyone can clone it
+    ["patches", "<relay-url>", ...], // this signals that patches are welcome and will be searched for in this relay
+    ["issues", "<relay-url>", ...] // this signals that bug reports, questions, comments and so on are welcome
   ]
 }
 ```
 
-The tags `web`, `clone`, `patches` and `issues` can be specified multiple times. Except `d`, all tags are optional.
+The tags `web`, `clone`, `patches` and `issues` can have multiple values.
+
+Except `d`, all tags are optional.
 
 ## Patches
 
--- a/34.md
+++ b/34.md
@@ -42,7 +42,13 @@ Patches can be sent by anyone to any repository. Patches to a specific repositor
   "content": "<patch>", // contents of <git format-patch>
   "tags": [
     ["a", "30617:<base-repo-owner-pubkey>:<base-repo-id>"],
-    ["p", "<repository-owner>"]
+    ["p", "<repository-owner>"],
+
+    // optional tags for when it is desirable that the merged patch has a stable commit id
+    ["commit", "<current-commit-id>"],
+    ["parent-commit", "<parent-commit-id>"],
+    ["commit-pgp-sig", "-----BEGIN PGP SIGNATURE-----..."],
+    ["committer", "<name>", "<email>", "<timezone offset in minutes>"],
   ]
 }
 ```


jb55:
awesome, this is very close to what I was converging on.

mikedilger:
When I looked into this topic, I discovered https://git-scm.com/book/en/v2/Git-on-the-Server-Smart-HTTP
where you can avoid SSH key management and use HTTP digest passwords for pushing (on linux git comes with a CGI server in /usr/bin/git-core/git-http-backend).  I thought that could work with NIP-98.

34.md
```diff
@@ -0,0 +1,79 @@
...
+## Repository announcements
+
+Git repositories are hosted in Git-enabled servers, but their existence can be announced using Nostr events, as well as their willingness to receive patches, bug reports and comments in general.
+
+```jsonc
+{
+  "kind": 30617,
+  "content": "",
+  "tags": [
+    ["d", "<repo-id>"],
```
Giszmo, line 20: A word on this, please. I can't find a repo-id definition

34.md
```diff
@@ -0,0 +1,79 @@
...
+  "tags": [
+    ["a", "30617:<base-repo-owner-pubkey>:<base-repo-id>"]
+  ]
+}
+```
+
+## Branch merge
+
+To be defined.
+
```
Giszmo, line 51: I would remove these lines and add some `## Outlook` at the bottom to mention what else this nip could cover. Furthermore are you thinking of a "merge request" event here? As the different clone urls in the original repo might use different branch names, the target needs one specific url to reference the branch to merge into. As branches can be renamed, these events have no absolute cryptographic reference unless we add one via the branches heads.

MRs call for code reviews ...


````suggestion
## Merge Request

Notify repository owner of an improvement under development.

```jsonc
{
  "kind": 31617,
  "content": "<Description using markdown>",
  "tags": [
    ["d", "<something unique / UUID>"],
    ["a", "30617:<base-repo-owner-pubkey>:<base-repo-id>"],
    ["title", "<merge request title>"], // 
    ["state", "[wip|ready|final]"], // wip: share an idea, knowing it's not ready to merge, ready: author thinks it can be merged, final: author doesn't intend to update the source repo after feedback
    ["source", "<url for git-cloning>", "<branch name>", "<current branch head sha1-hex>"], // 
    ["target", "<url for git-cloning>", "<branch name>", "<current branch head sha1-hex>"], // 
  ]
}
```

## Merge Request Review

```jsonc
{
  "kind": 2617,
  "content": "<summary/comment allowing markdown>",
  "tags": [
    ["a", "30617:<base-repo-owner-pubkey>:<base-repo-id>"],
    ["a", "31617:<merge-request-owner-pubkey>:<merge-request-d-tag>"],
    ["type", "[ACK|NACK|...]"],
    ["anchor", "<source-commit-id>", "<target-commit-id>"],
    ["source-comment", "<file-with-path>:<line>", "<comment>"],
    ["target-comment", "<file-with-path>:<line>", "<comment>"],
  ]
}
```

Both `a` tags are required. The other tags are optional. `type` and `anchor` may only be used once.

````

File: 34.md [hunk 3]
```diff
@@ -0,0 +1,79 @@
...
+## Patches
+
+Patches can be sent by anyone to any repository. Patches to a specific repository SHOULD be sent to the relays specified in that repository's announcement event's `"patches"` tag. Patch events SHOULD include an `a` tag pointing to that repository's announcement address.
+
+```jsonc
+{
+  "kind": 1617,
+  "content": "<patch>", // contents of <git format-patch>
+  "tags": [
+    ["a", "30617:<base-repo-owner-pubkey>:<base-repo-id>"]
```
Giszmo, line 43: Allow attribution for a patch

```suggestion
    ["a", "30617:<base-repo-owner-pubkey>:<base-repo-id>"],
    ["author", "<git-author>"] // typically "full name <name@e.mail>"
```

Giszmo (commented):
With a lot of stuff being optional in this nip already I took a shot at more stuff we'd definitely need long term but not urgently on the first shot.

buttercat1791:
This proposal has a lot in common with the [NIP-62](https://github.com/nostr-protocol/nips/pull/986) PR I opened recently.  Perhaps we can merge them.

I think NIP-62 proposes a general solution that can be applied to the Git use case, whereas NIP-32 is very Git-specific.  Git is a large use case, so it could make sense to have a NIP solely for that purpose; but in general, don't we want NIPs to be reusable and broadly applicable?

34.md
fiatjaf: Yes, the branch merge thing would probably have to specify a branch name and a URL from where to fetch it. The person calling for a branch merge must have a git server somewhere. I don't know the details.

I was also thinking of another kind for inline comments of files, but all of that can be done much later I think.

34.md
fiatjaf: Can't this be done at the commit level? The patches already names on `git format-patch`.

fiatjaf:
@buttercat1791 no, I think having NIPs not be general is better. Of course it depends and we must think of it case-by-case, but as a rule of thumb less general ends up being better.

34.md
fiatjaf: It's just an arbitrary string, as in all other `d` tags.

fiatjaf:
@jb55 what do you think is missing or could be changed here?

34.md
Giszmo: Wouldn't it makes sense to prescribe this? If we used the initial commit hash for example, finding/detecting forks would be easier. `fiatjaf:30617:helloWorld` might be a very different repo than `giszmo:30617:helloWorld` but `giszmo:30617:<sha256-of-bitcoin-root-commit>` ... would hopefully match between all forks.

34.md
Giszmo: The MR state was just me being creative. An issue would usually not include the fix and when you want to send a fix, you can do so with a patch. The MR with a final state would be better done with a patch but on the other hand, maybe you want to propose a 5GB change, so a patch wouldn't work but if you host a fork, you can propose the 5GB change that way.

34.md
```diff
@@ -0,0 +1,84 @@
...
+The tags `web`, `clone`, `patches` and `issues` can be specified multiple times. Except `d`, all tags are optional.
+
+## Patches
+
+Patches can be sent by anyone to any repository. Patches to a specific repository SHOULD be sent to the relays specified in that repository's announcement event's `"patches"` tag. Patch events SHOULD include an `a` tag pointing to that repository's announcement address.
+
+```jsonc
+{
+  "kind": 1617,
+  "content": "<patch>", // contents of <git format-patch>
```
Giszmo, line 41:
```suggestion
  "content": "<patch|patch-url>", // contents of <git format-patch> or a URL where that patch can be found as plain text or zipped plain text. This is a fallback for huge patches.
```

alexgleason:
Aside fom "head", the data format is almost like package.json

RandyMcMillan:
git commit proof of work miner

#gnostr 

https://github.com/gnostr-org/gnostr-legit

alexgleason:
It would be worth giving ForgeFed a look over: https://forgefed.org/spec/ 

GitLab is currently implementating this: https://gitlab.com/groups/gitlab-org/-/epics/11247

34.md
mikedilger: I like this idea a lot - something github doesn't do and we could - that allows people to find all the forks.  But this particular event contains data that would differ between forks.

SilberWitch:
@fiatjaf I would be interested to see your analysis on why there needs to be a specific NIP for each type of version control system, when we could easily define the same solution for multiple types.

Are we really to have a seperate NIP for each brand name covered?

Giszmo:
@alexgleason this nip is explicitly an interoperability layer between gitlab, gitea, github and your various local git instances.

From the document you linked:

> The gist of it is: what people really want is to have one global "Gitlab network" to be able to interact between various projects without having to register on each of their hosts.

The gist of this nip here: what people really want is to have one global "Git network" to be able to interact between various projects without having to register on each of their hosts.

34.md
fiatjaf: No, must be the full patch otherwise we will go mad.

fiatjaf:
@SilberWitch not for each version control system, just for git. No one uses the others.

git is very specific and an open protocol of its own and requires special handling. It's also used by a ton of people, so it makes no sense to try to shove it into some generic thing that wouldn't know about its intricacies. We would still need to write software specific for git with hardcoded git stuff in it, so why not define that in a specific NIP?

Now for some kind of plaintext diffing that doesn't come with all the crazy features that git has (including its network effect) we can come up with our own scheme, like the NIP-62 proposal linked above. Indeed it would be very useful given that there is basically no decent version-control of anything except code in this world. Many people have tried to apply git or git-like approaches to other things and as far as I know never had any success.

fiatjaf:
https://github.com/fiatjaf/nak, https://github.com/nbd-wtf/go-nostr and https://github.com/fiatjaf/gitstr are now accepting patches over Nostr. Check their READMEs.

vitorpamplona:
This looks really good. Can we find a common way to mark the patch/issue as merged/closed?

buttercat1791:
@fiatjaf As far as I can tell, the core concept behind both NIP-62 and NIP-34 is pretty much the same: define an event to announce something happening off of Nostr (such as an update to a Git repository) and offer clients hints as to where to find out more (such as a git server URL).

Having some common standard for announcing off-Nostr events would be valuable, I think.  Git-specific things, like patches, issues, etc. can and should be defined separately, but I think the underlying concept of announcing off-Nostr things is inherently reusable and should be defined as such.

34.md
mattn: Some OSS products which fork has become the official developer because the original author retired the development.

https://github.com/ctrlpvim/ctrlp.vim

34.md
```diff
@@ -0,0 +1,84 @@
...
+
+```jsonc
+{
+  "kind": 30617,
+  "content": "",
+  "tags": [
+    ["d", "<repo-id>"],
+    ["name", "<human-readable project name>"],
+    ["description", "brief human-readable project description>"],
+    ["head", "<sha1-hex>"], // the current repository commit head, as hex
```
jb55, line 23: what's this for? doesn't git handle this?

34.md
```diff
@@ -0,0 +1,84 @@
...
+    ["clone", "<url for git-cloning>"], // a url to be given to `git clone` so anyone can clone it
+    ["patches", "<relay-url>"], // this signals that patches are welcome and will be searched for in this relay
+    ["issues", "<relay-url>"] // this signals that bug reports, questions, comments and so on are welcome
+  ]
+}
+```
+
+The tags `web`, `clone`, `patches` and `issues` can be specified multiple times. Except `d`, all tags are optional.
+
+## Patches
```
jb55, line 34: how do you send a patch series?

34.md
gazhayes: Pull request events will need to be very specific about which repo they are targeting. Using the initial commit hash as the `d` tag makes it impossible to use this as the identifier for pull requests.

34.md
gazhayes: Isn't it simpler to send a pull request?

34.md
```diff
@@ -0,0 +1,84 @@
...
+Patches can be sent by anyone to any repository. Patches to a specific repository SHOULD be sent to the relays specified in that repository's announcement event's `"patches"` tag. Patch events SHOULD include an `a` tag pointing to that repository's announcement address.
+
+```jsonc
+{
+  "kind": 1617,
+  "content": "<patch>", // contents of <git format-patch>
+  "tags": [
+    ["a", "30617:<base-repo-owner-pubkey>:<base-repo-id>"],
+    ["p", "<repository-owner>"]
+  ]
```
gazhayes, line 45: Probably also want the commit that this patch is based on?

fiatjaf: Good point. How does `git send-email` handle that?

OK, I just checked and it doesn't add that automatically. How do people do it in practice?

34.md
fiatjaf: I don't know? How do people normally do it in `git send-email`? I know you can send multiple patches in the same email and you could do the same in the Nostr event. Then when applying with `git am -i <patch>` it works fine.

Or you could send multiple events? I'm not sure what is the best flow.

gazhayes:
I've been trying to solve the same problem in the context of Nostrocket, I've added my thoughts here: https://nostrocket.org/Snub/problems/c1cc39a3386b96d79dc5b5de335a88acd8476cce191efc76da79fc286955ddfb

34.md
fiatjaf: The idea was that people could optionally provide this to prevent untrusted git servers from serving bad code under their name.

34.md
gazhayes: > How do people do it in practice?

For projects that prefer patches rather than commits, the maintainers usually just try to apply the patch on the latest master and if it conflicts they reply and tell you to rebase it.

But for the use case here, that would just result in no one sending any more patches, so the best way to handle it is just copy/paste the latest HEAD that you're working on top of, and add it to a `base` tag (that's what I was using for nostrocket anyway).

Maintainers can then manually apply the patch on top of the correct commit and resolve any conflicts manually.

All of this just makes me think we should be using native git most of the time though rather than patches.

fiatjaf:
@gazhayes I left a comment on your post expressing my disagreements. But aside from that I agree with the rest. So your suggestion is just to do everything as I say here, but add a new kind for "pull request" that includes a URL to another repository from which the code is to be pulled? I like that. We can definitely add that to this NIP.

34.md
jb55: git send-email sends multiple emails when you give it multiple patches. it creates a proper email thread so that you can reply to each patch individually. the nostr equivalent would be building a nip-10 thread of patches.

jb55: > Isn't it simpler to send a pull request?

this implies you have write access to the repo in question, or that you have a hosted fork somewhere. it is much easier to just send patches as it requires no hosting.

34.md
jb55: > OK, I just checked and it doesn't add that automatically. How do people do it in practice?

you use `git-format-patch --base=HEAD^ HEAD`

this includes base-commit info in the patch itself, so you can know where the commit's parent.

it just includes it as text at the bottom of the patch: 

`base-commit: c67741983e3f07f2386eaa388cb8a1475e8e047`

34.md
jb55: the git url is authoritative though, someone could spoof the url, the hash would be different, and users would be none the wiser. maybe I'm still misunderstanding.

34.md
jb55: it might be interesting to make this explicit in a tag though!

jb55:
the equivalent in git+email is git-request-pull

34.md
```diff
@@ -0,0 +1,84 @@
...
+## Patches
+
+Patches can be sent by anyone to any repository. Patches to a specific repository SHOULD be sent to the relays specified in that repository's announcement event's `"patches"` tag. Patch events SHOULD include an `a` tag pointing to that repository's announcement address.
+
+```jsonc
+{
+  "kind": 1617,
+  "content": "<patch>", // contents of <git format-patch>
+  "tags": [
+    ["a", "30617:<base-repo-owner-pubkey>:<base-repo-id>"],
```
jb55, line 43: we probably also want a "version" tag. patches tend to have more than one version -v2, -v3, etc

jb55: I mean lots of this stuff already shows up in the patch itself, so not sure if worth it. but just something to consider.

34.md
```diff
@@ -0,0 +1,84 @@
...
+    ["p", "<repository-owner>"]
+  ]
+}
+```
+
+## Replies
+
+Replies are also Markdown text. The difference is that they MUST refer to one root thread event. Their root doesn't have to be a `kind:1621` _issue_, but also a `kind:1617` patch event.
```
jb55, line 66: I think the MUST restriction on the root thread is a bit much, you should be able to reply to patch 02/08 in a nip10 thread of patches no?

jb55: unless this means at least one root thread event along with the thing you're replying to. maybe this could be mentioned here.

jb55: the root item could be the cover letter. see `git format-patch --cover-letter ..branch` for an example.

34.md
DanConwayDev: Why not make it the root commit id? For forks it could be the id of the earliest unique commit. The repository COULD contain a config file which specifies this id so that clients MAY access the config to automatically identify the correct id.

34.md
DanConwayDev: I had this idea too. This obviously relies on a new 30617 event being issued every time a maintainer pushed to the git server.

Each maintainer would need to use a nostr git-remote-helper that proxies interaction with the git server and issues the event automatically. 

My implementation was a little more involved and allowed tracking additional branches. It also listed the tip's branch ancestors so that clients could tell how far behind they are without talking to the git server.
```
[
 ["main", "<commid_id>", "<parent commit_id shorthand>", "<grandparent commit_id shorthand>" ....]
]
```
I ultimately concluded that most projects that care about this attack vector are signing their commits which, assuming the signatures are being validated, addresses the concerns without having to introduce a new tool.

gazhayes:
> @gazhayes So your suggestion is just to do everything as I say here, but add a new kind for "pull request" that includes a URL to another repository from which the code is to be pulled? I like that. We can definitely add that to this NIP.

I also use a list of Maintainers in the Repo Announcement, the server allows these keys to push 
 (nip98). Clients treat the event as "replaceable" by any of the maintainer keys.

34.md
```diff
@@ -0,0 +1,84 @@
...
+  "kind": 1622,
+  "content": "<markdown text>",
+  "tags": [
+    ["a", "30617:<base-repo-owner-pubkey>:<base-repo-id>"],
+    ["e", "<root-event-id>"],
+    ["p", "<mentioned-pubkey>"]
+  ]
+}
+```
+
```
gazhayes, line 79:
```suggestion
## Pull Requests
{
    "kind": 1623
    "content": ""
    "tags": [
        ["commit": "<sha1-commit-hash>"], //MUST
        ["web": "url where the repo can be viewed"], //OPTIONAL
        ["git": "url of standard git API"], //MUST
        ["p": "maintainer(s) pubkey"], //OPTIONAL
        ["a", "30617:<base-repo-owner-pubkey>:<base-repo-id>"] //MUST
      ]
}
```

34.md
DanConwayDev: What about a pull request event which patches tag? later updates based on feedback or rebases could also be associated with the same pull request.

DanConwayDev:
> I also use a list of Maintainers in the Repo Announcement, the server allows these keys to push (nip98). Clients treat the event as "replaceable" by any of the maintainer keys.

This nip should support more than one maintainer. I also used a similar approach as you in my implementation but a list of npubs in a maintainers.yaml file in the repository would take precedent.

34.md
DanConwayDev: Each proposed commit / patch should have its own event so there is a record available for those who are interested in reviewing the of history how a change evolved over time before it got merged even if the contributor deletes the repo on their git server.

Patches that are too large for an event should have a special patch event type called a stub. This points to a git server for the full commit but contains most of the details of the patch but with only the large file diffs removed. 
This way the salient details will be retained but large file diffs (often binary blobs) removed.

I agree their should be a PR event to hang multiple patches on - see https://github.com/nostr-protocol/nips/pull/997#discussion_r1464671290.

fiatjaf:
I don't like the idea of having more than one maintainer. I think that is reminiscent of a GitHub feature that doesn't fit well in Nostr. The same concept can be done by having something like multiple people each announce the same repository URL with the same `"head"`, by demonstrating that they are in agreement about the state of the repository. At the same time in the repository code itself it can be written manually somewhere that Y and X are the maintainers.

jb55:
On Wed, Jan 24, 2024 at 02:11:37AM -0800, DanConwayDev wrote:
>@DanConwayDev commented on this pull request.
>
>What about a pull request event which patches tag? later updates based
>on feedback or rebases could also be associated with the same pull
>request.

This implies that there is always a pull request? That overcomplicates
things a bit. I think this is because most people are used to the GitHub
workflow vs the standard git workflow. In git you can just send a single
patch to someone and not open a "pull request". In git pull requests
mean a specific thing, its just a "hey pull my branch from over here".
Which is much different than sending patches and patch sets.

In git, code review is done on patches, not pull requests.

If you need to send an updated revision of your patch set, you send a
whole new set, invaliding the last one. You also include a patch
changelog which describes what has changes between patchset revision
1->2, 2->3, etc.

This is just not that common in GitHub land for some reason. Sending new
patch changelogs is not common, mainly because you just force push your
branch.

I think I much prefer just having patches and patch revisions, then you
don't need pull requests at all. You just need a way to show in you
client that the previous revisions have been invalidated, but you can
always reference them and their previous threaded code reviews
independently.

GitHub handles this very poorly imo. It doesn't even have threaded
review.

jb55:
Some more context on my anti-github-way-of-doing-things stance since I'm probably the minorty as we are on github while reviewing this:

from https://njump.me/nevent1qqsps7fyf8uy36u73v62x2hcmlplhqq75jyxy7jkuqhk8y5xz0nkc9czyqewrqnkx4zsaweutf739s0cu7et29zrntqs5elw70vlm8zudr3y2y097gc :

>Projects which switch to GitHub tend to have overall commit quality go down IMO, because the system
>
>(a) makes it nearly impossible to review commit messages, so people eventually degrade to writing really bad ones
>
>(b) makes it nearly impossible to review a rebased set of changes except redoing the entire review from square one, so people don't rebase
>
>(c) punishes both users and reviewers who want to work with a rebased patch series by displaying the series out of order -- even for a completely linear history (it resorts based on author timestamp, not even committer timestamp), and
>
>(d) punishes reviewers/users when they attempt to review individual commits by making it harder to see and follow these comments (though it has gotten much better on this front).
>
>There are combination effects too. People to write really bad commit messages for all the additional "fixups" they have.
>
>People notice that commits don't bisect nicely, and instead of understanding that the broken code review system they wrote was the problem, they instead offer a new "squash merge" option, thus destroying all the carefully separated commits that help people understand the individual steps toward the new feature and making it impossible for anyone in the future to review it incrementally.
>
>You may say it's a workflow choice for some people to just squash all their stuff together at their option, which would be fine, but the problem is most developers don't take the time to think, and someone in charge of the project notices that they keep getting un-bisectable meaningless commits unless they force everyone in the project to use squash merging.
>
>Now they are punishing me for creating clean separate commits and forcing them all to be squashed -- all as an ugly workaround to the basic tool being broken.
>
>You can work around this by making a long sequence of PRs, one per what you intend to be a commit, and try to track the hierarchy -- something that GitHub certainly doesn't make easy.
>
>And then each PR becomes a trivial small change, and you are back to merging individual commits, and writing your own tools to manage a hierarchy of PRs...and reviewers hate you if you do that because it's extremely onerous on them.
>
>GitHub PRs aren't just hard to use, they literally degrade the quality of the code for people who have to use it. I've seen it happen with many projects.
>

https://public-inbox.org/git/CABPp-BG2SkH0GrRYpHLfp2Wey91ThwQoTgf9UmPa9f5Szn+v3Q@mail.gmail.com/

another good talk on why the linux kernel uses patches+email instead of the PR model:

https://www.youtube.com/watch?v=L8OOzaqS37s

maybe this can be patches+nostr in the future. Let's not kill that by modeling this on the broken github model.

DanConwayDev:
I agree with that analysis. We don't have to recreate the way github does PRs.

What I am proposing is a event that serves as the root of a thread containing patches and conversations related to a desired change.

Often, as the conversation unfolds, patches need to be refactored and updated based on feedback and other changes made to the product. Having a root event that links this together is useful.

As I understand the patch model, each new update is issued in a self-contained patch-set without smooth connection to conversation that proceeded it after the previous iteration.

Perhaps we can try and take the best from both approaches. 

Could we flag the first patch event in a 'new desired change', in a way that that serves the purpose of a root event?

jb55:
On Wed, Jan 24, 2024 at 07:47:51AM -0800, DanConwayDev wrote:
>I agree with that analysis. We don't have to recreate the way github
>does PRs.
>
>What I am proposing is a event that serves as the root of a thread
>containing patches and conversations related to a desired change.
>
>Often, as the conversation unfolds, patches need to be refactored and
>updated based on feedback and other changes made to the product. Having
>a root event that links this together is useful.
>
>As I understand the patch model, each new update is issued in a
>self-contained patch-set without smooth connection to conversation that
>proceeded it after the previous iteration.

This is a client issue, typically you would match the different patch
threads by referencing the title and looking for a -v2

I just nuke a thread once I give feedback and ask for an update, so
older patch revisions are not that interesting. Tools that want to show
these historical revisions could just match based on the patch title or
search for them? 

If we wanted something more explicit, we could specify some kind of
tag that can be used to link different patchset revisions. Maybe this
can be generated canonically somehow, if not I think it would have to be
used provided?

I say this because usually it's a disaster to include v2 patches into
the same thread as v1, since things get all mixed up and is noisy.

My ideal scenario is having separate root threads for each patch
revision, where reviews can be isolated in different stages. You can
still reply to older messages from older revisions, but they are
isolated and whole threads can be archived idependently from each other.

I've envisioning a gitstr-like client where I manage code review similar to
how I do it with mutt+email, where I can nuke entire patch series reviews
when requesting a v2.

gazhayes:
@jb55 when you merge a patch, how do you make it so that the commit history shows the patch author rather than the maintainer who merged it?

DanConwayDev:
you need to apply the patch and update the 'committer' attached to the commit to reflect the 'author'.

DanConwayDev:
https://gitworkshop.dev is the web client I have been working on. It currently follows #1003 which is not all dissimilar to what @fiatjaf proposed.

jb55:
On Wed, Jan 24, 2024 at 11:54:09PM -0800, gsovereignty wrote:
>@jb55 when you merge a patch, how do you make it so that the commit
>history shows the patch author rather than the maintainer who merged
>it?

git-am and git-apply does this for you from the patch...

jb55:
On Thu, Jan 25, 2024 at 02:52:50AM -0800, DanConwayDev wrote:
>If you capture the 'committer' and pgp signature from the commit when
>creating the patch event you can fully recreate the commit so the
>commit id matches.

It's definitely an important feature for some projects like
bitcoin-core. I didn't realize you could recreate the commit id fully
just from these extra bits.

Very often in projects there are processes that mutate the commit:

- maintainers adding changelog entries

- committers fixing up conflicts and bugs without having to wait on the
   author

- patches can be rebased without waiting on the author, who might not
   have the same tree state which is extremely common in larger projects
   with lots of trees.

Mutability is really good in these scenarios. Otherwise you have really
long back-and-forth between the committer and author which makes things
extremely slow, which bitcoin-core devs know really well. Maybe that is
fine for projects like bitcoin, but that pretty much the exception.

I see this as more as something that would be good to support as a
feature-add vs forcing it on everyone from the start.

I personally have given up on pgp. It definitely shines in key rotation,
I give it that, but key rotation in pgp is a huge pain. Right now I
can't really rotate my pgp keys on my yubikey because I have to go
through 100 arcane gpg smartcard steps that I can't be bothered to
figure out again. nostr has already replaced pgp as my web of trust
model for most things.

And because of the need to rebase things so often PGP sigs end up
getting discarded anyway. Maybe we could somehow sign things using your
nostr key as a simpler alternative, or at least have a nip19 identifier
in the commit itself to link back to the note and patch discussion for
reference. This would at least give you a signed reference of the
orginal work that you can refer to even if the patch was rebased by the
committer.

There could even me tools that audit the diff between what the author
sent in and what was finally committed just from looking at the nostr
nip19 patch reference.

But yes, I still think preserving commits is useful for the
bitcoin-core-like scenarios! But let's not restrict ourselves to it.

	Will

34.md
Giszmo: wasn't familiar with format-patch. Please disregard.

34.md
Giszmo: Frequently dealing with MB patches containing hundreds of JPGs I guess we'll see about that but ok ... for now ...

Giszmo:
### Re: Critique on how GitHub does things

GitHub is the most successful code management tool and as such caters to maybe a broader audience than what any one of us might perceive as more efficient or more logical. I would try to not re-invent things on the protocol level and at least make sure to facilitate replication of the features people are used to on GitHub.

That said, a browser client can show a patch queue as a PR and allow code reviews just the same as they are on GitHub right now. Whether or not the applied patch appears squashed or the maintainer's client creates multiple commits is an implementation detail.

### Re: Multiple maintainers

I imagine we could have multiple maintainers using the same nsec via an nsec bunker. The project owner would have granular control over each maintainer's permissions without the need to extend relay features.

In general I strongly dislike that GitHub incentivizes multiple maintainers or contributors with special permissions in a repository and suspect nostr could provide the unifying layer that would make this unnecessary.

jb55:
On Thu, Jan 25, 2024 at 11:04:01AM -0800, Leo Wandersleb wrote:
>Frequently dealing with MB patches containing hundreds of JPGs I guess
>we'll see about that but ok ... for now ...

pull requests are better for this: a user hosts a repo somewhere and
dumps tons of data on in so that you can pull it down. I don't think we
want to encourage users to store image and binary data in nostr. 

I suspect some clients will try always have a pull request associated
with a patch or set of patches (github style), and some might just
support sending and dealing with patch, patchsets, and patch revisions
by themselves (mailing list style). Maybe it's best to support both
cases at the protocol level and just have clients build different
experiences for different dev workflows.

fiatjaf:
I was chatting with @DanConwayDev today and learned more about #1003.

The fundamental difference is that it has this `pr` event kind as its core building block that differs from this proposal here. The `commits` that proposal are pretty much like `patches` here. The patches reference the `pr`, so using the common Nostr thread model would yield something like this: image

Meanwhile I was trying the email flow a little bit and when there are multiple commit/patches git generates an email thread with one email message for each. However, if you specify the `--cover-letter` thing then git generates a blank message for you to fill in, which is set to be the first message in the thread -- and that corresponds to the `pr` event.

But then at this point I don't see why we can't just follow the normal Nostr thread model. Like @DanConwayDev suggested we can rename that `pr` kind to `cover_letter` and then it can be present or not. If it is present we use it as the root of the thread -- although clients can decide to ignore it if they want and just treat each patch as a standalone thing. If it is not present then clients just treat patches as standalone things.

image

Although this proposal is not clear about this I was always thinking that eventually people would submit normal plaintext comments as replies to patches and vice-versa.

One flow that seems to be useful in mind is someone opening an "issue", like a bug report, and then someone else commenting on it, then someone submitting a patch that fits in that same thread and fixes the bug. Does this make sense?

So in the end we have just a bunch of different kinds of stuff:

- `patch`, kind 1017
- `cover_letter`, kind 1018
- `branch_merge`, kind 1019
- `status`, kind 1020 (taken from #1003)
- `issue`, kind 1021
- `reply` (or just `comment`), kind 1022

And they all can refer to each other using NIP-10 rules and make a thread. The only rules would be that `cover_letter` and `issue` must always be the root -- and `reply` and `status` must never be the root. `patch` can be anywhere.

Does that make sense?

jb55:
On Thu, Jan 25, 2024 at 02:40:26PM -0800, fiatjaf_ wrote:
>And they all can refer to each other using NIP-10 rules and make a
>thread. The only rules would be that `cover_letter` and `issue` must
>always be the root -- and `reply` and `status` must never be the root.
>`patch` can be anywhere.

cover letters only really make sense for patches with more than 1 patch.
If you're sending a single patch I don't see why that couldn't be the
root of the thread. The content of the patch would be the "cover letter"
at that point.

jb55:
It's also common to send more than 1 patch without cover letters, this is very common. it encourages each patch to be self contained. In this situation the root of the thread is just the first patch. In email, the coverletter isn't different than a standard patch, it's just marked as "patch 0" instead of patch 1 and is skipped when applying patches. This saves us from using a new kind for cover letter.

34.md
cypherhoodlum: The `state` field is probably not the best way to approach this. For example, closing an issue or a PR doesn't have to publish a whole new Issue Event, having just one field changed. I think it's better to have a separate kind for closing issues and PRs.

fiatjaf:
Edited the NIP above to

- use just a single `relays` tag for the repository to monitor for patches and issues (two tags, `"patches"` and `"issues"`, was cumbersome, bloated and unnecessary).
- add optional fields copied from @DanConwayDev's proposal to allow a commit id to be stable when a patch is merged, which is required for the Bitcoin Core process and who knows where else.
- clarify that "reply" events should use the NIP-10 model for threading.

fiatjaf:
I think we should still add a "status" kind for maintainers to optionally notify everybody if a thing was merged or not or an issue was fixed or not and so on -- but I think we can do that later instead of overspecifying for now.

I'm not fully convinced it makes sense to have a "cover letter" event but I think it could be interesting if implemented as a fully optional thing -- but actually thinking now it could just be an "issue" event with patches referencing it. In any case we can also think about this later.

fiatjaf:
I just did this thing to browse patches on the web: https://patch34.pages.dev/
I'm not sure what it can evolve to.

34.md
```diff
@@ -42,7 +42,13 @@ Patches can be sent by anyone to any repository. Patches to a specific repositor
...
   "tags": [
     ["a", "30617:<base-repo-owner-pubkey>:<base-repo-id>"],
-    ["p", "<repository-owner>"]
+    ["p", "<repository-owner>"],
+
+    // optional tags for when it is desirable that the merged patch has a stable commit id
+    ["commit", "<current-commit-id>"],
+    ["parent-commit", "<parent-commit-id>"],
+    ["commit-pgp-sig", "-----BEGIN PGP SIGNATURE-----..."],
+    ["committer", "<name>", "<email>", "<timezone offset in minutes>"],
```
DanConwayDev, line 51: the committer timestamp is required as well

DanConwayDev:
diagram

This is how both the PR and patch model could bother be supported. I agree that a cover_letter should be optional - why repeat what should be in the commit message?

using a tag for either 'root' or 'root-version' provides flexibility for clients.

34.md
```diff
@@ -87,8 +87,9 @@ Replies are also Markdown text. The difference is that they MUST refer to one ro
 }
 ```
 
-## Things to be added later
+## Possible things to be added later
 
-- branch merge kind (specifying a URL from where to fetch the branch to be merged)
-- inline file comments kind
-- review type with multiple inline comments? -- or maybe this could just be an extension of the "reply" kind above
+- "status" kind (for letting people know a patch was merged or an issue was fixed or won't be fixed)
```
DanConwayDev, line 92: Status is quite important. Shouldn't we add this now? Its impossible to identify open proposals or issues unless you are a maintainer with a client that stores the status internally.

34.md
```diff
@@ -0,0 +1,95 @@
...
+    // ...
+  ]
+}
+```
+
+## Possible things to be added later
+
+- "status" kind (for letting people know a patch was merged or an issue was fixed or won't be fixed)
+- "branch merge" kind (specifying a URL from where to fetch the branch to be merged)
+- "cover letter" kind (to which multiple patches can refer and serve as a unifying layer to them)
```
DanConwayDev, line 94: as mentioned in https://github.com/nostr-protocol/nips/pull/997#issuecomment-1916213213 we should have a cover_letter as a separate kind than a patch because it is different in nature.

DanConwayDev (changes requested):
As a final note in my review: Should the key header items of the patch be optionally extract and put into nostr tags to make it easy for non-specialised clients to read or is this putting to much of a burden on clients that apply patches which would need to render it? 

Instead of repeating this data it could be extracted and removed. Here is the format:
```From [commit-id] Mon Sep 17 00:00:00 2001\nFrom: [author-name] <[author email]>\nDate: [authored-date]\nSubject: \[PATCH\] [commit-msg]\n---\n```

note date format here is: ```ddd, DD MMM YYY HH:mm:ss ZZ```

Example:
```
From 1263051aa4426937c5ef4f7616e06e9a8ea021e0 Mon Sep 17 00:00:00 2001\nFrom: William Casarin <jb55@jb55.com>\nDate: Mon, 22 Jan 2024 14:41:54 -0800\nSubject: [PATCH] Revert "mention: fix broken mentions when there is text is\n directly after"\n\nThis reverts commit af75eed83a2a1dd0eb33a0a27ded71c9f44dacbd.\n---\n
```

File: 34.md (hunk 5 above)
Comment (fiatjaf) [reply 34.md]: Makes sense. I've removed it.

File: 34.md (hunk 7 above)
Comment (fiatjaf) [reply 34.md:51]: I think for practical purposes this is better left to be read manually by the maintainer.

But in any case I've added this as an optional tag that can be used when people want a stable commit id, so it can be used for other purposes too I guess.

File: 34.md [hunk 14]
```diff
@@ -0,0 +1,95 @@
...
+
+```jsonc
+{
+  "kind": 30617,
+  "content": "",
+  "tags": [
+    ["d", "<repo-id>"],
+    ["name", "<human-readable project name>"],
+    ["description", "brief human-readable project description>"],
+    ["tag", "head", "<sha1-hex>"], // the current repository commit head, as hex
```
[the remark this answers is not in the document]
fiatjaf: I've removed it for now. There is probably a better way to do this. We'll see in the future.

34.md
```diff
@@ -0,0 +1,95 @@
...
+```
+
+## Issues
+
+Issues are Markdown text that is just human-readable conversational threads related to the repository: bug reports, feature requests, questions or comments of any kind. Like patches, these SHOULD be sent to the relays specified in that repository's announcement event's `"relays"` tag.
+
+```jsonc
+{
+  "kind": 1621,
+  "content": "<markdown text>",
```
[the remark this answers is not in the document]
fiatjaf: Do we really want Markdown here?

But yes, let's add a `title` tag.

34.md
```diff
@@ -0,0 +1,95 @@
...
+## Repository announcements
+
+Git repositories are hosted in Git-enabled servers, but their existence can be announced using Nostr events, as well as their willingness to receive patches, bug reports and comments in general.
+
+```jsonc
+{
+  "kind": 30617,
+  "content": "",
+  "tags": [
+    ["d", "<repo-id>"],
```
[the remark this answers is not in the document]
fiatjaf: What if I have two projects with the same root commit id?

34.md
```diff
@@ -0,0 +1,95 @@
...
+  "tags": [
+    ["a", "30617:<base-repo-owner-pubkey>:<base-repo-id>"],
+    ["p", "<repository-owner>"],
+
+    // optional tags for when it is desirable that the merged patch has a stable commit id
+    ["commit", "<current-commit-id>"],
+    ["parent-commit", "<parent-commit-id>"],
+    ["commit-pgp-sig", "-----BEGIN PGP SIGNATURE-----..."],
+    ["committer", "<name>", "<email>", "<timezone offset in minutes>"],
+  ]
```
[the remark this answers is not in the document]
fiatjaf: I'm confused by these quotes, `"["root"]`?

34.md
fiatjaf: What is the difference between a "cover letter" kind and an "issue" kind?

34.md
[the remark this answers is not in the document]
DanConwayDev: then there is a fork and the earliest unique commit id should be used instead. the repo event should contain a ["r","[commit-id]"] outlining this so the contributor doesn't need to identify it. 

Clients could make more sense of this if, in the case of a fork, 2 tags are used as follows: 

```suggestion
    ["d", "<repo-id>"], // the earliest unique commit id. usually the root commit id unless it's a fork.
    ["r", "<repo-id>"],
    ["r", "<first-commit-id>"], // if the first commit id isn't the earliest unique commit.
```

34.md
fiatjaf: Changed the text of this. See if it sounds better now.

34.md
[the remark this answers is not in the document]
fiatjaf: I think instead of writing all these `e` tags in the example we should just say these events should follow NIP-10 threading rules. It will be less confusing for people reading.

Now about the `t` tags I don't think they help. If a human is reading a list of patches or a thread with multiple patches and comments they will be able to see which patches are roots and which ones are v2, v3 and so, either from the context or from something written in the patch title. And if a machine is reading it or if someone wants to refer to a patch with absolute precision they should use the Nostr event id.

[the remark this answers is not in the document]
DanConwayDev: > I'm confused by these quotes, `"["root"]`?

typo fixed

34.md
DanConwayDev: A cover letter is always paired with a proposed change. An issue may result in zero or more changes in the future.

34.md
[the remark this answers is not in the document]
DanConwayDev: > I think instead of writing all these `e` tags in the example we should just say these events should follow NIP-10 threading rules. It will be less confusing for people reading.

I agree that what I wrote is confusing regarding threading however, its not clear how NIP-10 threading rules should be applied. Should the third patch in the set 'reply' to the second? a naive reading of the rules may suggest not unless it directly uses the change, but clients shouldn't be expected to automatically detect this. I think we this nip should clarify, for both consistency and as the decision effects comparability with the PR model.

The problem trying to be solved by stating that patches should reply to the previous patch in the PR model is as follows:
Assuming the commit id's cannot be replied upon, how can a client know if the patch is a intended to be used in combination with, or replace, another patch?

typo-PR-model

without the reply mechanism the typo fix 'amended' commit could be seen as a descendant, instead of a replacement by a client. Of course a human would see it, but we want clients to detect this and provide a smooth user experience.

> Now about the `t` tags I don't think they help. If a human is reading a list of patches or a thread with multiple patches and comments they will be able to see which patches are roots and which ones are v2, v3 and so, either from the context or from something written in the patch title. And if a machine is reading it or if someone wants to refer to a patch with absolute precision they should use the Nostr event id.

Here is the problem this is trying to solve: as a client I want to create a single subscription to return the only the root event of proposed changes so that I can display their titles as a list.

I suppose it is not too much of a problem to get all versions of proposed changes and filter them client side by filtering those that 'mention' other proposed changes. This would remove the `root-version` tag.

How else could this be achieved without the ["t","root"] tag or requesting every patch?

DanConwayDev:
Here is an idea for reducing duplicate patches in the patch model which wouldn't be feasible in the email client model:
patch-data-saving

jb55:
On Tue, Jan 30, 2024 at 08:31:35AM -0800, DanConwayDev wrote:
>Here is an idea for reducing duplicate patches in the patch model which wouldn't be feasible in the email client model:
>patch-data-saving

I think this just overcomplicates things. It's not uncommon that the
patch set is rebased each time they are sent anyways.

fiatjaf:
As @DanConwayDev pointed out it is important to have a way to identify the "root" of a series of patches and comments that refer to that in order to have UIs that display these patches in a coherent way.

I'm not thinking about just "patch sets", but also reviews of the same patch by the author after comments, or the cases in which someone might reply to someone else's patch proposing a different way to implement the same thing, or even amending someone else's patch with a patch to the patch -- I don't know.

We have some options:

1. enforce a tag `t=root` or something like that for the first patch in a series of patches so they can be queried on relays
2. download everything and assemble threads using NIP-10 rules on the client side
3. use a different kind to establish a root (either the "cover letter" kind with patches attached to it or the first patch in a series uses a different kind from the others)

@jb55 @gazhayes @vitorpamplona what do you think would be the best?

fiatjaf:
Actually now that I wrote that I think this might be a nice solution:

4. the first patch has an `a` tag pointing to the repository announcement, but the subsequent patches don't, they only refer to the initial patch, so they don't show in lists of patches for any given repo.

jb55:
On Wed, Feb 07, 2024 at 09:54:57AM -0800, fiatjaf_ wrote:
>1. enforce a tag `t=root` or something like that for the first patch in
>a series of patches so they can be queried on relays

This makes sense to me.

vitorpamplona:
I have very little knowledge of how git actually works (I never cared enough to learn the clusterfuck of commands that is git), so take this suggestion with a grain of salt. 

I would keep the Repo Announcement kind (30617) and create two chain heads as new event kinds: 
- Issue: (1621)
- Change Request (1618): -- new kind

These don't necessarily contain any description or patch but they SHOULD have a `subject` line. They just start a discussion thread in the repo. They are designed to help build the "GitHub" client and help repo owners navigate their issues and patches with sanity. Both of these will receive labels by the repo owner on their status (New, Accepted, Closed, etc).

From those two kinds, users can attach Comments/Replies (kind:1622) and Patches (1617). These events MUST point to an Issue OR a Change request which controls their lifecycle. People can submit a Patch during a discussion of an Issue (say to show how things could be done) and a Comment during the discussion of a Change Request. They are interchangeable. 

Both Comments and Patches should reference their parent event in the same way kind 1 replies work. So, patches can be applied to the main repo, but also to other patches (or after parent patches were applied to the repo). 

Once an Issue or a Change Request is marked as closed, all comments are replies are also closed.

vitorpamplona:
I don't know if NIP-34 is the place for this or not, but I need releases and release notes. :)

fiatjaf:
I think it is. I was thinking of having an event kind for tracking commits and tags. Like you could publish it on every commit, or on every tag so people following a repo could automatically update their local version, or CI services could have their actions triggered.

A tag event with a description is a "release".

vitorpamplona:
@DanConwayDev I see that you are using different event types for each status (Open, Closed, Resolved). Was that necessary?

DanConwayDev:
> @DanConwayDev I see that you are using different event types for each status (Open, Closed, Resolved). Was that necessary?

Originally I had a single status kind and the status name would be included as a hashtag. The approach of using separate kinds for each status, suggested by @fiatjaf, reduces the likelihood of status bloat.

I intend to add some optional tags to the applied status kind (1631):
1. `["applied-as-commit","<commit-id-1>", "<commit-id-2>",...] ` if the patch, or patches within the patch set, were applied
2. `["merge-commit","<merge-commit-id>"]` if the patches were recreated as commits, retaining their commit id, and merged.

I am currently reusing the same status kinds for root patches and issues. Perhaps that's less kinds for clients to worry about but the 'applied' status feels awkward as it really means 'resolved' in the context of an issue. Perhaps an additional kind could be added just for that.

It is also my intention to enables labels (with nip32) so that proposals and issues can be tagged with 'bug', 'feature' 'good-first-issue', 'vnext', 'repo-specific-label', etc. repo maintainers could optionally include a field in their repo event that specifies labels they would like to see used.

> I think it is. I was thinking of having an event kind for tracking commits and tags. Like you could publish it on every commit, or on every tag so people following a repo could automatically update their local version, or CI services could have their actions triggered.
> 
> A tag event with a description is a "release".

Are you suggesting an event per commit? whats the rationale?
My original [nip proposal 561](https://github.com/nostr-protocol/nips/pull/561) included kinds for commits, branches and tags. I moved away from that idea (and storing all commits diffs in events, removing the need for a git server) in favor of a more minimalist approach of including branch and tag states in the repo event as [mentioned in a comment](https://github.com/nostr-protocol/nips/pull/997#discussion_r1464484832). Repo events could be the authority, reducing trust in git_servers, allowing multiple mirrored git servers, which could be seamlessly swapped out by maintainers. This has the benefit of storing only a single replaceable event rather than 1000's of commit events, simplifying life for clients and leaving the heavy lifting of storing and serving git data efficiently to git server's which do this very well.

storing releases in a replaceable tag event is an interesting idea. links to files such as binaries, checksums could be added later, if some of these require the assistance of CI tools (eg platform specific binaries).

fiatjaf:
If no one opposes that I'm going to merge this and then @DanConwayDev can you open a PR adding the status stuff?

vitorpamplona:
Let's go! We got 3 clients implementing it.

gazhayes:
> Let's go! We got 3 clients implementing it.

make that 4. Nostrocket too, and also this service: https://nostrocket.org/products, so 5 depending on how you count.

fiatjaf:
Why late? No, the idea was that you would open the PR to `master`.

fiatjaf:
It's ridiculous that we're using GitHub PRs to define how we will get rid of GitHub PRs.

34.md
```diff
@@ -0,0 +1,102 @@
...
+    ["web", "<url for browsing>", ...], // a webpage url, if the git server being used provides such a thing
+    ["clone", "<url for git-cloning>", ...], // a url to be given to `git clone` so anyone can clone it
+    ["relays", "<relay-url>", ...] // relays that this repository will monitor for patches and issues
+  ]
+}
+```
+
+The tags `web`, `clone`, `relays` can have multiple values.
+
+Except `d`, all tags are optional.
```
ariard, line 32: i think you can recommend `d` to be a 80-bit / 128-bit random string.
that way minimize collision at the relay level
there is `name` for human-readable label.

34.md
```diff
@@ -0,0 +1,102 @@
...
+Patches can be sent by anyone to any repository. Patches to a specific repository SHOULD be sent to the relays specified in that repository's announcement event's `"relays"` tag. Patch events SHOULD include an `a` tag pointing to that repository's announcement address.
+
+```jsonc
+{
+  "kind": 1617,
+  "content": "<patch>", // contents of <git format-patch>
+  "tags": [
+    ["a", "30617:<base-repo-owner-pubkey>:<base-repo-id>"],
+    ["p", "<repository-owner>"],
+    ["p", "<other-user>"], // optionally send the patch to another user to bring it to their attention
```
ariard, line 45: can say a default limit (e.g `20`’s `other-user`).
relays can have local policy saying they allow more.
or price them with `zaps` or other micro-payments.
otherwise amplification attack vector for relay.

34.md
```diff
@@ -0,0 +1,102 @@
...
+    // for the first patch in a thread or series
+    ["t", "root"],
+
+    // optional tags for when it is desirable that the merged patch has a stable commit id
+    // these fields are necessary for ensuring that the commit resulting from applying a patch
+    // has the same id as it had in the proposer's machine -- all these tags can be omitted
+    // if the maintainer doesn't care about these things
+    ["commit", "<current-commit-id>"],
+    ["parent-commit", "<parent-commit-id>"],
+    ["commit-pgp-sig", "-----BEGIN PGP SIGNATURE-----..."], // empty string for unsigned commit
```
ariard, line 56: often on projects like linux, you can have commit signed by multiple authors / contributors.
so i think “commit-pgp-sig” + “committer” could be option duplicated.

34.md
```diff
@@ -0,0 +1,102 @@
...
+
+```jsonc
+{
+  "kind": 30617,
+  "content": "",
+  "tags": [
+    ["d", "<repo-id>"],
+    ["name", "<human-readable project name>"],
+    ["description", "brief human-readable project description>"],
+    ["web", "<url for browsing>", ...], // a webpage url, if the git server being used provides such a thing
```
ariard, line 23: i think this could document all you could use directly for `mailto` URI email address.
emails still widely used for project like linux.
also a fall-back if the list of original relay is perm down.

34.md
```diff
@@ -0,0 +1,102 @@
...
+
+Issues are Markdown text that is just human-readable conversational threads related to the repository: bug reports, feature requests, questions or comments of any kind. Like patches, these SHOULD be sent to the relays specified in that repository's announcement event's `"relays"` tag.
+
+```jsonc
+{
+  "kind": 1621,
+  "content": "<markdown text>",
+  "tags": [
+    ["a", "30617:<base-repo-owner-pubkey>:<base-repo-id>"],
+    ["p", "<repository-owner>"]
```
ariard, line 72: i think you could have a list of “p” - “other-user” in kind `1621` as generally you have many people working on large projects.

34.md
```diff
@@ -0,0 +1,102 @@
...
+
+```jsonc
+{
+  "kind": 1622,
+  "content": "<markdown text>",
+  "tags": [
+    ["a", "30617:<base-repo-owner-pubkey>:<base-repo-id>", "<relay-url>"],
+    ["e", "<issue-or-patch-id-hex>", "", "root"],
+
+    // other "e" and "p" tags should be applied here when necessary, following the threading rules of NIP-10
```
ariard, line 89: i think you could have “relay-local” threading policy on max default “p” elements.

34.md
```diff
@@ -0,0 +1,102 @@
...
+{
+  "kind": 30617,
+  "content": "",
+  "tags": [
+    ["d", "<repo-id>"],
+    ["name", "<human-readable project name>"],
+    ["description", "brief human-readable project description>"],
+    ["web", "<url for browsing>", ...], // a webpage url, if the git server being used provides such a thing
+    ["clone", "<url for git-cloning>", ...], // a url to be given to `git clone` so anyone can clone it
+    ["relays", "<relay-url>", ...] // relays that this repository will monitor for patches and issues
```
ariard, line 25: this is missing some `timestamp` / `expiry` notion.
you can have collaborative codes contributions spawning years on some project.
that ways can be used for policy migration if you wish to update `repo-id` / `relays` / `root`.

ariard (commented):
sounds good overall.

> I personally have given up on pgp. It definitely shines in key rotation,
> I give it that, but key rotation in pgp is a huge pain. Right now I
> can't really rotate my pgp keys on my yubikey because I have to go
> through 100 arcane gpg smartcard steps that I can't be bothered to
> figure out again. nostr has already replaced pgp as my web of trust
> model for most things.

pgp has a fundamental advantage over NIP-01 as an authenticated format message.
you have multiple signature scheme algorithm supported.
nostr has only one so far, schnorr bip340.

34.md
DanConwayDev: using human readable `d` identifiers follows a precedent established outside of nip34.
replaceable events are referenced with `<kind>:<pubkey>:<identifier>` so the collision problem is limited to repo events issued by the same pubkey.
The `r` tag with a `euc` marker (earliest unique commit) can also be used to identify the repository.

34.md
DanConwayDev: events that tag large numbers of pubkeys or events can be identified as spam either at the relay level or by clients.
I'm not sure there needs to be a default limit as other spam identification criteria can be used such as web of trust, reports, etc.

34.md
DanConwayDev: That's an interesting idea. git only supports a single pgp signature which in the git-over-email model is the committer rather than the author(s).
The main purpose of retaining the pgp signature is to retain the original commit id.
Including more than one pgp signature and committer would make it unclear which should be used to create the commit id.
What would be the value of including additional pgp signatures when git can't interpret them and all patches / responses are in signed nostr events?

34.md
DanConwayDev: Yes. The `p` tag can be used on any event to bring it to a users attention. The explicit inclusion of this only in the Patch section of the nip is potentially confusing.

34.md
DanConwayDev: all nostr events must include a `created_at` timestamp.
any nostr event with a kind in the 30000-40000 range is a 'parameterized replaceable event'. This means all metadata apart from the pubkey and `d` tag can be updated by issuing a new event with a larger timestamp.
a deletion event (nip09) would indicate the repository event has expired

PR #1004, opened 2024-01-23
#p refers to pubkeys, *not* "event pubkeys".

ibz:
"#p — a list of event pubkeys" - this sounds wrong. Events don't have pubkeys!


staab:
I think it actually does make sense, "event pubkeys" meaning "the value of the pubkey field on events". Maybe that could be clearer.

alexgleason (approved):
This wording is better.

In response to your comment, events _do_ have pubkeys. Both ways make logical sense, but the new wording is still better.

PR #1005, opened 2024-01-25
NIP-30: add supported kinds

AsaiToshiya:
Custom emoji is also used in `kind 7` (Reactions) and `kind 30315` (User Statuses).

### DIFF
--- a/30.md
+++ b/30.md
@@ -6,7 +6,7 @@ Custom Emoji
 
 `draft` `optional`
 
-Custom emoji may be added to **kind 0** and **kind 1** events by including one or more `"emoji"` tags, in the form:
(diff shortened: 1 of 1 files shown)


alexgleason (approved):

mikedilger:
There are a lot of event kinds where custom emoji makes no sense at all.... for just one example, in a NIP-46 NostrConnect event, or in a kind-10002 relay list.

Even though it seems tedious, we should list the kinds where it makes sense.  Adding the 2 kinds specified is okay by me.  Removing kind is not.

-1 as is.

AsaiToshiya:
@mikedilger You're right, so I added those two kinds instead of removing kinds.

PR #1006, opened 2024-01-26
NIP-XX: Benchmarking

mattn:


vitorpamplona:
I don't know what this NIP is doing. What benchmarking data does it have? And benchmarking of what? The relay? the Client? Which specific feature is it assessing the performance?

vitorpamplona:
Closing due to inactivity.

PR #1008, opened 2024-01-28
Fixes to NIP-46 based on my possibly incorrect understanding

mikedilger:
Implementing NIP-46, I ran into a number of things that were not clearly specified.  This is an attempt to clarify these.

In case I got these things wrong, please correct me.

@pablof7z

### DIFF
--- a/46.md
+++ b/46.md
@@ -30,13 +30,13 @@ The client generates a QR code in the following form (URL-encoded):
 nostrconnect://<client-key-hex>?relay=wss://...&metadata={"name":"...", "url": "...", "description": "..."}
 ```
 
-The signer scans the QR code and sends a `connect` message to the client in the specified relays.
+The signer scans the QR code and sends a `connect` message to the client in the specified relays using the request event format.
 
 ## Event payloads
 
 Event payloads are [NIP-04](04.md)-encrypted JSON blobs that look like JSONRPC messages (their format is specified inside the `.content` of the event formats below).
 
-Events sent by the client to the remote signer have the following format:
+Request events sent by the client to the remote signer have the following format:
 
 ```js
 {
@@ -50,7 +50,7 @@ Events sent by the client to the remote signer have the following format:
 }
 ```
 
-And the events the remote signer sends to the client have the following format:
+And reply events the remote signer sends to the client have the following format:
 
 ```js
   "pubkey": "<signer-key-hex>"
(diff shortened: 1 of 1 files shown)


46.md
```diff
@@ -30,13 +30,13 @@ The client generates a QR code in the following form (URL-encoded):
 nostrconnect://<client-key-hex>?relay=wss://...&metadata={"name":"...", "url": "...", "description": "..."}
 ```
 
-The signer scans the QR code and sends a `connect` message to the client in the specified relays.
+The signer scans the QR code and sends a `connect` message to the client in the specified relays using the request event format.
```
fiatjaf, line 33: what is the request event format?

fiatjaf: oh, I see

maybe we should rephrase this to say "RPC call" somehow.

mikedilger: Basically I'm just confused at how the signer can send a "connect" message to the client, when the signer's event format doesn't allow specifying a method.  It's a backwards command.  So I'm assuming it must use the other event format for this.  Given I don't actually know how this is supposed to be resolved I'm just not implementing this part of the spec for now.

mikedilger: Ok I'm implementing 2. I commented out my impl of flow 1 since I couldn't understand how to send a connect from the server to the client.  I was hoping someone did that and could explain what has actually happened in practice.

46.md
```diff
@@ -67,7 +67,7 @@ The signer key will always be the key of the user who controls the signer device
 ### Methods
 
 - **connect**
-  - params: [`pubkey`, `secret`]
+  - params: [`pubkey-hex`, `secret`]
```
monlovesmango, line 70: given that connect can be from either the signer or the client, do we know whether `pubkey-hex` changes in each scenario?

also any idea why we need `secrect`? assuming this is the client's private key..?

pablof7z: no, `secret` is a secret that can be provided out of band to the client from the signer; this is an easy way to identify the client in front of the signer.

Scenario:

A company provides authorization to 100 employees to sign using the conpany's key
They create 100 secrets and give a secret to each employee.
Each secret is redeemed without prior knowledge of which client-pubkey will do the redemption.

monlovesmango: where is this detailed?

is the secret for `connect` that is from client to signer only? or is it also for `connect` that is from signer to client as well?

I don't think i'm understanding, from the example you shared it seems like it is the employee's private key that can sign on behalf of the company's key. which sounds like the private key on the client side.

PR #1014, opened 2024-02-01
NIP-49: Fix correct PASSPHRASE to PASSWORD

ocknamo:
In NIP-49, the words PASSPHRASE and PASSWORD both existed to mean the same thing. 
After checking the scrypt documentation, password seemed to be correct.

FYI. https://nodejs.org/api/crypto.html#cryptoscryptpassword-salt-keylen-options-callback

So I unified the words into password.

### DIFF
--- a/49.md
+++ b/49.md
@@ -7,12 +7,12 @@ Private Key Encryption
 
 `draft` `optional`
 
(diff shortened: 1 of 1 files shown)


mikedilger:
Yeah that's fine.  The term passphrase was invented to nudge people towards choosing longer passwords.  It doesn't mean anything different.

Issue #1013, opened 2024-02-01
relax requirement of NIP-42 to include the relay


mleku:
the only things that are important about auth are challenge, pubkey and signature, the relay is not relevant

the requirement of the relay stops projects like bostr from being able to proxy your auth because there is no protocol with the client to accept a different relay address from an IP

the relay address can be fake anyway, it is not meaningful or anything related to security

by dropping the requirement of auth events to contain a relay tag, proxy relays become possible and this allows a relay to broadcast your events to your relay list - all automatically, and on the VPS running it, all that is required then is your pubkey as an access control whitelist, and a little more code in the proxy relay and you can fully proxy, and the proxy can also then do other things like caching your requests so it doesn't pester the other relays for them when it has them in cache

i am going to remove the restriction in my relay where it doesn't accept auth events with matching challenges because of the relay tag, or even the absence of the relay tag, only enforcing the challenge and the time window

the challenge is all of the security, and the relay tag just makes the auth protocol inflexible

vitorpamplona:
Ack. Nice.

Yonle:
Strongly disagree on this proposal.

This will give numerous of privacy issues (even in a small amount), for example there are some relays that is just another proxy of an nostr relay that all it does was forwarding any AUTH request from upstream to client in order to record who's the users are PMing to.

```
Victim <-> MITM Relay Proxy <-> Upstream
```

Of course the MITM relay proxy will not be able to read your PM content due to it being encrypted, However the MITM will know who are you talking to, which what some paid relays want to avoid to.

staab:
> we should require the relay address to be in the auth challenge

I agree, a client really needs to know who is asking for auth so they can decide whether to authenticate. Currently it's impossible for clients to know who they're authenticating with; adding the relay url to the challenge would allow clients to compare the challenge url with the connection url and decide what happens if they're different.

I'm currently working on a relay proxy as it happens, and the MITM attack is very real. If the proxy's connection with the destination relay is authenticated, they can hijack an authenticated session and exfiltrate whatever data they want, or publish unwanted events to the destination relay. This is true regardless of the format of the AUTH message. Proxies _must_ be trusted, which is hard if clients connect to whomever and authenticate freely. User-controlled white/black lists of relays is probably what's ultimately needed.

mikedilger:
[...]

PR #1016, opened 2024-02-01
Add support for proxying AUTH

staab:
Related to https://github.com/nostr-protocol/nips/issues/1013

I drafted this to see what it would look like. I sort of want relay proxies, but they may come at too high a cost. NACK unless someone changes my mind.

### DIFF
--- a/42.md
+++ b/42.md
@@ -23,7 +23,7 @@ A relay may want to require clients to authenticate to access restricted resourc
 This NIP defines a new message, `AUTH`, which relays CAN send when they support authentication and clients can send to relays when they want to authenticate. When sent by relays the message has the following form:
 
 ```json
-["AUTH", <challenge-string>]
+["AUTH", <challenge-string>, <relay-url>]
 ```
 
 And, when sent by clients, the following form:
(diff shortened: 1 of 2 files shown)


staab:
> Gotta figure out where to save the setting of what to proxy in each relay for the user.

That would be useful, but for now I was thinking the client would just make the decision.

Yonle:
The new scheme `["AUTH", "<id>", "<url>"]` is really a weak approach to handle this problem. The relay at this point asks you who are you connected to, not the relay tells you who you connected to.

This does not make sense as proxy could spoof the `<url>` regardless of the value no matter the value does not match to the current established `host`.

staab:
Yes, but the client would sign the spoofed value in the response, so the auth would only work for the specified relay, but the relay wouldn't have the challenge and would deny access.

staab:
Why the downvote? Am I wrong? If relays validate the url in the signed auth message, the relay url can't be spoofed by the proxy.

Yonle:
The fact that you allowed the spoofed value to be signed just make it sounded ridiculous.

staab:
That would be necessary for a proxy to operate. A client might legitimately want to do that if using a trusted proxy.

PR #1017, opened 2024-02-01
NIP-29: add to NIPs list, improve preamble phrasing

alexgleason:
This improves the intro of NIP-29 a little more, and adds a missing `m` (mime) value to the example event.

### DIFF
--- a/29.md
+++ b/29.md
@@ -6,7 +6,7 @@ Media Attachments
 
 Media attachments (images, videos, and other files) may be added to events by including a URL in the event content, along with a matching `imeta` tag.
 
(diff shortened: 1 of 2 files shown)


alexgleason:
Even though it stands for "image meta", I altered history and now it stands for "inline meta" for any media attachment.

staab (approved):
I love how you co-opted the imeta term to mean "inline" 😂

vitorpamplona:
It may be worth adding that `imeta` should not be added to NIP-04 DMs. 

[NIP-17](https://github.com/nostr-protocol/nips/pull/686) DMs can use `imeta`.

fiatjaf:
Wait, since when this is NIP-29? I didn't notice.
No, no, that's intolerable. I'll move this to 92.

staab:
😂 it was a complete accident I promise, I would never mess with The Best Nip Ever

PR #1022, opened 2024-02-05
Update 98.md

viapek:
Amended required tag 'u' and example event 'u' to 'url' to match linked implementation example.


fabianfabian:
👎

At this point 'u' is what is used in actual clients and servers and the linked implementation is wrong.

viapek:
OK :) Thanks.

PR #1023, opened 2024-02-05
NIP-46: replace npub1...#? notation with bunker://...

fiatjaf:
@pablof7z 
@erskingardner

### DIFF
--- a/46.md
+++ b/46.md
@@ -17,7 +17,7 @@ The client always starts by generating a random key which is used to communicate
 The remote signer generates a connection token in the form
 
 ```
-<npub1...>#<optional-secret>?relay=wss://...&relay=wss://...
+bunker://<hex-pubkey>?relay=wss://...&relay=wss://...&secret=<optional-secret>
 ```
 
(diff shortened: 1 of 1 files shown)


pablof7z (approved):

46.md
```diff
@@ -17,7 +17,7 @@ The client always starts by generating a random key which is used to communicate
 The remote signer generates a connection token in the form
 
 ```
-<npub1...>#<optional-secret>?relay=wss://...&relay=wss://...
+bunker://<hex-pubkey>?relay=wss://...&relay=wss://...&secret=<optional-secret>
```
erskingardner, line 20: Is `bunker` the right name? NIP-46 can be used by all sorts of remote signers, right? not all of them are nsecbunkers.

fiatjaf: It doesn't have to make sense, it's just what people are using. I like it. When I use it with [nak](https://github.com/fiatjaf/nak) I like to think it refers to `nak bunker`.

erskingardner: cool. definitely not a blocker. 🚢

erskingardner (approved):

Issue #292, opened 2023-02-23
Best way to propose adding `window.nostr.signSchnorr()`


dannydeezy:
Recently [Alby](https://github.com/getAlby/lightning-browser-extension)  added the function `window.nostr.signSchnorr` to their nostr key manager to enable us to turn a nostr key into a bitcoin wallet with [nosft](https://github.com/dannydeezy/nosft).

I'd like to formally propose `window.nostr.signSchnorr()` for other key managers - what's the best way to do this? Another NIP? an extension on an existing NIP?

chendatony31:
Same, adding `window.nostr.signEth` may turn it into an ethereum wallet. lol

dannydeezy:
> Same, adding `window.nostr.signEth` may turn it into an ethereum wallet. lol

hey now nobody is suggesting that :)

bloXmove:
signEth() would be awesome. Is anybody (willing to) work on that?

fiatjaf:
After thinking too much about this my conclusion is that it is an inversion of concerns (or whatever is the correct expression). This is a Bitcoin wallet functionality, not a Nostr functionality. Therefore it makes more sense for a Bitcoin wallet to implement importing a Nostr key and using it as a Bitcoin key than the opposite.

Alby is a Bitcoin wallet, so it may make sense there -- although I think it shouldn't live in the `nostr` namespace.

chendatony31:
> After thinking too much about this my conclusion is that it is an inversion of concerns (or whatever is the correct expression). This is a Bitcoin wallet functionality, not a Nostr functionality. Therefore it makes more sense for a Bitcoin wallet to implement importing a Nostr key and using it as a Bitcoin key than the opposite.
> 
> Alby is a Bitcoin wallet, so it may make sense there -- although I think it shouldn't live in the `nostr` namespace.

Agreed, should inject and use multi namespaces if support different protocols, such as `window.nostr`, `window.bitcoin`, `window.ethereum`, and so on. And that is what our wallet (TokenPocket) going to do.

bloXmove:
I agree that the container here is the wallet not nostr. 
Can you point me to the git of TokenPocket then?
Note: if one is to enter the Ethereum wallet space web3 and preferably WalletConnect connectivity is a must. As is the ability to hold at least ERC20 - preferably also ERC721. Otherwise there will be zero take up

chendatony31:
[...]

PR #1026, opened 2024-02-06
NIP-07 - Add signString

Egge21M:
In order to extend cross protocol interoperability, NIP-07 providers should be able to sign arbitrary strings and return the signature. This allows non-nostr-native apps to use nostr identities to do... things. For example Cashu web-wallets could use this to un-/lock eCash to nostr public keys. 

Alby has already implemented this as `signSchnorr', but it should be part of the specification.

There really isn't any overhead, as this is very similar to the signature part of signEvent, without the validation.

### DIFF
--- a/07.md
+++ b/07.md
@@ -20,6 +20,7 @@ Aside from these two basic above, the following functions can also be implemente
 async window.nostr.getRelays(): { [url: string]: {read: boolean, write: boolean} } // returns a basic map of relay urls to relay policies
 async window.nostr.nip04.encrypt(pubkey, plaintext): string // returns ciphertext and iv as specified in nip-04 (deprecated)
 async window.nostr.nip04.decrypt(pubkey, ciphertext): string // takes ciphertext and iv as specified in nip-04 (deprecated)
+async window.nostr.signString(string): string // returns a signature on an arbitrary string
 ```
 
 ### Implementation


callebtc:
ACK [2dad152](https://github.com/nostr-protocol/nips/pull/1026/commits/2dad152a67d29aeb05c0ce58fc9012ca08998a39)

Nostr npubs are becoming a de factor PGP replacement for many of us. I would like to be able to sign anything with the extension that already has my keys, including arbitrary data.

callebtc (approved):

EthnTuttle:
Would adding an indication as to what signature is being used be helpful? using a second argument can specify the signature type. 
`async window.nostr.signString(message: string, signatureType: string): string // returns a signature on an arbitrary string`

alexgleason:
Check the discussion in https://github.com/nostr-protocol/nips/issues/292

fiatjaf:
This is completely out-of-scope for Nostr.

Egge21M:
> It's super weird that Alby added it under the nostr namespace, it should have been added under some other generic namespace or window.bitcoin, I don't know.

I really don't think it is. It's a pretty standard pattern for a method to be defined in the same scope as its state / underlying data lives in. 

If the nostr object holds the respective private key, anything that accesses this key should live in the same space (at least as long as we do not have a getKey method, which would obviously defeat the purpose)

Egge21M:
Continuing this line of argumentation and adding an example. In https://github.com/nostr-protocol/nips/issues/292#issuecomment-1445502593 you said 
> it is an inversion of concerns

Meanwhile I believe quite the opposite. Actually the only way to not violate the Single responsibility principle is to implement it in the `nostr` namespace.

If you have the following structure
```go
type nostr struct { 
    nostrKey [32]byte
}
type bitcoin struct {
    bitcoinKey [32]byte
}
```
A method that accesses `nostrKey` should clearly be defined like this
```go
func (n nostr) SignString(s string) string {
    return createSignature(s, n.nostrKey);
}
```

starbackr-dev:
This is a huge unlock for using Nostr keys for other apps like cashu. Many key extensions such as Metamask has this and apps use it to validate the user locally. @fiatjaf pls revisit this again.

alexgleason:
@starbackr-dev Where does Metamask have this? I only see [`personal_sign`](https://docs.metamask.io/wallet/reference/personal_sign/) which prefixes the data with `0x` before signing for security reasons, but making it useless for applications like Nostr.

fiatjaf:
I have no power to force implementations to support this. I'm fine if they do, and you could probably reach out to extension authors and ask them to implement. Having this written in NIP is not a requirement for that.

I'm not saying should never write this on the NIP, just that I'm still very far from being convinced that this is somehow _related to Nostr_ so I wouldn't like that, but this is just my opinion and I'm open to changing my mind.

starbackr-dev:
> @starbackr-dev Where does Metamask have this? I only see [`personal_sign`](https://docs.metamask.io/wallet/reference/personal_sign/) which prefixes the data with `0x` before signing for security reasons, but making it useless for applications like Nostr.

I vaguely remembered this. but googling found this article where it specifically sign a one-time code to confirm the user has the private key. https://dzone.com/articles/spring-authentication-with-metamask

alexgleason:
> > @starbackr-dev Where does Metamask have this? I only see [`personal_sign`](https://docs.metamask.io/wallet/reference/personal_sign/) which prefixes the data with `0x` before signing for security reasons, but making it useless for applications like Nostr.
> 
> I vaguely remembered this. but googling found this article where it specifically sign a one-time code to confirm the user has the private key. https://dzone.com/articles/spring-authentication-with-metamask

If you just need to verify that a user owns a private key, `signEvent` can already do that. You could use it to sign a [NIP-98](https://github.com/nostr-protocol/nips/blob/master/98.md) event, for example. I have already implemented that in my projects.

Egge21M:
As soon as someone wants to build anything that locks anything to a nostr key, they will require this method.

Right now there are already two implementations of cross-protocol projects that I am aware of depending on it.

Deezy building a nostr ordinals marketplace and Cashu-address allowing users to receive Cashu on their npubs.

The amount of possible usecases is endless.

Why should the web be a second citizen in this because they, in contrary to native clients with key access, can not sign Arbitrary data?

alexgleason:
> As soon as someone wants to build anything that locks anything to a nostr key, they will require this method.
> 
> Right now there are already two implementations of cross-protocol projects that I am aware of depending on it.
> 
> Deezy building a nostr ordinals marketplace and Cashu-address allowing users to receive Cashu on their npubs.
> 
> The amount of possible usecases is endless.
> 
> Why should the web be a second citizen in this because they, in contrary to native clients with key access, can not sign Arbitrary data?

You didn't acknowledge how NIP-98 can solve this problem. Did you look at it?

If you're building something not over HTTP, you just need a new event type following a similar idea as NIP-98, but specific to your use-case.

alexgleason:
I mean, if you're ultimately just making them sign a string saying `My name is <name>, and I hereby attest I am accessing <resource> at <time>`, you could actually just put that into the `content` field of a kind 1 note and have them sign it. I'm not saying you should to that (you _should_ create a new event kind), I'm just trying to make a point about the semantics of this problem.

Egge21M:
> You didn't acknowledge how NIP-98 can solve this problem. Did you look at it?

Yes. I use it in most of my projects already. Cashu-Address authorization is purely NIP-98.

However Cashu Tokens for example are extremely sensitive to size. Including a serialised event in it (of which 90% is unnecessary overhead) is infeasible and will probably break the usecase.

Egge21M:
> If the intention was to make Cashu compatible with Nostr that would have been the most appropriate way and wouldn't require chasing extension developers to add a new potentially dangerous generic signing method.

It would be an optional method, no? This is not about chasing developers, but rather to establish a standard for anyone looking to implement it.

callebtc:
> If I'm not mistaken I told @callebtc it would be great if Cashu's P2PK scheme actually used a signed Nostr event in its authorization flow, instead of a custom message, but he didn't like it apparently and have no idea why.
> 
> If the intention was to make Cashu compatible with Nostr that would have been the most appropriate way and wouldn't require chasing extension developers to add a new potentially dangerous generic signing method.

We need Schnorr signatures from the private keys of the user. It is not related to the Nostr protocol or to nostr events etc.

It would have been absurd to monkey-patch the core of the Cashu protocol just to accommodate Nostr events inside ecash notes.

There is merit in allowing signatures from my nsec, precisely for other applications like Cashu (not nostr applications but other web applications) to be able to use the user's keys for other things. I don't see any reason why this simple but powerful option shouldn't be available. It works with Alby, so I guess that's not too bad.

As a user, I'd say: Let me sign what I want, I want total control over my own nsec. Maybe I just want to sign an email. Why not let me?

callebtc:
> If I'm not mistaken I told @callebtc it would be great if Cashu's P2PK scheme actually used a signed Nostr event in its authorization flow, instead of a custom message, but he didn't like it apparently and have no idea why.
> 
> If the intention was to make Cashu compatible with Nostr that would have been the most appropriate way and wouldn't require chasing extension developers to add a new potentially dangerous generic signing method.

The only thing we need here are Schnorr signatures from the private keys of the user. This is not directly related to the Nostr protocol or to Nostr events. Your suggestion back in the past disregarded any larger projects we were working on and it would have required us to monkey-patch the core of the Cashu protocol with an ugly special case without gaining anything really, so it wasn't a viable option.

In general, I see a lot of merit in allowing signatures from my nsec, precisely for other applications like Cashu (not nostr applications but other web applications) to be able to use the user's keys for other things. I don't see any reason why this simple but powerful option shouldn't be available. It works with Alby, so I guess the situation isn't too bad.

As a user, I'd say: Let me sign what I want, I want total control over my own nsec. Maybe I just want to sign an email. Why not let me? If we want nostr identities to be used in more places, we should consider this.

alexgleason:
There are many reasons not to do it... having multiple ways to do things leads to protocol bloat. In the future you describe where everybody is using `signString` to do all kinds of stuff, including send money, it could lead to the exact scenario Trezor warns against:

> It's not possible to sign an _arbitrary_ message with any sort of signature scheme by Trezor.
> 
> It would be really stupid to allow this: if the message is arbitrary, you can stuff in, say, a valid Bitcoin transaction. Then it's a matter of crafting a clever malware, telling the user: "Security check: please confirm the following characters on your Trezor screen to validate your wallet", and stealing their money.

[Source](https://www.reddit.com/r/TREZOR/comments/vrftwn/schnorr_signature_api/)

If you can `signString` why even have `signEvent` at all? Some extensions won't implement `signString`, leading to mixed experiences. But it would open the door to clients using `signString` directly, and forcing all extensions to implement it if they want compatibility. ie protocol bloat.

> However Cashu Tokens for example are extremely sensitive to size. Including a serialised event in it (of which 90% is unnecessary overhead) is infeasible and will probably break the usecase.

Acknowledged... but then I agree with fiatjaf it doesn't belong in the `window.nostr` namespace. Alby should have put it at `window.alby`. And if you're dealing with money you should probably have a `window.cashu` so the experience in the extension can be tailored to signing a money transaction.

fiatjaf:
> Your suggestion back in the past disregarded any larger projects we were working on and it would have required us to monkey-patch the core of the Cashu protocol with an ugly special case without gaining anything really, so it wasn't a viable option.

I'm not sure about that. I can't find the conversation, but from what I remember you didn't have it working yet and was making up a protocol for signing _something_ on the fly. That _something_ could have pretty much been a Nostr event. And you would have gained instant compatibility with all the Nostr extensions, bunkers, the Android Nostr event signing apps, and an easy way to get integrated into Nostr clients.

> As a user, I'd say: Let me sign what I want, I want total control over my own nsec. Maybe I just want to sign an email. Why not let me? If we want nostr identities to be used in more places, we should consider this.

You can have total control of your nsec, this just has nothing to do with a NIP.

fiatjaf:
> However Cashu Tokens for example are extremely sensitive to size. Including a serialised event in it (of which 90% is unnecessary overhead) is infeasible and will probably break the usecase.

You wouldn't have to embed the entire event in the cashu token. In fact you wouldn't have to embed anything except for the signature and the pubkey. Any software reading these tokens would know how to rebuild the serialized Nostr event by just concatenating strings and then check the signature on that.

> This is not about chasing developers, but rather to establish a standard for anyone looking to implement it.

Sure, but using a Nostr event wouldn't have precluded that at all. It would just require some concatenation of hardcoded strings. But I get it, it would be "ugly".

fiatjaf:
Anyway, just listen to what @alexgleason is saying.

Egge21M:
> Acknowledged... but then I agree with fiatjaf it doesn't belong in the `window.nostr` namespace. Alby should have put it at `window.alby`. And if you're dealing with money you should probably have a `window.cashu` so the experience in the extension can be tailored to signing a money transaction.

Please look at https://github.com/nostr-protocol/nips/pull/1026#issuecomment-1931545292

Implementing this anywhere else than window.nostr would be a clear violation of basic OOP principles. It would also lead to horrifying UX and DX, when it comes to extensions that manage a multitude of keys (like Alby for example). How would I choose wether to sign with a Bitcoin key or my nostr key, if signSchnorr would live inside window.alby?

For me it's really not about what is being signed, but instead which application state is being used for signing and where it belongs to.

starbackr-dev:
> Anyway, just listen to what @alexgleason is saying.

Good point. Can be used by hackers to sign a hidden event.

neilck:
I wrote the AKA Profiles Signer Extension (up to whopping 96 users). Arbitrary string signing would grant developers a lot of flexibility, but without a defined event with a Kind, it becomes very hard for an extension to communicate to users what they are signing, before they agree to sign.

callebtc:
> it becomes very hard for an extension to communicate to users what they are signing, before they agree to sign.

I get that. But how do you make sure that a rouge website isn't publishing fake posts from your nsec? Isn't this the same problem? Why would it be it more problematic that a third-party app can produce a signature on arbitrary data than producing signed nostr notes that you never wanted to sign? If the problem is about verifying the message before signing, then nobody seems to really care about this issue with regard of the hundreds of apps that are using NIP-07 today. @fiatjaf?

fiatjaf:
My extension displays to the user what they're signing and asks for confirmation beforehand.

TonyGiorgio:
> My extension displays to the user what they're signing and asks for confirmation beforehand.

Which every single sane person clicks always allow because every nostr client goes insane with their asks to even load the site. 

Concept ACK on the sign method. If you're going to base key identities off of the same crypto as Bitcoin, then it's a nostr functionality to do more advanced things with those keys. Not solely a "push it to Bitcoin" problem.

Egge21M:
It seems like most here are worried that users might "blindly" sign stuff, because a might not be as self explanatory as an event (especially if it is a digest).
Maybe there is a middle ground here. I think for most use cases (at least the ones that I have observed) the signed data doesn't have to be completely arbitrary. Most of them require a signature on a digest. So what do you think about this:

```js
signHash: async (digest, message, description?) => string
```

a) A provider can check wether the to-be-signed digest is actually the result of sha256(message).
b) They can display the message instead of the digest to the user, which might be more meaningful
c) A readable description can be added, which is then displayed ADDITIONALLY to the message. (Obviously does not solve trust issues).

neilck:
I would like to know more about the use cases you are thinking of, to continue the discussion.

PR #1027, opened 2024-02-06
NIP-114: ids_only filter

mmalmi:
When a filter has `ids_only: true`, the relay must return only the ids of the events that match the filter.

This allows a client to request the full event only when it is not already stored locally or received from another relay. It can save a lot of bandwidth, especially when connecting to multiple relays or subscribing to a lot of data.

### DIFF
--- a/114.md
+++ b/114.md
@@ -1,0 +1,27 @@
+NIP-114
+=======
+
+ids_only filter
+---------------
+
+### Description
+
+When a filter has `ids_only: true`, the relay must return only the ids of the events that match the filter.
+
+The ids are returned as `["HAVE", subscriptionID, eventID]` messages.
+
+Clients may request the full events by id with a `["GET", eventId]` message. Relays reply to `GET` requests with `["EVENT", "*", eventJson]` messages, where `*` is a placeholder subscription id.
+
+Clients may also request the full events by id with a normal `["REQ", subscriptionId, {ids}]` message.
+
+### Motivation
+
+`ids_only` filter allows a client to request the full event only when it is not already stored locally or received from another relay. This can save a lot of bandwidth, especially when connecting to multiple relays or subscribing to a lot of data.
+
+### Implementations
+
+- [@mmalmi/strfry](https://github.com/mmalmi/strfry) ([Pull request](https://github.com/hoytech/strfry/pull/99))
+- [@mmalmi/nostr-rs-relay](https://github.com/mmalmi/nostr-rs-relay) ([Pull request](https://github.com/scsibug/nostr-rs-relay/pull/186))
+- [Snort & Iris](https://git.v0l.io/Kieran/snort).
+  - [HAVE message handling](https://git.v0l.io/Kieran/snort/src/commit/edbfa02c5202c5a7035e146d446e7bf4bb3feb84/packages/system/src/connection-pool.ts#L74)
+  - [Adding ids_only filter](https://git.v0l.io/Kieran/snort/src/commit/edbfa02c5202c5a7035e146d446e7bf4bb3feb84/packages/system/src/query.ts#L397) for relays that support nip 114 


fiatjaf:
It's probably better to use a new verb for this instead of `REQ`, maybe just `HAVE` as in the response.

And instead of using `GET` why not `REQ` with `{ids:[]}`?

mmalmi:
Unlike `REQ`, `GET` doesn't create a subscription. In this case, we don't want subscriptions, because they would need to be separately closed and might exceed the relay's subscription limit.

Maybe the format should be `["GET", filters]` instead of `["GET", eventId]` so it could also be used in other requests where a subscription is not wanted.

Also `HAVE` messages could return an array of matching ids instead of just one id. Depends on the DB & buffering strategy how many ids the relay wants to put in one `HAVE` response.

If we want a new verb instead of using `REQ` + filter.ids_only, I think it should be something other than a reuse of `HAVE` because it needs to be handled differently. Maybe `SUB`?

monlovesmango:
>Unlike REQ, GET doesn't create a subscription. In this case, we don't want subscriptions

saying its not a subscription is just a superficial statement from the perspective of the client. from the relay side it takes the same amount of resources and I would think relay operators would definitely want to count this as another `REQ` subscription.

besides, using EOSE you can effectively get the same behavior you are describing imo. if relay doesn't support `EOSE` by now it probably won't support `GET` anytime soon either.

I like the ids_only in the filter but not the `GET` that is duplicative of `REQ`

I lean towards not having `HAVE` return an array, but might be good to use in conjunction with `EOSE`?

I don't like `SUB` for new verb if you want that for ids_only filter (since `REQ` is referred to as a sub a lot). maybe `GOT`?

vitorpamplona:
> One honest question: why not use negentropy?

Too complicated to code. There are no libs and these alternative sync options resolve most of the issue with 20 min implementations from scratch. 

If we want force Negentropy to be a mandatory part of Nostr, we lose most of the simplicity benefits of Nostr.

fiatjaf:
I don't think we should force negentropy to be mandatory, but this shouldn't be mandatory either. And if negentropy does the same that is achieved here more efficiently it is better to have only nip01 + negentropy instead of nip01 + ids_only + negentropy as thing to code.

By agree it is complicated. I was implementing it and stopped to give my brain some time to breath.

vitorpamplona:
> but this shouldn't be mandatory either

Maybe not this, but some type of easy SYNC should be mandatory in the long run. 

Many apps now are observing the benefits of a local cache with persistent databases. All those databases need to sync from time to time to make sure the user is not missing anything. Downloading everything, all the time is just not a scaleable solution to sync. 

A simple (30min implementation) sync solution does improve Nostr's decentralization while reducing the data bandwidth of the protocol. And both clients and relays waste less resources in doing so. 

I sync (re-download everything, from all ~3000 relays) ~2GB worth of events every other day for my user. I would prefer not doing that to relays, but I have to. The current state of sync is just dumb.

fiatjaf:
> I sync (re-download everything, from all ~3000 relays) ~2GB worth of events every other day for my user. I would prefer not doing that to relays, but I have to. The current state of sync is just dumb.

I'd guess most people aren't doing that, and isn't a requirement for having Nostr work, but I agree that when you need to do it it is dumb. If we're thinking about a good long term solution then I think negentropy looks better than this proposal though. Also even though it is hard to code it isn't _that_ hard and seems to be worth it for those strict use cases that really need it.

I would be interested in something that is better than negentropy if there is anything. I've played with https://github.com/sipa/minisketch and I was shocked about how well it worked -- but it has a lot of drawbacks in comparison with negentropy that make it unpractical. Also it is completely impossible to understand or implement although it is in C so it would be possible to use almost everywhere.

I guess we could be fine with a negentropy implementation in C and wrappers.

vitorpamplona:
I tried to implement last year's (it changed) Negentropy from scratch and gave up. I am not sure if a negentropy lib that "automatically" works with all database types will ever exist. The amount of work on just mapping the database structure and the relay messages to the needs of the library alone is a time-consuming process and it does require you to have a good clue of what the library is doing to assemble the right sets. I am not sure its going to be as simple as just adding it, calling a function, and boom, synced. 

However, If you are looking for the most efficient way and are willing to pay the price of protocol simplicity for it, negentropy wins. I just don't think Nostr is the place for that simplicity sacrifice. At least not on the mandatory part of Nostr. 

After testing a bunch of stuff, my recommendation was to do https://github.com/nostr-protocol/nips/pull/826 

Nostr.wine implemented it in 30 minutes and I did the same on Amethyst in about 1hr. It's not the most efficient way to solve Sync, but it is pretty darn good for the simplicity of the proposal.

mmalmi:
Negentropy or similar is good for syncing existing datasets, but this NIP is also useful if you have open subscriptions to many relays that might send the same new events many times over. That might not be a huge amount of data on a light client, but especially on relay-to-relay communications it can save bandwidth.

mmalmi:
> > Unlike REQ, GET doesn't create a subscription. In this case, we don't want subscriptions
> 
> saying its not a subscription is just a superficial statement from the perspective of the client. from the relay side it takes the same amount of resources and I would think relay operators would definitely want to count this as another `REQ` subscription.

It's not the same. Subscriptions persistently occupy memory on the relay and need to be checked against every time a new event is received. If `GET` is only for event ids, the lookup is usually very fast.
 
> besides, using EOSE you can effectively get the same behavior you are describing imo. if relay doesn't support `EOSE` by now it probably won't support `GET` anytime soon either.

Client side logic is a lot simpler and data is synced faster if there's `GET` that doesn't create a subscription. You can just ask right away for the not-haves. Otherwise, you have to buffer them and send them as a batch after some arbitrary timeout or debounce, which slows down the sync. Then you need to buffer again until the relay responds with `EOSE`, and as a fallback for not receiving `EOSE`, use a timeout.

monlovesmango:
I see, yeah I guess being able to make single `GET` is more nimble and lightweight.

>It's not the same. Subscriptions persistently occupy memory on the relay and need to be checked against every time a new event is received. If GET is only for event ids, the lookup is usually very fast.

even if the lookup is very fast, lookup for multiple `GET`s is still slower than a single `REQ` with only `id` array in the filter. just because it is closed right away shouldn't mean it is exempt from having a subscription ID. also not having a subscription ID would mean relays would have to rely solely on rate limiting to limit excessive `GET`s and there would be no way to communicate relay errors to the client with any specific `GET`.

I do agree that subscriptions persist on the relay which cause overhead, and it would be nice for a relay to know that "once I am done with this look up I can close this subscription right away".

>You can just ask right away for the not-haves. Otherwise, you have to buffer them and send them as a batch after some arbitrary timeout or debounce, which slows down the sync.

agree

>Then you need to buffer again until the relay responds with EOSE, and as a fallback for not receiving EOSE, use a timeout.

disagree, you can handle the events however you want the moment the `EVENT` is received from the relay just like if multiple `GET`s were sent if thats what you want to do. you would only need to wait for EOSE or timeout to closeout the subscription.

overall, it seems to me the main goal of the `GET` is to be able to make a `REQ` (for a specific event ID) which will close automatically upon completion. I think this is useful. however, I don't agree that just bc its only querying a single ID it should be exempt from relay max subscription limits and not have subscription ID. clients still need to be limited in how many queries they can send to the relay at once and relays still need to be able to communicate errors.

all this to say, I think `GET` should be more of a `REQ` + `CLOSE` message (and might be less ambiguous to name something like `REQCLOSE`). this way it will:

- give you the behavior you want for requesting a single ID with little overhead
- has the added benefit of allowing the auto close behavior for any filter (not just a single ID), which I think actually adds a lot of convenience to general querying (most `REQ`s probably don't need to stay open)
- preserves the subscription ID model (which imo `GET` shouldn't be exempted from just because it has little overhead)

vitorpamplona:
This proposal works for sporadic SYNCs involving thousands of events. Each `HAVE` is about 100 chars, including the WebSocket protocol overhead bytes. Which leads to 1MB transfer for a 10,000 event sync. Given that users on average are connected to 10 relays, that's 10MB data transfer costs for a single SYNC event with all relays. 

I was hoping to SYNC Amethyst's local database (~50,000 to 100,000 events) every time the user brings the app back to the foreground. That's a 5MB to 10MB download 10-20 times a day for each relay. A daily sync cost of 500MB - 2GB for all relays. 

So, I would still prefer https://github.com/nostr-protocol/nips/pull/826

Sidepoint: Amethyst does a TON (dozens of times a second) of `GET`s (`REQ` + `CLOSE`). If a subscription is indeed heavier to support, then we can start using GET directly independent of this `ids_only` proposal.

hoytech:
I was just reviewing @mmalmi 's [pull request to strfry](https://github.com/hoytech/strfry/pull/99) for this feature and have some thoughts:

* Yes, it might be useful to be able to stream just the `id`s of a query instead of full events. If you are subscribed to 10 relays and they usually have all events posted to each-other then you will be downloading/parsing each event 10 times.
  * For synchronising stored events upon connection, in my (obviously biased) opinion, negentropy is superior. However, negentropy does not help for reducing the overhead of multiple open subscriptions, as explained by @mmalmi above.
* I don't like `GET` and `HAVE`
  * They seem incidental to this use-case since we already can fetch events by `id` with regular `REQ`, as mentioned by @fiatjaf .
  * Not having a subscription ID means that clients can not reliably multiplex multiple independent tasks over a single websocket connection (which tasks should a `HAVE` be routed to?).
  * Without a subscription ID, how do you indicate that the relay does *not* have the `id` you requested? Normally you would know because you'd get an `EOSE` with no preceding event. Without subscription ID+`EOSE` you're in the bad state of original nostr protocol where you just had to wait "long enough".
* `ids_only` is not really a property of a filter, but instead the request. If a `REQ` had multiple filters and some had `ids_only` and others didn't, relays would have to track which filter an event came from in order to send the full event or just the `id` (and what if it matched multiple?)

It seems to me what is needed is a way to pass flags/options to the `REQ` command. Something like this:

```json
[
    "REQ+",
    "my-sub",
    {
        "auto_close": true,
        "ids_only": true
    },
    <filter1>,
    <filter2>,
    ...
]
```

monlovesmango:
really like @hoytech 's suggestion.

bezysoftware:
> really like @hoytech 's suggestion.

So do I. I like the 'REQ+' verb to make it simpler to maintain backward compatibility.

vitorpamplona:
> I don't like GET and HAVE

What would be the reply of the "REQ+" if you don't like `HAVE`? I suppose we need something different than `EVENT` otherwise parsers not expecting partial events with just the ID might crash out there.

hoytech:
I think it would be OK to just return the ID in the `EVENT` reply instead of the full JSON: Only clients that specifically requested `ids_only` would get the raw IDs.

vitorpamplona:
> I think it would be OK to just return the ID in the EVENT reply instead of the full JSON: Only clients that specifically requested ids_only would get the raw IDs.

Not a fan of having to use an external variable to facilitate parsing: clients either have to keep track of the sub's only_ids status or do a polymorphic parser without having the subtype specified in the message itself. 

Not a big deal by itself but these things make protocols more confusing over time. I think I would prefer a more clear distinction between the two modes in the message itself.

monlovesmango:
> What would be the reply of the "REQ+" if you don't like `HAVE`? I suppose we need something different than `EVENT` otherwise parsers not expecting partial events with just the ID might crash out there.

I do think a new relay to client message type for the id_only results like `HAVE` is also a good idea (not a fan sending IDs in the `EVENT` reply either). I would probably prefer `IDS` for the message type as that is more consistent with `EVENT`.

the relay can just send either `EVENT` or `HAVE` message type depending on what the parameters of the `REQ+` are.

vitorpamplona:
If we are going to do `REQ+`, can we put custom projections in the spec? 

This might be too much flexibility, but it would be very useful here. 

```json
    {
        "projection": ["id", "kind", "pubkey", "created_at"]
        "auto_close": true,
    },
```

We could get all ids, etc and progressively sync per kind, date, and author range without having to create multiple REQs in sequence to know which ids have which kinds, etc.

hoytech:
@monlovesmango - Ooooh yes, great point about `limit`: This is really something that should be at the `REQ` level, not per-filter. And yeah, `COUNT` too, I think (ie you could imagine composing `COUNT` with `auto_close`).

Although it would've been great to have a way to extend `REQ`, I'm still not actually convinced `REQ+` is a good idea. It might add more complexity than benefit. (?)

bezysoftware:
There were already multiple discussions about limit e, e.g. https://github.com/nostr-protocol/nips/issues/1368

@vitorpamplona and @staab have use cases for using a limit per filter

PR #1029, opened 2024-02-08
Adds Native Private Events

vitorpamplona:
This is a crazy new idea to add fully private events to Nostr in a backwards-compatible way where current relays won't even be able to verify private events. 

I did this first version as a change to NIP-42, but I am happy to create a new NIP if it makes more sense.

### DIFF
--- a/42.md
+++ b/42.md
@@ -6,7 +6,7 @@ Authentication of clients to relays
 
 `draft` `optional`
 
-This NIP defines a way for clients to authenticate to relays by signing an ephemeral event.
+This NIP defines a way for clients to authenticate to relays by signing an ephemeral event and a way to specify which events and private and MUST require authentication to be served.
 
 ## Motivation
 
@@ -17,6 +17,40 @@ A relay may want to require clients to authenticate to access restricted resourc
   - A relay may limit subscriptions of any kind to paying users or users whitelisted through any other means, and require authentication.
 
 ## Definitions
+
+### New field to identify private events
+
+Clients that want to force an authentication in any event can add a `requires_auth_by` with array of authorized pubkeys to the base event object. 
+
+```jsonc
+{
+  "id": <32-bytes lowercase hex-encoded sha256 of the serialized event data>,
+  "pubkey": <32-bytes lowercase hex-encoded public key of the event creator>,
+  "requires_auth_by" [<32-bytes lowercase hex-encoded public key>, ... ]
+  "created_at": <unix timestamp in seconds>,
+  "kind": <integer between 0 and 65535>,
+  "tags": [
+    [<arbitrary string>...],
+    // ...
+  ],
+  "content": <arbitrary string>,
+  "sig": <64-bytes lowercase hex of the signature of the sha256 hash of the serialized event data, which is the same as the "id" field>
+}
+```
+
+NIP-42 supporting relays must then include the `requires_auth_by` field as the 7th position in the serialization of the event. This guarantees that relays and clients that do not support this NIP cannot validate private events and will never be able to receive them, avoiding leaking the event structure to a public relay. 
+
+```js
+[
+  0,
+  <pubkey, as a lowercase hex string>,
+  <created_at, as a number>,
+  <kind, as a number>,
+  <tags, as an array of arrays of non-null strings>,
+  <content, as a string>,
+  <requires_auth_by, as an array of non-null strings> 
+]
+```
 
 ### New client-relay protocol messages
 
--- a/42.md
+++ b/42.md
@@ -22,6 +22,8 @@ A relay may want to require clients to authenticate to access restricted resourc
 
 Clients that want to force an authentication in any event can add a `requires_auth_by` with array of authorized pubkeys to the base event object. 
 
+Relays MUST authenticate connections with at least one of the pubkeys listed in `requires_auth_by` to return events.
+
 ```jsonc
 {
   "id": <32-bytes lowercase hex-encoded sha256 of the serialized event data>,


staab:
It's an interesting idea, I had thought of something similar where you publish events to a particular relay and include the relay's url in the hash/signature, so that that relay can validate the event, but if it's published elsewhere the signature would appear invalid. But it would be easy to attack by just validating the signature for all known relays, or the relay the event was obtained from. AUTH is better than that.

The thing I'm not sure about with this PR is that it relies on implementations to enforce it. Anyone who knows how to support this nip can validate the signature, it really doesn't change anything about how public the event is. You might as well not include it in the signature and add a `requires_auth_by` field, except for the compatibility reasons you mentioned. So I guess unless this solves something really specific, I don't see a reason to add it.

fiatjaf:
At first I thought this would work, even though it is a dirty hack, for practical reasons.

But now I'm thinking that since this requires new implementation effort to support it is not very different from just making a new kind and a relay that requires AUTH: in both cases it will be possible to propagate the private event to new relays, as long as someone writes software that understands the new scheme.

I guess this does at least protect existing relays from relaying private events unwillingly -- but only if we assume they are good citizens and don't want to implement the new scheme in a way that makes all private events public.

vitorpamplona:
@staab I think you were trying to solve for something different. 

This one just creates an additional layer of privacy. It doesn't block events from being re-broadcasted away from a specific relay. It's designed to not affect the decentralization of the Nostr events, while minimizing the chance that the event is seeing in public relays. 

The ideal use case is for GiftWraps, for instance. You can create wraps that only certain people can download. The encryption is what makes it private. 

After all, bad actor relays don't need to verify anything to re-broadcast stuff.

fiatjaf:
So here's another solution that is similar to this but doesn't require a new signature scheme:

We define a new kind range 40000-49999 for events that must always require `AUTH` to be published -- and the authed public key must be the same as the event author pubkey.

Relays should either reject events in that kind or reply to `EVENT` messages for events of that kind with an `["OK", <id>, false, "auth-required: ..."]` and then only accept the event after authentication from the published.

---

This wouldn't work tomorrow, as most relays would still keep accepting all events, but it would be an easy add to codebases (they could just default to rejecting) and most well-behaved relays would probably abide.

vitorpamplona:
I thought about adding a new kind range 40,000 to 50,000 for private events. But then I realized that all event kinds would just be duplicated in this new range. So, I think a new root field makes more sense. 

> and the authed public key must be the same as the event author pubkey.

I need more flexibility than that. The list of pubkeys allowed to download the event might not be related to the pubkey.

fiatjaf:
> I need more flexibility than that. The list of pubkeys allowed to download the event might not be related to the pubkey.

What is your specific use case?

fiatjaf:
> The list of pubkeys allowed to download the event might not be related to the pubkey.

My point is that access-control must not be tied directly to the event, it must be enforced by the relays. A relay should know, through any other means, that your events of a certain kind should be downloaded by a certain list of people.

The point is that these people cannot download these events and republish them to other rogue relays that won't enforce your rules: you must be the one deciding where to publish and do that wisely.

vitorpamplona:
> What is your specific use case?

Gift Wraps with medical information in them and Relationship Status events that will list the patients of a given doctor. Both are private, but I want to further minimize their presence in public relays.

vitorpamplona:
> Then what about a new tag, just like we already have expiration we can have ["segregation"] or something and then relays apply the AUTH rule to events with that.

Then is the tag added twice in the new serialization scheme?

fiatjaf:
> Gift Wraps with medical information (required_auth_by will be the p-tag) in them and Relationship Status events that will list the patients of a given doctor. Both are private/encrypted, but I want to further minimize their presence in public relays.

I see. Well, this use case is served by my approach of relying on some specific relays.

But my use cases of private groups or private feeds is not covered by `required_auth_by` since the membership in the groups can change.

fiatjaf:
> Then is the tag added twice in the new serialization scheme? I am fine with it. I just thought it would be a little weird/confusing.

No, I'm talking about not having a new serialization scheme.

vitorpamplona:
> Well, this use case is served by my approach of relying on some specific relays.

Yeah, I'd like to avoid that centralization point if we can. 

> But my use cases of private groups or private feeds is not covered by required_auth_by since the membership in the groups can change.

Agree. I don't think it matters for your use case because your auth is relay-specific and can be hardcoded into it. In your case, it would be better to have a version of the `required_auth_by` that uses `a`-tags to point to the list of people allowed to download the event. But since every event will have the same pointer it might defeat the purpose.

fiatjaf:
> Yeah, I'd like to avoid that centralization point if we can. Medical information should not be controlled by a single provider.

Why a single provider? You can publish the same event to multiple relays as long as it's you doing it.

vitorpamplona:
> No, I'm talking about not having a new serialization scheme.

The issue is that then all current relays would accept and return those events without any auth. Which defeats the purpose of the approach.

fiatjaf:
> The issue is that all current relays would accept and return protected events without any auth. Which defeats the purpose of the approach.

I know, I mentioned that this would require some time for relays to catch on with the code changes required. I assume they would all eventually do except malicious relays -- but your scheme is also vulnerable to malicious relays.

vitorpamplona:
> Why a single provider? You can publish the same event to multiple relays as long as it's you doing it.

I will need to duplicate/sync the permission table between relays. For every patient, there is about 30 pubkeys (an average over all patients) with access in it. And those 30 keys will change from time to time. Syncing all permissions between multiple relays in your model requires its own protocol.

arthurfranca:
Changing the serialization to make it incompatible with regular NIP-01 relays was really smart =o

However, to keep relay code simple, i think the spec ideally needs to take into account something really important @monlovesmango [said](https://github.com/nostr-protocol/nips/pull/841#discussion_r1400924039) when we were discussing NIP-42 flaws.

The conclusion is that relays must be able to know, just by looking at the REQ filter (**without inspecting events on DB**), which authed pubkey(s) can download the requested events OR if the requested events don't require AUTH.

In practice i think the above translates to the following:
1) the different serialization scheme should be applied on specifc **event kinds** (so relays know **which events** require AUTH), which could also be listed on a NIP-42 section. Though not a fixed kind range. The "restricted" kinds list would be updated as new NIPs define new kinds;
2) there must be an indexable tag present on the REQ filter meaning the same as your "requires_auth_by" (so relays know **which pubkey** must do AUTH). We could use uppercase `A` tag for that;
3) the custom serialization could be:

```js
JSON.stringify([
    0,
    event.pubkey,
    event.created_at,
    event.kind,
    event.tags,
    event.content,
    "a" <- meaning "a"uth required
])
```

So, in a nutshell, some specifc kinds must be queried by #A.
Invalid filter: `{ kinds: [1059] }` (gift wrap kind example)
Valid filter: `{ kinds: [1059], #A: ["<pubkeyX>"] }`
Valid filter: `{ #A: ["<pubkeyX>"] }`
Invalid filter: `{ kinds: [1059], #A: ["<pubkeyX>, <pubkeyY>"] }` // invalid cause one can't auth at the same time with two pubkeys

arthurfranca:
#1033 shows how the changes I suggested above could be used.

monlovesmango:
I think it would be better to define a way in the auth flow to authorize a 3rd party pubkey to access your events in real time.

this would allow us to have specific event range that requires authorization (like @fiatjaf suggested), but also allow others to access your events (like @vitorpamplona is looking for). other benefits:

- access is granted in real time, meaning there is no need to handle revocation of access
- no need to have custom serialization
- no need to store data to track who is authorized
- events are less likely to be accessed inappropriately with a predefined auth range
- defines a way to grant nostr connect signer access to your ephemeral events (so that you can use the signer over an authenticated relay)

caveats:

- would probably require auth range for ephemeral and replaceable events too..? 
- I haven't actually thought of a way to do this

generally I am against allowing arbitrary auth requirement for any event for the reason that @arthurfranca highlighted, the relay will have to run filter query before being able to ascertain whether auth is required or not. this is a waste of relay's resources. 

I also think a range of auth event kinds is preferable to kind by kind auth requirements, as the former only needs to be implemented once on relays and the later will be ongoing maintenance (and I have little faith that relays will dutifully keep up with all new kinds that need auth).

>But then I realized that all event kinds would just be duplicated in this new range

honestly I think any duplication that would happen would only happen when its actually appropriate, bc it means that kind is being used in a new and novel way so it should have a new kind anyway. for example, if you did have kind 1 events that you only want certain users to have access to, imo its good to make that as a new kind.

what do people think about this strategy? if some think this has promise I can try to create a flow for this.

arthurfranca:
Changed my mind about the range, it would indeed be good for removing relay maintenance burden. Events on this range besides requiring AUTH should be parameterized replaceable (d="" would make them act as just replaceable; if referencing it by id it would act as non-replaceable; expiration tag can make it act as ephemeral).

My current opinion still is that there shouldn't exist event kinds that sometimes require AUTH and sometimes don't because of the argument that ideally relays should know in advance if user should AUTH just by looking at the REQ filter, without needing to access events on DB.

If one sometimes want an event such as `kind:1` to require AUTH, #1033 makes it work without needing to recreate it using the new kind range.

And the new serialization is really great to auto-prevent someone from broadcasting the event to relays that wouldn't honor the event access rule cause cause they didn't implement NIP-42.

monlovesmango:
>And the new serialization is really great to auto-prevent someone from broadcasting the event to relays that wouldn't honor the event access rule cause cause they didn't implement NIP-42.

this implies that users are putting together the events themselves which is not true. clients only allow users to create certain types of events. if a client wants to add a kind that a user can create, they should be implementing that appropriately. also clients are likely to use libraries that will account for this. I understand this is nice built in error checking but from practical perspective this doesn't have much benefit imo.

>The issue is that all current relays would accept and return protected events without any auth. Which defeats the purpose of the approach.

generally I think changing fundamentals of the protocol to intentionally cause errors in order to enforce policy is a terrible approach and gives a false sense of security. yes this specific issue would be solved, but it wouldn't solve:
- relays that have updated their nostr library, which accounts for serialization, but haven't actually updated the auth piece
- relays that are malicious and can serialize the protected events but do not follow auth spec (this the group I actually worry about)

this could also lead to terribly confusing user experience if they are not using any relays that support these events.

what if we created an authorization specific delegation? the auth delegation token would be specific to signing for kind 22242 events and the conditions query string would inform the relay which of the delegator's events the delegatee has access to.

arthurfranca:
> what if we created an authorization specific delegation? the auth delegation token would be specific to signing for kind 22242 events and the conditions query string would inform the relay which of the delegator's events the delegatee has access to.

@monlovesmango Very interesting. The delegatee would auth as himself while adding a `delegation` tag with a token that could also include `created_at<...` rule. Great to share with each monthly paying subscriber and it wouldn't make the protected event any bigger (with too many `A:tag`/`requires_auth_by:field` entries). Would be great paired with #1033 cause the `kind:40042` event author may be an always online bot/dvm signing these tokens and sharing to paying users. From me it's an ACK to stuffing it on NIP-42 as well.

monlovesmango:
ok i'll write a nip up for this just to have it. my main (selfish) motivation for this was to enable nostrConnect signing messages over authenticated relays (since the client technically wouldn't be able to auth on its own).

arthurfranca:
@vitorpamplona is too quiet while we DEFACE!! his PR xD I'm curious to know what you think about the following:
1) using an `A` tag instead of `requires_auth_by` extra field? (so `#A` REQ filter would clearly be a hint for relays telling if the currently AUTHed user can access the requested events, without looking at DB)
2) using 40000-49999 kind range for events that **always** require AUTH? (so `kinds: [1, 40007]` can also be a hint telling if users must AUTH, without looking at DB)
3) using #1033 to retrofit current event kinds that **transiently** require AUTH access control (people can for example delegate access control of nip-94 premium pics to an always online bot/dvm that upon detecting payment adds the paying user's pubkey to the `kind:40042` `A` tag list. this `kind:40042` event can be owned by the bot/dvm while the `kind:1063` one by the real author)
4) with this setup, the new serializtion can just add some fixed string as the 7nth item
5) add the auth delegation token spice from @monlovesmango so that one can keep the number of `A` tags small if sharing these tokens

vitorpamplona:
> i'll write a nip up for this just to have it

Please do. I think everyone is trying to achieve something different with these protected events. And I don't yet know if the one solution for everyone would be better than specialized solutions to each of the problems. 

> #A REQ filter would clearly be a hint for relays telling if the currently AUTHed user can access the requested events, without looking at DB

I am not sure how that is true. Whatever goes on the #A filter will not be known to the client. 

I can't understand https://github.com/arthurfranca/nips/blob/acu/00.md I don't have a need to preview things before seeking the record.

monlovesmango:
>@vitorpamplona is too quiet while we DEFACE!! his PR

hehehe he is too kind to us

1. agree. my opinion is that it would be good to be a queryable tag, so that any auth'ed user that is tagged can query a relay for events that tag themselves and the relay knows right off that the user is allowed access to those events. otherwise if the tag is not queryable the relay must first query on other filter attributes and only after getting the results realize that authentication is needed.
2. agree
3. interesting, could work. don't necessarily like that it may break existing use cases, but also its a neat solution to create paid content without much infrastructure. I actually think the auth delegation token would be a much cleaner solution for this use case rather than putting the buyers in the event tags (I would add a query condition to the auth delegation token that allows granting access to just a single event ID which would be perfect here)
4. I am not a fan of changing the serialization. but also not understanding why you would want to add a fixed string to the serialization (I understand why vitor is proposing a change to the serialization, but what do you mean by fixed string?)
5. agree

arthurfranca:
> I am not sure how that is true. Whatever goes on the #A filter will not be known to the client.

You are right that it won't be known to the client. But i thought that if I'm an user and want to fetch someone else's proteced event for instance, my client will 1) use the `{ ..., #A: ["me"] }` filter or 2) check my (?)NIP-51 list of shared privkeys to event ids/addresses-the-shared-privkeys-can-access mapping in look for the right privkey to use on the  `{ ..., #A: ["right-privkey"] }` filter  3) my NIP-51 list of shared @monlovesmango's delegation tokens to event ids/addresses mappings. In fact the `{ ..., #A: ["me"] }`  filter would be the last thing the client would fallback to.

> I can't understand https://github.com/arthurfranca/nips/blob/acu/00.md I don't have a need to preview things before seeking the record.

Imagine a microblogging client fetching a `kind:1` feed and one of these events is supposed to be protected, just for paying subscribers. This event simply shows a preview while it points with a `full_event` tag to the full event, a `kind:40042` one (on the 40000-49999 range of events that require AUTH).  `kind:40042` events besides requiring AUTH and including `A` tag(s), they mimick the event structure of whatever other kind that is the original preview event - in this case it's the kind 1 structure.

`kind:1` events aren't on the 40000-49999 range, so the mentioned NIP would "upgrade" it with the access control feature. It is great for relays that will only need to require auth when they see `{ kinds: ["i'm on the 40000-49999 range", ...]}`

arthurfranca:
> 3) [...] I actually think the auth delegation token would be a much cleaner solution for this use case rather than putting the buyers in the event tags

@monlovesmango If one is sharing auth delegation tokens, the protected event would have a single `A` tag with the event author's pubkey. So it is compatible with the auth delegation tokens but would have the flexibility to use the simple approach of adding a bunch of pubkeys on `A` tags (similar to what vitor proposed with the `requires_auth_by` list of pubkeys that can access the event).

> I am not a fan of changing the serialization. but also not understanding why you would want to add a fixed string to the serialization (I understand why vitor is proposing a change to the serialization, but what do you mean by fixed string?)

Adding anything as 7nth element to the serialization array is enough to achieve what vitor wants. As I changed his `requires_auth_by` extra event field for `A` tags, I have to add something else as the 7nth element like:

```js
JSON.stringify([
    0,
    event.pubkey,
    event.created_at,
    event.kind,
    event.tags,
    event.content,
    "a" // a fixed string instead of <requires_auth_by, as an array of non-null strings>
])
```

<- to generate an "invalid"  event id for today relays validating it, like this PR is proposing.

arthurfranca:
Ok, people have spoken. Down with the new serialization scheme!

I guess these event authors should publish these events just to relays that implement (the to-be-updated) NIP-42?

42.md
```diff
@@ -18,6 +18,42 @@ A relay may want to require clients to authenticate to access restricted resourc
...
+  0,
+  <pubkey, as a lowercase hex string>,
+  <created_at, as a number>,
+  <kind, as a number>,
+  <tags, as an array of arrays of non-null strings>,
+  <content, as a string>,
+  <requires_auth_by, as an array of non-null strings> 
+]
+```
+
```
arthurfranca, line 56:
```suggestion
### Private Event Kind Range

Events in the kind range `40000-49999` are considered always private. They MUST have atleast one `A` tag. Each of these tags is set to a `pubkey` allowed to request the event after authenticating using the `AUTH` flow defined on below section.

These events should be treated by `relays` as parameterized replaceable events. When a new NIP chooses a kind number in this range for an event, it should define if `clients` will consider that the event will:

1) Act as a non-replaceable event, by being referenced by its `id` and by using a random `d` tag value OR;
2) Act as a replaceable event, by using a `d` tag set to an empty string OR;
3) Act as a parameterized replaceable event, by using a non-fixed `d` tag value OR;
4) Act as an ephemeral event, by using an `expiration` tag set to a value soon to expire, by being referenced by its `id` and by using a random `d` tag value.

```

arthurfranca: @vitorpamplona what do you think? Using an event kind in a specific kind range to signal expected relay behavior does have the benefit of not requiring relay builders to check every new NIP in look for how they should treat each newly added event kind.

Removed the serialization in favor of simply condering those relays that advertise NIP-42 support on "supported_nips" as able to enforce event access control.

arthurfranca: @vitorpamplona please take a closer look at the following paragraph "These events should be treated by `relays` as parameterized replaceable events. When a new NIP chooses a kind number in this range for an event, it should define if `clients` will consider that the event will: 1) **Act as a non-replaceable** event [...]"

A parameterized replaceable event is flexible enough to be treated by `clients` as non-replaceable, replaceable or ephemeral. While relays just treat them as parameterized replaceable ones.

vitorpamplona: nah, that is too confusing. Better not have a range at all then

arthurfranca: I mean, there is no need to create 4 ranges (regular, replaceable, param repl. and ephemeral) but now requiring AUTH.

vitorpamplona: Why not just add the AUTH requirement option to any event? There is no need to recreate event kinds that already exist just to make them authed.

vitorpamplona: > ideally relays should know in advance if user should AUTH just by looking at the REQ filter"

I don't think this is true. Once the relay gets the event that match a filter, it can remove the ones that have the `A` if the connection is not authed. Or if you are using an SQL database, it's super easy to know if the user is auth, include As, if it is not, then remove them.

arthurfranca: > Once the relay gets the event that match a filter, it can remove the ones that have the A if the connection is not authed

This is ambiguous behavior. The client won't be able to know the cause of the missing events. Was it because the event didn't exist on the relay or the client had no authorization to fetch them?
Also, will the specific relay implementation do as you said and reply with some of the events or instead simply send none? The client can't be sure so can't code effectively a retry strategy.

This kind of situation was the reason we had to add the `CLOSED` message to NIP-42.

arthurfranca: I mean, today if a client isn't authenticated and asks for `{ kinds: [1, 4] }`, the relay should send `CLOSED` instead of just the `kind:1` events.

And that being the case, for relay implementation simplicity, if it receives a REQ filter without `kinds`, it may choose to (and frankly should) send `CLOSED` right away. Better than getting events from DB just to later find out it includes a private event, using resources for no reason, just to later discard events and send `CLOSED`.

vitorpamplona: > This is ambiguous behavior. The client won't be able to know the cause of the missing events.

This already happens all the time. It's nothing new. Half of relays send missing events even though we have received them before with the same filter. 

> Also, will the specific relay implementation do as you said and reply with some of the events or instead simply send none?

Yep. Simply send an EOSE if you don't have any. It already happens today with DM relays that filter out GiftWraps to the pubkey alone. 

> I mean, today if a client isn't authenticated and asks for { kinds: [1, 4] }, the relay should send CLOSED instead of just the kind:1 events.

That's bad. It should send the 1s and the EOSE when done. We mix in a lot of filters to go around subscription and number of filter limits per relay. Almost all of our filters include private and public events. It's virtually impossible to have something like Amethyst without merging several data needs in the same subscription.

vitorpamplona: FYI, I just looked at the guidance for CLOSED and it doesn't include "relay should send CLOSED instead of just the kind:1 events". And I think that should never be the case. If the relay is refusing to fulfill the subscription, better find something better than not accepting one of the filters.

arthurfranca: There were 3 PRs and other ideas trying to fix ambiguous AUTH flow until @fiatjaf came up with the new message.

My understanding and why i liked the solution is that `CLOSED` refers to a subscription as a whole and the client retries a sub with exactly the same filter after authenticating. That's because the message includes the sub id and the reason is just a brief code plus free text (doesn't explain the reason in detail for a machine to read - e.g.: you will know you need to auth but not which events triggered it): `["CLOSED", "sub_ID", "reason"]`

> Almost all of our filters include private and public events

The idea is if you haven't already authenticated, you would receive a `CLOSED`, authenticate then retry it with a `REQ` with same filter.

If you get some events and then EOSE it means you don't need to retry and many clients will certainly close the subscription at this moment.

arthurfranca: Opened #1046 PR editing the example to make it more clear.

arthurfranca: Opened #1049 PR to better explain the motivation

mikedilger:
I don't think I'll be implementing this in gossip or chorus.  You can call my software "nostr classic" if you want.  If you send in events with broken signatures chorus and gossip will reject them as you desire thus you will have your privacy.

vitorpamplona:
Closing this due to unnecessary complexity of the solution.

PR #1030, opened 2024-02-08
relay-based event access-control with the `["-"]` tag

fiatjaf:
Read about it in the NIP body: https://github.com/nostr-protocol/nips/blob/protected-events-tag/70.md

This is inspired by https://github.com/nostr-protocol/nips/pull/1029, there is more discussion there.

### DIFF
--- a/70.md
+++ b/70.md
@@ -50,7 +50,7 @@ This NIP gives these authors and relays the means to know when a given event is 
 
 ## How
 
-The default behavior for most relays will be to reject all events with the `"-"` tag. Only relays previously programmed to enforce access-control on these events would have to write special code for handling them.
+The default behavior for relays will be to reject all events with the `"-"` tag. The only relays that should accept these events are those that have been explicitly programmed to enforce write access control for all events with the `"-"` tag.
 
 Authors of protected events would have to have some kind of relationship with some specific relays -- for example, using [NIP-29](29.md) -- that would enable relays to know to whom these events can be served once they're published by the author.
 
--- a/70.md
+++ b/70.md
@@ -1,0 +1,51 @@
+NIP-70
+======
+
+Protected Events
+----------------
+
+`draft` `optional`
+
+When the `protected` tag is present, relays must enforce that the event is being published willingly by its author. In order to do that relays must require authentication before accepting the event and verify that the authenticated public key matches the event author public key.
+
+## The tag
+
+The tag is a simple tag with a single item: `["protected"]`. It may be added to any event.
+
+## Example flow
+
+- User `79be667ef9dcbbac55a06295ce870b07029bfcdb2dce28d959f2815b16f81798` connects to relay `wss://example.com`:
+
+```jsonc
+/* user: */
+["EVENT",{"id":"cb8feca582979d91fe90455867b34dbf4d65e4b86e86b3c68c368ca9f9eef6f2","pubkey":"79be667ef9dcbbac55a06295ce870b07029bfcdb2dce28d959f2815b16f81798","created_at":1707409439,"kind":1,"tags":[["protected"]],"content":"hello members of the secret group","sig":"fa163f5cfb75d77d9b6269011872ee22b34fb48d23251e9879bb1e4ccbdd8aaaf4b6dc5f5084a65ef42c52fbcde8f3178bac3ba207de827ec513a6aa39fa684c"}]
+/* relay: */
+["AUTH", "<challenge>"]
+["OK", "cb8feca582979d91fe90455867b34dbf4d65e4b86e86b3c68c368ca9f9eef6f2", false, "auth-required: this event may only be published by its author"]
+/* client: */
+["AUTH", {}]
+["EVENT",{"id":"cb8feca582979d91fe90455867b34dbf4d65e4b86e86b3c68c368ca9f9eef6f2","pubkey":"79be667ef9dcbbac55a06295ce870b07029bfcdb2dce28d959f2815b16f81798","created_at":1707409439,"kind":1,"tags":[["protected"]],"content":"hello members of the secret group","sig":"fa163f5cfb75d77d9b6269011872ee22b34fb48d23251e9879bb1e4ccbdd8aaaf4b6dc5f5084a65ef42c52fbcde8f3178bac3ba207de827ec513a6aa39fa684c"}]
+["OK", "cb8feca582979d91fe90455867b34dbf4d65e4b86e86b3c68c368ca9f9eef6f2", true, ""]
+```
+
+## Use cases
+
+- private groups
+- private feeds
+- optimistic hiding of DMs or other encrypted material metadata
+
+## Why
+
+Many people want to restrict access to some of their data to a closed group: by having, for example, a private feed for paying subscribers.
+
+Even though it's ultimately impossible to restrict the spread of information on the internet (for example, one of the members of the closed group may want to take an event intended to be restricted and republish it to other relays), most relays would be happy to not facilitate the acts of these so-called "pirates", in respect to the original decision of the author.
+
+This NIP gives these authors and relays the means to know when a given event is not intended to be broadcasted to all and everywhere.
+
+## How
+
+In practice, authors of protected events would have to have some kind of relationship with some specific relays -- for example, using [NIP-29](29.md) -- that would enable relays to know to whom these events can be served once they're published by the author.
+
+If someone tries to copy the event and publish it to other relays that wouldn't enforce the rules intended by the author that won't work as these relays will reject the event.
+
+In practice there may be relays that accept these "pirated" events, but that isn't different from what happens in the current internet.
--- a/70.md
+++ b/70.md
@@ -12,7 +12,7 @@ A protected event is an event that can only be published to relays by its author
 
 The default behavior of a relay MUST be to reject any event that contains `["-"]`.
 
-Relays that want to accept such events MUST first require that the client perform the `AUTH` flow and then check if the authenticated client has the same pubkey as the event being published and only accept the event in that case.
+Relays that want to accept such events MUST first require that the client perform the [NIP-42](https://github.com/nostr-protocol/nips/blob/master/42.md) `AUTH` flow and then check if the authenticated client has the same pubkey as the event being published and only accept the event in that case.
 
 ## The tag
 
--- a/70.md
+++ b/70.md
@@ -6,11 +6,17 @@ Protected Events
 
 `draft` `optional`
 
-When the `protected` tag is present, relays must enforce that the event is being published willingly by its author. In order to do that relays must require authentication before accepting the event and verify that the authenticated public key matches the event author public key.
+When the `"-"` tag is present, that means the event is "protected".
+
+A protected event is an event that can only be published to relays by its author. That allows the author to choose who can access their events (either by knowing in advance the relay policy with regards to reads or by means of an external relationship with the relay that is not specified here).
+
+The default behavior of a relay MUST be to reject any event that contains `["-"]`.
+
+Relays that want to accept such events MUST first require that the client perform the `AUTH` flow and then check if the authenticated client has the same pubkey as the event being published and only accept the event in that case.
 
 ## The tag
 
-The tag is a simple tag with a single item: `["protected"]`. It may be added to any event.
+The tag is a simple tag with a single item: `["-"]`. It may be added to any event.
 
 ## Example flow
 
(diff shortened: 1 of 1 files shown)


70.md
```diff
@@ -0,0 +1,51 @@
+NIP-70
+======
+
+Protected Events
+----------------
+
+`draft` `optional`
```
pablof7z, line 7:
```suggestion
`draft` `mandatory`
```

pablof7z: I like the simplicity, but it's gotta be widely adopted for it to be useful.

fiatjaf:
Also remember that in most cases, specially if a group is small enough, there will be no incentive for receivers of these protected events to go around searching for a rogue relay to which publish the data against the will of the original author.

And also most normal users wouldn't also download a special client that knows what are these rogue relays and uses them specifically to download the events leaked from groups that would otherwise be private. A normal NIP-29 client, for example, would never even try to download events from groups to which it doesn't belong.

Exceptions are not the rule.

vitorpamplona:
> this NIP only works if it gets mass adoption by relays
> like https://github.com/nostr-protocol/nips/pull/566, and reuse the authorization scheme already described in those

Those two things are mutually exclusive. 

The `protected` tag doesn't reference which authorization scheme the relay should use. If we have separate implementations of https://github.com/nostr-protocol/nips/pull/566, how would it know how to apply? 

Clients can use this tag to not automatically re-broadcast the event and require choosing relays when the user tries to re-broadcast. However users might not know where the event should go to make the choice themselves.

It feels very abstract/general. Which is unlike you @fiatjaf

arthurfranca:
Handy NIP to make an event be hosted just on relays that support NIPs/behavior the author needs.
Though feels not that useful without an event access control spec for relays.

monlovesmango:
love this. very simple but powerful (but as with all nostr policy, only as powerful as its adoption). honestly surprised me that this hasn't already been done.

I kinda think this should just be added to the NIP42 generally, right? if so, my next question is what use case are you envisioning where we would have authenticated writes but not authenticated reads (which would need this separate NIP)? if not, what use case would have authenticated reads but not benefit from authenticated writes?

fiatjaf:
NIP-29 groups, both private and public, would benefit from authenticated writes. DMs would benefit from authenticated reads and writes. Just kind1 notes would benefit from authenticated writes/unauthenticated reads when the user wants to have some control of what they have posted and a little more assurance that deleting notes would be possible, for example.

fiatjaf:
@vitorpamplona the reason I think it's ok to be generic here is that I just expect all relays to implement this with a two-line check: `if protected tag is present reject` by default so that requires it being generic -- but I agree that being generic is often bad and that this is very unlike me. Maybe I'm wrong here though.

fiatjaf:
What do you think of using a tag like `["-"]` instead of `["protected"]` for brevity?

vitorpamplona:
But this requires changing the code: `if protected tag is present reject`

Most relays will not even realize that this PR is a thing. The default relay behavior of folks who do NOT implement this PR is to accept, not to reject. That's why I proposed changing the serialization: to reject by default on current relays and make it backward compatible (relays don't need to change).

vitorpamplona:
Same for all clients: The default behavior is to accept the protected event as a regular event and re-broadcast at will.

monlovesmango:
>That's why I proposed changing the serialization: to reject by default on current relays and make it backward compatible (relays don't need to change) and they shouldn't be working with "protected" events without implementing this or the other PR.

but there's nothing stopping a relay from updating its serialization (which is easy if they are using a nostr library that keeps up with nips) without actually implementing the auth in the PR. so still doesn't accomplish the goal of forcing non-supporting relays to protect these events.

vitorpamplona:
> so still doesn't accomplish the goal of forcing

Nothing will ever force relays to auth. That's not the point. If relays want to fuck around they can do it in a lot of ways. 

The only point of the serialization change is to make sure only clients and relays that have seen NIP-42 and are minimally aware of the difference between public and protected events can receive and rebroadcast these events.

monlovesmango:
>The only point of the serialization change is to make sure only clients and relays that have seen NIP-42 and are minimally aware of the difference between public and protected events can receive and rebroadcast these events.

yes agree. my point is you can update your nostr library dependency (which has accounted for new serialization) without being explicitly aware of the difference between public and protected events. most aren't writing their serialization code themselves (I think).

vitorpamplona:
Agree. My hope was that libraries requested an extra param (isAuthed: Boolean = false) to the serialization function which fails if the relay is not aware of NIP-42

fiatjaf:
> Most relays will not even realize that this PR is a thing. The default relay behavior of folks who do NOT implement this PR is to accept, not to reject. That's why I proposed changing the serialization: to reject by default on current relays and make it backward compatible (relays don't need to change) and they shouldn't be working with "protected" events without implementing this or the other PR.

I'm not sure about that. There are like 3 or 4 big relay implementations. If they implement this, which they can very easily, we give the network a couple of months and suddenly this change is in effect mostly everywhere. I think it's worth trying. If it fails then it fails, we don't lose anything.

Changing the serialization would achieve the same result at the end, but at a much bigger cost. It would make writing basic Nostr event creation libraries and using these libraries much more cumbersome for everybody forever.

vitorpamplona:
> There are like 3 or 4 big relay implementations.

I think that is the main difference between our experiences. I don't expect the current public relay codebases to be the same as the relay code that will work with the protected events. In fact, I don't want to change the public relay's code at all. 

> Changing the serialization would achieve the same result at the end

The difference is subtle but it's real. In this approach, there will be public relays, protected relays, and the ad-hoc relays that don't know the difference between the two because they have not read this PR yet. Those ad-hoc relays will re-broadcast everything protected. 

Same for clients. There will be clients that deal with the public events, clients that can deal with the protected events and those who don't know about this and will just scatter protected events everywhere. 

The proposal to change the serialization minimizes the existence of 3rd set of both relays and clients, better protecting events that should be protected.

monlovesmango:
>The proposal to change the serialization minimizes the existence of 3rd set of both relays and clients, better protecting events that should be protected.

agree this is the case. I think where we disagree is whether the better protection is enough to warrant such a drastic change to protocol fundamentals. IMO its not bc the network is always going to be vulnerable to malicious relays that don't follow the rules anyway, and this is the subset I am more concerned about. non-malicious actors that are ignorant of this NIP and accidentally spread protected events will diminish over time. I'd rather wait for that slow process to work itself out than force it with changes in serialization.

arthurfranca:
I wanna see pull requests adding the `protected` tag to 10 relay implementations then we have a deal @_@ hahah

vitorpamplona:
> whether the better protection is enough to warrant such a drastic change to protocol fundamentals.

Don't get me wrong. I think the idea of changing the serialization is crazy as well (as I wrote in the PR itself). I don't think it is a **change** in fundamentals. To me, it is the start of a second Nostr relay network that is designed for private events (medical information in my case) while not forcing users to centralize in the hospital's or country's private relay. The public network shouldn't change to stay as simple as possible. 

I would love to find a way to add to the serialization process a step where the relay has to prove one of the pubkeys was authenticated and in order to verify the event. In Clients, the client must prove it has access to the private key of one of the pubkeys listed in the event. That would be a dream. I would allow the event to be verified only on relays that are being used by the listed keys and those listed keys could move the event to any relay or client they want.

fiatjaf:
> I wanna see pull requests adding the protected tag spec to 10 relay implementations then we have a deal @_@ hahah

Working on that.

70.md
```diff
@@ -0,0 +1,59 @@
...
+
+Many people want to restrict access to some of their data to a closed group: by having, for example, a private feed for paying subscribers.
+
+Even though it's ultimately impossible to restrict the spread of information on the internet (for example, one of the members of the closed group may want to take an event intended to be restricted and republish it to other relays), most relays would be happy to not facilitate the acts of these so-called "pirates", in respect to the original decision of the author.
+
+This NIP gives these authors and relays the means to know when a given event is not intended to be broadcasted to all and everywhere.
+
+## How
+
+The default behavior for most relays will be to reject all events with the `"-"` tag. Only relays previously programmed to enforce access-control on these events would have to write special code for handling them.
```
monlovesmango, line 53:
```suggestion
The default behavior for relays will be to reject all events with the `"-"` tag. The only relays that should accept these events are those that have been explicitly programmed to enforce write access control for all events with the `"-"` tag.
```

70.md
```diff
@@ -0,0 +1,59 @@
...
+
+`draft` `optional`
+
+When the `"-"` tag is present, that means the event is "protected".
+
+A protected event is an event that can only be published to relays by its author. That allows the author to choose who can access their events (either by knowing in advance the relay policy with regards to reads or by means of an external relationship with the relay that is not specified here).
+
+The default behavior of a relay MUST be to reject any event that contains `["-"]`.
+
+Relays that want to accept such events MUST first require that the client perform the `AUTH` flow and then check if the authenticated client has the same pubkey as the event being published and only accept the event in that case.
```
monlovesmango, line 15:
```suggestion
Relays that want to accept such events MUST first require that the client perform the [NIP-42](https://github.com/nostr-protocol/nips/blob/master/42.md) `AUTH` flow and then check if the authenticated client has the same pubkey as the event being published and only accept the event in that case.
```

monlovesmango (approved):
maybe name this 'Propagation Protected Events' or 'Write Protected Events' to make it clear it this NIP doesn't affect read access but doesn't really matter.

70.md
```diff
@@ -0,0 +1,59 @@
...
+======
+
+Protected Events
+----------------
+
+`draft` `optional`
+
+When the `"-"` tag is present, that means the event is "protected".
+
+A protected event is an event that can only be published to relays by its author. That allows the author to choose who can access their events (either by knowing in advance the relay policy with regards to reads or by means of an external relationship with the relay that is not specified here).
```
arthurfranca, line 11: > A protected event is an event that can only be published to relays by its **author**

What if optionally the `"-"` tag can have a list of pubkeys that are allowed to (re)publish the event (which may have been authored by someone else)?

It would be in line with for example a gift wrap that is authored by a random key but is meant to be "owned" by the receiving user (set on the `p` tag).

staab: That would leak a lot of metadata if used with gift wraps, defeating the entire purpose.

vitorpamplona: only users authed by the p-tag should be able to download these events. Anyone can upload.

monlovesmango: >What if optionally the "-" tag can have a list of pubkeys that are allowed to (re)publish the event (which may have been authored by someone else)?

I thought of this too actually, I think it would work well. but figured that this NIP can always be extended to add this feature in the future so maybe just shelve it until theres an actual use case. most important thing right now is making rejection of events with `"-"` tag the default behavior.

70.md
monlovesmango: maybe the requirement that relays reject events with `"-"` tag by default should actually be added to NIP01, and then this NIP would define when events with `"-"` tag can be accepted (when the author matches authenticated user). that would suggest that rejection of events with `"-"` tag is expected behavior, while making it optional to accept them under very specific parameters.

70.md
arthurfranca: I thought that with e.g. `["-", "<author-pub>", "<receiver-pub>"]` the receiver could rebroadcast it to backup for example a local copy to another relay.

But yeah for gift wraps it doesn't make much sense as the receiver could re-wrap the seal. Just an idea but yeah haven't give it much thought.

fiatjaf:
Implemented here -- with the necessary `AUTH` support already: https://github.com/fiatjaf/khatru/commit/cd4c25c84519a4965fdb48f462111373d38a5eb8

monlovesmango:
how about "Publication Protected Events"? or "Publish Protected Events"?

fiatjaf:
Whatever is fine for me, but I would prefer a smaller but meaningful name. A single word.

nostr-wine:
Simple enough and should be useful if relays adopt it. We will add this tag for subscriber-only events on creatr.nostr.wine.

mikedilger:
+1. Simple enough to get implemented everywhere quickly.

Implemented in chorus here:
https://github.com/mikedilger/chorus/commit/1e2187b86c7a4cd8b33672cf7cf6f777f0baf526

gazhayes:
This is by far the simplest solution to the problem, anyone who disagrees is probably a statist.

fiatjaf:
Strfry PR: https://github.com/hoytech/strfry/pull/100

jb55:
This is live on the damus relay! I will also make a nip70 strfry plugin for https://github.com/damus-io/noteguard

fiatjaf:
This is implemented in Khatru, Chorus, nostr.wine and relay.damus.io.

bezysoftware:
Shouldn't the default behavior of dropping events (without auth) be mentioned in nip-01?

PR #1033, opened 2024-02-09
Access Control Upgrade

arthurfranca:
This uses https://github.com/nostr-protocol/nips/pull/1029#issuecomment-1936120725 enhancement idea from #1029 to easily turn a public event such as a `kind:1` or `kind:30023` one into an event that can only be fetched by some pubkeys.

Read [here](https://github.com/arthurfranca/nips/blob/acu/00.md)

### DIFF
--- a/00.md
+++ b/00.md
@@ -1,0 +1,63 @@
+NIP-XX
+======
+
+Access Control Upgrade
+----------------------
+
+`draft` `optional`
+
+This NIP defines how to upgrade a public event with the property of limiting access to itself.
+
+## Upgrading a Public Event
+
+A public event (with no access control), such as of `kind:1`,
+must be published or updated with a `full_event` tag pointing to a
+`kind:30042` event address.
+
+It may include one of the following tags holding a preview of the full event content:
+`summary`, `image`, `audio`, `video`. The first holds a text excerpt while the others an URL.
+
+Example:
+
+```json
+{
+  "kind": 1063,
+  "tags": [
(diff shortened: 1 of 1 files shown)


staab:
Interesting idea, but isn't backwards compatible because it hijacks event kinds, causing a broken experience for clients that don't know how to filter them out or request them.

arthurfranca:
@staab Could you review the last commit please? I got rid of the new tags.

staab:
No, I really think this essentially mangles stuff. It will just result in broken-looking notes. Plus, we would need conventions for how to hobble every event kind to reveal only data that should be revealed, since if some people do it differently from others that could also break things. I really don't think this will work.

arthurfranca:
@staab How about now with the new commit?

staab:
No, I mean the special casing is my point, you're describing what to do with a kind 1, but what about with a 30023, or a nip 72 video event, or anything else? What about events that can't be truncated without making them meaningless? The juice isn't worth the squeeze.

arthurfranca:
Don't know. Not convinced of the difficulty you point out. Note that only clients that want to enable features such as "Publish just to Close Friends" or "Publish just to Subscribers" or some other feature that requires access control, would have to go into the trouble of dealing with generating previews or "You don't have access to this event" type of messages.

Such a client that is also specialized in 30023 events would have not much problem in trimming a markdown content.
I don't know what a NIP-72 video event looks like but a NIP-53 client that wants to offer a feature that requires access control has the option to use the "recording" tag with a video clip of some frames or use the "image" tag with a frame pic etc... the specialized client builder would have to think when implementing the feature. However non supporting clients have to do nothing.

PR #1035, opened 2024-02-10
add nip-302 relay pools

Cameri:
This PR adds NIP-302 Relay Pools to the Nostr NIPs repository.

### DIFF
--- a/302.md
+++ b/302.md
@@ -1,0 +1,59 @@
+NIP-302
+=========
+
+Relay Pools
+-----------
+
+`draft` `optional`
+
+# Introduction
+
+This NIP introduces a system for the creation, management, and utilization of relay pools.
+
+# Specification
+
+## Creating a Relay Pool
+
+Users initiate a relay pool by creating a Nostr event with `kind` 30010.
+- The `content` field is the description of the relay pool.
+- A `d` tag has the name of the relay pool (e.g. awesome-pool) as value.
+- At least one `r` tag MUST be present with the third value as `"pool"`. These pool addresses are URLs clients can connect to.
+- Zero or more `r` tags with third value as `"member"`, empty string, or missing signify a relay pool member.
+- Zero or more `auth-required` tags with one of the following values: `nip-42` or `nip-98`. Clients wishing to connect to Relay Pools with an `auth-required` tag MUST support at least one of the NIPs before initiating a connection. 
+
+## Updating the Relay Pool
+
+To update the relay member list of a relay pool, pool addresses, a new event of kind 30010 with the same `d` tag MUST be published.
+Pool addresses and member relays are added or removed at any time.
+
+## Joining a Relay Pool
+
+Relay operators wishing to join a relay pool publish a Relay Pool Join Request event with `kind` 8000:
+- `p` tag referencing the pubkey of the Relay Pool (event kind 30010).
+- `a` tag addressing the kind 30010 event.
+- a single `r` tag with the relay address (e.g., wss://cool-relay.cool-domain.com).
+
+## Inclusion in the Relay Pool
+
+Relay Pool Operators MAY include new member relays by updating the Relay Pool event and adding their `r` tag from the kind 8000 event.
+
+### Leaving a Relay Pool
+
+Member Relay Operators can request removal from a relay pool by publishing an Event Deletion (kind 5) referencing their Relay Pool Join Request event (kind 8000).
+Upon receiving a kind 5 event, the Relay Pool Operator SHOULD issue a new event of kind 30010, removing the `r` tag that references the parting relay.
+
+# Verification
+
+Relay Pool Operators MAY require further steps as part of the application process (e.g. proof of work (mined event), payment or out-of-channel communication).
+Relay Pool Operators MAY respond with kind 1 note to Relay Pool Join Requests events by referencing the event kind 8000 and/or tagging the requester's pubkey.
+A Relay Pool is only successfully joined once a new Relay Pool event is published including the `r` tag from the Relay Pool Join Request.
+
+# Client Connection
+
+Clients can discover relay pools by subscribing to events of kind 30010.
+Clients MUST connect directly to the relay pool addresses in order to be routed to member relays.
+Relay Pools SHOULD distribute client connections among Relay Pool Members using a fair algorithm.
+
+## Proxying and Authentication
+Relay Pools MAY serve incoming WebSockets connections either by passthrough or by redirect.
+Relay Pools that require authentication (e.g. for paid relay pools) MUST support either [NIP-42](42.md), [NIP-98](98.md) or both.


vitorpamplona:
I am not sure I understand what the idea is. What's the difference between a Relay Pool and a Relay List? What is the pool doing that a list doesn't do?

staab:
Interesting idea, is there a discussion somewhere on why this is needed? I saw some stuff on nostr about the pool members replicating content across the pool to create a distributed relay of sorts. I think this could be useful to strike a balance between a few huge relays and many small ones, but I'm not sure how it would actually work.

Cameri:
@vitorpamplona a Relay Pool is managed by someone operating a pool of relays, to which clients are proxied to. Relay Lists are kind of a user thing? Let me know if that clears it up.

Cameri:
@staab recently here `note1wzcaccvyh6tcqyac4kvwffjrmx833azlgurq8hgr9hxpqp45fe6s88g8vn` but it's been talked about in many other threads and instances

vitorpamplona:
> a Relay Pool is managed by someone operating a pool of relays, to which clients are proxied to

So, is it just like a CDN for relays? 

Is the idea for Clients to connect with the first one that works or connect with all of them?

Cameri:
@vitorpamplona tl;dr the relay pool has pool addresses and distributes incoming client connections to the member relays. You can connect to one or as many pool addresses as the relay pool has available, up to you. What you should not do is connect to the list of member relays directly if the idea is to use the pool. It should be in the 302.md document I put together for your review.

staab:
So is the idea that the relays wouldn't accept connections directly? A pool would look to users exactly like a relay, so some indication that it's a pool and so has certain benefits would be necessary in order for it to be used. This sort of falls under the usual problem of relays being hard to differentiate from one another. I added relay reviews a long time ago to help with that, but much more needs to happen.

I'm not against this at all, maybe build it first then we can add it to nips?

vitorpamplona:
Ohh I see. This is the setup for Proxy relays, not a CDN type of thing. That makes more sense. 

> What you should not do is connect to the list of member relays directly

This was not clear in the text.

Cameri:
@staab a relay pool should look probably like 1 relay to clients. Relay pool reviews would work well, as it would help users  find good relay pools and less ones with a bunch of slow raspberry Pi 3s.

A paid relay pool would distribute its revenue among the member relays based on some reliable metric, and some HTTP header magic would ensure member relays know when the connection is coming from the pool. A member relay could definitely receive direct connections but bypassing the proxy means the relay won't get paid by the relay pool in this scenario.

Relay pools should help distribute traffic among new or smaller relays.

Cameri:
> > a Relay Pool is managed by someone operating a pool of relays, to which clients are proxied to
> 
> So, is it just like a CDN for relays?
> 
> Is the idea for Clients to connect with the first one that works or connect with all of the relays in the pool?

https://github.com/nostr-protocol/nips/pull/1035/files#diff-08b0b8ade803297426a58ca593248aa42940eb97fa6f24bcb30bcff8caef7ae8R54

I should probably rephrase this then or include what it is not.

Cameri:
> I'm not against this at all, maybe build it first then we can add it to nips?

That's the plan :D

vitorpamplona:
> I should probably rephrase this then or include what it is not.

I would just name the NIP Proxy Relays. A pool of relays can mean so many things.

staab:
I don't know, pools rely on proxying, but not vice versa. Proxies don't require coordination, which is basically all this pr adds.

Speaking of proxies, this would really only work for backends that don't require end user AUTH (proxies could auth as themselves with backends) since proxying auth isn't currently possible, and I'm not sure adding support would be a good idea. See #1016 for reasons.

Cameri:
@staab if I understood you correctly, this NIP as-is wouldn't work with relays that require AUTH. If AUTH is done only with the Relay Pool/Proxy and the proxy passes a header that the member relay trusts, that could work... with NIP-98 perhaps?

fiatjaf:
Maybe the NIP should define what it means by a "relay pool" or invent a new term if this is confusing.

Cameri:
I'll work on a POC and circle back. For the time being I'll assume relay members don't have AUTH or at least would disable auth for proxied requests.

fiatjaf:
How are these events supposed to be used? By whom?

Cameri:
> How are these events supposed to be used? By whom? From reading the NIP I would think clients can just ignore the existence of pools entirely as the pools will perform redirects and so on automatically.

the Relay Pool event (Ignore the name for now) defines a relay pool, and is published by an operator of a pool

the Relay Pool Join Event is published by a relay operator that wishes to join a pool

fiatjaf:
Oh, I see.

seanmurphy1661:
What ip address would the relay see/

Cameri:
> What ip address would the relay see?

It depends how the load balancing is done, whether by proxying or redirects. Redirecting would leak relay IPs but proxying would not. I'm intentionally leaving out how the load balancing within the relay pool will be done.

melvincarvalho:
- high level thoughts
  - like the idea of relay pools
  - relay pools are not just relay lists, they imply greater collaboration
  - relay pools are not the same as relay proxies, i dont think, but a relay proxy is one service that could be offered, there's more than one way to do proxies, eg haproxy
  - it would be good if relay pools were allowed to home page or git page, that users could look at
  - it would be great if relay pools could accept donations via a lightning address
  - the workflow of join, leave, invite etc could be different for each pool, but there could be common strategies, similar with revenue splits, pools could have their own governance
  - what to do about bad actors that want to join a pool then harvest user ip address and query history
  - relay pools can collaborate on more things, such as federated services, storage services, backup services, spam protection
  -

vitorpamplona:
Is the name pool coming from the idea of "mining pools"? 

The idea that a relay operator could join any pool out there to provide storage to the pool, reduce latency and increase speed to clients is great. I just don't know how to coordinate all of these needs.

Cameri:
> Is the name pool coming from the idea of "mining pools"?  That is now making more sense to me.

Indeed, the name mining pools inspired me to name this NIP.

Cameri:
> the relay pool join event is irrelevant to the core of the spec which is basically making relays be referred to by naddr instead of URL, and allowing clients to randomly pick from a list of URLs to use
> 
> 
> 
> I propose this specification be split in two, one for relay pools, and one for that functionality
> 
> 
> 
> also you can't use NIP-98 on WS

No

Cameri:
> the relay pool join event is irrelevant to the core of the spec which is basically making relays be referred to by naddr instead of URL, and allowing clients to randomly pick from a list of URLs to use
> 
> 
> 
> I propose this specification be split in two, one for relay pools, and one for that functionality
> 
> 
> 
> also you can't use NIP-98 on WS

This NIP is for creating relay pools and for relays to have a way to join them. Whether clients implement this NIP to find relays is secondary and a NIP just for clients is ridiculous as we don't split NIPs that way.


Again: No

arthurfranca:
Let's say relay pool `P` contains relays `A` and `B` and that both Alice an Bob have set `P` as their `write` relay.
Alice publishes eventAlice1 to `P` that ends up at `A`.
Bob publishes eventBob1 to `P` that ends up at `B`.

What happens when someone asks `P` for events using the `{ authors: [Alice, Bob] }` filter?

It looks like that the pool would only work if `A` and `B` had all relay pool events.

PR #1037, opened 2024-02-12
NIP01 Clarify missing 'd' tag in Param Repl Event is interpreted as empty string

monlovesmango:
clarify that when `d` tag is missing in parameterized replaceable events it will be interpreted as `d` tag with empty string

https://github.com/monlovesmango/nips/blob/NIP01-clarify-missing-d-tag/01.md


fiatjaf:
I don't think we should bloat the spec with this. The spec says there should be a `d` tag. There is no point in trying to accomodate for buggy clients. If relays want to do it they can come up with their own standard.

monlovesmango:
thas fair. closing.

PR #1040, opened 2024-02-12
File metadata sets

fr4nzap:
Groups of kind 1063 File Metadata events

### DIFF
--- a/51.md
+++ b/51.md
@@ -49,7 +49,7 @@ Aside from their main identifier, the `"d"` tag, sets can optionally have a `"ti
 | Curation sets | 30005 | groups of videos picked by users as interesting and/or belonging to the same category        | `"a"` (kind:34235 videos)                                                         |
 | Interest sets | 30015 | interest topics represented by a bunch of "hashtags"                                         | `"t"` (hashtags)                                                                  |
 | Emoji sets    | 30030 | categorized emoji groups                                                                     | `"emoji"` (see [NIP-30](30.md))                                                   |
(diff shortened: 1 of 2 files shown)


vitorpamplona:
Ack

fiatjaf:
Why is there an app identifier? That sounds wrong. Other apps are not supposed to see these lists?

fr4nzap:
> Why is there an app identifier? That sounds wrong. Other apps are not supposed to see these lists? In that case it's better to use NIP-78, no?

Hmm that's not what I mean by app identifier.

Take the example of a Sparrow Wallet release: https://github.com/sparrowwallet/sparrow/releases/tag/1.8.2

Every artifact (`sparrow-1.8.2-1.aarch64.rpm`, ...) would correspond to a kind 1063 file metadata event, and the  release itself would correspond to this kind 30063. In this case, the app identifier would be `com.sparrowwallet.sparrow`.

fiatjaf:
I see.

But then the idea here is that this thing should only be used for files relative to app releases? That's way too specific! Aren't there other use cases?

If this is the only stated use case that's fine too, but then rename the list as "release artifacts lists" or something like that.

fr4nzap:
Yes, that is my specific use-case for zap.store. I thought I'd leave it generic in case someone wants to use it for something else, like a photo album where each photo is a kind 1063. Good for me either way.

fiatjaf:
Wow, I didn't know you were working on that. Looks great.

I'd say we have a specific kind just for software binaries and stuff like that, not make it more generic than it should.

If other more generic use cases happen later we can add new kinds for them.

fr4nzap:
Thank you @fiatjaf ! Just pushed updated to make it specific

fiatjaf:
If you feel the need to update it as you develop things please come back with a new PR.

PR #1041, opened 2024-02-12
Add PGP as NIP-39 external identity

fr4nzap:
I thought it would make sense to add PGP here as it is an external identity, but open to move it somewhere else

Should the full PGP public key be included as well?

### DIFF
--- a/39.md
+++ b/39.md
@@ -20,6 +20,7 @@ A new optional `i` tag is introduced for `kind 0` metadata event contents in add
     ["i", "twitter:semisol_public", "1619358434134196225"],
     ["i", "mastodon:bitcoinhackers.org/@semisol", "109775066355589974"]
     ["i", "telegram:1087295469", "nostrdirectory/770"]
+    ["i", "pgp:A999B7498D1A8DC473E53C92309F635DAD1B5517", "xsFNBF2V8eEBEADmjYzGOpxEI0J7jQ1qFzlsrjF6NaBSq+UqKw..."]
   ],
   ...
 }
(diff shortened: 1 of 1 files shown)


fiatjaf (approved):

gzuuus:
It would also be very interesting to introduce external identities native to nostr, e.g. an external key event that signs your identity can be used as proof of external identity. In nip-41 pr #1032 I use this to determine a proof for the next master key rotation. For example `[ "i", "nostr:external-pubkey", "event-id-of-the-external-pubkey-signing-your-pubkey" ]`.

fr4nzap:
@gzuuus I lack the knowledge to fully understand your NIP-41 proposal but we 100% need a sound way of rotating keys in nostr. Thanks for writing that up

fr4nzap:
@Semisol I randomly found out that the merging of this PR has been reverted by you. Would you mind commenting here what is wrong and how in your opinion it should be fixed?

caesar:
It would be great to add OpenPGP keys here again somehow. It would be useful in order to add Nostr support to Keyoxide.

@franzaps I would have thought your second suggestion would be sufficient, I don't see that it should be necessary to create a Nostr event, signing a text string containing the OpenPGP fingerprint should be enough.
@Semisol do you have any input on that, since it was you that pointed out (correctly) that just signing the fingerprint wasn't enough?

One other thing, perhaps the prefix should be `openpgpfpr`, since that's a [standard](https://metacode.biz/openpgp/openpgp4fpr)?

franzaps:
@caesar this is old, it was merged then removed. It continued here https://github.com/nostr-protocol/nips/pull/1182 which we started using in zap.store and soon be merged into NIP-39 (will send PR)

caesar:
@franzaps thanks! I'll comment there.

PR #1042, opened 2024-02-12
NIP-97 Login with Nostr

Semisol:
This spec proposes a simple login flow to allow easy authentication to services like nostr.build without having to deal with DM verification or other confusing flows.

### DIFF
--- a/97.md
+++ b/97.md
@@ -24,9 +24,17 @@ The response MUST be a JSON object, with the format:
 ```json
 {
   "status": <"success" or "error">,
-  "message": <OPTIONAL string message to show user in the case of an error>
+  "message": <OPTIONAL string message to show user in the case of an error>,
+  "redirect_url": <OPTIONAL redirect URL>
 }
 ```
+
+For `redirect_url`, this should be used to bring the user to where they were while initiating the login process (now with the login complete).
+This should be used in a way that is compatible with a different device being used to authorize the sign in than the one being logged in on.
+An example for this is checking if the sign in was attempted from this device with a cookie, if so, redirect the user to where wanted to log in,
+otherwise show a "continue on device" message.
+
+The `redirect_url` SHOULD be shown to the user and in a way that indicates it will take the user back to the app, such as `Return to website`.
 
 HTTPS should always be used except for `.onion` services, which should be contacted using HTTP.
 


vitorpamplona:
Looks good, as long as the server doesn't use the identifier as the session cookie. 

`created_at` check (must be < 5 minutes from the request) could also be performed.

alexgleason:
Why not use [NIP-98](https://github.com/nostr-protocol/nips/blob/master/98.md)?

You can even use NIP-07 to sign NIP-98 messages.

fiatjaf:
I don't get this at all. What happens when you click? Who is going to handle that link?
I think it's better to let websites come up with their own login process using NIP-98.
Maybe recommend a common pattern on that NIP.

97.md
```diff
@@ -0,0 +1,25 @@
...
+
+A login URI is defined of the format `nostr+login:<domain>:<identifier>` where `domain` MUST be a valid DNS domain or .onion service. The `identifier` MUST only consist of `A-Z a-z 0-9 _ - .`.
+
+This login URI can be presented as a clickable link, a QR code or a copyable string.
+
+## Login process
+
+A client that wishes to log in to a service SHOULD display the domain associated with the service to the end user before allowing them to log in to prevent services showing a login string for another service.
+
+After the user approves the login, the client should send a POST request to `/.well-known/nostr-login` to the `domain` with the `i` query parameter set to the `identifier` and a valid NIP-98 authentication header present.
```
vitorpamplona, line 21: Change it to `challenge` to be consistent with NIP-42's flow.

97.md
```diff
@@ -0,0 +1,25 @@
+NIP-97
+======
```
vitorpamplona, line 2: Couldn't this just be added to NIP-98?

97.md
```diff
@@ -0,0 +1,25 @@
...
+
+Nostr Login
+-----------
+
+`draft` `optional`
+
+This NIP defines a method for a way for users to prove to apps that they control a certain private key. This can be used to enable login for services like the management panel of image hosts.
```
vitorpamplona, line 9: This NIP defines a login flow where apps can request other apps to prove the user controls a private key. The first app displays a link or a QR code with a `nostr+login:` URI. The app with login keys receives that information, signs a NIP-98 payload, and sends it back to the website of the first app, authenticating the user.

vitorpamplona (approved):
Just some nitpicks.

97.md
fiatjaf: Yes.

fiatjaf:
What about NIP-46?

mikedilger:
> not everyone has an extension. the point of this is if you don't have an extension, you could just scan a QR code/click a link and prove you control an npub without inputting your nsec everywhere.

I missed the part where you prove anything. Is that part of this NIP?  How does clicking a link prove anything? Maybe it is just missing a sentence where this is explained.

jb55:
cc @danieldaquino for review since you introduced a similar flow for damus purple

Egge21M:
> > not everyone has an extension. the point of this is if you don't have an extension, you could just scan a QR code/click a link and prove you control an npub without inputting your nsec everywhere.
> 
> 
> 
> I missed the part where you prove anything. Is that part of this NIP?  How does clicking a link prove anything? Maybe it is just missing a sentence where this is explained.
> 
> 

The request has a nip-98 header

97.md
```diff
@@ -0,0 +1,25 @@
...
+
```
danieldaquino, line 10: Would it be a good idea to add a bit more detail on the intended use cases?

The first time I read the document it was not clear that this login is meant for non-Nostr apps/clients. I was confused for a minute thinking that this was attempting to be a way to login into a Nostr client (where multiple Nostr events need signing)

97.md
danieldaquino: I think they are best as separate NIPs, because a client might want to implement NIP-98 for authenticated communication with a server (for any particular purpose), without needing to necessarily support to Nostr Login flows.

97.md
danieldaquino, line 21: We should also specify the return payload, to indicate success or failure. Perhaps something like:

Return HTTP code:
- 200: Login successful
- 401: NIP-98 auth failed
- ...

```
{
    "status": <"success" or "error">
    "redirect_url": <OPTIONAL_URL_TO_CONTINUE_LOGIN>
}
```

This would help clients know what to display to the user after they accept the login

97.md
danieldaquino: I like "Login with Nostr" too

97.md
danieldaquino: On top of my comment above, I believe whatever session auth token or secret auth information needs to be (or at least should be) on this payload (Maybe embedded into the `redirect_url` for simplicity), to avoid phishing attack vectors.

To understand the possibility of a phishing vulnerability, consider the following:
1. Alice clicks on a link to `fake.com/login` (A fake website imitating `real.com`)
2. `fake.com` detects Alice entered the website, and then crawls `real.com/login` to obtain the NIP-97 link `nostr+login:real.com:abc123`. It then copies and displays this URL as a QR code on `fake.com/login`
3. Alice, unknowingly, scans the QR code which opens her Nostr client. She sees a prompt "Do you want to login into `real.com`?"
4. She clicks "confirm". The Nostr client proceeds to send the NIP-98 authenticated request to `real.com`
5. In this particular implementation of `real.com` _(which meets the specs of this NIP-97 draft)_, as soon as the `real.com` server receives the NIP-98 auth'd event, it generates an auth session token and sends it to the original browser session that received the login URL — to allow the login to proceed on the same tab.
6. Thus, in this particular implementation, the attacker successfully tricked Alice into giving the them access to her account on `real.com`.

However, if `real.com` implemented it in such a way that provides the nostr client with the redirect url containing the auth session token info, this type of phishing attack would not be possible

danieldaquino: > that is the user's fault and therefore this is a non-issue
> 
> your proposal breaks logging in on desktop using a mobile device also

I see, you have a point. However, it is reasonable to allow developers to optionally implement the flow with redirect urls if the tradeoff makes sense for their particular use case. 

Besides, while @jb55 and I were developing a very similar flow for our Damus Purple checkout page, we found that such redirect URL mechanism helped improve the UX on mobile-only flows (Some early testers were confused when we did not have them)

The best solution is to make the `redirect_url` an optional field on the JSON response, and thus allow each website to choose whether or not to use it (Whichever works best for their use case).

danieldaquino (commented):
I really like this spec, and I think it will be very useful. 

Only thing I would add to it is an _optional_ `redirect_url` field to the JSON response from the server, but otherwise it looks good to me.

97.md
```diff
@@ -0,0 +1,33 @@
...
+A login URI is defined of the format `nostr+login:<domain>:<challenge>` where `domain` MUST be a valid DNS domain or .onion service. The `challenge` MUST only consist of `A-Z a-z 0-9 _ - .`.
+
+This login URI can be presented as a clickable link, a QR code or a copyable string.
+
+## Login process
+
+A client that wishes to log in to a service SHOULD display the domain associated with the service to the end user before allowing them to log in to prevent services showing a login string for another service.
+
+After the user approves the login, the client should send a POST request to `/.well-known/nostr-login` to the `domain` with the `i` query parameter set to the `challenge` and a valid NIP-98 authentication header present.
```
basantagoswami, line 21: `.well-known/nostr-login` should be changed to `.well-known/nostr/nip97` or `.well-known/nostr/login` to make the syntax similar to [NIP-96 HTTP File Storage Integration](https://github.com/nostr-protocol/nips/blob/master/96.md)

vitorpamplona:
@Semisol do you have any server where this was implemented already? I need to test Amethyst's implementation of this.

97.md
basantagoswami, line 21: NIP-96 uses `.well-known/nostr/nip96.json`, we should follow a similar pattern here: `.well-known/nostr/nip97` or `.well-known/nostr/login`

fabianfabian:
For testing: https://media.utxo.nl/nip97test/nip97login.php

TheRebelOfBabylon:
> For testing: https://media.utxo.nl/nip97test/nip97login.php

I don't think your server properly enforces NIP-98. It expects the event to have a `u` tag containing only the URL without the query parameters and then the `i` query parameter is in a separate tag. But if you look at the spec for NIP-98, the `u` tag should include the entire absolute request URL including query parameters.

fabianfabian:
> > For testing: https://media.utxo.nl/nip97test/nip97login.php
> 
> @fabianfabian I don't think your server properly enforces NIP-98. It expects the event to have a `u` tag containing only the URL without the query parameters and then the `i` query parameter is in a separate tag. But if you look at the spec for NIP-98, the `u` tag should include the entire absolute request URL including query parameters.

I'm checking the `i` tag for the challenge but I see the spec doesn't describe it like that, not sure what happened there probably I read it too quick, I'll update the test code later, this is the current test code: https://gist.github.com/fabianfabian/0c68d75aeb84860dcd7fc2995ec98af6

TheRebelOfBabylon:
I've also implemented a server for testing but only accessible via REST API: [https://git.rebelofbabylon.com/api/v1/login](https://git.rebelofbabylon.com/api/v1/login)

jb55:
> I really like this spec, and I think it will be very useful.
> 
> Only thing I would add to it is an _optional_ `redirect_url` field to the JSON response from the server, but otherwise it looks good to me.

agreed, this is pretty important.

jb55 (approved):
great!

melvincarvalho:
Doesnt look at secure as NIP-98.  What is actually proved?  Could the domain impersonate another users?  Could the domain steal sats?

melvincarvalho:
What are the security trade-offs?  I am assuming it's less secure than NIP-98, but it's not 100% clear how.  Perhaps a security considerations section in the NIP could explain more.

melvincarvalho:
Security concerns: Domain compromise could impact login systems, potentially harming Nostr's reputation. Motivation and security surface would benefite from further elaboration.

Existing efforts: Consider ongoing Nostr integration with Oauth/OIDC for new authentication methods.

Suggestion: Future auth NIPs should include a comparison section:

- Comparison with existing auth methods, e.g. nostrlogin, nip98, nwc, nsecbunker, there's more ongoing oauth and oidc integrations ive seen too
- Advantages offered
- Use cases
- Adoption status

Concept NACK: Needs clearer explanation and/or broader adoption to justify implementation (IMHO).

Just my 2 cents.

melvincarvalho:
Aside:  would we benefit from a label: **auth**, for auth related PRs?

danieldaquino (approved):

fabianfabian:
> > @Semisol do you have any server where this was implemented already? I need to test Amethyst's implementation of this.
> 
> I'll implement one on nostr.build in the next couple of days.

is this available on nostr.build? I want to try it out

jb55:
> Security concerns: Domain compromise could impact login systems, potentially harming Nostr's reputation. Motivation and security surface would benefit from further elaboration.
> 
> Existing efforts: Consider ongoing Nostr integration with Oauth/OIDC for new authentication methods.
> 
> Suggestion: Future auth NIPs should include a comparison section:
> 
> * Comparison with existing auth methods, e.g. nostrlogin, nip98, nwc, nsecbunker, there's more ongoing oauth and oidc integrations ive seen too
> * Advantages offered
> * Use cases
> * Adoption status
> 
> Concept NACK: Needs clearer explanation and/or broader adoption to justify implementation (IMHO).
> 
> Just my 2 cents.

The main usecase we had is that we wanted to explicitly give memberships to npubs to have the ability to display stars on profiles and to give them a purple member number, I see this as more of a thing for niche thing for specifically verifying that a user controls a key to an npub, not as a generic login mechanism, although its true its likely this will be abused for that.

I think pushing npubs as a login mechanism is a bit sketchy, because once your key leaks you basically have doors wide open into all the websites you visit. Maybe we should make this explicit in the NIP that this should be probably just be used as a two factor mechanism or for niche use cases like verifying a user controls a key to an npub? Not as a main login mechanism.

PR #1043, opened 2024-02-12
Add audio track NIP

staab:
Prior art:

- `32123` published by wavlake do zap splits wrong and are not published by the artist. Metadata is also json-encoded in the content.
- `31337` is published by zapstr.live and includes `c` tags which could be added here, although `t` probably would suffice. The `media` tag is missing lots of possible metadata. `subject` could be a good alternative to `title`.
- `1808` isn't replaceable, but includes different urls for download/stream, hash, and waveform.
- #923 re-defines metadata and doesn't support multiple urls for different uses.

This NIP relies on conventions established by NIPs 94 and 92 to support multiple urls and complete metadata simultaneously. It's extremely minimal, and can easily be extended to support:

- Categories (NIP 32, `t`, or `c` tag)
- Text tracks (via NIP 92 imeta)
- Albums (kind 30029 lists)
- Zap splits (as defined in NIP 52)
- File metadata via NIP 94


blastshielddown:
It would be helpful to state the main goal of this NIP. I am totally with you that it would be valuable to center around a standard way of sharing and describing media but without a clear purpose I'm afraid we'll just end up trading opinions about implementation details that might have vastly different objectives.

From my perspective, I would love to be able to publish events so that a user of any client could listen to the song(s) contained in the event and see any relevant metadata. I'm interested to hear if you have other use cases in mind.

A few questions and points of clarification:

1. Can you please elaborate on `wavlake do zap splits wrong`?
2. To the point that events "are not published by the artist", one of our major constraints here is most of our artists do not use Nostr. We would like to change this but it will take some time and will require artists opt-in and secure an nsec (as of now I'm not keen on storing keys for artists and signing on their behalf).
3. What is the problem with metadata being stored as a json in the event itself?

staab:
Sure, the goal of this nip is just to have a standard that everyone likes for publishing audio tracks. I think this probably requires that is be 1. replaceable, and 2. support multiple urls with different content types. So no current standard works yet. Of course there are slight differences between what exists so far:

- Stemstr is doing remixes, and so has `e` tags to link things together.
- Wavlake is re-posting as a feed, as a proxy for the artist, so pubkey != artist in this case
- zapstr.live is more straightforward.

But the three use cases have the same core, and should be interoperable. I would use a separate event kind referring to these base events to implement more targeted use cases.

1. According to [NIP 57](https://github.com/nostr-protocol/nips/blob/master/57.md) `zap` tags look like this: `["zap", "82341f882b6eabcd2ba7f1ef90aad961cf074af15b9ef44a09f9d2a8fbfbe6a2", "wss://nostr.oxtr.dev", "1"]` but wavlake publishes `["zap", "zap@www.wavlake.com", "lud16"]`. I get sort of why this probably happened, since not all artists have a pubkey, but in that case I'd just leave it off and make sure the profile's lud16 is the correct fallback.
2. Totally get this. The new format should support that use case. If the artist joins, the events could be deleted and re-posted, or we could add some kind of "redirect" tag that points to the new canonical event.
3. Nothing necessarily (this is how kind 0 works), but if we're starting from scratch a more idiomatic way to do nostr metadata is in tags. New specs should also try to make event `content` human-readable so social clients can read the events without special casing rendering for them.

joshr4:
Love to see the discussion on this. Yeah, Wavlake zaps ought to be updated now that the lud16 zap tag is no longer within spec. Wavlake zaps were implemented almost a year ago when the [NIP-57 spec had this](https://github.com/nostr-protocol/nips/blob/8b39976e78f90fe766ad7149e250777cddacbb5e/57.md#protocol-flow):

> Client calculates a recipient's lnurl pay request url from the zap tag on the event being zapped (see Appendix G), or by decoding their lud06 or lud16 field on their profile according to the [lnurl specifications](https://github.com/lnurl/luds)

> When an event includes a zap tag, clients SHOULD calculate the lnurl pay request based on it's value instead of the profile's field.

> [ "zap", "pablo@f7z.io", "lud16" ]

staab:
> Wavlake zaps were implemented almost a year ago when the [NIP-57 spec had this](https://github.com/nostr-protocol/nips/blob/8b39976e78f90fe766ad7149e250777cddacbb5e/57.md#protocol-flow)

Very interesting, I didn't remember that. Maybe we should even bring this back, it seems useful for zapping people that don't have nostr accounts, which is what wavlake uses it for I believe. It probably shouldn't have been removed. But that's all orthogonal to this discussion anyhow.

xx.md
```diff
@@ -0,0 +1,46 @@
...
+```json
+{
+  "id": <event_id>,
+  "pubkey": <author_pubkey>,
+  "created_at": <created_at>,
+  "kind": 31337,
+  "content": "Chill beats",
+  "tags": [
+    ["d", "<UUID>"],
+    ["subject", "Platinum Robots on the Moon"],
```
pablof7z, line 41: Why not `title` to keep it consistent with NIP-23 and others?

xx.md
```diff
@@ -0,0 +1,46 @@
...
+- `d` is a unique identifier. This MAY be generated by the client creating the event, or refer to an external GUID.
+- `media` is the url of the media. This SHOULD NOT be read, but is required for backwards compatibility reasons.
+- `imeta` is as described in NIP 92. Multiple `imeta` tags MAY be included to provide for different use cases
+  (e.g. streaming vs download).
+- `subject` is the title of the audio track
+
+The following tags are optional:
+
+- `website` is an external url to a website related to the track.
+- `creator` is the pubkey (or plain text name) of the track author(s).
```
pablof7z, line 28: I think this should be indexable, probably a `p` tag and if it's not a pubkey `n` or something else; the use case of "more songs by this artist" is very normal

arthurfranca: please use "something else" instead of `n` cause the tag is being used to add events to unbounded lists - set to the name of the list - like in [this example](https://github.com/nostr-protocol/nips/pull/761/files#diff-55a569d18c00c9db28ae64db89efdfd5cb644b26d2a9333df750e3ca385e5452R41-R42).

In theory, any event kind, specially replaceable ones, would be able to use the `n` tag for adding events to these lists.

xx.md
staab: I'm trying to keep this compatible with zapstr.live which uses subject. Also, subject is defined in nip 14 so it's not a terrible choice.

vitorpamplona: IMO `title` and `subject` are different things. Title is generally a name. Subject is a very short "what's this about".

Songs have title, not subject.

staab: Yes, I agree, but there are lots of events in the wild already that use subject.

staab:
@pablof7z it seems 31337 is used both in NIP 89 and for zapstr.live. Any preference for how to handle that collision? I think probably creating a new kind for this nip would probably be preferable.

pablof7z:
> @pablof7z it seems 31337 is used both in NIP 89 and for zapstr.live. Any preference for how to handle that collision? I think probably creating a new kind for this nip would probably be preferable.

NIP 89 doesn't use it, it just mentions it as an example 😅

staab:
> NIP 89 doesn't use it, it just mentions it as an example 😅

Oh hey, I didn't read closely enough, that's perfect. Back to using 31337. The one area of incompatibility is the `p` tag, since zapstr uses an unusual version of it. I've added `creator` tag with a required `p` tag so plain text is still supported.

For everyone else's reference, here's an example 31337:

```json
{
  "id": "4688579eb15107da01fb2fcd43b4e72c32fd0d7f75673b543c8c1e7d0a3fcb74",
  "pubkey": "38b405e49614835c083ae103398c510315c3a0366cfa27115995e79e92451058",
  "created_at": 1693135050,
  "kind": 31337,
  "tags": [
    ["d", "wrh81br2ifcfq11un6ah8"],
    ["p", "38b405e49614835c083ae103398c510315c3a0366cfa27115995e79e92451058", "author"],
    ["c", "Electronic"],
    ["media", "https://nostr.build/av/b057855ba5c38a598575e231a82175801d97a0a013d90c6b5761c05a40fecdf4.mp3", "http"],
    ["cover", "https://i.ytimg.com/vi/MCe2pAIMAzg/maxresdefault.jpg"],
    ["subject", "PilotRedSun - useful"],
    ["zap", ""]
  ],
  "content": "PilotRedSun - useful\n\nhttps://zapstr.live/t/naddr1qq2hwung8qckyu3jd9nxxen3xych2m3kv95rsqsqqvzqqqr6dyntly6j\n\nhttps://i.ytimg.com/vi/MCe2pAIMAzg/maxresdefault.jpg",
  "sig": "e040023e0435b787a64ef9bddff1dd6d22808279ac1e05a090406484aace2734e81f5832df4051f3e53134248f78781fc3b7b40993935146913fa1a63fd39499"
}
```

xx.md
staab: I've updated this to use creator + p tag (creator so a mark can be added, p tag so that we have a standard format). This is slightly incompatible with zapstr's current p tags, but not enough to break anything (zapstr's role will be interpreted as a relay hint).

xx.md
```diff
@@ -0,0 +1,56 @@
...
+    ["d", "<UUID>"],
+    ["c", "Chill Beats"],
+    ["creator", "<pubkey>"],
+    ["creator", "<plain text>", "feat"],
+    ["title", "Platinum Robots on the Moon"],
+    ["subject", "Platinum Robots on the Moon"],
+    ["published_at", "<published_at>"],
+    ["imeta", "url https://example.com/my-track.m3u8 x <hash> ..."],
+    ["p", "<pubkey>", "wss:/example.com", "jim"]
+  ]
```
blastshielddown, line 54: Per NIP-94, I think it would be useful to include the `m` tag here to indicate the data type. This would make the types more explicit for clients and help prevent content playback from breaking. Per that spec: "The [MIME types](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types) format must be used, and they should be lowercase."

```suggestion
    ["p", "<pubkey>", "wss:/example.com", "jim"]
    ["m", "<MIME type>"]
  ]
```

The addition could also allow this kind to be used for a variety of media. Don't see why this should be limited to only audio.

staab: `imeta` includes `m`, and I think it's necessary to structure it that way because you might have multiple urls for a single track with different mime types.

blastshielddown: Ah I missed that reference, thanks. That looks great. Maybe adding a link to NIP-92 in the tag descriptions would be helpful?

xx.md
staab: I've updated this to use `title` but require `subject` for backwards compatibility with zapstr.

pablof7z: > I'm trying to keep this compatible with zapstr.live which uses subject. Also, subject is defined in nip 14 so it's not a terrible choice.

NIP-14 is why I used `subject` at the time 😅

xx.md
pablof7z: sorry about the missing relay hint

is the idea then that each creator will be tagged twice? once as a `creator` and once as a `p`?

staab: Right, so you can query by `p`, but you can still tag people who weren't creators using `p` tags. Without the `creator` tag the semantics of `p` becomes overloaded, so it's necessary to denote the person's role in contributing to the track (and support plaintext creator names, which wouldn't have an accompanying p tag).

xx.md
```diff
@@ -0,0 +1,56 @@
...
+- `d` is a unique identifier randomly generated by the client.
+- `media` is the url of the media. This SHOULD NOT be read, but is required for backwards compatibility reasons.
+- `imeta` is as described in [NIP 92](./92.md). Multiple `imeta` tags MAY be included to provide for different
+  use cases (e.g. streaming vs download).
+- `title` is the title of the audio track
+- `subject` is the title of the audio track (deprecated, but required for compatibility)
+
+The following tags are optional:
+
+- `i` is an external GUID.
```
blastshielddown, line 28: What is the distinction between this tag and the `d` tag? In other words, what are some examples of needing both?

blastshielddown:
Not sure what the best practice is for this kind of thing, but since this is intended to be a parameterized replaceable event: would it be helpful to add the `a` tag to the example so it's explicit what needs to be included? Or maybe there is a better way to reference this requirement without also having also update this doc if NIP-01 changes, that also could work.

From the [NIP-01](https://github.com/nostr-protocol/nips/blob/master/01.md#tags) spec:
`["a", <kind integer>:<32-bytes lowercase hex of a pubkey>:<d tag value>, <recommended relay URL, optional>]`

xx.md
staab: I originally had it in the `d` tag, but thought it might be useful to be able to change the GUID if a different standard is chosen, or include multiple guids (e.g. music vs podcast if a track qualifies as both).

staab:
> would it be helpful to add the a tag to the example so it's explicit what needs to be included?

Just so people know what it looks like to `a` tag the event? I don't think that's necessary, since `a` tags are pretty well defined and widely used. I did add a reference to NIP 01 though so people know where to look up the term. Let me know if you had something else in mind.

xx.md
joshr4: I recall there being an actual `["d", <value>]` (in the deprecated NIP-33), but I don't even see a reference to a `d` tag in NIP-01. 

Are `d` tags used at all? Looks like the `d` tag value only exists within the `a` tag, but not as an actual tag itself.

It is only mentioned here in NIP-01:

>The a tag, used to refer to a (maybe parameterized) replaceable event
for a parameterized replaceable event: `["a", <kind integer>:<32-bytes lowercase hex of a pubkey>:<d tag value>, <recommended relay URL, optional>]`

>for kind n such that 30000 <= n < 40000, events are parameterized replaceable, which means that, for each combination of pubkey, kind and the d tag's first value, only the latest event MUST be stored by relays, older versions MAY be discarded.

joshr4: so is an `i` tag basically the new `d` tag? Or does NIP-01 need to be updated to explicitly define the `d` tag?

staab: `i` and `d` might be the same in some cases, but emerging best practices are to make `d` tags random UUIDs with no semantic meaning. `i` is a reference to an external source. In this case, it might even be beneficial to have more than one (multiple `d` tags would result in undefined behavior).

blastshielddown: So if I understand you correctly, it sounds like the `d` tag is mainly used as a way to identify the content contained in the event, and the `i` tag(s) can be used to reference related content?

The way I see us using this is to put our internal content identifier guids in the `d` tag and then use the `i` tag(s) for any related content. For example, if a podcast contained some songs, the guid for the podcast would go in the `d` tag and the song guids in the `i` tags. At this time I don't see any reason to create and use another random identifier for `d`

blastshielddown: Now I wonder if `i` makes sense here. Based on NIP-39 and the `i` tag being used for identities, I don't think it makes sense to use this tag for arbitrary id's. I'm still not totally clear on the use case for `i`

As for referencing other tracks that this piece of content might contain, that sounds like something we could add later as an extension of this NIP. I can see the eventual benefit but trying to get there now might add unnecessary complexity to this proposal.

staab: This is a different context, so it's ok if the meaning of `i` varies a bit. The sematics of tags are defined by the kind the tag is applied to. We only have so many one-letter tags, better to re-use i than introduce a new one.

blastshielddown:
If you're up for it @staab we could publish some test events using this draft spec and see how it behaves in practice

staab:
Go for it! I'm hoping to hit this in the next week or two. If you want to use Ainsley's music as a pilot I think that would be awesome.

xx.md
```diff
@@ -0,0 +1,56 @@
...
+
+```json
+{
+  "id": <event_id>,
+  "pubkey": <author_pubkey>,
+  "created_at": <created_at>,
+  "kind": 31337,
+  "content": "Chill beats",
+  "tags": [
+    ["d", "<UUID>"],
```
fiatjaf, line 45:
```suggestion
    ["d", "<id>"],
```

The `id` can be anything, we shouldn't constrain it to being an UUID, which is a huge ass sequence of characters that is not necessary in 99.99% of the cases.

blastshielddown:
I've just published a test event of one of Ainsley's tracks according to this draft spec, event id: `1d8d62cf5f11d6cb6ce9944170753a0fdbf25ec26573ff68b2eecc761323f46d`. It should be readable from our relay (relay.wavlake.com)

Let me know how it works @staab , and if you have any questions or feedback.

pablof7z:
> I've just published a test event of one of Ainsley's tracks according to this draft spec, event id: `1d8d62cf5f11d6cb6ce9944170753a0fdbf25ec26573ff68b2eecc761323f46d`. It should be readable from our relay (relay.wavlake.com)
> 
> 
> 
> Let me know how it works @staab , and if you have any questions or feedback.

https://njump.me/nevent1qqspmrtzea03r4ktdn5egstsw5aqlkljtmpx2ulldzewanrkzv3lgmgprpmhxue69uhhyetvv9ujumn0wdmksetjv5hxxmmdqgs8wk0myn8v2m79w4g82n9g7mfvvqvrmgjn0j8nsyy0mu5rkg9qukqnkg2aj

Few questions:

What's the a tag in this event?
Looks like it's missing the backwards-compatible media tag
Is it possible to have Ainsley Costello's pubkey tagged or you don't have that data mapped locally?

staab:
Mostly repeating what @pablof7z said:

- It looks like the `a` tag is pointing to the same event, which seems nonsensical to me.
- I'd like to see at least `creator` (and maybe `zap` too) pointing to Ainsley if possible.
- `media` and `subject` tags should be included for compatibility with zapstr.live
- Instead of using `artist` it would probably be better to do `["creator", <ainsley-pubkey-if-possible-or-name>, "Artist"]` following the spec: "`creator` is the pubkey or plain text name of the track creator. This tag MAY have a mark denoting the creator's role."

blastshielddown:
Thanks @staab and @pablof7z Just pushed up a revised [event](https://njump.me/nevent1qqsf6dmnulk86vtpwy7pnq6d8mygwed5efgsgd3z9u6njzqe2rejwgqprpmhxue69uhhyetvv9ujumn0wdmksetjv5hxxmmdqgs8wk0myn8v2m79w4g82n9g7mfvvqvrmgjn0j8nsyy0mu5rkg9qukqg64q2v).

Per your questions/comments:
- The `a` tag follows the [spec](https://github.com/nostr-protocol/nips/blob/master/01.md#tags) for parameterized replaceable events. From what I can tell it's accurate but maybe I'm misinterpreting the spec
- I'm going to leave the `zap` tag as-is for now (this approach remains the best way for us to tally zaps for tracks) but will add Ainsley's pubkey to an additional `creator` tag. This should make it easy for a user to zap the artist directly if they want to. Please note: many of our artists do not have npubs so this should not be an expectation (yet)
- Added `media`, and `subject` tags while removing `artist`, these were oversights

Additionally, I think it might be useful to add a `version` tag. I've been thinking this for a little while now as I've seen these NIPs evolve and introduce breaking changes. This would at least give clients some reliable expectation in case any event kind were to change in some subtle or dramatic way. I added one in this latest test event as an example.

staab:
> The a tag follows the [spec](https://github.com/nostr-protocol/nips/blob/master/01.md#tags) for parameterized replaceable events. From what I can tell it's accurate but maybe I'm misinterpreting the spec

`a` tags are intended to point to a different event, not itself. So replies to this event would include the tag, but the event itself wouldn't.

> I'm going to leave the zap tag as-is for now

I can't argue with this, but be aware that there probably aren't any clients that support the current format.

> Additionally, I think it might be useful to add a version tag.

This is an interesting idea, but I'm inclined to think it's not a good solution. If we're going to support multiple formats, we should just create a new kind. Shifting specs are bad, and while I haven't paid much attention to it in the past I think we should be more careful about changing stuff in incompatible ways going forward.

staab:
Went to go work on this, and a couple other things:

- It looks like that example event is p-tagging its own author, which doesn't make much sense to me.
- There's currently no way to query the `creator` field if there's no corresponding pubkey, which seems important. Maybe we could use NIP 32 labels? I just drafted https://github.com/nostr-protocol/nips/pull/1129 which relaxes the requirement for `L` tags.

fmaul:
I couldn't find any docs nor any example events for the `30029` album lists. Can you point me to those?

For audio tracks that are released on albums or EPs I would prefer to have an optional album / track number tag. I propose:
```json
"tags": [
    ["album", "Purple Rain", "01"],    // Album title and Track Number (may contain a leading zeros) 
]
```
This would make it possible to recreate full albums for a `creator` (Artist) from only the `31337` events. Otherwise multiple tracks from an album wouldn't have any way to identify which album they belong to.

staab:
> I would prefer to have an optional album / track number tag

Tracks could show up on multiple albums, with or without the explicit authorization of the event publisher. 30029 lists for a given track could be easily queried with `#a`, so I still think that's the better approach.

@blastshielddown any updates?

staab:
I agree it's getting quite messy. The change you suggested still doesn't solve the problem of indexing plain-text `creator`s, and I'm still not sure about overloading `p`. Maybe we should go whole-hog on NIP 32?

```
["L", "media/creator"],
["l", <pubkey>, "media/creator"],
["l", <plaintext>, "media/creator"],
```

The challenge this leaves is deduplicating creators referred to both by pubkey and by plaintext. But since pubkeys have profiles, maybe we could just say tracks should prefer pubkeys, and not include duplicates. Thoughts?

blastshielddown:
Sorry I don't follow how you can't index plain-text values in the `creator` tag. Wouldn't this also be true for an `l` tag?

Using `l` tags like you described is better, I think. I'd like if there were some flexibility in the naming (to allow for things like singer, writer, producer, etc) but that might be better suited for the `meta` tag later anyway.

staab:
> Sorry I don't follow how you can't index plain-text values in the creator tag. Wouldn't this also be true for an l tag?

Only single-letter tags get indexed by most relay implementations.

> I'd like if there were some flexibility in the naming (to allow for things like singer, writer, producer, etc)

We could totally do that, maybe `nipxx/singer` or something like that. It would be best to include those in the NIP when they become desired. For now I included `artist` and `record_label` in the NIP.

blastshielddown:
Why not just start using `music` as the namespace for these `l` tags? `music/artist` and `music/label` seem like good starting points.

blastshielddown:
Latest [sample event](https://njump.me/nevent1qqs2rgwn6eed6sh6r3r4d8wfcy6l9r2565gydk2uqrfwpkck7092kccpzemhxue69uhhyetvv9ujumn0wd68ytnzv9hxgq3qwavlkfxwc4hu2a2sw4x23akjccqc8k39xly08qgglheg8vs2pevqvzmex9) per this thread.

staab:
@blastshielddown sorry, I just realized I'm falling into the NIP 32 trap of using labels for actual values. I've updated the NIP to overload the `c` tag similar to how `l` tags are overloaded, but as a simple key/value pair thing. This covers genre and subgenre, as well as any roles any number of people might fulfill. This is sort of a nuclear option, but the design space was feeling too constrained, and I think this supports everything we need, and is properly indexed.

MerryOscar:
@staab just a thought - if you added a `medium` tag these events would work for podcast episodes too - `["medium", "music / podcast / audiobook"]`

staab:
> @staab just a thought - if you added a `medium` tag these events would work for podcast episodes too - `["medium", "music / podcast / audiobook"]`

We could, but we should probably just use a different `kind` if we want that. If you think this nip would work for podcasts I can add that.

blastshielddown:
> I've updated the NIP to overload the `c` tag similar to how `l` tags are overloaded

I think this is pretty straightforward, even if it is a "nuclear option" as you put it. So my understanding is:
- the `c` tag can be used for any kind of labeling or categorization: artist name, genre, label, publisher... the tag is wide open to supporting any namespace we can think up (though we probably should standardize a few obvious ones early on)
- the `p` tag will reference any creators involved if there is a `npub` for that creator. Additionally the `p` tag can be overloaded with the creator's readable name.

Should we require that name in the `p` tag matches at least one entry in the set of `c` tags?

Also, thinking a bit ahead here about genres: it might be useful to collaborate on a set of common music genres that could be shared across applications. Not required for this spec but good to be aware of for future reference: https://github.com/wavlake/genre-list

staab:
> Additionally the p tag can be overloaded with the creator's readable name.

I think this is a bad idea, the text name is in the `c` tag so it's still searchable. `p` tags should always be pubkeys.

> Should we require that name in the p tag matches at least one entry in the set of c tags?

I don't think so, since there are probably valid use cases for mentioning someone more casually without labeling them with a `c` tag. Again, `p` tags should just be `p` tags.

Other than that, I think we're on the same page. I went ahead and included a note about genre with a link to wavlake's list. I'm currently working on adding a custom feeds feature to Coracle that will be used to reference these events, I'll report back when that's done. Let me know when you've got support for this as well.

blastshielddown:
The comment about overloading the `p` tag with the creator's readable name was just based off this example in the spec:

`["p", "2a07724d42fd8004b5c97b62ba03b6baf3919f9e8211667039987866997e97ad", "wss://my-relay.com", "AC/DC"]`

What does that last item represent?

blastshielddown:
Just checking in on this since it's been a while. Is there anything else you need from us to start testing this kind? We can start publishing a few more if that helps.

staab:
I'm happy with it. One small change after talking to @MerryOscar the other day, `i` tags MAY have a url hint as the second argument to redirect people to a website if the client isn't opinionated about how to interpret the guid. I think now we just wait for adoption. This is basically done on my end, since [custom feeds](https://wikifreedia.xyz/cip-01/97c70a44366a6535c1) allow users to create feeds of these events on their own.

blastshielddown:
We're ready to start cutting over to this new kind soon. Updated a new [test event](https://njump.me/nevent1qqsqmv2pz46zpy2pfarxun3rnyk0e80n0072rqd62x65nsavctsdgwgprpmhxue69uhhyetvv9ujumn0wdmksetjv5hxxmmdqgs8wk0myn8v2m79w4g82n9g7mfvvqvrmgjn0j8nsyy0mu5rkg9qukqtvqdha) for people to review, please let me know if you spot any issues.

staab:
Awesome, I've created a feed that pulls this event (really, anything by Ainsely published by wavlake). You can view it here: 

https://coracle.social/notes/naddr1qvzqqqrujgpzp978pfzrv6n9xhq5tvenl9e74pklmskh4xw6vxxyp3j8qkke3cezqyd8wumn8ghj7urewfsk66ty9enxjct5dfskvtnrdakj7qg4waehxw309aex2mrp0yhxgctdw4eju6t09uq3camnwvaz7tmgdajxccn0vshxxmmjv93kcefww3hk7mrn9uqsuamnwvaz7tmwdaejumr0dshsqyfn8yerwd3jxcenxdej8qmnjvfsxcpmdvgf

arkin0x:
@staab Thanks for this! We are going to be using it. We'll utilize kind `31338` for podcasts on fanfares.io but they will have some additional tags because our episodes are paywalled:

```javascript
tags: [
        ["price", "1234", "SATS"]
        ["encrypted", "ChaCha20"]
        // ... and an i tag linking to the podcast:guid:<guid> in addition to the i tag linking to the episode
    ]
```

arkin0x:
For anyone wondering, I was curious about kind 1808 and after searching on nostr.band found that it's used by stemstr.app for audio stems.

README.md
```diff
@@ -175,6 +175,7 @@ They exist to document what may be implemented by [Nostr](https://github.com/nos
 | `31990`       | Handler information        | [89](89.md)              |
 | `34550`       | Community Definition       | [72](72.md)              |
 | `39000-9`     | Group metadata events      | [29](29.md)              |
+| `31337`       | Audio Track                | [72](72.md)              |
```
prcolaco, line 178: | `31337`       | Music Track                | [72](72.md)              |
| `31338`       | Podcast Episode            | [72](72.md)              |

51.md
```diff
@@ -49,6 +49,7 @@ Aside from their main identifier, the `"d"` tag, sets can optionally have a `"ti
 | Curation sets | 30004 | groups of articles picked by users as interesting and/or belonging to the same category      | `"a"` (kind:30023 articles), `"e"` (kind:1 notes)                                 |
 | Curation sets | 30005 | groups of videos picked by users as interesting and/or belonging to the same category        | `"a"` (kind:34235 videos)                                                         |
 | Interest sets | 30015 | interest topics represented by a bunch of "hashtags"                                         | `"t"` (hashtags)                                                                  |
+| Audio Playlist| 30029 | a list of kind `31337` audio tracks                                                          | `"a"` (kind:31337 audio tracks)                                                   |
```
prcolaco, line 52: Maybe here add also a list kind for Podcasts (episodes with kind:31338), or maybe call this `Audio Collection` and it could be applicable to both and maybe more kinds in the future? Because an Album/Release could also be applicable... also support `title` tag and `imeta` for enclosure/cover art, and maybe description in the `.content`...

prcolaco (changes requested):
love the simplicity of this proposal, just a few comments...

staab:
I split tracks into music vs podcast, @MerryOscar @arkin0x let me know if that messes anything up for you.

0a.md
```diff
@@ -0,0 +1,66 @@
...
+- `subject` is the title of the audio track (deprecated, but required for compatibility)
+
+The following tags are optional:
+
+- `i` is an external GUID in the format defined [here](https://github.com/MerryOscar/nips/commit/280eb498e0ac56b8f9356c1b7a88cc8b31579801).
+- `c` is the track's value as defined by a given category.
+- `website` is an external url to a website related to the track.
+- `duration` is the duration of the audio track, in seconds.
+- `published_at` is a timestamp representing the track's original publish date.
+
```
arkin0x, line 36:
```suggestion
- `price` is the value + UNIT, ie `[ "price", "1000", "SATS" ]
```

staab: This seems like it belongs somewhere else, since it may not be unique to audio events, and requires a lot of context to be used properly. Maybe draft up a new NIP that describes what fanfares is doing?

arkin0x: Sure, I just copied NIP-99 but we don't need to include it here.

staab: Yeah, I think there's more complexity involved with selling content vs other stuff, since I assume the content would be paywalled, and there would need to be a mechanism for unlocking it.

fiatjaf (approved):

51.md
```diff
@@ -43,16 +43,18 @@ For example, _relay sets_ can be displayed in a dropdown UI to give users the op
...
+| Curation sets         | 30005 | groups of videos picked by users as interesting and/or belonging to the same category        | `"a"` (kind:34235 videos)                                                         |
+| Interest sets         | 30015 | interest topics represented by a bunch of "hashtags"                                         | `"t"` (hashtags)                                                                  |
+| Emoji sets            | 30030 | categorized emoji groups                                                                     | `"emoji"` (see [NIP-30](30.md))                                                   |
+| Music Playlist        | 30037 | a list of kind `31337` music tracks                                                          | `"a"` (kind:31337 music tracks)                                                   |
+| Podcast Playlist      | 30038 | a list of kind `31338` podcast episodes                                                      | `"a"` (kind:31333 podcast episodes)                                               |
```
prcolaco, line 56: `kind:31338` instead of `kind:31333` in last column

README.md
```diff
@@ -195,6 +196,8 @@ They exist to document what may be implemented by [Nostr](https://github.com/nos
 | `34237`       | Video View Event           | [71](71.md)              |
 | `34550`       | Community Definition       | [72](72.md)              |
 | `39000-9`     | Group metadata events      | [29](29.md)              |
+| `31337`       | Music Track                | [xx](xx.md)              |
+| `31338`       | Podcast Episode            | [xx](xx.md)              |
```
prcolaco, line 200: change the NIP number and file name in the last column of lines 199 and 200

prcolaco (changes requested):

ebrakke:
Might be worth updating [NIP-73](https://github.com/nostr-protocol/nips/blob/master/73.md) with some of the external formats for audio tracks?
e.g. `mbid:release:a1812b30-e3ea-4ea7-b7bd-f2f3bfdc08f0` for a music brainz identifier, or `spotify:track:2liDEMyPVzDDLXKUnyiznA` to find on spotify? 
Unsure about the latter, since that would potentially require figuring out the external identifier for many different streaming platforms, but having the format for music brainz would be nice.

staab:
Fewer formats is probably better, especially if they're proprietary ids that are platform-specific (this is partly why the url hint component of nip 73 tags was added). But music brainz seems reasonable to add — would you be willing to open a separate PR for that?

Also, @blastshielddown @MerryOscar is this in production and ready for merge?

0a.md
```diff
@@ -0,0 +1,66 @@
+NIP-XX
```
prcolaco, line 1: also the title needs to be changed to `NIP-0A`...
great work guys! 👍

0a.md
```diff
@@ -0,0 +1,66 @@
...
+    ["c", "Pop", "genre"],
+    ["c", "EDM", "subgenre"],
+    ["c", "Columbia Records", "record_label"],
+    ["c", "AC/DC", "artist"],
+    ["i", "podcast:item:guid:123", "https://fountain.fm/episode/30uEXC25615Ze2ELjY2p"],
+    ["p", "2a07724d42fd8004b5c97b62ba03b6baf3919f9e8211667039987866997e97ad", "wss://my-relay.com", "AC/DC"],
+    ["title", "Platinum Robots on the Moon"],
+    ["subject", "Platinum Robots on the Moon"],
+    ["published_at", "<published_at>"],
+    ["imeta", "url https://example.com/my-track.m3u8", "x <hash>", ...]
```
arkin0x, line 63: @staab I'd like to include the podcast artwork as well as the individual episode artwork. What would be an example of how to include track artwork and podcast artwork and distinguish meaningfully between them so the client knows which is which? 

For example, when looking up a podcast episode via Podcastindex API, they provide an image for the episode and the podcast image as well. Sometimes they are the same if the podcast does not have unique artwork per episode, but they can be different. 

Would we differentiate this using summary or alt from NIP-94? (Seems a little messy)

arkin0x: For that matter, how do we indicate streaming vs download using an imeta tag?

> Multiple `imeta` tags MAY be included to provide for different use cases (e.g. streaming vs download).

staab: If necessary, we could add an attribute to imeta, something like `content disposition`, e.g. `["imeta", "url https://example.com/cover.png", "x <hash>", "disposition cover"]`. For streaming vs download we could do the same thing, but in the near term I think it'll be clear enough which url to use for what, m3u8 is suitable for streaming, mp4 for download for example.

arkin0x: That sounds good to me

0a.md
```diff
@@ -0,0 +1,66 @@
...
+- `media` is the url of the media. This SHOULD NOT be read, but is required for backwards compatibility reasons.
+- `imeta` is as described in [NIP 92](./92.md). Multiple `imeta` tags MAY be included to provide for different
+  use cases (e.g. streaming vs download).
+- `title` is the title of the audio track
+- `subject` is the title of the audio track (deprecated, but required for compatibility)
+
+The following tags are optional:
+
+- `i` is an external GUID in the format defined [here](https://github.com/MerryOscar/nips/commit/280eb498e0ac56b8f9356c1b7a88cc8b31579801).
+- `c` is the track's value as defined by a given category.
```
arkin0x, line 32: This seems too vague. Also, the category should definitely be the mark:

["c", "genre", "Pop"]

not

["c", "Pop", "genre"] <<< this is backwards compared to most other tags :(

arkin0x: - genre
- subgenre
- artist
- record_label
- producer

what other fields should be expected?

- composer
- album
- track_number
- tracks
- bpm
- year

arkin0x: https://soundcharts.com/blog/music-metadata might provide helpful insight on doing this right

staab: > ["c", "Pop", "genre"] <<< this is backwards compared to most other tags :(

This looks backward, but like with NIP 32 the intention is to be able to index the value, rather than the key. It's more important to be able to filter by category values (e.g. `#c: ["Pop"]`) than to filter by what categories are defined.

Also added a list of categories to the NIP.

arkin0x: > It's more important to be able to filter by category values (e.g. #c: ["Pop"]) than to filter by what categories are defined.

That's a good point. Too bad this makes the event.getMatchingTags function in NDK totally useless for parsing these audio events, but I can implement my own tag parsing.

Thanks for adding those categories.

arkin0x:
I'd like to see more clarification around the imeta disposition and other podcasting 2.0 features that are missing from this spec, such as defining timestamps and artwork changing based on those timestamps. I can work on it but I wanted to see what others thought. In it's current form, this NIP doesn't really provide a full equivalent for RSS-based podcasts.

staab:
I think we should merge this, then add additional features in additional NIPs. This is a good core feature set, and is far short of replacing podcasting 2.0, which I'm not sure is quite the same project.

staab:
@blastshielddown @MerryOscar just waiting on a final review from you guys.

0a.md
```diff
@@ -0,0 +1,80 @@
...
+rather than in a general micro-blogging context.
+
+The following kinds are defined by this NIP:
+
+- `31337` represents a music track
+- `31338` represents a podcast episode
+
+The following tags are required:
+
+- `d` is a unique identifier randomly generated by the client.
```
joshr4, line 22: So the publishing client will randomly generate the `d` tag? Can the GUID in the optional `i` tag be used here as the `d` tag value?

0a.md
```diff
@@ -0,0 +1,80 @@
+NIP-0a
+======
+
+Audio Events
+---------------
+
+`draft` `optional`
+
+This NIP defines several new event kinds representing different types of audio track. These events
+are _parameterized replaceable_ as defined in [NIP-01](./01.md) and deletable per [NIP-09](09.md).
```
joshr4, line 10: Being replaceable events, I assume there is no way to change the pubkey (author) of future events, since replaceable events are dependent on the original event's pubkey. 

I'm thinking through the problematic scenario of publishing audio tracks under a general Music Feed pubkey, and later supporting an Artist's ability to self-publish their audio tracks using their own pubkey, if/when they are onboarded to nostr and take custody of their own private key.

I suppose the older audio track event can be deleted when this happens, leaving the new Artist's self-published event? In this case, should the `d` tag be re-used or randomly re-generated?

0a.md
```diff
@@ -0,0 +1,80 @@
...
+- `record_label`
+- `producer`
+- `composer`
+- `album`
+- `track_number`
+- `tracks`
+- `bpm`
+- `year`
+
+Note that artists, producers, SHOULD be referred to using both a plaintext `c` tag AND a conventional `p` tag when pubkey is available. When doing so, the `p` tag's petname MUST be the same as the `c` tag's value. In this way `c` tags can be used to indicate genre, as well as artist, producer, etc.
```
joshr4, line 51: What are the consequences of a `p` tag's petname not matching a `c` tag's value?

If a pubkey updates their petname, must these audio tracks be replaced with an updated version to reflect the new petname to avoid a mismatch of `p` and `c` tags?

0a.md
staab: Yes, it should be random. Meaningful d-tags are an antipattern, since you can't really anticipate when something might change, even guids.

0a.md
staab: Correct, in that scenario you'd probably want to just delete the events and publish new ones. D tag persistence shouldn't matter because the pubkey would be changing anyhow.

0a.md
staab: Petnames are arbitrary, and aren't assigned by the pubkey owner. If they don't match, then clients would have to treat the `c` tag as an arbitrary string and the `p` tag as a normal mention.

sondreb:
Nostria has implemented support for kind 32100, which is a good specification for playlists. It relies on .m3u and .m3u8 text formats. These have existed for a very long time and can more easily be adopted and used in third party apps and services.

User's can even download (export) as .m3u files directly from these events.

It also means apps such as Kodi, could potentially implement support for Nostr to discover and open playlists.

I don't like the idea of implementing a custom playlist structure for Nostr.

[NIP-XX M3U Playlists over Nostr](https://github.com/nostr-protocol/nips/issues/1945)

https://en.wikipedia.org/wiki/M3U

maml:
Implementer datapoint: lightning.fm publishes kind 31337 tracks in production (music platform on Lightning + Nostr; catalog on an open-read relay plus public relays).

Because this PR and the community registry definition disagree about required tags, we emit a superset that both dialects parse: title and subject carry the same value, url and media carry the same value, and we include both the discrete url/x/m/size tags and the imeta form. The full tag contract we ship is documented at https://lightning.fm/interop.

From the publisher side, convergence on either dialect would be welcome; emitting both works but makes events noisier than they need to be. We will adjust our publisher to whatever this PR lands on, and we are glad to share anything useful from production usage.

staab:
I updated the NIP text to include most of what's on your interop guide, except for:

- `type` - the event kind should be sufficient to indicate that.
- `lightning_node_id` - is this for keysend? I figure zaps or some other keysend spec should be used in combination with zap split tags.
- The metadata tags I have kept as `c` tags (though I don't know if anyone actually uses them)
- I omitted `lyrics`, `explicit`, and `credits` since I don't know what the format would actually be.

At this point I think people should just abandon these kinds and pick something else. Pretty much every implementation is different, which makes interop impossible. This PR was an attempt to standardize and solve some problems (e.g. indexing), but I don't think anyone has really adopted it, perpetuating the n+1 standards problem.

On second thought, I'm going to close this and let someone else propose a format they actually implement (it's been years since I added support for audio stuff).

PR #1044, opened 2024-02-13
Bring back non-pubkey zap tags

staab:
[Last May](https://github.com/nostr-protocol/nips/pull/552#issuecomment-1561287442) we removed lightning address support from zap splits, which [broke Wavlake's implementation](https://github.com/nostr-protocol/nips/pull/1043#issuecomment-1940261772). Should we restore this behavior?

@jb55 @vitorpamplona @pablof7z

### DIFF
--- a/57.md
+++ b/57.md
@@ -169,14 +169,21 @@ A client can retrieve `zap receipt`s on events and pubkeys using a NIP-01 filter
 
 ### Appendix G: `zap` tag on other events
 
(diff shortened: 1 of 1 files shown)


staab:
@joshr4 @blastshielddown what are your thoughts on this? Would it greatly matter to you to bring back zap splits so that clients could start to support wavlake's zap tags, or would you be willing to migrate to the new version?

alexgleason:
Related: https://github.com/benthecarman/zapple-pay-backend/issues/11

It would make zap splits easier. But I don't think it's actually a good idea, but I don't think we should support two different ways.

vitorpamplona:
The main issue with the old way is that Clients can't identify who is the owner of the lnaddress is to set expectations correctly and get the zapping user approval before moving forward. 

In kind 1s, users either had to manually explain which lnaddress was owned by who in the `.content` or users simply wouldn't zap it. And when they did without knowing it was a split, many users got mad. 

Since the change, I have had zero issues of this type.

joshr4:
Yeah we can migrate easily enough, especially since we don't depend on that original split functionality today.

We're definitely interested in any current/future split conversations as they develop, thats one reason we had chosen that old way at the time

staab:
Cool, I'll close this then.

PR #1046, opened 2024-02-13
Minor NIP-42 Example Update

arthurfranca:
@fiatjaf I would like to make it clearer that the `CLOSED` message refers to a subscription as a whole.

So in the specifc AUTH flow case, client would retry the request with exactly the same filter after authenticating.

### DIFF
--- a/42.md
+++ b/42.md
@@ -66,11 +66,11 @@ Given that a relay is likely to require clients to perform authentication only f
 
 ```
 relay: ["AUTH", "<challenge>"]
-client: ["REQ", "sub_1", {"kinds": [4]}]
+client: ["REQ", "sub_1", {"kinds": [4, 1]}]
 relay: ["CLOSED", "sub_1", "auth-required: we can't serve DMs to unauthenticated users"]
 client: ["AUTH", {"id": "abcdef...", ...}]
 relay: ["OK", "abcdef...", true, ""]
-client: ["REQ", "sub_1", {"kinds": [4]}]
+client: ["REQ", "sub_1", {"kinds": [4, 1]}]
 relay: ["EVENT", "sub_1", {...}]
 relay: ["EVENT", "sub_1", {...}]
 relay: ["EVENT", "sub_1", {...}]


vitorpamplona:
I think this is the wrong way of solving this. The relay should send the Kind 1s back to the client and not send the kind 4s if they need authentication instead of just closing the entire subscription.

vitorpamplona:
> ["CLOSED", "sub_1", "auth-required: anything machines won't read"] you can't be sure about what events or rule triggered the AUTH request, just which subscription triggered it.

I don't need to know which ones triggered. I only have one option: authenticate with the logged-in user or not. If the user approves, that's what I send. If it is not enough to satisfy the server's auth needs then the server should just send what it can send.

arthurfranca:
Even if the relay sends some events then EOSE like you want, it is supposed to later send `CLOSED:auth_required` and then even if you authenticate, the sub will be already processed, you would have to send a new `REQ`.

But then the client wouldn't know how to alter the filter correctly to ask relays just for events that it haven't already received.

So it doesn't make sense to receive just some events. It is better to close sub right away, then the client can simply retry with the same filter after authenticating.

vitorpamplona:
Yep. I now think it doesn't make sense to CLOSE the subscription at all. Just keep the client subscribed to the events it has access to, whatever those may be.

fiatjaf:
I think it's better if relays just close everything. Clients should send different REQs for different purposes, so they can be closed in a more granular way.

vitorpamplona:
> I think it's better if relays just close everything. Clients should send different REQs for different purposes, so they can be closed in a more granular way.

Well, I just can't. The majority of relays limit clients to 10 subscriptions and 10 filters in each subscription. Amethyst alone needs a minimum of 40. So, I have to do a lot of work to get as much into a single filter as possible. 

But the worst part is that we never know if adding a kind to a current filter will CLOSE the subscription or not.

arthurfranca:
> Just keep the client subscribed to the events it has access to, whatever those may be

Doesn't work very well. For example, using the `{ kinds: [1, 4] }` case, consider relay is sending you just stored public events (kind 1) and you AUTH in the middle of it, what do you want to happen?

You can't expect relay to buffer private events (kind 4) it may have gotten from DB while you wasn't authenticated nor would it be reasonable to instead expect the relay to re-run the DB query. It messes with ordering too if you add `limit` to the equation.

So your authentication would probably just change the live events your receive after EOSE (it would include kind 4 events). Not a good flow.

vitorpamplona:
That's ok. We already repeat the filter after the AUTH has been confirmed anyway. Most relays don't wait for the AUTH and dont buffer the events like you said. After sending the AUTH, the relay never redoes the subscription. You have to manually register the filter again to receive the stores events.

arthurfranca:
> the worst part is that we never know if adding a kind to a current filter will CLOSE the subscription or not.

That's a good reason for a relay to be defensive and close the sub from its side right away if it sees a filter that it considers too broad, for example filters without `kinds` set.

I admit though that considering `e` tags don't include event kind info (just id), it may be too much to ask from clients to always add `kinds` to the filter. Although clients can probably infer most of the times from context which kinds are involved and do `{ ids: [...], kinds: [...] }` to give a hint to relays without requiring them to touch DB.

arthurfranca:
> We already repeat the filter after the AUTH has been confirmed anyway

This stays the same. The thing is before CLOSED existed, you couldn't know if a sub required AUTH to be fullfilled (relay may always send AUTH on connection start).

Now client code can do like "hey I'm requesting  DMs so if this sub requires AUTH I will authenticate and repeat the filter" or "I'm requesting just public events and this is not my paid relay so if this sub requires AUTH, I will give up"

PR #1047, opened 2024-02-14
New version of NIP-46

erskingardner:
I recently implemented the current version of NIP-46 in [nostr-ignition](https://github.com/erskingardner/nostr-ignition) and found it very painful. After talking to several other folks that have also implemented this NIP and found it painful, I decided to try and improve the NIP to make it more likely that remote signing is adopted by more of Nostr.

My proposal changes NIP-46 significantly based on a few principles. 
* Use separate event kinds instead of JSON-RPC style payloads with methods.
* Simplify payloads as much as possible, only using objects or arrays where absolutely necessary.
* Use NIP-44 encryption for payloads instead of NIP-04

I've talked to quite a few folks over the last few weeks on this so thanks for all the input @fiatjaf, @pablof7z, @staab, @nostrband, and others. 

I have a shed, it is ready for your bikes. 🚴‍♂️

### DIFF
--- a/46.md
+++ b/46.md
@@ -8,8 +8,9 @@ This NIP describes a method for 2-way communication between a remote signer and 
 
 ## Terminology
 
-- **Local keypair**: A local public and private key-pair used to encrypt content and communicate with the remote signer.
-- **Remote pubkey**: The public key that the user wants to sign as. The remote signer has control of the private key that matches this public key.
+- **Local keypair**: A local public and private key-pair used to encrypt content and communicate with the remote signer. Usually created by the client application.
+- **Remote user pubkey**: The public key that the user wants to sign as. The remote signer has control of the private key that matches this public key.
+- **Remote signer pubkey**: This is the public key of the remote signer itself. This is needed in both `create_account` command because you don't yet have a remote user pubkey.
 
 All pubkeys specified in this NIP are in hex format.
 
@@ -41,9 +42,9 @@ nostrconnect://<local-keypair-pubkey>?relay=<wss://relay-to-connect-on>&metadata
 
 In this last case, most often used to fascilitate an OAuth-like signin flow, the client first looks for remote signers that have announced themselves via NIP-89 application handler events.
 
-First the client will query for `kind: 31989` events that have a `k` tag of `24133`.
+First the client will query for `kind: 31990` events that have a `k` tag of `24133`.
 
-These are generally shown to a user, and once the user selects which remote signer to use and provides the remote pubkey they want to use (via npub, pubkey, or nip-05 value), the client can initiate a connection. Note that it's on the user to select the remote signer that is actually managing the remote key that they would like to use in this case. If the remote pubkey is managed on another remote signer, the connection will fail.
+These are generally shown to a user, and once the user selects which remote signer to use and provides the remote user pubkey they want to use (via npub, pubkey, or nip-05 value), the client can initiate a connection. Note that it's on the user to select the remote signer that is actually managing the remote key that they would like to use in this case. If the remote user pubkey is managed on another remote signer, the connection will fail.
 
 In addition, it's important that clients validate that the pubkey of the announced remote signer matches the pubkey of the `_` entry in the `/.well-known/nostr.json` file of the remote signer's announced domain.
 
@@ -55,7 +56,7 @@ Clients that allow users to create new accounts should also consider validating 
 
 1. Clients need several things to start: a local keypair and the details of the pubkey and remote signer the user would like to use.
    - The local keypair is roughly disposable. E.g. - it's ok to store it locally (ideally encrypted) for some period of time. Once authenticated, remote signers can choose to trust this local keypair for a period of time. This local keypair can be deleted/rotated at will by the client and will only trigger a new auth challenge or permissions on the remote signer.
-   - The remote pubkey is the pubkey that the user is trying to sign in / sign as. This can be fetched with an NIP-05 or entered directly by the user (e.g. by passing a `bunker://` token).
+   - The remote user pubkey is the pubkey that the user is trying to sign in / sign as. This can be fetched with an NIP-05 or entered directly by the user (e.g. by passing a `bunker://` token).
 2. Clients use the local keypair for encrypting request events that they send to remote signers.
 
 ### Remote signer-side
@@ -79,7 +80,7 @@ Coming soon...
     "kind": 24133,
     "pubkey": <local_keypair_pubkey>,
     "content": <nip04(<request>)>,
-    "tags": [["p", <remote pubkey>]],
+    "tags": [["p", <remote_user_pubkey>]], // NB: in the `create_account` event, the remote signer pubkey should be `p` tagged.
     "created_at": <unix timestamp in seconds>,
 }
 ```
@@ -104,8 +105,8 @@ Each of the following are methods that the client sends to the remote signer.
 
 | Command                  | Params                                            | Result                                                                 |
 | ------------------------ | ------------------------------------------------- | ---------------------------------------------------------------------- |
-| `create_account`         | `[<username>, <domain>, <optional_email>]`        | `<newly_created_remote_pubkey>`                                        |
-| `connect`                | `[<remote_pubkey>, <optional_secret>]`            | "ack"                                                                  |
+| `create_account`         | `[<username>, <domain>, <optional_email>]`        | `<newly_created_remote_user_pubkey>`                                   |
+| `connect`                | `[<remote_user_pubkey>, <optional_secret>]`       | "ack"                                                                  |
 | `sign_event`             | `[<json_stringified_event_to_sign>]`              | `json_stringified(<signed_event>)`                                     |
 | `ping`                   | `[]`                                              | "pong"                                                                 |
 | `get_relays`             | `[]`                                              | `json_stringified({<relay_url>: {read: <boolean>, write: <boolean>}})` |
@@ -124,7 +125,7 @@ Each of the following are methods that the client sends to the remote signer.
     "kind": 24133,
     "pubkey": <remote_signer_pubkey>,
     "content": <nip04(<response>)>,
-    "tags": [["p", <local keypair pubkey>]],
+    "tags": [["p", <local_keypair_pubkey>]],
     "created_at": <unix timestamp in seconds>,
 }
 ```
--- a/46.md
+++ b/46.md
@@ -69,7 +69,7 @@ nostrconnect://<local-keypair-pubkey>?relay=<wss://relay-to-connect-on>&metadata
 }
 ```
 
-#### Response
+#### Response event
 
 ```json
 {
@@ -82,6 +82,10 @@ nostrconnect://<local-keypair-pubkey>?relay=<wss://relay-to-connect-on>&metadata
     "tags": [["p", "eff37350d839ce3707332348af4549a96051bd695d3223af4aabce4993531d86"]], // p-tags the local keypair pubkey
 }
 ```
+
+#### Diagram
+
+![signing-example](https://i.nostr.build/P3gW.png)
 
 ## Request Events `kind: 24133`
 
@@ -167,6 +171,10 @@ An Auth Challenge is a response that a remote signer can send back when it needs
 ```
 
 Clients should display (in a popup or new tab) the URL from the `error` field and then subscribe/listen for another response from the remote signer (reusing the same request ID). This event will be sent once the user authenticates in the other window (or will never arrive if the user doesn't authenticate). It's also possible to add a `redirect_uri` url parameter to the auth_url, which is helpful in situations when a client cannot open a new window or tab to display the auth challenge.
+
+#### Example event signing request with auth challenge
+
+![signing-example-with-auth-challenge](https://i.nostr.build/W3aj.png)
 
 ## Remote Signer Commands
 
--- a/46.md
+++ b/46.md
@@ -15,6 +15,8 @@ This also allows us to incorporate [NIP-44](https://github.com/nostr-protocol/ni
 - **Local keypair**: A local public and private key-pair used to encrypt content and communicate with the remote signer.
 - **Remote pubkey**: The public key that the user wants to sign as. The remote signer has control of the private key that matches this public key.
 
+All pubkeys specified in this NIP are in hex format.
+
 ## Signer Discovery
 
 To initiate a connection between a client and a remote signer there are a few different options.
@@ -26,7 +28,7 @@ This is most common in a situation where you have your own nsecbunker or other t
 The remote signer would provide a connection token in the form:
 
 ```
-bunker://<hex-pubkey-you-want-to-sign-as>?relay=<wss://relay-to-connect-on>&relay=<wss://another-relay-to-connect-on>&secret=<optional-secret-value>
+bunker://<remote-pubkey>?relay=<wss://relay-to-connect-on>&relay=<wss://another-relay-to-connect-on>&secret=<optional-secret-value>
 ```
 
 This token is pasted into the client by the user and the client then uses the details to connect to the remote signer via the specified relay(s).
@@ -36,7 +38,7 @@ This token is pasted into the client by the user and the client then uses the de
 In this case, basically the opposite direction of the first case, the client provides a connection token (or encodes the token in a QR code) and the signer initiates a connection to the client via the specified relay(s).
 
 ```
-nostrconnect://<client-key-hex>?relay=<wss://relay-to-connect-on>&metadata=<json metadata in the form: {"name":"...", "url": "...", "description": "..."}>
+nostrconnect://<local-keypair-pubkey>?relay=<wss://relay-to-connect-on>&metadata=<json metadata in the form: {"name":"...", "url": "...", "description": "..."}>
 ```
 
 ### Remote signer discovery via NIP-89
@@ -83,7 +85,7 @@ Clients can subscribe to events of `kind: 24136`, `kind: 24137`, and `kind: 2413
 {
     "id": <id>,
     "kind": <kind for command>, // Each command has it's own kind
-    "pubkey": <local pubkey of the user>,
+    "pubkey": <local keypair pubkey>,
     "content": <nip44(<request>)>,
     "tags": [["p", <remote pubkey>]],
     "created_at": <unix timestamp in seconds>,
@@ -97,11 +99,11 @@ Clients can subscribe to events of `kind: 24136`, `kind: 24137`, and `kind: 2413
 | `24142` | `sign_event`             | nostr event to sign                                                             | JSON stringified object |
 | `24143` | `ping`                   | -                                                                               | null                    |
 | `24144` | `get_relays`             | -                                                                               | null                    |
-| `24145` | `nip04_encrypt`          | {third_party_pubkey: <third_party_pubkey>, content: <plaintext to encrypt>}     | JSON stringified object |
-| `24146` | `nip04_decrypt`          | {third_party_pubkey: <third_party_pubkey>, content: <ciphertext to decrypt>}    | JSON stringified object |
-| `24147` | `nip44_conversation_key` | third_party_pubkey                                                              | string                  |
-| `24148` | `nip44_encrypt`          | {third_party_pubkey: <third_party_pubkey>, content: <plaintext to encrypt>}     | JSON stringified object |
-| `24149` | `nip44_decrypt`          | {third_party_pubkey: <third_party_pubkey>, content: <ciphertext to decrypt>}    | JSON stringified object |
+| `24145` | `nip04_encrypt`          | {pubkey: <third_party_pubkey>, content: <plaintext to encrypt>}                 | JSON stringified object |
+| `24146` | `nip04_decrypt`          | {pubkey: <third_party_pubkey>, content: <ciphertext to decrypt>}                | JSON stringified object |
+| `24147` | `nip44_conversation_key` | <third_party_pubkey>                                                            | string                  |
+| `24148` | `nip44_encrypt`          | {pubkey: <third_party_pubkey>, content: <plaintext to encrypt>}                 | JSON stringified object |
+| `24149` | `nip44_decrypt`          | {pubkey: <third_party_pubkey>, content: <ciphertext to decrypt>}                | JSON stringified object |
 
 ## Response Event `kind: 24136`
 
--- a/46.md
+++ b/46.md
@@ -54,14 +54,14 @@ Clients that allow users to create new accounts should also consider validating 
 
 ### Client-side
 
-1. Clients need several things to start: a local keypair and the details of the pubkey and remote signer the user would like to use.
+1. Clients need several things to start: a local keypair and the details of the pubkey and relays used by the remote signer the user would like to use.
    - The local keypair is roughly disposable. E.g. - it's ok to store it locally (ideally encrypted) for some period of time. Once authenticated, remote signers can choose to trust this local keypair for a period of time. This local keypair can be deleted/rotated at will by the client and will only trigger a new auth challenge or permissions on the remote signer.
    - The remote user pubkey is the pubkey that the user is trying to sign in / sign as. This can be fetched with an NIP-05 or entered directly by the user (e.g. by passing a `bunker://` token).
 2. Clients use the local keypair for encrypting request events that they send to remote signers.
 
 ### Remote signer-side
 
-1. Remote signers manage the private key of the user that will be used for signing. There are many different remote signers out there (e.g. Nsecbunker, Amber, etc.) that operate in different ways and have different features.
+1. Remote signers manage the private key of the user that will be used for signing. There are many different remote signers out there (e.g. Nsecbunker, nsec.app, etc.) that operate in different ways and have different features.
 2. Remote signers wait for request events and then act.
 
 ### Example flow for signing an event
@@ -103,19 +103,19 @@ The `content` field is a JSON-RPC-like message that is [NIP-04](https://github.c
 
 Each of the following are methods that the client sends to the remote signer.
 
-| Command                  | Params                                            | Result                                                                 |
-| ------------------------ | ------------------------------------------------- | ---------------------------------------------------------------------- |
-| `create_account`         | `[<username>, <domain>, <optional_email>]`        | `<newly_created_remote_user_pubkey>`                                   |
-| `connect`                | `[<remote_user_pubkey>, <optional_secret>]`       | "ack"                                                                  |
-| `sign_event`             | `[<json_stringified_event_to_sign>]`              | `json_stringified(<signed_event>)`                                     |
-| `ping`                   | `[]`                                              | "pong"                                                                 |
-| `get_relays`             | `[]`                                              | `json_stringified({<relay_url>: {read: <boolean>, write: <boolean>}})` |
-| `get_public_key`         | `[]`                                              | `<hex-pubkey>`                                                         |
-| `nip04_encrypt`          | `[<third_party_pubkey>, <plaintext_to_encrypt>]`  | `<nip04_ciphertext>`                                                   |
-| `nip04_decrypt`          | `[<third_party_pubkey>, <ciphertext_to_decrypt>]` | `<plaintext>`                                                          |
-| `nip44_conversation_key` | `[<third_party_pubkey>]`                          | `<nip44_conversation_key>`                                             |
-| `nip44_encrypt`          | `[<third_party_pubkey>, <plaintext_to_encrypt>]`  | `<nip44_ciphertext>`                                                   |
-| `nip44_decrypt`          | `[<third_party_pubkey>, <ciphertext_to_decrypt>]` | `<plaintext>`                                                          |
+| Command                  | Params                                              | Result                                                                 |
+| ------------------------ | --------------------------------------------------- | ---------------------------------------------------------------------- |
+| `create_account`         | `[<username>, <domain>, <optional_email>]`          | `<newly_created_remote_user_pubkey>`                                   |
+| `connect`                | `[<remote_user_pubkey>, <optional_secret>]`         | "ack"                                                                  |
+| `sign_event`             | `[<json_stringified_event_to_sign>]`                | `json_stringified(<signed_event>)`                                     |
+| `ping`                   | `[]`                                                | "pong"                                                                 |
+| `get_relays`             | `[]`                                                | `json_stringified({<relay_url>: {read: <boolean>, write: <boolean>}})` |
+| `get_public_key`         | `[]`                                                | `<hex-pubkey>`                                                         |
+| `nip04_encrypt`          | `[<third_party_pubkey>, [<plaintext_to_encrypt>]]`  | `array of <nip04_ciphertext> results of encryption`                    |
+| `nip04_decrypt`          | `[<third_party_pubkey>, [<ciphertext_to_decrypt>]]` | `array of <plaintext> results of decryption`                           |
+| `nip44_conversation_key` | `[<third_party_pubkey>]`                            | `<nip44_conversation_key>`                                             |
+| `nip44_encrypt`          | `[<third_party_pubkey>, [<plaintext_to_encrypt>]]`  | `array of <nip44_ciphertext> results of encryption`                    |
+| `nip44_decrypt`          | `[<third_party_pubkey>, [<ciphertext_to_decrypt>]]` | `array of <plaintext> results of decryption`                           |
 
 ## Response Events `kind:24133`
 
@@ -156,7 +156,7 @@ An Auth Challenge is a response that a remote signer can send back when it needs
 }
 ```
 
-Clients should display (in a popup or new tab) the URL from the `error` field and then subscribe/listen for another response from the remote signer (reusing the same request ID). This event will be sent once the user authenticates in the other window (or will never arrive if the user doesn't authenticate).
+Clients should display (in a popup or new tab) the URL from the `error` field and then subscribe/listen for another response from the remote signer (reusing the same request ID). This event will be sent once the user authenticates in the other window (or will never arrive if the user doesn't authenticate). It's also possible to add a `redirect_uri` url parameter to the auth_url, which is helpful in situations when a client cannot open a new window or tab to display the auth challenge.
 
 ## References
 
--- a/46.md
+++ b/46.md
@@ -6,10 +6,6 @@ Private keys should be exposed to as few systems - apps, operating systems, devi
 
 This NIP describes a method for 2-way communication between a remote signer and a Nostr client. The remote signer could be, for example, a hardware device dedicated to signing Nostr events, while the client is a normal Nostr client.
 
-Currently, NIP-46 uses a JSON-RPC style flow where you pass encrypted commands in `kind:24133` events. This rewrite strives to remove as much of the JSON-RPC as possible and uses new event kinds that correspond to each command a client might send and the three types of responses that remote signers can send back.
-
-This also allows us to incorporate [NIP-44](https://github.com/nostr-protocol/nips/blob/master/44.md) encryption in a way that doesn't break current NIP-46 implementations.
-
 ## Terminology
 
 - **Local keypair**: A local public and private key-pair used to encrypt content and communicate with the remote signer.
@@ -45,21 +41,21 @@ nostrconnect://<local-keypair-pubkey>?relay=<wss://relay-to-connect-on>&metadata
 
 In this last case, most often used to fascilitate an OAuth-like signin flow, the client first looks for remote signers that have announced themselves via NIP-89 application handler events.
 
-First the client will query for `kind: 31989` events that have a `k` tag of `24133` and/or any of the new request kinds detailed below.
+First the client will query for `kind: 31989` events that have a `k` tag of `24133`.
 
 These are generally shown to a user, and once the user selects which remote signer to use and provides the remote pubkey they want to use (via npub, pubkey, or nip-05 value), the client can initiate a connection. Note that it's on the user to select the remote signer that is actually managing the remote key that they would like to use in this case. If the remote pubkey is managed on another remote signer, the connection will fail.
 
 In addition, it's important that clients validate that the pubkey of the announced remote signer matches the pubkey of the `_` entry in the `/.well-known/nostr.json` file of the remote signer's announced domain.
 
-Clients that allow users to create new accounts should also consider validating the availability of a given username in the namespace of remote signer's domain by checking the `/.well-known/nostr.json` file for existing usernames. Clients can then show users feedback in the UI before sending a `create_account` event to the remote signer and receiving an error in return.
+Clients that allow users to create new accounts should also consider validating the availability of a given username in the namespace of remote signer's domain by checking the `/.well-known/nostr.json` file for existing usernames. Clients can then show users feedback in the UI before sending a `create_account` event to the remote signer and receiving an error in return. Ideally, remote signers would also respond with understandable error messages if a client tries to create an account with an existing username.
 
 ## The flow
 
 ### Client-side
 
-1. Clients need two things to start: a local keypair and the pubkey that the user wants to use for signing events.
-   - The local keypair is roughly disposable. E.g. - it's ok to store it locally (ideally encrypted) for some period of time. Once authenticated, remote signers can choose to trust this local keypair for a period of time. This local keypair can be deleted/rotated at will by the client and will only trigger a new auth challenge for the user.
-   - The remote pubkey is the pubkey that the user is trying to sign in / sign as. This can be fetched with an NIP-05 or entered directly by the user.
+1. Clients need several things to start: a local keypair and the details of the pubkey and remote signer the user would like to use.
+   - The local keypair is roughly disposable. E.g. - it's ok to store it locally (ideally encrypted) for some period of time. Once authenticated, remote signers can choose to trust this local keypair for a period of time. This local keypair can be deleted/rotated at will by the client and will only trigger a new auth challenge or permissions on the remote signer.
+   - The remote pubkey is the pubkey that the user is trying to sign in / sign as. This can be fetched with an NIP-05 or entered directly by the user (e.g. by passing a `bunker://` token).
 2. Clients use the local keypair for encrypting request events that they send to remote signers.
 
 ### Remote signer-side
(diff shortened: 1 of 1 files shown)


46.md
```diff
@@ -1,98 +1,174 @@
-NIP-46
-======
+# NIP-46 - Nostr Remote Signing
 
-Nostr Connect
--------------
+## Rationale
 
-`draft` `optional`
```
pablof7z, line 7: is not a draft anymore. Nostr is ossifying

nostrband:
I don't like separate kinds for each method, relays are much more empowered to monitor/censor nip46 activity. I think making app request payload an array (many already are in your request kinds) and adding the method as the first element solves this and doesn't make this nip much more json-ish.

erskingardner:
> I don't like separate kinds for each method, relays are much more empowered to monitor/censor nip46 activity

I don't think censoring is an issue. It's trivial to have both parties on multiple relays, the chances of collusion across the entire set of relays is minimum. Also, for privacy sake, I am imagine a lot of remote signers will want to run their own relays, further reducing this as an issue.

46.md
```diff
@@ -1,98 +1,174 @@
...
+| Kind    | Command                  | Content field (always NIP-44 encrypted)                                         | Data Type               |
+| ------- | ------------------------ | ------------------------------------------------------------------------------- | ----------------------- |
+| `24140` | `create_account`         | {username: <username>, domain: <domain>, <any other arbitrary key/value pairs>} | JSON stringified object |
+| `24141` | `connect`                | -                                                                               | null                    |
+| `24142` | `sign_event`             | nostr event to sign                                                             | JSON stringified object |
+| `24143` | `ping`                   | -                                                                               | null                    |
+| `24144` | `get_relays`             | -                                                                               | null                    |
+| `24145` | `nip04_encrypt`          | {third_party_pubkey: <third_party_pubkey>, content: <plaintext to encrypt>}     | JSON stringified object |
```
alexgleason, line 100: I would go with just `pubkey` for these methods. I think the directionality is inherently confusing (you have to ask, "third-party from the perspective of whom?"), and no possible name can help it, so I'd go with the simpler name, matching nostr-tools.

vitorpamplona:
+1 

I am feeling a convergence between DVMs and NostrConnect services. 

JSON-RPC indeed sucks. It's like creating separate poorly designed little protocols everywhere. But adding stringified jsons in the `.content` doesn't seem that much better. 

NIP-44 encryption is good. 

Do we think that exposing kinds for each command could be a privacy issue? With these new kinds, a third-party entity connected to the relay can track how many signs, encrypts, and decrypts each key does. 

A way to hide all of the actual info could be done using a single Ephemeral GiftWrap that nip44 wraps any of the other request/response kinds which can then be in plain text (No need for seals here). Then the proposed `.content` json stringified fields could just be regular tags. Trackers could see wraps going back and forth but they wouldn't necessarily know what they are. In fact, no one would know the keys are even talking to a signer service.  

Now, this whole RPC ordeal makes me wonder if we shouldn't just specify a standard HTTP Services API for Nostr that simply takes an HTTP POST with a Nostr event and replies with another Nostr Event signed by the service. One entry point, one reply. Signers could just expose a Services Definition Event in Nostr pointing to their service URL. 

Anyway. Forward we go.

46.md
```diff
@@ -1,98 +1,174 @@
...
+    "created_at": <unix timestamp in seconds>,
 }
 ```
 
-And the events the remote signer sends to the client have the following format:
+| Kind    | Command                  | Content field (always NIP-44 encrypted)                                         | Data Type               |
+| ------- | ------------------------ | ------------------------------------------------------------------------------- | ----------------------- |
+| `24140` | `create_account`         | {username: <username>, domain: <domain>, <any other arbitrary key/value pairs>} | JSON stringified object |
+| `24141` | `connect`                | -                                                                               | null                    |
+| `24142` | `sign_event`             | nostr event to sign                                                             | JSON stringified object |
```
alexgleason, line 97: The "command" name like `sign_event` isn't actually used by the protocol anymore, right? So in the table you can just write an English statement like "Sign event"

alexgleason (commented):
This is so much better than the original. Thank you for doing this!

erskingardner:
> Do we think that exposing kinds for each command could be a privacy issue? With these new kinds, a third-party entity connected to the relay can track how many signs, encrypts, and decrypts each key does.

yeah, I thought about this a bunch and went back and forth. I agree there is a small amount of metadata leakage here but I think it's minimal to the point of not being an issue. 

I actually initially wrote a version of this with gift wrapping and using tags like you'd suggested but the net effect was that it felt more bulky/clunky for the small number of a calls that require more than just a simple string for the content. In the end I went with the lighter, slightly leaky version.

46.md
erskingardner: That's true. I kept the old command name here to help people make the connection that these new events are the same as the old commands.

46.md
erskingardner: the other option would be `recipient_pubkey` which is probably what I should have gone with in the first place. There are a lot of `pubkey`s in implementing NIP-46 so I think it's important to be as clear as possible, but maybe that's just a documentation thing. I'm happy with `recipient_pubkey` or just plain `pubkey`

alexgleason: `recipient_pubkey` still has the directionality problem. If I had to use an underscored name I'd call it `encryption_pubkey`. But it's ugly enough I'd still rather just call it `pubkey`. I think you made it clear enough in every other part of the document. It's actually clearer to be less clear here.

staab:
+1 on minimizing metadata leakage, I hadn't thought of that. Everything we can do to keep people's stuff private is something worth doing.

> a standard HTTP Services API for Nostr that simply takes an HTTP POST

I wouldn't mind such a thing, but I don't think it's the best solution in this case. Ideally, a signer would also be a relay, which means it could advertise itself and clients could send events to the signer directly, removing a hop and enabling cool things like relay proxying, which signers can do without escalating permissions.

erskingardner:
> +1 on minimizing metadata leakage, I hadn't thought of that. Everything we can do to keep people's stuff private is something worth doing.

My worry with requiring the gift-wrap stuff is that it adds significant weight to implementing this for a very small privacy improvement. Think about it, even if someone watching can see that you were decrypting or encrypting events or were asking for signatures, what use is that? They can't see to who, or when those messages were actually published (sign doesn't exactly equate to publish). 

I just think this is a over-optimization for now. If we have actual reason for concern in the future, it's a simple addition to add here that it's also possible to gift wrap these exact same events. Clients and remote signers would then only have to add one small change to listen for those event kinds and unwrap them on arrival.

Again doable now, but feels unnecessary.

staab:
Like Vitor said, I don't know if we need to gift wrap them, if we use the same kind for everything, just encrypt the content. I know that brings back the json rpc stuff, but it might actually be ideal. But gift wrap would probably be even better since it would even further reduce metadata leakage.

I don't know what particular attack vector there would be with people snooping, but what I learned working on NIP 44 is that a robust security model assumes all information leaked to an attacker can be used, if only through statistical analysis.

But maybe this is all irrelevant if signers are encouraged to run and recommend its own relay.

fiatjaf:
This can only work if everybody agrees to stop using the previous flow on the same day and turn on the new flow. Otherwise it's better to not even try or we will **forever have to implement two different protocols**.

nostrband:
> A way to hide all of the actual info could be done using a single Ephemeral GiftWrap that nip44 wraps any of the other request/response kinds which can then be in plain text (No need for seals here). Then the proposed `.content` json stringified fields could just be regular tags. Trackers could see wraps going back and forth but they wouldn't necessarily know what they are. In fact, no one would know the keys are even talking to a signer service.

Nsec.app runs a server that needs to know who is trying to talk to whom to wake up the service worker in user's browser through web push. That means client will have to supply the info on how to track peers to my server and that defeats the purpose of gift wrapping to hide metadata.

That itself is a pity, but adding even more metadata, publicly available, just bcs we don't like jsons, makes no sense to me. 

> Now, this whole RPC ordeal makes me wonder if we shouldn't just specify a standard HTTP Services API for Nostr that simply takes an HTTP POST with a Nostr event and replies with another Nostr Event signed by the service. One entry point, one reply. Signers could just expose a Services Definition Event in Nostr pointing to their service URL.

There can't be a service URL with nsec.app, it can only talk through relays, same with Mutiny, so RPC over Nostr is good, but let's not make it worse.

vitorpamplona:
To be clear on the `adds significant weight`. The GiftWrap version uses a stack of 3 events, instead of a stack of 2. However, there is a separation of responsibilities between the 3 events that might lead to more interesting private developments. 

This PR's version of `sign_event`

```js
{
  "id": <id>,
  "kind": 24142 // sign event
  "pubkey": <local pubkey of the user>, // visible
  "tags": [["p", <remote pubkey>]], // visible (can't add any extra sensitive params to the call)
  "created_at": <unix timestamp in seconds>, // visible
  "content": nip44Encrypt(JSON.stringify(
    {
      "id": <id>,
      "kind": <kind>,
      "pubkey": <pubkey of the remote signer>,
      "content": "my content",
      "tags": [<any tags>],
    }
  )),
}
```

becomes

```js
{
  "id": <id>,
  "kind": 21059 // new ephemeral gift wrap kind
  "pubkey": <random key>,
  "tags": [["p", <signer pubkey>]],
  "created_at": <random>,
  "content": nip44Encrypt(JSON.stringify(
    {
      "id": <id>,
      "kind": 24142 // sign event
      "pubkey": <local pubkey of the user>, // hidden sender key
      "tags": [], // additional parameters for the sign event command
      "created_at": <unix timestamp in seconds>, // date of the request
      "content": JSON.stringify(
        {
          "id": <id>,
          "kind": <kind>,
          "pubkey": <pubkey of the remote signer>,
          "content": "my content",
          "tags": [<any tags>],
        }
      ),
    }
  )),
}
```

Besides privacy, I do think that moving the RPC params from a json inside the `.content` to actual tags in the appropriate event that are also encrypted beneficial to the "Nostr-nativeness" of this approach.

fiatjaf:
Even though I like this new protocol more the coordination costs for migrating are too high and **unless everybody who was tagged replies with gigantic enthusiasm** I think it is not worth doing.

alexgleason:
> This can only work if everybody agrees to stop using the previous flow on the same day and turn on the new flow.

That's exactly what would happen. NIP-46 is not precious enough yet. Now is exactly the right time to change it.

46.md
```diff
@@ -1,98 +1,174 @@
...
 
-### Started by the signer (nsecBunker)
+This is most common in a situation where you have your own nsecbunker or other type of remote signer and want to connect through a client that supports remote signing.
 
-The remote signer generates a connection token in the form
+The remote signer would provide a connection token in the form:
 
 ```
-bunker://<hex-pubkey>?relay=wss://...&relay=wss://...&secret=<optional-secret>
+bunker://<hex-pubkey-you-want-to-sign-as>?relay=<wss://relay-to-connect-on>&relay=<wss://another-relay-to-connect-on>&secret=<optional-secret-value>
```
monlovesmango, line 29:
```suggestion
bunker://<remote-pubkey>?relay=<wss://relay-to-connect-on>&relay=<wss://another-relay-to-connect-on>&secret=<optional-secret-value>
```
might be good to stick with the terminology that is defined.

46.md
```diff
@@ -1,98 +1,174 @@
...
 
-### Started by the client
+### Direct connection initiated by the client
 
-The client generates a QR code in the following form (URL-encoded):
+In this case, basically the opposite direction of the first case, the client provides a connection token (or encodes the token in a QR code) and the signer initiates a connection to the client via the specified relay(s).
 
 ```
-nostrconnect://<client-key-hex>?relay=wss://...&metadata={"name":"...", "url": "...", "description": "..."}
+nostrconnect://<client-key-hex>?relay=<wss://relay-to-connect-on>&metadata=<json metadata in the form: {"name":"...", "url": "...", "description": "..."}>
```
monlovesmango, line 39:
```suggestion
nostrconnect://<local-keypair-pubkey>?relay=<wss://relay-to-connect-on>&metadata=<json metadata in the form: {"name":"...", "url": "...", "description": "..."}>
```

46.md
```diff
@@ -1,98 +1,174 @@
...
+| `24140` | `create_account`         | {username: <username>, domain: <domain>, <any other arbitrary key/value pairs>} | JSON stringified object |
+| `24141` | `connect`                | -                                                                               | null                    |
+| `24142` | `sign_event`             | nostr event to sign                                                             | JSON stringified object |
+| `24143` | `ping`                   | -                                                                               | null                    |
+| `24144` | `get_relays`             | -                                                                               | null                    |
+| `24145` | `nip04_encrypt`          | {third_party_pubkey: <third_party_pubkey>, content: <plaintext to encrypt>}     | JSON stringified object |
+| `24146` | `nip04_decrypt`          | {third_party_pubkey: <third_party_pubkey>, content: <ciphertext to decrypt>}    | JSON stringified object |
+| `24147` | `nip44_conversation_key` | third_party_pubkey                                                              | string                  |
```
monlovesmango, line 102:
```suggestion
```

there was a lot of conversation around how to structure the nip44 calls for nip07 and we ultimately landed on not exposing the conversation key (incase we want to use that for other schemes in the future) and instead make the nip44 calls take and return arrays (to mitigate latency for multiple calls, especially since conversation key is no longer exposed). I think nip46 should also follow that pattern.

heres the pr: https://github.com/nostr-protocol/nips/pull/940

46.md
```diff
@@ -1,98 +1,174 @@
...
+| `24148` | `nip44_encrypt`          | {third_party_pubkey: <third_party_pubkey>, content: <plaintext to encrypt>}     | JSON stringified object |
```
monlovesmango, line 103:
```suggestion
| `24148` | `nip44_encrypt`          | [{pubkey: <third_party_pubkey>, content: <plaintext to encrypt>}, ...]     | Array of JSON stringified objects |
```

46.md
```diff
@@ -1,98 +1,174 @@
...
+| `24149` | `nip44_decrypt`          | {third_party_pubkey: <third_party_pubkey>, content: <ciphertext to decrypt>}    | JSON stringified object |
```
monlovesmango, line 104:
```suggestion
| `24149` | `nip44_decrypt`          | [{pubkey: <third_party_pubkey>, content: <ciphertext to decrypt>}, ...]    | Array of JSON stringified objects |
```

46.md
```diff
@@ -1,98 +1,174 @@
...
+This NIP describes a method for 2-way communication between a remote signer and a Nostr client. The remote signer could be, for example, a hardware device dedicated to signing Nostr events, while the client is a normal Nostr client.
+
+Currently, NIP-46 uses a JSON-RPC style flow where you pass encrypted commands in `kind:24133` events. This rewrite strives to remove as much of the JSON-RPC as possible and uses new event kinds that correspond to each command a client might send and the three types of responses that remote signers can send back.
+
+This also allows us to incorporate [NIP-44](https://github.com/nostr-protocol/nips/blob/master/44.md) encryption in a way that doesn't break current NIP-46 implementations.
+
+## Terminology
+
+- **Local keypair**: A local public and private key-pair used to encrypt content and communicate with the remote signer.
+- **Remote pubkey**: The public key that the user wants to sign as. The remote signer has control of the private key that matches this public key.
```
monlovesmango, line 17:
```suggestion
All pubkeys specified in this NIP are in hex format
```

46.md
```diff
@@ -1,98 +1,174 @@
...
-  "pubkey": "<client-key-hex>"
-  "kind": 24133,
-  "tags": [
-    ["p", "<signer-key-hex>"]
-  ],
-  "content": "nip04_encrypted_json({id: <random-string>, method: <see-below>, params: [array_of_strings]})",
-  ...
+    "id": <id>,
+    "kind": <kind for command>, // Each command has it's own kind
+    "pubkey": <local pubkey of the user>,
```
monlovesmango, line 86:
```suggestion
    "pubkey": <local keypair pubkey>,
```

46.md
monlovesmango:
```suggestion
| `24145` | `nip04_encrypt`          | {pubkey: <third_party_pubkey>, content: <plaintext to encrypt>}     | JSON stringified object |
```
I think the key should just be `pubkey` but the placeholder in brackets should be descriptive like you have it

46.md
```diff
@@ -1,98 +1,174 @@
...
+| `24148` | `nip44_encrypt`          | {third_party_pubkey: <third_party_pubkey>, content: <plaintext to encrypt>}     | JSON stringified object |
+| `24149` | `nip44_decrypt`          | {third_party_pubkey: <third_party_pubkey>, content: <ciphertext to decrypt>}    | JSON stringified object |
+
+## Response Event `kind: 24136`
+
+```json
+{
+    "id": <id>,
+    "kind": 24136,
+    "pubkey": <pubkey of the remote signer>,
```
monlovesmango, line 112:
```suggestion
    "pubkey": <remote pubkey>,
```

46.md
```diff
@@ -1,98 +1,174 @@
...
+| ------------ | ------------------------ | ------------------------------------------------ | ----------------------- |
+| `24140`      | `create_account`         | newly created pubkey                             | string                  |
+| `24141`      | `connect`                | "ack"                                            | string                  |
+| `24142`      | `sign_event`             | nostr event with signature                       | JSON stringified object |
+| `24143`      | `ping`                   | "pong"                                           | string                  |
+| `24144`      | `get_relays`             | array of relay objects                           | JSON stringified object |
+| `24145`      | `nip04_encrypt`          | ciphertext                                       | string                  |
+| `24146`      | `nip04_decrypt`          | plaintext                                        | string                  |
+| `24147`      | `nip44_conversation_key` | conversation key                                 | string                  |
```
monlovesmango, line 130:
```suggestion
```

46.md
```diff
@@ -1,98 +1,174 @@
...
+| `24148`      | `nip44_encrypt`          | ciphertext                                       | string                  |
```
monlovesmango, line 131:
```suggestion
| `24148`      | `nip44_encrypt`          | [ciphertext, ...]                                       | Array of strings                  |
```

46.md
```diff
@@ -1,98 +1,174 @@
...
+| `24149`      | `nip44_decrypt`          | plaintext                                        | string                  |
```
monlovesmango, line 132:
```suggestion
| `24149`      | `nip44_decrypt`          | [plaintext, ...]                                        | Array of strings                  |
```

46.md
```diff
@@ -1,98 +1,174 @@
...
+
+## Auth Challenge Event `kind: 24137`
+
+All requests / commands can potentially trigger an auth challenge, which is a special kind of repsonse event that tells the client that the user needs to provide a password or another form of authentication before the remote signer will process their request. The response content is always a URL that the client should display to the user. This URL is a page that is created/controlled by the remote signer in a way that allows the signer to collect the needed data before proceeding.
+
+```json
+{
+    "id": <id>,
+    "kind": 24137,
+    "pubkey": <pubkey of the remote signer>,
```
monlovesmango, line 142:
```suggestion
    "pubkey": <remote pubkey>,
```

46.md
```diff
@@ -1,98 +1,174 @@
...
-  "content": "nip04_encrypted_json({id: <request-id>, result: <string>, error: <reason-string>})",
-  ...
+Error response (a separate kind in order to make errors easier to handle on the client side).
+Content is NIP-44 encrypted string in the following format: `CODE: Error message text`. [Error codes](#error-codes) are detailed below.
+
+```json
+{
+    "id": <id>,
+    "kind": 24138,
+    "pubkey": <pubkey of the remote signer>,
```
monlovesmango, line 158:
```suggestion
    "pubkey": <remote pubkey>,
```

46.md
```diff
@@ -1,98 +1,174 @@
...
+    "content": <nip44(<request>)>,
+    "tags": [["p", <remote pubkey>]],
+    "created_at": <unix timestamp in seconds>,
 }
 ```
 
-And the events the remote signer sends to the client have the following format:
+| Kind    | Command                  | Content field (always NIP-44 encrypted)                                         | Data Type               |
+| ------- | ------------------------ | ------------------------------------------------------------------------------- | ----------------------- |
+| `24140` | `create_account`         | {username: <username>, domain: <domain>, <any other arbitrary key/value pairs>} | JSON stringified object |
```
monlovesmango, line 95:
```suggestion
| `24140` | `create_account`         | {username?: <username>, domain?: <domain>, <any other arbitrary key/value pairs>} | JSON stringified object |
```

monlovesmango (commented):
this is beautiful. huge improvement from previous nip46 and is actually consistent. thank you. also love the remote signer discovery via NIP-89. so cool.

as for leaking meta data, my preference would be to just communicate with the signer over a relay that enforces authentication for these event kinds. this way it is opt in and no overhead/complexity in data structure. I am working on a PR to spec out "authentication delegation tokens" which would work well for this, basically you could create delegation tokens that would be specifically for signing event kind 22242 and the query conditions would dictate what sensitive data the delegatee/local pubkey can access on behalf of the delegator/remote pubkey (for this nip would be the signer request/response kinds). hopefully that makes sense, but will try to get that PR out asap.

pablof7z:
/> Now, this whole RPC ordeal makes me wonder if we shouldn't just specify a standard HTTP Services API for Nostr that simply takes an HTTP POST with a Nostr event and replies with another Nostr Event signed by the service. One entry point, one reply. Signers could just expose a Services Definition Event in Nostr pointing to their service URL. 
> 
> 
> 
No, we would lose behind-router signers that can easily expose an HTTP endpoint.

pablof7z:
This is a massive improvement. The JSONRPC-on-top-of-nostr is why this very valuable NIP has sat idle for basically a year. Pretty much all implementations that actually do NIP-46 just use NDK, to the point where a typo on the create_account (wrong kind) went undetected for a couple months with a various instances already out there.

I've spoken with a lot of the people that have asked questions on how to implement it and it's always a pretty painful process because of how unnecessarily complex it is due to the very unnostr nature of it.

If we want more adoption that aren't like farcaster/hypercore's "just use the SDK, bro" we need to reduce the complexity where we can.

Massive enthusiastic ACK

staab:
> it's always a pretty painful process

I don't disagree with your main point, but I will say the main source of pain for me was not json rpc, but figuring out which key to use when, which nip05 to check and which key to pull from it, which kind to subscribe to, etc. Most of which are silent failures, making it very hard to debug. You've answered my questions about the general flow, but I can't help thinking there must be a better way. That would be the big thing to improve imo.

Currently it's something like this:

- Fetch nip 89 handlers from some relays.
- Infer the domain name from the nip05 of the handler.
- Fetch the handler domain's nostr.json.
- Get the signer's pubkey using the `_` entry
- Get the relays the signer listens on specifically for `create_account` requests (different relays may be used for all other requests). This comes from the `relays` field I believe, not the `nip46` one.
- Ask the user to select a domain
- Check to see if the user's name + signer domain has a nip05 record
- If not, use the signer's nostr.json `relays` to broker a `create_account` request addressed to the _signer_'s pubkey.
- If there is a record, use the _user_'s nip46 relay selection from the signer's domain's nostr.json to broker a `connect` addressed to the _user_'s pubkey.
- In either case, make sure to start a REQ with the _client_'s p-tag before sending the events or else you might miss them since they're ephemeral.

The main difficulty with the flow above is that it's not clear who you're asking for what. The mental model I would _like_ would be to create a channel between the client and signer, and use the same model as DMs to p-tag the recipient (rather than the user).

But Pablo pointed out to me that you might have multiple signers listening for your events, in case you want redundancy, so you can't just p-tag the signer. You also might not know the signer's pubkey if you're not starting with a domain name. So there's a certain amount of irreducible asymmetry that is ok as long as we can explain it thoroughly (along with explanations as to why it's so weird).

Some improvements we can make to that process though:

1. It's not clear that the signer's domain should be inferred from the 31990 handler's nip05 — for example, I used my personal nip05 for coracle's listing rather than _@coracle.social. This could confuse clients if I created a coracle signer. A "signer_domain" field or tag on kind 31990s would be clearer.

2. Rather than mashing the signer's relays and pubkey into nostr.json alongside everyone else's nip05, I would do something like:

```json
{
  "nip46": {
    "signer": "<signer pubkey>",
    "relays":  ["wss://relay.my-signer.com"],
  }
}
```

I don't think users should customize relays in the signer's nostr.json file, since they can already do that via nip 65 (maybe we could add a "signer" mark) if they want clients to send requests to all signers listening for the pubkey. But communicating with signers over user-selected relays doesn't really make sense to me because signers would have to listen on the superset of all user-selected relays, some of which the signer might not have permission to read from. So it seems like you're implicitly selecting a signer anyway based on the intersection of the user's and signer's choice of relays. What should probably be done is have users publish nip 65 `signer` relay selections reflecting the relays the signers they are registered with use (which signers can do automatically since they have the user's private key).

3. I'd like to see `create_account` gracefully handle a record already existing for the username so that if a user has explicitly selected a signer domain, a client doesn't have to check the nip05 record to figure out whether they're already signed up and choose a login method based on that. This gives the bunker the opportunity to provide a better UX than a client might, and reduce latency.

----

TLDR; The one-user-to-many-concurrent-signers paradigm (enforced by the original bunker approach, which has to work passively) is the source of a lot of this complexity, and conflicts with the oauth-like flow in which a signer has been explicitly selected by the user. Maybe we need to split this protocol into two different ones?

mikedilger:
I feel tired. But if you need me to rescue nostr by dropping everything and redoing nip46 yet again to meet this new protocol, I can't say no.  If my kid is stuck at school and missed the bus, I guess I have to go pick him up.  But it is annoying.

I was a little annoyed at switching to "bunker://" but it was a quick fix and luckily I read about it. How many implementations didn't know about that change?

I didn't read through all of this, but it doesn't look like a simple change.  Granted NIP-46 had some interpretation problems. But I don't think the thing that it was desperately needing was more kind numbers. Who knew?

IETF standards are a fucking pain in the ass to work out, hugely detailed, massively pedantic, battles rage between representatives from various corporations and interest groups about stupid details that nobody cares about and you will pluck your eyes out before it's over.  But the reason they endure that pain is because once a standard is out there, it never changes. That's one thing I prefer with IETF RFCs.  Nostr standards are a moving target that nobody can keep up with unless they work 28 hours per day just to keep up.

I guess my point is this:  Take more time making sure that a standard will be maintainable and supported forever before merging it, and make new standards in new NIPs.  This PR I think should be a new NIP that replaces NIP-46, but NIP-46 should remain behind for people who didn't upgrade to the replacement.

erskingardner:
> So there's a certain amount of irreducible asymmetry that is ok as long as we can explain it thoroughly (along with explanations as to why it's so weird).

I'm planning on adding a few diagrams and some more detailed descriptions of the flow you outlined in your note for sure. Just didn't want to do that work until we hammered out the details here. 

@staab I also think your point about making sure signers and users who use remote signing are signaling their relays of choice is a good one. We can add a section on relay best practices.

> With a comprehensive and well-written spec we can get away with the current way things are very easily going forward.

I disagree with this. The current spec is just half-baked. Especially in responses; the response object is being used in different ways in different places. e.g., auth challenges coming through via the `error` field. We still have a small enough number of people who have implemented this that we can change things and vastly improve this without _too_ heavy a lift in changing SDKs and client code.

All that said, There is a more minimal version of changes we could make that would reduce the blast radius here: 

Instead of using kinds for each of the different commands, we keep a more JSON RPC structure for requests and responses (keeping the current kind there) and only add new kinds for errors and auth challenges (you could even drop the error kind and only have a new kind for auth challenges). Personally, I don't like this and would prefer to personally spend my time helping all implementations upgrade to the new version but, if we do end up with that simplified set of changes, I still think it would be a huge improvement over what we have.

46.md
erskingardner: Why would these be optional?

46.md
erskingardner: Yeah – @staab flagged that discussion with me earlier. This might be reopening a painful wound but I'm not convinced that where the Nip44 calls in Nip07 stuff landed applies in the same way here. 

Remote signers are signing, encrypting, and decrypting for potentially many, many users instead of just one local user (as is the case in NIP-07). This means that you're shifting a potentially huge compute workload onto remote signers with the encryption stuff (since signing has to obviously stay remote). I completely understand why you all ended up doing away with the conversation key and allowing arrays for encryption the NIP-07 stuff, but I think it's probably more appropriate here to do the opposite. 

Asking a remote signer to decrypt your entire inbox each time your load up your DMs (which is how naive clients will do it) across thousands or eventually millions of users would require huge compute for remote signers, not to mention added network latency.

The idea of a rouge client doing something nefarious with your conversation key is a problem that, while I'm sure it'll happen eventually and is very bad, it's also something that is more easy to manage/solve with good UI/UX (remote signers can display warnings or limit the scope of usage of this method).

My main fear is that we'll end up with very few (read: centralized) remote signers if they have to be beasts of compute to manage all the remote work they have to do for clients.

46.md
erskingardner: This would actually be the pubkey of the remote signer (the bunker) since these response objects are coming from the bunker/remote signer

mikedilger:
I'm on the verge of leaving the nostr community.  If you can't get your shit together, I don't want to be walking through it.  This is NOT NIP-46.  This is something new.  EITHER modify the wording of NIP-46 to make it more explicit and clear, or make a different NIP that doesn't require me to spend another week rewriting everything again.  FOR FUCKS SAKE.

mikedilger:
I don't mind this new thing. I just don't think it should be an edit to NIP-46. That's all.  Sorry I'm so angry... sometimes I get angry.

nostrband:
> With a comprehensive and well-written spec we can get away with the current way things are very easily going forward.

Exactly!

fiatjaf:
I don't think there should be a new NIP for doing the same thing this is already doing. I think since both @mikedilger and @nostrband are not enthusiastic about the change (and honestly I am not very keen to rewriting my two libraries either) we should just give up now and leave things as they are.

Would be nice to get a nice rewording and clarification of the current spec.

erskingardner:
I'm not ready to bail on this yet. I think this (or the simplified solution I proposed above) would both be significantly better for the tradeoff of reimplementing in a few libraries (2 of which I'll be doing myself) and clients (again, a few here too).

v0l:
I don't think there is anything wrong with NIP-46 in its current form, its just a bit confusing because the spec is terrible.

I believe this all came about because of the oAuth flow, which i myself just read the doc on it today and it leaves out some important details like the `url` being in the `error` field... which seems kinda important.

I also dont like that oAuth flow is using the same id twice which already had a reply.

It would be better to just change oAuth flow since its new and leave everything else alone.

I would rather see a reply for auth like this:
```json
{ 
  "id": "connect-id-or-create_account-id",
  "result": "auth_url",
  "error": "oAuth flow requested",
  "url": "https://popup.com",
  "next": "next-id"
}
```
`next-id` being the reply after external auth flow, `next-id` could also simply be `connect-id + auth` or some other random value to make sure its unique enough for the client (i use UUID so its fine for me, but if clients use a counter then it might be a problem)

jb55:
On Thu, Feb 15, 2024 at 12:20:14AM -0800, Michael Dilger wrote:
>I'm on the verge of leaving the nostr community.  If you can't get your
>shit together, I don't want to be walking through it.  This is NOT
>NIP-46.  This is something new.  EITHER modify the wording of NIP-46 to
>make it more explicit and clear, or make a different NIP that doesn't
>require me to spend another week rewriting everything again.  FOR FUCKS
>SAKE.

You don't have to let people who do PRs to the nips repo affect you at
all. People write new spec and change specs all the time and it doesn't
bother me because I just ignore it. If a change is introduced that isn't
backward compatible than it's wrong.

I don't have the energy anymore to argue for people to not break things,
so now I just wait a good year and watch the same spec get refactored 3
times before I consider implementing it.

erskingardner:
Just so it's clear. This is backwards compatible. It uses new event kinds and leaves the original ones in place.

AsaiToshiya:
I agree with @mikedilger. I think add this as a new NIP and mark the current NIP-46 as deprecated, so everyone can be happy.

fiatjaf:
Adding a new NIP and using different kind numbers **is not the same as being backwards compatible**. The point is that existing implementations will stop working with implementations that change. The numbers are irrelevant.

Suppose we add this new spec using different kind numbers and as NIP-64 instead of NIP-46.

Now we will have some clients implementing NIP-46 and others implementing NIP-64 and they won't be compatible. JSON-RPC requests of NIP-46 will go unanswered from NIP-64 bunkers and NIP-46 bunkers will not be able to be used from NIP-64 clients. Then clients will do the most reasonable thing: implement both specs, and now everybody is happy! Except that for the rest of our days every single new codebase will have to implement both NIP-46 and NIP-64 just to be safe.

erskingardner:
Ok ok. This is a waste of human energy. I'm going to simply rewrite the current spec (broken as I find it) to make it more understandable. 

How do people feel about moving the OAuth stuff out into a separate doc? I know that @staab feels strongly that's the right way. I tend to agree. Remote signing if left in it's current form (but removing OAuth stuff) is very narrow and can be made quite clear. Trying to force auth challenges into that model makes is feel much more broken. If we treat that as a separate thing it might make this much better?

fiatjaf:
I would very much prefer them both in the same NIP since they're deeply intertwined.

staab:
> How do people feel about moving the OAuth stuff out into a separate doc? I know that @staab feels strongly that's the right way.

I do think the OAuth-like flow should be a separate protocol, for the following reasons:

- It relies on DNS. If the provider has the ability to configure a DNS record and serve a nostr.json file, they have the ability to run a public-facing server. This makes the need for the intermediary relay unnecessary. The signer could just receive events directly by websocket or even http request.
- In an oauth-like flow, the user has explicitly requested a particular signer. This means that passive signing (i.e. signers listening for events tagging the user) is not only unnecessary, it might even lead to unexpected behavior, like multiple signers sending back the same event, wasting resources, or sending multiple `auth_url`s leading to poor user experience. Clients could defend themselves by filtering unwanted responses out, but why bother.
- Targeting a single signer results in a much simpler mental model for who you're actually addressing, which relays to use, and which keys to use (the three actors keep the same roles rather than switching them around).

There's plenty of common ground between the two, but duplicating the signer functions is probably better than trying to shoehorn both models into the same spec.

erskingardner:
@nostrband @pablof7z @fiatjaf @mikedilger @staab @vitorpamplona @v0l @alexgleason @monlovesmango @jb55 @AsaiToshiya 

Thanks for all the feedback. 

I've dropped all the new stuff (other than adding the `create_account` command that is languishing in another PR) and just tried to document in a much more clear way how NIP-46 works.

[New file here](https://github.com/nostr-protocol/nips/blob/5727c55d4887be1742b3f7a9f52c9628f9c4c722/46.md)

nostrband:
> * In an oauth-like flow, the user has explicitly requested a particular signer. This means that passive signing (i.e. signers listening for events tagging the user) is not only unnecessary, it might even lead to unexpected behavior, like multiple signers sending back the same event, wasting resources, or sending multiple `auth_url`s leading to poor user experience. Clients could defend themselves by filtering unwanted responses out, but why bother.

Users can put their nsecs into several bunkers and each could send replies to the same request and no way to avoid that, so clients should dedup replies. NDK already does that for replies, I've submitted a patch for it to ignore repeated auth_url too, other implementations should do the same. Sorry for off topic, but maybe it's not - maybe it should be mentioned it the spec.

46.md
monlovesmango: maybe I misinterpreted @pablof7z ? this comment is why I thought these were optional: https://github.com/nostr-protocol/nips/pull/976#discussion_r1445399476

also, if these are mandatory it would mean they also have to be valid and I think we need to spec out better error handling for client.

erskingardner: this is out of date now. I'm going to resolve it.

46.md
```diff
@@ -1,98 +1,162 @@
...
-The signer scans the QR code and sends a `connect` message to the client in the specified relays.
+### Remote signer discovery via NIP-89
+
+In this last case, most often used to fascilitate an OAuth-like signin flow, the client first looks for remote signers that have announced themselves via NIP-89 application handler events.
+
+First the client will query for `kind: 31989` events that have a `k` tag of `24133`.
+
+These are generally shown to a user, and once the user selects which remote signer to use and provides the remote pubkey they want to use (via npub, pubkey, or nip-05 value), the client can initiate a connection. Note that it's on the user to select the remote signer that is actually managing the remote key that they would like to use in this case. If the remote pubkey is managed on another remote signer, the connection will fail.
+
+In addition, it's important that clients validate that the pubkey of the announced remote signer matches the pubkey of the `_` entry in the `/.well-known/nostr.json` file of the remote signer's announced domain.
```
staab, line 48: What domain is this? Right now it's inferred from the nip05 in the kind 31990 event, but I think some new tags should be introduced, maybe `signer_domain`, `signer_relays`. Domain would be optional to support the non-publicly-accessible use case. `signer_relays` would replace the `nip46` field in `nostr.json`. This way, DNS is not a hard dependency for discoverable signers. If a domain is provided, you'd still want to validate that the domain's pubkey matches the author's pubkey.

46.md
```diff
@@ -1,98 +1,162 @@
...
-nostrconnect://<client-key-hex>?relay=wss://...&metadata={"name":"...", "url": "...", "description": "..."}
+nostrconnect://<local-keypair-pubkey>?relay=<wss://relay-to-connect-on>&metadata=<json metadata in the form: {"name":"...", "url": "...", "description": "..."}>
 ```
 
-The signer scans the QR code and sends a `connect` message to the client in the specified relays.
+### Remote signer discovery via NIP-89
+
+In this last case, most often used to fascilitate an OAuth-like signin flow, the client first looks for remote signers that have announced themselves via NIP-89 application handler events.
+
+First the client will query for `kind: 31989` events that have a `k` tag of `24133`.
```
staab, line 44: Should be 31990 I believe

46.md
monlovesmango: neutering these calls to only take single arguments doesn't really fix this concern though. clients will still request for the same number of decrypts/encrypts. if a signer wants to rate limit how many decryptions are done per minute, it can easily do that even if the requests are submitted by array by only returning the first few elements decrypted.

I don't think its fair to make these calls less functional just to protect one particular use case of a signer. resource allocation is something an nsecbunker needs to handle, and we shouldn't use specifications as a proxy to handle this. besides, this is very unfair to small personal signers which would greatly benefit from these calls to take arrays.

as far as centralization, the most decentralized remote signer model is if everyone has their own app that manages their signing rather than relying on any external service. by tailoring these calls for nsecbunker I feel like you are actually favoring more centralized solutions.

monlovesmango: also, if we expose conversation key (which is needed for efficiency if we only allow single encryption/decryption) then there is no possibility of using the conversation key as some sort of shared secret key between too users in the future. imo this could have really creative applications.

fiatjaf:
> I do think the OAuth-like flow should be a separate protocol, for the following reasons:
> 
> - It relies on DNS. If the provider has the ability to configure a DNS record and serve a nostr.json file, they have the ability to run a public-facing server. This makes the need for the intermediary relay unnecessary. The signer could just receive events directly by websocket or even http request.
> - In an oauth-like flow, the user has explicitly requested a particular signer. This means that passive signing (i.e. signers listening for events tagging the user) is not only unnecessary, it might even lead to unexpected behavior, like multiple signers sending back the same event, wasting resources, or sending multiple auth_urls leading to poor user experience. Clients could defend themselves by filtering unwanted responses out, but why bother.
> - Targeting a single signer results in a much simpler mental model for who you're actually addressing, which relays to use, and which keys to use (the three actors keep the same roles rather than switching them around).
> There's plenty of common ground between the two, but duplicating the signer functions is probably better than trying to shoehorn both models into the same spec.

This makes sense now that I'm reading, but I think the ship has sailed.

However a good custodial bunker signer provider can still just point at itself as the only relay capable of delivering these signatures and the client woulds only connect to it, send NIP-46 events and receive responses and it would work.

46.md
```diff
@@ -1,98 +1,162 @@
...
-`draft` `optional`
+Private keys should be exposed to as few systems - apps, operating systems, devices - as possible as each system adds to the attack surface.
 
-This NIP describes a method for 2-way communication between a **remote signer** and a normal Nostr client. The remote signer could be, for example, a hardware device dedicated to signing Nostr events, while the client is a normal Nostr client.
+This NIP describes a method for 2-way communication between a remote signer and a Nostr client. The remote signer could be, for example, a hardware device dedicated to signing Nostr events, while the client is a normal Nostr client.
+
+## Terminology
+
+- **Local keypair**: A local public and private key-pair used to encrypt content and communicate with the remote signer.
+- **Remote pubkey**: The public key that the user wants to sign as. The remote signer has control of the private key that matches this public key.
```
staab, line 12: "remote keypair" sounds like it could be referring to the signer rather than the user. What about:

```suggestion
- **Client keypair**: A local public and private key-pair generated by the client, used to encrypt content and communicate with the remote signer.
- **Signer pubkey**: The public key belonging to the remote signer. This is either explicitly provided to the client, or discovered by accessing the signer's `/.well-known/nostr.json` file and looking up the `nip05._` entry.
- **User pubkey**: The public key that the user wants to sign as. The remote signer has control of the private key that matches this public key.
```

46.md
```diff
@@ -1,98 +1,162 @@
...
 ```
 
-The signer scans the QR code and sends a `connect` message to the client in the specified relays.
+### Remote signer discovery via NIP-89
+
+In this last case, most often used to fascilitate an OAuth-like signin flow, the client first looks for remote signers that have announced themselves via NIP-89 application handler events.
+
+First the client will query for `kind: 31989` events that have a `k` tag of `24133`.
+
+These are generally shown to a user, and once the user selects which remote signer to use and provides the remote pubkey they want to use (via npub, pubkey, or nip-05 value), the client can initiate a connection. Note that it's on the user to select the remote signer that is actually managing the remote key that they would like to use in this case. If the remote pubkey is managed on another remote signer, the connection will fail.
```
staab, line 46: > and provides the remote pubkey they want to use (via npub, pubkey, or nip-05 value)

This is a viable flow, but not I think the best UX. Best would be for a user to provide a username + domain so users don't have to remember their pubkey. Providing a different nip05 than the one you're logging in to would be very confusing. If this is a "create account" flow, the signer should return an error if the username is already taken.

But an optimistic connection request should also be supported. This would work the same as the `create_account` request, except that whether the username exists or not, the signer should return an auth_url rather than an error. If the username exists, the auth_url would be straightforward login screen. If the username doesn't exist, the signer should display a modified sign up screen with an option to add an existing private key. To make it clear, the `create_account` verb could be renamed to `connect_with_username` or something. The idea is that signers should follow postel's law and do their best not to return an error if possible.

Here's how I would write this:

```suggestion
The client MAY display the list of domains to the user so they can select the signer they prefer to use. Clients SHOULD allow users to enter a custom domain in case there is a signer that wasn't discovered.

In addition to the signer domain, the user must provide some way to identify their account, any of:

- Pubkey/npub
- NIP 05 address (even if the domain differs from the signer)
- Username (recommended)

If a NIP 05 address is provided, the client should de-reference it to find the user's pubkey and display an error if no record exists. If a pubkey is provided (either via NIP 05 address or directly), the client should then attempt to `connect` with the signer by p-tagging the provided **user pubkey**.

If a bare username is provided, the client should send a `connect_with_username` request p-tagging the **signer pubkey**. The signer should then return an `auth_url` which the client can use to redirect the user to complete the process. This auth_url should be a sign in page if the username exists in the signer's database, and a sign up page if the username doesn't exist yet. If desired, the signer MAY provide a field on the sign up form that allows users to enter an existing private key to custody. A client may exert more control over this flow if desired by checking the signer's `nostr.json` for a record containing the provided username, and showing a different form client-side before redirecting to the signer.

The username-based flow is optional, clients should be prepared for signers to omit support for it, either by having no publicly accessible domain, or by ignoring/rejecting the `connect_with_username` request.
```

46.md
```diff
@@ -1,98 +1,162 @@
...
+
+In addition, it's important that clients validate that the pubkey of the announced remote signer matches the pubkey of the `_` entry in the `/.well-known/nostr.json` file of the remote signer's announced domain.
+
+Clients that allow users to create new accounts should also consider validating the availability of a given username in the namespace of remote signer's domain by checking the `/.well-known/nostr.json` file for existing usernames. Clients can then show users feedback in the UI before sending a `create_account` event to the remote signer and receiving an error in return. Ideally, remote signers would also respond with understandable error messages if a client tries to create an account with an existing username.
```
staab, line 50: This could be removed if my suggestion above is accepted.

46.md
```diff
@@ -1,98 +1,162 @@
...
-  "tags": [
-    ["p", "<signer-key-hex>"]
-  ],
-  "content": "nip04_encrypted_json({id: <random-string>, method: <see-below>, params: [array_of_strings]})",
-  ...
+    "id": <id>,
+    "kind": 24133,
+    "pubkey": <local_keypair_pubkey>,
+    "content": <nip04(<request>)>,
+    "tags": [["p", <remote pubkey>]],
```
staab, line 82: Unless it's a create_account/connect_with_username, in which case it's the signer's pubkey. This is my very least favorite part of the spec.

staab: What about adding somewhere: `p tag MAY be the signer's pubkey, or if known, the user's pubkey. Using a user's pubkey makes it possible to connect to any signer that custodies the user's key, but if a specific signer has already been selected, it may be preferable to continue p-tagging the signer instead.` This could actually fix all my major objections to the current form of the protocol.

staab (changes requested):

monlovesmango:
as for the rewrite, I think its great. most of what I liked about this PR was just that it was clear, not the different event kinds.

some feedback I have:

- nip44 encrypt/decrypt calls should take arrays and getCoversationKey should not be exposed (see novel above)
- auth url should be new kind or have different payloads so it is explicit that this response should be handled differently. for example, for nip04 decryption if the plaintext just happens to be a URL how is the client supposed to know whether that is an auth URL response or a real response?
- why not upgrade encryption of content to nip44?
- `create_account` should also be for login of existing bunker users who don't have nip05
- mention that request can also be authored by signer under request details for only the `connect` method , and response can be authored by client only in response to signer `connect` request

staab:
> This makes sense now that I'm reading, but I think the ship has sailed.

Yeah, despite brainstorming my ideal world in public, I think we should stay conservative in how we change this NIP.

46.md
monlovesmango: I think this is more confusing. signer pubkey and user pubkey are the same no? no point creating different names.

staab: Ok, fair, but your confusion illustrates my point. The signer in this case is the service that custodies the user's key. Maybe "bunker" but that's a pretty idiosyncratic term.

46.md
```diff
@@ -1,98 +1,162 @@
...
+- `method` is the name of the method/command (detailed below).
+- `params` is a positional array of string parameters.
+
+### Methods/Commands
+
+Each of the following are methods that the client sends to the remote signer.
+
+| Command                  | Params                                            | Result                                                                 |
+| ------------------------ | ------------------------------------------------- | ---------------------------------------------------------------------- |
+| `create_account`         | `[<username>, <domain>, <optional_email>]`        | `<newly_created_remote_pubkey>`                                        |
```
monlovesmango, line 107: per @pablof7z all these params are optional: https://github.com/nostr-protocol/nips/pull/976#discussion_r1445399476

I think it would be good to be able to discern that from the spec.

also @alexgleason [suggested](https://github.com/nostr-protocol/nips/pull/976#discussion_r1448119529) it should be 'name' to be consistent with nip05 
```suggestion
| `create_account`         | `[<optional-name>, <optional-domain>, <optional_email>]`        | `<newly_created_remote_pubkey>`                                        |
```

46.md
monlovesmango: I like `signer pubkey` best.

46.md
erskingardner: I'm unclear as to how this would work if you didn't pass a username or domain option? Are you just then saying to a bunker, create me a new set of keys on your domain without a username?

46.md
erskingardner: I don't think that you could _just_ tag the signer in the `p` tag. There needs to be at least one `p` tag with the remote pubkey for most commands. Otherwise signers don't know which pubkey to use to decrypt the payload. 

I understand what you mean with `create_account` using the signer's pubkey and that being confusing. I can definitely add a comment about that.

erskingardner: also @staab I believe that you are still using the remote user pubkey in the `p` tag for the connect method.

46.md
erskingardner: Should that get added here or in NIP-89?

46.md
erskingardner: Updated the language in the latest.

erskingardner:
> nip44 encrypt/decrypt calls should take arrays and getCoversationKey should not be exposed (see novel above)

I still think `getConversationKey` is very valid and useful. Sounds like it's not a settled discussion on the nip-07 thread either?

> auth url should be new kind or have different payloads so it is explicit that this response should be handled differently. for example, for nip04 decryption if the plaintext just happens to be a URL how is the client supposed to know whether that is an auth URL response or a real response?

That was my original idea that was roundly rejected. In your example, clients would know because the result field is `auth_url` and the error field is the URL for the auth challenge. With decryption, the results are the results. 

> why not upgrade encryption of content to nip44?

Again, I wanted to do that but sounds like people are more interested in not changing that yet.

> create_account should also be for login of existing bunker users who don't have nip05

Why? That doesn't make any sense to me.

> mention that request can also be authored by signer under request details for only the connect method , and response can be authored by client only in response to signer connect request

I'm not sure what you mean here. Can you elaborate?

46.md
pablof7z: Yeah, I think that's perfectly valid, the client would simply not know which username was created but it can fetch it via nip05 since it has the pubkey

All these fields should be optional

fiatjaf: "the client would simply not know which username was created but it can fetch it via nip05 since it has the pubkey"

That's not great. NIP-05 queries can only be done by name (using the query string `?name=`). Services are not required to expose the full list of pubkeys to everybody.

Also if you don't know the domain name even that would be impossible to query.

`name` and `domain` cannot be optional.

fiatjaf:
We could just merge this because it is much better than what we currently have and then get PRs improving it later. Any objections?

fiatjaf:
About the NIP-44 calls, I think we should have them like NIP-07, and services who want to provide the conversation key can do it and reject all encrypt/decrypt requests -- while others can not expose the conversation key and just handle arrays with tons of encrypt/decrypt requests.

Clients should first try to get the conversation key, and if that fails try sending arrays of messages to be decrypted.

46.md
pablof7z: I think this should be removed or moved to a "best practices"; the checking is a nice-to-have for a better UX, but is in no way required for this flow.

46.md
```diff
@@ -1,98 +1,163 @@
...
+In addition, it's important that clients validate that the pubkey of the announced remote signer matches the pubkey of the `_` entry in the `/.well-known/nostr.json` file of the remote signer's announced domain.
+
+Clients that allow users to create new accounts should also consider validating the availability of a given username in the namespace of remote signer's domain by checking the `/.well-known/nostr.json` file for existing usernames. Clients can then show users feedback in the UI before sending a `create_account` event to the remote signer and receiving an error in return. Ideally, remote signers would also respond with understandable error messages if a client tries to create an account with an existing username.
+
+## The flow
+
+### Client-side
 
-## Event payloads
+1. Clients need several things to start: a local keypair and the details of the pubkey and remote signer the user would like to use.
```
pablof7z, line 57: No, the user just needs to know the remote user pubkey and the relays in which to talk to, both things can come from the `bunker://` or if using a nip-05 from the `nip46` entry in the nostr.json.

46.md
```diff
@@ -1,98 +1,163 @@
...
+
+```json
+{
+    "id": <request_id>,
+    "result": "auth_url",
+    "error": <URL_to_display_to_end_user>
+}
+```
+
+Clients should display (in a popup or new tab) the URL from the `error` field and then subscribe/listen for another response from the remote signer (reusing the same request ID). This event will be sent once the user authenticates in the other window (or will never arrive if the user doesn't authenticate).
```
pablof7z, line 159: Clients can add an extra `redirect_uri` to get the bunker to redirect the user to that URI after account creation. This is useful in popup/new-tab-blocking contexts where all the client is able to do is redirect the user to a different URL.

46.md
erskingardner: Agree, I don't think these can be optional.

46.md
pablof7z: The entire concept of the `remote signer pubkey` is ONLY useful when you need to perform tasks on the bunker (like `create_account`) itself.

the `p` tag should always be either the local user pubkey or remote user pubkey (i.e. the counterparty of the communication)

If we moved the `create_account` into a separate part of the NIP this might become way simpler since we don't even need to mention until that part that the signer itself can have a pubkey.

46.md
erskingardner: I think it should stay with the rest of the "discovery" section so that it's coherent. I've tried to make it clear that this is all a suggestion. I can strengthen the language there.

46.md
erskingardner: And they need a disposable set of keys to communicate with the bunker.

I'll make the source of these details more clear.

46.md
erskingardner: I'll add a note.

arthurfranca:
> @monlovesmango  if we expose conversation key [...] clients can forever read your private messagesg. getCoversationKey should not be exposed 
> @erskingardner I still think getConversationKey is very valid and useful
> @fiatjaf services who want to provide the conversation key can do it and reject all encrypt/decrypt requests

From [nip-07 discussions](https://github.com/nostr-protocol/nips/pull/940#issuecomment-1878551791), getConversationKey can be made secure when NIPs using NIP-44 encryption define a `salt` to use when generating the conversation key (and pass it as argument to getConversationKey).

A good default is using the involved event's `created_at` first 4 digits from the left as `salt`, which would make the getConversationKey different every 12 days. So a conversation key wouldn't be able to decrypt events outside this time window.

First step would be merging #1020, then we should add the salt argument to NIP-46 and NIP-07 getConversationKey calls.

arthurfranca:
Pardon me. @staab is right. If the salt is based on `created_at`, the malicious client can just ask the nip07/46 thing to generate a bunch of conversation keys related to many timestamps.

The `salt` is still useful for [other use case](https://github.com/arthurfranca/nips/blob/priv-dm/43.md#session-channel) but that's another subject.

arthurfranca:
@staab the salt instead could be a random value on a public event tag though like `["salt", "37981873721398"]`. This way the malicious client currently using nip07/46 would not be able to ask to generate conversation keys to decrypt future events.

Just like not revealing conversation key is useful to protect just future events aswell isn't it? Cause clients using nip07/46 can ask to decrypt any previous ciphertext/event.

Am i right?

staab:
Sure, but now you have a linked list, which introduces new failure modes. I haven't yet seen a use of salt that solves a real problem that doesn't also have the same problems explicit key exchange does. Maybe one exists, I mean, this is not my wheelhouse, but I think the only real solution to forward secrecy is a real ratchet. Maybe salt is part of that, but I don't know.

pablof7z:
Guys, let's not bring the salt and nip-44 discussion to this PR; whatever is decided wrt NIP-07's NIP-44 interface this NIP should do the same.

monlovesmango:
> I still think `getConversationKey` is very valid and useful. Sounds like it's not a settled discussion on the nip-07 thread either?

yep, I just added it back to the nip07 pr.

> That was my original idea that was roundly rejected. In your example, clients would know because the result field is `auth_url` and the error field is the URL for the auth challenge. With decryption, the results are the results.

hmm where was it roundly rejected? maybe I missed it. I think most were against creating new kinds for existing functionality, but the auth URL would be net new functionality and so shouldn't duplicate anything that already exists. at the very least, even if we don't want a new kind we can put auth url in a new field like `auth_url` rather than put it in `result`. offers a lot more clarity for little/no cost.

> Again, I wanted to do that but sounds like people are more interested in not changing that yet.

oh I hadn't seen that.

> > create_account should also be for login of existing bunker users who don't have nip05
> 
> Why? That doesn't make any sense to me.

what is the current easy login flow for an existing bunker user that has no nip05 identifier? I haven't seen that accounted for yet. basically #3 from @staab 's https://github.com/nostr-protocol/nips/pull/1047#issuecomment-1944454074.

> > mention that request can also be authored by signer under request details for only the connect method , and response can be authored by client only in response to signer connect request
> 
> I'm not sure what you mean here. Can you elaborate?

right now request events are explicitly defined as:
```json
{
    "id": <id>,
    "kind": 24133,
    "pubkey": <local_keypair_pubkey>,
    "content": <nip04(<request>)>,
    "tags": [["p", <remote_user_pubkey>]], // NB: in the `create_account` event, the remote signer pubkey should be `p` tagged.
}
```
however when a signer sends a `connect` request to the client, the `pubkey` will actually be remote_pubkey and the `tags` will actually be [["p", <local_keypair_pubkey>]].

the same for a response event for a signer sent `connect` request, the `pubkey` and `tags` will have local_keypair_pubkey and remote_pubkey respectively.

monlovesmango:
@pablof7z really like how the gist is arranged, agree its clearer. for the "The flow" section under #2 I would add that pubkey can also be obtained by signer sent `connect` request

@fiatjaf no objections to merging as is

erskingardner:
I just moved things around in line with @pablof7z's gist to give the Signer commands it's own section and to move OAuth-like stuff to it's own section as well. 

I have also purposefully updated the NIP44 commands to be "potential future addition" so that we can merge this without committing to the form of those commands/responses yet.

AFAICT, this is perfectly inline with what is out there in the wild now, just way better explained.

vitorpamplona:
This is still so confusing. Let's say the user is doing the NIP-05 auth. Client gets the key from `/.well-known/nostr.json` and what? How does the client register the local keypair with the remote signer? 

Is the `OAuth-like Flow` a requirement for this type of login? Otherwise, how does the user accept the new local key pair in their signer service? Do they have to have the signer open somewhere to approve the new local key pair?

erskingardner:
> Is the OAuth-like Flow a requirement for this type of login? Otherwise, how does the user accept the new local key pair in their signer service? Do they have to have the signer open somewhere to approve the new local key pair?

It's not a requirement but it definitely makes it easier. You're correct, that anytime you're using a remote signer, you will need to allow a new local client from the signer. That "allow" flow will be different for each type of signer. Nsecbunker's default is to see a request show up in the Admin UI and then approve. If you're using the OAuth-like flow, Nsecbunker sends back an auth challenge so that the user can visit a webpage (hosted by the bunker) to approve (it's using the same mechanism under the hood). 

You can imagine other signers, like phones, etc that might send you a push notification and then you'd approve there. But yeah, there's always a step on the remote signer to allow the access.

46.md
```diff
@@ -1,98 +1,223 @@
...
+## Appendix
+
+### NIP-05 Login Flow
+
+Clients might choose to present a more familiar login flow, so users can type a NIP-05 address instead of a `bunker://` string.
+
+When the user types a NIP-05 the client:
+
+- Queries the `/.well-known/nostr.json` file from the domain for the NIP-05 address provided to get the user's pubkey (this is the **remote user pubkey**)
+- In the same `/.well-known/nostr.json` file, queries for the `nip46` key to get the relays that the remote signer will be listening on.
```
vitorpamplona, line 200: how do you query for the NIP-46 key?

pablof7z: it works just like `relays`, you want to get the `nip46.$pubkey` to get the array of relays for this pubkey:

```bash
$ wget -q -O - https://highlighter.com/.well-known/nostr.json | jq .names.pablo
"bab85564b5d198c3d3a9f856a3fe9dc8e2f9ca3952ba6e7dd60bc98030321103"

$ wget -q -O - https://highlighter.com/.well-known/nostr.json | jq .nip46.bab85564b5d198c3d3a9f856a3fe9dc8e2f9ca3952ba6e7dd60bc98030321103
[
  "wss://relay.nsecbunker.com",
  "wss://relay.damus.io"
]
```

vitorpamplona: ohh the nip46 key is the service provider key. ok.

vitorpamplona: oh wait, there is a nip46 field as well. Interesting.

staab: @vitorpamplona my comments are lost in the maelstrom, but I propose to change this to put most of it on the handler event.

monlovesmango:
since we are adopting @pablof7z 's restructuring, can we also simplify the terminology section as he suggested in https://github.com/nostr-protocol/nips/pull/1047#issuecomment-1952334959. so only need local pubkey and remote pubkey (no longer need to distinguish between remote user pubkey and remote signer pubkey)

monlovesmango:
@alexgleason maybe we can move the conversation around how nip44 calls within nip46 should be structured to https://github.com/nostr-protocol/nips/pull/983.

think this convo should stay focused on updating NIP-46 clarity. the nip44 calls discussion seems to get pretty involved.

PR #1051, opened 2024-02-15
Deprecates NIP-26

vitorpamplona:


melvincarvalho:
Quite useful.  Isnt it in use.  And it's a simple way to do delegation, which is needed.  What other way is there to do delegation?

vitorpamplona:
Most relays and clients don't implement it and those who did are removing it. 

I am not sure delegation is necessary. 

The deprecation makes a statement that people are looking for a better/less cumbersome way to do this.

jb55:
> I am not sure delegation is necessary.

delegation is quite useful for things like a twitter crossposter: give a service a signed delegation and they can create posts on your behalf. This is much simpler than alternative approaches like nsecbunker which requires it to be "online".

vitorpamplona:
> A few people are removing it and therefore its deprecated?

Yep, if people don't want to use it, there is no point in keeping it. I also checked nostr.wine's massive database and it looks like the last non-test delegated event was signed last September. Meaning, that no one is using this. Maybe @nostrband can give us more info. 

> This is much simpler than alternative approaches like nsecbunker which requires it to be "online".

Sure but today's model requires every client and relay to implement it otherwise no one will see your crosspost. Since we can't control all clients, most users immediately see that as a deal breaker and go for an nsecbunker solution. Brands are particularly inclined to just use the nsecbunker instead.

basantagoswami:
> I also checked nostr.wine's massive database and it looks like the last non-test delegated event was signed last September.

As far as I can remember the delegation event was created to support minds.com's nostr integration. They are most likely still producing events that use it

vitorpamplona:
> As far as I can remember the delegation event was created to support minds.com's nostr integration. They are most likely still producing events that use it

Can you find these events? I think users realized that even if these events have been created, very few clients and relays implement this NIP which means no one is seeing their posts. 

I couldn't find anyone using this lately. But maybe they have their own relays and clients that work with it and have a separate nostr community somewhere.

basantagoswami:
> Can you find these events?

They have a relay (wss://relay.minds.com/nostr/v1/ws) mentioned in their [developer docs](https://developers.minds.com/docs/decentralization/nostr/#how-do-i-view-my-posts-using-a-nostr-client). We might find some there

arthurfranca:
I also thought NIP-26 was dying but I agree it could be good if relays and clients supported it.

> delegation is quite useful for things like a twitter crossposter: give a service a signed delegation and they can create posts on your behalf. This is much simpler than alternative approaches like nsecbunker which requires it to be "online".

Imo the worse about nsecbunker is that in practice most of them will be custodial. Shared nostr accounts, like from companies, could have their privkey leaked by the nsecbunker provider.

The bad thing about NIP-26 is that a token can't be revoked, for example, when an employee gets fired. Though it could be less of a problem if using very narrow created_at limits.

vitorpamplona:
> Imo the worse about nsecbunker is that in practice most of them will be custodial. 

I intend to create a standalone Bunker Android app that receives `kind:24133` via push notification. It wouldn't work well for decrypting chat messages (too many requests via push will get you blocked), but simple singing and encrypting works. Then all you need is a trustless server to send those `24133` events via Push. The key is on the phone. 

Maybe that's @pablof7z's vision for the nsec bunker as well. The key doesn't need to be online. 

> The bad thing about NIP-26 is that a token can't be revoked

Yep. NIP-26 essentially grants full signing control for a given time window forever.

staab:
I think we should keep this. It's simple, and even if it's not well-supported, it could prove useful.

vitorpamplona:
> It's simple, and even if it's not well-supported, it could prove useful.

I don't think it is that simple of a scheme. Its simplicity is misleading. 

Imagine having to deal with delegated keys signing for NWC payments on behalf of the NWC key. Or Group admins delegating control over the group admin key to others. Bunkers with delegated keys for both the local keys and the main key. 

Also, DMs from delegated keys should be merged with the main user's conversation. But which key do you encrypt the reply to? Both? What if there are 5 delegatees in the same "chat room"? Do we confuse users by putting all messages in the same room or by creating separate rooms whose names are the exact same (the main key) while the main key is actually not seeing the conversation?  What do we do on GiftWraps?

jb55:
On Wed, Feb 21, 2024 at 12:21:25PM -0800, Vitor Pamplona wrote:
>> The bad thing about NIP-26 is that a token can't be revoked
>
>Yep. NIP-26 essentially grants full signing control for a given time window forever.

this is not true. you would almost always have an expiring delegation condition.

jb55:
unless you're referring to sending backdated posts but that would easily be preventable using a bit of code.

arthurfranca:
> Imagine having to deal with delegated keys signing for NWC payments [...] Or Group admins delegating control over the group admin key to others. Bunkers with delegated keys for both the local keys and the main key.

Yeah some use cases are hard to implement and probably won't be supported. I think the two use cases that make sense the most are loading delegated events on feed/user profile and DMs.

> Also, DMs from delegated keys should be merged with the main user's conversation. But which key do you encrypt the reply to?

I think clients could present delegated DMs clearly as coming from a specific delegatee. Like on websites with embedded chat widgets where you talk to Alice, representing Brand XYZ and if you refresh the page maybe you will end up talking with John, still representing the same Brand XYZ.

vitorpamplona:
All these suggestions look good but without actual implementations we wouldn't know if they work or if they just create more problems.

These suggestions highlight my point that this NIP is a poorly defined mess.

I am still trying to find people actually using it. 

We have a 2+ clients rule here. If no one is using it as it is defined, we should deprecate it.  

Somebody else can then send a PR with the fixes and wait for 2 clients and relays to properly implement it before merging again.

fiatjaf (approved):

alexgleason (approved):
NIP-46 is a better path forward. The idea of implementing this seems like hell, and a security nightmare. Minds has implemented ActivityPub and I don't think their Nostr relay is even active anymore.

AsaiToshiya (approved):

insanity54:
Halt. I think all ya'll are moving too fast for this new Nostr ecosystem where adopting developers are just starting to come online.

I am building an X to Nostr crossposter and NIP-26 is the only solution I have found where I don't have to custody the user's nsec. I'd like to teach good practice of nsec self-custody and I as a service provider want to get permissions to create kind:1 events on the user's behalf for something like 30 days at a time. This way I custody a delegated child key, but not the master key which I know I should never have access to.

NIP-26 solves my problem **today**. NIP-46 doesn't solve my problem at all, because I cannot use it to create an automated solution that can publish Nostr events.

There is no NIP-26 alternative for automated, non-custodial event publishing and therefore I do not think it should be deprecated.

vitorpamplona:
The expiration dates don't work because as soon as you delegate the key, the sub key can always post in the approved period forever (people can post in the past). 

And no one ever wants that. They want a revocation mechanism that actually revokes privileges. 

Also, there is no relay implementation that supports nip 26, so there is no "this works for me today".

jb55:
@vitorpamplona i've never understood this criticism, all we would need to do is spec a max post in the past duration. is this your only concern?

greenart7c3:
There's other issues like clients can't filter for delegation tags.
All clients and relays must implement this nip for it to be usable.
When I was trying to implement it there's was only 1 or 2 relays with delegation support.
Only gossip had delegation support, so only people that used gossip could see with who they were talking to.
If people really want nip 26 we should change it to use 1 letter tag, but we still have to change all clients, relays, paid relays and file storage services

vitorpamplona:
> I've never understood this criticism, all we would need to do is spec a max post in the past duration.

I can only judge what's written. Some folks suggested specific changes but proposals presented as PRs also do not work in my mind. 

> is this your only concern?

No, I think delegation will never work until we figure out how to share encrypt/decrypt with both keys. 

Besides that, I have several questions on what to do when the NIP-26 is used in tandem with other nips. 
- Can the sub key kind:5 delete the posts of the main account? 
- Can a delegated key send a NWC payment on behalf of the main key? 
- What happens when a zap split names a sub account, should we search for the root key's ln address before paying? 
- Should we allow delegations on NIP-98?
- Can NIP-29 groups be managed by sub keys? 
- Do sub-keys automatically get access to all NIP-42 relays as the main account?
- When assembling Wiki pages do you consider delegated keys as the same? 
- Should the main account see the subaccount's drafts with delegation? Which encryption should we use for that?

nostrband:
> I am building an X to Nostr crossposter and NIP-26 is the only solution I have found where I don't have to custody the user's nsec. 

Use nip-46 - connect to user's key store and talk to it whenever a signature is needed, if it's offline - retry later.

kehiy (approved):
implementing nip-26 on the relay side will hurt performance and require many security and database checks and conditions. i think we must deprecate this as well. people can still use remote singing as an alternative and later we can work on a new model for delegation which is simpler. there are ideas that can help us to do that, like using temp subkeys of a parent key and keep a reference of events signed by these subkeys somewhere that can be resolved by clients (like nip-05) and then we can see which of subkeys or events are approved by the main key. 

notes:
1. owner of main key can add/remove subkeys and events. (where client makes query to.)

2. owner of main key should sign the list of approved events and pubkeys.

3. we can use a replaceable event for this purpose instead of some server to return it.

using this model they can even deny some old post or manage these kind of stuff.

staab (approved):

greenart7c3 (approved):

PR #1052, opened 2024-02-15
Breaking changes list

mikedilger:
This is my attempt to start a list of potentially breaking changes. When we merge in a new NIP that we think might have a breaking change, even a subtle one, we can include it on this list.

Nostr changes fast, but most changes are typo corrections, and then there is a lot of stuff added that doesn't break anything. Developers who are too busy to keep up all the time can really benefit from this list to make sure they have kept up.

I think enough of us care that we can keep this list maintained.

### DIFF
--- a/BREAKING.md
+++ b/BREAKING.md
@@ -5,13 +5,13 @@ reverse chronological order.
 
 | Date        | Commit    | NIP      | Change |
 | ----------- | --------- | -------- | ------ |
-| 2024-02-15  | afbb8dd0  | [NIP-39](39.md) | PGP identity was removed |
-| 2024-02-07  | d3dad114  | [NIP-46](46.md) | Connection token format was changed |
-| 2024-01-30  | 1a2b21b6  | [NIP-59](59.md) | 'p' tag became optional |
-| 2023-01-27  | c2f34817  | [NIP-47](47.md) | optional expiration tag should be honored |
-| 2024-01-10  | 3d8652ea  | [NIP-02](02.md) | list entries should be chronological |
-| 2024-01-10  | 3d8652ea  | [NIP-51](51.md) | list entries should be chronological |
-| 2024-01-01  | 27fef638  | [NIP-57](57.md) | optional P tag added |
+| 2024-02-15  | [afbb8dd0](https://github.com/nostr-protocol/nips/commit/afbb8dd0) | [NIP-39](39.md) | PGP identity was removed |
+| 2024-02-07  | [d3dad114](https://github.com/nostr-protocol/nips/commit/d3dad114)  | [NIP-46](46.md) | Connection token format was changed |
+| 2024-01-30  | [1a2b21b6](https://github.com/nostr-protocol/nips/commit/1a2b21b6)  | [NIP-59](59.md) | 'p' tag became optional |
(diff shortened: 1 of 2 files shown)


mikedilger:
Such a list is not a license to break things!  I think of this as a list of shame.

staab:
Also refer to https://nostrability.com/ for interoperability problems in the wild. The tricky thing is that some of those problems are related to poor specs. Either we keep the specs intact and work around those problems, or we break the spec to fix those problems. I don't know what the right answer is.

mikedilger:
I think whichever answer causes the least disruption and incompatibility.

BREAKING.md
```diff
@@ -0,0 +1,21 @@
...
+This is a history of NIP changes that potentially break pre-existing implementations, in
+reverse chronological order.
+
+| Date        | Commit    | NIP      | Change |
+| ----------- | --------- | -------- | ------ |
+| 2024-02-15  | [afbb8dd0](https://github.com/nostr-protocol/nips/commit/afbb8dd0) | [NIP-39](39.md) | PGP identity was removed |
+| 2024-02-07  | [d3dad114](https://github.com/nostr-protocol/nips/commit/d3dad114)  | [NIP-46](46.md) | Connection token format was changed |
+| 2024-01-30  | [1a2b21b6](https://github.com/nostr-protocol/nips/commit/1a2b21b6)  | [NIP-59](59.md) | 'p' tag became optional |
+| 2023-01-27  | [c2f34817](https://github.com/nostr-protocol/nips/commit/c2f34817)  | [NIP-47](47.md) | optional expiration tag should be honored |
+| 2024-01-10  | [3d8652ea](https://github.com/nostr-protocol/nips/commit/3d8652ea)  | [NIP-02](02.md) | list entries should be chronological |
```
vitorpamplona, line 12: how is this order breaking things? Before there was no order. And it's a SHOULD, not a MUST.

BREAKING.md
```diff
@@ -0,0 +1,21 @@
...
+| 2024-01-10  | [3d8652ea](https://github.com/nostr-protocol/nips/commit/3d8652ea)  | [NIP-51](51.md) | list entries should be chronological |
+| 2024-01-01  | [27fef638](https://github.com/nostr-protocol/nips/commit/27fef638)  | [NIP-57](57.md) | optional P tag added |
```
vitorpamplona, line 14: How does an added tag break things?

vitorpamplona (commented):
This is a good-to-have list but some items in the list don't seem to be breaking changes to me.

BREAKING.md
mikedilger: Hmm. I guess that one is entirely safe.

BREAKING.md
vitorpamplona: > I would rather have false positives than false negatives.

Maybe we need a rating system for how bad it is based on how many clients were/are expected to be broken. 

Having a massive list of mostly minor things will not be that helpful for the average dev. We should focus on the changes that broke everything and keep a list of the minor stuff on the side.

mikedilger: Could use 'breaking' and 'advisory'.  Don't want to be too complex.

fiatjaf:
I'll merge now since this is not a breaking change and then you can propose changes @vitorpamplona.

PR #1053, opened 2024-02-15
NIP-49 [breaking]: password normalized to unicode NFKC format

mikedilger:
This will aid compatibility.

Shocking for me to suggest a breaking change after my hissy fit last night. But I'm actually ok with important breaking changes as long as we document them (see my other PR here https://github.com/nostr-protocol/nips/pull/1052).

### DIFF
--- a/49.md
+++ b/49.md
@@ -12,7 +12,7 @@ This NIP defines a method by which clients can encrypt (and decrypt) a user's pr
 Symmetric Encryption Key derivation
 -----------------------------------
 
-PASSWORD = read from the user
+PASSWORD = Read from the user. The password should be unicode normalized to NFKC format to ensure that the password can be entered identically on other computers/clients.
 
 LOG\_N = Let the user or implementer choose one byte representing a power of 2 (e.g. 18 represents 262,144) which is used as the number of rounds for scrypt. Larger numbers take more time and more memory, and offer better protection:
 


vitorpamplona:
ACK. 

Welcome to the club of NIP breakers.

vitorpamplona:
Good to go from here: https://github.com/vitorpamplona/amethyst/commit/6553f3bec5e43446ace0a6221407e898e9d65d69

mikedilger:
So I would have chosen NFKC from the outset (and this PR is currently written with NFKC) if this had never been deployed, but maybe NFC is better since this is a breaking change and NFC will break less sitautions. NFC only changes things to their canonical form and doesn't fold distinct variants into simpler variants.

Anybody else have an opinion on that?
https://www.unicode.org/reports/tr15/

vitorpamplona:
NFKC is ok for me.

fiatjaf:
Nostr is a shit show.

fiatjaf:
Can we get at least one test vector?

mikedilger:
What language? Kotlin? Go? JavaScript?

fiatjaf:
Go, but I was reading about it and this normalization stuff is used for a bunch of different things, like turning `á` into `a`? Anyway, I think I can figure it out with with some test vectors.

mikedilger:
I am working on test vectors.

mikedilger:
Here is a starting point test vector.  If someone enters the first password below (I've listed it on 3 lines in 3 different ways) the NFKC algorithm should convert it into the second password below (again listed on 3 lines in 3 different ways)

"ÅΩ"
U+212B U+2126
UTF-8 bytes: [0xE2, 0x84, 0xAB 0xE2, 0x84, 0xA6]

"ÅΩ"
U+00C5 U+03A9
UTF-8 bytes: [0xC3 0x85, 0xCE, 0xA9]

PR #1055, opened 2024-02-16
BREAKING.md: update the history

AsaiToshiya:
I will add NIP changes until March 2023.


AsaiToshiya:
Done, but I think there are probably some changes that are missing or unnecessary.

fiatjaf:
This list of changes is full of very meaningless things, but thank you for compiling it. Must have been a tiresome job.

fiatjaf:
@mikedilger

AsaiToshiya:
I don't know if it's meaningless, but it's definitely a tiresome job to have to keep updating.

PR #1060, opened 2024-02-18
Create nip1962.md

gazhayes:


fiatjaf:
I'm afraid this will make breaking changes sound ok because there is a standard way to make them.

fiatjaf:
I like this, but maybe these shouldn't be NIPs, they should be NOPs and then we could have a central directory that points to Nostr events (making it clear who is owner of each NOP and what is its current status)? Or what exactly are you proposing?

PR #1062, opened 2024-02-20
Add ninvite and relay access requests

staab:
This is probably poorly organized, suggestions on where to put the various parts of this welcome. This PR does two things:

- Adds a bech32 entity, `ninvite` which encodes a few different entities. This is meant to help with onboarding and sharing content outside of nostr.
- Adds a relay access request event for closed relays that work something like pyramid, or some other invite link based scheme.

I'm adding support for both to Coracle this week.

https://coracle.us-southeast-1.linodeobjects.com/invites.mov

### DIFF
--- a/xx.md
+++ b/xx.md
@@ -1,0 +1,57 @@
+NIP-XX
+=======
+
+Invite Links
+------------
+
+This NIP defines a new bech32-encoded entity, `ninvite1` and a new event representing an invite code which
+will grant access to a given relay.
+
+## Invite entity
+
+A `ninvite` is a bech32-encoded entity which includes one or more pubkeys, relay urls, and group ids.
+
+```typescript
+InvitePointer = {
+  people?: string[]
+  relays?: {
+    url: string
+    claim?: string
+  }[]
+  groups?: {
+    address: string
+    relay: string
+    claim?: string
+  }[]
+}
+```
+
+The contents are a binary-encoded list of `TLV` (type-length-value), with `T` and `L` being 1 byte each (`uint8`, i.e. a
+number in the range of 0-255), and `V` being a sequence of bytes of the size indicated by `L`.
+
+These possible standardized `TLV` types are indicated here:
+
+- `0` is _optionally_ a pubkey. This may be included multiple times.
+- `1` is _optionally_ a tuple of `url` and `claim` joined using a `|`. This may be included multiple times.
+- `2` is _optionally_ a tuple of `address`, `relay`, and `claim` joined using a `|`. This may be included multiple times.
+
+Invite links MAY be shared as bare nostr links, but more often SHOULD be baked into a handler url, since that will be more
+useful to users coming from a non-nostr context.
+
+## Relay access request
+
+User MAY request access to relays using an ephemeral `kind 28935` event sent ONLY to the relay in question. Relay access requests
+MUST have a `claim` tag containing an arbitrary string. These events MUST NOT be stored or relayed to subscribers.
+
+```json
+{
+  "kind": 28934,
+  "content": "Optional message",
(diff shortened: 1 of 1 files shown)


xx.md
```diff
@@ -0,0 +1,63 @@
...
+Invite Links
+------------
+
+In order to facilitate exchange of nostr recommendations, it's useful to have a single link with multiple
+entities encoded in it. This can be especially helpful as part of an onboarding process, to get new users
+started with recommendations hand-picked by another user.
+
+This NIP defines a new bech32-encoded entity, `ninvite1` and a new event representing an invite code which
+will grant access to a given relay.
+
```
vitorpamplona, line 13: I would get rid of the two initial paragraphs and simplify them to: 

This NIP defines a new bech32-encoded entity, `ninvite1`, to encode recommendations for existing and new nostr users.

vitorpamplona:
I wonder if this should be a bech32 link or if it should just be a full event kind that is signed by the sender so that you can show the sender's information when the friend opens the link. 

"Jon is recommending the following people, relays and groups"

If it is an event, the ninvite could just be a regular nevent.

staab:
> I wonder if this should be a bech32 link or if it should just be a full event kind that is signed by the sender so that you can show the sender's information when the friend opens the link.

Interesting idea, showing sender would be kind of nice. But:

> If it is an event, the ninvite could just be a regular nevent.

I don't think this is workable, since there are secret invite codes in the entity, which you wouldn't want to publish.

> optional "expire" timestamp and "limit" integer parameters.

What are these for?

xx.md
```diff
@@ -0,0 +1,57 @@
...
+A `ninvite` is a bech32-encoded entity which includes one or more pubkeys, relay urls, and group ids.
+
+```typescript
+InvitePointer = {
+  people?: string[]
+  relays?: {
+    url: string
+    claim?: string
+  }[]
+  groups?: {
```
AsaiToshiya, line 21: Sorry for the basic question, but what exactly does "groups" refer to?

staab: Any replaceable event that can be "joined". So my intention here is NIP 72 communities and (draft) NIP 87 groups, but anything else analogous that comes up.

AsaiToshiya: Is the reference to those events set to "address"?

staab: Yep, same value as an `a` tag

AsaiToshiya: I understand and thanks for your answers.

vitorpamplona:
>> If it is an event, the ninvite could just be a regular nevent.

> I don't think this is workable, since there are secret invite codes in the entity, which you wouldn't want to publish.

With private zaps, we encrypt a fully signed event and bech32-encode the ciphertext with `pzap1`. You could do a version of that were a query param transfers a password in a way that anyone with the link can decrypt it. You could even re-use (49.md)[49.md] strong password derivation scheme to make it super secure if you want. 

Something like: 

`nostr:nsecinvite1<bech32>?password=1234`

Then you can: 
1. Secure with a password
2. Verifiable with a signature
3. Extensible to include any tags and avoid the TLV part.

staab:
Yeah, at that point we should just do a regular query string. Which I might actually do, because I'm not sure this has to be interoperable, since a client recommendation is just the link's base url, and is just as relevant as relay/group/pubkey recs.

vitorpamplona:
It feels like to make it interoperable we would need to design invitation payloads for every event kind. Then if the Client happens to support that particular kind, it can show the invitation but if not then it can send it to other clients. 

njump could then point each invite item to a separate client. 

I don't know...

staab:
Closed in favor of #1078 and #1079

PR #1063, opened 2024-02-20
NIP-07: add NIP-44 calls

alexgleason:
Very simpler alternative to #940

It does the same thing as NIP-04 but with the better encryption.

### DIFF
--- a/07.md
+++ b/07.md
@@ -20,8 +20,8 @@ Aside from these two basic above, the following functions can also be implemente
 async window.nostr.getRelays(): { [url: string]: {read: boolean, write: boolean} } // returns a basic map of relay urls to relay policies
 async window.nostr.nip04.encrypt(pubkey, plaintext): string // returns ciphertext and iv as specified in nip-04 (deprecated)
 async window.nostr.nip04.decrypt(pubkey, ciphertext): string // takes ciphertext and iv as specified in nip-04 (deprecated)
-async window.nostr.nip44.encrypt(pubkey, plaintext): string // returns ciphertext and iv as specified in nip-44
-async window.nostr.nip44.decrypt(pubkey, ciphertext): string // takes ciphertext and iv as specified in nip-44
+async window.nostr.nip44.encrypt(pubkey, plaintext): string // returns ciphertext as specified in nip-44
+async window.nostr.nip44.decrypt(pubkey, ciphertext): string // takes ciphertext as specified in nip-44
 ```
 
 ### Implementation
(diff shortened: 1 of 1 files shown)


staab (approved):

staab:
We've come full circle 😂

monlovesmango:
I won't stand in the way of this if this is what most want. I still do think the batched calls are useful and would be nice to align with NIP46 (especially bc I think NIP46 is the future and NIP07 will become increasingly obsolete, and having them be similar increases support for both).

@alexgleason I am curious what your opinion is on exposing getConversationKey?

vitorpamplona:
We are debating too many hypotheticals. 

Does anybody have a Client implementation that calls NIP-44 methods with arrays? Is there a Signer that takes in and displays the array to the user for approval? If not, then let's merge this. 

If the need for arrays comes up, I am sure we can just add new methods to deal with that need.

staab:
Well, there is this: https://github.com/fiatjaf/nos2x/pull/47/files, but I don't think it's deployed yet. I also used the same approach with alby. But neither is released, I'm just waiting for this discussion to resolve. So yeah, let's just go back to NIP 04 and add bulk options if we need them.

monlovesmango (approved):

fiatjaf:
What is the argument against having an optional `getConversationKey()` method? I never read that. No one likes that anymore?

vitorpamplona:
The best argument for not needing `getConversationKey` in this PR is that no client needs it yet. :)

staab:
> What is the argument against having an optional getConversationKey() method? I never read that. No one likes that anymore?

You log into `evil client`. It fetches private messages, and asks you for conversation keys for each. As part of that giant wall of alby requests, it also asks for every conversation key for every likely/possible contact, and (if using time-based salt) across every epoch from now until 10 years in the future. Then it stores them on a server. Now it has access to all your messages in perpetuity with any specified counter party, and you have no idea.

fiatjaf:
Fine.

staab:
Updated my PRs to [nos2x](https://github.com/fiatjaf/nos2x/pull/57) and [alby](https://github.com/getAlby/lightning-browser-extension/pull/2653).

alexgleason:
Looks like https://github.com/fiatjaf/nos2x/pull/57 is merged. #983 is merged. Everyone is pretty much in agreement at this point. This should be ready to merge.

erskingardner:
CleanShot 2024-02-22 at 08 26 59

That changeset... a thing of beauty.

PR #1064, opened 2024-02-20
NIP-57: Define private zaps

benthecarman:
Been working on zap stuff and noticed these weren't defined in the NIP. One thing I left out was how to derive the encryption key for private zaps. The [current standard](https://github.com/rust-nostr/nostr/blob/5bc7193d9ab5dafdc8df7a276d3f2f79c5512120/crates/nostr/src/nips/nip57.rs#L302) for this is really ugly and does not work for things like NIP-07 extensions or nsecbunker use cases. Been trying to come up with a better way to do it but don't think there is a way to do it deterministically without direct access to the private key

### DIFF
--- a/57.md
+++ b/57.md
@@ -183,6 +183,82 @@ When an event includes one or more `zap` tags, clients wishing to zap it SHOULD 
 
 Clients MAY display the zap split configuration in the note.
 
-## Future Work
+### Appendix G: Anon Zap Request Event
 
-Zaps can be extended to be more private by encrypting `zap request` notes to the target user, but for simplicity it has been left out of this initial draft.
+The same as a zap request, with the `anon` tag added.
+
+Example:
+
+```json
+{
+  "kind": 9734,
+  "content": "Zap!",
+  "tags": [
+    ["relays", "wss://nostr-pub.wellorder.com", "wss://anotherrelay.example.com"],
+    ["amount", "21000"],
+    ["lnurl", "lnurl1dp68gurn8ghj7um5v93kketj9ehx2amn9uh8wetvdskkkmn0wahz7mrww4excup0dajx2mrv92x9xp"],
+    ["p", "04c915daefee38317fa734444acee390a8269fe5810b2241e5e6dd343dfbecc9"],
+    ["e", "9ae37aa68f48645127299e9453eb5d908a0cbb6058ff340d528ed4d37c8994fb"]
+    ["anon"]
+  ],
+  "pubkey": "97c70a44366a6535c145b333f973ea86dfdc2d7a99da618c40c64705ad98e322",
+  "created_at": 1679673265,
+  "id": "30efed56a035b2549fcaeec0bf2c1595f9a9b3bb4b1a38abaf8ee9041c4b7d93",
+  "sig": "f2cb581a84ed10e4dc84937bd98e27acac71ab057255f6aa8dfa561808c981fe8870f4a03c1e3666784d82a9c802d3704e174371aa13d63e2aeaf24ff5374d9d"
+}
+```
+
+### Appendix H: Private Zap Request Event
+
+A `private zap request` is used to send a zap to a user but only revealing the sender's identity to the receiver. 
+An `anon` tag is added to the `zap request` event with the encrypted `private zap request` event.
+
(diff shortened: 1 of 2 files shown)


staab:
What about something like `["anon", nip04.encrypt(ephemeral_sk, recipient_pk, sender_pk)]`? That would be possible with basic nip 04 support, and would only reveal the sender to the recipient.

benthecarman:
You need to encrypt a signed event, otherwise you can make it look like the zap came from someone else. The other problem is that if I send a private zap and drop the `ephemeral_sk` you can no longer see that it was your private zap. Current clients handle this by having a deterministic way to generate the `ephemeral_sk`, however, it is not pretty and doesn't work for things like NIP-07

jb55:
On Tue, Feb 20, 2024 at 02:28:20PM -0800, benthecarman wrote:
> Been trying to come up with a better way to do it but don&#39;t think
> there is a way to do it deterministically without direct access to the
> private key

This is why I never spec'd this in its current state btw. The web people
would get mad at me.

vitorpamplona:
I was hoping to use the `get_conversation_key` from NIP-44 with the `salt` exposed as a parameter as in #1020 to deterministically create keys. Then the encryption would use the `get_conversation_key(my_priv_key, my_pub_key, event.id || event.created_at)` as the new private key.

But:
1. I am not sure if that is a safe way to generate new private keys. 
2. We haven't decided if exposing conversation keys to clients is a good idea.

benthecarman:
If NIP 7 and 46 had something like `create_hmac_from_secret` that'd be ideal for here and potentially other user cases. Could try proposing that

vitorpamplona:
Same issue we get from the get conv key discussion:  Clients should not have access to any keys when using nip7 or nip46. Having keys exposed means that now users must trust the client to manage keys correctly and not trusting clients is the main reason to use nip7 and nip46.

benthecarman:
If it's an hmac it's just a hash of your key, you'd only be trusting them with whatever that is used for.

vitorpamplona:
> I feel like that's what the extension permissioning is for. Otherwise we would have to add an optional param to every sign/decrypt/encrypt function to derive sub keys.

Agree. But keep in mind that NIP-46 and NIP-07 do not offer the same permission support as NWC tends to do. 

> add an optional param to every sign/decrypt/encrypt function to derive sub keys.

I am frankly not sure if this is bad. We could overload these methods with additional key derivation params. As long as nips follow the same key-derivation (say... an hmac) when they needed it, it could work really well and satisfy the key security people seem to want.

arthurfranca:
> a deterministic way to generate the ephemeral_sk

Considering that conversationKey [can’t be used as private key](https://github.com/nostr-protocol/nips/pull/945#issuecomment-1882545014), the key derivation could use `hkdf(sk, salt)` (taken from [here](https://github.com/nostr-protocol/nips/pull/761#discussion_r1485667225) - without the sha256 part cause the hashing was for another use case). Then this PR should define a salt such as "<event_being_zapped.pubkey>" or something else.

NIP-07/46 would need getPublicKey/signEvent/(nip44.)encrypt/decrypt methods like the following:
1) `withDerivedKey({ salt: "..."', method: 'nip44_encrypt' }, pk, plaintext)` OR
2)  `nip44.encrypt(pk, plaintext, { withDerivedKey: { salt: '...' } )` OR
3)  `nip44.deriveAndEncrypt(derivationSalt, pk, plaintext)`

jb55:
This should be a new NIP imo, not a fan of extending the same nip over and over.

jb55:
I've used this to create a damus DIP with extended info on how we generate deterministic private zap keys:

https://github.com/damus-io/dips/blob/master/03.md#deterministic-ephemeral-keys

generalsunion:
> He estado trabajando en temas de zap y he notado que no están definidos en el NIP. Una cosa que dejé fuera fue cómo derivar la clave de cifrado para zaps privados. El [estándar actual](https://github.com/rust-nostr/nostr/blob/5bc7193d9ab5dafdc8df7a276d3f2f79c5512120/crates/nostr/src/nips/nip57.rs#L302) para esto es realmente feo y no funciona para cosas como extensiones NIP-07 o casos de uso de nsecbunker. He estado tratando de encontrar una mejor manera de hacerlo, pero no creo que haya una manera de hacerlo de manera determinista sin acceso directo a la clave privada.

PR #1069, opened 2024-02-22
Clarify behaviour of kind:5 events for replaceable events

basantagoswami:


vitorpamplona:
I would clarify that regular replaceable events (non-parametrized ones - without the `d` tags), including `kind:0` and `kind:3`  also follow the same rule. Meaning the existence of the `d`-tag is irrelevant.

PR #1074, opened 2024-02-22
Use `q` instead of `e` tags for quote reposts

jb55:
Damus does this, other clients should too!

### DIFF
--- a/18.md
+++ b/18.md
@@ -20,9 +20,10 @@ reposted.
 
 ## Quote Reposts
 
-Quote reposts are `kind 1` events with an embedded `e` tag
-(see [NIP-08](08.md) and [NIP-27](27.md)). Because a quote repost includes
-an `e` tag, it may show up along replies to the reposted note.
+Quote reposts are `kind 1` events with an embedded `q` tag of the note being
+quote reposted. The `q` tag ensures quote reposts are not pulled and included
+as replies in threads. It also allows you to easily pull and count all of the
+quotes for a post.
 
 ## Generic Reposts
 


vitorpamplona:
Is the goal to replace the `mention` marker in [NIP-10](10.md) or to use `q` alongside it? Do we just duplicate to avoid breaking the past? 

Is there a difference between quote posting and mentioning? I am not sure. 

Can `q` point to other kinds or is it just for kind 1s?

jb55:
The mention marker doesn't solve:

1. not pulling down quotes when you don't need to
2. adding the ability to only pull a list of quote reposts (which could be substantially larger for viral posts and could kill pulling large threads)

jb55:
> Can q point to other kinds or is it just for kind 1s?

I'm only considering kind1 at the moment.

alexgleason (approved):

vitorpamplona:
> I'm only considering kind1 at the moment.

This means that we will have to keep both `mention`s and `q` at the same time. Which is fine. That will make it backward-compatible at least.

fabianfabian:
👍 
Nostur changed to `q` instead of `e` a few weeks ago

mikedilger:
You should also edit NIP-10 where it says:

> Those marked with "mention" denote a quoted or reposted event id.

And perhaps also edit NIP-27 to add more detail around this paragraph:

> Including [NIP-10](https://github.com/nostr-protocol/nips/blob/master/10.md)-style tags (["e", <hex-id>, <relay-url>, <marker>]) for each reference is optional, clients should do it whenever they want the profile being mentioned to be notified of the mention, or when they want the referenced event to recognize their mention as a reply

Because otherwise the amount of contradictory and confusing statements NIPs gets out of hand.

alexgleason:
> This change doesn't solve this either, because for compatibility we'll have to include the `e` tag still.

Can't you just include a NIP-21 URI in the content? Excluding "e" events for quotes in tags doesn't really break anything. Users are still notified by the "p" tag. All it does is frees the "e" tag from being queried in the wrong scenarios.

vitorpamplona:
> > The mention marker doesn't solve:
> > 
> > 1. not pulling down quotes when you don't need to
> 
> This change doesn't solve this either, because for compatibility we'll have to include the `e` tag still.

The one thing that the old `[n]` references did was to make sure Clients created `e` tags correctly. Since we migrated to `nostr:` uris, there is no way to get an accurate number of quoted notes through tags. 

I trust neither `q` nor `mention` to build these counters. I just download everything and parse the content to see if the author actually quoted the other post. 

> Excluding "e" events for quotes in tags doesn't really break anything.

It breaks down those clients that followed NIP-10 and simply filters by `e` to bring all replies and quotes of an event. If we change to `q`, they are not going to download quotes anymore.

staab:
> This change doesn't solve this either, because for compatibility we'll have to include the e tag still.

To be clear, I'm not saying `q` is a bad idea. But if we can't fix kind 3 relays, we can't fix this either. Which way, modern man?

jb55:
> > The mention marker doesn't solve:
> > 
> > 1. not pulling down quotes when you don't need to
> 
> This change doesn't solve this either, because for compatibility we'll have to include the `e` tag still.

It just fixes the "bug" of quote reposts getting pull down with the thread? Why is the `e` reference needed and what does it break?

jb55:
One thing I can see it "breaking" is if clients rely on this for pulling and counting quote reposts? fwiw they would already be missing quotes from clients who aren't adding the e tag (nostur, damus, maybe others).

staab:
> One thing I can see it "breaking" is if clients rely on this for pulling and counting quote reposts? fwiw they would already be missing quotes from clients who aren't adding the e tag (nostur, damus, maybe others).

Right. But before `e` tag was added there was no way to do so right? So removing the `e` tag would break any clients that count quotes. But does anyone actually do that?

jb55:
it would already be broken, since there are clients that are already not sending `e` tags. it would back backwards compatible for older threads because they could keep the code that counts pulled quote reposts?

basantagoswami:
> Can't you just include a NIP-21 URI in the content?

That's pretty much how it works on Twitter too. You can quote tweet by copy pasting a tweet's URL. I think this is the best way of doing quotes.

vitorpamplona:
I am less concerned about breaking the quote count and more concerned with Clients simply not downloading any quoted posts if the `e` tag is not mentioned anymore. 

The quote count will always be broken and adding `q` will not fix it.

jb55:
does amethyst put quoted reposts in the thread? I'm not sure I understand the concern.

vitorpamplona:
> does amethyst put quoted reposts in the thread?

No. The simple answer is we just ignore `mention` tags when building the thread.

jb55:
then switching to `q` would allow you to not pull the quotes instead of ignore them. 

>  more concerned with Clients simply not downloading any quoted posts if the e tag is not mentioned anymore

this is the right behavior? it's fixing the bug of downloading stuff you don't need to in a thread. again, if they actually want to download the quote reposts and not put them in the thread then it would only be for counting quotes. I'd argue this is the incorrect way to do this because of filter limits that could disrupt these two different types of queries.

vitorpamplona:
> then switching to `q` would allow you to not pull the quotes instead of ignore them.

Not really. It won't change anything for me. I don't use Nostr filters for the local cache. On top of that, there are thousands of `mention`s out there. I will still need to remove them to avoid breaking the thread view of the past. Maybe one day when absolutely everyone makes the transition to `q` and all the past events get deleted then sure. Meanwhile, there is still going to be a lot of `mention`s to deal with. 

> this is the right behavior?

It was written in NIP-10. I am not sure if it is right or not, but it was clearly the guidance. If people have been following it, then it will break them.

jb55:
I forgot @vitorpamplona now runs the nips repo.

vitorpamplona:
> I forgot @vitorpamplona now runs the nips repo.

Jesus, dude. I don't have a strong preference in this battle. We can do `q` or `mention`s. To me these are not mutually exclusive. In fact, I suggested doing both in my 2nd reply to avoid breaking anything.

jb55:
you make nostr development miserable, I genuinely dislike working on this project because of you. I have left the nostr-protocol organization and will no longer submit nips. enjoy wasting other peoples time and continuing to push your bs into every single issue and PR.

vitorpamplona:
Well, I am not here to make anyone's life miserable. I will stop providing feedback to your posts. If anyone else feels the same, just let me know and I will stop it with you as well. It's all good.

staab:
@jb55 what is the difference between this breaking change and removing kind 3 relays? Do you want pushback against breaking changes or don't you?

jb55:
I've documented this damus behavior as DIP-02:

https://github.com/damus-io/dips/blob/master/02.md

mikedilger:
I've only ever tried to follow these ever-changing NIPs. I'm not wed to what gossip currently does, but I'm not going to change it haphazardly.  NIP-08, NIP-10, NIP-18, and NIP-27 all have their fingers in this pie, and now DIP-02 does too.  That's 5 conflicting standards I'm supposed to comply with.

I will just wait until things settle down and then I'll change gossip to whatever the answer is. Far too many cooks in this kitchen.

tyiu:
lol please reopen this PR. It makes sense to me. It does make it more annoying to need to support both ways but what's being proposed is the right way to do it. Eventually, q tags will be the only way people do quote reposts for new events. It'll be up to clients to decide if it wants to keep supporting the old way. We shouldn't be afraid of making breaking changes. In a decentralized world, there are no completeness guarantees.

alexgleason:
@tyiu :100:

mikedilger:
I agree. Re-open this. I think 'q' is a good idea.  Far better than unclear 'e' tags ('e' is far too general)

AsaiToshiya:
> It does make it more annoying to need to support both ways but

> It'll be up to clients to decide if it wants to keep supporting the old way.

I agree with @tyiu.

fabianfabian:
To me, the only problem with `q` is that it was not described anywhere, or at least I could not find it, I’m surprised its already in use for over 200+ days apparently.

First time I heard about it was:
nostr:note14nwedvff0kn6cgpv9z88d0rash4m4najxwd3sd7txkstlnyp0fvqv3wzcx

This was when other clients were also using `e` with `"mention"` and seemed to work fine. 

The reason I changed to `q` recently was because

1. Complaints: nostr:note12aktw523yugu8ek944hecatunsvcfuasc4326wmz5hawnlq5d0pq9y3pv0
2. Changing to `q` did not break anything, but did fix it for other client
3. It is better if you want to fetch only the things you need (I wish we also had this for root posts without replies)

alexgleason:
3 approvals, and based on the comments every person who voiced concern also voiced approval. So I'm merging it.

mikedilger:
It's not wrong, but it failed to update NIP-10 or NIP-27 or BREAKING.md.  I guess I'll open another PR.

PR #1079, opened 2024-02-23
Add relay access requests

staab:
The use case here is private relays that can be joined by a user with minimal friction. The idea is that a user can auth using an invite code instead of a challenge. Relays should then save that code and use the normal challenge flow in the future. Currently implemented in Coracle and Triflector/Khatru.


vitorpamplona:
Are you using `claim` just because the `challenge` has different semantics like the one below? 

> The challenge is valid for the duration of the connection or until another challenge is sent by the relay.

For clients, it doesn't matter if it is a `claim` or `challenge`. They could be just one tag. If the `challenge` comes from the relay in one connection, its duration is for the connection. If the challenge comes from an external source, the challenge might be long-lived. 

I assume relays store the `claim` and as soon as somebody joins with a valid claim, it accepts that pub key as invited. That could work with `challenge` as well. As long as the relay has it stored somewhere, it doesn't need to be the temporary challenge it creates for the connection.

staab:
Interesting. I've gone through about 4 designs with "claim" before getting here, so it hadn't occurred to me that they could be the same thing. But I think it would be good to keep them separate, because a claim _grants_ access, while a challenge only prevents replay attacks.

fiatjaf:
Why not use a different dedicated event kind and send these claims with a simple `["EVENT", ...]`?

I imagine the idea behind using `AUTH` is that these events won't be broadcasted, but a relay that supports the `claim` protocol would also not broadcast these events -- and even if it did the claim would already have been claimed at the time anyone else got it.

monlovesmango:
just created #1082 for authentication delegation. I think could accomplish what you are trying to doing with claim? but is also supposed to be more general purpose and extensible as we think of new things we want to delegate authentication for.

monlovesmango:
I think new event kind makes sense, but why not send it back in a `["AUTH", ...]` message instead? would prevent non-supporting relays from storing these, and would fit into supporting relays' auth flow better.

staab:
I also went down that path, but it's much hard to add support to relays for new verbs than for new events.

monlovesmango:
for relays that have not implemented the "AUTH" verb, what use case do claims have?

monlovesmango:
> > for relays that have not implemented the "AUTH" verb, what use case do claims have?
> 
> None

then why not use "AUTH" verb instead? if you use "EVENT", relays will have to add access handling to "EVENT" messages and clients are forced to check NIP11.

staab:
It actually adds a lot of complexity. Auth is invalid if there is no claim, and no challenge, but it can be valid if either is valid, but what if one is invalid? Using a separate verb or event is much cleaner.

monlovesmango:
how is "AUTH" with new event kind more complex than "EVENT" with new event kind? the only difference between the two is that "EVENT" handling will now need to inject itself into access control flow and for un-authenticated connections it will now have to check the event kind before being able to reject the event.

staab:
That's something that happens anyway when requesting DMs etc for a particular user, auth is triggered based on event inspection.

monlovesmango:
yes!

staab:
Interesting. I suppose that doesn't seem any worse than anything else, but I don't think we can assume that relays will handle a new event kind properly. @fiatjaf told me, and I think I agree, that he thinks the `AUTH` verb was a mistake (at least the one that clients publish), because it doesn't really carry any semantics on its own. This is especially true if we add more kinds to the auth verb. I'm inclined to use EVENT, and not expand the weirdness of what AUTH means.

alexgleason:
I was planning to also do something very similar to this. But I was thinking of doing a DVM flow so you can get feedback events including the ability to charge for access.

staab:
You're saying the DVM would be *in* the relay? I like that a lot.

alexgleason:
> You're saying the DVM would be _in_ the relay? I like that a lot.

Yes exactly. It would work really well for this. Invite code is an optional param. Feedback events are also basically needed for a flow like this, because they could demand an access code (or not), or tell you if the access code expired, or any number of other things. Then you also get the "success" event baked into the flow.

Check also: https://github.com/nostr-protocol/data-vending-machines/pull/25 

In Ditto, a traditional "registration" would be a combination of a NIP05 request and a relay access request.

vitorpamplona:
Doing authorization tasks via events makes more sense than coding separate schemes in the Nostr protocol. `Claim` is an authorization protocol, not authentication. Authentication probably should remain in the core protocol, but authentication can be managed by a separate DVM-like thing. 

But maybe the DVM spec is too general and not general enough for authorization. Making a NIP designed to run "jobs" into a more general "bot that replies to stuff" is challenging and one of the reasons NIP-46 and NIP-47 are not DVMs.

vitorpamplona:
Yes, but HTTP-like calls fit just the bare minimum of computer science needs. In practice, it's more like forcing everything to be async, atomic, and UDP-like (DVMs might not reply to a request). Sure, devs can make anything work in that setting, you just need to deal with the side effects manually. 

For instance, there is no concept of a transaction involving multiple calls. The way we solved it via DVMs was to have a temporary replaceable event kind to assemble the atomic call. Then multiple DVM requests change that event. Clients can watch the event changing to make sure the DVM processed the request and when everything is ready, hit the button and make another DVM run it atomically. It's like coding in Assembly in 2024. 

Authorization in production is one of these things that frequently requires adjusting many permissions in an all-or-nothing way.

alexgleason:
@vitorpamplona I think I understand what you're saying, but is it actually a response to using a NIP-90 flow for registering an account on a relay? If so I don't get the objection to it, if any. It seems like you're just saying you can't use DVMs to do literally everything. But you can use it for this.

vitorpamplona:
Yes. It's not an objection. It's just like we sometimes use the word DVM to represent any type of Bot in Nostr for lack of a better term. And blindly applying the concept of DVMs into these bots can sometimes create more problems than solutions.

vitorpamplona:
However, I have to say that I don't like the idea of bloated relays (relay + DMV together). I think they are inevitable for some applications but it's way too centralizing, giving too much power to relay operators, for my taste.

alexgleason:
DVM design is modeled after AI use-cases, but the 59xx range is appropriate to solve common Nostr problems outside of what regular clients can do. We already have scheduled posting in there: https://github.com/nostr-protocol/data-vending-machines/blob/master/kinds/5905.md 

There are only 100 numbers in the range. But I think "relay access request" and "NIP-05 registration" are widespread enough to deserve inclusion, and NIP-90 offers the best Nostr-native flow for doing it.

I will also add, the DVM doesn't necessarily need to be "baked-in" to the relay. It should just be a bot that updates a pubkey whitelist the relay reads from. It's pubkey should match the NIP-11 pubkey. But it _could_ be baked into a relay.

alexgleason:
@staab Check out: https://github.com/nostr-protocol/data-vending-machines/pull/26

mleku:
just random thought but i already have got relay identities and a filter using khatru's extensible function arrays one is an AuthCheck for OnConnect and the second is a Chat that goes in the StoreEVent array

these can be seen at https://mleku.dev/git/replicatr in the main.go around line 190 you find these two methods referenced (they live inside the /app directory)

there is the requirement for the relay to have an identity key, and a user then follows the relay ID (just make a link to their account profile) and then you message it, it would be simpler and not require touching the NIPs at all to just make it into a chat control feature

the reason why i am constantly nagging everyone to get their clients supporting - PROPERLY - NIP-42 is precisely to enable this kind of control interface, first and foremost, the metadata leak prevention is secondary to my reasons

with this you then just plug in an ACL system through the chat (which i am also in the process of doing) and you could easily have it so you just message the relay ID with your invite code and it automatically adds you to the ACL and voila

no nips need to be tweaked for this

i considered making special events that would maybe need to eventually go into the NIPs but they can just be internal and not necessarily go on the NIPs at least not right away, that would be internal only, require NIP-42 to function outside of the chat interface, and let you store the ACL changes as special kind events

this has a universal applicability, i'd be really chuffed if we could get this to become a standard thing because it would make relay management so much simpler, you can do all kinds of cool stuff with a secured chat interface

also, yes, it does have a workaround to prevent replay attacks, if the user hasn't already been authed via NIP-42 in the DM it can send a challenge string with instructions to respond with it and then it auths the connection

staab:
It seems simpler, but you're really creating an unspecified sub-protocol nested inside chat. Which, if it's to be interoperable, needs to be specified anyway. So we may as well specify it at the top level.

mleku:
> It seems simpler, but you're really creating an unspecified sub-protocol nested inside chat. Which, if it's to be interoperable, needs to be specified anyway. So we may as well specify it at the top level.

it's just a CLI interface

it's too simple to need complex specification when it can be self documenting as well

i say don't bloat the spec if you don't have to, the smaller the spec the better it is supported

mikedilger:
Using EVENT with an ephemeral kind and not having any searchable tags makes the event less likely to be broadcast to other client's REQs, but of course the NIP should also say not to do it (some relays will serve a request for just kind).

This logic transfers over to local delete. Instead of a new verb, it can also be an ephemeral event with no single-letter tags which the relay shouldn't broadcast (if it does it's no huge issue though).

I think this is a reasonable PR, so +1. Is it implemented somewhere?

fiatjaf:
I like this but now I don't understand what is the difference between 28934 and 28935. Also how can you claim access to anything with an invoice? If you're paying to join the relay shouldn't the relay be giving you the invoice and you be paying it?

staab:
They're basically the same event, and you could use the same kind for both, but semantically they're different so I opted for a new kind. In the case of an invite, you'd do a REQ for a 28935, share the url+claim (rather than the event), and then it could be used as the payload of a 28934 sent to that same relay. For invoices, you'd do the same thing, but pay the invoice before sending the claim. The relay would then look up the invoice and see whether it was paid. But I'm not sure about that flow, it's not something I'm currently working on, and overloading the claim field would require parsing the claim, which seems like a bad idea to me.

staab:
I've removed the invoice stuff, we can add new use cases with new tags if we need them.

fiatjaf:
So Bob does a REQ for a 28935, the relay supposedly gives him a thing it just generated on the fly, Bob takes the claim tag out of that, then shares it with Alice, Alice makes a 28935 with that claim tag and sends to the relay?

Looks pretty good to me.

staab:
Updated to include membership lists, join/remove events, and leave requests (like nip 29).

staab:
This NIP is fully supported by https://github.com/coracle-social/flotilla and https://github.com/coracle-social/zooid

fiatjaf:
I don't understand one thing: if you get an invite 28934 by querying the relay, how can you get a 28935 later in order to use that invite? The spec says the 28935 must be signed by the relay key.

staab:
It sounds like you have the kinds backwards, 28935 are requested from the relay and include an invite code, which can later be used as the `claim` in a 28934 which is signed by the user and sent to the relay.

fiatjaf:
You are totally right!

I got confused, but still my question is: who signs the event that is sent when claiming the access?

fiatjaf:
Should clients rely on 43 being signaled in NIP-11 before they display a "create invite" button?

staab:
> I got confused, but still my question is: who signs the event that is sent when claiming the access?

The relay who wants to issue a claim is the one who signs the `28935` with the relay's `self` pubkey. The `28935` should include a newly generated (or cached, or static) `claim` tag that can be shared. The user who wants to gain access is the one who signs the `28934` with the previously shared `claim` tag.

> Should clients rely on 43 being signaled in NIP-11 before they display a "create invite" button?

It depends, in flotilla I generate an invite link without a claim if nip 43 isn't supported and tell the user who requested it that "additional steps might be required".

43.md
```diff
@@ -0,0 +1,142 @@
...
+```
+
+## Join Request
+
+A user MAY send a `kind 28934` to a relay in order to request admission. It MUST have a `claim` tag containing an invite code. The event's `created_at` MUST be now, plus or minus a few minutes.
+
+```jsonc
+{
+  "kind": 28934,
+  "pubkey": "<nip11.self>",
```
fiatjaf:
```suggestion
  "pubkey": "<user>",
```

43.md
```diff
@@ -0,0 +1,142 @@
...
+  "tags": [
+    ["-"],
+    ["claim", "<invite code>"],
+  ],
+  // ...other fields
+}
+```
+
+Note that these events are in the `ephemeral` range, which means relays must explicitly opt-in to this behavior by generating claims on the fly when requested. This allows relays to improve security by issuing a different claim for each request, only issuing claims to certain users, or expiring claims.
```
fiatjaf, line 128:
```suggestion
Note that these events are in the `ephemeral` range, which means relays must explicitly opt-in to this behavior by generating claims on the fly when requested. This allows relays to improve security by issuing a different claim for each request, only issuing claims to certain users, or expiring claims.

Clients MUST only offer the ability to generate invite codes if the relay advertises support for `43` in its `nip11.supported_nips` array.
```

staab:
Pushed an update with slightly modified versions of your suggestions

fiatjaf:
This is implemented in https://github.com/fiatjaf/pyramid now.

PR #1082, opened 2024-02-25
NIP43 Authenticatio Delegation

monlovesmango:
NIP to allow delegation of NIP42 authentication to other key pairs. can be used to create login tokens for relay access or grant permissions to restricted events (which is another pr). 

https://github.com/monlovesmango/nips/blob/NIP43-auth-delegation/43.md

### DIFF
--- a/43.md
+++ b/43.md
@@ -1,0 +1,159 @@
+NIP-43
+======
+
+Delegated Authentication of Clients to Relays
+-----------------------------------
+
+`draft` `optional`
+
+
+This NIP defines how authentication can be delegated so that a user or client can authenticate with a relay using other keypairs.
+
+## Motivation
+
+A user may want to delegate authentication to another keypair in order to selectively grant access to relay restricted resources. For example,
+
+ - A user may want to login to clients with a NostrConnect remote signer over an authenticated relay that requires membership to send ephemeral events
+ - A user may want to authorize others to access a particular [NIP-71 Restricted Event](https://github.com/nostr-protocol/nips/blob/master/71.md) in exchange for payment or other services
+ - A user may want to create subscription services to certain kinds of [NIP-71 Restricted Event](https://github.com/nostr-protocol/nips/blob/master/71.md) in exchange for payment or other services
+
+## Terminology
+
+- **delegator**: The user who would like to delegate authentication to another key pair.
+- **delegatee**: The alternate key pair the user has delegated authentication to.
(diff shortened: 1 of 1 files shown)


arthurfranca:
A `#<one-letter-tag>` condition could be useful.

For example an event with `["s", "platinum-subscribers"]` tag only meant for users holding a token with a `{ #s: ["platinum-subscribers"] }` condition.

monlovesmango:
yeah perhaps, but would also need to be integrated into restricted events NIP.

I think both are already introducing enough complexity so I would prefer not to add more. however, I did make the 'mode condition' in this NIP specifically so we can extend this to allow for different types of permissions to restricted events in the future.

arthurfranca:
I mean, still considering what you said "The filter condition ALWAYS IMPLICITLY includes the filter criteria "authors": ["${delegator pubkey}"].

So it would still be limited to delegator events which include such `s` tag value.

Your NIP-43 PR wouldn't need to change.

monlovesmango:
yes, true. was just thinking about this myself. actually don't think I would need to change anything other than add it as another option to the filter condition right?

43.md
```diff
@@ -0,0 +1,159 @@
...
+
+#### Filter Condition
+
+The filter condition is OPTIONAL and only applies if the mode condition is set to `1`. It defines what a `REQ` sent by a delegatee can filter for within [NIP-71 Restricted Events](https://github.com/nostr-protocol/nips/blob/master/71.md). The filter condition is a JSON stringified filter object as defined in [NIP-01](https://github.com/nostr-protocol/nips/blob/master/01.md), however only the following attributes are valid within the filter condition:
+
+- `"ids"`
+- `"kinds"`
+- `"since"`
+- `"until"`
+
```
arthurfranca, line 76: Here, this would ne enough

```suggestion
- `"#<single-letter (a-zA-Z)>"`

```

monlovesmango:
really like the idea @arthurfranca ! being able to allow access to different subscription levels or different sets of content is so neat.

staab:
How is this different from NIP 26? I would like users to be able to log in as a separate pubkey, but without revealing their identity, which this doesn't support. The way I imagine that is you would claim a badge with a one-off private key, then authenticate (with a relay, third party service, whatever) using that key to prove you have the badge. You could reveal your main pubkey or not in that scenario.

PR #1083, opened 2024-02-25
NIP71 Restricted Events

monlovesmango:
NIP to create restricted events which can only be requested by authenticated users. Can be used to limit metadata leakage or grant access to paid content.

references to NIP-43 are for pr #1082 (https://github.com/monlovesmango/nips/blob/NIP43-auth-delegation/43.md)

nice version: https://github.com/monlovesmango/nips/blob/NIP71-restricted-events-by-tag/71.md

most seemed to like the tag based identification of restricted events so that what I did, but also created a range based version: https://github.com/monlovesmango/nips/blob/NIP71-restricted-events-by-range/71.md


fiatjaf:
How is this better than https://github.com/nostr-protocol/nips/pull/1030?

arthurfranca:
> How is this better than https://github.com/nostr-protocol/nips/pull/1030?

Not better nor worse, they have different uses.
#1030 prevents rebroadcasts from people that aren't the event author; this one doesn't.
This one prevents people that aren't the author or p-tagged from downloading the event; #1030 doesn't.

arthurfranca:
I'm ok with this solution to event access control. Both this one and #1049 achieve the same goal of adding event access control in a way that enables relays to know if user (and which user) should authenticate by simply inspecting the request filter.

Main difference is this one uses a new filter field while #1049 a new kind range.

monlovesmango:
>How is this better than https://github.com/nostr-protocol/nips/pull/1030?

its not better, its just supposed to be complementary.

PR #1087, opened 2024-02-26
Fully-editable Kind1

vitorpamplona:
This opens the gates of hell for editable kind1 clients. 

Read [here](https://github.com/vitorpamplona/nips/blob/editable-kind1/37.md)

Recap of the 3 options: 
- Updates kind1 to accept edits with full history #1090
- Updates kind1 to accept edits #1089 
- Creates a new content-modifiable short note kind #1088 
- Creates a new fully modifiable short note kind #1087

PR #1089, opened 2024-02-26
Content-editable kind1s without creating a new kind1. 

vitorpamplona:
This option applies content updates directly into kind1. 

Instead of creating separate short post clients with a different kind, this one modifies the behavior of the current kind1. 

Read [here](https://github.com/vitorpamplona/nips/blob/content-editable-kind1-2/37.md)

Recap of the 3 options: 
- Updates kind1 to accept edits with full history #1090
- Updates kind1 to accept edits #1089 
- Creates a new content-modifiable short note kind #1088 
- Creates a new fully modifiable short note kind #1087

### DIFF
--- a/37.md
+++ b/37.md
@@ -16,7 +16,7 @@ This NIP adds content modification directly to `kind:1` short notes.
   "kind": 31010,
   "pubkey": "<32-bytes hex-encoded public key of the event creator>",
   "tags": [
-    ["d", "<kind_10_event_id>"]
+    ["d", "<kind_1_event_id>"]
   ],
   "content": "this is a modified note",
   // ...other fields
--- a/37.md
+++ b/37.md
@@ -1,0 +1,26 @@
+
+NIP-37
+======
+
+Editable Short Notes
+--------------------
+
+`draft` `optional`
+
+This NIP adds content modification directly to `kind:1` short notes. 
+
+`kind:31010` stores the updated content with a `d` tag pointing back to the `kind:1` id. 
+
+```js
+{
+  "kind": 31010,
+  "pubkey": "<32-bytes hex-encoded public key of the event creator>",
+  "tags": [
+    ["d", "<kind_10_event_id>"]
+  ],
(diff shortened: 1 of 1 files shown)


arthurfranca:
+1 to this version

Using a regular event as anchor does solves the parameterized replaceable event problem of making the `.created_at` mean "updated at" which breaks the since/until filtering expectations.

This one is also backwards-compatible.

vitorpamplona:
> This one is also backwards-compatible.

I don't know. Clients that don't support edits, will display the wrong version. So, I wouldn't call it backward compatible. 

The other PRs do it without changing kind 1. Those posts won't show in current clients, but no other new kind would as well. That's just the expected behavior at this point.

arthurfranca:
The other PRs downside is that they split the microblogging scene. Old clients will appear as broken cause they would miss more and more notes as time passes and people start using clients that have note editing feature.

staab:
As usual, I object to replaceables on the grounds that they destroy information. I would prefer non-replaceable events so that it is possible to find out which version of an event someone replied to. Also NACK kind 1s are beautiful as they are. "Author comments" would be a better approach, because they would not need to be fetched by old clients (since the original content is immutable), but authors could still make corrections.

alexgleason:
> I object to replaceables on the grounds that they destroy information.

That's an implementation detail, though. `{ kinds: [0], authors: [alex] }` could easily return multiple versions on the relay side. However, `{ kinds: [0] }` should return only the latest profile for each user.

staab:
> could easily return multiple versions on the relay side.

That's a hack to get around the awful semantics of replaceables. If we want something like this we should introduce "projections" as a first-class feature (which I also object to if they're going to be added to relays, but would be fine to add via DVMs or other services).

vitorpamplona:
Ok, PR https://github.com/nostr-protocol/nips/pull/1090 does it without replaceables.

vitorpamplona:
Closing in favor of #1090

PR #1092, opened 2024-02-26
Initial draft of the nostr-spaces proposal

ilmaoun:
The following proposal is an ambitious attempt to bring the twitter/X spaces feature into the nostr network. I believe we're on the verge of drastically changing the landscape of social networks, and this nip is but a small step in the path to the new communication revolution.

The Nostr Spaces Protocol (NIP) is designed to extend the capabilities of the Nostr ecosystem by introducing a framework for real-time peer-to-peer audio broadcasting. Leveraging the power of WebRTC for media transmission and utilizing Nostr for signaling, the protocol facilitates decentralized, secure, and scalable audio communication channels. This document outlines the specifications of the Nostr Spaces Protocol, detailing its architectural design, operational mechanisms, event types, and interactions between entities within the system. The objective is to provide developers, relay operators, and users with a clear understanding of how to implement, interact with, and benefit from Nostr spaces dedicated to audio broadcasting.

I have started working on an implementation of this nip using JavaScript but I must admit, since I'm a backend engineer who have mainly worked with Java, the quality of the code might not be the best, and a large chunck of it is still to be implemented and will be reviewed as the project advances.

If anyone wants to help me out, here is the github repository: https://github.com/ilmaoun/nostr-spaces-tools

I'm looking forward for your feedback and suggestions.


MrZarp:
have you seen https://nostrnests.com/?

ilmaoun:
> have you seen https://nostrnests.com/?

Yes, checkout my NIP, I don't think it's the same approach. What I'm suggesting is using relays as signaling servers. The space network will be maintained by the peers exclusively.

ilmaoun (commented):
Nostr spaces nip

staab:
Do you have something built? I would strongly suggest starting with what nostr nests has and build incrementally from there, rather than creating an entirely separate standard.

ilmaoun:
> Do you have something built? I would strongly suggest starting with what nostr nests has and build incrementally from there, rather than creating an entirely separate standard.

Hello, 
I’m still working on the implementation of the library but I’m close to the testing phase.
The approach I suggest is a bit different than nostr nests. The idea is not to have a backend at all, all the action happens on the client side. The different peers collaborate with each other to broadcast the audio to the whole network.

fiatjaf:
The problem is that this approach doesn't work and will lead to a very poor experience, because WebRTC doesn't work, you really need a server for peers to connect to.

ilmaoun:
> The problem is that this approach doesn't work and will lead to a very poor experience, because WebRTC doesn't work, you really need a server for peers to connect to.

Can you please elaborate on that? Why do you think it wouldn’t work? Have there been any similar attempt that failed?

Issue #760, opened 2023-09-04
Referencing Podcasts in Nostr Events


MerryOscar:
It would be great if nostr apps / clients / services could reliably reference podcasts in nostr events.

This would enable:

- using nostr for cross-app podcast comments
- using nostr for livestream chat that can be surfaced in podcast apps
- clipping / highlighting services to reference the source material
- nostr clients to open podcast links in the user's app of choice (_e.g. fountain link 1-click open in overcast_)

Because podcasting is fragmented across many apps and service providers - the only real way to reliably reference a show or episode is to use the `<podcast:guid>` and the `<item>` `<guid>` [podcast namespace](https://podcastindex.org/namespace/1.0#guid) elements which exist in the rss feed. The `guid` at both the feed and episode level are supposed to be honoured when podcasters switch hosting providers meaning that any external references to these guids won't break if the podcast moves to a different rss feed url.

From reading through the NIPs I think there are 3 ways to potentially do this:

---

### Option 1 - `r` Tags:

```
{
  "tags": [

	# podcast feed reference
        [“r”, “podcast:guid:123”]

        # podcast episode reference
        [“r”, “podcast:item:guid:1234”]
  ],
  "content": "This was a great episode!",
  ...
}
```

- Benefits:
  - very simple
- Drawbacks:
  - kind of arbitrary?

---

### Option 2 - [NIP-32](https://github.com/nostr-protocol/nips/blob/master/32.md) Labels:
```
{
  "kind": 1,
  "tags": [

	# podcast feed label
	["L", "podcast:guid"],
	["l", "123", "podcast:guid"],

	# podcast episode label
	["L", "podcast:item:guid"],
	["l", "1234", "podcast:item:guid"],

  ],
  "content": "This was a great episode!",
  ...
}
```

- Benefits:
  - specifically linked to a defined namespace
- Drawbacks:
  - not the right use of NIP-32?

---

### Option 3 - [NIP-48](https://github.com/nostr-protocol/nips/blob/master/48.md) Proxy Tags:
```
{
  "kind": 1,
  "tags": [

	# podcast feed reference
	["proxy", "podcast:guid:123", "podcast"],

	# podcast episode reference
	["proxy", "podcast:item:guid:1234", "podcast"],

  ],
  "content": "This was a great episode!",
  ...
}
```

- Benefits:
  - specifically linked to a defined namespace
- Drawbacks:
  - not really referring to a protocol

---

### Feedback / Discussion:

[...]

PR #1093, opened 2024-02-28
NIP-F4: podcasts

fiatjaf:
text: https://github.com/nostr-protocol/nips/blob/podcasts/F4.md

These are mostly ideas and if someone has better ones I'm ready to change everything.


alexgleason:
@daveajones is building ActivityPub podcast integration with Lightning. Nostr integration would be more seamless with Lightning.

mterenzio:
There's a bunch of movement in the Podcast 2.0 space and I think it would
be a mistake to try and support a bunch of it, but a transcript tag might
be useful for a number of reasons including being able to index for better
search results given the advantages of decoupling the episodes from the
larger feed

mterenzio:
>Please don't try to replace RSS.

I agree with this 120% . But not sure it's a fair description of what is being proposed. Relays do a lot more than accommodate social interaction IMHO

54.md
```diff
@@ -0,0 +1,80 @@
...
+{
+  "id": "55807e7d5cd90d0303d7dce7397f996fdbaed8697903f326c7cf8ad999b9de3d",
+  "pubkey": "79be667ef9dcbbac55a06295ce870b07029bfcdb2dce28d959f2815b16f81798",
+  "kind": 54,
+  "created_at": 1700682555,
+  "tags": [
+    ["title", "<episode title>"],
+    ["image", "<optional episode image>"],
+    ["description", "<a brief description>"],
+    ["audio", "https://.../", "<optional_media_type>"], // can be specified multiple times
```
staab, line 45: I would suggest `imeta` instead, since it can contain more metadata. Users might want to download a lower-fidelity version of the episode for example.

54.md
```diff
@@ -0,0 +1,80 @@
...
+    ["t", "<optional tag>"], // can be specified multiple times
+    ["alt", "<optional NIP-31 short description for displaying in incompatible clients>"]
```
staab, line 47: There are some other fields in #1043 that might be worth adding (creator, website, duration, published_at). If they're useful for music publishers, they're probably useful for podcast publishers. But they can be added later too.

In particular, I think an `i` tag (or whatever) referring to the RSS version's GUID would be useful for linking duplicate content across protocols. Single-letter, because you might want to find an event by guid.

54.md
```diff
@@ -0,0 +1,80 @@
...
+  "created_at": 1700682556,
+  "tags": [
+    ["e", "55807e7d5cd90d0303d7dce7397f996fdbaed8697903f326c7cf8ad999b9de3d"],
+  ],
+  "content": "<plain text content>",
+  "sig": "...",
+}
+```
+
+(It's unclear if it's best to use this or a generic kind for commenting on stuff that isn't `kind:1`.)
```
staab, line 72: If in doubt, create a new kind

staab (changes requested):
I like this, with some minor changes. I think we're slowly discovering a common pattern for linking to non-native media types.

Edit: I agree with what was said above. RSS should not be replaced, but maybe mirrored/enhanced by bridging it to nostr.

fiatjaf:
> It's been tried many times and always fails.

Can you give me some examples besides JSONFeed?

Maybe because all of the attempts didn't offer anything new, but were just syntax sugar?

In any case, Nostr is trying to replace Twitter, Facebook, Wikipedia, blogs, Instagram, YouTube, Twitch already. Is RSS so much more impossible than this?

fiatjaf:
Also, Atom has successfully replaced RSS (for a percent of the sites) in one of the most stupid moves of the entire internet: they literally made the exact same thing but with different names and got away with it and now every client has to support the two protocols forever, amazing.

The creation of Atom is only comparable to the creation of Deno: the thing that was invented by creator of Node to fix all the problems associated with Node, but after many years ended up in the exact same place as that.

daveajones:
> Can you give me some examples besides JSONFeed?

XMPP was going to kill RSS.  Then AMP.  But, what I meant was mostly API based services.  Twitter is an example of an "RSS killer" because of their API back when it was fully open.  Also YouTube.  Spotify aimed to "move beyond the limitations" of RSS podcasting and have now given up that endeavor and re-embraced RSS by buying Megaphone.  There is also an RSS 3.0 spec that never went anywhere. 

> Is RSS so much more impossible than this?

Nothing is impossible. 😉 Just unlikely enough to make it a big waste of time IMO.

> Also, Atom has successfully replaced RSS...

RSS and ATOM are interchangeable in my mind.  I use the term "RSS" as a catch-all to mean any static uri based XML feed of published items.  I agree that ATOM was dumb.

> Maybe because all of the attempts didn't offer anything new, but were just syntax sugar?

There is a $2 billion dollar podcasting industry built on the creation AND consumption of RSS feeds with years of know-how on both sides.  It's not a question of technical merit.  It's a question of inertia.  RSS podcasting is a 2000 pound flywheel.  It's just not going to change, because after all of the hours of dev work to switch to something new, you'll end up back with essentially the same thing at base level:  publishing audio urls to a podcast app.  

It took us 3 years of advocacy to get people to build new podcast features _on top_ of the existing RSS.  Getting them to replace RSS entirely would be a feat of evangelism to rival the New Testament.  If you can pull it off, you'll have my eternal respect.  But, I've been in the trenches of this for a long time now and I think there are bigger fish to fry.  Sometimes the best format just doesn't win.  HTTP, SMTP, etc.  None of them are ideal, but they are all heavy flywheel's.

mterenzio:
I don't see the potential here as competing with RSS, more like
getting your existing podcasts into a new platform. You wouldn't maintain
anything extra besides setting it up the way you set it up for Apple, e.g.

Once the data is rolling into Apple, or anyone that creates a directory of
podcasts, they don't simply store the XML, they index it and make it
available to their platform in the best way.

In Nostr that's events on relays. It's opt in anyway.

fiatjaf:
> Getting them to replace RSS entirely would be a feat of evangelism to rival the New Testament.

I must say I laughed a lot from this line here, ahahah

I don't know, I agree with everything you said but I still think we can pull it off, but let's give it more time and thought. I agree there are other areas that could be prioritized first and RSS is already good enough.

54.md
KotlinGeekDev: Might it not be better to add (website, creator) as tags to the Kind 0 event?

staab: No, because a single pubkey might publish events on behalf of multiple creators (e.g. wavlake). Website should also be more than a domain name, maybe that's the wrong word.

staab: Ideally yes, but wavlake currently publishes everything under the same key. Maybe it would be better to use a separate key for each creator even if you're publishing on behalf of someone else, but we probably shouldn't force that here.

KotlinGeekDev: I see.

KotlinGeekDev: As for Wavlake, could some form of delegation work, ideally in the manner of @alexgleason 's mostr?

mterenzio:
One of things ATOM did do that the RSS (2.0) spec didn't was move the
channel info into the feed item, so that an item could move away from feed
and still have all channel info inside it.

Eventually RSS 2.0 could also do that by adding the ATOM namespace.

Nostr events are more loosely coupled than an RSS feed.

The point being that there is great advantage to getting the item or event
with all the channel info in it, so they can be aggregated into
multi-source feeds without having to do extra queries back to the channel
(profile here) to get that info

54.md
staab: Yeah, I think use of the `proxy` tag could be appropriate here, although I don't think it needs to be in the spec.

KotlinGeekDev: Got it.

mterenzio:
Discussion about whether trying to replace RSS is a good idea or not aside,
I _do_ have a use case where I'd like to make a feed of URLs.

My initial thinking was to use lists and just replace the list every time
there was a new entry, but if this was more generic or I could
alternatively use a `["type", "feed"] `  that might be better than a
replaceable list, or not

fiatjaf:
What is the use case?

mterenzio:
A service that monitors your timeline and creates a digest of articles that
your followers link to. Nuzzel for Nostr. It's a port of something I built
for Mastodon, but since we are on Nostr, it seems a better fit that the
results would be an event or group of events, rather than an email. Rabble
also suggested a long form content event which also has advantages. I'd
like to support all options

fiatjaf:
Digests of articles people link to? I don't quite get it. Why not just reply to the notes that link to articles with a digest of these articles?

mterenzio:
No, like all the articles shared by your friends (and perhaps friends of
friends) in a nice newsletter type format. See Nuzzel for prior art
https://daringfireball.net/linked/2021/05/05/nuzzel

staab:
It might be a good idea to model this as closely as is reasonable after [podcasting 2.0's spec](https://podcasting2.org/podcast-namespace)

blastshielddown:
> it could be better to just keep using RSS but add Nostr for social interaction on top

This seems simpler.

A thought: [NIP-84](https://github.com/nostr-protocol/nips/blob/master/84.md) (or "highlights") could be extended for a use case like this, when you want to reference something that has a relatively stable URI that is not a Nostr event but you want to enable interactions around it. For podcasts, you could point the `r` tag to the feed URL (or better yet, include the podcast UUID per the Podcasting 2.0 spec in a `d` or `i` tag). You also could include references to the episode and more granular details like timestamp in supplemental tags, to point clients to exactly the spot a user wants to share. The NIP-84 spec already makes some references to media so this would follow its original intent.

MerryOscar:
no matter what happens with this proposal or any other proposal that attempts to link nostr to rss - the simplest primitive that we should try and agree on is how to reliably reference podcasts in nostr events.

Because there are so many existing nips that are relevant to podcasts (nip01 notes, nip53 live events, nip38 statuses, nip58 badges, nip84 highlights) I think the easiest way to do this is with tags.

I outlined [some options](https://github.com/nostr-protocol/nips/issues/760) for this last year - and it seems the preferred approach was to use `r` tags to reference podcasts via the existing `guids` that are defined in the rss feed:

```
  "tags": [

	# podcast feed reference
        ["r", "podcast:guid:123"]

        # podcast episode reference
        ["r", "podcast:item:guid:1234"]
  ]
```

pablof7z:
I just skimmed a bit of this very funny discussion, here are my two sats:

comparing previous attempts to replace RSS for podcast distriibution with replacing it with nostr is not a good comparison:

Replacing with nostr is not simply a schema modification that might be simpler/saner/faster/more-extensible than RSS 2.0. A debatable marginal improvement over RSS.

Podcasts on nostr is not about the schema, is about the social graph and network effects of all other use cases. The schema changing is a mere byproduct but leveraging the rest of your social graph is an orders-of-magnitude improvement over RSS.

Publishing podcasting data as nostr events will lead to apps that fully leverage nostr instead of shoehorning some nostr coating on top of existing implementations.

All the use cases we're rebuilding on nostr seem like a Quixotesk pursuit at first, but that's the nature of overtaking any network effect.

MerryOscar:
@pablof7z 

> Podcasts on nostr is not about the schema, is about the social graph and network effects of all other use cases. 

You still get the social graph and network effects if you use nostr as the social layer but keep rss as the hosting mechanism. If interacting with podcasts on nostr requires the audio to be hosted on nostr - then you don't get any network effects because 90% of the content is not hosted on nostr

If social features in a nostr podcast app only work for <10% of the shows then it's a terrible experience for users

If you can allow nostr social features to apply to _any_ existing podcasts then you can provide a great experience that doesn't require any buy-in from the existing podcast industry - it just works out of the box

There are 400m+ podcast listeners that are using podcast apps with zero social features and nostr is a great solution for them - but if using these features only applies to a tiny subset of what they listen to then it's going to be pretty hard to get them to use it

We use `r` tags for linking to websites so why not use them to link to podcasts which are just xml websites

mterenzio:
>
> We use r tags for linking to websites so why not use them to link to
> podcasts which are just xml websites


What you you be referencing once the episode is no longer in the feed? I
recognize it's a guid but for it to be meaningful you need the surrounding
context and metadata, which is basically duplication of the RSS no?

pablof7z (approved):

PR #1100, opened 2024-03-05
BREAKING.md: add NIP-18 change

AsaiToshiya:
No one seems to care, but a list that is not updated has no value.


mikedilger:
I care. And I didn't know that was merged so now I'm off to switch up my code.  Thanks.

PR #1103, opened 2024-03-06
Add DM relays to nip 65

staab:


fiatjaf (approved):

fiatjaf:
If I understood correctly this is great, let's ship it.

65.md
```diff
@@ -6,11 +6,9 @@ Relay List Metadata
 
 `draft` `optional`
 
-Defines a replaceable event using `kind:10002` to advertise preferred relays for discovering a user's content and receiving fresh content from others.
+This NIP defines a series of replaceable event kinds for advertising user relay preferences, each with a particular purpose. These events MUST include a list of `r` tags with relay URIs. The `.content` field is not used.
```
dluvian, line 9: > These events MUST include a list of `r` tags

I think there was a discussion to deprecate r tags because of indexing. If we introduce new kinds we should go with `relay` instead of `r` imo

65.md
```diff
@@ -26,39 +24,53 @@ The `.content` is not used.
...
-When broadcasting an event, Clients SHOULD:
+When seeking private or encrypted events authored by a user, Clients SHOULD use the `outbox` relays defined by
+the author's `kind:10050` event, falling back to the `write` relayd defined by author's `kind:10002` event.
 
-- Broadcast the event to the WRITE relays of the author
-- Broadcast the event all READ relays of each tagged user
+When broadcasting a public event, Clients SHOULD:
+
+- Broadcast the event to the `write` relays of the author
+- Broadcast the event all `read` relays of each tagged user
```
dluvian, line 52:
```suggestion
- Broadcast the event to all `read` relays of each tagged user
```

65.md
```diff
@@ -26,39 +24,53 @@ The `.content` is not used.
...
-
-4. DMs SHOULD only be broadcasted to the author's WRITE relays and to the receiver's READ relays to keep maximum privacy. 
-
-5. If a relay signals support for this NIP in their [NIP-11](11.md) document that means they're willing to accept kind 10002 events from a broad range of users, not only their paying customers or whitelisted group.
+## Final Considerations
 
-6. Clients SHOULD deduplicate connections by normalizing relay URIs according to [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986#section-6).
+1. Clients SHOULD guide users to keep relay lists small (2-4 relays).
+2. Clients SHOULD spread an author's relay metadata events to as many relays as viable.
+3. Relay lists events should primarily be used to advertise the user's preferred relays to others.
```
dluvian, line 73:
```suggestion
3. Relay list events should primarily be used to advertise the user's preferred relays to others.
```

dluvian:
You forgot to add `kind 10050` to the [Event Kinds table](https://github.com/nostr-protocol/nips?tab=readme-ov-file#event-kinds)

65.md
staab: I'm willing to change it, it's annoying though because it introduces an inconsistent use of tags in a single NIP. It might be useful to have

vitorpamplona:
I am good with this. I don't really know why using two kinds but sure.

staab:
> Came to the conclusion that users don't need `dm` relays. Simply pick good `read` relays that also treat DMs well (NIP-42).
> 
> Besides that, two lists and write/outbox usage overlap contribute to make things more complex for little gain.

I actually agree with this. The benefits of a separate list is that you can avoid broadcasting messages to more relays than necessary, enhancing privacy. The downside is that clients that don't support the new list won't know to look elsewhere for DMs, and will start missing them. @fiatjaf can you make the case for this change improving deliverability rather than making it worse?

fiatjaf:
I've been talking about using specific DM relays for many months, and my entire hope for NIP-44 + gift-wraps was that we would change to using specific DM relays.

As far as I know only Amethyst and Coracle implement the new DM scheme and it isn't specified anywhere so this shouldn't break anything for NIP-17.

staab:
Let's add the new relay kind to NIP 17 instead @vitorpamplona. Where to save hints should be defined by the relevant NIP, not added to NIP 65. That limits the blast radius, because otherwise you end up with tricky compatibility issues like the one I mentioned.

fiatjaf:
@arthurfranca relays should be able to specialize. We gain literally nothing from just merging together two entirely different functions in the same group of relays. If we can specialize, why not?

Also the presence of the new kind serves as a signal that this user is ready to accept NIP-17 DMs, otherwise the user shouldn't even be contacted.

staab:
> Also the presence of the new kind serves as a signal that this user is ready to accept NIP-17 DMs, otherwise the user probably shouldn't even be contacted.

This is a great idea

vitorpamplona:
ohh, so the new event kind is for DMs only and not any GiftWrap? I don't think I see a difference between DMs and any other GiftWrap payload in terms of privacy or which relay to use. They are all private.

staab:
NIP 87 uses gift wrap, but they're not addressed to users, but shared keys, so there wouldn't ever be a NIP 65 available. Instead, the group meta's relays should be used. That's a pretty good example of appropriate relay selection being defined in the relevant NIP.

vitorpamplona:
Cool I am going to add this over there. 

But here's the catch, for DMs, there is no Outbox. Everything is an inbox.

fiatjaf:
> But here's the catch, for DMs, there is no Outbox. Everything is an inbox.

Good point.

Everything is an inbox and you can just connect to that single relay and also publish your outgoing DMs to it as well as access everything you receive from it.

A specialized relay could also allow you to delete directly from it even though you're not the immediate author of the events, it becomes your personal inbox.

vitorpamplona:
added: https://github.com/nostr-protocol/nips/pull/686/files

arthurfranca:
> relays should be able to specialize. We gain literally nothing from just merging together two entirely different functions in the same group of relays. If we can specialize, why not? Clients can easily automate the creation of these lists.

Just that don't see much to specialize in when relaying DMs compared to what a `read` relay already provides rather than requiring `AUTH` when client requests gift-wraps (or kind:4).

> Also the presence of the new kind serves as a signal that this user is ready to accept NIP-17 DMs

The user can accept NIP-17 dms if it is currently using a client that supports it. A list won't guarantee anything imo. It depends solely on NIP-17 adoption ratio.

vitorpamplona:
> Just that don't see much to specialize in when relaying DMs compared to what a read relay already provides rather than requiring AUTH when client requests gift-wraps (or kind:4).

I think the main issue is really the difference between a public `read` (no auth) and a private `inbox` (authed). Authed relays can do both `read` and `inbox`. Non-authed relays should only do `read`s (only public events).

vitorpamplona:
Actually it's about 3 relay types: 

Public `read` (anyone can download events that p-tagged you). Public `read`s are used when the author's `write` doesn't have the post. The client should start looking at other relays from the p-tag list. It's important that this type of relay remains open/unauthed. But this relay can accept only events by or about a given public key. And for privacy, it should not return DMs or GiftWraps until the connection is authed by the author. 

Private `inbox` relays might require auth to download any event: They are not supposed to be used by third parties to find posts. They can accept posts by an author or where the author is p-tagged, but they do not respond unless authed. 

While `read` can be seen as a superset of `inbox`, there might be use cases where the author doesn't want to allow the download of public events (Zaps for instance). 

So, in practice, the two relay types intersect, but each has something unique going for them.

PR #1105, opened 2024-03-06
NIP-50: search by nip05 domain

alexgleason:
Add a search extension to NIP-50 like `domain:gleasonator.dev` to include only events from users with a valid nip05 name on the domain `gleasonator.dev`

### DIFF
--- a/50.md
+++ b/50.md
@@ -47,3 +47,4 @@ Relays SHOULD exclude spam from search results by default if they support some f
 
 Relay MAY support these extensions:
 - `include:spam` - turn off spam filtering, if it was enabled by default
+- `domain:<domain>` - include events from users whose valid nip05 domain matches the domain


fiatjaf (approved):

nostr-wine:
We have recently started experimenting with event classification (https://docs.nostr.wine/api/classification) on our nostr.wine relay.

I'm not sure if this is the right place, but is this an opportunity to include a few additional search extensions for those? We'd love to find a way to make topic and other advanced search more accessible on clients. We were thinking topic, sentiment (negative, neutral, positive), language, and a nsfw boolean.

topic:<topic> - include only events within a classified topic (short string, one to two words?)
sentiment:<negative, neutral, or positive> - include only events of a specific sentiment
language:<two letter language code?> - include only events of a specified language
nsfw:<true/false> - include or exclude nsfw events

fiatjaf:
Yes, I think we could include all these.

alexgleason:
@nostr-wine Good idea. Let's open a separate PR for those.

PR #1110, opened 2024-03-07
NIP-87: Ecash Mint Discoverability

benthecarman:
This is mostly just a fork of NIP-89 but changing the wording about application handling to about discovering ecash mints.

Hopefully this can help cashu/fedimint apps discover mints to connect to.


87.md
```diff
@@ -0,0 +1,126 @@
...
+  "content": "<optional-kind:0-style-metadata>",
+  "tags": [
+    ["d", <random-id>],
+    ["u", "https://cashu.example.com", "cashu"],
+    ["nuts", "1,2,3,4,5,6,7"]
+  ]
+}
+```
+
+Fedimints SHOULD publish `kind:38173` events to announce their capabilities and how to connect to them.
```
TonyGiorgio, line 66: There's nothing here that really connects the mint broadcast event to the recommendation notes, besides one of the invite strings. I don't see any purpose for these mint events if it's not even recommended that it should be queried in the first place and nothing verifiably links to them.

starbackr-dev:
just a question. Can we make this generic for any outside application can use? NIP-89 is Nostr specific Apps and NIP-87 is rest of the world? Can potentially used by App Store, AI agents, and any other non nostr Apps.

87.md
benthecarman: The only real way to link them would be to tag the event id, however, then you cannot update the mint's event if anything changes (new features, fees, etc)

benthecarman: The reason I did it this way so that you can recommend a mint without it announcing itself. Also it saves on the required number of queries to be able to find a mint.

However you do bring up a good point of being able to point to the mint's event. Kept the `u` tag stuff but also added a NIP-89 style `a` tag to be able to recommend based on the mint's event.

staab:
If there were an e-cash specific nostr event kind, you could just use NIP 89. You could even make up a fake kind and use that to signal ecash mint support.

benthecarman:
> If there were an e-cash specific nostr event kind, you could just use NIP 89. You could even make up a fake kind and use that to signal ecash mint support.

need a way to pass around fedimint invite codes/cashu urls, you can't encode that in an event kind

staab:
I see, maybe this is different enough from NIP 89 to warrant a new nip. You could add a new `mint_url` tag or something though.

benthecarman:
> You could add a new mint_url tag or something though.  

that is what the `u` tag is for. Needs to be a single letter so it is indexed by relays

87.md
```diff
@@ -0,0 +1,129 @@
...
+Nostr's discoverability and transparent event interaction is one of its most interesting/novel mechanics.
+This NIP provides a simple way for users to discover ecash mints recommended by other users and to interact with them.
+
+### Parties involved
+
+There are three actors to this workflow:
+
+* An ecash mint operator, announces their mint and its capabilities.
+    * Publishes `kind:38173`, detailing how to connect to it and its capabilities.
+* user A, who recommends an app that handles a specific event kind
```
TonyGiorgio, line 22: You have quite a few places that still uses the word `app` or `application` when you should be using `mint`

87.md
```diff
@@ -0,0 +1,129 @@
...
+```
+
+* `content` is an optional `metadata`-like stringified JSON object, as described in NIP-01. This content is useful when the pubkey creating the `kind:38173` is not an application. If `content` is empty, the `kind:0` of the pubkey should be used to display application information (e.g. name, picture, web, LUD16, etc.)
+
+## User flow
+A user A who uses a non-`kind:1`-centric nostr app could choose to announce/recommend a certain kind-handler application.
+
+When user B sees an unknown event kind, e.g. in a social-media centric nostr client, the client would allow user B to interact with the unknown-kind event (e.g. tapping on it).
+
+The client MIGHT query for the user's and the user's follows handler.
```
TonyGiorgio, line 95: All of this needs updating for using `mint` instead of application.

87.md
```diff
@@ -0,0 +1,129 @@
...
+  "kind": 38173,
+  "pubkey": "<application-pubkey>",
+  "content": "<optional-kind:0-style-metadata>",
+  "tags": [
+    ["d", <random-id>],
+    ["u", "fed11abc..", "fedimint"],
+    ["u", "fed11xyz..", "fedimint"],
+    ["modules", "lightning,wallet,mint"]
+  ]
+}
```
TonyGiorgio, line 85: Should the federation ID be referenced and published here in some way?

Federations invite codes can change and there could be many. It is going to be very difficult to aggregate and parse them all when they could be referencing the same federation. The federation ID is what the user is "joining" not just a single invite code.

87.md
```diff
@@ -0,0 +1,129 @@
...
+```json
+{
+  "kind": 18173,
+  "tags": [
+    ["u", "fed11abc..", "fedimint"],
+    ["u", "https://cashu.example.com", "cashu"]
+  ],
+  ...
+}
+```
```
TonyGiorgio, line 111: Same thing here. The user recommends a federation mint. Not a federation invite code.

fiatjaf:
I don't see why not use NIP-89 and add some extra tags in there. NIP-89 is already being used for multiple weird things that aren't Nostr clients properly, like DVMs and custodial bunker signer servers.

benthecarman:
> I don't see why not use NIP-89 and add some extra tags in there. NIP-89 is already being used for multiple weird things that aren't Nostr clients properly, like DVMs and custodial bunker signer servers.

DVMs and nsec bunkers fit well into here because they are nostr native, they can just say they handle a specific event kind. This doesn't work as well for this use case because we'd have to assign a fake kind number to represent this which is ends up being basically the same as this, creating a new kind number

benthecarman:
After implementing this made some changes:

- split cashu and fedimint events, don't see the point in making them the same, normally you are going to query for one or the other
- made it so the `d` tag on the announcement event should have meaning (federation id or cashu pubkey)
- cleaned up some text

fiatjaf:
Maybe we should make this into a big "discoverability" NIP and have a table like in NIP-51 with all the possible kinds of things that can be discovered? Starting with Cashu and Fedimint but then it could have other stuff, I have some in my mind already. This could actually be big.

fiatjaf:
But I think we can merge this as is now and reword later without breaking.

benthecarman:
> I like the 'd' tag as cashu pubkey which should be constant for any mint. Starting to work on this to create mint review site for cashu.network assuming this will be merged shortly. 

What is the reasoning for that, if it's just a fixed value then it doesn't really have any meaning. May as well assign meaning to it.

staab:
> What is the reasoning for that, if it's just a fixed value then it doesn't really have any meaning. May as well assign meaning to it.

Just in response to this comment, I've learned it's best to keep d tags meaningless. Any semantic piece of information can change, even a pubkey.

benthecarman:
Okay, can change to a different tag then

gudnuf:
Should there ideally only be one info event for each mint, and should that only be published by the mint operator? Is there any way to verify the key publishing the event is related to the mint?
What if I really like a mint, but there is no info event for it, how do I recommend it?

otech47:
does this NIP consider the use case where an ecash user wants to be paid in ecash and wants others to know that these are the federations that they want to receive ecash from? like if i want to pay someone in ecash, can i take someone's npub and query for these event kinds and get the federations i can go join and generate ecash for to send to the person and pay them?

or is it more general just "here i recommend these federations"?

or is this just for federations themselves to advertise their mint? or both?

starbackr-dev:
> Should there ideally only be one info event for each mint, and should that only be published by the mint operator? Is there any way to verify the key publishing the event is related to the mint? What if I really like a mint, but there is no info event for it, how do I recommend it?

Yes. Cashu mint info endpoint has a field called nostr with npub. This npub should match the info event publisher npub. Otherwise we will not even show that as available mint on the mint review site. 
[
      "nostr",
      "npub..."
    ]

but there is no info event for it, how do I recommend it? - it is up to the mint operator to publish the event. if the mint operator wants to keep the mint private, then user cannot recommend it..

starbackr-dev:
> or is this just for federations themselves to advertise their mint? or both?

Both

> or is it more general just "here i recommend these federations"?

yes. This NIP is just for information about the mint and reviews/recommendations from other users. The use case you are mentioning should be implemented by the wallet providers.

gudnuf:
> Yes. Cashu mint info endpoint has a field called nostr with npub. This npub should match the info event publisher npub. Otherwise we will not even show that as available mint on the mint review site. [ "nostr", "npub..." ]

This is reasonable, it should be explicit in the nip.

>  This NIP is just for information about the mint and reviews/recommendations from other users. 

To enhance the "just for information", what do you think about including some extra metadata about the recommendation? Possibly a review (ie. "This mint is awesome!") in the content field of the `18173` and a "stars" tag to rate the overall quality of the mint? I'm thinking in terms of how can clients have more data to rank the mints, rather than just a simple recommendation.

gudnuf:
> wants others to know that these are the federations that they want to receive ecash from

Wallet can also simply swap for a trusted mint once payment is received

otech47:
> Wallet can also simply swap for a trusted mint once payment is received

yes but intra-mint ecash payments saves the user a swap fee so having a good protocol to advertise to others which mint is preferred for receiving can sometimes be cheaper for both sender and receiver?

87.md
```diff
@@ -0,0 +1,129 @@
...
+User B's wallet client queries for `kind:18173` events, looking for recommendations for ecash mints.
+
+```json
+["REQ", <id>, [{ "kinds": [18173], "authors": [<user>, <users-contact-list>] }]]
+```
+
+User B, who follows User A, sees that `kind:18173` event and tries to connect to the corresponding mints.
+
+### Alternative query bypassing `kind:18173`
+Alternatively, users might choose to query directly for `kind:38173` for an event kind. Clients SHOULD be careful doing this and use spam-prevention mechanisms or querying high-quality restricted relays to avoid directing users to malicious handlers.
```
alexlwn123, line 125: Would it make sense for `kind:38173` to contain some kind of attestation from the mint custodians to independently verify its authenticity? 

Could be optional if it's a hassle for some mints.

87.md
```diff
@@ -0,0 +1,129 @@
...
+
+```json
+{
+  "kind": 38173,
+  "pubkey": "<application-pubkey>",
+  "content": "<optional-kind:0-style-metadata>",
+  "tags": [
+    ["d", <federation-id>],
+    ["u", "fed11abc.."],
+    ["u", "fed11xyz.."],
```
elsirion, line 87: An invite code is just a bech32 encoded federation ID and a list of API urls. The proposed encoding is thus highly redundant: each invite contains the federation id again and possibly overlapping API URLs. Instead I'd recommend to go with:

```
    ["d", <federation-id>],
    ["u", "wss://guardian-1.url"],
    ["u", "wss://guardian-2.url"],
```

We do include the guardian id for each URL in the invite, but I don't think it's strictly necessary. If we want full future feature-parity (if that ID ever gets used for anything) we could encode the guardian id in the URL announcement.

```
    ["d", <federation-id>],
    ["u", "1,wss://guardian-1.url"],
    ["u", "2,wss://guardian-2.url"],
```

To decode an invite code use `fedimint-cli dev decode-invite-code`:

```
$ fedimint-cli dev decode-invite-code fed11qgqpw9thwvaz7te3xgmjuvpwxqhrzw33xvenvve0qyqjpq7rkttxqknn0p30lsyxnljwqw0ff2l55t7rc6xxgup8zfguc9dhh9d3yd
{
  "url": "ws://127.0.0.1:13363/",
  "federation_id": "83c3b2d6605a737862ffc0869fe4e039e94abf4a2fc3c68c6470271251cc15b7"
}
```

Our support for multiple URLs in dev tooling may still be a bit lacking support for it was added recently. Currently you'd call `fedimint-cli invite-code <guardian-id>` for each guardian and then decode the invite, getting their API URL.

87.md
```diff
@@ -0,0 +1,129 @@
...
+## Example
+
+### User A recommends some mints
+User A might be a user of a cashu mint. Using a client, user A publishes an event recommending the cashu mint they use.
+
+```json
+{
+  "kind": 18173,
+  "tags": [
+    ["u", "fed11abc..", "fedimint"],
```
elsirion, line 104: Why would the recommendation contain the full invite code? Feels like the federation ID would be sufficient since in the second step the user client would query the 38172/38173 event anyway?

87.md
elsirion: > The reason I did it this way so that you can recommend a mint without it announcing itself

If you want recommendations for private mints you should really consider not using invite codes since they always contain the necessary information to join a federation. https://github.com/nostr-protocol/nips/pull/1110/files#r1524194996

87.md
benthecarman: I think this is doable for cashu, they can just use the same key for ecash to sign the nostr event (might just need to invert for xonly stuff). For fedimints I'm not sure if there is an easy way to prove ownership, maybe signing with the BLS keys, however this could end up being a ton of signatures.

cc @elsirion

87.md
benthecarman: > Why would the recommendation contain the full invite code? Feels like the federation ID would be sufficient since in the second step the user client would query the 38172/38173 event anyway

my main reasoning was so that you can recommend a federation even if they never announced themselves. Ideally every federation does but at least for now this is a manual process and to bootstrap this it'd be easier to support both options.

87.md
benthecarman: It seems like we'd need the index to build the invite code

https://github.com/fedimint/fedimint/blob/ca6176c76de0e92c68392e82b4d8ab1a7c7df87b/fedimint-core/src/api.rs#L827

benthecarman: I think this makes sense to me, my only concern would be if there would ever be changes to the encoding / extra data added to the invite code. Then we'd have to change this encoding here as well. Encoding the invite code gives us better guarantees of backwards compatibility

87.md
elsirion: Ah, the federation is supposed to announce itself? I thought the recommending user would do that if not the case already.

87.md
elsirion: There's little sense in signing imo. The federation id commits to the federation config. So even if the API URLs get spoofed the attacker can only:

* Return valid config with different API endpoints that will be used after that
* Return invalid config which will be rejected

So the only thing an attacker can achieve is making you ask more API endpoints for the correct config til one returns it.

elsirion: Just joining a federation one doesn't know is a horrible idea anyway, so I only expect this query to be used when the (short) federation id is known but no API URL. That could be useful to make e-cash notes smaller while still being able to claim notes from unknown federations.

87.md
```diff
@@ -0,0 +1,129 @@
...
+### Recommendation event
+```json
+{
+  "kind": 18173,
+  "pubkey": <recommender-user-pubkey>,
+  "tags": [
+    ["u", <recommended-cashu-mint-url>, "cashu"],
+    ["u", <recommended-fedimint-invite-code>, "fedimint"],
+    ["u", <recommended-cashu-mint-url>, "cashu"],
+    ["a", "38173:fedimint-pubkey:<d-identifier>", "wss://relay1", "fedimint"],
```
gudnuf, line 38: Why multiple u tags for one recommendation?
Can it be one u tag per recommendation and then include metadata like a review and rating, or anything else to help clients rank mints? Would also make handling the events simpler

benthecarman: It is a list, so you include all your recommendations in one event.

review and ratings don't really make sense for this use case, you either trust them with your money or you don't/

starbackr-dev: I think we need comments and reviews because there will be several mints that I like/use but I rate them based on stars and put my comment for what I like or what I use. NIP-89 (nostrapp.link) uses kind 1985 .  See the sample event below. We should use the same format for ratings and recommendations. 

{
  "created_at": 1710460808,
  "content": "The best text 2 speech I;ve used.. ",
  "tags": [
    [
      "l",
      "review/app",
      "social.coracle.ontology",
      "{\"quality\":1}"
    ],
    [
      "L",
      "social.coracle.ontology"
    ],
    [
      "p",
      "cf13a4f45628931f1cdbe7d31f923b6f6884d41644ab2cc055a0e0805b77cb9e"
    ],
    [
      "a",
      "31990:cf13a4f45628931f1cdbe7d31f923b6f6884d41644ab2cc055a0e0805b77cb9e:D94C72C241D1C73B2C36EEA73E9FB"
    ]
  ],
  "kind": 1985,
  "pubkey": "d8a2c33f2e2ff3a9d4ff2a5593f3d5a59e9167fa5ded063d0e49891776611e0c",
  "id": "95f5d58a7c9c2caed4f77d403c6e1846c71644997c87ed112f569413ec13bf9b"
}

gudnuf: > review and ratings don't really make sense for this use case, you either trust them with your money or you don't/

A mint could have lots of downtime, poor lighting routing, slow responses etc, but still be trusted to hold my money

pablof7z:
> Maybe we should make this into a big "discoverability" NIP and have a table like in NIP-51 with all the possible kinds of things that can be discovered? Starting with Cashu and Fedimint but then it could have other stuff, I have some in my mind already. This could actually be big.

Great idea. Lots of things feel like they fit in NIP-89, I had the same with DVMs (NIP-90); the same thing that happened with NIP-51 and we ended up overloading that one senseless.

A general discoverability NIP that doesn't try to shoehorn into "applications" is a great idea.

87.md
starbackr-dev: @benthecarman I propose the following changes to recommendation event kind 18173. 

Add content tag to provide verbal comments on the recommendation.
Add 'l' tag include review stars like this. Quality can range from 0 to 1. 

{"l","<type>", "{"quality":1}"}

pablof7z: Oh god, please don't 😔

gudnuf: I would propose something more like:

Content field is for verbal comments/reviews. 
Add a tag like "rating" that could be 0 to 100: ["rating", 42]... no need to index the rating and I like whole numbers. 

Cc @starbackr-dev @benthecarman

staab: Use the reviews NIP: https://github.com/nostr-protocol/nips/pull/879/files

starbackr-dev: Yes. Kind 1986 will exactly fit this. Let's use that for recommendation, comments or review and use 18173 for mint information.

benthecarman:
Updated to be more extendable and to do away with the list of recommendations and instead be a single event for each recommendation with a corresponding `k` and `d` tag pointing to the thing you are recommending. This should make it easier for clients to not accidentally overwrite people's recommendation list as well as make it simpler to find recommendations for things you are looking for.

87.md
gudnuf: How do I then signal that a mint is untrusted? There needs to be a way to explicitly say that certain services are bad. Either a new kind for *dissuasion* or the *recommendation* needs to be generalized to *advice* with positive or negative feedback.
I see that ratings may be overkill because as @benthecarman said, "you either trust them with your money, or you don't", but we should know which mints are **not** trusted with our money.

staab: Leave a bad review?

gudnuf: And you propose using nip85 separately to do that? Nip87 for positive recommendations and nip85 for negative reviews?

benthecarman: Added a line for the recommendation events to have the content be a review. This should suffice

87.md
```diff
@@ -0,0 +1,136 @@
...
+    ["u", <recommended-fedimint-invite-code>],
+    ["a", "38173:fedimint-pubkey:<d-identifier>", "wss://relay1"]
+  ],
+  "content": "I trust this mint with my life"
+}
+```
+
+The recommendation event is a parameterized-replacable event so that a user can change edit their recommendation without creating a new event.
+
+The `d` tag in `kind:38000` is the `kind:38173`/`kind:38172` event identifier this event is recommending, if no event exists, the `d` tag can still be calculated from the mint's pubkey/id.
```
gudnuf, line 46: What is expected if the mint does not share their pubkey? Lots of the Cashu mints I see do not have a pubkey at their /info endpoint

benthecarman: That's the deprecated spec, should be on /v1/info

gudnuf: Yea that too, I think it's mostly lnbits mints. 
ie. https://legend.lnbits.com/cashu/api/v1/Na859jHdYk9EjwSiBXurtp/v1/info does not return a pubkey.
I've just been falling back to random d tag

benthecarman: cc @callebtc maybe be a bug here?

benthecarman:
This is now live in mutiny and bitcoinmints.com

87.md
callebtc: > What is expected if the mint does not share their pubkey? Lots of the Cashu mints I see do not have a pubkey at their /info endpoint

LNbits mints do not have their own pubkey since we don't use it in the protocol yet (this was reserved for future use cases). We currently use the mint URL as an identifier of the mint. Maybe that's a way?

TonyGiorgio:
I think there needs to be a way to remove recommendations as well as remove listing a mint in general. Does this NIP support that?

Giszmo:
I'm surprised to not see @franzaps and his https://github.com/franzaps/nips/blob/applications/82.md mentioned here ...

benthecarman:
wow thank you @pablof7z

pablof7z:
> wow thank you @pablof7z

was fucking time! I came to look up the schema and realized it wasn't merged yet 😂

Issue #1106, opened 2024-03-06
NIP-46 secret


alexgleason:
We can skip the whole initial handshake of "connect" -> "ack" in NIP-46 by including a private key in the `bunker://` URL _instead_ of a `secret`.

The client has to sign events using that private key. It skips an entire layer of complexity between the client and the signer.

fiatjaf:
This is a good idea.

fiatjaf:
The main benefit is not that you can skip `connect`. I think we shouldn't skip that. The main benefit is that you get a single string credential that you can store and reuse.

fiatjaf:
A client that uses a bunker to connect would presumably store that URL locally somehow, say, in `localStorage`. And today they would have to generate a secret key and then store it separately. with this they would just either generate and store the two things together, or they would just use the secret key from user input, not generate.

Am I thinking wrongly about this?

alexgleason:
I agree @fiatjaf 

In the current flow you generate a random keypair and then use the "secret" provided by the bunker to authenticate with it. Now the bunker can trust your keypair.

But if the bunker just gives you the keypair to begin with, it already trusts it.

I don't get what's not to get.

brugeman:
Whether one stores one or two strings to me is a triviality not worth the ink. I get what you're proposing Alex, I don't get the benefit. 

I do get the reduced security though - today if bunker url with a single-use token leaks (which it will, bcs bunker urls are there to be copy-pasted), at least hopefully token has already been redeemed or expired, so it's not a big deal. If bunker url with private key leaks then much more damage can happen - that's the reason Pablo came up w/ a redeemable token afaik.

Oh btw the old spec had nothing about this 'secret' token, and the new one just mentions it without explaining why it's there. It's a single-use token, redeemed on the first connect - the connection perms are assigned to the npub that connected and token is no longer valid. Pablo came up with this token and that's how nsecbunker handles it.

Everything you can do with passing a private key you can also do with a token, but token is also more secure. Or is there a use case I'm not considering? Or does the current spec imply a different meaning for the secret token which I'm not aware of?

alexgleason:
I private key IS a token. You can revoke it just like a token.

brugeman:
Ok I revoked the token - I no longer accept 'connect' call with this private key as a token. But what you're proposing is that client keeps using that private key and signer keep allowing this private key to sign. Which means if this bunker url with that token leaked, anyone else can use it and signer will allow them access to sign too. No?

[...]

PR #1111, opened 2024-03-07
Add optional `&clientkey=...` to NIP-46

fiatjaf:
This allows a single bunker://... string to be self-contained, storeable and pre-authorized.

closes https://github.com/nostr-protocol/nips/issues/1106

### DIFF
--- a/46.md
+++ b/46.md
@@ -25,7 +25,7 @@ This is most common in a situation where you have your own nsecbunker or other t
 The remote signer would provide a connection token in the form:
 
 ```
-bunker://<remote-pubkey>?relay=<wss://relay-to-connect-on>&relay=<wss://another-relay-to-connect-on>&secret=<optional-secret-value>
+bunker://<remote-pubkey>?relay=<wss://relay-to-connect-on>&relay=<wss://another-relay-to-connect-on>&secret=<optional-secret-value>&clientkey=<optional-hex-key-to-use-as-the-clients-key>
 ```
 
 This token is pasted into the client by the user and the client then uses the details to connect to the remote signer via the specified relay(s).


46.md
```diff
@@ -25,7 +25,7 @@ This is most common in a situation where you have your own nsecbunker or other t
 The remote signer would provide a connection token in the form:
 
 ```
-bunker://<remote-pubkey>?relay=<wss://relay-to-connect-on>&relay=<wss://another-relay-to-connect-on>&secret=<optional-secret-value>
+bunker://<remote-pubkey>?relay=<wss://relay-to-connect-on>&relay=<wss://another-relay-to-connect-on>&secret=<optional-secret-value>&clientkey=<optional-hex-key-to-use-as-the-clients-key>
```
alexgleason, line 28: I would go with `sk` or `secretKey` personally. I think it's more important that it's secret than which side of the flow it's on.

```suggestion
bunker://<remote-pubkey>?relay=<wss://relay-to-connect-on>&relay=<wss://another-relay-to-connect-on>&secret=<optional-secret-value>&sk=<optional-hex-key-to-use-as-the-clients-key>
```

alexgleason (approved):

46.md
fiatjaf: I just thought that `clientkey` reduces the chances of someone actually writing the main user secret key in there.

alexgleason: Fair enough.

mikedilger:
I'm ok with this but as implied by @brugeman, specifying the client key directly leaves an infinite window for a leaked key (via a leaked bunker URL) to be used by an attacker, whereas specifying the secret and having it consumed closes that window during the connect. Maybe that isn't concerning to some, which is why I'm ok with this, but I prefer the connect sequence even with the extra round-trip.

fiatjaf:
Yeah, maybe people can use persistent secrets if they want this functionality.

PR #811, opened 2023-10-10
NIP-106 Decentralized Web Hosting on Nostr

studiokaiji:
# NIP-106 Decentralized Web Hosting on Nostr

By recording HTML, CSS, and JS on the Nostr relay, it becomes possible to create a decentralized web hosting solution that eliminates the need for centralized servers. Web servers or Nostr clients retrieve these recorded data, transform them into appropriate forms, and deliver them.

## Reasons for Hosting on Nostr

- Tamper-resistant through public key-based integrity
- Fault tolerance through deployment on multiple relays
- Resistance to blocking due to the distribution of web servers or clients
- Faster retrieval speed compared to IPFS's DHT

## Proposed Approach

Each HTML, CSS, and JS file is assigned a `kind` for identification.

- HTML: `kind: 5392`
- CSS: `kind: 5393`
- JS: `kind: 5394`

The "content" field contains the content of the file. However, internal links (`href`, `src`, etc.) referenced within should be replaced with event IDs.

Example: `<link rel="stylesheet" href="066b7ca0b167f0adad5c6d619ab1177050423e3979e83b8dfa069992533bdcf5">`

### Implementation on Web Server or Client

Access events using `/e/{nevent}`. Since event `nevent` are specified for each internal link, opening an HTML file enables automatic retrieval of data from this endpoint.

By using `nevent`, you can expedite the retrieval of relay information by including it internally, which offers the advantage of speed. However, it is also acceptable to implement the retrieval in the usual hex format as an option.

**Implementation Example (Golang)**

```go
r.GET("/e/:nevent", func(ctx *gin.Context) {
  nevent := ctx.Param("nevent")

  // parse nevent
  _, res, err := nip19.Decode(hexOrNevent)
    if err != nil {
      ctx.String(http.StatusBadRequest, "Invalid nevent")
      return
    }

  data, ok := res.(nostr.EventPointer)
  if !ok {
    ctx.String(http.StatusBadRequest, "Failed to decode nevent")
    return
  }

  id := data.ID
  allRelays = append(allRelays, data.Relays...)

  // Fetch data from nostr pool
  ev := pool.QuerySingle(ctx, allRelays, nostr.Filter{
    Kinds: []int{
      consts.KindWebhostHTML,  // 5392
      consts.KindWebhostCSS,  // 5393
      consts.KindWebhostJS // 5394
    },
    IDs:   []string{id},
  })

  if ev != nil {
    // Return data with content-type adapted to kind
    switch ev.Kind {
[...]

PR #1112, opened 2024-03-07
NIP-83: JavaScript Registry

alexgleason:
JavaScript is unique in that you can import source code directly by URL. This NIP makes it possible to publish JavaScript code to Nostr and then use it within web browsers and Deno using an HTTP gateway.

Inspiration: #742 #811 https://github.com/denoland/deno/discussions/22779

[Read here](https://github.com/nostr-protocol/nips/blob/1158921a12a36c1db3d6f8cf92fa78415f54d345/83.md)

### DIFF
--- a/83.md
+++ b/83.md
@@ -1,0 +1,120 @@
+NIP-83
+======
+
+JavaScript Registry
+-------------------
+
+`draft` `optional`
+
+JavaScript source files may be stored by relays, and then imported into web browsers or development environments.
+
+This NIP defines two kinds:
+
+- `8394` - JavaScript source file.
+- `8395` - TypeScript source file.
+
+In both cases, the `content` field contains the source code of the file.
+
+```json
+{
+  "kind": 8394,
+  "id": "c17cf2f43580ad8238703ea32fb55c90c93402ca1f7d38085381f32c0f00e459",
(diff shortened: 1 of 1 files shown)


83.md
```diff
@@ -0,0 +1,120 @@
...
+  "sig": "b01ab6e2273bc33594dde5634c68add95c6273b8a864bf8f4c8a85564db3e3df2c9981a3c5d8694cc4fa0e0f984635d6b51d5ca352eff96b7cb789c39429d303"
+}
+```
+
+## Immutability
+
+Source code events are considered immutable, and should NOT be deleted by supported clients or relays in response to kind `5` deletion requests.
```
AsaiToshiya, line 44: I agree about immutability, but not being able to delete it seems a bit overkill 😅

vitorpamplona:
I don't think relays should have the responsibility of serving content via HTTP. It's just not what they are designed for.

AsaiToshiya:
In the future, I think it would be useful to include tags such as `name` and `version` which correspond to fields in `jsr.json`, and also by doing so can be distinguished from `kind: 5394` in NIP-106.

alexgleason:
This was not a serious proposal, more of a thought experiment. Never did anything with it.

PR #1113, opened 2024-03-08
NIP-50: add event classification extensions

nostr-wine:
As previously discussed in the other NIP-50 PR with @alexgleason and @fiatjaf...

We have recently started experimenting with event classification (https://docs.nostr.wine/api/classification) on our nostr.wine relay. 

I've added language, sentiment, and nsfw to extensions. I also updated the wording of the domain: extension to include ONLY events from users instead of just "include events". Let me know if I'm misunderstanding that.

I'm not sure of the best format for everything (is ISO 639-1 a good choice for language codes?) but wanted to get something started.

### DIFF
--- a/50.md
+++ b/50.md
@@ -48,7 +48,6 @@ Relays SHOULD exclude spam from search results by default if they support some f
 Relay MAY support these extensions:
 - `include:spam` - turn off spam filtering, if it was enabled by default
 - `domain:<domain>` - include only events from users whose valid nip05 domain matches the domain
-- `topic:<topic>` - include only events within a classified topic (short string, one to two words)
 - `language:<two letter ISO 639-1 language code>` - include only events of a specified language
 - `sentiment:<negative/neutral/positive>` - include only events of a specific sentiment
 - `nsfw:<true/false>` - include or exclude nsfw events (default: true)
--- a/50.md
+++ b/50.md
@@ -47,4 +47,8 @@ Relays SHOULD exclude spam from search results by default if they support some f
 
 Relay MAY support these extensions:
 - `include:spam` - turn off spam filtering, if it was enabled by default
-- `domain:<domain>` - include events from users whose valid nip05 domain matches the domain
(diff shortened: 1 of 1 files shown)


50.md
```diff
@@ -47,4 +47,8 @@ Relays SHOULD exclude spam from search results by default if they support some f
 
 Relay MAY support these extensions:
 - `include:spam` - turn off spam filtering, if it was enabled by default
-- `domain:<domain>` - include events from users whose valid nip05 domain matches the domain
+- `domain:<domain>` - include only events from users whose valid nip05 domain matches the domain
+- `topic:<topic>` - include only events within a classified topic (short string, one to two words)
```
alexgleason, line 51: How it topic different from hashtag, which you can get with `#t` filters? It seems too niche - I would remove it.

The other ones look good to me.

alexgleason: Unless somehow this is supposed to find things that in theory could be tagged with a hashtag but aren't? :thinking: How would that work?

nostr-wine: We are doing our own topic classification after the event has been posted based on the content of the event and the links. It has nothing to do with hashtags created by the user.

nostr-wine: See https://api.nostr.wine/classification?topic=food as an example.

nostr-wine: Good question and I don't have an obvious solution. I was thinking we'd leave it open ended because different providers will likely have different classified topics.

For nostr.wine specifically, we could expose an API with the most popular classified topics.

alexgleason: Cool. I would personally omit this one from the PR until that's sorted out. I think it's too ambiguous for other implementations, whereas the other extensions you proposed are unambiguous and self-explanatory.

alexgleason (approved):

50.md
nostr-wine: Ok - done. 

I do think we need some way for clients to query for topics but I really don't want to be the one to argue about or decide where it should go.

PR #1116, opened 2024-03-09
NIP-26: Increase percentage of relays supporting it from 0.01% to 100%

arthurfranca:
Currently almost no relay supports NIP-26 (e.g.: nostream removed it and strfry doesn't have it). By renaming `delegation` tag to `D`, all relays automatically have basic support to delegated events.

Minds client (the first and only client implementation?) won't break because they certainly use their own relay (which doesn't need to change) for that.

### DIFF
--- a/26.md
+++ b/26.md
@@ -8,11 +8,11 @@ Delegated Event Signing
 
 This NIP defines how events can be delegated so that they can be signed by other keypairs.
 
-Another application of this proposal is to abstract away the use of the 'root' keypairs when interacting with clients. For example, a user could generate new keypairs for each client they wish to use and authorize those keypairs to generate events on behalf of their root pubkey, where the root keypair is stored in cold storage. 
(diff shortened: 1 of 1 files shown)


vitorpamplona:
Mind's relay seems offline.

This PR is why I wanted to deprecate NIP-26. We need to kill it to bring it back in a way that it's actually useful.

arthurfranca:
> We need to kill it to bring it back in a way that it's actually useful.

This one with #1114 could fix most problems. Do you have something better in your mind?

#1067 would be a good optional addition aiming an ideal future where most outbox relays would help with some quirks.

vitorpamplona:
> Do you have something better in your mind?

Not really. For me, this approach of delegation is just plain wrong. But I am willing to be convinced by some good implementations out there.

fiatjaf:
Two years from now will be as good as a time to think about these matters as now, and maybe until that time someone comes up with a better idea.

arthurfranca:
Closing this in favor of an updated version of #1114

PR #1117, opened 2024-03-10
NIP-01 Reword filter limits

### LINKED ISSUE #1115: NIP-01: `limit` in multiple filters
How should relays interpret the `limit` specification if present in multiple filters? There seem to be contradicting specs regarding the `limit` filter field if given by multiple filters

```json
[
  {"limit": 2, …},
  {"limit": 40, …}
]
```

1. NIP-01 specifies that we can use multiple filters in a subscription

  > A REQ message may contain multiple filters. In this case, events that match any of the filters are to be returned, i.e., multiple 
  filters are to be interpreted as || conditions.

2. NIP-01 also specifies that a `limit` determines the `<maximum number of events relays SHOULD return in the initial query>`

Is the limit to be applied for each filter? So we return 40-42 events depending on whether the matching events overlap.
Or should the maximum `limit` of any given limit be relevant, since the idea is to clamp the event count of the initial query?

schmijos:
Fixes #1115 by uniform glossary and using "each"

### DIFF
--- a/01.md
+++ b/01.md
@@ -112,30 +112,30 @@ Relays expose a websocket endpoint to which clients can connect. Clients SHOULD 
 Clients can send 3 types of messages, which must be JSON arrays, according to the following patterns:
 
   * `["EVENT", <event JSON as defined above>]`, used to publish events.
-  * `["REQ", <subscription_id>, <filters1>, <filters2>, ...]`, used to request events and subscribe to new updates.
+  * `["REQ", <subscription_id>, <filter1>, <filter2>, ...]`, used to request events and subscribe to new updates.
   * `["CLOSE", <subscription_id>]`, used to stop previous subscriptions.
 
 `<subscription_id>` is an arbitrary, non-empty string of max length 64 chars. It represents a subscription per connection. Relays MUST manage `<subscription_id>`s independently for each WebSocket connection. `<subscription_id>`s are not guaranteed to be globally unique.
 
-`<filtersX>` is a JSON object that determines what events will be sent in that subscription, it can have the following attributes:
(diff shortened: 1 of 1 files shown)


01.md
```diff
@@ -127,13 +127,13 @@ Clients can send 3 types of messages, which must be JSON arrays, according to th
...
 
-Filter attributes containing lists (`ids`, `authors`, `kinds` and tag filters like `#e`) are JSON arrays with one or more values. At least one of the arrays' values must match the relevant field in an event for the condition to be considered a match. For scalar event attributes such as `authors` and `kind`, the attribute from the event must be contained in the filter list. In the case of tag attributes such as `#e`, for which an event may have multiple values, the event and filter condition values must have at least one item in common.
+Filter attributes containing lists (`ids`, `authors`, `kinds` and tag conditions like `#e`) are JSON arrays with one or more values. At least one of the arrays' values must match the relevant field in an event for the condition to be considered a match. Scalar event attributes such as `pubkey` and `kind` must match the corresponding filter condition exactly. In the case of tag attributes such as `#e`, for which an event may have multiple values, the event and filter condition values must have at least one item in common.
```
schmijos, line 136: `authors` was wrong here. Replaced with `pubkey`.

fiatjaf: This sounds more confusing than before.

If there is a filter `{"kinds": [1,2,3]}` how does an event `{"kind": 1}` match it exactly?

schmijos: I see your point. I deleted the sentence and included the information in the filter format example.

schmijos:
Closing for lack of interest.

bezysoftware:
Oh no, this actually makes a lot of sense, I was also struggling with the filters wording especially around the `limit` field.

PR #1126, opened 2024-03-18
Added bot field to denote automated npubs

### LINKED ISSUE #624: Extending kind:0 - Mark bots.

SilberWitch:
This is an extension of the Kind 0 fields, to allow profiles to self-indicate if they are a bot.

## Description

There are an increasing number of automated npubs that are simply newsfeeds or cross-posting from other places, such as RSS or Twitter. This is completely legitimate and many users find them helpful, to consolidate their various reading lists, but it creates confusion when there are two npubs with the same name (one human, one automated), or when people try to respond to these bots, as if they were humans.

There are also an increasing number of chatbots, infobots, DVMs, etc. that could make their status more clear to users and clients.

Until now, profiles denote this status in freetext descriptions, but that makes it difficult to query or filter, and limits the ability to easily create clients with newsfeeds or bot-labels.

## Notes for Reviewer
Closes #624

### DIFF
--- a/24.md
+++ b/24.md
@@ -16,6 +16,7 @@ These are extra fields not specified in NIP-01 that may be present in the string
   - `display_name`: an alternative, bigger name with richer characters than `name`. `name` should always be set regardless of the presence of `display_name` in the metadata.
   - `website`: a web URL related in any way to the event author.
   - `banner`: an URL to a wide (~1024x768) picture to be optionally displayed in the background of a profile screen.
+  - `bot`: a boolean to clarify that the content is entirely or partially the result of automation, such as with chatbots or newsfeeds.
 
 ### Deprecated fields
 


finrodfelagund97 (approved):

mleku:
may i suggest an alternative which adds a tag field same kind of thing as the event tags (i personally think that was a mistake combining - could now be replaced with NIP-42 labels)

yes, the kind 0 needs more fields but i don't think just one more is really gonna help long term... and a tags field is opaque to clients that don't understand tags they just ignore them, but new fields may break things

mattkwarren:
I won't comment on the technical aspect of how you want to implement this idea, because I don't know enough about the Nostr protocol to do so (yet). However, the idea itself is wonderful. An actual field would allow clients to do useful things with the data, like label posts as bot posts, warn in comment sections, etc. Doing those things using free text in a profile section is virtually impossible, or at least not as reliable, as you point out. I would love to see it if it won't harm the protocol in some other way I can't see.

fiatjaf (approved):

alexgleason (approved):

staab (approved):

mleku:
> > yes, the kind 0 needs more fields but i don't think just one more is really gonna help long term... and a tags field is opaque to clients that don't understand tags they just ignore them, but new fields may break things
> 
> Wouldn't tags be more appropriate for an event with other (often human-readable) content? Here, the metadata is the content.

i'm talking about futureproofing it for easier additions in the future

with tags in place the modifications to the NIP are simpler and require no change in the fundamental parsing logic for the JSON, tags with unrecognised first fields will just be ignored by most clients, but the tags themselves won't be unexpected

in comparison, adding an extra field in here that isn't expected could cause a bad implementation to fail to parse much more easily, such as for example trying to marshal it into a struct without the extra field, the JSON marshaling implementation may by default throw an error if it finds a field it doesn't recognise from the expected target variable

a tag list, on the other hand, won't be mangled or cause an error, just that there will be no logic yet in the downstream code that will respond to the additional, unrecognised tag

it's a bit of an edge case but one that can be avoided forevermore if we add a tags list to the event instead of piecemeal addition of fields

alexgleason:
Whether kind 0 should use tags instead of json-stringified objects has been discussed many times. Obviously tags are better because they can be indexed. But it's not worth doing it two different ways (and creating inconsistency) just for this field. If we want to switch kind 0 to be tags, that's a separate discussion. And if a service really needs to index bots, I would suggest using a NIP-50 search extension like `{ search: "bots:true" }`

IngwiePhoenix:
> Also, this is not a good argument against it. Kind 0 content is essentially free-form JSON. If implemenations don't account for there being extra fields, the implementation is the one at fault.

imo if one wants to strictly marshal JSON into a struct, it should be for the "common" object (id, sig, pubkey, ...) but not for the .content of kind 0. This is actually a huge problem when working with Matrix, and why there are only so few implementors and what is Nostr's biggest strength as a protocol.

Is there currently a discussion thread on tags in kind 0? I would be curious to read about this more. Thing is that I do see an advantage for searchability here by querying relays directly (`["REQ","nak",{"kinds":[0],"#t":["artist","musician"]}]`). So if there is one, please do link it - thanks! ^^

SilberWitch:
> Is there currently a discussion thread on tags in kind 0? I would be curious to read about this more.

If not, perhaps you should start a new PR. I would also be interested in continuing the discussion elsewhere.

mleku:
>  the implementation is the one at fault.

that is just dodging the problem that a tag list is inherently constituted of things that may or may not be understood by the client

maybe in the languages you are used to working with, these extra tags are no problem but in Go if i make a struct with tags for these expected fields and someone adds a new one it breaks my marshaling code

and conversely, on a free form tag based list of lists structure, if the first field is not recognised, it will just be ignored, it is already defined in the structure as a list of lists of strings so there is nothing syntactically different

but key/value fields are syntactical in structure, i'm sure that other languages json handling also has this, ones that are more strict with typing

PR #1128, opened 2024-03-18
wiki relay list

pablof7z:
Reserving `kind:10008` for wiki relay sets


fiatjaf:
This is fine, but why not update it on https://github.com/nostr-protocol/nips/pull/787 (you can push directly to that branch).

I had reserved 10101 for authors and 10102 for relays there, but we can change the numbers.

pablof7z:
damn, I hadn't seen that; ok, changing my kind and closing this

PR #1130, opened 2024-03-19
Add open-ended hint mechanism

staab:
This introduces the `~` tag, which allows events to refer to indexes where clients might be able to find additional related events.

This is purposely open-ended, since many types of indexes can be created, but the main goal is to replace or complement existing relay hints with redundant indexes that are less prone to link rot. In other words, if damus' relay goes down, many relay hints will become dead and can't be updated. Index hints allow for pointing to other sources than relays for fetching events, which may have different trade offs (for example, gundb indexes are publicly writable, and so prone to attack, but also maintainable by anyone).

The main events recommended are kind 0/3/10002, so that clients can find user relay preferences and profiles and then load events directly from relays, but other use cases are possible. For example, it is currently hard to find reply parents if the relay hint is unhelpful. Ad-hoc indexes can be created by clients which contain just the ancestors of a given note.

For example, for an event with the following tags:

```json
"tags": [
  ["e", "<event_id>", "wss://relay.example.com", "root"],
  ["~", "<event_id>", "gundb"]
]
```

If `event_id` is not found on `relay.example.com`, it may be found at `gun.get("<event_id>")`, which would return an index with a single event (the parent event).

Other index types will have different trade-offs from relays, which is a good thing. For example, gundb keys are publicly writable, so they can be both attacked and maintained by anyone. Other index types that are authenticated-write could be created though.

### DIFF
--- a/xx.md
+++ b/xx.md
@@ -16,26 +16,24 @@ This NIP introduces the `~` tag, used to recommend durable indexes which can be 
 
 Indexes may directly provide events published by a given pubkey, or references to other indexes. An index's content is a JSON-encoded object with the following properties optionally defined:
 
-- `events`: an object mapping pubkeys to arrays of events signed by that pubkey. Indexes MAY include as many events as desired, but SHOULD prioritize kind 0 and kind 3.
+- `events`: an array of events. Indexes MAY include as many events as desired, but SHOULD prioritize kind 0, 3, and 10002 so that profiles can be loaded quickly while clients fetch additional events from relays directly.
 - `indexes`: an object referring to additional indexes, keys of which are index types, and values are arrays of index addresses.
 
 Example:
 
 ```
 {
-  "events": {
-    "b0635d6a9851d3aed0cd6c495b282167acf761729078d975fc341b22650b07b9": [
-      {
-        "kind": 3,
-        "content": "Hello",
-        "tags": [],
-        "created_at": 1710607821,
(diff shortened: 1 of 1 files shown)


fiatjaf:
If you write `"gundb"` then we need a standard for gundb. If someone writes something else we need a standard for that something else. And these standards should be in this NIP.

I don't know why you wrote "gundb". Is that a thing that works?

alexgleason:
Isn't `-` already used for private events?

staab:
> Isn't - already used for private events?

Tilde, not dash.

> If you write "gundb" then we need a standard for gundb. If someone writes something else we need a standard for that something else. And these standards should be in this NIP.
> 
> I don't know why you wrote "gundb". Is that a thing that works?

From the NIP:

> ## `gundb`
> 
> [GunDB](https://gun.eco/) is an implementation of a DHT, and can be used to provide an index. A gundb index address may be any string.

GunDB is what Iris used to be built on. It's a web-first DHT, and "works". It's just a low-hanging fruit first pass at this. Other things I could see working:

- A huge file accessible over torrent with lots of events in it
- A json file accessible by URL
- An IPFS file

alexgleason:
Also isn't this similar to [NIP-48](https://github.com/nostr-protocol/nips/blob/master/48.md)?

staab:
> Also isn't this similar to [NIP-48](https://github.com/nostr-protocol/nips/blob/master/48.md)?

Very interesting, it's basically the same. Maybe we can just use that. I'll think about it.

PR #1133, opened 2024-03-22
Add optional_requested_permissions

nostrband:
This is implemented in nsec.app, nostr.band, Coracle and Nostrudel, so maybe it's time to update the NIP.


vitorpamplona:
Ack. Default permissions are key to good user experiences.

PR #1135, opened 2024-03-23
nip-64: inbox model

fiatjaf:
read here: https://github.com/nostr-protocol/nips/blob/inbox-model/64.md

from the text:

> The approach described in this NIP doesn't replace any of the other models for following people that are already in use in Nostr, and doesn't claim to be strictly superior to the Outbox Model, but just provides an alternative way to do things.

> In reality, clients should implement both approaches and can never assume others will implement this.

based on https://github.com/nostr-protocol/nips/discussions/1134

### DIFF
--- a/64.md
+++ b/64.md
@@ -1,0 +1,65 @@
+NIP-64
+======
+
+Inbox model
+-----------
+
+`draft` `optional`
+
+Suppose **Sarah** wants to subscribe to the notes **Walter** writes.
+
+She must let Walter know she is following him. The first step is to find Walter's _inbox list_:
+
+```jsonc
+{
+  "kind": 10064,
+  "pubkey": "<walter>",
+  "tags": [
+    ["relay", "wss://walter.inbox"]
+  ]
+}
+```
+
+Now Sarah can publish a `kind:6401` _follow intent_ to `wss://walter.inbox` tagging Walter:
+
+```jsonc
+{
+  "kind": 6401,
+  "pubkey": "<sarah>"
+  "tags": [
+    ["p", "<walter>"],
+    ["relay", "wss://sarah.inbox"],
+    ["relay", "wss://other.inbox"]
+  ]
+}
+```
+
+Sarah also publishes these to her own inbox relays (`wss://sarah.inbox` and `wss://other.inbox`) so they are aware.
+
+From the Walter side, whenever he is about to publish a note, he downloads all the `kind:6401` events from his inbox relays and sends his note to all the relays he finds in these.
+
+Whenever Sarah's inbox relay receive a note from Walter, they will know that Sarah is following him, so they index that note in a way that it is associated with Sarah. If the relays receive a note from someone else they can simply reject the note.
+
+When Sarah turns on her client she will connect to her inbox relay, perform [NIP-42](42.md) `AUTH` and create a subscription that doesn't specify `"authors"` (for example, `["REQ", "_", {}]`). The relays, aware of who she is and who she is following, can deliver to her just Walter's notes.
+
+If Sarah ever decides to stop following Walter, she can send a `kind:6402` _unfollow intent_ to both Walter's relays and her own, so her relay can start rejecting (and maybe delete previous) Walter's notes, and Walter can know to not send her his notes anymore. As an optimization, Walter's inbox relay can just delete Sarah's `kind:6401` upon receiving a `kind:6402`.
+
+```jsonc
+{
+  "kind": 6402,
+  "pubkey": "<sarah>"
+  "tags": [
+    ["p", "<walter>"]
+  ]
+}
+```
+
+---
+
+The approach described in this NIP doesn't replace any of the other models for following people that are already in use in Nostr, and doesn't claim to be strictly superior to the Outbox Model, but just provides an alternative way to do things.
+
+In reality, clients should implement both approaches and can never assume others will implement this.
+
+For example, if Sarah wanted to follow Wesley and he didn't have a `kind:10064` she would have to do the normal "outbox" approach of subscribing to his relays and fetching his notes from there.
+
+In the same vein, if for any reason Sarah noticed that she had stopped receiving notes from Walter for a long time (~7 days) it would make sense for her to go to his relays (obtained from [NIP-65](65.md), [NIP-05](05.md), relay hints or other means) and check if he is publishing to these. In the affirmative case, she would turn Walter into an "outbox follow" instead of an "inbox follow".


staab:
I don't really see what this accomplishes. All it does is make an empty filter meaningful for auth'd users. This seems like a job for a content recommendations DVM instead, or a tool that can pull Walter's notes into Sarah's relay. But if she wants to retrieve Walter's content she should just ask for it.

64.md
```diff
@@ -0,0 +1,65 @@
...
+
+`draft` `optional`
+
+Suppose **Sarah** wants to subscribe to the notes **Walter** writes.
+
+She must let Walter know she is following him. The first step is to find Walter's _inbox list_:
+
+```jsonc
+{
+  "kind": 10064,
```
alexgleason, line 15: Inboxes and outboxes are both just relays in reality. Put another way, an outbox is a relay with a "write" marker, and an inbox is a relay with a "read" marker. So this nip can reuse kind 10002 events from NIP-65.

64.md
```diff
@@ -0,0 +1,65 @@
...
+  "pubkey": "<walter>",
+  "tags": [
+    ["relay", "wss://walter.inbox"]
+  ]
+}
+```
+
+Now Sarah can publish a `kind:6401` _follow intent_ to `wss://walter.inbox` tagging Walter:
```
alexgleason, line 23: Why not use kind 10 and 11 events? https://github.com/nostr-protocol/nips/pull/349

64.md
fiatjaf: I think we gain nothing by conflating two different kinds of relays. Relays specified on kind 10002 are already there and already being used. Relays that implement this NIP would require a little bit of special behavior, so it's better to signal them properly.

Also I think it's very important that users announce that they support this NIP before others start engaging with them through it, having a special kind published serves that purpose.

64.md
fiatjaf: These kinds are already being used by NIP-29.

vitorpamplona:
The idea is good and it serves as an alternative to the pull model. We can have both running at the same time. 

To @staab's point, we can also use the subscription events to help a service provider or a DVM to redistribute events where they are supposed to be. 

But I do think the unfollow event should just be a deletion request.

mikedilger:
The inbox model has different characteristics from the outbox model. Neither replaces the other and both can co-exist.

... but the outbox model is superior of course.

I'm not sure we need a separate list of inbox relays for the inbox model. It adds complexity, and I don't see the benefit.

64.md
```diff
@@ -0,0 +1,55 @@
...
+}
+```
+
+Sarah also publishes these to her own inbox relays (`wss://sarah.inbox` and `wss://other.inbox`) so they are aware.
+
+From the Walter side, whenever he is about to publish a note, he downloads all the `kind:6401` events from his inbox relays and sends his note to all the relays he finds in these.
+
+Whenever Sarah's inbox relay receive a note from Walter, they will know that Sarah is following him, so they index that note in a way that it is associated with Sarah. If the relays receive a note from someone else they can simply reject the note.
+
+When Sarah turns on her client she will connect to her inbox relay, perform [NIP-42](42.md) `AUTH` and create a subscription that doesn't specify `"authors"` (for example, `["REQ", "_", {}]`). The relays, aware of who she is and who she is following, can deliver to her just Walter's notes.
```
arthurfranca, line 43: This breaks how absent `authors` filter works.

Instead of NIP-42, better to filter like this `{ "nip64": "<sarah-pubkey>", kinds: [1], ...etc }`. The `nip64` key clearly asks for relay to change how an absent `authors` filter should be treated.

It is ok to let anyone see the "feed" as Sarah cause follow list is public.

64.md
arthurfranca: > it's very important that users announce that they support this NIP

After #991 gets merged, you can pick a new relay usage flag to mean "nip64 relay" instead of introducing `kind:10064`.

64.md
fiatjaf: This introduces unnecessary complexity. A new term to remember, libraries would have to add support for this and everything would be awful.

I don't see any issue with using an empty `"authors"` and expecting the relay to treat that in a certain way, as long as the relay is carefully chosen to be treated differently by the client. A relay is already allowed to return anything they want _as long as it matches the filter_, it isn't mandated or expected to _always return everything_.

There are already other experiments out there of relay that behave differently when `AUTH`ed and when not and I hope to see more.

Also specifying the `AUTH` stuff from the beginning is nice as it allows for private feeds in cool ways that, although not explicitly specified in this NIP, could be implemented somewhat easily.

64.md
fiatjaf: We could use a different tag in kind 10002 if people prefer that, but I honestly thinks it's always better to use a different event kind when the use case and implementation contexts are different. It's not a super strong argument, more like a rule-of-thumb I came up with, same as I said in https://github.com/nostr-protocol/nips/pull/686#discussion_r1536542654

fiatjaf:
I guess we could remove the special handling of events by the inbox relays and the blank filter behind `AUTH` and this NIP could still work somewhat (one would have to pick a permissive relay for their inbox, then query it manually with an `"authors"` filter later), but it would lose much of its appeal to me and probably not worth doing.

64.md
fiatjaf: Immediately after writing the comment above I read this note from @mikedilger: https://njump.me/nevent1qy88wumn8ghj7mn0wvhxcmmv9uqjqamnwvaz7tmwdaehgu3wv45kuatwv3a8wctw0f5kwtnnwpskxef0qyfhwumn8ghj7mmxve3ksctfdch8qatz9uqzpv48826guadrdezm5eyswj6m2vjgq9pewzz309v427jcv33yrgqdlvlk6v

mikedilger: I must have missed this entire paragraph. You are suggesting to put the smarts in the relay.  I thought Walter would just tag Sarah when he puts her notes into her inbox relay.  If the relay really handles the subscriptions as a service without the tagging, then of course this is not synonymous with a NIP-65 inbox relay and it should be specified differently.

fiatjaf: I don't think there should be any tagging. Walter writes one note, signs it once and sends it to all the relays of people who have previously announced their intent to follow him (and also sends it to his own outbox relays). If he had to write a different note for each I think this would be a cumbersome nightmare.

alexgleason:
I'm still thinking about this NIP. My original suggestion in #1134 was just that clients should try harder (a "free-for-all"). But I like that you envisioned something more.

I think this will be a slow burn. It's the start of something. Next someone needs to make an all-in-one "inbox/outbox" package in some programming language (TypeScript) that does all the hard parts of this. What is maybe missing is the yin-yang of inbox/outbox used together.

vitorpamplona:
> I guess we could remove the special handling of events ... probably not worth doing

I would never use that special filter anyway. All I need to do is add the user's own inbox relay to the relay list of the usual `authors: ` filter. If the client knows about the subscriptions and is sure the writer is using a compliant client, it can even switch to the user's relays. Rergardless, the filter stays the same.

64.md
dcadenas: This could be useful https://undocumented.nostrkinds.info/

alexgleason: 6401 is within the reserved Nostr Job Result range. :facepalm:

PR #1136, opened 2024-03-23
potentially superfluous/erroneous colon ":" character in NIP-01 non-parameterized replaceable-event "a" tag specfication

avlo:
**_non-parameterized_** replaceable-event "a" tag specification contains (potentially) superfluous/erroneous colon ":" character after hex pubkey and prior to field separator ","

```java
["a", <kind integer>:<32-bytes lowercase hex of a pubkey>:, <recommended relay URL, optional>]
                                                         ↑
                                          superfluous/erroneous inclusion
```

in contrast, **_parameterized_** replaceable event, _correctly_ specifies colon ":" character, as it is necessarily followed by a `<d tag value>`

```java
["a", <kind integer>:<32-bytes lowercase hex of a pubkey>:<d tag value>, <recommended relay URL, optional>]
                                                         ↑
                                                necessary inclusion
```

### DIFF
--- a/01.md
+++ b/01.md
@@ -79,7 +79,7 @@ This NIP defines 3 standard tags that can be used across all event kinds with th
 - The `p` tag, used to refer to another user: `["p", <32-bytes lowercase hex of a pubkey>, <recommended relay URL, optional>]`
 - The `a` tag, used to refer to a (maybe parameterized) replaceable event
     - for a parameterized replaceable event: `["a", <kind integer>:<32-bytes lowercase hex of a pubkey>:<d tag value>, <recommended relay URL, optional>]`
-    - for a non-parameterized replaceable event: `["a", <kind integer>:<32-bytes lowercase hex of a pubkey>:, <recommended relay URL, optional>]`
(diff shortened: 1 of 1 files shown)


vitorpamplona:
At the time, we designed the non-parameterized replaceable-event spec to be just like the parameterized one with a d-tag equals to "", which creates that extra `:` at the end.

fiatjaf:
Yes, I think the ending colon is correct.

mikedilger:
Rust allows and prefers trailing commas.  Javascript disallows them.  Rust is correct.  So therefore this colon is correct.

avlo:
> At the time, we designed the non-parameterized replaceable-event spec to be just like the parameterized one with a d-tag equals to "", which creates that extra `:` at the end. In that way, there are always two `:` for parsers to find.

Understood & thanks for your reply, clarification and historical context.  Should either spec/parser become unwieldy/awkward, I'm confident the community is aware of generalized spec + parser alternatives.

avlo:
> Rust allows and prefers trailing commas. Javascript disallows them. Rust is correct. So therefore this colon is correct.

"Allows"/"Prefers" are design/use-case considerations...  not language considerations (or language restrictions).  See  @vitorpamplona reply, above for more info.

PR #1137, opened 2024-03-25
Add used fields to Calendar Time Event in NIP-52

vicariousdrama:
Cornychat.com relies on a `t` tag of `audiospace` for filtering.  Flockstr.com relies on optional and previously undocumented `audio` and `image` tags for rendering.

### DIFF
--- a/52.md
+++ b/52.md
@@ -90,7 +90,7 @@ The list of tags are as follows:
 * `end` (optional) exclusive end Unix timestamp in seconds. If omitted, the calendar event ends instantaneously.
 * `start_tzid` (optional) time zone of the start timestamp, as defined by the IANA Time Zone Database. e.g., `America/Costa_Rica`
 * `end_tzid` (optional) time zone of the end timestamp, as defined by the IANA Time Zone Database. e.g., `America/Costa_Rica`. If omitted and `start_tzid` is provided, the time zone of the end timestamp is the same as the start timestamp.
-* `about` (optional) description of the calendar event
+* `summary` (optional) brief description of the calendar event
 * `image` (optional) url of an image to use for the event
 * `location` (optional, repeated) location of the calendar event. e.g. address, GPS coordinates, meeting room name, link to video call
 * `g` (optional) [geohash](https://en.wikipedia.org/wiki/Geohash) to associate calendar event with a searchable physical location
@@ -113,7 +113,7 @@ The following tags are deprecated:
     ["d", "<UUID>"],
 
     ["title", "<title of calendar event>"],
-    ["about", "<details of the calendar event>"],
+    ["summary", "<brief description of the calendar event>"],
     ["image", "<string with image URI>"],
 
     // Timestamps
--- a/52.md
+++ b/52.md
@@ -90,10 +90,12 @@ The list of tags are as follows:
 * `end` (optional) exclusive end Unix timestamp in seconds. If omitted, the calendar event ends instantaneously.
 * `start_tzid` (optional) time zone of the start timestamp, as defined by the IANA Time Zone Database. e.g., `America/Costa_Rica`
 * `end_tzid` (optional) time zone of the end timestamp, as defined by the IANA Time Zone Database. e.g., `America/Costa_Rica`. If omitted and `start_tzid` is provided, the time zone of the end timestamp is the same as the start timestamp.
+* `about` (optional) description of the calendar event
+* `image` (optional) url of an image to use for the event
 * `location` (optional, repeated) location of the calendar event. e.g. address, GPS coordinates, meeting room name, link to video call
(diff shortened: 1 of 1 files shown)


staab:
Using `t` for metadata not meant to be read by users is a mistake. Imagine a valley girl shouting "hashtaaag #audiospace!" as a litmus test. I would prefer corny chat switch to using NIP 32 labels if possible.

vicariousdrama:
It's early enough that I can change the implementation in Corny Chat for this to be able to identify audiospace events.

If `t` is exclusively for hashtags, then I need to change some others in use as well

For context, Corny Chat is also adding a `t` tag for `cornychat.com` which could probably be an `l` tag instead. In addition, a `deleted` tag is being added which I may be able to refactor into using `expires`

vicariousdrama:
Kind of a pain to sort out what tag should be used as looking at NIP32, labels are a kind 1985 event, and I need to add this meta information to kind 31923

staab:
See "self-reporting" in NIP 32

vicariousdrama:
Switched to label tags and namespace in place of hashtag tags

52.md
```diff
@@ -90,9 +90,12 @@ The list of tags are as follows:
 * `end` (optional) exclusive end Unix timestamp in seconds. If omitted, the calendar event ends instantaneously.
 * `start_tzid` (optional) time zone of the start timestamp, as defined by the IANA Time Zone Database. e.g., `America/Costa_Rica`
 * `end_tzid` (optional) time zone of the end timestamp, as defined by the IANA Time Zone Database. e.g., `America/Costa_Rica`. If omitted and `start_tzid` is provided, the time zone of the end timestamp is the same as the start timestamp.
+* `about` (optional) description of the calendar event
```
tyiu, line 93: Why do we need `about` if we already have `content`?

52.md
```diff
@@ -110,6 +113,8 @@ The following tags are deprecated:
     ["d", "<UUID>"],
 
     ["title", "<title of calendar event>"],
+    ["about", "<details of the calendar event>"],
```
tyiu, line 116: Ditto.

52.md
AsaiToshiya: If so, I recommend that documenting the `about` tag as deprecated in NIP-24.

fiatjaf: Unfortunately some people are incapable of reading specs. Let's get rid of `content` and just use `about` everywhere?

tyiu: I disagree. Let's just update Flockstr to use content. I'm building a Comingle, a mobile events app. All the calendar events I've seen have content populated already. It's a trivial change and won't break existing clients. @zmeyer44  what do you think?

zmeyer44: I like the idea of content being used for a more expressive and detailed account of the event (possibly in Markdown) while about can serve as a short description, perhaps something that would be used in a preview card.

tyiu: Not opposed. [Long-form notes](https://github.com/nostr-protocol/nips/blob/master/23.md) do that. It uses a summary tag for a short description, and content for the full Markdown.

zmeyer44: I'm for changing 'about' to 'summary' if that would make things more clear and consistent

52.md
vicariousdrama: I also like supporting summary as a tag

52.md
tyiu: I've released Comingle to Apple TestFlight, which supports `summary`.

52.md
```diff
@@ -90,9 +90,12 @@ The list of tags are as follows:
 * `end` (optional) exclusive end Unix timestamp in seconds. If omitted, the calendar event ends instantaneously.
 * `start_tzid` (optional) time zone of the start timestamp, as defined by the IANA Time Zone Database. e.g., `America/Costa_Rica`
 * `end_tzid` (optional) time zone of the end timestamp, as defined by the IANA Time Zone Database. e.g., `America/Costa_Rica`. If omitted and `start_tzid` is provided, the time zone of the end timestamp is the same as the start timestamp.
+* `about` (optional) description of the calendar event
+* `image` (optional) url of an image to use for the event
```
tyiu, line 94: I've added support for `image` in Comingle.

tyiu:
Any other changes needed on this PR? Can we push this through?
@vicariousdrama @staab @zmeyer44 @fiatjaf @AsaiToshiya

tyiu:
I think we should push this PR through after replacing the `about` tag with `summary`, as discussed above. After that, we can merge the PR. @vicariousdrama

52.md
vicariousdrama: I'm changing the `about` field to `summary` in the next release of Corny Chat

PR #1142, opened 2024-03-28
Clarifies relays to be used for NIP-28

vitorpamplona:
Choosing relays for NIP-28 has been quite a challenge. With these changes, the owner of the chat can recommend relays in the metadata events.


staab (approved):
This has been a long time coming

PR #1144, opened 2024-03-28
NIP-59 as Primitives

vitorpamplona:
This tries to describe NIP-59 in terms of primitives and not as a fixed protocol (rumor -> seal -> wraps)

### DIFF
--- a/59.md
+++ b/59.md
@@ -6,8 +6,7 @@ Gift Wrap
 
 `optional`
 
-This NIP defines a protocol for encapsulating any nostr event. This makes it possible to obscure most metadata
-for a given event, perform collaborative signing, and more.
+This NIP defines event kinds and procedures for encapsulating other nostr events with variying degrees of privacy. The goal is to provide tools to obscure the right amount of metadata for separate use cases.
 
 This NIP *does not* define any messaging protocol. Applications of this NIP should be defined separately.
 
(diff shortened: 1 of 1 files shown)


59.md
```diff
@@ -6,157 +6,193 @@ Gift Wrap
...
-Encryption is done following [NIP-44](44.md) on the JSON-encoded event. Place the encryption payload in the `.content`
-of the wrapper event (either a `seal` or a `gift wrap`).
+`.content` contains the cyphertext of the [NIP-44](44.md) encryption of the JSON-encoded event. 
 
-# Other Considerations
+To increase privacy, `.created_at` MAY not be accurate. 
 
-If a `rumor` is intended for more than one party, or if the author wants to retain an encrypted copy, a single
-`rumor` may be wrapped and addressed for each recipient individually.
+Relays SHOULD only serve `kind 1059` events intended for the marked recipient based on user AUTH.
```
staab, line 73: Please no, this breaks nip 87 groups. Or at least requires the user to AUTH using the shared key when fetching group events.

59.md
```diff
@@ -6,157 +6,193 @@ Gift Wrap
...
-`rumor` may be wrapped and addressed for each recipient individually.
+Relays SHOULD only serve `kind 1059` events intended for the marked recipient based on user AUTH. 
 
-The canonical `created_at` time belongs to the `rumor`. All other timestamps SHOULD be tweaked to thwart
-time-analysis attacks. Note that some relays don't serve events dated in the future, so all timestamps
-SHOULD be in the past.
+Clients SHOULD only send wrapped events to destination relays that offer this protection.
 
-Relays may choose not to store gift wrapped events due to them not being publicly useful. Clients MAY choose
+Relays MAY choose not to store Gift-wrapped events due to them not being publicly useful. Clients MAY choose
```
staab, line 77:
```suggestion
Relays MAY choose not to store gift-wrapped events due to them not being publicly useful. Clients MAY choose
```

59.md
```diff
@@ -6,157 +6,193 @@ Gift Wrap
...
+  "pubkey": "<random-pubkey>",
+  "created_at": randomTimeUpTo2DaysInThePast()
+  "kind": 1059,
+  "tags": [
+    ["token", "<destination's push notification token>"],
+  ],
+  "content": nip44Encrypt(notificationEvent, pubkeyB, privkeyRandom)
+  "sig": "<random, one-time-use pubkey signature>"
+}
+```
```
staab, line 137: I don't think this needs to be included.

staab (changes requested):
Mostly an improvement, but please don't try to sneak in gating 1059's based on `p` tag, I don't think there is consensus on how best to do this yet.

vitorpamplona:
> gating 1059's based on p tag

That seems to be already on the text, but I didn't add anything to that account.

59.md
vitorpamplona: interesting. I thought this was already in the text, but we can roll it back.

59.md
vitorpamplona: I don't know.. the text without it feels like people are required to do rumor -> seal -> giftwrap which is not the case at all. Otherwise, every example is an example of the rumor -> seal -> giftwrap scheme, which I am starting to think it should be only used for DMs and other super private stuff. The average case for GiftWraps is much bigger than the private protocol.

vitorpamplona: Somebody could even use Seals with the receiver info in it for use cases where the sender and receiver can be connected but the rest of the event can never leak.

59.md
```diff
@@ -6,157 +6,193 @@ Gift Wrap
...
+The inner event MUST always be unsigned.
 
-## 3. Gift Wrap Event Kind
+To increase privacy, `.created_at` MAY not be accurate. 
 
-A `gift wrap` event is a `kind:1059` event that wraps any other event. `tags` SHOULD include any information
-needed to route the event to its intended recipient, including the recipient's `p` tag or [NIP-13](13.md) proof of work.
+## The Gift Wrap Event Kind
+
+A `gift wrap` event is a `kind:1059` event that wraps any other event. `tags` SHOULD include any information needed to route the inner event to its intended recipient, including the recipient's `p` tag or [NIP-13](13.md) proof of work.
```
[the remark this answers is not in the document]
vitorpamplona: added

59.md
vitorpamplona: Changed to MAY

vitorpamplona:
I think when we moved the DM protocol to this NIP, most of the hyper private needs came in as well. Most use cases don't need that level of privacy, like the nip 87 groups

59.md
staab: Ah, my bad, you're right. I guess we can leave it. I may have to re-think some things.

vitorpamplona: We should get rid of it if it is getting on your way. I tried to soften the requirements to MAY, but let me know if it is still not what you need.

staab: We should probably get rid of it. It doesn't belong in this NIP at the very least. I could see gift wraps being created with no `p` tag, preferring instead some other way of addressing recipients. It definitely limits the extent to which this NIP can be used in creative ways.

At the same time, this is probably what relays will do by default, so implementers should probably think about how they can work around the recipient auth limitations 🤷‍♂️

staab: I know there have been other proposals to solve this. Maybe `-` is a good solution, but I'm not sure the nuances right now.

fiatjaf (approved):

fiatjaf:
Let's merge this?

staab:
I still don't like "3. Relays MAY only serve `kind 1059` events intended for the marked recipient based on user AUTH" because it breaks nip 87 groups. But this is done in practice, so it should stay in here. Maybe I can switch to a different kind for grouo messages. This reduces the anonymity set, but it will still essentially work.

59.md
```diff
@@ -6,157 +6,193 @@ Gift Wrap
...
   "sig": "<random, one-time-use pubkey signature>"
 }
 ```
 
-# Encrypting Payloads
+`.content` contains the cyphertext of the [NIP-44](44.md) encryption of the JSON-encoded event. 
+
+To maximize privacy: 
+1. `.created_at` MAY not be accurate. 
+2. The `extension` tag can be used to minimize the existence of the event after a certain date.
```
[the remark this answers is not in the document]
vitorpamplona: fixed

vitorpamplona:
> I still don't like "3. Relays MAY only serve `kind 1059` events intended for the marked recipient based on user AUTH" because it breaks nip 87 groups. But this is done in practice, so it should stay in here. Maybe I can switch to a different kind for group messages. This reduces the anonymity set, but it will still essentially work.

So, members AUTH with their usual key and not the secret key that would match the p-tag of the wrap? Doesn't that require a custom relay to know which members are authorized to download which wrapped events?

staab:
> So, members AUTH with their usual key and not the secret key that would match the p-tag of the wrap? Doesn't that require a custom relay to know which members are authorized to download which wrapped events?

They auth with their regular key, yeah, and request wraps with the assumption that they're not gated. This was true when I built the implementation and spec, but it no longer holds. Users of NIP 87 now have to choose relays that don't enforce auth for 1059s, or else their groups stop working. The alternative would be for users to auth with the group key when downloading messages, but that would require closing and re-opening their connections. Not impossible I guess.

PR #1145, opened 2024-03-28
NIP-95 Revisit

arthurfranca:
Read [here](https://github.com/arthurfranca/nips/blob/nip-95-revisit/95.md)

I was reading #345 new comments and came up with this spec.

Differences:
- a random pubkey (not the uploader's one) used just once as author of the file event(s) is what identifies the file
- file can be made of multiple chunks, all with the same above pubkey
- new NIP-65 [flag](https://github.com/nostr-protocol/nips/pull/991) to configure user's "file relays" cause most relays won't accept NIP-95 events
- `nfile` entity

### DIFF
--- a/95.md
+++ b/95.md
@@ -131,4 +131,8 @@ These two tag values can instead be encoded on an `nfile`.
 Upon seeing such tags, in order to download the file, client should search for `kind:1195` events
 by author (`f` tag value) on the `nip95u` uploader's "file relays".
 
+Filter example:
+
+`{ authors: ["<same-random-pubkey-for-all-event-chunks>"], kinds: [1195] }`
+
 It must join the `.content` values from all the `kind:1195` events then convert it to bytes to recreate the file.
--- a/95.md
+++ b/95.md
@@ -24,7 +24,7 @@ into multiple chunks or use a single event.
 The event `.pubkey` is NOT the uploader's one but a randomly generated pubkey. All chunks
 use the same pubkey. The pubkey MUST NOT be reused on events other than the
 `kind:1195` event set of a single file.
-**The pubkey is what is used to uniquely identify the file made of a set of chunk events**
+**The pubkey is what is used to uniquely identify the set of chunk events that make up the file**
 
 The event `.content` is the base64-encoded file data.
 
@@ -70,10 +70,29 @@ Before uploading `kind:1195` events of a specifc file,
 user MUST publish a single `kind:1095` "Uploader" event,
 which is an event authored by the **user's main pubkey**,
 with empty `.content` and an `f` tag referencing the `kind:1195`
-event set (the `kind:1195` event(s) `.pubkey`).
+event set (the `kind:1195` event(s) `.pubkey`). It also
+has a `x` tag with the full file's hex-encoded sha256 hash that the
+file relay may later validate.
 
 Its purpose is to register the user as an uploader of the file on a specific relay and to request permission
 for uploading the file chunks.
+
+Example:
+
+```js
+{
+  "id": "<id>",
+  "pubkey": "<user's-main-pubkey>"
+  "kind": 1095,
+  "tags": [
+    ["f", "<same-random-pubkey-for-all-event-chunks>"],
+    ["x", "<Hash SHA-256>"]
+  ],
+  "content": "",
+  "created_at": <timestamp>,
+  "sig": "<signature>"
+}
+```
 
 When receiving the `kind:1095` event, relay may answer with one of the following messages and prefixes:
 
@@ -98,14 +117,14 @@ Clients may download `kind:1095` events from user's file relays to list all user
 A [NIP-19](19.md) `nfile` bech32-encoded entity may be embedded directly on notes using a [NIP-21](21.md) URL.
 It translates to two values:
 
-1) The `kind:1195` set's pubkey, identifying the file;
+1) The `kind:1195` set's pubkey, that is the address of the file chunks;
 2) The uploader's pubkey, to be able to find the right "file relays" to download the file from.
 
 ## Download
 
-This NIP introduces an `f` tag filled with the `kind:1195` pubkey, uniquely identifying the file,
+This NIP introduces an `f` tag filled with the `kind:1195` pubkey, uniquely identifying the file chunk set,
 and a `nip95u` tag set to the pubkey of the user who uploaded the file. Both are to
-be present if using thee NIP-94-like `kind:1094` event (instead of the `url` tag).
+be present if using the NIP-94-like `kind:1094` event (instead of the `url` tag).
 
 These two tag values can instead be encoded on an `nfile`.
 
--- a/95.md
+++ b/95.md
@@ -19,16 +19,17 @@ The chunk events may be sent on separate websockets connections for parallel upl
 ## Event
 
 An event of `kind:1195` represents a file chunk. A client may choose to split a file
-into multiple chunks or use a single event.
+into multiple chunks or use a single event. If using chunks, all of them MUST
+have the same size, except for the last one.
 
-The event `.pubkey` is NOT the uploader's one but a randomly generated pubkey. All chunks
-use the same pubkey. The pubkey MUST NOT be reused on events other than the
-`kind:1195` event set of a single file.
-**The pubkey is what is used to uniquely identify the set of chunk events that make up the file**
+The event `.pubkey` may be the uploader's one or a randomly generated pubkey.
+
+All chunks share the same [NIP-61](61.md) unbound list `n` tag that is set to `"<toHex(sha256(full-file))>-<chunk-byte-size>"`
+(here the chunk size is always the greater one, because the last chunk size may be smaller than the others).
 
 The event `.content` is the base64-encoded file data.
 
-It has a `c` tag. Its first falue is the chunk number while the second one is the total number of chunks.
+It has a `c` tag. Its first value is the chunk number while the second one is the total number of chunks.
 The client can use the `c` tag to ask for missing chunks when resuming a download.
 
 Example:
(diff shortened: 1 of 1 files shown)


vitorpamplona:
File relays and nfile are good ideas, but why the random pubkey? I don't get it.

arthurfranca:
> but why the random pubkey?

The NIP is missing a section that I'm gonna add soon. The idea is that the same chunk event set may have multiple "owners/uploaders" to avoid duplication. Because of that, one reason for using a random pubkey is that the chunk event author does not need to be the main pubkey of the user who first uploaded the file.

The random pubkey would be used as author of all chunk events of a single file and then should never be reused as author of anything else. To get all chunks, client filter by `{ authors: ["<the-random-pubkey>"], kind: [1064] }`. This way the pubkey would identify this "version" of a file (the same file can have many "versions", e.g. it can be split into 3 chunks or 1 or 10 which would be 3 versions).

But some user may misbehave and reuse the key so maybe it is not a good way to group all chunks of a file.

vitorpamplona:
Interesting, so the point is to find a way to query the chunks of a single file without knowing all the event ids beforehand and without allowing other people to add a malicious chunk in the middle of your file: 
```js 
{ authors: ["<key-per-file>"], kind: [1064] }
```

Meaning: If the event header has all the ids in a list, no one can add a malicious chunk, but you have to create a filter with all the ids from that event and that filter can be huge. 
```js 
{ ids: [<huge list of chunk event ids>] }
```

Alternatively, chunk events can tag an unbound list. But, since it's very easy for anyone else out there to create a new malicious chunk and also point to the same list, the filter must include the author of the header: 

```js
{ #n: ["<list name>"], authors: ["owner"], kind: [1064] }
```

vitorpamplona:
`n` could be the file hash that is already computed for the header event.

95.md
```diff
@@ -0,0 +1,108 @@
...
+Relay File Storage
+------------------
+
+`draft` `optional`
+
+This NIP defines a way to store binary data on relays. It supports parallel uploads and resumable downloads.
+
+## Upload
+
+To upload a file, first client must convert its bytes to base64. It may do it in chunks made of multiples of 3 bytes or in one go.
```
vitorpamplona, line 13: 3 bytes?

95.md
```diff
@@ -0,0 +1,108 @@
...
+## Metadata
+
+Metadata should be added using [NIP-94](94.md) or other metadata-related NIP.
+
+## File Relay
+
+Client should upload to user's "file relays", which use the [NIP-65](65.md) `f` flag.
+When downloading a file uploaded with this NIP, it should search on the uploader's "file relays".
+
+**Relays must NOT honor `kind:5` deletion events referencing file chunk events.** Deletion
```
vitorpamplona, line 58: No need for this. I might want to get a separate chunk of the file in each relay. Deleting chunks should be possible and it should be fine.

95.md
```diff
@@ -0,0 +1,108 @@
...
+Its purpose is to register the user as an uploader of the file on a specific relay and to request permission
+for uploading the file chunks.
+
+When receiving the `kind:1065` event, relay may answer with one of the following messages and prefixes:
+
+- `["OK", "<kind:1065-event-id>", false, "auth-required: ..."]`: Authentication is required to register an uploader;
+- `["OK", "<kind:1065-event-id>", true, "uploaded: ..."]`: The corresponding `kind:1064` file chunks are already uploaded, trying to re-upload them will fail;
+- `["OK", "<kind:1065-event-id>", true, "upload: Missing chunks 1, 2, 7, 10"]`: File isn't uploaded yet or incomplete, user is allowed to upload it on this ws connection;
+
+Trying to send a `kind:1064` event before a `kind:1065` one should fail.
```
vitorpamplona, line 80: I don't like these custom behaviors for relays.

95.md
arthurfranca: multiples of 3 cause base64 needs atleast 3 bytes to encode. it could be 255000 bytes per chunk for example

NfNitLoop:
I'm new to the Nostr community, so I don't know how things are done here, but FYI, making an entirely new PR and re-using a NIP number[^1] that's already in progress (and has implementations in the wild) sort of comes off like you're trying to hijack authorship and/or the conversation. (I assume good will here, so I'm not claiming that's what you're doing! But if someone didn't have good will, that's how this would seem.) Just wanted to let you know.

A more collaborative approach might be to target these changes to be merged into the ongoing work over in https://github.com/nostr-protocol/nips/pull/345.
* Collaborates more directly w/ the ongoing conversation. 
* Easier to see the diffs from that document in the PR.
* Shares authorship.

However, there are a lot of issues that folks are pointing out here that have already been resolved in that other PR, so I'm not sure how likely the owner of that PR might be to merge these changes. 😅 


[^1]: Though, these things do seem to stay in-progress for a while, and there doesn't seem to be a great way to "grab" a NIP number while working through things, so I could see this happening accidentally as well.  But in this case you're clearly referencing the other one.

95.md
arthurfranca: I thought of the following flow, maybe it is bogus:
- An userA uploads a file.
- Another userB sees it on his client and asks client to copy and upload it too (to make it his file) and both users happen to use the same file relay.
- UserB client instead of re-uploading it, just registers userB as an owner/uploader of the file that is already on the relay.

Now that userA and userB both own the file, we can't let userA delete the chunks.. userA can just unregister himself as not an uploader/owner anymore. The file is deleted if there is no registered uploader.

95.md
arthurfranca: The reason for this is that a file relay can't let a client send a big event (a file chunk) just to later reply that the client/user had no authorization to do it. It wastes relay resources so I imagine file relay will start with a capped max ws message size for every new ws connection until it sees a `kind:1065` event "asking" for authorization to upload a `kind:1064` event.

95.md
```diff
@@ -0,0 +1,108 @@
...
+a file relay may charge each of them half of the costs to host the file.
+
+Relay should delete the `kind:1065` event if no corresponding `kind:1064` event is uploaded within a reasonable period.
+
+Clients may download `kind:1065` events from user's file relays to list all user files.
+
+## Download
+
+This NIP introduces a NIP-94 `f` tag filled with the `kind:1064` pubkey, uniquely identifying the file,
+and a `nip95u` tag set to the pubkey of the user who uploaded the file.
```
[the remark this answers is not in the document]
arthurfranca: True

arthurfranca:
@NfNitLoop I get your point but authors aren't listed anymore on NIPs. I did steal many ideas while adding some of my own and glued them together here cause there were many changes that needed to be placed together to make it cohesive and would be hard to explain and ask them to be considered separately there at #345. 

My goal is solely to help come up with the best version of a NIP we could. This one was my vision of how it could look like.

I can change the NIP number and the kinds.. doesn't matter, I just put the text here for whoever may be insterested to discuss if it is better, worse or if could be improved further or ditched in favor of a better version.

arthurfranca:
@vitorpamplona I think I may have confused you by reusing the `kind:1065` to mean something other than file metadata. 

On this NIP, `kind:1065` means "uploader" and would look like this:
```js
{
  kind: 1065,
 pubkey: "<uploader-main-pubkey>",
  ...,
  tags: [
    ["f", "<key-per-file>"]
  ]
}
```

There could have a NIP-94 event or a copy with another number like you suggested that would have the metadata tags like:
```js
{
  kind: 10xx,
  pubkey: "<user-main-pubkey>",
  ...,
  tags: [
    ["f", "<key-per-file>"],
    ["nip95u", "<uploader-main-pubkey>"], // most times it is the same as "<user-main-pubkey>"
    ["size", "..."],
    ["dim", "..."],
    ["blurhash", "..."]
  ]
}
```

NIP-94 event isn't required cause nfile with or without NIP-54 inline metadata could be used instead inside a `kind:1` for example.

I will change all the kind numbers.

95.md
vitorpamplona: Are you trying to register many owners for shared chunks and only allow deletion when all owners request or stop using the chunk?

arthurfranca: Yes, anyone the file relay authorizes can become an owner ("uploader) of a file chunk set.

There is an "uploader event" for that. When no "uploader event" is present on the relay anymore (deleted with `kind:5`), the file relay is free to automatically delete the chunk set (not using `kind:5` here, just auto-deletes).

95.md
arthurfranca: Bad english, tried to improve text

vitorpamplona:
The gains of using random pubkeys to represent files are still not clear to me though.

arthurfranca:
@vitorpamplona It is true that a file chunk set fits well into the [unbound list](https://github.com/nostr-protocol/nips/pull/784/files#diff-468091cdc566caf598af0eed51638d38573d377e486e6322ae4d6a56090adf09R32-R34) spec, that addresses a set with the owner pubkey + n tag (that could be set to the sha256 as you said).

But it may not be a perfect fit. Because a pubkey may (don't know why it would want to do it but it is possible) upload the same file with the same hash twice but with a different set of chunks.

Example:
First time it sends 3 chunks of 9 bytes
Second time it sends 1 chunk of 27 bytes

Now we got two versions of the same file. 4 chunks that shouldn't belong on the same unbound list.

vitorpamplona:
Yeah, that would require not using the hash of the file as the name of the unbound list, but we could do `hash`+`blocksize` and then have 2 entries in the header pointing to the two unbound list the receiver can choose to download.

arthurfranca:
Right, I will edit it.

arthurfranca:
@vitorpamplona now it is using unbound list. One thing left is adding `nfile` to NIP-19 but not sure yet how to do it.

@NfNitLoop @frbitten what do you think of this version of NIP-95?

arthurfranca:
Reviewing this I think it has some problems:
1) on upload: serializing a somewhat big payload (chunk event) to sign it is bad;
2) on download: need to put file chunk content in memory at once to send it as a nostr event, though small chunks wouldn't be that of a problem;
3) storage: it feels wrong to possibly store many versions (sets of chunks of variable chunk sizes) of the same file (same sha256). somehow the identifier should be just the sha256 hash instead of it plus chunk size;
4) of course the base64 encoding/decoding step isn't ideal too;

NIP-96 is better. #719 may be good too

PR #1146, opened 2024-03-29
NIP35: Relay-Specific Notes

jb55:
This is a very early/experimental idea on relay-specific notes. Initial feedback welcome!

### DIFF
--- a/35.md
+++ b/35.md
@@ -1,0 +1,46 @@
+
+NIP 35
+======
+
+Relay-Specific Notes
+--------------------
+
+This NIP describes a method for which notes are only valid on the relays that they are specifically destined for. It does this via a fairly extreme measure: adding an additional data to the note-id commitment. This additional data ensures that the note is not valid on existing relays and clients.
+
+New and updated clients/relays that support this will be able to accept and filter notes that are destined for them, enabling many new use cases such as more private relays.
+
+## Modified note commitment
+
+To support relay specific notes, a new JSON field is added to the note called `to`. `to` is a list of strings of relays which the note is destined for. For example:
+
+```json
+{
+  "pubkey": "...",
+  "to": ["wss://relay.jb55.com",""]
+}
+```
+
+## Calculating the note id
+
+The new note id calculation is like so:
+
+```
+[
+  1,
+  <pubkey, as a lowercase hex string>,
+  <created_at, as a number>,
+  <kind, as a number>,
+  <tags, as an array of arrays of non-null strings>,
+  <content, as a string>,
+  <to, as an array of non-null strings>,
+]
+```
+
+## Client validation
+
+Clients SHOULD reject relay-specific notes where the relay it is receiving the note from isn't in the `to` field.
+
+## Server validation
+
+Servers MUST reject relay-specfic notes that are not intended for them by comparing the hostname of the relay to the relays in the `to` field.
+
--- a/README.md
+++ b/README.md
@@ -50,6 +50,7 @@ They exist to document what may be implemented by [Nostr](https://github.com/nos
 - [NIP-31: Dealing with Unknown Events](31.md)
 - [NIP-32: Labeling](32.md)
 - [NIP-34: `git` stuff](34.md)
+- [NIP-35: Relay-Specific Notes](35.md)
 - [NIP-36: Sensitive Content](36.md)
 - [NIP-38: User Statuses](38.md)
 - [NIP-39: External Identities in Profiles](39.md)


alltheseas:
What happens if I am connected to relay.jb55.com, receive a note intended only for this relay, and I reply broadcast the parent note to all my other relays, most of which do not recognize this NIP/spec?

staab:
I had this exact same idea a few months back. I'm still not sure if it's good or not. Possible downsides:

- If a relay goes away, all its content dies with it
- Proxies can't function as simply
- This can be easily be bypassed by just not validating the url (this point conflicts with my first one. One possible way to fix it would be to make `to` implicit, so you'd have to know what relay the note came from in order to validate the signature, or you could brute force the sig against all known relays)

> What happens if I am connected to relay.jb55.com, receive a note intended only for this relay, and I reply broadcast the parent note to all my other relays, most of which do not recognize this NIP/spec?

The relays will reject it because the signature will be invalid.

jb55:
> I had this exact same idea a few months back. I'm still not sure if it's good or not. Possible downsides:
> 
> * If a relay goes away, all its content dies with it

seems like a feature to me!

> * Proxies can't function as simply

true, haven't thought of a good solution for this yet but maybe its not too big of a deal. this is why I put a "SHOULD" in the client validation wording...

> * This can be easily be bypassed by just not validating the url (this point conflicts with my first one. One possible way to fix it would be to make `to` implicit, so you'd have to know what relay the note came from in order to validate the signature, or you could brute force the sig against all known relays)

this is where both client and server validation is important, both client and server would have to not do validation to propagate the note outside the intended location.

itawa-panema:
As users can not trust clients or relays to not (re-)propagate their events to relays that doesn't support this NIP, I wonder what would be the real use cases of this?

staab:
> seems like a feature to me!

You're right, this is a great addition.

jb55:
> As users can not trust clients or relays to not (re-)propagate their events to relays that doesn't support this NIP, I wonder what would be the real use cases of this?

What do you mean? These are not valid nostr events to current relays. They are not propagatable.

itawa-panema:
> > As users can not trust clients or relays to not (re-)propagate their events to relays that doesn't support this NIP, I wonder what would be the real use cases of this?
> 
> What do you mean? These are not valid nostr events to current relays. They are not propagatable.

That's a good point but it doesn't prevent "rogue" relays to keep the events anyway (received from candid clients). So I was interested on what would be the use cases (as the privacy can't be one)

vitorpamplona:
I got a lot of flack for suggesting something similar on #1029, but I do think the change in the serialization algorithm is the answer for relay-controlled events. 

I just don't know if it is a good idea to give that much power to relays. 

For the specifics of this proposal, changing the serialization to block events from reaching current relays will only work until StrFry includes the `to` field in their serialization without checking the current domain name of the relay. Then more relays will accept everything.

staab:
> For the specifics of this proposal, changing the serialization to block events from reaching current relays will only work until StrFry includes the to field in their serialization without checking the current domain name of the relay. Then more relays will accept everything.

As @jb55 points out though, that would be non-standard behavior, and well-behaved clients would consider these invalid events. So it requires collusion between clients and relays to bypass this. In the aggregate, this will I think allow for pretty robust relay de-commodification.

Maybe the biggest problem is that the correct relay has to be selected by the client signing the event, which might not align with the events the recipient relays actually want to store.

alltheseas:
> The relays will reject it because the signature will be invalid.

Do all relays check signatures today?

vitorpamplona:
> As @jb55 points out though, that would be non-standard behavior, and well-behaved clients would consider these invalid events. So it requires collusion between clients and relays to bypass this. In the aggregate, this will I think allow for pretty robust relay de-commodification.

I don't know... I don't trust people to run well-behaved anything these days... Especially when protected data (and thus more valuable data) is involved. 

But I am not against it if somebody wants to push this forward.

jb55:
> > The relays will reject it because the signature will be invalid.
> 
> Do all relays check signatures today?

yes

alltheseas:
> > > The relays will reject it because the signature will be invalid.
> > 
> > Do all relays check signatures today?
> 
> yes

Given this, does a single adversarial relay, or relay + client combo undo this NIP's stated benefit?

staab:
> Given this, does a single adversarial relay, or relay + client combo undo this NIP's stated benefit?

No, because it will still work for anyone who cares enough to run a compliant relay or client.

alltheseas:
> > Given this, does a single adversarial relay, or relay + client combo undo this NIP's stated benefit?
> 
> 
> 
> No, because it will still work for anyone who cares enough to run a compliant relay or client.

Sorry for the left curve questions. Last one: if your company had "trade secrets" would you be confident in using this approach - i.e. only post these to this relay only?

jb55:
> > > Given this, does a single adversarial relay, or relay + client combo undo this NIP's stated benefit?
> > 
> > No, because it will still work for anyone who cares enough to run a compliant relay or client.
> 
> Sorry for the left curve questions. Last one: if your company had "trade secrets" would you be confident in using this approach - i.e. only post these to this relay only? 

You still would want AUTH or vpn/ip blocking as well, this doesn't prevent leaking anything. Screenshots can always do this. What this does stop is making notes trivially propagatable via "broadcast" and "repost" features in clients. The ease of propagatability is a huge downside that makes private use cases not feasible on nostr.

In the repost case, you cannot repost relay-specific notes, because the inner signature doesn't validate. Same as broadcast.

Another example: it's not easy to make someone's comment appear in many different discord servers. In their case it's not a problem to enforce this because it's centralized, but it would be nice to have a similar mechanism in a decentralized network.

Initially I thought of just adding a "to" tag, but everyone would ignore it so that's a dead end. This is a similar problem with @fiatjaf 's `-` proposal.

fiatjaf:
One problem with this is that I can't change my mind and publish my events to other relays after creation, everything is set in stone.

I can't migrate my events from one relay to another or other basic things. Depending on the situation this might lead to people copying events and resigning them with different ids, breaking references and all that.

fiatjaf:
The [`["-"]` approach](https://github.com/nostr-protocol/nips/pull/1030) solves that by given the event creator agency over where the event can go at any point in time. I also think it's saner in the long run, as it would take just a few lines of code to hardcode it in the 3 or 4 relay implementations that exist (which after some time will make their way into currently deployed relays).

Here's the implementation on Khatru: https://github.com/fiatjaf/khatru/commit/cd4c25c84519a4965fdb48f462111373d38a5eb8 (this implementation already includes the `AUTH` stuff, a simpler implementation would be to just block events with the `["-"]` tag, which would be really 3 or 4 lines).

Meanwhile this proposal requires cumbersome changes in all of the dozens of Nostr libraries in existence so they can understand the new signature algorithm.

fiatjaf:
Now I'm thinking of a way to bootstrap the `["-"]` mechanism such that it will only work today on relays that voluntarily decide to implement it: clients can just send the event with some invalid part in it, like a single extra letter appended to the `"sig"` field. And compliant relays will remember to check that and remove that extra letter when the received event has a `["-"]` tag. Relays can just append that extra letter when sending these events to clients too, so only compliant clients will understand these.

Eventually as relays catch on with the implementation we can remove this hack and start sending normal signatures everywhere.

vitorpamplona:
Oh boy, are we getting 3 different serialization changes?

pablof7z:
> Now I'm thinking of a way to bootstrap the `["-"]` mechanism such that it will only work today on relays that voluntarily decide to implement it: clients can just send the event with some invalid part in it, like a single extra letter appended to the `"sig"` field. And compliant relays will remember to check that and remove that extra letter when the received event has a `["-"]` tag. Relays can just append that extra letter when sending these events to clients too, so only compliant clients will understand these.
> 
> Eventually as relays catch on with the implementation we can remove this hack and start sending normal signatures everywhere.

this sound like versioning the protocol in a terrible way

fiatjaf:
> this sound like versioning the protocol in a terrible way

I agree, I very much prefer that relays just implement handling the `["-"]` properly.

jb55:
> like a single extra letter appended to the `"sig"` field. And compliant relays will remember to check that and remove that extra letter when the received event has a `["-"]` tag. Relays can just append that extra letter when sending these events to clients too, so only compliant clients will understand these.

This was my initial idea when I started thinking about this, but I then realized it would be trivial for people to strip that a re-propagate it.

> I can't migrate my events from one relay to another or other basic things. Depending on the situation this might lead to people copying events and resigning them with different ids, breaking references and all that.

yes this is either crappy or really good depending on how much you want these notes to be propagated.

Maybe if the relays had associated pubkeys instead it would alleviate these issues?

jb55:
I didn't see the `-` spec until now. I think it's fine and is much simpler. although not sure how long it will take for all relays to implement. strfry still doesn't have AUTH support :(

This spec doesn't need all relays to update so it works out of the gate which is nice.

35.md
```diff
@@ -0,0 +1,46 @@
...
+Relay-Specific Notes
+--------------------
+
+This NIP describes a method for which notes are only valid on the relays that they are specifically destined for. It does this via a fairly extreme measure: adding an additional data to the note-id commitment. This additional data ensures that the note is not valid on existing relays and clients.
+
+New and updated clients/relays that support this will be able to accept and filter notes that are destined for them, enabling many new use cases such as more private relays.
+
+## Modified note commitment
+
+To support relay specific notes, a new JSON field is added to the note called `to`. `to` is a list of strings of relays which the note is destined for. For example:
```
AsaiToshiya, line 14: What about using clearer `relays` instead of `to`?

danieldaquino:
I like this spec more and more. The only downside in my mind was illustrated in this comment from @fiatjaf:

> One problem with this is that I can't change my mind and publish my events to other relays after creation, everything is set in stone.
> 
> I can't migrate my events from one relay to another or other basic things. Depending on the situation this might lead to people copying events and resigning them with different ids, breaking references and all that.

However, I just had a thought that perhaps this is not an issue, but actually a strength, a feature. It is technically possible to migrate things to a different set of relays by recreating events and references "en masse" — it only requires either one of two things:
- Cooperation from everyone involved (to make the new signatures)
- Intentional programming on particular clients and relay configurations to accept address discrepancies, some internal "knowledge" of said data migration

If there is general consensus among a controlled group (e.g. a company), these methods above are not very difficult — provided enough tooling to do this.

Therefore, this kind of friction may sound bad, but could actually be a strength that protects private data, by requiring a degree of consensus before moving data somewhere that could be considered less private by members of such a group of people.

Example: If I post something meant to go to "privatecompanyrelay.com", but someone wants to move it to "lessprivaterelay.com", it is harder to do so without my consent. If they have my consent, I can approve the migration by signing the updated event

alltheseas:
In the context of a company, rarely are these operated by consensus. Usually there is one or a few decision maker.

In the context of a e.g. community such as the ones @staab is building for, perhaps there is more of a need for consensus. Maybe I'm wrong as communities have leaders as well.

danieldaquino:
> In the context of a company, rarely are these operated by consensus. Usually there is one or a few decision maker.

True, so in practice the decision maker of a company would ask team members to re-sign their events to include the new relay(s), and those team members would most likely just comply.

Worst case scenario _(e.g. if a private key is lost and re-signing is not an option)_, the company's relays/clients could be programmed to accept the old pre-migration notes (i.e. explicit exceptions to the rules in this NIP)

> In the context of a e.g. community such as the ones @staab is building for, perhaps there is more of a need for consensus. Maybe I'm wrong as communities have leaders as well.

In the context of a community, a lack of migration consensus means that propagating data from the non-consenting author would become a bit more difficult, which is also a plus, as the author may not want their notes to be on the new location.

Overall, data migration is still possible under this NIP, it just requires a bit of a process/formality, which I consider to be a good thing in the context of handling private data and avoiding leaks.

vitorpamplona:
The most difficult part of this approach is to make sure privacy expectations are correctly set among participants: users/company/community, meaning: 
1. These are not private events by any means. Anyone can read them when they leak. 
2. They can be verified as authentic when leaked. 
3. Somebody can put up a relay that simply copies all these events to a new relay for public consumption in real time. That new relay would simply ignore the rule to verify if it is one of the `to` domains in the URL. It's very easy to imagine a NostrLeaks relay that disregards the `to` rule. 

I run 3 private relays for companies. Two use normal relays but one of them runs #1029 (validates per user, instead of per relay), which has similar privacy framework as the proposal here. It was not fun when they discovered that events could leak and be verifiable. Most companies can deal with a "screenshot" leaking, but if a signed package leaks, things get to a whole new level of problems. So, managing expectations is important. 

In my experience, user-initiated migration is not likely to happen within companies because they own your communications when inside of it. But company-initiated migrations will likely happen (e.g. when changing the relay's DNS name, branding changes, etc). And when it happens, the company won't ask past employees to re-sign payloads. The new relay code will have an exception to accept the old and the new DNS domain as valid. So, I do expect `to` exceptions to become the norm.

danieldaquino:
> The most difficult part of this approach is to make sure privacy expectations are correctly set among participants: users/company/community, meaning:
> 
> 1. These are not private events by any means. Anyone can read them when they leak.

If they leak, anyone can read them, yes. But the same is true about leaked emails or any kind of private communications. Emails and other private platforms prevent unauthorized access and accidental leaks, but they cannot prevent an internal malicious actor from intentionally leaking private info.

The good thing about this NIP is that it prevents accidental leaks by:
- Not being compatible with existing clients/servers that do not know or care about this NIP.
- Forcing clients and servers to implement logic that explicitly handles these private data rules — if they want to be compatible with this NIP


> 3. Somebody can put up a relay that simply copies all these events to a new relay for public consumption in real time. That new relay would simply ignore the rule to verify if it is one of the `to` domains in the URL. It's very easy to imagine a NostrLeaks relay that disregards the `to` rule and captures all of the corporate notes it can.

In this scenario, are you assuming this to be an outsider attack (e.g. An outsider waiting for accidental leakages) or an insider attack (e.g. A malicious person with access to a private group)?

> I run 3 private relays for companies. Two use normal relays but one of them runs #1029 (validates per user, instead of per relay), which has similar privacy framework as the proposal here. It was not fun when they discovered that events could leak and be verifiable. Most companies can deal with a "screenshot" leaking, but if a signed package leaks, things get to a whole new level of problems. So, managing expectations is important.

Interesting. I don't quite understand what difference it makes in practice, can you please provide some examples of how digitally signed communications leaking becomes a much bigger problem than normal data leaking?

> In my experience, user-initiated migration is not likely to happen within companies because they own your communications when inside of it. But company-initiated migrations will likely happen (e.g. when changing the relay's DNS name, branding changes, etc). And when it happens, the company won't ask past employees to re-sign payloads. The new relay code will have an exception to accept the old and the new DNS domain as valid. So, I do expect `to` exceptions to become the norm in the long run.

Fair point, perhaps that will be the standard strategy people will adopt under this NIP

jb55:
On Wed, Apr 17, 2024 at 10:29:57AM -0700, Daniel D’Aquino wrote:
>> I run 3 private relays for companies. Two use normal relays but one of them runs #1029 (validates per user, instead of per relay), which has similar privacy framework as the proposal here. It was not fun when they discovered that events could leak and be verifiable. Most companies can deal with a "screenshot" leaking, but if a signed package leaks, things get to a whole new level of problems. So, managing expectations is important.
>
>Interesting. I don't quite understand what difference it makes in practice, can you please provide some examples of how digitally signed communications leaking becomes a much bigger problem than normal data leaking?

This started making me think of strategies of tweaking signatures with
private information so that even if the note was leaked it would not be
possible to verify the signature without this tweaked key. I am not a
cryptographer so I'm not sure how you would do this, but deniability
would be a really cool feature.

vitorpamplona:
> This started making me think of strategies of tweaking signatures with private information so that even if the note was leaked it would not be possible to verify the signature without this tweaked key.

Yeah, we need to find a way to do this. There are many schemes that can improve the situation but because the user is verifying the payload in their device, the user's device must know how to reassemble the payload. And once it knows that, the client can leak the reassembled payload. It's better, but not really solved. 

A separate way would be to have a verifier device: a USB-C dongle that verifies payloads for the user's device. In that case, the relay can put a secret into the device (which can be unique per user) and use that secret to reassemble the payload and verify inside of it. Since the user never gets to see the secret, it is impossible to leak a verifiable version of the payload. 

That device can also be connected and rotate secrets at every connection to the relay. In that way, if the user leaks the unverifiable payload, you can actually know which person did it and when.  From that perspective, it is possible to revoke access to verifiable content in real-time. Which is a thing many companies will ask for (e.g. important when firing bad employees).

jb55:
I found this ! https://en.wikipedia.org/wiki/Undeniable_signature

jb55:
not sure how amenable it is to nostr though ...

jb55:
another dumb thing I thought of is creating a subidentity (perhaps something like a delegation) that creates notes on the main keys behalf, but after some epoch it leaks the key, so that in a leak sitation, you can't tell if the leaked data contains forged notes for that epoch

interactively, the relay could prevent accepting notes from older epochs.

danieldaquino:
If deniability is a concern for a particular user, wouldn't a simpler solution for them be to simply use separate public/private key identities for different purposes?

e.g. Alice uses `npub123` for her public profile, and `npub456` within her company's internal relay. In practice it would be harder to prove that `npub456` actually belongs to her in the case of a leak

I am not convinced yet that this is a problem that needs solving at the cryptographic protocol level, I don't even see plausible deniability as a must-have feature for the use cases we have been discussing _(e.g. If the Krabby Patty secret formula leaked, it is bad regardless of whether or not there is a signature on it)_; I worry that it could lead to over-engineering.

danieldaquino:
> > I worry that it could lead to over-engineering.
> 
> Frankly, all of these ideas (this PR, #1029 and #1030) are over-engineering Nostr to please very specific use cases where that small delta between what we have today and the PR becomes significant. For everybody else, the gains are clearly way too small to justify the investment.

Most relays and clients don't need to implement this NIP and probably wouldn't want to. Most apps/relays today implement the public social network use case, and for that use case this NIP might not be in their interest to implement. That's ok.

This NIP would only appeal for new apps that want to serve more private use cases, such as internal company communications, private groups between friends/family, etc.

Not all clients/relays need to implement this. That's ok, it's an optional NIP.

> For instance, the delta of value between having a regular relay, where users can concentrate public posts while trusting clients to post to the right relay on their own, and this PR, which adds a little crypto blocker until everyone codes that into their relays, is extremely small. Users of this PR really need to value that blocker much more than the average Nostr user to justify this approach. Same story for #1029 and #1030.

If I want to run a private Nostr app for internal company comms or a small private group between friends, I would value a mechanism like this tremendously. If I tried to do that today with the apps/relays available, data could get accidentally leaked very easily. All it would take is for one member of the group to inadvertently add a connection to a public relay and accidentally post or rebroadcast notes to those relays.

The great thing about this NIP is that applications that want to work with this it will need to _explicitly_ handle these "to" fields.

It's similar to how type-safe programming languages prevent type errors by forcing the developer to explicitly think about and manage variable types.

> From what I see out there, there is an unrefined need for these things but no PR seems presents a "clear-dunk" solution to the need. They are all partially solving it with some side effects on each.
> 
> I think that's why most of us are still on the "I am not sure if this is good or not" mindset.

Whether a solution is clear-dunk depends on the specific use case as well. For private groups where there is some decent amount of mutual trust and connection (e.g. companies, family, friends) I think this would be perfectly suitable.

For other use cases, maybe not so much? but I think that's fine, they can create/use other NIPs

vitorpamplona:
Agree that this is optional. But that doesn't mean it's good. Having a bunch of optional crap no one uses is a problem. 

> For private groups where there is some decent amount of mutual trust and connection (e.g. companies, family, friends) I think this would be perfectly suitable.

Let's find groups that want to implement/use these ideas, then. My current assumption is that companies, family, friends do not want the same thing out of Nostr clusters. For instance, for the company I am running #1029 for internal communications, public verifiable events are a massive legal problem. Friends have more flexibility than family and families have more flexibility than companies to accept the way public notes are used. Private group notes will require a whole different approach. At this point, I will be surprised if a single solution works for them all.

jb55:
well by definition this is a new version/fork of the protocol. I think that's fine because v1 is not private to begin with and there's no way to make it so, at least with all the proposals I've seen. Leaking and spreading is just way too easy on nostr.

Having notes that are protocol-incompatible with public notes is a feature! but we also don't have to throw away all the nips for private-nostr, so I see this as a diverging fork for sure.

happy to close this and create a fork of the repo instead if this doesn't really jive here.

danieldaquino:
> For instance, for the company I am running https://github.com/nostr-protocol/nips/pull/1029 for internal communications, public verifiable events are a massive legal problem.

@vitorpamplona, can you please help me understand this point, perhaps with analogous examples (if you can't share the specifics)? I don't fully understand it yet.

When I think of the corporate use case, I imagine the corporation would want it to avoid leaking trade secrets or IP (e.g. Proprietary code). If that leaks, it is bad either way. An adversary who gets that info could use that proprietary trade secret (e.g. code) and run it, copy it, etc. regardless of whether there is a cryptographic signature on it. Similar with user data. If — for example — a customer email list leaked of out a private server, the consequences of potential spam/scam risk are still there regardless of whether that payload is cryptographically signed or not.

I am not denying that this could be an issue (I can see it probably would be in certain cases), I just want to learn more about it so that I can fully understand your point of view. Can you please elaborate on the use cases/scenarios where this would become a big issue, so I can better understand them?

jb55:
I just had another dumb yet simple idea:

You could have an unsigned/trusted mode where after you AUTH, all notes sent afterwards can be unsigned. The server at least knows the person sending notes on the socket has control of the key, but other clients trust that the server isn't forging anything.

Everything is deniable in that case, but clients would need to created an untrusted/unsigned DB separate from their trusted one, which adds some complexity.

jb55:
I'm starting to like this idea way more than changing the id commitment. It prevents broadcast and adds repudiation, at the risk of relay tampering which is kind of crappy, but might be ok in many trusted environments like corporate relays?

jb55:
I will create a separate NIP for this because the tradeoffs are way different

alltheseas:
> I just had another dumb yet simple idea:
> 
> You could have an unsigned/trusted mode where after you AUTH, all notes sent afterwards can be unsigned. The server at least knows the person sending notes on the socket has control of the key, but other clients trust that the server isn't forging anything.
> 
> Everything is deniable in that case, but clients would need to created an untrusted/unsigned DB separate from their trusted one, which adds some complexity.

Did you just recreate twitter?

jb55:
On Tue, Apr 23, 2024 at 06:47:27AM GMT, alltheseas wrote:
>> I just had another dumb yet simple idea:
>>
>> You could have an unsigned/trusted mode where after you AUTH, all notes sent afterwards can be unsigned. The server at least knows the person sending notes on the socket has control of the key, but other clients trust that the server isn't forging anything.
>>
>> Everything is deniable in that case, but clients would need to created an untrusted/unsigned DB separate from their trusted one, which adds some complexity.
>
>Did you just recreate twitter?

more like IRC :P

danieldaquino:
> It's a risk management problem and not necessarily a legal problem. And no one likes to take unnecessary risks. Especially when other solutions are available in the market.

@vitorpamplona makes sense, thanks for the in-depth explanation! I can see this being a problem especially for larger corporations that likely have more strict risk management.

Perhaps @jb55's suggestion would be a good one for that particular market, or something in between where signatures are stripped (or notes deleted) over time following some retention schedule.

But I admit I do not have a perfect solution for that particular concern, given that Nostr is built on signed events. 😅

vitorpamplona:
> Perhaps jb55's suggestion would be a good one for that particular market

I think these protections are not that useful to make internal comms for companies, but they might work well for communities and other activities where information leaks are not a big deal. We gotta find the right target for each approach.

jb55:
after thinking about it some more, giftwraps are probably better and safer unsigned events, even if the scheme is a bit more complex. with the nostrdb changes I am thinking of I think this complexity can be greatly mitigated for end-user applications.

jb55:
I don't like this anymore

Issue #541, opened 2023-05-19
Thou shall not use single letter tags


alexgleason:
At some point I saw a comment stating that single letter tags should ONLY be used if it's expected that relays will index them with filters like `"#x": []`. Give there are only 26 letters of the alphabet, this is a point I strongly agree with. So I think during NIP feedback we should be mindful of this. Thoughts?

staab:
I think we are, I don't think I've often seen single-letter tags proposed when indexing is not a concern.

telamon:
But can we have dual emoji please?

It's an very space efficient technique to encode category/topic in an nonoffensive manner.
#🇸🇪😎

PR #1147, opened 2024-03-30
NIP-01: filters can query any tag

alexgleason:
Filters such as `{ "#proxy": ["https://example.com/123"] }` are valid. It doesn't mean the relay will index them - this is never a guarantee (regardless of whether the tag is single-letter or not).

### DIFF
--- a/01.md
+++ b/01.md
@@ -124,7 +124,7 @@ Clients can send 3 types of messages, which must be JSON arrays, according to th
   "ids": <a list of event ids>,
   "authors": <a list of lowercase pubkeys, the pubkey of an event must be one of these>,
   "kinds": <a list of a kind numbers>,
-  "#<single-letter (a-zA-Z)>": <a list of tag values, for #e — a list of event ids, for #p — a list of pubkeys, etc.>,
+  "#<tag>": <a list of tag values, for #e — a list of event ids, for #p — a list of pubkeys, etc.>,
   "since": <an integer unix timestamp in seconds, events must be newer than this to pass>,
   "until": <an integer unix timestamp in seconds, events must be older than this to pass>,
   "limit": <maximum number of events relays SHOULD return in the initial query>


fiatjaf (approved):

arthurfranca:
It seems to encourage a bad practice. Why filter by an unindexed tag, specially when client can't rely on broad relay support?

alexgleason:
NIP-01 doesn't say which tags should or shouldn't be indexed by relays, or guarantee that relays need to return anything at all.

[`matchFilters`](https://github.com/nbd-wtf/nostr-tools/blob/master/filter.ts#L29) doesn't enforce any tag name.

It's not "encouraging a bad practice", it's just the truth.

arthurfranca:
> NIP-01 doesn't say which tags should or shouldn't be indexed by relays

From NIP-01 text: "As a convention, all single-letter (only english alphabet letters: a-z, A-Z) key tags **are expected to be indexed** by relays"

This is the main detail to remember when picking tag names for a new event spec.

alexgleason:
Yes, I was the one who pushed for that: https://github.com/nostr-protocol/nips/issues/541

The reason I opened this issue was because someone argued that `{ "#proxy": ["https://example.com/123"] }` is not considered a valid Nostr filter according to the NIP-01 spec. Do you think `matchFilters` should throw if you try to match against that filter?

Like you said, the text already clarifies that only single-letter tags should be expected to be indexed. This is about what's considered a valid filter.

arthurfranca:
Most relays will simply ignore the `#proxy` part and use the rest of the filter keys.

In the context of a personal relay it could filter by that custom key though so I guess it could be a valid check for `the matchFilters` fn.

But imo shouldn't be included on NIP-01.

alexgleason:
Look at my change again. I am not suggesting that people filter with multi-letter tags. I am replacing an unnecessary restriction (simplifying) with something that is already defined in other places in the document.

MerryOscar:
I'm not sure if this is relevant but thought I'd share my experience with trying to reference podcasts in nostr events - https://github.com/nostr-protocol/nips/issues/760

Although there is still no consensus on how to do this - the two options I came up with that seemed to fall within the existing nostr spec were either using `r` tags: `[“r”, “podcast:guid:123”]` or NIP-32 labels - `["l", "123", "podcast:guid"]`.

The reasoning behind this was so the tags would be indexed by relays - but it would be easier if the single tag filter requirement were relaxed in the way this PR proposes. Then the existing `proxy` tag `["proxy", "podcast:guid:123", "podcast"]` would work and not all relays would have to index the tags - only the ones that cared about podcasts.

mikedilger:
As a datapoint, chorus skips these when deserializing from json (if there is more than one letter after the hash character). Which probably isn't the right place to do it, but that's an internal problem to chorus.

So no problems here.

vitorpamplona:
+1

I kinda wish we had both `.tags` and `.indexes` and separate business logic from db-indexing labels. But it's too late for that.

alexgleason:
> update NIP-11 with what tags for what kinds are indexed?

That's a great idea @Semisol

alexgleason:
Closing because it's old and has merge conflicts, but it's still true.

PR #1148, opened 2024-03-30
Signed Filters

vitorpamplona:
The goal here is 2 fold: 
- Allow subscriptions for multiple users in the same connection. 
- Get rid of the auth round trip

We can use an `AUTHED_REQ` message type as well. 

I don't know if this is the right way to do it, but it is what I am testing at the moment.

### DIFF
--- a/63.md
+++ b/63.md
@@ -1,8 +1,8 @@
 NIP-63
 ======
 
-Pre-authed Filters
-------------------
+Signed Filters
+--------------
 
 `draft` `optional`
 
--- a/63.md
+++ b/63.md
@@ -8,7 +8,7 @@ Pre-authed Filters
 
 This NIP declares the use of event kind `33012` to represent one or more Nostr filters. 
 
-The `d` tag represents the subscription id and the `filter` tag describes a (NIP-01)[01.md] filter. 
+The `d` tag represents the subscription id and the `filter` tag contains the JSON-stringified (NIP-01)[01.md] filter. 
 
 Supporting relays interpret the tags of this event as if they were submitted in the `REQ` scheme while automatically authenticating the signer. 
 
--- a/63.md
+++ b/63.md
@@ -1,0 +1,33 @@
+NIP-63
+======
+
+Pre-authed Filters
+------------------
+
+`draft` `optional`
+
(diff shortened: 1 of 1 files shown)


fiatjaf:
I don't like this, I think `AUTH` achieves the same thing in a simpler and more flexible way.

> Allow subscriptions for multiple users in the same connection.

I don't know if this is a problem. You can always open multiple connections, and what are the use cases for having the same process listen to things on behalf of many users? A custodial Nostr server of some sort?

> Get rid of the user auth round trip since the user might/will not be online.

If a user is delegating listening to relays to a third party it is better to come up with a way to tell that relay that the third party will be allowed to subscribe to the user's events. Incidentally this also fixes the previous point as the third party can be allowed to subscribe to multiple users' events.

vitorpamplona:
> I think AUTH achieves the same thing in a simpler and more flexible way.

This is not a replacement for AUTH. It's a way to use AUTH by a third party. 

> I don't know if this is a problem.

This is a problem. Amethyst Push System has to authenticate for ~145,000 active users of the notification system. There is no way we can open 145K connections to each of the 3000 relays out there. 

> If a user is delegating listening to relays to a third party it is better to come up with a way to tell that relay that the third party will be allowed to subscribe to the user's events.

This is how we tell them.

fiatjaf:
> This is a problem. Amethyst Push System has to authenticate for ~145,000 active users of the notification system. There is no way we can open 145K connections to each of the 3000 relays out there.

It sounds like you're doing something very wrong here.

> This is how we tell them.

This is a very bad way of telling them. A single event published by the user saying something like `"consider AUTH from <amethyst-server-pubkey> as if it was mine"` is much better.

mikedilger:
I'm not against this proposal. It is a form of delegation, and being more specific about what is being delegated is a good thing IMHO so that as Vitor says, the middle-man can only do what the user authorized them to do.

> Amethyst Push System has to authenticate for ~145,000 active users of the notification system. There is no way we can open 145K connections to each of the 3000 relays out there.

It would be far easier for the phone to pull from a small set of inbox relays. But you do you. This isn't a nostr feature I will be using.

PR #1152, opened 2024-04-01
Add Corny Chat Slide Set and Link Set kinds

vicariousdrama:
Kind 30388 - Corny Chat Slide Set is a list that includes ordered references to image URLs and optionally a corresponding caption.  It most closely resembles an album of images.

Kind 31388 - Corny Chat Link Set is a list that includes ordered references to URIs intended for hyperlinking from a web browser and/or a corresponding caption.  It is similar to the existing kind 30003 for Bookmark Sets, but only intended to store URIs .

NIP 51 is updated to enumerate both and provide the expected tag items


staab:
These are very specific kinds. I wonder if they could be made more generic? Also, I'm curious how they would be linked to the room.

vicariousdrama:
They are generic enough for others to use for albums and links.  So they can be renamed if so desired.

They are _not_ linked to rooms, though Corny Chat's implementation does set the default value of the `d` tag to reference the room id. Users can change that though when setting the name of the list and about information.

fiatjaf:
I don't know if they should be made generic, but definitely having the client trademark in the kind description is not a good idea.

@vicariousdrama would you be willing to define what an audio-space-room is and how one can implement a client that is compatible with Corny Chat?

vicariousdrama:
It's probably easier if I just remove the "Corny Chat" terms from these. 

An audio space room is a long lived online area where people can meet to discuss topics in audio form. The room is customizable to the owner's preferences. This includes metadata about the room, optional images that can be displayed in the room by a presenter, hyperlinks for resources made available to room participants, and more.

Corny Chat allows users to export portions of a room configuration for later import (in the same room, or another room, or for other purposes). There are certain tags that Corny Chat will assign, as well as expect to be present in a Slide Set or Link Set for export/import.  Based on NIP32, Corny Chat sets labels denoting the set as "Corny Chat Slide Set" or "Corny Chat Link Set" along with the corresponding kind.  If these aren't present, they are passed over for imports.

fiatjaf:
I'm not against merging these, but I think it's impossible for anyone to implement a Corny-Chat-compatible client today without reverse-engineering Corny Chat, so I don't see the point of standardizing just these lists and not the rest.

vicariousdrama:
I'll be adding documentation to the Corny Chat repository that will cover the nostr integrations. There's no reason to duplicate all that in the NIPS repo itself.  I'm not sure what level of detail should be here in the NIPs

fiatjaf:
Once you have those things written then please link to them from the list of kinds in the README, like nostrocket and lightning.pub have done. No NIP, just a specification of the kinds you're using.

vicariousdrama:
Thanks @fiatjaf.  Should the kind numbers be retained in NIP-51 as well, or removed?

fiatjaf:
I think removed makes more sense, but I'm not sure.

vicariousdrama:
Bump

AsaiToshiya (approved):
But I prefer to remove them that are no consensus as NIP from NIP-51.

51.md
```diff
@@ -54,7 +54,9 @@ Aside from their main identifier, the `"d"` tag, sets can optionally have a `"ti
 | Kind mute sets | 30007 | mute pubkeys by kinds<br>`"d"` tag MUST be the kind string                                  | `"p"` (pubkeys)                                                                   |
 | Interest sets | 30015 | interest topics represented by a bunch of "hashtags"                                         | `"t"` (hashtags)                                                                  |
 | Emoji sets    | 30030 | categorized emoji groups                                                                     | `"emoji"` (see [NIP-30](30.md))                                                   |
+| Slide Set | 30388 | user-defined ordered list of captioned image urls | `"r"` (image URLs) |
 | Release artifact sets | 30063 | groups of files of a software release  | `"e"` (kind:1063 [file metadata](94.md) events), `"i"` (application identifier, typically reverse domain notation), `"version"`  |
+| Link Set | 31388 | user-defined ordered list of captioned hyperlinks | `"r"` (web page URLs) |
```
fiatjaf, line 59:
```suggestion
```

51.md
```diff
@@ -54,7 +54,9 @@ Aside from their main identifier, the `"d"` tag, sets can optionally have a `"ti
 | Kind mute sets | 30007 | mute pubkeys by kinds<br>`"d"` tag MUST be the kind string                                  | `"p"` (pubkeys)                                                                   |
 | Interest sets | 30015 | interest topics represented by a bunch of "hashtags"                                         | `"t"` (hashtags)                                                                  |
 | Emoji sets    | 30030 | categorized emoji groups                                                                     | `"emoji"` (see [NIP-30](30.md))                                                   |
+| Slide Set | 30388 | user-defined ordered list of captioned image urls | `"r"` (image URLs) |
```
fiatjaf, line 57:
```suggestion
```

Issue #1154, opened 2024-04-01
Type of NIP-46 message has changed


alexgleason:
- Original document: https://github.com/nostr-protocol/nips/blob/4bb393735ea501f0bac3e2cb255ae9a3caab6a68/46.md
- New version: https://github.com/nostr-protocol/nips/blob/master/46.md

New version says `[<json_stringified_event_to_sign>]`

Old version just says `[event]`

Did we change this on purpose? Or it slipped through the cracks, or what?

alexgleason:
I'm implementing a new Nostr Connect thing and it's incompatible with my old thing.

alexgleason:
It was changed in https://github.com/nostr-protocol/nips/pull/1047 @erskingardner 

I think this was just overlooked, right? We should change it back.

fiatjaf:
It was always implemented as the stringified garbage.

erskingardner:
Yeah, I just tried to make the original doc more clear. Anything you put in an even must be stringified to be valid JSON.

PR #1155, opened 2024-04-01
NIP-46: restore original types

alexgleason:
In https://github.com/nostr-protocol/nips/pull/1047 we agreed to leave the protocol the same, and just clarify its usage. However a change slipped through, which I think wasn't intentional, because nobody building existing NIP-46 apps has updated AFAIK, and I don't think they knew they were supposed to.

Fixes https://github.com/nostr-protocol/nips/issues/1154

### DIFF
--- a/46.md
+++ b/46.md
@@ -59,11 +59,11 @@ nostrconnect://<local-keypair-pubkey>?relay=<wss://relay-to-connect-on>&metadata
     "content": nip04({
         "id": <random_string>,
         "method": "sign_event",
-        "params": [json_stringified(<{
+        "params": [{
             content: "Hello, I'm signing remotely",
             pubkey: "fa984bd7dbb282f07e16e7ae87b26a2a7b9b90b7246a44771f0cf5ae58018f52",
             // ...the rest of the event data
-        }>)]
+        }]
     }),
     "tags": [["p", "fa984bd7dbb282f07e16e7ae87b26a2a7b9b90b7246a44771f0cf5ae58018f52"]], // p-tags the remote user pubkey
 }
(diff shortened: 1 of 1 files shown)


alexgleason:
This NIP is such a shit show. Now it's possible this revision could break an implementation developed in the past 40 days by restoring the behavior from 1 year ago.

alexgleason:
cc @erskingardner @nostrband 

Are yall stringifying these params or not?

erskingardner:
There weren't any changes, I only made things in the NIP more explicit. You can't put anything into an event that isn't stringified if you want it to be valid JSON. If you weren't doing it explicitly, then whatever language/framework/library you're using was probably doing it for you.

erskingardner:
ahhh, ok. I see what you mean now. yeah, my interpretation of how I'd written it was not as two steps but you're right, that is confusing and doing it twice does change things.

erskingardner (approved):

nostrband:
All params are strings, putting an event to signEvent without stringify is a breaking change. Result is a string, removing stringify from there is a breaking change too. I don't like it, but that's how it's implemented everywhere.

alexgleason:
@fiatjaf Why did you close this? I'm confused.

Is it true people have been stringifying against the spec all along? I don't believe it.

fiatjaf:
Unfortunately it is. I assumed it wasn't because the spec didn't say so, but when I went to look in order to implement it -- much before the NIP-46 rewrite -- it was implemented like that in ndk, nsecbunker and snort, and that's how I implemented it in nostr-tools and go-nostr and these things all interoperated correctly.

alexgleason:
You're right, I just saw: https://github.com/nostr-dev-kit/ndk/blob/1fdc61736f33cc9ebc822e8bb265d5d10ccf1c5a/ndk/src/signers/nip46/rpc.ts#L14 

Alright. 🤷

PR #1157, opened 2024-04-02
add kind 10033 private follows list

haorendashu:
Add kind 10033 private follows list who shoud be encrypted and put in to the event content. When querying feeds, we should mix these private follows list with the kind 3's list.

The things we put into private follows list's what we want to follow and don't want others know.

### DIFF
--- a/51.md
+++ b/51.md
@@ -20,18 +20,19 @@ Standard lists use non-parameterized replaceable events, meaning users may only 
 
 For example, _mute list_ can contain the public keys of spammers and bad actors users don't want to see in their feeds or receive annoying notifications from.
 
-| name           | kind  | description                                                 | expected tag items                                                                |
-| ---            | ---   | ---                                                         | ---                                                                               |
-| Mute list      | 10000 | things the user doesn't want to see in their feeds          | `"p"` (pubkeys), `"t"` (hashtags), `"word"` (lowercase string), `"e"` (threads)   |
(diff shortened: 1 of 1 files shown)


mikedilger:
This PR makes sense, but there is another possibility.

Gossip has been using kind 30000 FollowSets to allow people to define multiple sets of people to follow. Each of these includes private entries.

Since then I've been thinking about the kind-3 follow list as the publicly declared list of people followed, it's purpose being to inform others about who I follow (so they can compute web of trust, etc), rather than to provide a local feed (even though I do allow people to see a local feed from it).  Given that distinction, at least for gossip client, kind-3 doesn't really need private entries.

However, most clients don't implement/use kind 30000 FollowSets and so if the just want private entries (without the multiple lists feature), this PR makes sense.

vitorpamplona:
+1 for 30000 FollowSets. You can build several lists on Listr.lol and select them on Amethyst. 

A single follow list for everything doesn't really work. Zap.stream follows are different than Kind 1 follows, which are different than audio/podcast follows. A user that uses multiple apps has a lot of trouble establishing follow lists for separate event kinds. 

Also, kind 3 WAS just "contacts" and not follows. That made more sense that the current language.

haorendashu:
It's nice to know that kind 30000 had bean used by clients. I would like to use kind 30000 too.

PR #1163, opened 2024-04-04
NIP-29: Pin Messages

sepehr-safari:
Admins may pin some of the messages to the top of the group.


sepehr-safari:
> The pinned messages are supposed to be private to the group

No, they are not.
assume that there are lots of messages in a group (each has an event id) and the owner decided to pin one of those messages to the top of the group (imagine it like telegram pinned messages in groups). so the owner simply publishes a moderation action and reference that specific event to be pinned.

sepehr-safari:
@fiatjaf thoughts?

fiatjaf:
Looks good to me. @pablof7z was also interested in this.

fiatjaf (approved):

pablof7z:
Would be better to have an update-pin-list and have a payload of the entire pin list event

Say you want to reorder the list; doing with add/remove would be super annoying. Or empty the list.

fiatjaf:
That's true.

sepehr-safari:
> Would be better to have an update-pin-list and have a payload of the entire pin list event

updated.

PR #1166, opened 2024-04-09
Nip15

gzuuus:
Revision of nip 15:
- Modification of some elements and tags
- Introduction of variable products
- Clean up
- Rewrite some parts to make them easier and clearer to understand

[view](https://github.com/gzuuus/nips/blob/nip15/15.md)


fiatjaf:
@motorina0 @talvasconcelos @arcbtc

15.md
```diff
@@ -79,7 +79,7 @@ Fields that are not self-explanatory:
   ...
 }
 ```
- - the `d` tag is required, its value MUST be the same as the stall `id`.
+ - the `d` tag is required, its value MUST be the same as the stall `id`. (NOTE: What is the reason for having the same value of `id` in the content, and `d` tag? Can we get rid of the `id` and just have the `d` tag?)
```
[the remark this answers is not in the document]
gzuuus: Let's do it, I have not been able to find a reason to have them duplicated.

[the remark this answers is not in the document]
gzuuus: I also think that `stall_id` can be replaced by an tag:`a` so that we can index the products in a stall.

motorina0:
cc: @ibz

15.md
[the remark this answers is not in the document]
motorina0: > Let's do it, I have not been able to find a reason to have them duplicated.

Not all redundancy is bad. 
The event content will be used by the actual app and it needs those `id`s. Yes, you can grab the event tags and alter the event content, but that just makes the merchant&client code more complicated.
Its a separation of concerns, tags are `nostr` protocol related, content is application logic related.

15.md
```diff
@@ -75,21 +74,21 @@ Fields that are not self-explanatory:
...
 
 **Event Content**
 
 ```json
 {
-  "id": <string, id generated by the merchant (sequential ids are discouraged)>,
-  "stall_id": <string, id of the stall to which this product belong to>,
+  // "stall_id": <string, id of the stall to which this product belong to>,
   "name": <string, product name>,
+  "type": <simple | variable | variation>,
```
motorina0, line 91:
```suggestion
  "type": <string, ["simple" | "variable" | "variation"]>,
```

15.md
```diff
@@ -223,11 +227,19 @@ The below JSON goes in `content` of [NIP-04](https://github.com/nostr-protocol/n
 {
   "id": <string, id of the order>,
   "type": 2,
+  "status":<String, "pending" | "processing" | "completed" | "cancelled" | "refunded" | "failed">;,
```
motorina0, line 230:
```suggestion
  "status":<string, "pending" | "processing" | "completed" | "cancelled" | "refunded" | "failed">;,
```

15.md
```diff
@@ -121,21 +120,26 @@ Fields that are not self-explanatory:
    - to calculate the total cost of shipping for an order, the user will choose a shipping option during checkout, and then the client must consider this costs:
      - the `base cost from the stall` for the chosen shipping option
      - the result of multiplying the product units by the `shipping costs specified in the product`, if any.
+- `type`: It determines whether a product is a simple product, a variable product, or a variation of a variable product. (i.e.: a shirt sold in a single colour is a simple product, a shirt sold in more than one colour is a variable product, and has variations). To link the variable product within its variations, the variation product event should contain an `a` tag pointing to the event coordinate of the variable product.
```
motorina0, line 123: 1. how is the merchant keeping the inventory. If he has 3 red shirts and 2 blue ones then the client will see one product with 5 or two different products
1. how does the merchant (or anyone) decide that the blue shirt is a variation of the red one or vice-versa? What does it happen if there are no more red shirts (the "simple" product)?

motorina0: Isn't this request just a "group by" functionality? 
So web clients can display:
 - in the `product list` multiple products as one
 - in the `product details` the different grouped products that the user can select from

15.md
```diff
@@ -75,21 +74,21 @@ Fields that are not self-explanatory:
...
+ - the `d` tag is required and represents the stall id.
 
 ### Event `30018`: Create or update a product
 
 **Event Content**
 
 ```json
 {
-  "id": <string, id generated by the merchant (sequential ids are discouraged)>,
-  "stall_id": <string, id of the stall to which this product belong to>,
```
motorina0, line 91: should not be removed, see my https://github.com/nostr-protocol/nips/pull/1166#discussion_r1559095810

15.md
gzuuus: Products that are variations of a variable product have a tag:`a` in their tags pointing to the product of which they are a variation. So if a person sees the variable product in a stall, this being a variable product means that it has some variations so the client should query for events that point to the variable product, using a filter. Ideally once the user selects the product to go browse its product page, although this is up to the developers of each client and how they want to treat variable products. 

The quantity of each variation, as well as the price is specified in the event of each variation, so you can have 5 red shirts and 2 purple shirts and the purple one is a bit more expensive. 

You can also use the `"specs"` field in the variable product to give a preview of the different variations that can be found.

15.md
gzuuus: Just [answered](https://github.com/nostr-protocol/nips/pull/1166#discussion_r1559206125)

15.md
motorina0: - how is this different than a `group by`?
- the "variations" approach is much more tangled

15.md
motorina0: This is a nitpick improvement unrelated with the other changes in this PR.
Its a NACK from my side.

15.md
```diff
@@ -47,7 +47,6 @@ A merchant can publish these events:
 
 ```json
 {
-  "id": <string, id generated by the merchant. Sequential IDs (`0`, `1`, `2`...) are discouraged>,
   "name": <string, stall name>,
   "description": <string (optional), stall description>,
   "currency": <string, currency used>,
```
Quin5, line 52: Why is there a single currency for the stall? The merchant should be able to specify multiple currencies that they are willing to accept.

motorina0: The `currency` can be `BTC` or a fiat `currency`.
The payment only takes place using BTC.  **There is no payment in fiat**.

The fiat currency serves as an indicator, it will be converted to `sats` at the moment the invoice is created.

Having the merchant set **multiple currencies** (lets say Turkish Lira and USD)would:
 - require the merchant to keep track of the exchange rate between these currencies and update the prices accordingly
 - allow the clients to select the fiat currency that has the best exchange rate against BTC, thus paying less

However, on the client side (not merchant side), the application can take one fiat currency (lets say USD) and show the user the value in any currency it requires (eg: Turkish Lira)

Quin5: An indicator of what?

How would the value of this currency be determined without the merchant being required to keep track of it?

ibz: I don't understand how this is even a question. You have one currency that the seller is willing to accept: sats and one fiat currency that the seller is willing to have the price fixed in. The clients can display anything else, but that is a UI issue, not a protocol issue. You can't have more than one fiat currency here, because fiat currencies fluctuate against each other, so it would be impossible to make the conversion to sats at the time of sale!

motorina0: > Why is there a single currency for the stall? The merchant should be able to specify multiple currencies that they are willing to accept.

The merchant does not only specify the currency, it also has to specify the actual value.
Eg for today:
 - 10 USD
 - 300 Turkish Lira
 
 What will be the prices tomorrow if the Turkish Lira depreciates by 25%?

Quin5: > I don't understand how this is even a question. You have one currency that the seller is willing to accept: sats and one fiat currency that the seller is willing to have the price fixed in. The clients can display anything else, but that is a UI issue, not a protocol issue. You can't have more than one fiat currency here, because fiat currencies fluctuate against each other, so it would be impossible to make the conversion to sats at the time of sale!

Thank you, it makes more sense now.  If I understand correctly, "currency" is an indicator of BTC value in the merchant's preferred fiat currency.  Basically a, "merchant's price display preference", yes?

Quin5: > > Basically a, "merchant's price display preference", yes?
> 
> Not really. It is the amount of fiat currency that the merchant is willing to sell the item for. As a merchant you can say you want to sell the T-shirt for 20 USD or for 30k IDR... and it will always sell for that amount regardless of the USD <=> IDR conversion. This is not about price display. It is about _price_. Price display can be anything, up to the UI implementation.

Is there a portion of this NIP that explains how the BTC amount is updated to stay at the fiat value (which I'm assuming would be in the Product definition); or is this something each marketplace client will determine on their own?

motorina0: > Is there a portion of this NIP that explains how the BTC amount is updated to stay at the fiat value

No, it is not necessary for that to be part of the protocol.

Please experiment with one of the implementations (plebian-market or the LNbits nostrmarket extension), it will make things clearer.

ibz: The BTC amount is not updated at any point... it is simply the fiat amount converted to BTC at the time the buyer tries to buy something.

Quin5: The "converted" part is what I'm asking about.  I don't see that mentioned in this NIP.  Is there an assumed mechanism?

ibz: NIPs define the communication protocol between clients. Currency conversions are not part of the protocol. The seller could be manually opening coinbase.com and looking up the price. It really does not matter.

15.md
gzuuus: How can you implement 'group by' here?
I think the design is extensible, flexible and simple enough to bring in the feature of variable products and their variations.

15.md
gzuuus: Yes, as others have commented, prices are fixed in a currency and converted to the market value of BTC at the moment when a user wants to buy the product. I also think that from a protocol perspective, there is no need to specify how to convert between currency/BTC, this is something specific and the responsibility of the platform

15.md
gzuuus: I would say that there is no necessity to have that redundant information in the events, there is no separation of concerns or something around that, same information duplicated two times in the same event, from my point of view it just creates ambiguity that can lead to inconsistencies

15.md
```diff
@@ -346,24 +339,28 @@ Bids are simply events of kind `1021` with a `content` field containing the stri
...
 
-An additional `message` field can appear in the `content` JSON to give further context as of why a bid is `rejected` or `pending`.
+An optional `message` field can appear in the `content` JSON to provide further context for why a bid is `rejected` or `pending`.
 
-Another thing that can happen is - if bids happen very close to the end date of the auction - for the merchant to decide to extend the auction duration for a few more minutes. This is done by passing a `duration_extended` field as part of a bid confirmation, which would contain a number of seconds by which the initial duration is extended. So the actual end date of an auction is always `start_date + duration + (SUM(c.duration_extended) FOR c in all confirmations`.
+If bids are placed very close to the auction's end date, the merchant can extend the auction duration by updating the `duration` field in the auction event.
```
ibz, line 363: Auctions cannot be edited after they have received any bids, so how would this work?

gzuuus: There is nothing that stop someone to edit auctions after the first bid, also I'm purposing to do bids differently, instead of using a `e` tag what it is inappropriate to reference p r events (since relays are not supposed to store versions of the p r events) use a 'snapshot' of the auction as stringified auction event in the content and set the amount in a tag `amount` as there is already a standard `amount` tag

ibz: In the existing NIP, auctions cannot be edited after they get bids. Period. If you edit an auction, it becomes a new product, which has no bids. All bids are attached to a particular version of the auction *by design*.

What you are proposing is to make all clients need to validate this assumption rather than having it simply enforced as part of the protocol. Furthermore, the design gets even more complicated since not only every client needs to verify that the auction did not change, but they should exclude certain fields - like the end time - from the said validation.

ibz: > relays are not supposed to store versions

There is only one version, because auctions are not supposed to be edited after they receive bids!

gzuuus: When editing an auction, whether a bid has been placed or not, it doesn't convert into a new product... This is a parametrizable replaceable event, so forcing clients to behave in a certain way is a solution that raises many potential edge cases. You can't force a user not to change the event, even if clients doesn't allow it, since they can always sign a new version of the same event using their private keys. How do you handle this? You can't. 

What I propose is to use the correct coordinates to point to the auction, using a tag 'a' and storing a stringified version of the current auction event in the bid content (this event is tamper-evident, as it can always be verified that the auctioneer signed the event correctly). This way, the auctioneer can't change anything, and bidders have more security about what they're bidding on. It's also a more robust way to keep track of the auction version being bid on, and use this to put bids on pending or for any other reason. Validating is better than forcing, as forcing is impossible on Nostr, if the user possesses their key pair. The design becomes more robust, not more complicated, by using this snapshot and tag 'a' instead of a tag 'e'. All clients can validate whether a bid is valid or not.

ibz: >  solution that raises many potential edge cases

Like what? The current protocol and implementation is specifically designed to avoid edge cases because the auction is simply impossible to edit. If you edit it, it becomes a new object with no bids attached. Remember, the event ID is derived from all the event members, so any change to the event means the ID gets changed. This serves as a proof that the bid is attached to a specific version of the auction.

The new proposal seems to introduce a lot of potential issues while solving no real problem.

But I wish other reviewers would also comment, especially since I am not the person deciding whether to merge something or not.

PR #1167, opened 2024-04-09
NIP-46 : Fix use of confusing ambiguous term

kuiperanon:
It's very confusing as to whether `<remote-pubkey>` refers to remote user pubkey vs remote signer pubkey. This is complicated further by the typo in the explanation of "remote signer pubkey", which makes it hard to infer which would be used. Can we just be explicit here?

### DIFF
--- a/46.md
+++ b/46.md
@@ -25,7 +25,7 @@ This is most common in a situation where you have your own nsecbunker or other t
 The remote signer would provide a connection token in the form:
 
 ```
-bunker://<remote-pubkey>?relay=<wss://relay-to-connect-on>&relay=<wss://another-relay-to-connect-on>&secret=<optional-secret-value>
+bunker://<remote-user-pubkey>?relay=<wss://relay-to-connect-on>&relay=<wss://another-relay-to-connect-on>&secret=<optional-secret-value>
 ```
 
 This token is pasted into the client by the user and the client then uses the details to connect to the remote signer via the specified relay(s).


mikedilger:
+1

fiatjaf:
Wait, I merged this, but now I'm confused about what should actually be the pubkey there.

pablof7z:
it's the remote user pubkey

mikedilger:
Hang on. If it is not the remote-SIGNER-pubkey, how does the client know which signer to 'p' tag in their connect?  They specify the remote-USER-pubkey in the connect command.  So I think this is written wrong.

kuiperanon:
> Hang on. If it is not the remote-SIGNER-pubkey, how does the client know which signer to 'p' tag in their connect? They specify the remote-USER-pubkey in the connect command. So I think this is written wrong.

I would like confirmation. But I think the "remote signer pubkey" is only used for 24133 kind events that are sent directly to the remote signer such as for creating an account:

> Remote signers might support additional commands when communicating directly with it. These commands follow the same flow as noted above, the only difference is that when the client sends a request event, the p-tag is the pubkey of the remote signer itself and the content payload is encrypted to the same remote signer pubkey.

kuiperanon:
> Hang on. If it is not the remote-SIGNER-pubkey, how does the client know which signer to 'p' tag in their connect? They specify the remote-USER-pubkey in the connect command. So I think this is written wrong.

Yeah, I'm not sure why we need to put the remote-user-pubkey in the connect command parameters, since it's already in the p tag. Maybe it's a reason that's going over my head.

PR #1168, opened 2024-04-09
Relational Databases

vitorpamplona:
This NIP offers a plug-and-play method to export/sync information from existing systems using relational databases into Nostr. It describes a generic `ROW` event type to be used by all tables and all database types. 

Read [here](https://github.com/vitorpamplona/nips/blob/relational-database/41.md)

### DIFF
--- a/41.md
+++ b/41.md
@@ -1,0 +1,37 @@
+NIP-41
+======
+
+Relational Database
+-------------------
+
+`draft` `optional`
+
+This NIP offers a plug-and-play method to export/sync information from existing systems using relational databases into Nostr. 
+
+Event kind `31200` represents a row in a database. Each value of the row is added as a tag whose name is the column's name. 
+
+Values can be public or encrypted to the user. 
+
+```js
+{
+  "kind": 31200,
+  "tags": [
+    ["d", "<row-private-key>"]
+    ["n", "<table-name>"]
+    ["u", "<database-name>"]
+    ["<column-name>", "<column-row-value>"]
+    // other public columns
+  ],
+  "content": nip44Encrypt(JSON.stringify((
+    [
+      ["<column-name>", "<column-row-value>"]
+      // other private columns
+    ]
+  )),
+  // .. other fields
+}
+```
+
+Inserts and updates are supported via Nostr's regular signing operations and deletions via [NIP-09](09.md).
+
+The pubkey MAY be own by the system and represent several users from that system. 


vicariousdrama:
Interesting. How about handling datatypes, sizes, null values, etc.

Is this just intended as an export for import? If so, why JSON instead of either a) a binary blob, or b) existing tools for data export as text

vitorpamplona:
> How about handling datatypes, sizes, null values, etc.

Everything has to become a string :( 

> Is this just intended as an export for import?

vitorpamplona:
Right now, I am just doing a simple DB Trigger to push changes from 5 tables into Nostr. Then using a tool I developed in nostr to generate a report based on that data. It's obviously not the most performant thing ever, but it seems to work well for small to medium data sets.

itawa-panema:
I don’t know if it is a good ID but I would use the concatenation of "database name" + "table name" + "row id" for the “d” tag to ensure uniqueness.

vitorpamplona:
> "database name" + "table name" + "row id"

oh yeah! good catch!

vicariousdrama:
I also agree that the d tag should be a composite key. In many databases, primary keys end up as being an incrementing number if not a GUID.

vitorpamplona:
Yeah, my case was using UUIDs for PKs, so I didn't even notice that issue. :)

vicariousdrama:
Is this export row by row? 
Or multiple rows in a table?
Or multiple tables in a database?

vitorpamplona:
> Is this export row by row?

Right now it is a trigger in Postgres that runs individually for every insert, update or delete of each row. 

For those wondering, I did consider creating one event kind for each table and turn all foreign keys into a-Tags to make it more "Nostr native". But it is hard to justify using new event types if there is no expectation of interoperability with anything else. The tables were designed for that application alone.

staab:
But why?

vitorpamplona:
> But why?

Well, I need to bring an existing system into Nostr without breaking the production code. So, this is my first step. The next part is to replace the db with Nostr events entirely. Then phase the server out completely and keep only the JS client connecting directly to Nostr.

staab:
This is just a way to use nostr relays as a database. The schema is a proprietary format, so I don't see what defining how to fit the schema into nostr accomplishes. You're not going to get any kind of interoperability from this, which is the point of NIPs.

vitorpamplona:
> You're not going to get any kind of interoperability from this

Yep, this is just like [NIP-78](78.md), but for generic DBs instead.

staab:
So just use NIP 78 for your proprietary schema events

vitorpamplona:
> So just use NIP 78 for your proprietary schema events

I could use the same kind for sure, but since events from this PR can be massive, I thought a new kind could allow relays to block these kinds of objects without affecting kind `30078` applications, which seem to be much leaner than the ones from this PR.

mikedilger:
My first instinct is in favor of this, of a new kind, and of such a mapping.

Issue #1118, opened 2024-03-10
REMOVE command


mikedilger:
I wish to delete events on my relay. Maybe because they are spam. After the relay AUTHs me, it will accept DELETE commands which simply pass event IDs.

These are not delete events!  They do not propagate to other relays. Clients never see these.  It is just a command to delete the local event from the specific relay.

But we do not have a DELETE command in nostr.

Do people think this command is a good idea, or is it too much out of scope for nostr? Most clients probably would never issue DELETE commands. It could well be done in some out-of-band way too.

alexgleason:
The way I'm doing it is the relay can have admin pubkeys configured, and then admins can do normal kind 5 deletions of events they don't own.

vitorpamplona:
I like the idea that Clients should be able to manage personal relays via Nostr-native interfaces (via events, either in a RPCs protocol like NIP-46 or by modifying a replaceable event the relay is watching, etc). That could be a DVM-like system running in or outside the relay that accepts requests to adjust the relay's properties from a given set of admin pubkeys. A Delete method could just be one of the available commands to Clients.

 It would be awesome if all relay devs could agree on a set of methods for personal relays.

mikedilger:
Having thought a bit more, I think it should be called REMOVE to avoid confusion.

I would *love* if someone built a web-based relay moderation system, that popped up alerts for 1984 events, showed a moderation queue, did REMOVE commands, etc... and such a tool being 100% independent of the relay implementation.

alexgleason:
Check out how I'm doing relay moderation here: https://docs.soapbox.pub/ditto/nsec/#event-deletions

The relay has its own nsec and publishes events from it. Kind 5, 3, 10000, 10002, 10006, etc all have a semantic meaning when published from the relay nsec.

vitorpamplona:
> I would love if someone built a web-based relay moderation system, that popped up alerts for 1984 events, showed a moderation queue, did REMOVE commands, etc... and such a tool being 100% independent of the relay implementation.

If you make Chorus notify and reply, I will code the client on Amethyst.

mikedilger:
> Check out how I'm doing relay moderation here: https://docs.soapbox.pub/ditto/nsec/#event-deletions
> 
> The relay has its own nsec and publishes events from it. Kind 5, 3, 10000, 10002, 10006, etc all have a semantic meaning when published from the relay nsec.

That is one way to do it, but not my preference.

[...]

PR #1169, opened 2024-04-10
REMOVE command

mikedilger:
Proposal for a REMOVE command, so that moderation tools can be relay-implementation agnostic.

Prior discussion: https://github.com/nostr-protocol/nips/issues/1118

### DIFF
--- a/XX.md
+++ b/XX.md
@@ -1,0 +1,32 @@
+NIP-XX
+======
+
+REMOVE command
+--------------
+
+`draft` `optional`
+
+This NIP defines a new nostr command `REMOVE` of the following format:
+
+```json
+["REMOVE", <remove-request-id>, <filters1>, <filters2>, ...]
+```
+
+to be sent by a client to a relay.
+
+The relay should respond with a `REMOVED` message in the following format:
+
+```json
+["REMOVED", <remove-request-id>, <true|false>, <message>]
+```
+
+If the client is AUTHenticated and authorized to remove the specified events from the relay, the relay should remove them and respond with true.
+
+Relays not supporting this NIP will not comprehend the `REMOVE` message and will not send back a `REMOVED` response. Therefore clients supporting this NIP must tolerate situations where no response is obtained.
+
+This NIP defines no rules for the format of the `REMOVED` message field.
+
+Motivation
+----------
+
+It would be beneficial to the nostr ecosystem to have a generic means of moderating events on relays, because it allows for the creation of moderation tools that work on all supporting nostr relays, rather than just on a particular relay implementation.
--- a/XX.md
+++ b/XX.md
@@ -26,6 +26,8 @@ Relays not supporting this NIP will not comprehend the `REMOVE` message and will
 
 This NIP defines no rules for the format of the `REMOVED` message field.
 
+Relays that support this NIP should signal this in their NIP-11 relay information document.
+
 Motivation
 ----------
 
--- a/XX.md
+++ b/XX.md
@@ -30,6 +30,8 @@ This NIP defines no rules for the format of the `REMOVED` message field.
 
 Relays that support this NIP should signal this in their NIP-11 relay information document.
 
+NIP-42 AUTH SHOULD be used to authenticate the client user. The means of determining if a client user is authorized to remove events is outside the scope of this NIP.
+
 Motivation
 ----------
 
--- a/XX.md
+++ b/XX.md
@@ -22,6 +22,8 @@ The relay should respond with a `REMOVED` message in the following format:
 
 If the client is AUTHenticated and authorized to remove the specified events from the relay, the relay should remove them and respond with true.
 
+If the relay does not remove all the matching events for some reason, it should ideally remove none of them and send `false` in its `REMOVED` response.
+
 Relays not supporting this NIP will not comprehend the `REMOVE` message and will not send back a `REMOVED` response. Therefore clients supporting this NIP must tolerate situations where no response is obtained.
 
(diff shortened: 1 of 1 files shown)


pablof7z:
why introduce a new command? why not just a kind that implies a relay-admin's action? we'll definitely want more relay admin-commands, introducing a new verb for each seems unnecessary

e.g. kind: xxx, e-tagging event id 1,2,3 and 4 deletes them from the db when sent by an authorized pubkey.

alexgleason:
> why introduce a new command? why not just a kind that implies a relay-admin's action? we'll definitely want more relay admin-commands, introducing a new verb for each seems unnecessary

I agree with this. Check my solution here: https://docs.soapbox.pub/ditto/nsec/#event-deletions

pablof7z:
yeah, I was going to propose using kind 5 actually and then thought there could be push back on overloading the functionality of kind 5, but I think this makes sense

mikedilger:
It feels to me like "I have a hammer therefore everything is a nail". We have an event structure, therefore everything must be an event.

Little about this activity requires what an event gives:  persistence, sharing with others, time-stamping, signing, etc.... except maybe signing and only if AUTH isn't used.

Everything has to fit into an event-shaped hole.  And to me, the proliferation of unnecessary events is clutter.

But yes, if protocol extension is not possible, then this can be done via events.  Just not my preference.

mikedilger:
But if we are going to do this with an event, let's make it an ephemeral event not a kind 5 which propagates and tells others on other relays about a deletion which in this case would be invalid (signed by the wrong author).

vitorpamplona:
Many relays in production today already have keys to control themselves (sending invoices via DM, publishing a signed DNS record, etc). Alex's idea would fit in nicely. We can make a NIP-46/47-like protocol to control the relay via events. 

And because these events are public, people can check what the relay owner is deleting/censoring by just querying the key.

alexgleason:
> It feels to me like "I have a hammer therefore everything is a nail". We have an event structure, therefore everything must be an event.

Look at kind 5 events as they are. It's an event the user submits to the relay asking the relay to delete their event. It's not a different verb. The only difference between a user deleting an event and an admin deleting an event is who is doing it. It's simpler and more consistent to use a pattern similar to what already exists.

> And because these events are public, people can check what the relay owner is deleting/censoring by just querying the key.

Agree with this. Admin deletions should be transparent. Then it can easily be viewed in a relay-admin UI. Win-win.

There is really no harm overloading kind 5's either. Queries for kind 5's will always contain either ids or authors.

pablof7z:
> > It feels to me like "I have a hammer therefore everything is a nail". We have an event structure, therefore everything must be an event.
> 
> Look at kind 5 events as they are. It's an event the user submits to the relay asking the relay to delete their event. It's not a different verb. The only difference between a user deleting an event and an admin deleting an event is who is doing it. It's simpler and more consistent to use a pattern similar to what already exists.
> 
> > And because these events are public, people can check what the relay owner is deleting/censoring by just querying the key.
> 
> Agree with this. Admin deletions should be transparent. Then it can easily be viewed in a relay-admin UI. Win-win.
> 
> There is really no harm overloading kind 5's either. Queries for kind 5's will always contain either ids or authors.

Agreed with everything @alexgleason said here.

mikedilger:
> Look at kind 5 events as they are. It's an event the user submits to the relay asking the relay to delete their event. It's not a different verb. The only difference between a user deleting an event and an admin deleting an event is who is doing it. It's simpler and more consistent to use a pattern similar to what already exists.

This isn't true at all.  A kind-5 event is a proper event, spread between all parties on the network, to remove an event from everywhere, signed by the original event author authorizing such an action.

A relay admin cannot delete other people's events with a kind-5 event, and such an event has no force in any other jurisdiction, such an event clutters things, and such an event being persistent and copied to other jurisdictions appears to be abuse, an attack, and people might sanction such an admin who is trying to delete other people's events illegitimately.

Furthermore, this STILL is a "we must use events for everything" false belief that nobody seems to be reexamining.  It's akin to the situation that once port 80 was used for www, all the other unix ports were no longer used, everything had to be forced over port 80. I don't know why humans behave this way.

I'm fully in favor of relays having pubkey identities. But that doesn't allow someone to delete someone else's event with a kind-5.

mikedilger:
Let me make the distinction a bit clearer.  Events are transmissible things that flow around the nostr network, notes transmitted through relays.  What I'm proposing is not something that should flow around the network.  So it doesn't make sense to me to shoehorn it into an event, despite prior art by various people who did similar things.

mikedilger:
In fact I really don't want people seeing that I removed their event from my relay.  I'm supposed to create a digitally signed event that proves I did it?  Fuck that.

I'm just going to go around nostr and delete events from my relay using a different protocol.  And you guys can do whatever, I don't really care.  I just thought we could share tooling, but I guess we can't.

fiatjaf:
I understand your concerns and how `["REMOVE"]` can be clearer, but to me it seems less practical and harder to get compatibility across implementations, thinking above how many other commands we would have to add in order to get a full admin dashboard for a relay, like `["BLOCKIP"]` or `["INCREASEPRICE"]`, I don't know (silly examples, but you get the point).

I think it is implied in the suggestions above that events that are to be treated as "commands" should probably not be kept or at least not returned in queries or forwarded to other relays, they can just be received by one specific relay, acted upon, then deleted.

mikedilger:
It's fine. I closed the PR. But this issue is non-negotiable on my end. I will not be digitally signing such commands which might leak in the case of a bug. I'll just add undocumented non-standard private extensions.

PR #1172, opened 2024-04-13
Kind mute sets

### LINKED ISSUE #1156: Mute by kind
It would be nice to be able to mute particular pubkey's kind like "mute retweets" in X (Twitter).

- Reposts (kind 6, 16)
- Reactions (kind 7)
- Zaps (kind 9735)
- etc

### JSON

Based on [NIP-51](51.md) sets.

```json
{
  "kind": 30007,
  "content": "<encrypted tags>",
  "tags": [
    ["d", "<kind>"],
    ["p", "<pubkey 1>"],
    ["p", "<pubkey 2>"]
  ]
}
```

### Tags

- `d` (MUST): kind string
- `p`s (MAY): pubkeys
  - Same as kind 10000

SnowCait:
Fix #1156

### JSON Example

```json
{
  "kind": 30007,
  "content": "<encrypted tags>",
  "tags": [
    ["d", "<kind>"],
    ["p", "<pubkey 1>"],
    ["p", "<pubkey 2>"]
  ]
}
```

### DIFF
--- a/51.md
+++ b/51.md
@@ -48,6 +48,7 @@ Aside from their main identifier, the `"d"` tag, sets can optionally have a `"ti
 | Bookmark sets | 30003 | user-defined bookmarks categories , for when bookmarks must be in labeled separate groups    | `"e"` (kind:1 notes), `"a"` (kind:30023 articles), `"t"` (hashtags), `"r"` (URLs) |
 | Curation sets | 30004 | groups of articles picked by users as interesting and/or belonging to the same category      | `"a"` (kind:30023 articles), `"e"` (kind:1 notes)                                 |
 | Curation sets | 30005 | groups of videos picked by users as interesting and/or belonging to the same category        | `"a"` (kind:34235 videos)                                                         |
+| Kind mute sets | 30007 | mute pubkeys by kinds<br>`"d"` tag MUST be the kind string                                  | `"p"` (pubkeys)                                                                   |
 | Interest sets | 30015 | interest topics represented by a bunch of "hashtags"                                         | `"t"` (hashtags)                                                                  |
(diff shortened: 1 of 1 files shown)


fiatjaf (approved):

AsaiToshiya:
+1. 

I feel like it is more generic mute list than the sets. How about using the `k` tag after all?

SnowCait:
Adding both `k` tag and `p` tag to kind 10000 makes it complex and will breaks existing clients.

Expanding `k` tag to be able to contain pubkey would cause conflicts between NIPs in the future. (It may be okay if the `k` tag has a different meaning for each NIP, but I don't like it.)

AsaiToshiya:
It was not communicated properly. What I want to say is this: 

```json
{
  "kind": 30007,
  "content": "<encrypted tags>",
  "tags": [
    ["d", "<identifier>"],
    ["k", "<kind>"],
    ["p", "<pubkey 1>"],
    ["p", "<pubkey 2>"]
  ]
}
```

How about using the `k` tag like this? Without the `k` tag, it just looks like a generic mute list, not the sets.

SnowCait:
I see.

If `<identifier>` is free input, multiple events will be created for a kind.
It makes clients complex to use kind mute sets.

If `<identifier>` is same as `<kind>`, please see #1156 discussion and edit history.

AsaiToshiya (approved):
Ok, I guess I didn't understand Alex's comment correctly.

SnowCait:
Implemented in [nostter](https://github.com/SnowCait/nostter) and [lumilumi](https://github.com/TsukemonoGit/lumilumi).

PR #1173, opened 2024-04-14
EOSE extensions

Semisol:


pablof7z:
Concept NACK -- this adds too much complexity that I don't think is warranted, and adds a path for ever increased complexity with the `x_` prefixes and protocol fragmentation.

Even if we didn't have another way of doing pagination (and we do), I think this approach is too complex.

v0l:
I think its better to just include a new message response instead of lumping it into `EOSE` 

```json
["CURSOR", <id>]
```
MUST be sent before `EOSE`, clients can resume fetching results by sending another request with the cursor ID `["REQ",{"cursor": <id>}]`

Something like that, i dont think you need those other extensions on a way to continue reading the cursor.

fiatjaf:
This is a bad idea. Relays are not databases, but you can already paginate with `until`. No relay that exists today implements persistent cursors and they probably shouldn't. Keep some weird state in between queries? That's unnecessary cumbersomeness.

But yes, if you really want cursors then `["CURSOR"]` messages make more sense.

And what is the point of `x_` prefixed extensions? Has that worked well for HTTP? "Let's standardize the unstandardized!" makes no sense.

mleku:
yeah, i think it's unnecessary, one or two events sent twice is no big deal, just shift the "until" timestamp to the oldest (last) returned

cursors and pagination are state and if you try to set arbitrary markers other than timestamps you have a problem of race conditions and total event ordering

this is definitely functionality that belongs in a cache not in a relay

mikedilger:
I am against this proposed solution.

I am against any stateful queries on relays. 

Chorus database is not SQL and I can't just whip up some SQL to do it.

Pagination isn't ideal right now but it is workable.

vitorpamplona:
Sending the cursor is better than cutting clients off when relays hit their internal LIMIT config as it happens today. Many relays send an EOSE at 500 events even though they have more. It's up to clients to figure out if it is a real EOSE or it is fake. That is bad. The current way to figure out if an EOSE is real is to repeat the filter with a separate until and see if 0 events return. If not, keep changing `until` at every EOSE. 

I am ok with extending EOSE but I do not understand the benefits of the auth extension as proposed.

mleku:
> > I am against any stateful queries on relays.
> 
> There is no state that you need to store. How many times do I have to say this? You can very well do this on a K/V database also.

it's state, the query is state, and the only way to implement what you describe requires storing at least all of the IDs of the events that the query matched

more useful would be to propose exactly that, a new variant like count but instead of count it's just an ID list, then the client is free to paginate it as it likes

PR #1175, opened 2024-04-14
NIP-35: Torrents

v0l:
Making a NIP for torrents with a few updates

read: https://github.com/nostr-protocol/nips/blob/feat/nip35/35.md

### DIFF
--- a/35.md
+++ b/35.md
@@ -38,7 +38,7 @@ In some cases the url mapping isnt direct, mapping the url in general is out of 
 
 ```json
 {
-  "kind": 2003,›
+  "kind": 2003,
   "content": "<long-description-pre-formatted>",
   "tags": [
     ["title", "<torrent-title>"],
--- a/35.md
+++ b/35.md
@@ -11,7 +11,7 @@ This NIP defined a new `kind 2003` which is a Torrent.
 `kind 2003` is a simple torrent index where there is enough information to search for content and construct the magnet link. No torrent files exist on nostr.
 
 ## Tags
-- `i`: V1 BitTorrent Info Hash, as seen in the [magnet link](https://www.bittorrent.org/beps/bep_0053.html) `magnet:?xt=urn:btih:HASH`
+- `x`: V1 BitTorrent Info Hash, as seen in the [magnet link](https://www.bittorrent.org/beps/bep_0053.html) `magnet:?xt=urn:btih:HASH`
 - `file`: A file entry inside the torrent, including the full path ie. `info/example.txt`
 - `tracker`: (Optional) A tracker to use for this torrent
 
@@ -19,9 +19,9 @@ In order to make torrents searchable by general category, you SHOULD include a f
 
 ## Tag prefixes
 
-Tag prefixes are used to label the content with references, ie. `["t", "imdb:1234"]`
+Tag prefixes are used to label the content with references, ie. `["i", "imdb:1234"]`
 
-- `tcat`: A comma separated text category path, ie. `["t", "tcat:video,movie,4k"]`, this should also match the `newznab` category in a best effort approach.
+- `tcat`: A comma separated text category path, ie. `["i", "tcat:video,movie,4k"]`, this should also match the `newznab` category in a best effort approach.
 - `newznab`: The category ID from [newznab](https://github.com/Prowlarr/Prowlarr/blob/develop/src/NzbDrone.Core/Indexers/NewznabStandardCategory.cs)
 - `tmdb`: [The movie database](https://www.themoviedb.org/) id.
 - `ttvdb`: [TV database](https://thetvdb.com/) id.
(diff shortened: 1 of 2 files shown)


vitorpamplona (approved):

fiatjaf (approved):

v0l:
@hzrd149

hzrd149:
This looks great. however It could be worth mentioning k:2004 events that dtan and noStrudel use for comments on torrents since I think they could be useful for other clients wanting to implement this nip

The comment events are similar to kind 1 and follow NIP-10

utxo-one:
Great NIP!

hzrd149:
looks good, I like the addition of the comments :+1:

flox1an:
looks good! Makes sense to have the concatenation of the category hierarchy to make filtering work on the relays.

flox1an:
FYI this PR uses `i` tags for external references: https://github.com/nostr-protocol/nips/pull/1185

Could make sense for `imdb`, `tmdb`, .. and the like.

jb55:
what about v2?

jb55:
v2 is a lot more interesting that v1: merkle torrents so you can incrementally update a torrent file over time while sharing parts, etc. the same file can share seeders across many different torrents. I having been looking at v2 for notedeck file sharing, but it looks like this is hardcoded to v1.

fiatjaf:
Is anyone using v2?

jb55:
they should be, but it is not widely adopted yet no.

melvincarvalho:
Great idea.  What about a tag for thumbnail or images?  Some trackers and clients have a thumb which makes the navigation UX much better.  Or hover over for a larger image.

jb55:
it would be useful if the torrent metadata was embedded in the note itself, bencoded or as json. That way you could express v1 or v2 torrents, and you could see all of the files, their merkle root hashes (in the case of v2), and the corresponding trackers.

Then nostr clients could open the note as a torrent, allowing you to download individual files from bittorrent or even blossom nodes by looking at file hashes.

image

jb55:
Hmm started diving into the spec and the pieces root may not be the same has the file hash, so this may foil my plan to have it compatible with blossom. darn...

fiatjaf:
Regardless of that I like the idea of having the metainfo inside the event.

jb55:
although one interesting observation is by breaking a file into 16KiB blocks you can always reconstruct the corresponding bittorrent v2 file hash (V2FH) for a given file: 

image

so a smart blossom seeder that wanted to help seed v2 torrents would:

- split entire file into `N` 16 KiB blocks, hash these blocks individually, and then hash these all together into the bittorrent-v2 sha2 file hash (`V2FH`)
- store this root hash alongside the standard sha2 hash of the file
- A hybrid torrent/blossom client could check to see if a blossom server has a file from some torrent by making a HEAD request for the `V2FH`

Let's say the client is only missing the 2nd block from the torrent file, we can make a range request against the blossom server for the second block:

```
GET /5584a1cc51f777b13b9de2a6a1012ee0e7a93652143c415901b8fe005ed1124c HTTP/1.1
Host: blossom.xyz
Range: bytes=16384-32768
````

Where `5584a1cc51f777b13b9de2a6a1012ee0e7a93652143c415901b8fe005ed1124c` is the `VTFH`

Then we can hash the response and match it against our torrent metadata `piece layers` to confirm it's correct.

This is getting a bit off topic, but just wanted to put this out for consideration when it comes to torrents + nostr.

**Comment by v0l** [comment]
> compatible with blossom

Not a huge fan of Blossom, there is a lot of overlap with NIP96, it would have been nicer to reuse that spec and expand.

> it would be useful if the torrent info-dictionary was embedded in the note itself,

Sounds good to me, i can add this to the spec if you still think its a good idea?

**Comment by arthurfranca** [comment]
> Not a huge fan of Blossom, there is a lot of overlap with NIP96, it would have been nicer to reuse that spec and expand.

Finally someone said it! Never got people praising blossom. It is a NIP-96 rip-off hurting interop.

Blossom spec is exactly the same as NIP-96 but with small changes that bring incompatiblity like `kind:10063` for discovery instead of `kind:10096` and JSON responses that have the same fields from NIP-96 but with different structure for no reason.

Blossom only additions that could be simply added to NIP-96: a route to list uploaded items of a pubkey, a HEAD one to check for an item existence.

**Comment by jb55** [comment]
blossom, nip-96, blobstr, it's all the same, the only relevant thing to my proposal above is and endpoint for fetching data in a CAS-way. blobstr is the most minimal that is needed to support this: 

- https://github.com/nostr-protocol/nips/pull/1098 <- blobstr

**Comment by v0l** [comment]
> blossom, nip-96, blobstr, it's all the same, the only relevant thing to my proposal above is and endpoint for fetching data in a CAS-way. blobstr is the most minimal that is needed to support this:
> 
> * [Draft for file archival and retrieval NIP #1098](https://github.com/nostr-protocol/nips/pull/1098) <- blobstr

Fair enough, ive been building a Blossom/NIP-96 server over here: https://git.v0l.io/Kieran/void-cat-rs would be cool to trial the bittorrent compat with this.

One thing im not sure of is how you get the hash of each file from the torrent, since we're only storing the root hash, do we also need to store the hash of each file? because we have the `file` entry already, maybe we should include the `sha256` in those tags?

And once we have a way to link these torrent files which servers would you lookup?

File: 35.md [hunk 1]
```diff
@@ -0,0 +1,70 @@
...
+    ["t", "4k"],
+  ]
+}
+```
+
+## Torrent Comments
+
+A torrent comment is a `kind 2004` event which is used to reply to a torrent event.
+
+This event works exactly like a `kind 1` and should follow `NIP-10` for tagging.
```
arthurfranca, line 66: Save the ~planet~nostr, guys! Support #1233 with a :+1:? =D

hzrd149: I have not read through the entirety of #1233 but I think there is merit in having a specific kind for torrent commends and using a simple schema like NIP-10 for marking replies and threads

It seems reasonable that a NIP would define its primary event kind (in this case `2003`) and a comment / reply kind that is specifically used for it. It seems like a better solution then using `k` tags as a "meta" kind

vitorpamplona: Thread roots should be defined by the NIP. Replies can be generalized. 

Otherwise, should we have reactions, zap, labels, and report kinds for each NIP as well? Replies are no different than reactions.

v0l: Seems reasonable, but im not sold on the new tag structure

v0l: Going to merge anyway, we can always change it later once your PR is merged, nobody is really using comments on torrents yet. I only see 24 comments so far.

PR #1178, opened 2024-04-15
Add previous tag

v0l:
Add a previous tag to replaceable events which prevents updates when the id does not match.

### DIFF
--- a/01.md
+++ b/01.md
@@ -99,6 +99,8 @@ And also a convention for kind ranges that allow for easier experimentation and 
 
 In case of replaceable events with the same timestamp, the event with the lowest id (first in lexical order) should be retained, and the other discarded.
 
+When the `previous` tag exists on a replaceable event relays SHOULD use this id as the id of the version the client wishes to replace, if the id does not match the stored version, relays MUST reject the update.
+
 When answering to `REQ` messages for replaceable events such as `{"kinds":[0],"authors":[<hex-key>]}`, even if the relay has more than one version stored, it SHOULD return just the latest one.
 
 These are just conventions and relay implementations may differ.


01.md
```diff
@@ -99,6 +99,8 @@ And also a convention for kind ranges that allow for easier experimentation and
 
 In case of replaceable events with the same timestamp, the event with the lowest id (first in lexical order) should be retained, and the other discarded.
 
+When the `previous` tag exists on a replaceable event relays SHOULD use this id as the id of the version the client wishes to replace, if the id does not match the stored version, relays MUST reject the update.
```
fiatjaf, line 102:
```suggestion
A `previous` tag in the format `["previous", "<id>", "<id>", ...]` may be specified containing any number of ids. When it is present relays should only accept the new event if they know at least one of the ids.
```

fiatjaf:
Initially I liked this idea, but isn't this solvable by the client just querying the relay before publishing? The logic for implementing this on the client side will be harder than what it is today and will lead to clients implementing it wrongly and users losing data anyway.

Instead of

"why did the people I followed last week vanished today?"

we'll have

"oh, didn't I follow this person yesterday, why is it not showing up now?"

Because the update that added the new follow was rejected.

v0l:
> Initially I liked this idea, but isn't this solvable by the client just querying the relay before publishing

Yes and this is a way to force clients to implement a query to the relays first, im sick of these issues persisting in my own clients and i want to make it harder for me to replace these events.

> Because the update that added the new follow was rejected.

If it was reject this needs to be handled in the client, again i want to make it harder for myself.

vitorpamplona:
There is a risk that this will make replaceables diverge further away from each other. 

Also, clients who are not implementing it will modify the list while keeping the `previous` intact. Very few clients discard tags they don't recognize from replaceables. They just try to keep as many of them as possible. Which creates a problem for this one. Since the previous won't be updated, these clients will never be able to update replaceables with an existing `previous`.

Issue #1170, opened 2024-04-11
Does anybody else see the need for a location-aware kind1 feed?


vitorpamplona:
I want to post things that should only be shown to users in my neighborhood/city. They are not supposed to show on everybody's default feed. 

For instance, If I want to make a question about a school nearby, I don't want that question to be a kind1 and go into everybody's feed. It's not for that audience. 

Shall we create a kind1 copycat event type that requires location?

mterenzio:
I've been thinking about how to best do that with existing NIPS and this
seems way easier. Interested to hear what the drawbacks might be

v0l:
Kind 1 is broadcast, so yea i guess you need another kind

mterenzio:
How would we handle the challenge of the clients knowing what relays to
connect to since one would be interested in events based on location and
not individuals? Unless some relays were dedicated to locations? Or else a
version of tag groups that somehow works alongside the location data

vitorpamplona:
Quick PR for this: https://github.com/nostr-protocol/nips/pull/1180

vitorpamplona:
Solved with NIP-22

PR #1180, opened 2024-04-16
Adds Topical Notes

vitorpamplona:
The goal here is to create Notes that are only displayed if you follow certain topics from certain relays. 

With this, users can post into their City's relays or into a Hashtag's relays without polluting their main feed. 

Solves #1170 

Read [here](https://github.com/vitorpamplona/nips/blob/location-based-kind1/54.md)

### DIFF
--- a/54.md
+++ b/54.md
@@ -1,0 +1,60 @@
+NIP-54
+======
+
+Topical Notes
+-------------
+
+`draft` `optional`
+
+This NIP enables authors to post exclusively into topics, without polluting their main feed. Authors can post to locations, hashtags, labels, and urls in a way that only followers of those topics will receive their posts. 
+
+## Topical Note Kind
+
+Kind `21` defines a `kind:1`-equivalent event that MUST NOT be displayed unless the user follows one of the `g`, `t`, `r` or `l` tags. 
+
+```jsonc
+{
+  "kind": 21,
+  "tags": [
+    ["g", "<geohash>"],
+    ["t", "<hashtag>"],
+    ["r", "<url>"],
+    //.. 
+    ["L", "<label namespace>"],
+    ["l", "<label>", "<label namespace>"],
+  ],
+  "content": "This is a note for followers of the hashtag, location OR url. My followers should not see this.",
+  //...
+}
+```
+
+`.content` MUST be plain text and tags that mention, quote and assemble threading structures MUST follow [NIP-10](10.md).
+
+## Topical Relay Setup
+
+Kind `30021` defines a topic follower. It registers a list of relays to aggregate events for each topic. 
+
+The `d` tag of this kind is equal to `name:value` of the topics used in kind `21`. 
+
+```js
+{
+  "kind": 30021,
+  "tags": [
+    ["d", "g:<my city's geohash>"],
+    ["r", "wss://my-city-owned-relay"],
+    ["r", "wss://independent-but-city-based-relay"],
+    // ... other tags
+    ["title", "Name of this topic"]
+  ],
+  "content": nip44Encrypt([
+    ["r", "wss://my-city-owned-relay2"], // private relay
+    ["title", "Name of this topic"] // private topic title
+    // ... other tags
+  ]),
+  //...other fields
+}
+```
+
+The `.content` contains a list of private tags.
+
+Clients MUST seek and broadcast events from and to the relays defined for each topic. 
--- a/54.md
+++ b/54.md
@@ -28,27 +28,33 @@ Kind `21` defines a `kind:1`-equivalent event that MUST NOT be displayed unless 
 }
 ```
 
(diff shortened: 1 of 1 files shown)


staab:
Check to make sure those kinds aren't used anywhere else, they seem pretty likely to have already been nabbed. Also potentially of concern is location privacy, this pretty thoroughly doxxes users who use this feature, maybe @arkin0x has thought about that.

Otherwise, I think this is a really interesting idea. It pushes for de-commodification of relays, since if relays don't establish whether they support these events for a given topic/location, any popular tag will ultimately have 30021 events on all relays. I'll be interested to see how this works out.

vitorpamplona:
> Check to make sure those kinds aren't used anywhere else

Yep, they seem to be clean. I have a little crawler that hits every single visible relay for this: https://crawler.amethyst.social/. There is no kind 21 event out there. Kind 30021 has a total of 9 events that look like tests from June last year.

vitorpamplona:
> Also potentially of concern is location privacy

On the privacy question, we could use random d-tags (UUID) with the following label declared in the encrypted part. Maybe that is even better because it allows the user to follow many tags/labels (different spellings, languages, etc) for the same "topic".

staab:
Sure, but that wouldn't solve privacy problems for publishers of kind 21s. The only way to do that would be to encrypt the note for desired recipients. Obviously not feasible on an individual basis, so it would require some kind of key exchange. I don't know if we need that for this to be useful, I just know that I personally would not publish a kind 21 with a g tag.

vitorpamplona:
ohh I see. Yes, these are for the folks that want to make public events. 

Although the `g` tag means they are interested in a location or that the post was for a location. It might not necessarily mean that they are based on that location. Of course, the tag's multi-resolution options also help to not pinpoint somebody's exact location with high precision.

54.md
```diff
@@ -0,0 +1,60 @@
...
+```jsonc
+{
+  "kind": 21,
+  "tags": [
+    ["g", "<geohash>"],
+    ["t", "<hashtag>"],
+    ["r", "<url>"],
+    //.. 
+    ["L", "<label namespace>"],
+    ["l", "<label>", "<label namespace>"],
```
arkin0x, line 24: Can we be a bit more explicit that only 1 of these tags is required?

Is it possible that one would ever use 2 or more?

54.md
```diff
@@ -0,0 +1,60 @@
...
+}
+```
+
+`.content` MUST be plain text and tags that mention, quote and assemble threading structures MUST follow [NIP-10](10.md).
+
+## Topical Relay Setup
+
+Kind `30021` defines a topic follower. It registers a list of relays to aggregate events for each topic. 
+
+The `d` tag of this kind is equal to `name:value` of the topics used in kind `21`.
```
arkin0x, line 37:
```suggestion
The `d` tag of this kind is equal to `tag:value` of the topics used in kind `21`. 
```

How do we handle multiple topics here? Comma separated?

arkin0x:
This is cool!

So, is the `g` tag required for usage of topics, or is it only required if you choose to subscribe to a place? IMHO this seems useful even if people don't use the location aspect.

I'm not sure how useful the `g` tag actually is because you could also just make the topic "#newyork" and that would probably be more intuitive than picking the geohash containing New York. The interface a client uses to select the topic tags would need a map interface for the `g` tag stuff and it could be fairly unintuitive to the pleb who doesn't already understand geohashes.

I'm not against the `g` tag being part of this because it's a cool idea but I think that if you want to see what people are posting via geohash it's easier to query the `g` tag for kind1 notes.

To address @staab 's question, subscribing to a `g` topic indicates your interest in an area; typically this would mean you live there or someone you know lives there. I think the key is to make people aware of this with good UI and let them make informed decisions, and also limiting the precision of the `g` tags in the client so that people don't accidentally subscribe to a place that is too specific. Geohashes with a length of 5 is my personal maximum for location-sharing comfort zone for accuracy; I only allow 1-5 in Yondar. (Higher means more precise which means less privacy.)

54.md
vitorpamplona: Yes, they can do one or more if you consider translations and spellings, like: 

```jsonc
{
  "kind": 21,
  "tags": [
    ["t", "Amethyst"],
    ["t", "amethyst"],
    ["t", "Ametista"],
    ["t", "ametista"],
  ],
  //...
}
```

54.md
vitorpamplona: I just replaced the name:value for a UUID and added the following tags directly on the 30021 event as well.

mterenzio:
Certainly no client should publish exact location unless that was clearly
what the user intended, but certainly geohash has a place and utility
beyond a tag for some applications

vitorpamplona:
Replaced by NIP-22

PR #1182, opened 2024-04-17
Linked cryptographic identities

franzaps:
[Rendered](https://github.com/fr4nzap/nips/blob/linked-cryptographic-identities/69.md)

### DIFF
--- a/69.md
+++ b/69.md
@@ -1,0 +1,96 @@
+NIP-69
+======
+
+Linked Crytographic Identities
+------------------------------
+
+`draft` `optional`
+
+Nostr users may have other crytographic identities such as PGP or SSL certificates. This NIP introduces a replaceable event of kind `10069` that links these identities with a Nostr pubkey.
+
+The event MAY contain zero or more `i` tags. Each tag MUST have, in the following order:
+ - A type (listed below) and a public key fingerprint that MUST NOT contain spaces, joined by a `:`
+ - A signature of the message (described below) in base64 format, unwrapped
+ - The full public key in base64 format, unwrapped
+
+Types: `pgp`, `x509`
+
+Message: `"By signing this message I confirm that I control the private key for the Nostr public key <npub>"`, where `npub` is the bech-32 encoded form of the `pubkey` of the event
+
+```jsonc
+{
+  "kind": 10069,
+  "tags": [
+    ["i", "<type>:<fingerprint>", "<signature>", "<public key>"],
+    ["i", "<type>:<fingerprint>", "<signature>", "<public key>"],
+    // ...
+  ],
+  // ...
+}
+```
+
+## Revocation
+
+The event MAY be updated by removing `i` tags that no longer apply.
+
+## Examples
+
+### PGP
+
+```jsonc
+{
+  "kind": 10069,
+  "tags": [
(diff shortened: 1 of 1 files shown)


vitorpamplona:
Nice! Is it necessary to index the `<type>:<fingerprint>`? Do we expect people to query by it?

fr4nzap:
@vitorpamplona Thanks and yes, absolutely. This way coming across any PGP key or certificate in the wild one can check if that identity is also on nostr.

In my particular case for zap.store, it's one of the ways to link APKs with developers on nostr. 

@fiatjaf this is my second attempt after failure with NIP-39, what do you think?

dskvr:
Looks good. Seems like this could also link other nostr pubkeys, I could use this now. 

> Is it necessary to index the <type>:<fingerprint>

Necessary for my use case as well.

69.md
```diff
@@ -0,0 +1,96 @@
...
+`draft` `optional`
+
+Nostr users may have other crytographic identities such as PGP or SSL certificates. This NIP introduces a replaceable event of kind `10069` that links these identities with a Nostr pubkey.
+
+The event MAY contain zero or more `i` tags. Each tag MUST have, in the following order:
+ - A type (listed below) and a public key fingerprint that MUST NOT contain spaces, joined by a `:`
+ - A signature of the message (described below) in base64 format, unwrapped
+ - The full public key in base64 format, unwrapped
+
+Types: `pgp`, `x509`
```
dskvr, line 16: Should keys be enumerated? As drafted, this NIP can work for any key that can produce a signature.

fr4nzap: Not listing them would lead to undesired inconsistencies (example, someone prefixes the `i` tag with `gpg:` instead of `pgp:`).

We can add a nostr type but there again it's important to define the exact format of the event to be signed.

dskvr: Why would it be a problem if a client prefixed the `i` tag with `gpg` for a `gpg` signature?

fr4nzap: So clients can easily find these events?

In my particular case I am indexing all sorts of release artifacts (APKs, DMGs, AppImages, etc) some signed with certificates and others with PGP. Given an artifact, I want to be able to quickly look up if its signer also controls a nostr key. If we use `pgp` or `gpg` interchangeably now clients have to make two different requests. Don't you think it makes sense?

Of course this is nostr and people can ultimately publish whatever they want

dskvr: Not necessarily. You: "I support PGP proof", Them: "I have GPG proof", You: "..."

> If we use pgp or gpg interchangeably now clients have to make two different requests.

While this may be superfluous since your case seems to flag encryption support, if you wanted to be extra safe you could do `{ "i#": { 'pgp:<key>', 'gpg:<key>'}` ... `'pgp:<key>' or 'gpg:<key>'` in a single `REQ`.

> Of course this is nostr and people can ultimately publish whatever they want

And it is because of this I questioned enumeration.

fr4nzap: I think it's good to have some basic agreement on these labels! I will change it to "recommended" types so it's just that, a recommendation, and let's add a `nostr` one.

franzaps:
@Semisol if you confirm it will get merged I can make a PR to NIP-39

franzaps:
@staab @fiatjaf are you active here? This one is good to merge. I'm happy to create a PR on NIP-39 if I get a confirmation that it will actually get merged.

staab:
It seems like this does the same thing as NIP 39? I like the dedicated event as an addition, but I would add it as an addition to NIP 39, unless I'm misunderstanding its purpose. Is this implemented anywhere?

franzaps:
> It seems like this does the same thing as NIP 39? I like the dedicated event as an addition, but I would add it as an addition to NIP 39, unless I'm misunderstanding its purpose. Is this implemented anywhere?

It used to be in NIP-39, but was flawed, and semisol silently removed it instead of asking for a fix. This is a new proposal. I'm fine with either kind 10069 or kind 0. I can create a PR to NIP-39 but I first wanted to make sure it will be merged.

Yes it's being implemented! Developers on zap.store will start using this very soon. I started producing these events but only on relay.zap.store.

caesar:
I'd love to see a standardised way to claim an OpenPGP identity in Nostr, but I have a few thoguhts:

- I think it would make more sense to add these identities to NIP-39 instead of adding another message type; I don't understand the benefit to there being a new message type.

- I think the verbiage of the English-language message to be signed is unnecessarily verbose. It would probably make sense for the sake of consistency to use the same proof string used by NIP-39: `Verifying that I control the following Nostr public key: "<npub encoded public key>"`

- I think it would probably make sense to use the prefix `openpgp4fpr`, which is a [standard](https://metacode.biz/openpgp/openpgp4fpr). Note that `PGP` and `GPG` are the names of software, the encryption standard they implement is called `OpenPGP`.

franzaps:
@caesar Yes this will be added to NIP-39. I don't mind updating the spec to what you suggest.

franzaps:
Closing in favor of https://github.com/nostr-protocol/nips/pull/1335

PR #1184, opened 2024-04-20
NIP-61: Danmaku

CodyTseng:
Danmaku (弾幕 だんまく; 弹幕; literally translated as "bullet curtain" and figuratively as "barrage") is a subtitle system in online video platforms that originates from Japan and popularised in Mainland China. Such system allows user to post moving comments onto a playing video that are synchronized to the video timeline. The comments are typically presented as “shooting” across the screen, resembling a barrage.

This NIP is being implemented and experimented with in [danmakustr](https://github.com/CodyTseng/danmakustr), but it's still a work in progress. Feedback is welcome

WechatIMG40

### DIFF
--- a/61.md
+++ b/61.md
@@ -1,0 +1,36 @@
+NIP-61
+======
+
+Danmaku
+-------
+
+`draft` `optional`
+
+This NIP defines a new kind of event, `2333`, for danmaku.
+
+Danmaku (弾幕 だんまく; 弹幕; literally translated as "bullet curtain" and figuratively as "barrage") is a subtitle system in online video platforms that originates from Japan and popularised in Mainland China. Such system allows user to post moving comments onto a playing video that are synchronized to the video timeline. The comments are typically presented as “shooting” across the screen, resembling a barrage.
+
+### Format
+
+The format uses a regular event kind `2333`.
+
+The `.content` of these events is the content of the danmaku comment.
+
+The list of tags are as follows:
+* `i` (required) universally unique identifier of the video. The format is `<platform (lowercase)>:<identity>` where `<platform>` is the platform of the video and `<identity>` is the unique identifier of the video on that platform. For example, a YouTube video with the identifier `aA-jiiepOrE` would have the identifier `youtube:aA-jiiepOrE`.
+* `time` (required) the time of the video when the danmaku comment is sent in seconds.
+
+```jsonc
+{
+  "id": "<...>",
+  "pubkey": "<...>",
+  "kind": 2333,
+  "tags": [
+    ["i", "youtube:aA-jiiepOrE"],
+    ["time", "3.018597"]
+  ],
+  "content": "Hello Nostr!",
+  "sig": "<...>",
+  "created_at": 1713368851
+}
+```
--- a/README.md
+++ b/README.md
@@ -69,6 +69,7 @@ They exist to document what may be implemented by [Nostr](https://github.com/nos
 - [NIP-57: Lightning Zaps](57.md)
 - [NIP-58: Badges](58.md)
 - [NIP-59: Gift Wrap](59.md)
+- [NIP-61: Danmaku](61.md)
 - [NIP-65: Relay List Metadata](65.md)
 - [NIP-72: Moderated Communities](72.md)
 - [NIP-75: Zap Goals](75.md)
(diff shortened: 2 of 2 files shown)


staab:
I've been waiting for someone to build this, I think it's a great idea. I haven't looked closely at it, but you should consider modeling this as an extension (using a new kind) of [NIP 84](https://github.com/nostr-protocol/nips/blob/master/84.md).

fiatjaf:
I feel like this should go on that videos NIP.

CodyTseng:
> you should consider modeling this as an extension (using a new kind) of [NIP 84](https://github.com/nostr-protocol/nips/blob/master/84.md?rgh-link-date=2024-04-20T19%3A26%3A40Z).

Could you please provide a brief explanation of how to extend NIP-84? Is it accomplished by adding a `comment` tag? I'm not very familiar with NIP-84, but it seems like it might not be suitable for comments.

> I feel like this should go on that videos NIP.

Are you referring to the `Live Streaming chat message` in [NIP-53](https://github.com/nostr-protocol/nips/blob/master/53.md)? I do see many similarities between the two. However, I'm unsure if incorporating this feature into NIP-53 would be appropriate.

staab:
> Could you please provide a brief explanation of how to extend NIP-84

Reviewing it, I think the two are conceptually similar, but it would be hard to modify NIP 84 to work for this use case. I actually like this NIP better on closer inspection.

61.md
```diff
@@ -0,0 +1,36 @@
...
+======
+
+Danmaku
+-------
+
+`draft` `optional`
+
+This NIP defines a new kind of event, `2333`, for danmaku.
+
+Danmaku (弾幕 だんまく; 弹幕; literally translated as "bullet curtain" and figuratively as "barrage") is a subtitle system in online video platforms that originates from Japan and popularised in Mainland China. Such system allows user to post moving comments onto a playing video that are synchronized to the video timeline. The comments are typically presented as “shooting” across the screen, resembling a barrage.
```
staab, line 11: Background can be elided, how comments are displayed should be up to the client (I would find this very useful, but would _not_ want the comments to be flying past).

61.md
```diff
@@ -0,0 +1,36 @@
...
+
+### Format
+
+The format uses a regular event kind `2333`.
+
+The `.content` of these events is the content of the danmaku comment.
+
+The list of tags are as follows:
+* `i` (required) universally unique identifier of the video. The format is `<platform (lowercase)>:<identity>` where `<platform>` is the platform of the video and `<identity>` is the unique identifier of the video on that platform. For example, a YouTube video with the identifier `aA-jiiepOrE` would have the identifier `youtube:aA-jiiepOrE`.
```
staab, line 20: Why not just use a `r` with the url of the video instead of inventing our own new data format? If we're talking podcast episode guids or something that's different, but for referring to centralized entities where the client is bundled with the data, I'd just use a url.

61.md
```diff
@@ -0,0 +1,36 @@
...
+* `time` (required) the time of the video when the danmaku comment is sent in seconds.
```
staab, line 21: I would like to see an optional `duration` tag. This would be useful for slightly different use cases from danmaku, for example the little comment/link boxes that used to hover over youtube videos.

staab (approved):

61.md
CodyTseng: Sure. I'm worried that some people might not know what "danmaku" refers to, so I added some background information. Indeed, not all danmaku are flying past, lol. This should be handled by the client.

61.md
CodyTseng: Because the URL for the same video can vary with different parameters, and sometimes unexpected ones at that. The URL for the same video can be different on mobile and desktop platforms, and also differ between China and the US. So, I think having a simple ID convention would be better. If someone wants to use the URL directly, that's also possible because it still conforms to the <a>:<b> format.

61.md
CodyTseng: Great idea. We can also add optional tags like color, speed, and so on.

61.md
staab: Maybe a quick definition would be useful

61.md
staab: Interesting. I feel like this points to a bigger problem with standardization; regular urls basically treat closed services like protocols, which you're just making explicit here. I'm not sure how to solve this long-term.

CodyTseng: I believe platforms are unlikely to change IDs easily, so for now, using IDs seems to be a better approach than using URLs. Indeed, this cannot completely solve the problem.

61.md
CodyTseng: Alright, I'll add a brief definition.

PR #1185, opened 2024-04-20
External Content IDs

MerryOscar:
There are certain established global content identifiers that would be useful to reference in nostr events so that clients can query all events associated with these ids.

- Book [ISBNs](https://en.wikipedia.org/wiki/ISBN)
- Podcast [GUIDs](https://podcastnamespace.org/tag/guid)
- Movie [EIDRs](https://www.eidr.org)

Formalising the way these ids are referenced in nostr events would make it much easier for developers to build open versions of popular services like IMDB, GoodReads etc. built on nostr.

### DIFF
--- a/XX.md
+++ b/XX.md
@@ -1,0 +1,31 @@
+NIP-XX
+======
+
+External Content IDs
+-------------------------
+
+`draft` `optional`
+
+There are certain established global content identifiers that would be useful to reference in nostr events so that clients can query all events assosiated with these ids.
+
+- Book [ISBNs](https://en.wikipedia.org/wiki/ISBN)
+- Podcast [GUIDs](https://podcastnamespace.org/tag/guid)
+- Movie [EIDRs](https://www.eidr.org)
+
+Since the `i` tag is already used for similar references in kind-0 metadata events it makes sense to use it for these content ids as well.
+
+
+## Supported IDs
+
+### Books:
+
+- Book ISBN: `["i", "book:isbn:123"]`
+
+### Podcasts:
+
+- Podcast Feed GUID: `["i", "podcast:guid:123"]`
+- Podcast Item GUID: `["i", "podcast:item:guid:123"]`
+
+### Movies:
+
+- Movie EIDR: `["i", "movie:eidr:123"]`
--- a/24.md
+++ b/24.md
@@ -40,4 +40,5 @@ tags
 These tags may be present in multiple event kinds. Whenever a different meaning is not specified by some more specific NIP, they have the following meanings:
 
   - `r`: a web URL the event is referring to in some way
+  - `i`: an external id the event is referring to in some way - see [NIP-XX](XX.md)
   - `title`: title of the event


staab (approved):
I've added support for this to #1043. I think this is better than bare guids, because a single event may point to multiple guids, and it would be nice to know which one is which.

XX.md
```diff
@@ -0,0 +1,31 @@
...
+## Supported IDs
+
+### Books:
+
+- Book ISBN: `["i", "book:isbn:123"]`
+
+### Podcasts:
+
+- Podcast Feed GUID: `["i", "podcast:guid:123"]`
+- Podcast Item GUID: `["i", "podcast:item:guid:123"]`
```
Giszmo, line 27: Please use actual examples. ISBN and GUID have formats.

I think it would also be required to mention what to do if a standard uses `:` as separator itself like in IPv6: ` 2001:0db8:0000:0000:0000:ff00:0042:8329` is a valid IPv6.

Casing should also be addressed to not get both ISBN and isbn.

```suggestion
- Book ISBN: `["i", "book:isbn:0-201-63361-2"]`

### Podcasts:

- Podcast Feed GUID: `["i", "podcast:guid:396d9ae0-da7e-5557-b894-b606231fa3ea"]`
- Podcast Item GUID: `["i", "podcast:item:guid:396d9ae0-da7e-5557-b894-b606231fa3ea"]`
```

MerryOscar: @Giszmo thanks for the suggestions I've added real examples for all 3.

I also switched the movie references to use `isan`s rather than `eidr`s because they are simpler and easier to ignore multiple edits of the same movie.

For standards that use `:` in the value my assumption is that's fine as the value is always the last thing in the string - I can add some text about this if you think that's necessary?

Giszmo: I think it would be indeed helpful to clarify that : must not get escaped in any way. That should be enough hint for implementers to consider this case.

alexgleason:
This is very close in what it does to NIP-48 proxy tags. I wonder if they should be combined? Or even if NIP-48 should be deprecated in favor of this?

MerryOscar:
> This is very close in what it does to NIP-48 proxy tags. I wonder if they should be combined? Or even if NIP-48 should be deprecated in favor of this?

@alexgleason yes I did consider just using NIP-48 proxy tags for this use case - see https://github.com/nostr-protocol/nips/issues/760 for the thinking. But in the end I understood NIP-48 to be more about referencing other _posts_ from different protocols whereas this idea is just a link to some external content id. You might actually have a `proxy` tag and an `i` tag in the same nostr event - for example if you were using a book review site that supported both nostr and activitypub.

Also the `proxy` tag isn't indexed by relays default (although it would be great if this requirement was relaxed!)

XX.md
```diff
@@ -0,0 +1,35 @@
...
+- Movie [ISANs](https://en.wikipedia.org/wiki/International_Standard_Audiovisual_Number)
+
+Since the `i` tag is already used for similar references in kind-0 metadata events it makes sense to use it for these content ids as well.
+
+
+## Supported IDs
+
+### Books:
+
+- Book ISBN: `["i", "book:isbn:9780765382030"]` - https://isbnsearch.org/isbn/9780765382030
```
v0l, line 22: I dont think it makes sense to prefix the id with the media, it should just be prefixed with the database type ie `isbn:watever` because in some cases those databases contain all kinds of media

MerryOscar: @v0l yes that's a great point - I've updated the `isbn` and `isan` examples to remove the media prefix - https://github.com/nostr-protocol/nips/pull/1185/commits/d33b223c6afe2be3d197d2cdb4686961a6fb2b29

For the `podcast:guid` case I've left it in because `podcast:guid` is equivalent to `isbn` in denoting the namespace that is used.

v0l: It is useful to have a 2nd level for those other media types, this is what im doing in the Torrent NIP, i have `ttvdb:series:123` and `ttvdb:movies:123` because those sets can overlap, if books or other media overlaps the ids then you basically need this.

MerryOscar: I think for the `isbn` and `isan` there will never be any overlap

MerryOscar:
Fountain is now using these `i` tags to reference podcasts via the `podcast:guid` namespace. Wavlake is also supporting with their audio tracks. You can see an example on Fountain here - https://fountain.fm/show/vnmoRQQ50siLFRE8k061 (see the latest note on the activity tab)

@staab @Semisol @v0l do you see any objection to merging this? (still needs a nip number which I can add)

staab (approved):

arkin0x:
FanFares is using this 👍 @staab

staab:
Great, that's the 2 client threshold, @MerryOscar if you could update to add a NIP number, I think we can merge this.

v0l:
DTAN also uses `i` tags for IMDB links etc https://git.v0l.io/Kieran/dtan/src/commit/dd31d27914902fb025da95b4a84f6b70c939976d/src/nostr-torrent.ts#L200

MerryOscar:
@staab FYI I added a `podcast:publisher:guid` example - both Fountain and Wavlake are using these - https://github.com/nostr-protocol/nips/pull/1185/commits/14c3310039cf7ac8140e6e27f5a7c47ec8f47a30

staab:
Cool, last thing is to change XX to a nip number of your choice

MerryOscar:
@staab I went with NIP-73 as didn't see any open PRs with this number - https://github.com/nostr-protocol/nips/pull/1185/commits/4d9eb32dc9062868d0d61c15456935dad39364de

Happyzippy:
For references to external products GTIN is missing.
This would enable open versions of pricerunner, product review/information sites

Example:
`gtin:3046920028004`

Issue #1183, opened 2024-04-18
Private Nostr


jb55:
I wanted to create a tracking issue that collects some of the approaches for making notes more private, or gives more control over how notes are viewed and by who. I have expressed interest in doing a nostriga talk along these lines so I would love to gather all of the approaches for the talk if I end up doing it.

- https://github.com/nostr-protocol/nips/pull/1029 @vitorpamplona 
- https://github.com/nostr-protocol/nips/pull/1146 @jb55
- https://github.com/nostr-protocol/nips/pull/1030 @fiatjaf 
- https://github.com/nostr-protocol/nips/pull/875 @staab 

Please comment here if you have your own approach you want to add to this list!

jb55:
added!

vitorpamplona:
Looks complete. The only other way would be to use Encrypted Group messages on https://github.com/nostr-protocol/nips/pull/686 but that's quite a privacy overkill for most of these "workgroup" notes. 

You might want to separate encrypted vs non-encrypted stuff for the talk. The number of encrypted options alone can get overwhelming.

vitorpamplona:
I am not sure if this helps, but I used this image in the past to try to map out all of these solutions

![Screenshot 2024-04-18 at 6 08 31 PM](https://github.com/nostr-protocol/nips/assets/532031/f616fc41-9e39-4f3a-925a-7f1466b89ff2)

arthurfranca:
#1083

SnowCait:
I just created [Notestr](https://github.com/SnowCait/notestr) for private notes (PoC). The flow is [here](https://gist.github.com/SnowCait/33f59ef20bae0c075b7c9cf51766dca7#file-privateaccountpossibility-md).
You can try it at https://notestr.pages.dev/example which allows your followees to read your private notes (through ephemeral events tentatively).

I think it would be easier to control private notes if they are all in one server.
The servers should be decentralized and you can use any server as NIP-96.

erskingardner:
Just added https://github.com/nostr-protocol/nips/pull/1206

melvincarvalho:
[...]

Issue #1139, opened 2024-03-27
Long-form publications


alexgleason:
NIP-23 defines long-form content, eg habla.news

These posts have an author. But it should be possible to group posts by publication so you can build something like Substack or Ghost.

As I'm writing this, I realize maybe it's the job of the publication to identify which events should belong in it, rather than for the event to mark itself as part of a publication. Maybe as a new type of list.

vitorpamplona:
Looks like a NIP-72 Community to me :)

alexgleason:
Is that what habla.news is doing here? Seems like a good idea in theory. Not sure they're doing it successfully here.

![image](https://github.com/nostr-protocol/nips/assets/3639540/f85252d4-f5cb-40af-8a70-66e61be7b48f)

staab:
> I realize maybe it's the job of the publication to identify which events should belong in it, rather than for the event to mark itself as part of a publication.

I'd say this is the better approach. Lists, labels, or a special-purpose relay could all be used.

vitorpamplona:
> Is that what habla.news is doing here? Seems like a good idea in theory. Not sure they're doing it successfully here.

Yeah, I don't think their implementation is focused on your own Substack or Ghost communities. It's focused on adding posts to more general communities, like "Linux". It's almost like topics right now.

vitorpamplona:
> Lists, labels, or a special-purpose relay could all be used.

For a single author Substack, yes. For multiauthor, then a Community is needed.

staab:
> For multiauthor, then a Community is needed

Not at all, the publisher would curate the content in any of those ways. NIP 72 communities are no different, since moderators have to authorize posts otherwise everyone would be able to post.

vitorpamplona:
> Not at all, the publisher would curate the content in any of those ways. 

And if you had multiple publishers, the user has to follow to each individually. The point of NIP 72 is to label who to follow so that users don't need to manually go after whoever is the most up-to-date list of publishers of a given substack.

staab:
> up-to-date list of publishers of a given substack

Maybe I don't understand, isn't a "substack" or "blog" a publisher? So even if there are multiple authors, there's only ever a single list, and it's authoritative.

vitorpamplona:
> Maybe I don't understand, isn't a "substack" or "blog" a publisher? So even if there are multiple authors, there's only ever a single list, and it's authoritative.

[...]

Issue #583, opened 2023-06-05
NIP-23 used for private journal


marksftw:
Could NIP-23 be used as a private journal entry? I’m thinking it is encrypted and only readable by the author’s key. This would allow for a Nostr-based journal or diary app. Self custody, no need to store your private thoughts on a cloud service.

markskram:
Combo of NIP-04, NIP-23, and NIP-33?

Encrypted to self, long form, able to update.

PR #1186, opened 2024-04-21
Update NIP-23 to define encrypted long form content

space-shell:
**Overview**

This change updates an existing protocol definition to allow for encrypted long form content.

> Considering that we currently have applications utilizing the event kind `30024` for long form content and to avoid clients accidentally being presented with encrypted strings it may be preferable to replace the `encrypted` tag for another event kind all together e.g. event `30025` to represent encrypted long form content.

**Potential use cases**

 - Article paywalls: utilizing [NIP-46](/nostr-protocol/nips/blob/master/46.md) authors can selectively allow decryption acces to private / premium content
 - Private journaling: utilizing relays to distribute and access private journal entries to be viewed and edited via multiple clients.
 - DVM Publication: users may request that DVM's ( [NIP-90](/nostr-protocol/nips/blob/master/90.md) ) output their content to Nostr as a private article.
 
 **Relates to**
 
 https://github.com/nostr-protocol/nips/issues/1183
 https://github.com/nostr-protocol/nips/issues/1139
 https://github.com/nostr-protocol/nips/issues/583

### DIFF
--- a/23.md
+++ b/23.md
@@ -6,21 +6,19 @@ Long-form Content
 
 `draft` `optional`
 
-This NIP defines `kind:30023` (a _parameterized replaceable event_) for long-form text content, generally referred to as "articles" or "blog posts". `kind:30024` has the same structure as `kind:30023` and is used to save long form drafts.
+This NIP defines `kind:30023` (a _parameterized replaceable event_) for long-form text content, generally referred to as "articles" or "blog posts". `kind:30024` has the same structure as `kind:30023` and is used to save long form drafts. `kind:30025` is used for encryped long form content and also has the same structure as `kind:30023` with the difference being that the content is encrypted via NIP-44 encryption.
 
 "Social" clients that deal primarily with `kind:1` notes should not be expected to implement this NIP.
 
 ### Format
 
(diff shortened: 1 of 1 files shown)


space-shell:
Thanks @vitorpamplona for the quick feedback. I think that the new kind is the better option, I'll make the change to this proposal.

My thinking is that via [NIP-46](https://github.com/nostr-protocol/nips/blob/master/46.md), one could utilize an nsecBunker to manage permissions to articles using a remote key 'owned' by the author.

For private journaling, individuals could encrypt and publish articles to themselves.

fiatjaf:
> Article paywalls: utilizing NIP-46 authors can selectively allow decryption acces to private / premium content

I don't think encryption is a necessary or sufficient condition for sharing premium content.

> Private journaling: utilizing relays to distribute and access private journal entries to be viewed and edited via multiple clients.

This is a larger problem that @vitorpamplona for example is trying to solve on another NIP. Just encryption doesn't solve it.

I personally generally think these things are better done on proprietary platforms or separate protocols outside of Nostr.

> DVM Publication: users may request that DVM's ( NIP-90 ) output their content to Nostr as a private article.

This should be specified as a DVM response format, not on NIP-23.

PR #1189, opened 2024-04-22
Spreadsheets

vitorpamplona:
Adds a simple Spreadsheet format. 

Read [here](https://github.com/vitorpamplona/nips/blob/spreadsheets/73.md)

### DIFF
--- a/73.md
+++ b/73.md
@@ -15,18 +15,21 @@ Event kind `35337` describes a workbook with `data` tags that contain the value 
   "kind": 35337,
   "tags": [
     ["d", "<unique identifier>"],
+    ["title", "Name of this spreadsheet"] // private title
     ["data", "<sheet name>", "<column letter>", "<row number>", "<value>"]
     ["data", "<sheet name>", "<column letter>", "<row number>", "<value>", "<style>"]
     ["style", "<TBD>"] // Need to specify all options here. 
   ],
   "content": nip44Encrypt([
-    ["title", "Name of this topic"] // private title
+    ["title", "Name of this spreadsheet"] // private title
     ["data", "<sheet name>", "<column letter>", "<row number>", "<value>"] // private data
     // ... other tags
   ]),
   // ...
 }
 ``` 
+
+Sheets MUST be rendered in the order of their tags. 
 
 ## Example
 
--- a/73.md
+++ b/73.md
@@ -1,0 +1,48 @@
+NIP-73
+======
+
+Spreadsheets
+------------
+
+`draft` `optional`
+
+This NIP provides a simple way to save spreedsheets on Nostr. 
+
+Event kind `35337` describes a workbook with `data` tags that contain the value of each cell as well as optional styling. 
+
+```js
+{
+  "kind": 35337,
+  "tags": [
+    ["d", "<unique identifier>"],
+    ["data", "<sheet name>", "<column letter>", "<row number>", "<value>"]
+    ["data", "<sheet name>", "<column letter>", "<row number>", "<value>", "<style>"]
+    ["style", "<TBD>"] // Need to specify all options here. 
+  ],
+  "content": nip44Encrypt([
+    ["title", "Name of this topic"] // private title
+    ["data", "<sheet name>", "<column letter>", "<row number>", "<value>"] // private data
+    // ... other tags
+  ]),
+  // ...
+}
+``` 
+
+## Example
+
+```json
+{
+  "id": "32360b52b11616ea331aacac516494e36bd4079d8908edc8f26ad1e4acab5a53",
+  "kind": 35337,
+  "tags": [
+    [ "d", "SheetStr Demo" ],
+    [ "data", "Sheet1", "J", "25", "3" ],
+    [ "data", "Sheet1", "J", "26", "5" ],
+    [ "data", "Sheet1", "J", "27", "=SUM(J25:J26)" ]
+  ],
+  "created_at": 1713819120,
+  "content": "",
+  "pubkey": "460c25e682fda7832b52d1f22d3d22b3176d972f60dcdc3212ed8c92ef85065c",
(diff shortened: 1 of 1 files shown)


73.md
```diff
@@ -0,0 +1,51 @@
...
+This NIP provides a simple way to save spreedsheets on Nostr. 
+
+Event kind `35337` describes a workbook with `data` tags that contain the value of each cell as well as optional styling. 
+
+```js
+{
+  "kind": 35337,
+  "tags": [
+    ["d", "<unique identifier>"],
+    ["title", "Name of this spreadsheet"] // private title
```
theborakompanioni, line 18:
```suggestion
    ["title", "Name of this spreadsheet"] // public title
```

73.md
```diff
@@ -0,0 +1,140 @@
...
+  ], keyPair.privateKey, keyPair.publicKey),
+  // ...
+}
+``` 
+
+To decrypt, receivers SHOULD find the ciphertext for their key, decrypt that to get the shared private key and use the shared private key to decrypt the `.content`.
+
+Clients SHOULD include the author as a `p` tag to save the secret that will allow the author to decrypt and modify the event further. 
+
+## Sharing Encrypted Spreadsheets with Editor permission.
```
abhay-raizada, line 108: no way to share public sheets with editor permissions?

vitorpamplona: We can, just use the same model of singing with the secret and putting the secret on p-tags of the authorized users, but put all the data on the public tags.

abhay-raizada: so the content on the shared editabale public sheets are not encrypted? isn't clear in the spec. Also in regular public sheets the sheet data is in the tags field, but in an shared editable public sheet the sheet data is in the content key? doesn't seem right

vitorpamplona: Think like this: 
- You can put content publicly on the "data" tag or privately on the `.content`. That defines who can see it. 
- You can sign the event with the secret or with the user's main key. If signed with the user's, no one else can change it. If signed with the secret, everyone with the secret can change it.

vitorpamplona: It's just weird to create UI/UX where you have both private and public data at the same time. But one could.

abhay-raizada: okay i think i get it, so in the case of public shared forms, the public data is in the tags field, as well as the p-tags? would the order matter in this case (the order of the p-tags relative to data tags)

vitorpamplona: Order doesn't matter, but the p-tags must be public otherwise people won't have a way to query for events that they can view/change from relays.

73.md
```diff
@@ -0,0 +1,149 @@
...
+
+```json
+{
+  "id": "32360b52b11616ea331aacac516494e36bd4079d8908edc8f26ad1e4acab5a53",
+  "kind": 35337,
+  "tags": [
+    [ "d", "SheetStr Demo" ],
+    [ "data", "Sheet1", "J", "25", "3" ],
+    [ "data", "Sheet1", "J", "26", "5" ],
+    [ "data", "Sheet1", "J", "27", "=SUM(J25:J26)" ]
```
abhay-raizada, line 40: are these formula's standardized? also what about custom formulas, are they out of scope for the nip?

73.md
```diff
@@ -0,0 +1,190 @@
...
+val viewingKeyPair = nostr.generateKeyPair()
+
+{
+  "pubkey": edittingKeyPair.publicKey
+  "kind": 35337,
+  "tags": [
+    ["d", "<unique identifier>"]
+    ["p", "<pubkey 1>", "<relay url>", nip44Encrypt(edittingKeyPair.privateKeyHex, edittingKeyPair.privateKey, "<pubkey 1>") ]
+    ["p", "<pubkey 2>", "<relay url>", nip44Encrypt(edittingKeyPair.privateKeyHex, edittingKeyPair.privateKey, "<pubkey 2>") ]
+    ["p", "<pubkey 3>", "<relay url>", nip44Encrypt(viewingKeyPair.privateKeyHex,  edittingKeyPair.privateKey, "<pubkey 3>") ] // view only
```
arthurfranca, line 138: I see 2 problems here:
1) Editor doesn't know if there is a viewing key, so attempting to decrypt content will fail if there is a viewer
2) Editor doesn't know the viewing key when there is one, so editor can't add new viewers nor decrypt content

Although cumbersome, editor does could decrypt all `p` tags till reaching the third one and finding the viewingKeyPair.privateKeyHex. Though if an editor removes all viewers without re-encrypting the content, the content may be lost (cause the viewing key would be gone) so we also need an easy way to know how many viewers exist so to know when to re-encrypt the content.

One way of dealing with all of this is adding "Editor" and "Viewer" markers to the p tag. Here markers make no harm afaik.

arthurfranca:
It may be a good idea to choose a list to hold all sheet references the user is interested in, because one can't rely on authors filter and anyone can add the user to a `p` tag on a random sheet.

73.md
vitorpamplona: > Editor doesn't know if there is a viewing key

They have to decrypt all p-tags to see if there is a viewing key. 

> Editor doesn't know the viewing key when there is one

Editors can decrypt the viewing key. 

>  "Editor" and "Viewer" markers

There is no way to enforce that these markers actually represent what they are supposed to. So, it's better to just decrypt all tags.

vitorpamplona:
> It may be a good idea to choose a list to hold all sheet references the user is interested in

Yep, I was thinking about an unbound list that just points to all "accepted" spreadsheets.

73.md
```diff
@@ -0,0 +1,190 @@
...
+Event kind `35337` describes a workbook with `data` tags that contain the value of each cell as well as optional styling. 
+
+```js
+{
+  "kind": 35337,
+  "tags": [
+    ["d", "<unique identifier>"],
+    ["title", "Name of this spreadsheet"], // public title
+    ["data", "<sheet name>", "<column letter>", "<row number>", "<value>"],
+    ["data", "<sheet name>", "<column letter>", "<row number>", "<value>", "<style>"],
```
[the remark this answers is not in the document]
vitorpamplona: The issue is how to not break the map between these names and the formulas, since formulas use the names instead of indexes... If one client wants to start with A while others want to start with a letter of their alphabet, they wouldn't be able to translate indexes without mapping back to A first and then translating the formula.

PR #1191, opened 2024-04-23
Add ontolo to nip 32

staab:


staab:
@erskingardner

erskingardner (approved):

staab:
We can add a table when it's needed. Namespace and url are both there, and the point of linking is to not have to put the categories into the NIP (ontolo doesn't currently have a list of categories).

PR #1192, opened 2024-04-23
Shared replaceables

vitorpamplona:
The idea here is to add replaceable events that can be changed only by a list of pre-defined keys without changing relay implementations and without using a separate key to sign the event on behalf of the owners, like how https://github.com/nostr-protocol/nips/pull/1015 suggested. 

Imagine a Google Doc/Spreadsheet that is constantly being updated by a list of users. 

Read [here](https://github.com/vitorpamplona/nips/blob/shared-replaceables/68.md)

### DIFF
--- a/68.md
+++ b/68.md
@@ -1,0 +1,86 @@
+NIP-68
+======
+
+Shared Replaceables
+-------------------
+
+`draft` `optional`
+
+This NIP establishes the encoding mechanism to create, filter and reference replaceable events whose ownership are shared among several keys. 
+
+The initial creator adds all owners to the `d` tag, separated by `|`. The presence of `|` with valid keys for each but the last position, directs clients to work with this event as a shared replaceable. 
+
+```js
+{
+  "kind": 3xxxx,
+  "tags": [
+    ["d", "<pubkey1>|<pubkey2>|<pubkey3>|<identifier>"],
+    // ...
+  ],
+  //...
+}
+```
+
+To change owners, users MUST create a new `d` tag with new owners and MAY delete the previous one. 
+
+# Referencing & Indexing
+
+Events that reference the most recent version of a shared replaceable MUST add `a` tags to each owner. 
+
+Reply example:
+
+```jsonc
+{
+  "kind": 1,
+  "tags": [
+    ["a", "3xxxx:<pubkey1>:<pubkey1>|<pubkey2>|<pubkey3>|<identifier>", "<optional-relay-url>", "reply"],
+    ["a", "3xxxx:<pubkey2>:<pubkey1>|<pubkey2>|<pubkey3>|<identifier>", "<optional-relay-url>", "reply"],
+    ["a", "3xxxx:<pubkey3>:<pubkey1>|<pubkey2>|<pubkey3>|<identifier>", "<optional-relay-url>", "reply"]
+  ],
+  "content": "a reply",
+  // ...
+}
+```
+
+## Deletion
+
+Owners can only unilaterally delete their own events: 
+
+```jsonc
+{
+  "pubkey": "<pubkey1>",
+  "kind": 5, 
+  "tags": [
+    ["a", "3xxxx:<pubkey1>:<pubkey1>|<pubkey2>|<pubkey3>|<identifier>", "<optional-relay-url>"],
+  ],
+  "content": "Deletes my versions of this repleceable",
+  // ...
+}
+```
+
+Adding other pubkeys to the `a` tag won't have any effect. 
+
+# Filtering
+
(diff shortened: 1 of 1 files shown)


fiatjaf:
I feel like it would be much simpler for everybody involved if this was done with a shared private key or a FROST key.

vitorpamplona:
> I feel like it would be much simpler for everybody involved if this was done with a shared private key or a FROST key.

I went there but I was way more complicated to manage the key than I initially thought. Each Doc with a different group creates a key, which then the user has to manage all of these keys. It's a lot of work. 

And you lose track of who did what change in the doc. Of course, you can have a separate system that takes a signed payload from each user and resigns as the shared key, so that you can keep track of the changes, but that is even more work.

vitorpamplona:
Frankly, I wish we had a way to build 1 of N signatures, in the sense that the `.pubkey` in Nostr is the sum of all owner's pubkeys, but the signature comes from just one of the owners. It would be perfect for replaceable events.

fiatjaf:
> I went there but I was way more complicated to manage the key than I initially thought. Each Doc with a different group creates a key, which then the user has to manage all of these keys. It's a lot of work.

I see, it makes sense that this would be a hassle.

> Frankly, I wish we had a way to build 1 of N signatures, in the sense that the .pubkey in Nostr is the sum of all owner's pubkeys, but the signature comes from just one of the owners. It would be perfect for replaceable events.

I'm pretty sure FROST can do that. But you wouldn't know who signed.

frbitten:
You could create a new kind group for events that need to keep a history of all changes.  In the case of kind 3XXXX, only the last change would be kept.

vitorpamplona:
> I'm pretty sure FROST can do that. But you wouldn't know who signed.

How do we test this?

DanConwayDev:
Here's an alternative approach.
Instead of including the pubkeys in the `d` identifier, include them as an array in the in an `owners` tag.

Benefits:
1) allows new owners to be added and old owners to be removed (or remove themselves)
2) maintains signing transparency (who signed what update unlike FROST)
3) allows users to decide which owner they trust and always use their 'owners' list.
4) allows forks: alice and bob are working on a thing. carole forks it by using the same `d` identifier but only listing herself in the `owners` tag

this is how the `maintainers` tag works in nip34

fiatjaf:
> How do we test this? Do we know anyone with experience on FROST to create some Nostr events? :)

Haven't you seen https://github.com/nickfarrow/frostr? I thought everybody knew about this. Some people played with it at the time, but it was a very manual process. I couldn't myself. So not great, but it was just an experiment. Still it proved it was possible.

vitorpamplona:
> Haven't you seen https://github.com/nickfarrow/frostr?

Oh yeah, I remember seeing that. At the time, there were a lot of questions and people weren't sure if the way it was coded was secure. Do you know if anyone is using that version? Also, should we store all those FROST secrets inside Nostr? And can we reuse those secrets into many keys? (many shareables in our case here).

vitorpamplona:
> Instead of including the pubkeys in the d identifier, include them as an array in the in an owners tag.

The issue there is the mess created by intermittent relays while validating the updates to the `owner` tags. For instance: 

1. UserA creates a replaceable with UserB as owners. 
2. UserB updates the replaceable and adds UserC as owners. 
3. UserC updates the replaceable with new changes. 

Now if the event 2 is unavailable at the time (relay is busy, UserB deleted it or something else), the client has a problem that UserC added himself to the replaceable. There is no way for the client to verify that UserC was authorized by previous owners to update the record and the client SHOULD just reject UserC's change because the same state would exist if UserC is an attacker and is inserting a phishing link to rug people. 

Nostr Replaceables do not allow owner change. This PR would be similar with the only difference that it would be multiple owners and not a single one.

fiatjaf:
@nickfarrow can you enlighten us?

vitorpamplona:
> As in, posted within events? I guess could be done if encrypted and you trust relays to persist them..

I am thinking on a regular kind 10xxx event that holds FROST public keys as tags and private secrets as a NIP-44-encrypted JSON-stringified tag array and in the `.content` of the event. So that everyone knows your FROST public nonce to allow them to create shared-events with you without reaching out to you every time. But I am not sure if reusing the public key so much reduces security in the long run. 

> has no accountability for who signed

We could add a `signed by` tag that has the real pubkey of the author and a signature of the hash of the (kind + FROST pubkey + d-tag + created at of the FROST event).

vitorpamplona:
@fiatjaf testing the idea of using the old #357 model as a shared replaceable for spreadsheets. https://github.com/nostr-protocol/nips/pull/1189 

It might actually work quite well. The shared key is kept in each version of the replaceable.

PR #1194, opened 2024-04-24
Require tags to have at least one string

### LINKED ISSUE #1193: Empty tag
Should empty tag be allowed in events? Like:

```json
{
      "id": "11f22b4683f0aaf7060778cfafd2d62ae34648e43029bc75ee9e1bcd19366849",
      "pubkey": "97c70a44366a6535c145b333f973ea86dfdc2d7a99da618c40c64705ad98e322",
      "created_at": 1712189994,
      "kind": 10002,
      "tags": [
        [], // <==
        [
          "client",
          "Coracle",
          "31990:97c70a44366a6535c145b333f973ea86dfdc2d7a99da618c40c64705ad98e322:1685968093690"
        ]
      ],
      "content": "",
      "sig": "a4b3a034963702e7b8bb514dec692741fdd00ba41d4d4cb73decb753b2c226097ce8d0d66a25dea2896d06a5b981c4eeec498a29fe70ff21193ad1d995b1ccd8"
}
```

In `rust-nostr` libs empty tags aren't allowed, the same in `strfry` and `chorus` relays. But relays like `nostr-rs-relay` allow them.

Currently NIP01 specify this for tags: "Each tag is an array of strings of arbitrary size". So apparently empty tag should be allowed (?)

Ref https://github.com/rust-nostr/nostr/issues/411

Giszmo:
fixes #1193

### DIFF
--- a/01.md
+++ b/01.md
@@ -56,7 +56,7 @@ To prevent implementation differences from creating a different event ID for the
 
 ### Tags
 
-Each tag is an array of strings of arbitrary size, with some conventions around them. Take a look at the example below:
+Each tag is an array of one or more strings, with some conventions around them. Take a look at the example below:
 
 ```jsonc
 {


mikedilger:
I don't read this change as requiring you to block them.  Some software already errors when these are present.  Just don't create them.

fiatjaf:
Why would anyone create them except in case of a bug? Who is creating them?

fiatjaf:
I don't know, I think it's a bad idea to clutter the spec with a bunch of useless recommendations that are obvious to anyone. Makes things harder to read for no gain.

staab:
I made [the one cited](https://github.com/nostr-protocol/nips/issues/1193#issue-2260849525). Probably a bug, I almost always use Coracle's dev branch.

fiatjaf:
OK, so you go to jail.

Giszmo:
So I would close this and resolve #1194 accordingly, too?

staab:
I'm ok with merging this, it's a slight improvement and guideline for publishers, even if it doesn't prevent people from breaking receiver assumptions

staab (approved):

basantagoswami:
> Makes things harder to read for no gain.

I think in this case, it rather makes the spec a bit easier to read, at least for me

PR #1196, opened 2024-04-24
Nip05 Signed Metadata

trbouma:
This NIP proposes to add signed metadata to a NIP-05 request. This could be a new nip or added to NIP-05


staab:
What does this accomplish that regular kind 0's don't? I wouldn't want my NIP 05 provider to be providing metadata about me that isn't signed.

vitorpamplona:
Is the goal to let the nip05 provider describe who the user is or for users to describe themselves in these fields?

trbouma:
> Is the goal to let the nip05 provider describe who the user is or for users to describe themselves in these fields?

It really depends on the nip 05 provider. For example, I might want to provide my profile information via nip-05 provider who then can emit a kind 0 note on my behalf, or, when a nip 05 request is mad,e use the additional metadata information for some specific purpose.

trbouma:
> What does this accomplish that regular kind 0's don't? I wouldn't want my NIP 05 provider to be providing metadata about me that isn't signed.

Yeah, I agree with that. Profile information should not be updated if is not signed by the user. This approach allows for additional information that can be provide that might not be related to the user, but rather about the NIP-05 provider. The kind 0 format is a generic way to structure this information, but doesn't necessarily mean that a kind 0 event will be transmitted.

vitorpamplona:
I am not sure if I understand the benefits of making it this general/insecure. Is anyone you know trying to use this? Which version would they want to implement?

trbouma:
You can see a live example here:

https://staging.nimo.cash/.well-known/nostr.json/?name=trbouma

vitorpamplona:
Sure, but who is reading the metadata from the json and how are they presenting on their screens if not signed by users themselves?

fiatjaf:
Return a relay URL in the nostr.json file, tell clients to connect to that and fetch kind 0.

trbouma:
> Sure, but who is reading the metadata from the json and how are they presenting on their screens if not signed by users themselves?

@vitorpamplona - It's really intended for parties that exist outside of the nostr ecosystem to find info. Originally I was going to do a W3C did:web implementation, but realized it would be just as easy to extend NIP-05 first.

trbouma:
> I am not sure if I understand the benefits of making it this general/insecure. Is anyone you know trying to use this? Which version would they want to implement?

One of the use cases is for automated nostr accounts, or accounts that are used by services to backend communicate with other services via relays. A decentralized push notification service for apps, for example. An app instance might generate its own npub/nsec and use an app-specific relay. The additional metadata could help the app to query other app-generated npubs.

PR #1199, opened 2024-04-24
nip34: simplify `r` tag for earliest unique commit

fiatjaf:


fiatjaf:
@DanConwayDev

DanConwayDev:
ACK
Nice simplification. This works with gitworkshop.dev and ngit.

DanConwayDev:
I was just implementing the `euc` marker on ngit.
`rust-nostr` assumes a tag reference with a marker is a relay tag. It throws a `invalid relay metadata: euc` error. @yukibtc
Might other clients / libraries make a similar assumption. 
So long as the error is handled well this shouldn't break anything. I thought I'd raise it in any case.

yukibtc:
> I was just implementing the `euc` marker on ngit.
> `rust-nostr` assumes a tag reference with a marker is a relay tag. It throws a `invalid relay metadata: euc` error. @yukibtc
> Might other clients / libraries make a similar assumption?
> So long as the error is handled well this shouldn't break anything. I thought I'd raise it in any case.
> 
> 

Some days ago I've rewritten the `Tag` in `rust-nostr`, it will no longer throws these type of errors from v0.31

PR #1200, opened 2024-04-25
Resumes

vitorpamplona:
Adds kind 11000 as the pubkey's resume. It uses a structure format that Clients can render in any way they see fit. 

Read [here](https://github.com/vitorpamplona/nips/blob/resume/55.md)

### DIFF
--- a/55.md
+++ b/55.md
@@ -8,8 +8,6 @@ Structured Resumes
 
 This NIP defines event kind `11000` as the pubkey's resume. Life acomplishments are described as tags and a mini bio is written in `.content`. 
 
-
-
 ```jsonc
 {
   "kind": 11000,
--- a/55.md
+++ b/55.md
@@ -1,0 +1,71 @@
+NIP-55
+======
+
+Structured Resumes
+------------------
+
+`optional`
+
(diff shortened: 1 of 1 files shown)


alltheseas:
It seems to me a strength of nostr lacking in this proposal is referrals/recommendations/WOT. 

For instance, I would refer @robagreda as a UI/UX/design professional from our mutual experience working together, and I would publicly sign this referral
using my keys. 

cc @fr4nzap

vitorpamplona:
We could use NIP-32 labels from third parties for endorsements on the skill level. 
+ a new event kind to serve as a testimonial/recommendation

tyiu:
Resume or CV?

vitorpamplona:
> Resume or CV?

No idea. Both?

alltheseas:
what is the difference?

thibistaken:
This looks quite interesting. I built https://plebwork.com/ a marketplace for companies to work with bitcoin talent, and we've been looking into adding Nostr to improve the product experience. When it comes to referrals, endorsements and reviews, marketplaces generally build walled gardens to increase switching costs, so this would be quite useful obviously. What do you think is the next phase for this NIP? Curious to see if we can help in any way we can based on what we're building.

Issue #1198, opened 2024-04-24
Paths and Queries in Relay URLs


mikedilger:
A relay is defined by its websocket server, which means that the protocol, host and port of a URL matter in differentiating relays.

Path and query do not determine which websocket to connect to, they are submitted after connection.

Two different URLs with the same (protocol + host + port) point at the same relay, but may trigger different relay-specific behavior when addressed with a specific path and/or query.

I find this very difficult to deal with. Nostr spec refers to relay URLs without any reference to path and/or query.  I've presumed different URLs are different relays, even if they only differ in path or query.  But this has caused problems:

- Too many connections to the very same host/relay, giving "429 Too Many Requests"
- Fetching many times from the same host/relay over multiple websocket connections
- Posting the same event many times to the same host/relay
- In gossip it is very difficult for users to manage hundreds of variations of the same relay in the relay pages. I want to set rank=0, but there are far too many pages I have to do this on, and if a new npub path comes up it won't apply to those.

If however I don't preserve the path and query, then the relay doesn't get the benefit of having the path and query (whatever functionality it was intending to provide by having it).

Before making massive sweeping changes to gossip I'd like to hear what others think.

staab:
Interesting question. Even if there are valid use cases for paths/query strings, a client can't tell if these are minor configuration options, or consist of an entirely separate data set. Separate thought, attackers could just generate lost of variations of a single relay url and DOS the relay (and a user's client potentially).

My first thought would be to prohibit paths or query strings and say clients MUST normalize to just the domain name. This wouldn't preclude relays from parameterizing their url via subdomain, but failures would happen at the DNS level, rather than what currently happens with relays which accept connections at any path. This would of course break current implementations like nostr.wine.

This seems mostly like a relay problem. Relays should reject connections on anything other than a valid, meaningful path. If a client is connected to the same relay twice, the second connection should be rejected (unless the path indicates a different "virtual relay").

fiatjaf:
Path differentiation is crucial for any relay who wants to offer different services in the same process, like different filtering options, without having to bloat the protocol with unsupported flips and switches.

A simple example:

 - `wss://relay.com/` -> returns all notes
 - `wss://relay.com/pt` -> returns only notes in Portuguese
 - `wss://relay.com/jp` -> returns only notes in Japanese

staab:
Why not `wss://jp.relay.com`? I'm not saying this is the right solution, but you don't necessarily need paths.

[...]

PR #1201, opened 2024-04-25
nip11: "virtual_paths"

fiatjaf:
https://github.com/nostr-protocol/nips/issues/1198#issuecomment-2077040526

This is just a stupid starting point just to get something concrete, please change it before merging @mikedilger @v0l @staab @nostrband

### DIFF
--- a/11.md
+++ b/11.md
@@ -60,6 +60,30 @@ The relay MAY choose to publish its software version as a string attribute.  The
 Extra Fields
 ------------
 
+### Virtual Relays
+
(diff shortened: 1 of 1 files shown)


fiatjaf:
As far as I know filter.nostr.wine has paths for each pubkey and relay.nostr.band has paths for nip32-labeled content.

I'm working on a relay with the sole purpose of having virtual relays of multiple types with a shared database. I was thinking of putting these in subdomains, but maybe paths are better.

staab:
> As far as I know filter.nostr.wine has paths for each pubkey and relay.nostr.band has paths for nip32-labeled content.

I know about those, I'm more wondering what an example of a path/qp that doesn't indicate a virtual relay would be.

v0l:
lgtm

PR #1202, opened 2024-04-25
nip46: signer should fill in `pubkey`, `id` and `sig` on `sign_event`

fiatjaf:
This is how it works on NIP-07 and should be how it works on NIP-46 too.


fiatjaf:
@pablof7z @staab @mikedilger @nostrband

staab (approved):

mikedilger:
gossip also allows you to elide the created_at field (it will use the current time), but we don't have to put that in the spec.

mikedilger (approved):

nostrband:
ack

nostrband (approved):

PR #1205, opened 2024-04-27
Speedrunning Run Submission Event

BlueZTV:
Draft event for submitting a speedrun.

Information: https://github.com/BlueZTV/nips/blob/master/76.md

### DIFF
--- a/76.md
+++ b/76.md
@@ -12,7 +12,7 @@ Each game's leaderboard has optional Categories, Variations, Platform, Level, Re
 ```jsonc
 {
   "kind": 7602,
-  "d", "<Uniquie Identifier of the game's leaderboard>", //E.g. Among Us' Steam App ID is 945360
+  "d", "<Unique Identifier for the game>", //E.g. Among Us' Steam App ID is 945360
   "tags": [
     ["c", "<Category Identifier the submission is for>", <Long Name>], //E.g. A%G (Any% Glitchless), AlAc (All Achievements)
     ["v", "<Variation of the category the submission is for>", <Long Name>], //E.g. for Celeste AnyP (Any%), ARB (All Red berries), TE (True Ending) etc.
--- a/76.md
+++ b/76.md
@@ -6,8 +6,8 @@ Speedrunning Leaderboard Entry
 
 `draft` `optional`
 
-This NIP defines event kind `7602` as an entry towards a Game's speedrunning leaderboard, specified via a `d` tag, e.g. a game's Steam App ID. 
-Each game's leaderboard has optional Categories, Variations, Platform, Level, Region and Versions to create alternate leaderboards for one game.
+This NIP defines event kind `7602` as a submission to a game's speedrunning leaderboard, specified via a unique `d` tag, e.g. a game's Steam App ID. 
+Each game's leaderboard has optional Categories, Variations, Platform, Level, Region and Versions tags to allow for the creation of alternate sub-leaderboards under one game.
 
 ```jsonc
 {
@@ -19,6 +19,7 @@ Each game's leaderboard has optional Categories, Variations, Platform, Level, Re
     ["p", "<Platform identifier>"], //E.g. N64, EMU
     ["reg", "<3 Letter ISO Country Code>"], //E.g. JPN, AUS
     ["lvl", "<Level>"], //E.g. FG, W1L1, F3OS
+    ["ver", "<Version>"], //Version the submission is ran on e.g. 1.4.0.0
     ["rlt", "<Start Unix Time>", "<Stop Unix Time>"], //Real Life Time, subtract Start from Stop to get actual time
     ["igt", "<Start Unix Time>", "<Stop Unix Time>"] //In Game Time
   ],
--- a/76.md
+++ b/76.md
@@ -22,8 +22,9 @@ Each game's leaderboard has optional Categories, Variations, Platform, Level, Re
     ["ver", "<Version>"], //Version the submission is ran on e.g. 1.4.0.0
     ["rlt", "<Start Unix Time>", "<Stop Unix Time>"], //Real Life Time, subtract Start from Stop to get actual time
     ["igt", "<Start Unix Time>", "<Stop Unix Time>"], //In-Game Time
-    ["proof": "<Link to proof e.g. Twitch Vod>"]
+    ["proof": "<Link to proof>"] //e.g. Twitch Vod, replay file
   ],
+  "content": "<Description for run>", //User written comment about run if desired
   "created_at": "<Unix Time>",
   "pubkey": "<Public Key>",
   "sig": "<signature>"
@@ -43,6 +44,7 @@ An example of a potential Super Mario 64 120 Star N64 run submission.
     ["rlt", "1587370820", "1587376601"],
     ["proof": "https://www.twitch.tv/videos/2034126985"]
   ],
+  "content": "WR run",
   "created_at": "1714228879",
   "pubkey": "1a6ea253df444f434417bdcaf9d4aef75f00921965e1afd836da0ae9cd1a0e074cb3490bb4cc26ead75af",
   "sig": "076b3405407f26b87ff9333fa336c1c6fbdfc27f2f"
--- a/76.md
+++ b/76.md
@@ -1,0 +1,66 @@
+NIP-76
+======
+
+Speedrunning Leaderboard Entry
+------------------
+
+`draft` `optional`
+
+This NIP defines event kind `7602` as an entry towards a Game's speedrunning leaderboard, specified via a `d` tag, e.g. a game's Steam App ID. 
+Each game's leaderboard has optional Categories, Variations, Platform, Level, Region and Versions to create alternate leaderboards for one game.
+
+```jsonc
+{
+  "kind": 7602,
+  "d", "<Uniquie Identifier of the game's leaderboard>", //E.g. Among Us' Steam App ID is 945360
+  "tags": [
+    ["c", "<Category Identifier the submission is for>", <Long Name>], //E.g. A%G (Any% Glitchless), AlAc (All Achievements)
+    ["v", "<Variation of the category the submission is for>", <Long Name>], //E.g. for Celeste AnyP (Any%), ARB (All Red berries), TE (True Ending) etc.
+    ["p", "<Platform identifier>"], //E.g. N64, EMU
+    ["reg", "<3 Letter ISO Country Code>"], //E.g. JPN, AUS
+    ["lvl", "<Level>"], //E.g. FG, W1L1, F3OS
+    ["rlt", "<Start Unix Time>", "<Stop Unix Time>"], //Real Life Time, subtract Start from Stop to get actual time
+    ["igt", "<Start Unix Time>", "<Stop Unix Time>"] //In Game Time
+  ],
+  "proof": "<Link to proof e.g. Twitch Vod>",
+  "created_at": "<Unix Time>",
+  "pubkey": "<Public Key>",
+  "sig": "<signature>"
+}
+```
+Tags can be entered multiple times if required such as a game having multiple varieties under one category, e.g. Pizza Tower Full Game Any% has Playable Character and Glitches? varieties.
+
+An example of a potential Super Mario 64 120 Star N64 run submission.
+```jsonc
+{
+  "kind": 7602,
+  "d", "SM64",
+  "tags": [
+    ["c", "120S", "120 Star"],
+    ["p", "N64"],
+    ["reg", "JPN"],
+    ["rlt", "1587370820", "1587376601"],
+  ],
+  "proof": "https://www.twitch.tv/videos/2034126985",
+  "created_at": "1714228879",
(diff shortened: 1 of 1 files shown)


76.md
```diff
@@ -0,0 +1,67 @@
...
+
+`draft` `optional`
+
+This NIP defines event kind `7602` as a submission to a game's speedrunning leaderboard, specified via a unique `d` tag, e.g. a game's Steam App ID. 
+Each game's leaderboard has optional Categories, Variations, Platform, Level, Region and Version tags to allow for the creation of alternate sub-leaderboards under one game.
+
+```jsonc
+{
+  "kind": 7602,
+  "d", "<Unique Identifier for the game>", //E.g. Among Us' Steam App ID is 945360
```
alexgleason, line 15: Is there any standard for globally unique game IDs? 

I like this idea a lot.

76.md
```diff
@@ -0,0 +1,67 @@
...
+    ["c", "<Category Identifier the submission is for>", <Long Name>], //E.g. A%G (Any% Glitchless), AlAc (All Achievements)
+    ["v", "<Variation of the category the submission is for>", <Long Name>], //E.g. for Celeste AnyP (Any%), ARB (All Red berries), TE (True Ending) etc.
+    ["plt", "<Platform Identifier>"], //E.g. N64, EMU
+    ["reg", "<3 Letter ISO Country Code>"], //E.g. JPN, AUS
+    ["lvl", "<Level>"], //E.g. FG, W1L1, F3OS
+    ["ver", "<Version>"], //Version the submission is ran on e.g. 1.4.0.0
+    ["rlt", "<Start Unix Time>", "<Stop Unix Time>"], //Real Life Time, subtract Start from Stop to get actual time
+    ["igt", "<Start Unix Time>", "<Stop Unix Time>"] //In-Game Time
+  ],
+  "proof": "<Link to proof e.g. Twitch Vod>",
```
fiatjaf, line 26: "proof" should be a tag

76.md
fiatjaf: "d" should be a tag

76.md
BlueZTV: My bad, I've moved it in the tags.

76.md
BlueZTV: > Is there any standard for globally unique game IDs?
> 
> I like this idea a lot.

Cheers, appreciate the comment.
Regarding your question about Game IDs, I'm unsure whether you're asking about a Unique Game ID standard in general or if I had a specific standard in mind. For a Game ID system that exists in general, other than implementations for specific platforms (such as [Dolphin Emu's Nintendo GameID system](https://wiki.dolphin-emu.org/index.php?title=GameIDs)) or auto-increment systems made by individual sites ([ESRB Example](https://www.esrb.org/ratings/38140/), [Steam App ID's](https://store.steampowered.com/app/945360/), [The GamesDB IDs](https://thegamesdb.net/game.php?id=96733) and [Roblox's Game ID's](https://www.roblox.com/games/6516141723/)) there doesn't seem to be an existing standard in place. At least not one I could find that works across all platforms, stores and generations. 
If you were asking whether I had a particular standard in mind, I didn't really, beyond Platform ID's (e.g. Steam and Roblox App IDs), commonly accepted community names for games and clients displaying the Game ID somewhere so it's easily accessible for other implementations, sorry for the half-baked initial idea. 

I've somewhat thought of a potential system:
- The first 6 characters could be either the first 2 characters from the first 2 words of the title and the first 2 characters of the last word in the title, the first 3 letters from each word if it's only 2 words long or the first 6 characters of the title if it's only one word, If the title is one word and less than or equal to 6 characters, it'll just be the title. All capitalised.
- Game Number/Version, 0 if not specified.
- Release year for the initial release in the country with the earliest release (Month could be added too, I wasn't able to find an example that required it though).

For example, Five Nights at Freddy's would be FINIFR02014 and Five Nights at Freddy's 2 would be FINIFR22014 as they were released the same year. [Spider-Man (1982)](https://en.wikipedia.org/wiki/Spider-Man_(1982_video_game)) would be SPIMAN01982 and [Spider-Man (2000)](https://en.wikipedia.org/wiki/Spider-Man_(2000_video_game)) would be SPIMAN02000. [Spider-Man 2: Enter: Electro (2000)](https://en.wikipedia.org/wiki/Spider-Man_2:_Enter:_Electro) would be SPMAEL22000 and [Spider-Man 2: The Sinister Six (2000)](https://en.wikipedia.org/wiki/Spider-Man_2:_The_Sinister_Six) would be SPMASI22000.

I'm very new to nostr development, so if there are any suggestions for an actual/improved standard, I'm completely open to suggestions.

BlueZTV: Thanks for revisiting this. Steam URIs are definitely a thing and could work well for games that are on Steam. For games that aren't on Steam, I feel the URL solution has some issues that aren't complete show stoppers but could cause confusion in the long term.
<details><summary>1. URL preferences</summary>
<p>
A lot of games speedran are older console games and not available on modern platforms. A public tiered preference/ranking system for domains/schemes would somewhat mitigate this but it's still possible someone when making a client/mod to submit runs uses the wrong URL e.g. The game's Wiki instead of Wikipedia. This would also happen less as the system becomes more popular and there are multiple sources for Game URLs. Generating a tiered ranking system for URLs also has it's own troubles, the language would have to be very clear and leave little room for confusion. A URL system, with the exception of games on Steam if it has the highest priority, would be a first in best dressed in terms of picking which URL represents that game.
</p>
</details>  
<details><summary>2. Based on centralized sources</summary>
<p>
Highly unlikely to happen but it's possible Steam or other platforms used for URLs could change their systems or shut down. It wouldn't outright break the Game ID system but it may cause confusion and is more likely to lead to split leaderboards.
I do realise that by creating a standard in a NIP that is also a form of centralisation, however, a custom naming standard not reliant on outside sources, once proven to be valid and reliable can be cemented and shared outside of its initial sphere of influence without fear of change.
If a game were to get removed off platforms or banned in a certain country (unfortunately happens a surprising amount in Australia) it could cause confusion. A recent example is Helldivers 2 getting delisted from Steam in countries without PSN, I'm sure it wasn't completely gone from the site's backend but it's still a possibility to consider.
</p>
</details> 
<details><summary>3. Barrier to entry</summary>
<p>
Just potentially an issue if someone develops an indie game and publishes a few speedruns amongst their friend group with a random Game ID as they don't yet have a website or haven't published to steam, if they do decide to publish and it becomes popular this could lead to either a non-compliant Game ID or split leaderboards. This situation can happen regardless of standards as people can submit any Game ID, URLs just introduce a barrier for compliant entries.
</p>
</details> 
<details><summary>4. Uniformity and appearance</summary>
<p>
This is mostly personal preference and opinion. I think mixing and matching various URIs and URLs as primary identifiers sounds unappealing.
</p>
</details> 
<details><summary>Points against my own standard</summary>
<p>
The naming system I pitched isn't fool proof by any means. There's interpretation in a games name e.g. Marvel's Spider-Man on the PS4, most people just call it Spider-Man although by the rules I mentioned above it would be MASPMA002018, Wikipedia even has it's page name as just Spider-Man so someone may interpret it as SPIMAN002018. I'm working on adjusting the amount of information required and wording around the interpretation of game names. My proposed standard also reveals nothing about the game other than the required information in it's creation, whereas a URL/URI system may be able to utilise APIs via parsing out the Game's ID from the provided URL to obtain further information where applicable. This could make client development easier and allow for more dynamic and reactive leaderboards.

The main reason I've been using Spider-Man games as examples is because there's an abundance of them with a lot of similar names, I don't have an obsession with Spider-Man.

I also think my proposed standard isn't the best in terms of functionality or appearance, what I think would be better is if there would be some way to have game information posted as a note for submissions to reference, I'm just unsure of if and how this would be accomplished in a method that doesn't just lead to having 7 different notes about the same game with different information.
</p>
</details> 

All of these issues would be massively mitigated once a client is developed with runs submitted as it can be used as a refence for developing other clients or individuals making submissions. 
I'm not meaning to completely dismiss a URL/URI system as I feel it would work fine for majority of cases if the wording was clear, I just wanted to highlight some potential considerations. I appreciate the back and forth though as I feel it's all helpful to produce a better end product, I've got a few more ideas that will depend on having a solidly established Game identification system.

Semisol (changes requested):

76.md
alexgleason: Check https://github.com/nostr-protocol/nips/pull/1185 as well

They're doing stuff with `isbn:` and stuff

Ambiguity is a problem without URIs. Perhaps even worse, because it's missing context for what that ID actually is.

Same argument for mixing IDs without URIs. The URIs is what gives them uniformity.

It's common to use URIs in decentralized systems. This is the exact use-case we created them for.

I imagine whoever publishes the first speedruns that matter, people will just follow what they put as the ID.

The only other thing I can think of is that somebody creates a "Game" event to define the game, and then other people tag that event with their submission. It could actually work kind of like NIP-54 [wikis](https://github.com/nostr-protocol/nips/blob/master/54.md). But then people will need to put something as the shared identifier for the game. And then we have the issue all over again.

76.md
```diff
@@ -0,0 +1,69 @@
...
+    ["lvl", "<Level>"], //E.g. FG, W1L1, F3OS
+    ["ver", "<Version>"], //Version the submission is ran on e.g. 1.4.0.0
+    ["rlt", "<Start Unix Time>", "<Stop Unix Time>"], //Real Life Time, subtract Start from Stop to get actual time
+    ["igt", "<Start Unix Time>", "<Stop Unix Time>"], //In-Game Time
+    ["proof": "<Link to proof>"] //e.g. Twitch Vod, replay file
+  ],
+  "content": "<Description for run>", //User written comment about run if desired
+  "created_at": "<Unix Time>",
+  "pubkey": "<Public Key>",
+  "sig": "<signature>"
```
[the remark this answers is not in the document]
BlueZTV: I've removed these lines.

76.md
BlueZTV: > I imagine whoever publishes the first speedruns that matter, people will just follow what they put as the ID.

I ultimately agree. It's a first in best dressed situation, which I mentioned.

>The only other thing I can think of is that somebody creates a "Game" event to define the game, and then other people tag that event with their submission. It could actually work kind of like NIP-54 wikis. But then people will need to put something as the shared identifier for the game. And then we have the issue all over again.

I addressed this solution as well in my response, just poorly worded. Fair enough if you didn't read my whole response, keeping things concise was never a specialty of mine.
If it were to be a wiki/event that runs referenced, could it have multiple URLs/URIs/IDs in the tags of the event or be modified to include this functionality? Then whenever a fractured leaderboard occurs the event could be edited to include that identifier thereby grouping the runs. I like the idea of group collaboration/rating, although I don't know enough about the applications of nostr wikis to ultimately gauge their viability for this.

You only address URIs here, when even in your initial response you acknowledge games that aren't on Steam or other platforms that don't use URIs.
For games such as Super Mario 64, .kkrieger and the indie game my mate's made but hasn't published to any sites yet, how would you go about it?
There's also games such as Hades that came out on Epic Games and then released on Steam after a period of exclusivity. Not an immediate problem because people can just reference other submissions for the ID, however, for people who just read the NIP which would presumably say to use Steam as a preference it would lead to split leaderboards.

It would be lovely if there was a system like isbn for games. I read through the link you mentioned and the podcast guid system stood out as one that may be applicable, just use a different namespace UUID and `<game:UUID>` for the `g` tag. But then there's the case of what do we use as the input, which leads us back to the start, it's like some horrid ouroboros of unique identifiers.

I'm fine to go with a URI/URL system, with the GUID even if everyone reckons that's for the best. I just wanted to outline some potential concerns I had with it.

alexgleason:
I still think this is a good idea. Blocked by the fast there's no good way to refer to video games in a standardized way.

Also, there's no open API for getting video game metadata like there is [OpenLibrary API](https://openlibrary.org/developers/api) for books. I think that problem needs to be solved first. I researched it, and there are like 5 different scummy closed platforms with paid API subscriptions. Gamers unite.

BlueZTV:
> I still think this is a good idea. Blocked by the fact there's no good way to refer to video games in a standardized way.
> 
> Also, there's no open API for getting video game metadata like there is [OpenLibrary API](https://openlibrary.org/developers/api) for books. I think that problem needs to be solved first. I researched it, and there are like 5 different scummy closed platforms with paid API subscriptions. Gamers unite.

Hey, me too gamer.
Do you think it's a problem that could be solved on nostr if we just made our own system?
It could then also allow for other things, such as high scores or reviews.
I don't know, I haven't thought this in a while, but I'd love to have a more in-depth chat anytime if you think it's possible.

alexgleason:
I ended up making heavy usage of wikidata identifiers in other Nostr projects, and for games I think this is the way. Eg:

```json
["i", "https://www.wikidata.org/wiki/Q732943", "The Legend of Zelda: Majora's Mask"]
```

BlueZTV:
Thanks for still being here Alex.
So people would just make a new Wikidata entry for games that aren't on there yet? That seems like a fine middleground I suppose, but I like Vitor's idea of just making our own system. I had Gemini grossly throw a thing together with the basics which just got the ball rolling. I'm lazy.

Could it not just point to NIP-54 wikis of the games and achieve a similar result without pointing to external sites?

PR #1210, opened 2024-05-01
NIP-85: Attestation of DNS-based identity providers

aidik:
The impact of NIP-05 identifiers is not as big as it could be. Spammers can spawn an unlimited amount of NIP-05 verified npubs with one or just a few domains. So a simple verified NIP-05 does not add much trust to the network. 

This NIP adds a process to chain NIP-05 providers and increases trust similar to how certificates work. Allowing users to select trusted identity providers based on various criteria, e.g. manual verification process, content rules, community standards, etc. thereby bringing the freedom of choice to the user.

Here is a simple implementation on the NIP-05/85 provider side:

NIP-05 verification:
https://anarcho.capital/.well-known/nostr.json?name=aida

NIP-85 attestation verification:
* for individual NIP-05 https://anarcho.capital/.well-known/nostr.json?attest=aida@anarcho.capital
* for whole domain https://anarcho.capital/.well-known/nostr.json?attest=anarcho.capital

### DIFF
--- a/85.md
+++ b/85.md
@@ -1,0 +1,108 @@
+
+NIP-85
+======
+
+Attestation of DNS-based identity providers
+--------------------------------------------
+
+`draft` `optional`
+
+NIP-85 extends the well established NIP-05 in the way that clients would be able to verify not only the `<local-part>` but also assess the `<domain>` part based on published attestations from other NIP-05 identity providers. This will allow clients to select trusted identity providers which then will extend to all of their attestees.
+
+#### Example
+
+If a client sees an event like this:
+
+```json
+{
+  "pubkey": "b0635d6a9851d3aed0cd6c495b282167acf761729078d975fc341b22650b07b9",
+  "kind": 0,
+  "content": "{\"name\": \"bob\", \"nip05\": \"bob@example.com\"}"
+  ...
+}
+```
+
+It will make a GET request to `https://example.com/.well-known/nostr.json?name=bob` and get back a response that will look like
+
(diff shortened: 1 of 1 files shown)


aidik:
HI @Semisol, can you be more specific? How would a badge solve the links of trusts proposed here?
```
Me ----- [ NIP 5 Provider A ] ----- [ NIP 5 Provider B] ----- You 
```
I trust A,
You trust B,
A and B trust each other
therefore we are mutually attested

chmac:
> If the pubkey matches the one given in `"names"` (as in the example above) that means the association is correct and the client can continue to verify attestations from the `"attestators"` attribute which contains an object with two optional attributes. One equal to the (`<local-part>` and `<domain>`) `"nip05"` identifier  with value consisting of an array of URLs of attestators attesting this individual `"nip05"` identifier. The second attribute is equal to the `<domain>` and the value is an array of URLs attesting the whole `<domain>` of the DNS-based identity provider.

Couple of minor typo fixes

Giszmo:
NACK

Does anybody **trust** any nip05 provider wholesale? Is that even the point of nip05? I see this nip as promoting centralization in nip05 providers for no good reason.

The only nip05 identifiers that instill trust in me are one-offs or really tiny ones where I know the user list is hand picked. A verified elon@x.com would have my attention (but not necessarily my trust) but once Xitter opens the floodgate and gives each handle a nip05 id, I would not trust them wholesale anymore.

PR #1215, opened 2024-05-02
NIP-56: allow tagging third-parties in reports

alexgleason:
When creating a report, `p` tags are used to indicate the user(s) **being reported**.

This change introduces capital `P` tags to **notify third-parties about the report**. These parties are not subjects of the report themselves, but it is rather an indication that the author of this report wants those additional people to know.

I am using `P` tags for a community Nostr relay with a moderation interface, where users can report events to the relay, and then an admin can delete those events from the community relay.

This inclusion has other possible use-cases, such as notifying bots or kicking off webhooks that can trigger other side-effects.

### DIFF
--- a/56.md
+++ b/56.md
@@ -37,6 +37,9 @@ Some report tags only make sense for profile reports, such as `impersonation`
 `l` and `L` tags MAY be also be used as defined in [NIP-32](32.md) to support
 further qualification and querying.
 
+Capital `P` tags MAY also be included on a report to notify third-parties who
+may be able to act upon the report, such as relay admins.
+
 Example events
 --------------
 


alexgleason:
@staab @vitorpamplona @mikedilger @fiatjaf and anyone else working on Nostr moderation stuff

staab:
This seems fine, but I also feel a little nervous about tagging people in reports who don't have anything to do with it. This could be misinterpreted pretty easily, and get some people in trouble. Probably not a real objection though. I also don't like the capital P tag, just for aesthetic/semantic reasons.

Maybe a better way to do this would be a new `notification` event? Just a new kind + e/a tag and a p tag for recipient? Those could then be gift-wrapped for privacy if desired. This would decouple the report itself from anyone who might want to know about it, and make the same pattern applicable to any kind.

alexgleason:
Notification event type seems like overkill unless it also solves other problems (maybe it could?). But that's a whole other thing to be spec'd out.

I'm open to using a tag other than `P` if you have a better suggestion. I went with it due to its use in NIP-57 for consistency, as an "alternate p-tag". It seems like the right tag to use.

mikedilger:
If you use 'P', existing code won't recognize it and people that you are trying to notify won't see it.  It seems like the reverse makes more sense, but indicates that report shoudn't have been using 'P' tags to start with.  If the report event was defined with 'p' being the people reported (I didn't look I'm trusting you) then they are also currently being notified (no wonder the uproar I've seen).

alexgleason:
Yes @mikedilger I also thought it's weird we are essentially notifying the people who are reported.

I concluded that kind 1984 requires special treatment.

alexgleason:
@rabble Is this what you're doing in Nos? Can we have interoperability?

https://gleasonator.dev/@rabble@nos.social/posts/ac5f16e2cb12238bf82506fbcf171d242d84c80e9880ed5c32031a8cb36eaff8

PR #1220, opened 2024-05-03
Wikilink tags and Trust Rank NIP (and how to apply it to wikis)

arthurfranca:
Edited TLDR:

When user gives a thumbs up/down to an article, wiki client auto sets the trust rank for the wiki author relative to the said article (and to some other closely related articles; most likely the first two article links).

By looking at an user network (him and his follows) of trust metrics, client can figure out the best article version to show.

### DIFF
--- a/54.md
+++ b/54.md
@@ -31,6 +31,69 @@ The content should be Markdown, following the same rules as of [NIP-23](23.md), 
 
 One extra functionality is added: **wikilinks**. Unlike normal Markdown links `[]()` that link to webpages, wikilinks `[[]]` link to other articles in the wiki. In this case, the wiki is the entirety of Nostr. Clicking on a wikilink should cause the client to ask relays for events with `d` tags equal to the target of that wikilink.
 
+#### Wikilink Tags
+
+The one to two most prevailing wikilink `d` tag references of an article should be set to the `y` and `z` tags.
+The `y` tag is set to the most prevailing one while `z` is set to the least of the two.
+
+A prevalent wikilink is a reference that is able to categorize the article into a specifc topic. These are usually positioned
+on the top of the article on an introductory paragraph.
+
+A Nostr wiki client may try to auto assign those tags. For example, by finding the first wikilink's paragraph on the article
+then considering the up to first two wikilinks on the current paragraph as the prevailing wikilinks.
+
+Example:
+
+```js
+{
+  "kind": 30818,
+  "tags": [
+    [ "d", "john-doe" ],
+    [ "title", "John Doe" ],
+    [ "y", "internet-celebrity" ],
+    [ "z", "nostr" ],
+  ],
+  "content": "# John Doe\n\n" +
+    "John Doe is an [influencer[internet-celebrity]] in the [Nostr[nostr]] scene. He is known for his [memes[meme]].\n" +
+    "# Early Life\n\n" +
+    "He was born in [Los Angeles[los-angeles]] ...",
+  // other fields
+}
+```
+
+Those tags are handy when searching for topics related to an article.
+They may be used as labels to tell two articles with the same `d` tag apart.
+They are also used when [ranking authors](#author-ranking).
+
+#### Author Ranking
+
+When a reader adds a [NIP-25](25.md) reaction to an article,
+the client may help them with adding/updating a [NIP-64](64.md) trust rank
+on the main and related topics regarding the article author.
+
+For example, if an user **likes** the above article, the author may be auto ranked as follows:
+
+```js
+{
+  "kind": 30382,
+  "tags": [
+    ["d", "<wiki-article-author-pubkey>"],
+    // "x" is the marker for the main article topic (its `d` tag "john-doe")
+    ["T", "3:wiki:john-doe", "x"],
+    // even though this wiki article author may not have written about "internet-celebrity" yet (or never will),
+    // this rank may be used to consider the relevance of a "x"-marked trust rank eventually authored
+    // by them (or by someone even deeper, if following another y/z rank) on the "internet-celebrity" topic
+    ["T", "2:wiki:internet-celebrity", "y"],
+    ["T", "1:wiki:nostr", "z"]
+  ],
+  // ...other fields
+}
+```
+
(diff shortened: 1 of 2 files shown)


54.md
```diff
@@ -31,6 +31,71 @@ The content should be Markdown, following the same rules as of [NIP-23](23.md),
...
+    "John Doe is an [influencer[internet-celebrity]] in the [Nostr[nostr]] scene. He is known for his [memes[meme]].\n" +
+    "# Early Life\n\n" +
+    "He was born in [Los Angeles[los-angeles]] ...",
+  // other fields
+}
+```
+
+Those tags are handy when searching for topics related to an article.
+They may be used as labels to tell two articles with the same `d` tag apart.
+They are also used when [ranking authors](#author-ranking).
```
dluvian, line 66: Having `y` and `z` tags instead of just `t` tags for these use-cases feels overly complicated.

64.md
```diff
@@ -0,0 +1,38 @@
...
+
+```js
+{
+  "kind": 30382,
+  "tags": [
+    ["d", "<pubkey>"],
+    ["T", "3:wiki:role-playing-video-game", "x"], // has written a based wiki article about RPG
+    ["T", "2:driver"], // polite and skilled driver but doesn't turn the air conditioner on
+    ["T", "3:buyer"], // no problems when selling to this person
+    ["T", "-2:seller"] // had a bad experience when buying from this person or company
```
dluvian, line 28: I would prefer a different `kind` for each trust rank application or another indexable tag that sets the application like `["D", "wiki"]`. Otherwise a wiki client would just discard the `driver`, `buyer`, and `seller` tags when updating a wiki value, because these are not useful for the client, leading to interoperability issues.

54.md
arthurfranca: Agree. Though used `r` instead of `t`. People tend to over use hashtags and I think the `d` value would end up being a `t` tag too, which is not the intent.

64.md
dluvian: My point remains. Leaving junk data untouched would require client devs to write extra code for no benefit to their own clients. It's also more bandwidth efficient if I can query only the tags that are of interest to me.

64.md
```diff
@@ -0,0 +1,41 @@
...
+`draft` `optional`
+
+A trust rank is a public way to individually assign a quantifiable level of trust on specifc topics to a nostr identity.
+Ranks can be used to discover trusted people or content on a common topic.
+
+## Event Structure
+
+The `kind:30382` is used to assign public relationship statuses between the event author and a pubkey set to the `d` tag.
+This NIP assigns one or more trust ranks by adding `T` tags with different topic names to such event
+using the following structure: `["T", "<int_value_ranging_from_-3_to_3>:<topic-name>", "<context_dependent_marker>"]`.
```
dluvian, line 16: What do you think about `["T", "<topic-name>", "<int_value_ranging_from_-3_to_3>", "<context_dependent_marker>"]`?
This way you could easily query a pubkey's trust rank for a topic.

64.md
arthurfranca: > Leaving junk data untouched would require client devs to write extra code for no benefit

Even if there was a kind just for ranking authors of wiki articles, the client would have to fetch the latest version and don't touch ratings on topics other than the current one. Similar work.

> [...] indexable tag that sets the application like ["D", "wiki"] [...] It's also more bandwidth efficient if I can query only the tags that are of interest to me

One isn't supposed to fetch all wiki-related trust scores but do one article at a time like this:
```
{ kinds: [30382], authors: ['<my>', '<follows'>, '<...>'], tags: [
  '3:wiki:nostr',
  '2:wiki:nostr',
  '1:wiki:nostr',
]}
```
The above will return just people with positively ranked knowledge on the encyclopedic nostr topic (randed by one's follows). Those with the `d` marker will lead to a recommended article variant. Pubkeys ranked with `r` markers can be used on a follow-up query in the authors filter till reaching a `d` marker to lead to a recommended article variant.

64.md
arthurfranca: Your example doesn't allow to filter on relays by trust score.

The reason the score has few possible values (-3 to 3) and goes together with the topic name is that it enables filtering on relays by both (always picking one or more specific topics).

Filter examples:
All "nostr": `{ #T: ['-3:wiki:nostr', '-2:wiki:nostr', "...", '3:wiki:nostr'] }`
Positive "nostr": `{ #T: ['1:wiki:nostr', '2:wiki:nostr', '3:wiki:nostr'] }`
Negative "nostr": `{ #T: ['-3:wiki:nostr', '-2:wiki:nostr', '-1:wiki:nostr'] }`

arthurfranca:
Am I expecting too much from this or am I right to expect these changes to:
1) Allow people to be rated as `<insert-topic>` gurus after writing good wiki articles on said topic
2) Which motivates `<insert-topic>` influencers to write based wiki articles
3) Which gonna expand human public knowledge base
4) And as a bonus could give us better notes when clicking a hashtag `#<insert-topic>` (notes from those gurus)
5) And then all wiki articles would finally be wiped out by another Damus Relay database reset 😅️

@fiatjaf @pablof7z @vitorpamplona @staab @mikedilger

64.md
dluvian: > Even if there was a kind just for ranking authors of wiki articles, the client would have to fetch the latest version and don't touch ratings on topics other than the current one. Similar work.

No. All wiki ratings are relevant to a wiki client, every other ranking is junk. Same with the legacy follow lists. If a client does not implement following relays and hashtags than it would just ignore these tags, why would the client cache these irrelevant tags? This leads to clients nuking each others lists. 
Nip51(Lists) does it right, having a different kind for each use-case because managing one giant list is hell, same applies to these rankings.
Also it's much more efficient if you don't have to constantly refetch and validate the event when a marketplace client changes tags that have nothing to do with your wiki use-case.
Doesn't have to be a new kind, maybe put it in the d tag like  `["d", "wiki:<pubkey>"]`

arthurfranca:
Gonna separate the trust nip part

PR #1222, opened 2024-05-04
NIP-41: Unix Domain Sockets

ChristianChiarulli:
Example using unix domain sockets with two desktop apps where one is acting as a remote signer: https://twitter.com/chrisatmachine/status/1786549455967154312

### DIFF
--- a/55.md
+++ b/55.md
@@ -10,7 +10,7 @@ For the purpose of this NIP, we will define a standard way to communicate via Un
 
 ## Defining the file path for UDS
 
-NIP-55 defines a standard way to communicate via Unix Domain Sockets (UDS). UDS needs a file path as a qualifier so that both client and server can agree on a place to exchange information.
+UDS needs a file path as a qualifier so that both client and server can agree on a place to exchange information.
 
 The file path will be chosen as `/tmp/nip55`
 
--- a/55.md
+++ b/55.md
@@ -1,0 +1,11 @@
+# NIP-55 - Unix Domain Socket Communication
+
+## Defining the file path for UDS
+
+NIP-55 defines a standard way to communicate via Unix Domain Sockets (UDS). UDS needs a file path as a qualifier so that both client and server can agree on a place to exchange information.
+
+The file path will be chosen as `/tmp/nip55`
+
+## Communication
+
+There are potentially many ways to communicate over this socket, one way would be the JSON-RPC collection found in NIP-46.
--- a/55.md
+++ b/55.md
@@ -12,7 +12,9 @@ For the purpose of this NIP, we will define a standard way to communicate via Un
 
 UDS needs a file path as a qualifier so that both client and server can agree on a place to exchange information.
 
-The file path will be chosen as `/tmp/nip55`
+The system wide file path will be chosen as `/tmp/nip55`
+
+The user specific file path will be chosen as `~/.nostr/nip55`
 
 ## Communication
 
--- a/55.md
+++ b/55.md
@@ -12,9 +12,9 @@ For the purpose of this NIP, we will define a standard way to communicate via Un
 
 UDS needs a file path as a qualifier so that both client and server can agree on a place to exchange information.
 
-The system wide file path will be chosen as `/tmp/nip55`
+The system wide file path will be chosen as `/tmp/nip55.sock`
 
-The user specific file path will be chosen as `~/.nostr/nip55`
+The user specific file path will be chosen as `~/.nostr/nip55.sock`
 
 ## Communication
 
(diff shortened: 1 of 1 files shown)


fiatjaf:
@mikedilger

vitorpamplona:
Nice! Let's go!

vitorpamplona:
Why the link to Twitter, though? :)

ChristianChiarulli:
Thanks @vitorpamplona, also I've had a lot of issues with kind 1 web clients.

tvolk131:
A few thoughts:

* Should we define a message structure and encoding? I really like what NIP-46 uses for requests and responses, where they both share a designated kind, and the content field contains a NIP-04 encrypted JSON-RPC 2.0 request or response. When using JSON-RPC 2.0 over byte sockets, we also need a message delineation mechanism. See [here](https://www.simple-is-better.org/json-rpc/transport_sockets.html) for some possible options - I'd vote for the first option, shutdown after every request/response, for its simplicity.
* Building on this, should we have socket namespaces? We could use `/tmp/nip55-kind{kind number}` instead of `/tmp/nip55`. This way there could be more than one server on a single device while also keeping the purpose of each one clear.

Here's a reference implementation in Rust containing a working server and client which uses the message structure and encoding mentioned above: https://crates.io/crates/nip-55

vitorpamplona:
You can also use the Android Signer spec as a reference: https://github.com/nostr-protocol/nips/pull/868 

The interesting part of the Android spec is that it has two modes: with UI and without it. It always tries to do the operation without transferring the user to the other app and only if it fails, it brings the signer UI up.

mikedilger:
Cool.

It's not nostr, because it's not websocket communication through relays.

Maybe nostr should be a layered specification, with the top event layer defined in one spec, relay-client messages in another lower spec, websocket transport in a lower yet spec.   In that world, this demo would be using not just a different transport, but a different sub-event protocol.

vitorpamplona:
> Maybe nostr should be a layered specification, with the top event layer defined in one spec, relay-client messages in another lower spec, websocket transport in a lower yet spec.   In that world, this demo would be using not just a different transport, but a different sub-event protocol.

IMHO, it already is like that. The NIP description is just not up to speed. The real Nostr Protocol is only the event definition/sign/verify. Everything else is layer 2+.

ChristianChiarulli:
@mikedilger does that call into question the validity of nip-7 since it is not websocket communication through relays?

mikedilger:
Yeah I think nip-7 and nip-6 too are not really nostr. It doesn't mean we can't put it in a nip, we are sloppy about a lot of things.

alexgleason:
Why not just do local http?

ChristianChiarulli:
@alexgleason one reason is that I don't want to deal with figuring out how to get the server and client to agree on a port.

@mikedilger right, maybe it's worth outlining some place in the future to put things like this nip, nip-7 etc, since standards like this allow more streamlined development for clients and greater security.

tvolk131:
> Why not just do local http?

I see two reasons:

1. For local communication, UDS is simpler and more performant.
2. If we want socket namespaces like I mentioned above (`/tmp/nip55-kind{kind number}` instead of just `/tmp/nip55`) then I'm not sure how this would translate to HTTP. With UDS there are effectively port _names_ rather than port _numbers_ which allows for easy extensibility without the need for a discovery mechanism

tvolk131:
It looks like Windows has supported UDS since 2017 when they added support for the SOCK_STREAM socket type (see [here](https://devblogs.microsoft.com/commandline/af_unix-comes-to-windows/)). We could include Named Pipes as an alternative/fallback for Windows (although I don't think that's necessary - I just figured it's worth mentioning).

SOCK_STREAM is analogous to TCP and is what we'd want to use for JSON-RPC 2.0 as I mentioned [here](https://github.com/nostr-protocol/nips/pull/1222#issuecomment-2096914751).

ChristianChiarulli:
I think maybe we should narrow it down to just passing the JSON defined in NIP-46. Maybe in the future other events could be passed via UDS but for now I think this is fine.

I'll think about multi user, I'm open to suggestions here.

This NIP does not apply to non Unix systems.

vitorpamplona:
> I'll think about multi user, I'm open to suggestions here.

I would switch `/tmp/...` to a `.nostrsigner.sock` in the user's home directory. In that way, system apps can have access to it if you give the right permissions to the file. Otherwise, only the user's apps can interact with it.

ChristianChiarulli:
@vitorpamplona I think that's probably a good idea to manage it in the users home directory, maybe we could define two standard locations in this nip, one that is system wide and one user specific?

ChristianChiarulli:
would a concrete implementation in JS or Python be helpful to add to the nip?

PR #1226, opened 2024-05-08
rename "parameterized replaceable event" to "tagged-replaceable event"

fiatjaf:
Because it's impossible to write and takes two minutes to say that other word, "tagged" is nicer. If you have a better name please let me know.

### DIFF
--- a/01.md
+++ b/01.md
@@ -77,9 +77,9 @@ This NIP defines 3 standard tags that can be used across all event kinds with th
 
 - The `e` tag, used to refer to an event: `["e", <32-bytes lowercase hex of the id of another event>, <recommended relay URL, optional>]`
 - The `p` tag, used to refer to another user: `["p", <32-bytes lowercase hex of a pubkey>, <recommended relay URL, optional>]`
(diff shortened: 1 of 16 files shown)


fiatjaf:
Maybe "addressable event"?

alexgleason:
What does `d` actually stand for?

vitorpamplona:
`tagged` is not that much better. :(

patrickReiis:
What about "**triple** replaceable event"? Considering the fact that a "parameterized replaceable" is a combination of 3 things (pubkey, kind & `d` tag)

alexgleason:
> `tagged` is not that much better. :(

Why not? It's a 7 syllable difference. Ambiguity?

I was thinking d-tagged replaceable event, or even just d-replaceable event. But I still don't know why we use that letter. Maybe we should invent a backronym.

arthurfranca:
> What does d actually stand for?

"d"eduplication id

vitorpamplona:
d-anchored event?

fernandolguevara:
named replaceable event?

staab:
I have an idea, what about "parameterized replaceable event"?

PR #1228, opened 2024-05-09
Shared replaceables via Event-owned keys

vitorpamplona:
This is the 4th proposal to solve the shared replaceable problem. IMHO this scheme is the winner.

In this one, we create private keys for each shared event and host those keys in the event itself. 

Read [here](https://github.com/vitorpamplona/nips/blob/shared-replaceables-key/68.md)

This was extracted from the https://github.com/nostr-protocol/nips/pull/1189 PR because it serves many other use cases. 

Live Demo: https://sheetstr.amethyst.social/
Code: https://github.com/vitorpamplona/sheetstr

---

As a reminder, the three other proposals are: 
1. Using a shared d-tag scheme https://github.com/nostr-protocol/nips/pull/1192 
2. Using a DVM to hold the shared private key https://github.com/nostr-protocol/nips/pull/1015
3. Using a NIP-46 Signer to hold the shared key (undocumented)

### DIFF
--- a/68.md
+++ b/68.md
@@ -1,0 +1,101 @@
+NIP-68
+======
+
+Shared Replaceables
+-------------------
+
+`draft` `optional`
+
+This NIP creates replaceable events that can be changed by any public key in the list of editors. Editors can also add and remove new editors. 
+
+Every shared replaceable event MUST be signed with it's own private key. The event owns itself. 
+
+The event's private key MUST be shared with all editors through `p` tags. The key is [NIP-44](44.md)-encrypted to each editor and placed as the 4th element in a regular `p` tag.
+
+```js
+val edittingKeyPair = nostr.generateKeyPair()
+
+{
+  "pubkey": edittingKeyPair.publicKey
+  "kind": 3xxxx or 1xxxx,
+  "tags": [
+    ["d", "<unique identifier>"]
+    ["p", "<pubkey 1>", "<relay url>", nip44Encrypt(edittingKeyPair.privateKeyHex, edittingKeyPair.privateKey, "<pubkey 1>") ]
+    ["p", "<pubkey 2>", "<relay url>", nip44Encrypt(edittingKeyPair.privateKeyHex, edittingKeyPair.privateKey, "<pubkey 2>") ]
+  ],
+  "content": "",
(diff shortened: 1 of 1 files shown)


mikedilger:
I think the general idea of this NIP is fine.

But I think we may have some issues with how tags are defined.

1) Are they kind-specific? Do they have different meanings based on the kind?  In this PR and many others, they are not kind-specific (another example would be the proof of work).
2) If the first field is not kind-specific, should we be defining kind-specific 4th fields?  It seems strictly that we could, but also that this level of complexity is probably going to bite one or two developers down the road.
3) If that field isn't kind specific, are we sure all the other NIPs and PRs have not made a claim on that slot already?

vitorpamplona:
Yeah, I went back and forth in that as well. I don't think anyone is using the 4th element for a p-tags but who knows. 

We could duplicate tags with a new tag name. So, the event would `p` tag everyone, but the keys could be in a `key` tag, like: 

```js
val editingKeyPair = nostr.generateKeyPair()

{
  "pubkey": editingKeyPair.publicKey
  "kind": 3xxxx or 1xxxx,
  "tags": [
    ["d", "<unique identifier>"]
    ["p", "<pubkey 1>", "<relay url>" ],
    ["p", "<pubkey 2>", "<relay url>" ],
    ["key", "<pubkey 1>", nip44Encrypt(editingKeyPair.privateKeyHex, editingKeyPair.privateKey, "<pubkey 1>") ],
    ["key", "<pubkey 2>", nip44Encrypt(editingKeyPair.privateKeyHex, editingKeyPair.privateKey, "<pubkey 2>") ],
  ],
  "content": "",
  "sig": signWith(editingKeyPair.privateKey)
  // ...
}
```

Initially I thought the duplication could become wasteful in cases of large groups, but maybe it's ok.

vitorpamplona:
@Semisol there is a proposal to use #875's kind 24 to transfer the keys inside wraps. It's a little more cumbersome and there is an issue with rotating keys (how to know which version of the replaceable should use which version of the key kind), but it's possible as well. 

I don't know.. I like the simplicity of putting everything in the main event as a base option and then exploring more private ways inside each NIP.

PR #1232, opened 2024-05-11
NIP-47: fix NWC connection string example

SamSamskies:
The format in the example given doesn't work with any of the Alby NWC tools.


benthecarman (approved):
Ack, this is what everyone does

PR #1236, opened 2024-05-14
NIP-96: List files / rewording, no_transform

v0l:
Adding an endpoint to list uploads

Also included a bit of rewording, this NIP is far too long.

read: https://github.com/nostr-protocol/nips/blob/nip96-sync/96.md

### DIFF
--- a/96.md
+++ b/96.md
@@ -84,46 +84,43 @@ it must use the "api_url" field instead.
 
 See https://github.com/aljazceru/awesome-nostr#nip-96-file-storage-servers.
 
+
+## Auth 
+
+When indicated, `clients` must add an [NIP-98](98.md) `Authorization` header (**optionally** with the encoded `payload` tag set to the base64-encoded 256-bit SHA-256 hash of the file - not the hash of the whole request body).
+
 ## Upload
 
-A file can be uploaded one at a time to `https://your-file-server.example/custom-api-path` (route from `https://your-file-server.example/.well-known/nostr/nip96.json` "api_url" field) as `multipart/form-data` content type using `POST` method with the file object set to the `file` form data field.
+`POST $api_url` as `multipart/form-data`.
 
-`Clients` must add an [NIP-98](98.md) `Authorization` header (**optionally** with the encoded `payload` tag set to the base64-encoded 256-bit SHA-256 hash of the file - not the hash of the whole request body).
-If using an html form, use an `Authorization` form data field instead.
+**AUTH required**
 
-These following **optional** form data fields MAY be used by `servers` and SHOULD be sent by `clients`:
-- `expiration`: string of the UNIX timestamp in seconds. Empty string if file should be stored forever. The server isn't required to honor this;
-- `size`: string of the file byte size. This is just a value the server can use to reject early if the file size exceeds the server limits;
-- `alt`: (recommended) strict description text for visibility-impaired users;
-- `caption`: loose description;
-- `media_type`: "avatar" or "banner". Informs the server if the file will be used as an avatar or banner. If absent, the server will interpret it as a normal upload, without special treatment;
+List of form fields:
+- `file`: **REQUIRED** the file to upload
+- `caption`: **RECOMMENDED** loose description;
+- `expiration`: UNIX timestamp in seconds. Empty string if file should be stored forever. The server isn't required to honor this.
+- `size`: File byte size. This is just a value the server can use to reject early if the file size exceeds the server limits.
+- `alt`: **RECOMMENDED** strict description text for visibility-impaired users.
+- `media_type`: "avatar" or "banner". Informs the server if the file will be used as an avatar or banner. If absent, the server will interpret it as a normal upload, without special treatment.
 - `content_type`: mime type such as "image/jpeg". This is just a value the server can use to reject early if the mime type isn't supported.
-
 
 Others custom form data fields may be used depending on specific `server` support.
 The `server` isn't required to store any metadata sent by `clients`.
 
-Note for `clients`: if using an HTML form, it is important for the `file` form field to be the **last** one, or be re-ordered right before sending or be appended as the last field of XHR2's FormData object.
-
 The `filename` embedded in the file may not be honored by the `server`, which could internally store just the SHA-256 hash value as the file name, ignoring extra metadata.
-The hash is enough to uniquely identify a file, that's why it will be used on the "download" and "delete" routes.
+The hash is enough to uniquely identify a file, that's why it will be used on the `download` and `delete` routes.
 
-The `server` MUST link the user's `pubkey` string (which is embedded in the decoded header value) as the owner of the file so to later allow them to delete the file.
-Note that if a file with the same hash of a previously received file (so the same file) is uploaded by another user, the server doesn't need to store the new file.
-It should just add the new user's `pubkey` to the list of the owners of the already stored file with said hash (if it wants to save space by keeping just one copy of the same file, because multiple uploads of the same file results in the same file hash).
+The `server` MUST link the user's `pubkey` string as the owner of the file so to later allow them to delete the file.
 
-The `server` MAY also store the `Authorization` header/field value (decoded or not) for accountability purpose as this proves that the user with the unique pubkey did ask for the upload of the file with a specific hash. However, storing the pubkey is sufficient to establish ownership.
+### Response codes
 
-The `server` MUST reject with 413 Payload Too Large if file size exceeds limits.
-
-The `server` MUST reject with 400 Bad Request status if some fields are invalid.
-
-The `server` MUST reply to the upload with 200 OK status if the `payload` tag value contains an already used SHA-256 hash (if file is already owned by the same pubkey) or reject the upload with 403 Forbidden status if it isn't the same of the received file.
-
-The `server` MAY reject the upload with 402 Payment Required status if the user has a pending payment (Payment flow is not strictly required. Server owners decide if the storage is free or not. Monetization schemes may be added later to correlated NIPs.).
-
-On successful uploads the `server` MUST reply with **201 Created** HTTP status code or **202 Accepted** if a `processing_url` field is added
-to the response so that the `client` can follow the processing status (see [Delayed Processing](#delayed-processing) section).
+- `200 OK`: File upload exists, but is successful (Existing hash)
+- `201 Created`: File upload successful (New hash)
+- `202 Accepted`: File upload is awaiting processing, see [Delayed Processing](#delayed-processing) section
+- `413 Payload Too Large`: File size exceeds limit
+- `400 Bad Request`: Form data is invalid or not supported.
+- `403 Forbidden`: User is not allowed to upload or the uploaded file hash didnt match the hash included in the `Authorization` header `payload` tag.
+- `402 Payment Required`: Payment is required by the server, **this flow is undefined**.
 
 The upload response is a json object as follows:
 
@@ -179,10 +176,12 @@ The upload response is a json object as follows:
 
 Note that if the server didn't apply any transformation to the received file, both `nip94_event.tags.*.ox` and `nip94_event.tags.*.x` fields will have the same value. The server MUST link the saved file to the SHA-256 hash of the **original** file before any server transformations (the `nip94_event.tags.*.ox` tag value). The **original** file's SHA-256 hash will be used to identify the saved file when downloading or deleting it.
 
-`Clients` may upload the same file to one or many `servers`.
+`clients` may upload the same file to one or many `servers`.
 After successful upload, the `client` may optionally generate and send to any set of nostr `relays` a [NIP-94](94.md) event by including the missing fields.
 
 Alternatively, instead of using NIP-94, the `client` can share or embed on a nostr note just the above url.
+
+`clients` may also use the tags from the `nip94_event` to construct an `imeta` tag
 
 ### Delayed Processing
 
@@ -219,7 +218,7 @@ However, for all file actions, such as download and deletion, the **original** f
 
 ## Download
 
-`Servers` must make available the route `https://your-file-server.example/custom-api-path/<sha256-file-hash>(.ext)` (route taken from `https://your-file-server.example/.well-known/nostr/nip96.json` "api_url" or "download_url" field) with `GET` method for file download.
+`GET $api_url/<sha256-hash>(.ext)`
 
 The primary file download url informed at the upload's response field `nip94_event.tags.*.url`
 can be that or not (it can be any non-standard url the server wants).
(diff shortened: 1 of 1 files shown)


96.md
```diff
@@ -275,6 +274,35 @@ The successful response is a 200 OK one with just basic JSON fields:
...
+Example Response:
+```js
+[
+  { 
+    "id": "<sha256-hash>",
+    "nip94_event": {...},
+    "expires": 1715691139, // unix timestamp
+    "size": 123456, 
+    "alt": "a meme that makes you laugh",
+    "caption": "haha funny meme"
```
arthurfranca, line 294: I admit placing metadata at `nip94_event.tags` wasn't the brightest idea. At the time I just wanted to make sure servers used the same metadata field names and values from NIP-94.

If you want to keep same structure from upload response (without deprecating nip94_event key and introducing an alternative structure for metadata), I think it should be like below: 

```suggestion
    "nip94_event": {
      "tags": [
        ["size", "123456"],
        ["alt", "a meme that makes you laugh"],
        ["expiration",  "1715691139"]
        // ...other metadata
      ]
      "content": "haha funny meme" // caption
    }
```

arthurfranca:
Much better than before for sure!

Regarding the listing route, the response is JSON which may eat too much memory when generating it if user has uploaded too many files.

One way to solve this is to let the server optionally add a `"nextPage": "<url-to-list-remaining-files>"` to the response if it wants to split response in pages.

vitorpamplona (approved):

fiatjaf:
@fishcakeday

96.md
```diff
@@ -275,6 +274,35 @@ The successful response is a 200 OK one with just basic JSON fields:
 }
 ```
 
+## Listing files
+
+`GET $api_url`
```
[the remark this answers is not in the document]
v0l: Yea, next pass i will add pagination

arthurfranca:
@v0l For the listing route, please check my review comment above. You are exposing fields that are already part of the nip94_event object.

96.md
```diff
@@ -275,6 +272,43 @@ The successful response is a 200 OK one with just basic JSON fields:
 }
 ```
 
+## Listing files
+
+`GET $api_url?page=x&count=y`
```
arthurfranca, line 277: Cool but if client asks for page=1&count=20 and the server returns 10 items, the client doesn't know if there existed just 10 items or if the server max page size (count) is 10. So client won't know if it should check for next page or not.

One way of solving this is if the response is a jsob object instead of json array, something like this:
```js
{
  count: 10, // page size, not necessarily the number of items returned, which may be lower
  data: [
    { "id": ...,  ... },
    { "id": ...,  ... },
   // ...
  ]
}
```

This way client will know the max page size is 10 instead of 20.

quentintaranpino:
No one is counting much on my opinion but it's fine with me 😀😅🤓

nostrcheck-server will implement it.

96.md
```diff
@@ -275,6 +272,43 @@ The successful response is a 200 OK one with just basic JSON fields:
...
+]
+```
+
+`<sha256-hash>` is the **original hash**, ie. `ox`
+
+`nip94_event` is the same as in the upload result.
+
+### Query args
+
+- `page` page number (`offset=page*count`)
```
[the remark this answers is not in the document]
v0l: Yea i think this makes sense, every time cursor is present you know there are more results.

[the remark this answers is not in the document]
v0l: For example, UI loads page 0 on load, response should contain total number of pages (on in the nip yet) and UI can display page numbers

nostrband:
Could we add an optional param to request "no transformation" from the server, so that GET requests would always serve the original file by default? And if it's not supported by the server it could reply with some 40x code. It seems like this single addition would make this nip supercede blossom.

v0l:
> Could we add an optional param to request "no transformation" from the server, so that GET requests would always serve the original file by default? And if it's not supported by the server it could reply with some 40x code. It seems like this single addition would make this nip supercede blossom.

I think i would rather drop the entire "original file" concept and just allow uploads with no compression, i think the `ox` tag was a big mistake.

When you upload a file to some server and it gets compressed then that file should be the original hash.

sant0s12:
@v0l Perhaps the sever could reply with a hash for each of the variants it intends to store so that the client can verify that the have not changed.

v0l:
> nip-96 is mainly for media.

The title of the NIP is "File Server Integration"

v0l:
@fishcakeday i dont think anybody was arguing about removing the resize/compress step, i think its vital, not sure where you are going with these comments.

sant0s12:
@fishcakeday 
> What if variant computed on-demand as it should be? Did you take into account the time it will take to compute and the resources it will require to do it? Did you consider video?

Fair, you are right.

@v0l 
> I think i would rather drop the entire "original file" concept and just allow uploads with no compression, i think the ox tag was a big mistake.

Yes I think the server should just return one hash that has different meanings:

- If the uploader requests no transformation, then this is the hash of the original file and it is expected that the server will continue serving a file that matches this.
- Otherwise (default) this is not expected to match any files served since the server might apply any transformations it wants. This can be viewed more like an id, although in practice the server may use the hash of the initially compressed file or whatever, as long as it is unique.

nostrband:
> > Could we add an optional param to request "no transformation" from the server, so that GET requests would always serve the original file by default? And if it's not supported by the server it could reply with some 40x code. It seems like this single addition would make this nip supercede blossom.
> 
> I think i would rather drop the entire "original file" concept and just allow uploads with no compression, i think the `ox` tag was a big mistake.
> 
> When you upload a file to some server and it gets compressed then that file should be the original hash.

How does "allow uploads with no compression" differ from my proposal for "no transformation" option? I'm missing the point maybe.

I don't care what hash the server returns, it may lie, and that's what I need to protect against. A client knows the file hash, uploads the file, and then hash is published somewhere, and then everyone can download using that hash and verify file's integrity, and can reupload the same file everywhere and everyone can verify it. AFAIU that's the point of blossom. I thought a single optional upload param could work for this. Or not?

v0l:
> How does "allow uploads with no compression" differ from my proposal for "no transformation" option? I'm missing the point maybe.

The reason i want this is to allow copying of files between NIP-96 servers without modification so the hash is the same, in the same way that blossom allows cloning of files, if 2 different implementations process the media in different ways then the resulting hash might be different if you re-upload the original.

If the server returns the file hash (after processing) then i take this as the "original" and then replicate it to other servers (keep in mind that i think `ox` is a mistake), i think this is better than uploading an un-compressed image to multiple servers, you ask your first preference server to compress the media then you copy it, other servers can always reject the upload if they dont want to allow replication of files (no-compress).

Probably each implementation needs to check if an image/video is "compressed enough" to allow replication. 

At the end of the day, the user is still going to be paying for space (usually) so it shouldn't matter if its compressed or not.

96.md
```diff
@@ -84,46 +82,43 @@ it must use the "api_url" field instead.
...
-- `content_type`: mime type such as "image/jpeg". This is just a value the server can use to reject early if the mime type isn't supported.
+List of form fields:
 
+- `file`: **REQUIRED** the file to upload
+- `caption`: **RECOMMENDED** loose description;
+- `expiration`: UNIX timestamp in seconds. Empty string if file should be stored forever. The server isn't required to honor this.
+- `size`: File byte size. This is just a value the server can use to reject early if the file size exceeds the server limits.
+- `alt`: **RECOMMENDED** strict description text for visibility-impaired users.
+- `media_type`: "avatar" or "banner". Informs the server if the file will be used as an avatar or banner. If absent, the server will interpret it as a normal upload, without special treatment.
+- `content_type`: mime type such as "image/jpeg". This is just a value the server can use to reject early if the mime type isn't supported.
```
nostrband, line 104: Suggested "no transformations" option:

+- `no_transform`: "true" asks server to do no transformations and serve the uploaded file as is, may be rejected. By default it's up to the server.

sant0s12:
@v0l 
> If the server returns the file hash (after processing) then i take this as the "original" and then replicate it to other servers (keep in mind that i think ox is a mistake), i think this is better than uploading an un-compressed image to multiple servers, you ask your first preference server to compress the media then you copy it, other servers can always reject the upload if they dont want to allow replication of files (no-compress).

This assumes that the first server will not change the file again or serve different versions so I'm not sure that your scheme will work. I would suggest compressing client side and then sending the result of that to whichever servers you want with the "no modification" flag set.

96.md
```diff
@@ -84,46 +82,43 @@ it must use the "api_url" field instead.
...
-
-On successful uploads the `server` MUST reply with **201 Created** HTTP status code or **202 Accepted** if a `processing_url` field is added
-to the response so that the `client` can follow the processing status (see [Delayed Processing](#delayed-processing) section).
+- `200 OK`: File upload exists, but is successful (Existing hash)
+- `201 Created`: File upload successful (New hash)
+- `202 Accepted`: File upload is awaiting processing, see [Delayed Processing](#delayed-processing) section
+- `413 Payload Too Large`: File size exceeds limit
+- `400 Bad Request`: Form data is invalid or not supported.
+- `403 Forbidden`: User is not allowed to upload or the uploaded file hash didnt match the hash included in the `Authorization` header `payload` tag.
+- `402 Payment Required`: Payment is required by the server, **this flow is undefined**.
```
nostrband, line 122: Contd "no transformations":

+- `422 Unprocessable Content`: An option "no_transform=true" was specified, but server rejected, response body may contain an explanation.

nostrband (approved):

nostrband:
> > How does "allow uploads with no compression" differ from my proposal for "no transformation" option? I'm missing the point maybe.
> 
> The reason i want this is to allow copying of files between NIP-96 servers without modification so the hash is the same, in the same way that blossom allows cloning of files, if 2 different implementations process the media in different ways then the resulting hash might be different if you re-upload the original.
> 
> If the server returns the file hash (after processing) then i take this as the "original" and then replicate it to other servers (keep in mind that i think `ox` is a mistake), i think this is better than uploading an un-compressed image to multiple servers, you ask your first preference server to compress the media then you copy it, other servers can always reject the upload if they dont want to allow replication of files (no-compress).
> 
> Probably each implementation needs to check if an image/video is "compressed enough" to allow replication.
> 
> At the end of the day, the user is still going to be paying for space (usually) so it shouldn't matter if its compressed or not.

I've drafted a "no_transform=true" option above. 

Could it be used for copying of the compressed file? You upload to a trusted server, get the compressed file and it's hash (when it's ready?) and then can publish the compressed hash, while client/server may reupload compressed version with "no_transform=true".

sant0s12:
> > This assumes that the first server will not change the file again or serve different versions
> 
> Once the file is compressed, the hash is returned, you put this hash in your posts as the `x` tag and so clients check when they fetch this file that the hash is correct, the server cannot modify it again or it will break.
> 
> This is how i have implemented it in `void-cat-rs` i don't return the `ox` or use it at all, there is only 1 hash, and it is the hash of the file after compression.

As far as I understand (please correct me if I'm wrong), @fishcakeday would like to serve dynamic variants of the same file, which I can see why it would be useful (especially for video). Your approach only works if the server only compresses the file exactly once and serves that forever.

96.md
sant0s12: Would it make sense to have a `no_transform` field in the `plans` to let the client know beforehand?

v0l:
@sant0s12 on the subject of versions, i don't really see how this can work, to me each version is its own upload, if you want multiple versions then you need to upload using the `type` or another spec should be created to ask the server to resize the image for you.

> Your approach only works if the server only compresses the file exactly once and serves that forever.

Yes this is how it should work, you cannot modify it after the fact as some sort of optimization because the hash will not be the same, the point of this spec is to avoid the media rugpull, you cannot serve whatever you want it has to be the same media that the user uploaded in the first place.

v0l:
@sant0s12 @nostrband @fishcakeday this is completely off topic for this PR btw, i was going to make another PR to add the `no_transform` option later but here we are lol

sant0s12:
Yes, let's discuss this in a new PR. @v0l do you want to open one?

quentintaranpino:
I take this opportunity to show you another PR I did some time ago, where I tried to unify the URL's as blossom (before blossom 😅🤓)

https://github.com/nostr-protocol/nips/pull/1097

If someone wants to open a debate there I'll be happy to adapt the idea to what the consensus seems appropriate.

96.md
v0l: I believe the 400 response should cover this

arthurfranca:
I will just take the opportunity too xD (sorry for the PR offtopic) to explain why the `ox` tag concept exists. It is simply that from the user point of view, they have a file that when sent to multiple file servers can then be represented by different `x` tags. It is easier for the user to anchor all these variants by a single `ox` tag, cause for the user it represents the same one thing they intended to upload. With the no-transform PR, `ox` and `x` will be the same, no problem.

About the different (image) versions like thumbnails, there is the `?w=<width>` param on NIP-96. Server should return the version it has (or generate it on the fly) that is closest to the asked width.

For videos, i don't know what similar query param could be useful to hint server could send a lower quailty version, maybe `?w=<width>` too, but now representing not the video container's width, but the max device width (for cellphones it would be its height, cause of portrait mode).

arthurfranca:
> video can be an hour-long thing that will take a lot of time to transcode

@fishcakeday hmm I guess file server may want to pre-generate (I see on-the-fly is just for images) some even lesser quality versions than the main one that it could serve depending on the `?w=<devide-width>` used when client asks for download?

arthurfranca:
@sant0s12 while the server is free to modify further the file after the upload (no one can stop it) and even should if it wants to free  up some space, the user wants to have a way to detect that, just in case the server is swapping their flower image to a porn one ^^

sant0s12:
@arthurfranca Client side checks only make sense under the assumption of immutability. If the server is free to change the image, then the client should not check (and vice versa, i.e., only check when the user previously specified `no_transform`).

I just think that this should be explicit in the NIP.

arthurfranca:
@sant0s12 I mean just that if the user adds `x` tag(s) to the event and if the served image doesn't match it/them, the user can say "I'm not accountable for the image content".

Though yeah user could be adding bogus `x` tag to the event on purpose 🤔️. Well you are right, it doesn't make sense because of that.

Adding `x` tag (or some simple flag tag like `["immutable", "1"]`, as event may already have the ox with the same value) just in case of asking for `no_transform` seems correct.

quentintaranpino (approved):

sant0s12:
@v0l
While I disagree that expecting immutability when not specifying `no_transform` makes little sense, I don't care too much. My main issue is that I think that what you just said is not clearly codified in the NIP and should be included to avoid confusion.

arthurfranca:
Yeah the transformed file hash has its uses. Sometimes I forget important details and jump to conclusions. It is just that it is not exactly a proof that the server changed the file after the upload because user can lie when setting the `x` tag to the event.

96.md
v0l: @arthurfranca look ok now?

arthurfranca: perfect

arthurfranca:
imo ready to merge

v0l:
added to void-cat-rs: https://git.v0l.io/Kieran/void-cat-rs/commit/5547673331a8ffebe73df7139eca2bfaef5357c4

quentintaranpino:
Hello

@nostrband @v0l @sant0s12 @fishcakeday @Semisol @vitorpamplona 

Added **list files** API and **no_transform** to nostrcheck.me, the other instances (those installed by others) will have it implemented as soon as version 0.6.0 is released in 2-3 weeks.

PR #1242, opened 2024-05-17
External computation off relays

AndrewRyanChama:
This nip describes a way to perform computation externally, and communicate the results over nostr.  Computation is necessary for certain functions, and nip-29 proposes an alternative model of computation where the relay acts as the computer.  Unlike nip-29, this nip proposes that computation should be owned by a keypair, while the inputs and results are decentralized across multiple relays.

A Computer can be used to add interactive functionality to existing nips

- nip-72: A computer can manage the `kind:34550`, giving a way to edit the community outside of just a single private key.  The computer can also automate the approval of posts with `kind:4550` as timeline state.
- #875 : The computer can validate the claim tag tag and automatically grant membership.
- (upcoming nip made by me)

### DIFF
--- a/xx.md
+++ b/xx.md
@@ -1,0 +1,80 @@
+
+NIP-XX
+======
+
+External Computation Off Relays
+-------------------------------
+
+For many social applications, implementation requires the execution of a state machine. For example, a group messaging application requires the handling of invites, bans, permission levels, and invite links. Each of these changes depend on the current state of group membership to be possible and will in themselves change the state of the group membership.
+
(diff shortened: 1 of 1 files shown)


vitorpamplona:
Isn't this [NIP-90](90.md)?

AndrewRyanChama:
@vitorpamplona @Semisol This may seem similar to nip-90/DVM on the surface, but it's actually quite distinct

Overall nip-90 is a sort of rpc job model where where users use nostr as a transport layer to call a different service.  The user makes a request and gets a response back.  The example given is that a user will request image generation and get the image back.

This nip is for running a state machine *on nostr* for enablement of social features, such as group user management and moderation.  Computation is modeled as several independent state machines, and each state machine includes an overall state object as well as a timeline view that can be paged through.

The critical parts of this NIP that make it work is the use of the 3XXXX parameterized replaceable events to store the state, as well as a way to page through the timeline events.
This enables the computer to assist with creation of nip-72 moderation events, as well as managing the state of group membership with #875 invite requests.  This would not be possible with nip-90

PR #1243, opened 2024-05-17
Fix connection string protocol description.

braydonf:
The example was fixed yesterday (https://github.com/nostr-protocol/nips/commit/b1f771302a6a1a94c888b619e08a9c302df85fc2) however the description remained the same. 

Without the `//` several Nostr clients will either error and say to include `//` (e.g. Stacker News) or silently fail (e.g. Amethyst).

Perhaps both should be acceptable, however most implementations assume the use of `//`.


AsaiToshiya:
I think this needs to be fixed here as well:

https://github.com/nostr-protocol/nips/blob/bff2e9a28adcb5a6935157441825ca4fe1329a25/47.md?plain=1#L405

PR #1214, opened 2024-05-02
NIP-71: NUDs

fiatjaf:
Permissionless NIPs.

https://github.com/nostr-protocol/nips/blob/nuds/71.md

@pablof7z @mleku @jb55 @staab @vitorpamplona @Semisol @arthurfranca @mikedilger

PR #1251, opened 2024-05-23
Add CIP-01

staab:
Adds a link to my [custom feed nip](https://wikifreedia.xyz/cip-01/97c70a44366a6535c1) over on wikifreedia. See https://github.com/nostr-protocol/nips/pull/1181 for more details and discussion


arthurfranca:
What about using #1214? ~CIP-01~NUD: Custom Feed (singular instead of plural as with wikipedia stuff)

staab:
Forgot about that, updated

fiatjaf:
Are we NUD yet?

PR #1254, opened 2024-05-24
NIP-71: remove `aes-256-gcm` tag

AsaiToshiya:
For the same reason as #948.

> 1. No one seems to be using it.
> 2. People who wish to use encryption can now use the new NIP-44 scheme.
> 3. The existence of the tag as is creates more questions than solutions.


fiatjaf:
I think having events published is not enough for us to say this stuff is actually being used. It's still more in an experimentation phase than anything, so it should be ok to change. Unless I'm mistaken.

PR #1255, opened 2024-05-24
Mention extra metadata fields may be set

patrickReiis:
When someone new wants to learn about Nostr, most probably this person will start reading by order starting with NIP-01, therefore it is good to keep some NIP utilities mentioned so the user can figure it out about them in a fluid way, without having to be inspecting all the NIPs list.


patrickReiis:
Thanks for merging & reviewing! @alexgleason @AsaiToshiya @staab

PR #1256, opened 2024-05-24
Right to Vanish

vitorpamplona:
Adds a special event kinds for relays to allow for 
- Full deletion of an account to specific relays 
- Full deletion of an account to ALL relays

Read [here](https://github.com/vitorpamplona/nips/blob/right-to-vanish/62.md)


62.md
```diff
@@ -0,0 +1,63 @@
...
+  "pubkey": <32-bytes hex-encoded public key of the event creator>,
+  "tags": [],
+  "content": "<reason or note>",
+  //...other fields
+}
+```
+
+The tag list MUST be empty.
+
+Content MAY include a reason or a legal notice to the relay operator.
```
fiatjaf, line 55:
```suggestion
Content MAY include a reason to the relay operator. 
```

fiatjaf (approved):

fiatjaf:
@rabble @arthurfranca

arthurfranca:
Starting over from scratch from one point in time onwards is a good feature.

Considering that an user could change their mind by using a `kind:5` to delete the `kind:62`, scoping it to a relay is ok too. Or else user could be just punising theirself by forbidden backing up their old events on that one relay, forever.

@vitorpamplona have you considered merging the two kinds into a single parameterized replaceable event with the `d` tag set to `""` if for all relays or to `"<relay-url>"` if for a specific relay? It feels like a good fit.

RandyMcMillan:
Add a switch so that if tags is empty - delete all

if a tag list is included - delete only those tags associated with the pubkey.

a similar switch for removal of specific types can be implemented as well.

sant0s12:
I also like @arthurfranca's idea but instead of "" meaning all, have something like "ALL_RELAYS".
This is a harder mistake to make than using kind 63 instead of 62 and avoids having an extra kind.

mplorentz (approved):
This is nice! Required, I think, for relays in the EU that want to be GDPR compliant. Also the Apple App Store has a rule that you must offer the user a way to delete their account, and this seems like a best-effort way to adhere to that rule, given that you can't force other people to delete your data.

mplorentz:
CC @jb55 I think you publish some kind of account deletion event in Damus to comply with the Apple rules. Curious what you think of this.

arthurfranca:
> Ok, unified the two kinds in one with the ALL_RELAYS tag as requested

But it was supposed to be a `d` tag instead of `relay` one, a parameterized replaceable event (PRE).

A PRE fits nicely because the proposed event is unique per relay url.
Also, the fact that user events get deleted from the oldest one up until the proposed event's `.created_at` timestamp, makes it ok to overwrite the timestamp with a newer `.created_at` value from a PRE replacing an older version of itself.

vitorpamplona:
> But it was supposed to be a d tag instead of relay one, a parameterized replaceable event (PRE).

It can't. The event can have more than one target relay.

arthurfranca:
> Relays MAY store the deletion request for bookkeeping and ensure past events are not re-broadcasted into the relay.

If I'm relay "A", references to relays "B", "C", ..., "Z" are bytes I don't want to store.

Contrary to what happens in most other use cases, this one benefits from splitting many references into individual events.

vitorpamplona:
> If I'm relay "A", references to relays "B", "C", ..., "Z" are bytes I don't want to store.

If that is the case, you probably also don't want to maintain an entry in the replaceable address index for this event in case the user sends a new one. 

Deletion events should not be replaceable.

arthurfranca:
> informs a specific relay to delete everything [...] until its .created_at

An event replacing the older one would effectively extend the "until" period. Isn't it desired?

arthurfranca:
Replaceable events infra is already in place (NIP-01), no extra lines of code for the replacement logic.

vitorpamplona:
> Replaceable events infra is already in place (NIP-01), no extra lines of code for the replacement logic.

My point is that we don't need to do replacement + deletion logics. Deletion logic is already enough. 

Clients that don't implement replaceable events don't need to implement it.

arthurfranca:
Though.. a newer `d-tag=ALL_RELAYS` may replace an older `d-tag=my-relay`
And there is wss:// vs ws://, also trailing slash or not.

You have a point, the replacement logic can't rely solely on PRE flow.

arthurfranca:
> Deletion logic is already enough.

Got it, if it replaces everything that came before, it replaces older 62. I was wrong.

NfNitLoop:
I don't really believe in a universal "right to vanish" for distributed protocols.

1) Cache invalidation is a very complex problem. Throwing a "delete" event (be it kind 5, or this kind, or others) into the ether gives users a false sense of security.  OK, you've *said* you want to delete some/all events, but do all the relays that store your events implement this NIP? Do they store the deletion request[^1] indefinitely so that things don't show up again? Did your message even make it to all the relays where your content was cached? If relays are free to not store deletion requests[^1] or not implement this NIP, you don't actually get a full "delete".

2) Single/Mass deletion gives a false sense of security in the case of key compromise. If a user sees a post that they didn't post, and they delete it, they may think they've "dealt with" the issue, when in reality their key is still compromised.

3) It's also an avenue for social abuse. Users can post awful content, then just delete it after the fact to avoid any consequences. Social media is *social*. If we have a conversation and you say awful things, you shouldn't get to just mind-wipe the internet of your actions.

4) It can also introduce a potential for abuse. If relays DO store deletion events, now users can just fill relays with deletion events which they have to hold forever so that those events don't come back?

---

All that said, I suppose for those who *do* want to allow for deletion, or those that *have* to in order to comply with some laws, it's good to have a standard for this.

Some feedback:

I feel like "Delete Account" is a bad name for this. You're still going to have your pubkey, and all of your followers. And all of your content, on relays that aren't targeted.  It's more like "Delete Content Before X From Server Y".  And "Right To Vanish" is … just the same thing targeted at all relays. Maybe "Delete Content [From Relay]" and "Reset Account" are better terms for what these two cases accomplish?

If you want a real [Delete Your Account][2], it should be an irrevocable revocation of the pubkey. Something like:

```
{
  "kind": whatever, // new kind # for pubkey revocation?
  "pubkey": pubkey,
  "content": seckey
}
```

Advantages:
1. There's no coming back from this one. The secret key is out there, so anyone can now impersonate that account.
2. The former user of the account gains plausible deniability for any old content. It could've been forged and backdated with their secret key, which is now in the public domain.
3. Relay operators will *want* to store these kinds of revocations and delete other events, because they are possibly inauthentic, since the key is in the public domain.
4. Key holders are less likely to abuse this than kind 5/62 deletion events, because the cost is losing their social network and truly starting over again from scratch.




[^1]: re: "Relays MAY store the deletion request for bookkeeping and ensure past events are not re-broadcasted into the relay."

[2]: https://x.com/HillaryClinton/status/740973710593654784?lang=en

IngwiePhoenix:
If I was a maintainer, I'd ACK.
This'd be neat for GDPR compliance, of sorts, and can easily be advertised by relay documents. 👍

Neigdoig:
I'm actually surprised that this idea was considered, despite the thought it wouldn't happen. Yes, you can thank me for that idea.

erskingardner:
I like this. 👍

alltheseas:
It's a welcome feature.

Given the recent terminology change on [nip-09 ](https://github.com/nostr-protocol/nips/commit/0ee877275a339d7a29915ddb6a4b37feb714c61d) from "delete" to "deletion request" is any terminology update needed here? 

E.g. "right to vanish" -> "request to vanish"

bezysoftware:
Request to vanish sounds good

vitorpamplona:
Adjusted

mplorentz (commented):
Oops, somehow I posted a review on this PR which was intented for https://github.com/planetary-social/nos/pull/1518. Anyway we are implementing this in Nos now. strfry PR to follow. I would like to see this merged in the near future.

62.md
```diff
@@ -0,0 +1,59 @@
+NIP-62
+======
+
+Request to Vanish
```
kehiy, line 4: this mostly called right to be forgotten in the internet. i think we can consider this as a better name. RTBF.

62.md
```diff
@@ -0,0 +1,59 @@
...
+
+`draft` `optional`
+
+This NIP offers a Nostr-native way to request a complete reset of a key's fingerprint on the web. This is a legally binding procedure in some jurisdictions and thus supporters of this NIP should truly delete events from their database. 
+
+## Request to Vanish from Relay
+
+Kind `62` requests a specific relay to delete everything, including [NIP-09](09.md) Deletion Events, from the `.pubkey` until its `.created_at`. 
+
+```js
```
kehiy, line 15:
```suggestion
```jsonc
```

62.md
```diff
@@ -0,0 +1,59 @@
...
+
+Paid relays or relays that restrict who can post MUST also follow the request regardless of the user's status. 
+
+Clients SHOULD send this event to the target relays only. 
+
+## Global Request to Vanish
+
+To request ALL relays to delete everything, the event MUST include a `relay` tag with the value `ALL_RELAYS` in uppercase. 
+
+```js
```
kehiy, line 47:
```suggestion
```jsonc
```

62.md
```diff
@@ -0,0 +1,59 @@
...
+  "pubkey": <32-byte hex-encoded public key of the event creator>,
+  "tags": [
+    ["relay", "ALL_RELAYS"]
+  ],
+  "content": "<reason>",
+  //...other fields
+}
+```
+
+Clients SHOULD broadcast this event to as many relays as possible.
```
kehiy, line 59: is it ok with line 41?

62.md
vitorpamplona: We can do either. I was reticent to use "forgotten" because deleting stuff from relays doesn't mean you will be "forgotten". It just means that your data was deleted from that instance or instances that received this event and comply to it.

62.md
kehiy: makes sense. thanks a lot.

62.md
kehiy: i agree and it's logical. but i think this is same for other stuff on web and internet, and using `forgotten` just makes it easier ti understand for people from everywhere. im more agree with `forgotten` or maybe something between. we can think more about the name.

kehiy: found this: https://en.wikipedia.org/wiki/Right_to_be_forgotten.
they call it vanish as well. we can use one of them and include other one on details of standard. maybe.

bezysoftware:
What would deleting a `kind:62` event do? Would it basically allow user to republish their old (backed up) events?

vitorpamplona:
Either that or we just don't allow deletions. Deleting the kind 62 allows the user the fix potential mistakes, but it also allows an attacker to got access to an nsec to restore everything to its full glory. 

I am leaning more towards disallowing kind 62 to be deleted in the same way we don't allow kind:5 to delete other kind 5s

bezysoftware:
> Either that or we just don't allow deletions. Deleting the kind 62 allows the user the fix potential mistakes, but it also allows an attacker to got access to an nsec to restore everything to its full glory.
> 
> I am leaning more towards disallowing kind 62 to be deleted in the same way we don't allow kind:5 to delete other kind 5s

Agree with disallowing to delete kind 62. Either one should be probably mentioned in nip to avoid ambiguity

mplorentz:
This NIP has been deployed in the latest version of Nos and [relay.nos.social](https://github.com/planetary-social/nosrelay). 

Unfortunately we didn't make a PR directly to strfry. We instead went with a strfry plugin that pushes kind 62 events to a redis stream where they are processed by all our systems (NIP-05, push notification service, and a script that deletes events from strfry). But it's all open source and we have a docker compose file if anyone wants to copy it.

@vitorpamplona have you added this to Amethyst yet?

alexgleason (approved):
I think this is needed and the framing of it is right.

vitorpamplona:
> @vitorpamplona have you added this to Amethyst yet?

Not yet, but will do soon. 

I usually wait for somebody else to code the NIPs I propose to confirm that my idea is not just complete trash. :)

staab:
This has the same problem that allowing tagged users to delete events has, which is that it completely breaks all use of shared keys. In #875, all members of a group share a key. This allows anyone to post to the group, but with this PR it would allow all members of the group to delete all the other members' posts. While NIP 87 is likely going away, I also don't think it's the only valid use of a shared pubkey. We should see if the gift-wrap part of this PR can be handled some other way.

62.md
```diff
@@ -38,6 +38,8 @@ Relays MAY store the signed deletion request for bookkeeping.
 
 Paid relays or relays that restrict who can post MUST also follow the request regardless of the user's status. 
 
+Kind `5` deletions MUST not delete kind `62`s.
```
dcadenas, line 41: Should we add that kind `62` should not delete kind `62` for the same reasons?

bezysoftware: I think it's fine that a newer 62 deletes an older 62. It effectively supercedes it

dcadenas: So is it implied that kind `62` events should be kept and be the only event that is kept in the relay for the deleted pubkeys? If that's the case it seems there's no clarification about that in the copy. In other words, `MUST not delete kind 62s` seems to hint that the vanish requests should be kept. Maybe the intent was that it should be kept until being processed?

62.md
```diff
@@ -0,0 +1,61 @@
...
+}
+```
+
+The tag list MUST include at least one `relay` value. 
+
+Content MAY include a reason or a legal notice to the relay operator. 
+
+Relays MUST fully delete any events from the `.pubkey` if their service URL is tagged in the event. 
+
+Relays SHOULD delete all [NIP-59](59.md) Gift Wraps that p-tagged the `.pubkey`, deleting all DMs to the pubkey.
```
bezysoftware, line 33: 1. Should this also mention the deprecated `kind:4`? It's basically the same thing right?
2. This wording implies tagged gift wraps can be deleted even if their service URL is not tagged in the event (this is mentioned only for deleting by `pubkey`). Is this intended? If so, why?

62.md
bezysoftware: The idea is that the vanishing request cannot be undone. Whether the event itself is kept by the relay should be an implementation detail.
@vitorpamplona perhaps this could be reworded same way as in NIP-09 where it says `Publishing a deletion request event against a deletion request has no effect. Clients and relays are not obliged to support "unrequest deletion" functionality.`?

vitorpamplona: Just adjusted

62.md
vitorpamplona: No, kind:4s that were sent to you are owned by the sender. GiftWraps are not owned by anyone (random signer).

vitorpamplona: adjusted the wording for DMs

bezysoftware:
This has been deployed to https://relay.netstr.io/ 
I wrote a few gherkin scenarios for this NIP if anyone interested: https://github.com/bezysoftware/netstr/blob/main/test/Netstr.Tests/NIPs/62.feature

staab:
> it completely breaks all use of shared keys

Never mind, shared keys are already broken. What we need to fix it is a mechanism to say events _can't_ be deleted by the author, or the recipient. I rescind my objection.

kehiy:
i think it can be replaced by #1509 since it would support deletion for all events as well. to specify relays we already have nip-70 and `"-"` tag which relays can interpret. adding new rules to do the same thing is unnecessary complexity imho.

staab:
I support this, similar to #377 though it should probably allow users to specify all past data, or a complete burn of their key.

staab:
@kehiy I think the filter version is actually much more complex (it leaks metadata about deleted events, how is "search" interpreted, it's not easily indexable, it's going to be more computationally intensive for relays to implement), and isn't really justified by use case. The way I see it, we need the following types of deletes:

- Delete this event. Bulk deletes are weird, if you published a bunch of events, just tag each one. It's still much lighter than the original payloads were. With one exception...
- Delete everything on my pubkey until now. A "delete account" option. If this event uses the `expiration` tag, you could even (optionally) allow re-publishing old events after the delete expires.
- Burn this pubkey. A permanent "delete account" option, for when a key is compromised.

staab:
> soft-bricking the npub

I think this is actually a desirable feature. I agree that it would suck to do this accidentally, but it's probably better to have it.

kehiy:
> @kehiy I think the filter version is actually much more complex (it leaks metadata about deleted events, how is "search" interpreted, it's not easily indexable, it's going to be more computationally intensive for relays to implement), and isn't really justified by use case. The way I see it, we need the following types of deletes:
> 
> * Delete this event. Bulk deletes are weird, if you published a bunch of events, just tag each one. It's still much lighter than the original payloads were. With one exception...
> * Delete everything on my pubkey until now. A "delete account" option. If this event uses the `expiration` tag, you could even (optionally) allow re-publishing old events after the delete expires.
> * Burn this pubkey. A permanent "delete account" option, for when a key is compromised.

i dont think if it's too complex. you can look at my implementation on immortal which currently performs the deletion of current events well. soon ill support keeping the deleted filter as well. i believe both of them at the same time won't be too complex. about performance, im using a mongo db with multiple database calls and it still performs well. soon ill refactor it to make on call and do the whole deletion which is much more efficient. and it would support all three cases you mentioned and more.

this article explains this `filter` tag better and answers your question about search and other fields on deletion filter:

https://njump.me/nostr:naddr1qq2nyv3ndf342d6ewcm52dny2d2rvc26w4452qghwaehxw309aex2mrp0yhxummnw3ezucnpdejz7q3qhtdamfg82u4nj7zjqj82wnew7wke9vd2cp7r6nsaast5arxujc4qxpqqqp65wvdt5af

kehiy:
> There must be a "proof of freshness", to prevent dating events to in the future, and soft-bricking the npub. A public key should not be able to be irreparably damaged due to a malicious or defective client (outside of full compromise of the nsec).

one of good approaches for this is @fiatjaf's idea which was using relay contact and asking them for a change. if someone accidentally freezed, deleted, limited their public-key.

> the main idea was asking the relay operator for account deletion itself from fiatjaf on nsf meeting. but this would work for denial of a deletion or freeze. for deletion itself i like the filter model.

kehiy:
> I find it hard to see how relays will honor a feature that can be abused to hold anyone's npub by ransom just because your web client was able to sign an event. Most people blindly allow apps to sign all kinds of events, and most extensions don't even offer a per-kind option.

if you mean in case of filter deletion, relay can set a min pow for deletion request or ask for payments. to prevent huge checks and resources usage.

kehiy:
> Asking relay operators to undo it doesn't scale though.

i agree.

vitorpamplona:
> I was referring to a malicious web client signing a delete everything event on behalf of someone that is dated to in the future, and holding their identity for ransom

I am not sure how proof of freshness solves this. If there is such a client, they can do the same with the past events. They can sign and wait to use in the future as well. The same client can start posting stuff the user didn't approve. If the issue is ransom from a malicious client and that client has the key, I don't think there is anything we can do to protect the user. They can literally do whatever they want with the account.

kehiy:
@Semisol if your signer/client publishs unexpected deletions on your behalf, then it would do more as well.

check the code, use a safe signer, restrict writing levels to be safe from this. we already have people who loose their follow lists. ive got my profile cleaned multiple time juts because some old web clients considered my old pubkey kind 0 as their base to create new one.

mikedilger:
Implemented in chorus.

staab (approved):

62.md
```diff
@@ -0,0 +1,61 @@
...
+
+Relays MAY store the signed request to vanish for bookkeeping.
+
+Paid relays or relays that restrict who can post MUST also follow the request to vanish regardless of the user's status. 
+
+Publishing a deletion request event (Kind `5`) against a request to vanish has no effect. Clients and relays are not obliged to support "unrequest vanish" functionality.
+
+Clients SHOULD send this event to the target relays only. 
+
+## Global Request to Vanish
```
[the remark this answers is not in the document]
mikedilger: Clients may add that. Relays shouldn't require it.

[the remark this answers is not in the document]
kehiy: i mean something like this:

a relay receives a right to vanish. is it protected? so ask for auth. its not? so execute it.

then we don't need to have `"relays"` tag.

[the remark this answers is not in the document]
vitorpamplona: The relays tag exists to avoid client mistakes. Most clients auth automatically. So if they happen to re-broadcast the event to a different relay, they would also auth without even thinking that their actions are changing what the user wanted (deletion in just the previous relay). 

The meaning of what the author wanted to do should not be dynamic (auth-based).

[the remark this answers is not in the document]
kehiy: makes sense. should a relay black list the user to prevent future incoming events from them as well?

kehiy (approved):
implemented on https://github.com/dezh-tech/immortal/pull/119

62.md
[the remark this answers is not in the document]
mikedilger: Yes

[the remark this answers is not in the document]
mikedilger: > Relays MUST ensure the deleted events cannot be re-broadcasted into the relay.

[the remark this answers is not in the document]
vitorpamplona: Do it in a similar way we delete replaceable events: everything up to the deletion event created_at gets blocked but the author can use the key after that date.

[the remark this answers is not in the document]
kehiy: thanks. its already done.

staab:
This is implemented in flotilla.social

mikedilger:
@vitorpamplona can you fix merge conflicts?  This is implemented in at least 2 places and has 5 approvals, I think it should be merged.

vitorpamplona:
Good to go!

quentintaranpino:
This will be implemented in the nostrcheck-server relay module.

cypherhoodlum:
This will be implemented on Nestr github-like client

dtonon:
It seems a quite dangerous NIP, a bugged or malicious client can definitively nuke your profile just using a remote signer. Scary.
I would at least add a delay to the final action (e.g. 24h window) and a mechanism to undo the request.

staab:
> I would at least add a delay to the final action (e.g. 24h window) and a mechanism to undo the request.

I agree. I also don't think the "undelete" part of the NIP makes sense, because this NIP is not a defense against a leaked key. But I don't feel strongly enough to open a PR.

vitorpamplona:
> I would at least add a delay to the final action (e.g. 24h window) and a mechanism to undo the request.

I think that is a relay operator's decision to offer or not the 24h window.

mikedilger:
If we had master keys and sub-keys, this could require the (more secure) master key to avoid the danger since clients wouldn't be given the master key. But we still have no movement towards a master-key sub-key system. So currently irrelevant.

Issue #225, opened 2023-02-06
Questioning JSON peculiarities


burdiyan:
This was initially part of https://github.com/nostr-protocol/nostr/issues/127, but I decided to separate it as these more subjective concerns of mine.

I don’t know if I’m the only one here that thinks this, but I found many things in Nostr's requirements about JSON to be peculiar and unusual. I also couldn’t find any justifications for these choices in the NIPs, so maybe it would be worth adding something to make it clearer.

### Tags

Why event tags aren’t just fields on the JSON instead of being an array of arrays? The most confusing part for me was to find that inside the subscription query request, they are indeed fields on the JSON prefixed with `#`. Why couldn’t they be the same in the event structure?

Something like:

```
{
    "id": "...",
    "kind": 1,
    "content": "...",
    "#p": [[<public key>, ...]], // Or object instead of array.
    "#e": [<event-id>, ...],
    ...
}
```

The fact that tags are “tuples” (in fact they are like function calls with some expected [arity](https://en.wikipedia.org/wiki/Arity)) makes them much harder to work with and to evolve, IMO. E.g. the `p` tag can have 2 or 3 “arguments” (with or without petname). I think this would be so much easier to extend and work with if they were JSON objects instead of tuples. In typed languages like Go, people normally prefer to work with structs rather than tuples. So converting these arrays into structs would require custom marshaling/unmarshaling code, which is always cumbersome. Having these as optional top-level fields in the event's JSON object would be so much explicit and easier to work with, IMO.

### Structured Content

The most surprising and confusing part for me was to find out that content is strictly defined as string. So, when working with structured content you’d have to encode JSON as a string and put it into something that’s already a JSON, which feels a bit weird.

Again, I’m sure there’re reasons for these choices, but I couldn’t find any in the NIPs.

fiatjaf:
I agree the choices feel odd, but they were made such that the event structure was predictable and static and that hashing the event was simple and yielded a deterministic result.

The `#` in filters exists to denote that these keys represent tags.

Also for some weird reason these "tuples" or "function calls" appeal to me very much, because there is no doubt about the name of the property, you can't type the wrong name and get away with it, you don't have to memorize the names, and also you don't have to transfer the names around.

burdiyan:
> I agree the choices feel odd, but they were made such that the event structure was predictable and static and that hashing the event was simple and yielded a deterministic result.

[...]

PR #1258, opened 2024-05-25
NIP-XX: Improved event signing scheme

sant0s12:
Alright buckle up, here I come with a wacky idea that I found out through #225. When I saw it at the time I thought that it would break backwards compatibility but to my surprise I think it does not have to.

The basic idea is to have a new signing scheme that allows for arbitrary JSON fields, which IMO would make better use of the fact that nostr events are just JSON objects than using tags as is done now. The current signing scheme is still used but I propose adding a new field that holds the signature over all of the objects fields as done by [Perkeep](https://perkeep.org/doc/json-signing/). Therefore, existing clients can continue to work the same.

I thought about doing the same thing for the Id field since it is currently also not aware of any other fields besides the ones described in NIP-01. However, I am unsure about the utility of this since it would increase the size at no clear advantage.

Please let me know your thoughts and concerns.

### DIFF
--- a/XX.md
+++ b/XX.md
@@ -1,0 +1,51 @@
+NIP-XX
+======
+
+Improved event signing scheme
+----------------------------------------------------
+
+`draft` `optional`
+
+This NIP describes a new event signature scheme that provides greater flexibility
+than the existing scheme by allowing signing JSONs with arbitrary properties while
+providing backwards compatibility. It is based on [Perkeep's JSON signing](https://perkeep.org/doc/json-signing/).
+The signature scheme remains the same as the one described in NIP-01.
+
+## Signing
+
+This NIP adds a new signature property to the event object: `sig_v2`.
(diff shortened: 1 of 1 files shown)


vitorpamplona:
> I am unsure about the utility of this since it would increase the size at no clear advantage.

Correct. I don't see the point of this. 

There are major problems on today's hashing/signing procedure: 
1. Requires JSON parsing and serializers which are extremely slow. 
2. Requires assembling the NIP-01 array to hash, which uses a TON of memory creation and deletion just to hash it. 
3. SHA256 can process batch arrays, but batch processing is slower than a call with the complete byte array. 
4. We use Hex, which is a wasteful format for keys inside a JSON. 

Those are extremely important limitations since we are expected to re-hash and verify every single event coming in. I would try to create the fastest verifier if I had a chance to re-do nostr.

Issue #1095, opened 2024-03-01
NIP-46 still uses NIP-04 encryption


alexgleason:
NIP-04 is considered harmful. It could leak your secret key. We did so much work to ensure we made the right choices this time in NIP-44 and even did a proper security audit.

NIP-46 still depends on the insecure encryption. This goes against the entire fundamental purpose of NIP-46, which is to protect your private key.

There was pushback against changing NIP-46 because it would result in breaking changes. But a breaking change to NIP-44 is necessary. Otherwise why did we do all of that? And what's the point of protecting your key using encryption that could reveal your key?

So now we are stuck: not wanting to change NIP-46 so it doesn't have breaking changes, but using an outdated and deprecated form of encryption that goes against the goals of NIP-46.

What should we do?

mekefly:
赞成修改，现在使用人数少的时候修改是成本更小的

fiatjaf:
I'll believe it is insecure when I see it.

mekefly:
Agree to make change,or, we can create a new nip and write a unrecommended  on nip46,I will follow up and modify my client

alexgleason:
> I'll believe it is insecure when I see a stolen key.

Then why did we do all of that? I thought it was the whole point.

cc @paulmillr

paulmillr:
@fiatjaf 

>I'll believe it is insecure when I see a stolen key.

Then it would be too late.

You know there are people who can steal your keys using radio, right? https://eprint.iacr.org/2015/170 Have you “seen” it?

Why did Signal and iMessage switch to post quantum cryptography? There are no pq computers today. Have you seen them? 

@alexgleason i’m a security guy who saw a problem and wanted to get it done. If the community is not interested in this,  there’s not much I can do 🤷‍♂️ It was complicated enough to even release nip44 out. I don’t want to spend more time agitating for “using” it.

what can be done by someone: modifying 46 to include proper encryption.

mikedilger:
I wouldn't worry about the breaking change. NIP-46 was broken recently. What does it matter that we break it again? Just please document it in BREAKING.md (which isn't being kept up to date so far)

tyiu:
[...]

PR #1259, opened 2024-05-26
NIP-46, NIP-90: "encrypted" tag to indicate the encryption method used

alexgleason:
I remembered that NIP-90 has an "encrypted" tag with no value, and it got me thinking we could utilize that tag as a more general-purpose "hey the thing in the content is encrypted" indicator, for events that may or may not have encrypted data.

Then I realized, since it has no value, we could add a `"nip04" | "nip44"` to indicate the type of encryption used.

Then I realized we could also use this tag on NIP-46 to help ease the transition in https://github.com/nostr-protocol/nips/issues/1095 https://github.com/nostr-protocol/nips/pull/1248

As I'm typing this I realize, that we could also use this on NIP-51 lists which also have an "optionally encrypted" criteria, and might also want to benefit from knowing which encryption is used.

So basically it's one simple idea that solves a lot of problems.

### DIFF
--- a/46.md
+++ b/46.md
@@ -56,7 +56,7 @@ nostrconnect://<local-keypair-pubkey>?relay=<wss://relay-to-connect-on>&metadata
 {
     "kind": 24133,
     "pubkey": "eff37350d839ce3707332348af4549a96051bd695d3223af4aabce4993531d86",
-    "content": nip04({
+    "content": encrypt({
         "id": <random_string>,
         "method": "sign_event",
         "params": [json_stringified(<{
@@ -66,7 +66,10 @@ nostrconnect://<local-keypair-pubkey>?relay=<wss://relay-to-connect-on>&metadata
             created_at: 1714078911
         }>)]
     }),
-    "tags": [["p", "fa984bd7dbb282f07e16e7ae87b26a2a7b9b90b7246a44771f0cf5ae58018f52"]], // p-tags the remote user pubkey
+    "tags": [
+        ["p", "fa984bd7dbb282f07e16e7ae87b26a2a7b9b90b7246a44771f0cf5ae58018f52"], // p-tags the remote user pubkey
+        ["encrypted", "nip04"] // specifies the encryption used
+    ],
 }
 ```
 
@@ -76,13 +79,32 @@ nostrconnect://<local-keypair-pubkey>?relay=<wss://relay-to-connect-on>&metadata
 {
     "kind": 24133,
     "pubkey": "fa984bd7dbb282f07e16e7ae87b26a2a7b9b90b7246a44771f0cf5ae58018f52",
-    "content": nip04({
+    "content": encrypt({
         "id": <random_string>,
         "result": json_stringified(<signed-event>)
     }),
-    "tags": [["p", "eff37350d839ce3707332348af4549a96051bd695d3223af4aabce4993531d86"]], // p-tags the local keypair pubkey
+    "tags": [
+        ["p", "eff37350d839ce3707332348af4549a96051bd695d3223af4aabce4993531d86"], // p-tags the local keypair pubkey
+        ["encrypted", "nip04"], // specifies the encryption used
+    ]
 }
 ```
+
+#### Encryption
+
+An `encrypted` tag MAY be included in kind 24133 events to specify the encryption method used.
+The tag is in the form:
+
+```
+["encrypted", "<encryption>"]
+```
+
+Where encryption is one of:
+
+- `nip04` - NIP-04 encryption
+- `nip44` - NIP-44 encryption
+
+If the tag is not included, the encryption method is assumed to be NIP-04 for backwards-compatibility.
 
 #### Diagram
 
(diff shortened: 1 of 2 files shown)


alexgleason:
I modified NIP-51 as well.

fiatjaf:
This doesn't solve anything because NIP-44 support still has to be hardcoded.

alexgleason:
The "encrypted" tag lets users (or clients) optionally encrypt data in certain events.

The same tag can also be used to indicate the encryption method used.

Of course people sill have to implement NIP-44. But don't you see the bigger picture of why this tag is a good idea?

To put it another way: should we remove the `encrypted` tag from NIP-90? What would happen if we removed it?

This is consistency and unification of events with optionally-encrypted payloads.

fiatjaf:
I had written a long answer, but decided to not post it. I've written these same arguments way too many times and I don't want to bore the reader.

staab:
I don't think adding a value helps at all, because it's easy to detect the difference between 04 and 44 (04 has `?iv=` in the ciphertext), and nip 44 is versioned.

vitorpamplona:
> (04 has `?iv=` in the ciphertext)

Not only has `?iv` but iv is fixed in length. So, this is faster: 

`v[v.length-28] == '?' && v[v.length-27] == 'i' && v[v.length-26] == 'v' && v[v.length-25] == '='`

alexgleason:
The main thing I was trying to solve is having an optionally-encrypted event that may be either nip04, nip44, or not encrypted at all. Then I realized the "encrypted" tag could be useful to put on any encrypted event. It seems better than detecting the encryption to go with the author's intention.

fiatjaf:
It's not detecting based on the author's intention, but based on a standard everybody follows.

vitorpamplona:
> Then I realized the "encrypted" tag could be useful to put on any encrypted event. It seems better than detecting the encryption to go with the author's intention.

The issue is that the encrypted tag might not be aligned with the `.content` (bugs, idiots, refusal to implement, ...). So, even if you know things have been `encrypted` by a given standard, you still need to check if the encryption is there, by that standard. If that is the case, then just test it directly. There is no need for the flag.

alexgleason:
I feel like yall are missing my point.

I created a new NIP-78 event. This event is **optionally encrypted**:

```js
{
  kind: 30078,
  tags: [['d', 'pub.ditto.pleroma_settings_store']],
  content: JSON.stringify(settingsStore),
}
```

or:

```js
{
  kind: 30078,
  tags: [['d', 'pub.ditto.pleroma_settings_store'], ['encrypted', 'nip44']],
  content: signer.nip44.encrypt(JSON.stringify(settingsStore)),
}
```

To me this is obviously valuable. And the `encrypted` tag can unify NIP-90 and NIP-51 at the very least.

My mistake in this MR was touching NIP-46, which I know brings up a lot of feelings.

vitorpamplona:
I would never use the tag. 

I would just check if: 
1. `.content` starts with `{` -> unencrypted. 
2. `.content` "has" `iv` -> nip04
3. `.content` has nip44 version -> nip44. 

The tag is just noise. 

> My mistake in this MR was touching NIP-46, which I know brings up a lot of feelings.

alexgleason:
@vitorpamplona I coded that and then reverted it because it seems wrong for clients to "guess" about the encryption method being used. It tickles that part of my brain that wonders how I can exploit it, and it seems to not be a best practice regarding security.

Also, if the event content is a string that isn't a JSON-stringified object, an `encrypted` tag can indicate whether the content should be used directly as the subject even if it _looks_ like an encrypted string. I imagine this is why NIP-90 has it - so a "decrypt" DVM could treat the content as the direct subject, for example.

vitorpamplona:
> @vitorpamplona I coded that and then reverted it because it seems wrong for clients to "guess" about the encryption method being used.

I don't think there is an alternative. Resilient clients will have to guess or end up not displaying some content. Knowing the mess already out there, I can't trust the signing client to set the `encrypted` tag correctly. You can't even trust if the tag will be there. So, to me, this is not really an option. You will have to "guess it". And I encourage other clients to do the same.

mikedilger:
> > (04 has `?iv=` in the ciphertext)
> 
> Not only has `?iv` but iv is fixed in length. So, this is faster:
> 
> `v[v.length-28] == '?' && v[v.length-27] == 'i' && v[v.length-26] == 'v' && v[v.length-25] == '='`

Be careful to not run Vitor's code without first checking that the content is at least 28 characters long.

mikedilger:
>     2. `.content` "has" `iv` -> nip04

Future bug: I mentioned kim iversen and then Amethyst broke. It thought the content was encrypted because I typed "iv".

alexgleason:
I should be able to paste an entire nip04 ciphertext into a note without a breaking.

vitorpamplona:
Hum.. I think it's something else. The IV check is not in production yet 🤔

staab:
> I should be able to paste an entire nip04 ciphertext into a note without a breaking.

Coracle detects encryption _type_ based on `iv=?` but only attempts decryption in situations where encryption makes sense. Encrypting kind 1 contents without wrapping via nip 59 would be very weird behavior. For kind 3s, `content` is either json or encrypted tag lists (that's what the spec says, but does anyone do that?), so if you try to parse json first, you won't have any problems. So in practice I don't think there are any collisions, and going forward NIP 44 ciphertext is much easier to identify.

PR #1266, opened 2024-05-28
NIP-116: Event paths

mmalmi:
Event kind `30079` denotes an event defined by its *event path* rather than its event kind. Event paths are a scalable alternative to numeric event kinds, and help make Nostr a true "everything app".

### DIFF
--- a/116.md
+++ b/116.md
@@ -1,0 +1,56 @@
+NIP-116
+=======
+
+Event paths
+-----------
+
+### Description
+
+Event kind `30079` denotes an event defined by its *event path* rather than its event kind.
+
+The *event directory path* is included in the event path, specified in the event's `d` tag. For example, an event path might be `user/profile/name`, where `user/profile` is the directory path.
+
+Relays should parse the event directory from the event path `d` tag and index the event by it. Relays should support "directory listing" of kind `30079` events using the `#f` filter, such as `{"#f": ["user/profile"]}`.
+
+For backward compatibility, the event directory should also be saved in the event's `f` tag (for "folder"), which is already indexed by some relay implementations, and can be queried using the `#f` filter.
+
+Event content should be a JSON-encoded value. An empty object `{}` signifies that the entry at the event path is itself a directory. For example, when saving `user/profile/name`: `Bob`, you should also save `user/profile`: `{}` so the subdirectory can be listed under `user`.
+
+In directory names, slashes should be escaped with a double slash.
+
+### Example
+
+#### Event    
+```json
+{
+    "tags": [
+        ["d", "user/profile/name"],
+        ["f", "user/profile"]
+    ],
+    "content": "\"Bob\"",
+    "kind": 30079,
+    ...
+}
+```
+
+#### Query
+```json
+{
+    "#f": ["user/profile"],
+    "authors": ["[pubkey]"]
+}
+```
+
+### Motivation
+
+To make Nostr an "everything app," we need a sustainable way to support new kinds of applications. Browsing Nostr data by human-readable nested directories and paths rather than obscure event kind numbers makes the data more manageable.
+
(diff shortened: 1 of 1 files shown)


arthurfranca:
> Event kind `30079` denotes an event defined by its *event path* rather than its event kind.

But if you want interoperability, don't you need to reserve the `event path` for a specific usage (by merging a NIP or writing a NUD)? If yes, isn't this the same as reserving an `event kind` number?

vitorpamplona:
> Event paths are a scalable alternative to numeric event kinds

We do not have a scaling problem in numeric kinds. Pretending there is one is not helpful.

But I am not opposed to creating a directory structure on an event kind. It's just not solving what you think it is solving. 

Either way, directories are very similar to #784, you can just use `n` instead of `f`. 

> Using `d`-tagged replaceable events for reactions, such as `{tags: [["d", "reactions/[eventId]"]], content: "\"👍\"", kind: 30079, ...}` would make un-reacting trivial: just publish a new event with the same `d` tag and an empty content. Toggling a reaction on and off would not cause a flurry of new reaction & delete events that all need to be persisted.

This is dumb. It's just going to create a ton of empty events that client's don't need to download. It's better to have proper deletions than an empty event. See #1263

AsaiToshiya:
If the interoperability is not required, I think #995 will help.

mmalmi:
> > However, it avoids the situation where a user toggles e.g. "like" button on and off, creating lots of "like" and "delete like" events that relays have to store.
> 
> I don't get this. Relays don't have to store these likes and deletions in the same way they don't have to store the history of replaceables (which are also going back and forth on toggles). They can just delete them forever. It's up to them.

If the deletion event is not stored, the deleted event can reappear from a relay or client that doesn't support deletions or just hasn't received the deletion.

fiatjaf:
#1263

mmalmi:
I could use the existing event kind `30078` and possibly `n` tags instead of `f`.

Then the only relay side change in this NIP would be this:

> Relays should parse the event directory from the event path `d` tag and index the event by it. Relays should support "directory listing" of kind `30079` events using the `#f` filter, such as `{"#f": ["user/profile"]}`.

But it can be done later if this arrangement becomes popular. Avoids a little bit of redundancy.

Replaceable events for lists and reactions is more of a client side suggestion. It could be done parallel to current follow lists and profiles. You could restore your follow list or profile from the replaceable event side even if there's a kind 0 or 3 overwrite.

"Directory names" and numeric kinds are both possible on Nostr, so we can see which one works better, and maybe they have different purposes.

vitorpamplona:
> Relays should parse the event directory from the event path d tag and index the event by it. Relays should support "directory listing" of kind 30079 events using the #f filter, such as {"#f": ["user/profile"]}.

What does this mean? What do they actually have to do to support it? Is it a filter with `startsWith` for f-tags?

mmalmi:
> > Relays should parse the event directory from the event path d tag and index the event by it. Relays should support "directory listing" of kind 30079 events using the #f filter, such as {"#f": ["user/profile"]}.
> 
> What does this mean? What do they actually have to do to support it? Is it a filter with `startsWith` for f-tags?

startsWith for d-tags. It would just save the redundancy in tags: instead of `tags: [['f', 'user/profile'], ['d', 'user/profile/name']]` you could leave out the 'f' tag altogether.

vitorpamplona:
> startsWith for d-tags.

mmalmi:
Actually not exactly startsWith: "#f": ['user'] should return an event with ["d", "user/profile"] but not ["d", "user/profile/name"]. So it requires a custom index for "f" to be efficient.

> We have been requesting this for months (for geotags), but I don't think it's going to happen :(

Link? Sounds like this should be an efficient startsWith lookup if just "d" is indexed on the relay.

vitorpamplona:
> Link? Sounds like this should be an efficient startsWith lookup if just "d" is indexed on the relay.

It was def in this repo, but I am unable to find it :( 

FYI, pubkey prefixes were also removed for the same reason: https://github.com/nostr-protocol/nips/pull/703

PR #1270, opened 2024-05-30
NIP-54: add `c` tag

AsaiToshiya:


fiatjaf:
I don't think this is a good idea nor I see anyone using it.

This sounds like a `t` tag with a different name.

fiatjaf:
I found two events now from @hzrd149 with a `c` tag `"Nostr"`, with an uppercase `N`. I don't see the point of doing this.

staab:
tags > categories, sorry Pablo.

staab:
Yeah, 31337 audio events also use `c`. This is probably good to merge.

arthurfranca:
Imo instead of categories, articles should [link to other articles](https://github.com/arthurfranca/nips/blob/trust-rank/54.md#closely-related-article-tags).

I explain it better with an example [here](https://github.com/nostr-protocol/nips/pull/1214#discussion_r1621230511) on the NUD discussion.

fiatjaf:
This is ridiculous because it's just `t` tags with a different name. This doesn't pass criteria 2 for acceptance of NIPs.

Also yes, categories in wiki world are just articles and links. We can have articles declare whom they're linking to, if that's what you want, it makes more sense, but I don't think we should have that either.

`c` tags are easily spammable and they require even serious users to be extra diligent in classifying their articles, so it will never work for making actual categories. That's why no one really takes `t` tags seriously as a way to organize things, they're a hint at best. `l` labels (tags assigned by a third-party curator) are probably the best possible solution to this (even if they're still not great).

AsaiToshiya:
Oh, the `c` tag was Wikifreedia's own implementation. I thought it was a consensus specification somewhere.

pablof7z:
> tags > categories, sorry Pablo. @staab 

They (could be) different.

my thinking here is that you can have dozens of different tags in an event, but a `c` should be a single one.

The (very hopeful) reasoning is that constraining the UI to have a single `c` could prevent fragmentation around topics too much and hurt discoverability.

When I spoke with the wikipedia founder categories was a very important topic he kept pointing to and told me that the category list was the front page of wikipedia for years until the woke took over.

pablof7z:
> `c` tags are easily spammable

spam is always possible regardless of `c`, `t` or any letter the alphabet people come up with. If a user spams you mute the pubkey if they are in your WoT. This is completely indepedent of `c`.

> and they require even serious users to be extra diligent in classifying their articles

you say that as if it's a bad thing; if a user has to XOR categorize their article as `History` or `Business & Entrepreneurship` then briefly pausing to consider it is a good thing.

> That's why no one really takes `t` tags seriously as a way to organize things

`t` are, by their very nature, very prompt to fragmentation because of how users interact with them

it's an empty field that takes anything so you might as well put all variations of all possible things remotely related to it

for example, I would tag this issue as:

```js
["c", "nips" ],

["t", "wiki" ],
["t", "nips" ],
["t", "nostr" ],
["t", "nud" ],
["t", "nuds" ],
["t", "nudes" ],
["t", "development" ],
["t", "dev" ],
["t", "wikinips" ],
["t", "nip-54" ],
["t", "nip54" ],
["t", "nip 54" ],
["t", "fiatjaf hates freedom" ],
["t", "hodlbod loves labeling people" ]
```

Obviously there are still a bunch of problems with `c` such as trying to standardize the value (like we currently do with the `d` tag)

vitorpamplona:
From your example, `c` seems to similar to `n`  from #784 to group things in sets.

Otherwise `t`, `c`, and `n` are all syntactically the same but semantically different. Any documentation MUST lay out their semantics very well.

fiatjaf:
Thank you for the explanation, @pablof7z, but I still think it's a bad approach.

It makes much more sense to me for someone else to look at a bunch of articles about different alien civilization in Star Wars written by others and create a _collection_ named "Alien Civilizations in Star Wars", maybe using a NIP-51 list or some other approach, and place them all inside.

Or do it the @staab way and just go around browsing articles and issuing labels to them such that someone else can later browse their labels and see that there are many articles with the same label.

Or perhaps not even that, someone could just create a new wiki article named "Alien Civilizations in Star Wars" and link to those other articles. This has the advantage of being very simple, very manual and do not require any extra software features. I have the feeling that 99% of the people prefer _manual_ things, as they are more human, and it's only programmers who like to create excessive structurization.

staab:
> tags > categories

I stand by this, but I do think categories are conventional enough it would be fine to add this.

PR #1272, opened 2024-05-30
Local Relay list

vitorpamplona:
Adds a local relay list anchored by the device ID from each OS. Ideally, all apps installed on the same device/user can access that relay.


vitorpamplona:
I could use some help on this. I don't know how to get a Device ID from the browser for Clients running in JavaScript. :(

staab:
Why would we need support in the protocol for this? Wouldn't this list only be relevant to the device the user is currently using?

vitorpamplona:
> Wouldn't this list only be relevant to the device the user is currently using?

Yep, it's only for the device, but for all apps running in that device.

vitorpamplona:
For instance, People are using [Citrine](https://github.com/greenart7c3/Citrine) on Android as a local relay. All clients running on Android should enable that relay when using that phone. The idea is to avoid having to set up the relay separately in each client.

vitorpamplona:
>  is there no way to advertise locally running services on a device?

There is on Android and I guess we can use local files on UNIX systems to do so. I have no idea how iOS, Web or Windows Apps will access those files, though. We might need to create a different way to advertise a local relay for each OS. I was trying to just normalize it all in one event.

DanConwayDev:
Can't you just check for a relay running on `localhost:4869`? That's the port both Citrine and https://github.com/CodyTseng/nostr-relay-tray run on.

vitorpamplona:
What happens if the user installs two local relay apps? Or Citrine + Umbrel?

DanConwayDev:
the one on `4869` would be the primary local relay but apps could support a custom local relay and store this as a local setting in the app.

PR #1274, opened 2024-05-31
NIP-163 (draft): Identicon Sentinel, for easily spotting spoofs

arcbtc:
For adding an easily recognizable identicon sentinel made from a hash of an npub.

Spoofing accounts is easy using the same profile pic,  banner and closely related nip05, but having a little visual reference made from an npub would  make the spoofs easier to spot.

Whatever identicon library is used, all clients should use the same library, <a href="https://laurentpayot.github.io/minidenticons/">minidenticons</a> is an excellent choice, they are easy to identify even if smaller on the client.

image


arcbtc:
You can make a closely related nip05, so should be used for everyone.

In twitter I have spoofs dm'ing people constantly, whereas nostr folks are more astute, as the world turns to nostr this seems like low hanging fruit for battling spoofs.

arcbtc:
> This has been tried lots of times on various clients
I would argue this justifies it being a standard, clearly there is demand.

staab:
> I would argue this justifies it being a standard, clearly there is demand

Yes, and github does it etc. But I've never once identified someone by their identicon, unless it's because they use an identicon while everyone else uses regular profile pictures. Maybe it's just how my brain works. But normally profile pictures vary greatly in style, composition, brightness, saturation, color, etc. If there were an identicon library that selected from 100+ different styles (this guy's a robot, that guy's a space invader, etc), that would have a chance of working with human pattern matching. But even then, I don't know if I would add this to coracle for aesthetic reasons.

vitorpamplona:
I think private petnames are much more effective than this. @vicariousdrama has been using #761 to do that. 

We just need to encourage users to change or add to a user's name.

PR #1284, opened 2024-06-03
More precisely document since/until bounds.

NfNitLoop:
Lower in this document, this is expressed as "since <= created_at <= until".

But the English description said that these values had to be "newer than" or "older than".

These two were in conflict. I assume the inclusive behavior is the intended one, since that documentation was added later to clarify. This just makes both cases match.

### DIFF
--- a/01.md
+++ b/01.md
@@ -125,7 +125,7 @@ Clients can send 3 types of messages, which must be JSON arrays, according to th
   "authors": <a list of lowercase pubkeys, the pubkey of an event must be one of these>,
   "kinds": <a list of a kind numbers>,
   "#<single-letter (a-zA-Z)>": <a list of tag values, for #e — a list of event ids, for #p — a list of pubkeys, etc.>,
-  "since": <an integer unix timestamp in seconds. Events must have a created_at >= this to pass>,
+  "since": <an integer unix timestamp in seconds. Events must have a created_at >= to this to pass>,
   "until": <an integer unix timestamp in seconds. Events must have a created_at <= to this to pass>,
   "limit": <maximum number of events relays SHOULD return in the initial query>
 }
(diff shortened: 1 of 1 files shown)


AsaiToshiya (approved):

01.md
```diff
@@ -125,8 +125,8 @@ Clients can send 3 types of messages, which must be JSON arrays, according to th
   "authors": <a list of lowercase pubkeys, the pubkey of an event must be one of these>,
   "kinds": <a list of a kind numbers>,
   "#<single-letter (a-zA-Z)>": <a list of tag values, for #e — a list of event ids, for #p — a list of pubkeys, etc.>,
-  "since": <an integer unix timestamp in seconds, events must be newer than this to pass>,
-  "until": <an integer unix timestamp in seconds, events must be older than this to pass>,
+  "since": <an integer unix timestamp in seconds. Events must have a created_at >= this to pass>,
+  "until": <an integer unix timestamp in seconds. Events must have a created_at <= this to pass>,
```
alexgleason, line 129:
```suggestion
  "until": <an integer unix timestamp in seconds. Events must have a created_at <= to this to pass>,
```

01.md
```diff
@@ -125,8 +125,8 @@ Clients can send 3 types of messages, which must be JSON arrays, according to th
   "authors": <a list of lowercase pubkeys, the pubkey of an event must be one of these>,
   "kinds": <a list of a kind numbers>,
   "#<single-letter (a-zA-Z)>": <a list of tag values, for #e — a list of event ids, for #p — a list of pubkeys, etc.>,
-  "since": <an integer unix timestamp in seconds, events must be newer than this to pass>,
-  "until": <an integer unix timestamp in seconds, events must be older than this to pass>,
+  "since": <an integer unix timestamp in seconds. Events must have a created_at >= this to pass>,
```
alexgleason, line 128:
```suggestion
  "since": <an integer unix timestamp in seconds. Events must have a created_at >= to this to pass>,
```

alexgleason:
I wish there was a way to say "newer or the same age" more clearly in English.

I think the words "newer" and "older" are more friendly terms than >=, <=. It requires less brain power to understand the concept.

NfNitLoop:
I accepted the proposed updates.

> I think the words "newer" and "older" are more friendly terms than >=, <=

They mean different things, though.  You can't say that X is "newer than" Y if they have the same age.

tyiu (approved):
Ship it. This change is strictly better than what's currently written.

AsaiToshiya:
If you have problem with the merge, please create a PR again.

PR #1292, opened 2024-06-07
Adds Order of results for NIP-50 Search

vitorpamplona:
Searching for `fiatjaf` right now brings back the `n` most recent metadata changes. If clients limit by 100

```
["REQ", "mysub", {"kinds":[0],"limit":100,"search":"fiatjaf"}]
```

The results don't include @fiatjaf. 

Either NIP-50 relays are returning based on `.created_at` OR the scoring algorithm is not very good.

Pinging NIP-50 operators: @nostrband @nostr-wine @Semisol

### DIFF
--- a/50.md
+++ b/50.md
@@ -26,9 +26,8 @@ Relays SHOULD interpret the query to the best of their ability and return events
 Relays SHOULD perform matching against `content` event field, and MAY perform
 matching against other fields if that makes sense in the context of a specific kind. 
 
-Results SHOULD be returned in the order of their matching score, not by the usual `.created_at`. 
-The `limit` filter SHOULD be applied after sorting by matching score.
-
+Results SHOULD be returned in descending order by quality of search result (as defined by the implementation),
+not by the usual `.created_at`. The `limit` filter SHOULD be applied after sorting by matching score.
 A query string may contain `key:value` pairs (two words separated by colon), these are extensions, relays SHOULD ignore 
 extensions they don't support.
 
--- a/50.md
+++ b/50.md
@@ -26,6 +26,9 @@ Relays SHOULD interpret the query to the best of their ability and return events
 Relays SHOULD perform matching against `content` event field, and MAY perform
 matching against other fields if that makes sense in the context of a specific kind. 
 
+Results SHOULD be returned in the order of their matching score, not by the usual `.created_at`. 
+The `limit` filter SHOULD be applied after sorting by matching score.
+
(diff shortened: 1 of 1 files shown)


staab:
This is hilarious. I see this more as a bug of the implementation rather than the spec, but clarifying doesn't hurt. I do think that the definition of `matching score` should be up to the implementation, since it may not be strictly based on text matching, but on web of trust or some other heuristic.

vitorpamplona:
> This is hilarious.

When life gives you lemons, we make lemonade. 

> definition of matching score should be up to the implementation

Yeah, I think Clients just expect the limit filter to be by quality, not FIFO.

50.md
```diff
@@ -26,6 +26,9 @@ Relays SHOULD interpret the query to the best of their ability and return events
 Relays SHOULD perform matching against `content` event field, and MAY perform
 matching against other fields if that makes sense in the context of a specific kind. 
 
+Results SHOULD be returned in the order of their matching score, not by the usual `.created_at`. 
+The `limit` filter SHOULD be applied after sorting by matching score.
+
```
staab, line 31:
```suggestion
```suggestion
Results SHOULD be returned in descending order by quality of search result (as defined by the implementation),
not by the usual `.created_at`. The `limit` filter SHOULD be applied after sorting by matching score.

```

staab (approved):

nostr-wine:
We were indeed sorting these by created_at to keep it consistent with expectations from other filters. Just deployed an update to all regions of nostr.wine and filter.nostr.wine so all results should be sorted by score now.

The real (or so they say) fiatjaf is now the 6th result for us. Not ideal, but better...

mikedilger:
I think this issue was just a bug, and this change to NIP-50 is half right and half wrong.

I like clarifying how LIMIT should be applied, but I don't think we should have changed the sort order.  I think clients may expect the sort order to be `created_at`.  I would have considered search to either match or not match and not really to provide an alternate sort order.

I think there is a common relay bug with LIMIT (which I will have relay-tester test for).  LIMIT should always work AS IF the code loaded all the events in created_at DESC order, and only afterwards did it take the most recent LIMIT events.  It is easy to try to optimize code in a way that fails to yield what that naive straightforward algorithm yields (I had that bug in chorus initially).

staab:
The root problem here I think is that search shouldn't be bundled with relays, because it's essentially different from other REQs.

mikedilger:
All this time (until I read this PR) I thought SEARCH was a command, not a field in a REQ. I never looked at NIP-50 I just presumed it would have been done that way.

PR #1293, opened 2024-06-07
NIP-09: clarify "a" tag deletions

alexgleason:
Deleting a replaceable event is not well defined. It should delete all versions of the event up to the created_at timestamp of the deletion event itself, to allow the author to publish that kind of event in the future.

### DIFF
--- a/09.md
+++ b/09.md
@@ -32,6 +32,8 @@ Relays SHOULD delete or stop publishing any referenced events that have an ident
 
 Relays SHOULD continue to publish/share the deletion events indefinitely, as clients may already have the event that's intended to be deleted. Additionally, clients SHOULD broadcast deletion events to other relays which don't have it.
 
+When an `a` tag is used, relays SHOULD delete all versions of the replaceable event up to the `created_at` timestamp of the deletion event.
+
 ## Client Usage
 
 Clients MAY choose to fully hide any events that are referenced by valid deletion events.  This includes text notes, direct messages, or other yet-to-be defined event kinds.  Alternatively, they MAY show the event along with an icon or other indication that the author has "disowned" the event.  The `content` field MAY also be used to replace the deleted events' own content, although a user interface should clearly indicate that this is a deletion reason, not the original content.


staab (approved):

vitorpamplona (approved):

staab:
@jb55 @v0l does this pass the sniff test?

mikedilger (approved):

alexgleason:
@fiatjaf You good with this? Asking since you had also opened https://github.com/nostr-protocol/nips/pull/1263

vitorpamplona:
Can we make sure StrFry operates in this way?

alexgleason:
> Can we make sure StrFry operates in this way?

strfry does not support "a" tag deletions: https://github.com/hoytech/strfry/blob/master/src/events.cpp#L332

As much as I love strfry for its ingenuity and efficiency, it should not be regarded as the flagship relay implementation. It has quite a few bugs and is not actively being developed at the pace most of us are moving at.

I don't think I will use strfry for future projects. It's actually not that hard to build a relay - way easier than building a client. Between a 150 line [controller](https://gitlab.com/soapbox-pub/ditto/-/blob/main/src/controllers/nostr/relay.ts?ref_type=heads) and 400 line [storage adapter](https://gitlab.com/soapbox-pub/nostrify/-/blob/main/src/NDatabase.ts?ref_type=heads) I have a relay that works better than strfry and supports all these features. It's not as efficient with storage but the functionality is better.

vitorpamplona:
I agree but the reality is that, right now, everyone runs StrFry. It has become the de-facto standard for Nostr. :(

staab:
Lack of support on the relay side shouldn't block a-tag deletions. Clients that want to support them should be downloading them anyway, so it shouldn't make any real difference in end-user experience.

alexgleason:
strfry does not even properly support NIP-01 due to mishandling `since` and `until` filters. See: https://github.com/hoytech/strfry/pull/110

What strfry does should basically not influence what we do with NIPs.

Exactly @staab, was typing this at the same time.

vitorpamplona:
I just hope this doesn't become yet another NIP-26 that exists only on paper.

alexgleason:
"a" tag deletion is already in the spec. This is just a clarification of what is already documented, matching what all known implementations are already doing. Whether a-tag deletions are wrong does not really pertain to this MR. If we decide later it is still wrong for some reason, we can remove the original text. But as long as a-tag deletions are in this document, it's helpful to have the clarifying statement.

AsaiToshiya:
Just to confirm, as explained in NIP-01, basically when a new replaceable event is published, older versions should be deleted, right?

https://github.com/nostr-protocol/nips/blob/d251ca0da5372ba0ae0df76754d96068266a6778/01.md?plain=1#L96-L98

PR #1298, opened 2024-06-10
nip54: change to asciidoc

fiatjaf:
== Why?

Because Markdown is somewhat of a mess. It's fine when you just use the core features, but when you start to add tables and other stuff it quickly degrades. And there had been talks already of how to add description lists, footnotes, sidebars and tables to these wiki articles so we would not be able to stick to mostly plain text anyway.

Asciidoc has a strict spec, multiple implementations in many languages, and support for features that are very much necessary in a wiki article, like _sidebars_, _tables_ (with rich markup inside cells), many levels of _headings_, _footnotes_, _superscript_ and _subscript_ markup and _description lists_. It is also arguably easier to read in its plaintext format than Markdown.

Read this if you're still not convinced: https://docs.asciidoctor.org/asciidoc/latest/asciidoc-vs-markdown/

Caveat: I don't have much experience with Asciidoc (if I had I might have proposed this NIP with Asciidoc since the beginning anyway).

Caveat: unfortunately Asciidoc is extensible, so it is not guaranteed that implementations will all implement the same extensions and stay compatible forever. However the extensions at least are well-delimited, named and specified and must be added explicitly by the developer (unlike on Markdown world where each library will add ad-hoc extensions by themselves), so that should be less bad. Also it's always clear when an extension is being used since its usage is delimited in tagged blocks in the source text. In any case I think we won't need any extensions for a long time.

== Migration path

Option 1:: Just drop Markdown support and be done with it. Everything will still be readable.
Option 2:: Documents created up to the date this document is merged should still be supported in the existing wiki clients. Documents with a higher timestamp shouldn't. Clients can drop Markdown support entirely in the future when that becomes a more reasonable idea.

@hzrd149 @pablof7z @SilberWitch

### DIFF
--- a/54.md
+++ b/54.md
@@ -28,13 +28,13 @@ Articles are identified by lowercase, normalized ascii `d` tags.
 
 ### Content rules
 
-The content should be Markdown, following the same rules as of [NIP-23](23.md), although it takes some extra (optional) metadata tags:
+The content should be Asciidoc, following the same rules as of [NIP-23](23.md), although it takes some extra (optional) metadata tags:
 
   - `title`: for when the display title should be different from the `d` tag.
   - `summary`: for display in lists.
   - `a` and `e`: for referencing the original event a wiki article was forked from.
 
-One extra functionality is added: **wikilinks**. Unlike normal Markdown links `[]()` that link to webpages, wikilinks `[[]]` link to other articles in the wiki. In this case, the wiki is the entirety of Nostr. Clicking on a wikilink should cause the client to ask relays for events with `d` tags equal to the target of that wikilink.
+One extra functionality is added: **wikilinks**. Unlike normal Asciidoc links `[]()` that link to webpages, wikilinks `[[]]` link to other articles in the wiki. In this case, the wiki is the entirety of Nostr. Clicking on a wikilink should cause the client to ask relays for events with `d` tags equal to the target of that wikilink.
 
 Wikilinks can take these two forms:
 
@@ -86,12 +86,12 @@ This is a stronger signal of trust than a `+` reaction.
 
 This marker is useful when a user edits someone else's entry; if the original author includes the editor's changes and the editor doesn't want to keep/maintain an independent version, the `link` tag could effectively be a considered a "deletion" of the editor's version and putting that pubkey's WoT weight behind the original author's version.
 
-Why Markdown?
+Why Asciidoc?
 -------------
 
-If the idea is to make a wiki then the most obvious text format to use is probably the mediawiki/wikitext format used by Wikipedia since it's widely deployed in all mediawiki installations and used for decades with great success. However, it turns out that format is very bloated and convoluted, has way too many features and probably because of that it doesn't have many alternative implementations out there, and the ones that exist are not complete and don't look very trustworthy. Also it is very much a centralized format that can probably be changed at the whims of the Wikipedia owners.
+Wikitext is [garbage](nostr:nevent1qqsqt0gcggry60n72uglhuhypdlmr2dm6swjj69jex5v530gcpazlzsprpmhxue69uhhyetvv9ujumn0wdmksetjv5hxxmmdqy28wumn8ghj7un9d3shjtnyv9kh2uewd9hsygpm7rrrljungc6q0tuh5hj7ue863q73qlheu4vywtzwhx42a7j9n5ueneex) and Markdown is not powerful enough (besides being too freeform and unspecified and prone to generate incompatibilities in the future).
 
-On the other hand, Markdown has proven to work well for small scale wikis and one of the biggest wikis in the planet (which is not very often thought of as a wiki), [StackOverflow](https://stackoverflow.com) and its child sites, and also one of the biggest "personal wiki" software, [Obsidian](https://obsidian.md/). Markdown can probably deliver 95% of the functionality of wikitext. When augmented with tables, diagram generators and MathJax (which are common extensions that exist in the wild and can be included in this NIP) that rate probably goes to 99%, and its simplicity is a huge benefit that can't be overlooked. Wikitext format can also be transpíled into Markdown using Pandoc. Given all that, I think it's a reasonable suspicion that mediawiki is not inherently better than Markdown, the success of Wikipedia probably cannot be predicated on the syntax language choice.
+Asciidoc has a strict spec, multiple implementations in many languages, and support for features that are very much necessary in a wiki article, like _sidebars_, _tables_ (with rich markup inside cells), many levels of _headings_, _footnotes_, _superscript_ and _subscript_ markup and _description lists_. It is also arguably easier to read in its plaintext format than Markdown (and certainly much better than Wikitext).
 
 # Appendix 1: Merge requests
 Users can request other users to get their entries merged into someone else's entry by creating a `kind:818` event.


pablof7z:
@vitorpamplona has also added support for `30818`.

vitorpamplona:
Clients will end up having to support every format anyway, so bring it on.

fiatjaf:
> I like asciidoc better. But what about this: wikifreedia.xyz/nud:-nostext/npub1jlrs53pkdfjnts29kveljul2sm0actt6n8dxrrzqcersttvcuv3qdjynqn

Looks like a simpler and more strict Markdown. Could be good for notes, but it's not powerful enough for wiki articles.

vitorpamplona:
> I know you were going to say this, but let me just say again that you are wrong!

Prove me wrong and everybody wins.

SilberWitch:
> Looks like a simpler and more strict Markdown. Could be good for notes, but it's not powerful enough for wiki articles.

Yeah, I'd go nuts trying to format articles with so few options. Like trying to edit a magazine or encyclopedia with nothing but .txt files.

SilberWitch:
Interesting. I'm apparently using some asciidoc formatting, thinking it was markdown, like _italics_ . That explains why some of the clients aren't displaying them as expected.

I've seen the ==Header format in the articles I've referenced from Medium.com, so they seem to be using Asciidoc.

I support this change. 🤔 It's still early and markdown files stay readable, even without the rendering.

SilberWitch:
> Interesting. I'm apparently using some asciidoc formatting, thinking it was markdown, like _italics_ . That explains why some of the clients aren't displaying them as expected.
> 
> I've seen the ==Header format in the articles I've referenced from Medium.com, so they seem to be using Asciidoc.
> 
> I support this change. 🤔 It's still early and markdown files stay readable, even without the rendering.

Apparently, GitHub allows for both, as the italics rendering is correct.

SilberWitch:
Is this going to happen?

fiatjaf:
Yes!

fiatjaf:
https://wikistr.com/ has transitioned already and I'm slowly working on the change for Wikifreedia.

SilberWitch:
Okay, I saw that my files looked funny on wikistr. Now I'm sort of trapped, since some of them have changed, but some not, so I guess I'll just wait until everyone moves and then reformat.

SilberWitch:
@hzrd149 are you also aware of this change?

SilberWitch:
Yeah, it's a bit of a hurdle.

SilberWitch:
We've added an optional "markup" tag to our 30041, so that someone can define if they used asciidoc, markdown, LaTeX, PlantUML, AsciiMath or whatnot in their article event, with markdown the default, if it's undefined.

We have to support a wider array because we're using the events for complex documents like scientific research papers and etc.

Might be an option for other article events. 🤷‍♀️

pablof7z:
> We've added an optional "markup" tag to our 30041, so that someone can define if they used asciidoc, markdown, LaTeX, PlantUML, AsciiMath or whatnot in their article event, with markdown the default, if it's undefined.
> 
> We have to support a wider array because we're using the events for complex documents like scientific research papers and etc.
> 
> Might be an option for other article events. 🤷‍♀️

Different formats should have different kinds; otherwise you just raise the bar to support a specific kind and increase the complexity for rendering it. Overloading kinds with different, even if adjacent, meanings and/or renderings is an anti-pattern.

SilberWitch:
Then we should reject this PR and stick with only markdown and come up with a different kind for Asciidoc.

SilberWitch:
We are basing our Kind 30041 off of this one and we assumed that allowed for markdown default.

fiatjaf:
I believe without Asciidoc (or some other markup language) the Nostr free encyclopedia will never attract the methodical people who love categories and structure. Markdown is too limited and clients will naturally start adding ad-hoc extensions to it to accommodate some needs, and these extensions will not be as complete as Asciidoc, but also they won't be standard enough and there won't be WYSIWG editors that support them uniformly anyway, so nothing will work. @hzrd149 

If I manage to finish my Quill integration then we'll have at least one WYSIWG editor for Asciidoc.

SilberWitch:
Is NIP-23 also going to Asciidoc or staying Markdown?

fiatjaf:
That should stay in Markdown.

cesardeazevedo:
I am currently working in a WYSIWG editor for nostr based on tiptap / prosemirror, It supports markdown already and I would love to support Asciidoc too, and I agree this should be in a different kind.

SilberWitch:
Can we just have two different wiki kinds, one for markdown and one for asciidoc? Then the writer is free to pick the one they prefer.

It will usually be markdown, but some more complex documents could use asciidoc.

fiatjaf:
No. It must be one. Having two just doubles the work for everybody. We're still very early here and we can transition. But if everybody wants to remain on Markdown then please let me know so I can stop working on Asciidoc crap.

SilberWitch:
I don't see how we don't offer some event with asciidoc, long-term, without cutting off advanced publishing use cases. We were considering having asciidoc-blocks in markdown docs, but it seems to be easier, the other way around.

I also saw that there are PlantUML and LaTeX extensions for Asciidoc, which is awesome.

SilberWitch:
Yeah, okay, just looked at my wiki pages under Asciidoc and they look nearly the same. Only the more advanced stuff that only I'm using is even different.

Most people could keep writing markdown and wouldn't even notice the difference.

fiatjaf:
https://github.com/fiatjaf/quill-asciidoc
https://github.com/fiatjaf/svelte-asciidoc
https://github.com/pablof7z/wiki/pull/32

SilberWitch:
Okay, GitCitadel is going to host Wikistr, we're rewriting our docs into asciidoc, and our products will use @fiatjaf 's asciidoc libraries.
Done.

fiatjaf:
Well, I'm sorry if there are bugs, please let me know if you hit any.

SilberWitch:
No problem. I'm a tester.

NielLiesmons:
I'm not convinced of this switch.
There's more benefit in:
- keeping articles and wikis the same format (supporting one ≅ supporting both)
- using the most popular markup language and the tools built on it
Then going for asciidocs because of: 
- tables that display badly and nobody uses outside of a few use cases (which Github Markdown already solved for)
- very limited extensibilty relative to the functionality that Nostr can offer

Very much prefer this scenario: 
https://njump.me/naddr1qvzqqqr4gupzp22rfmsktmgpk2rtan7zwu00zuzax5maq5dnsu5g3xxvqr2u3pd7qq9y6ctjddjx7amw9d8qntf0u3
But I'll be fine with Asciidocs + Nostr Widgets too. No worries.

SilberWitch:
I'm assuming articles will eventually migrate over to asciidoc, as well, as @fiatjaf has a natural incentive to create something that handles special Nostr rendering or widgets or whatever, so that we can have a common [Nostr-native parser](https://github.com/fiatjaf/svelte-asciidoc), and that'll be more attractive than each dev wasting his energy coming up with the upteenth flavor of markdown.

At the moment, the pages render differently, depending upon which app you look at, which is upsetting to see because it feels like formatting is random. This is a problem other apps don't have because they assume you upload and reformat the md file separately at each different app, as they have no common protocol and most writers only upload to one place. We upload to one place and then view on n number of places.

Here are two examples of a wiki page I wrote in markdown, displayed in [Nostrudel](https://next.nostrudel.ninja/#/wiki/topic/nkbip-01?pubkey=dd664d5e4016433a8cd69f005ae1480804351789b59de5af06276de65633d319) and [Wikifreedia](https://wikifreedia.xyz/nkbip-01/laeserin@getalby.com) (as markdown) and in [Wikistr](https://wikistr.com/nkbip-01*dd664d5e4016433a8cd69f005ae1480804351789b59de5af06276de65633d319) (as asciidoc). There's hardly any difference for basic md, the resulting data is a content string in json (not a md or adoc file, unlike on GitHub), and the display is html.

pablof7z:
Wikifreedia now uses asciidoc

pablof7z:
> I am currently working in a WYSIWG editor for nostr based on tiptap / prosemirror, It supports markdown already and I would love to support Asciidoc too, and I agree this should be in a different kind.

Are you currently working on a WYSIWG editor for asciidoc?

staab:
> Are you currently working on a WYSIWG editor for asciidoc?

 I'm contributing to the editor, he's currently supporting plaintext and markdown, but I think asciidoc could be added without too much trouble. 

https://github.com/cesardeazevedo/nostr-editor

cesardeazevedo:
First version of nostr-editor was just published a few minutes ago as I want to collect more feedback, I am planning to support asciidoc soon, demo can be found here https://cesardeazevedo.github.io/nostr-editor/

PR #1302, opened 2024-06-11
Clarifies which relays to put in the zap request

vitorpamplona:
When paying zap splits, the relay list of the zap request should include the post author so that the amount of received zaps can be counted as well as the inbox relays from each of the recipients so that they can receive notifications.

### DIFF
--- a/57.md
+++ b/57.md
@@ -28,7 +28,7 @@ Having lightning receipts on nostr allows clients to display lightning payments 
 
 A `zap request` is an event of kind `9734` that is _not_ published to relays, but is instead sent to a recipient's lnurl pay `callback` url. This event's `content` MAY be an optional message to send along with the payment. The event MUST include the following tags:
 
-- `relays` is a list of relays the recipient's wallet should publish its `zap receipt` to. Note that relays should not be nested in an additional list, but should be included as shown in the example below.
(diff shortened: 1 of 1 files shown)


57.md
```diff
@@ -28,7 +28,7 @@ Having lightning receipts on nostr allows clients to display lightning payments
 
 A `zap request` is an event of kind `9734` that is _not_ published to relays, but is instead sent to a recipient's lnurl pay `callback` url. This event's `content` MAY be an optional message to send along with the payment. The event MUST include the following tags:
 
-- `relays` is a list of relays the recipient's wallet should publish its `zap receipt` to. Note that relays should not be nested in an additional list, but should be included as shown in the example below.
+- `relays` is a list of relays the recipient's wallet should publish its `zap receipt` to. The list SHOULD include the recipient's READ relays from [NIP-65](65.md), the sender's READ relays and the post author's READ relays (e.g receiver and author are different when paying zap splits)
```
alexgleason, line 31:
```suggestion
- `relays` is a list of relays the recipient's wallet should publish its `zap receipt` to. The list SHOULD include the recipients' READ relays from [NIP-65](65.md), the sender's READ relays, and the post author's READ relays (e.g receivers and author may be different when paying zap splits)
```

alexgleason (approved):

PR #1303, opened 2024-06-11
Adds Proxy and Broadcasting relay lists.

vitorpamplona:
Adds lists for specific proxy/broadcasting relays and guides clients on how to work with them. 

Read [here](https://github.com/vitorpamplona/nips/blob/broadcasting-proxy-relay-lists/74.md)

### DIFF
--- a/74.md
+++ b/74.md
@@ -1,0 +1,41 @@
+NIP-74
+======
+
+Proxy and Broadcasting Relay Lists
+----------------------------------
+
+`draft` `optional`
+
+Kind `10017` indicates the user's preferred relays for proxy downloads. Such services save the user's bandwidth by utilizing a single connection to a proxy relay which reaches out to all of the other relays. 
+
+Kind `10018` indicates the user's preferred relays for broadcasting relays. Such services improve the user's privacy by utilizing a broadcasting relay to distribute an event to all of the appropriate relays without requiring the user to connect with each of those relays.
+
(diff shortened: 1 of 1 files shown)


74.md
```diff
@@ -0,0 +1,41 @@
...
+  "kind": 10018,
+  "tags": [],
+  "content": nip44Encrypt(JSON.stringify([
+    ["relay", "wss://mybroadcaster.relay.com"]
+  ]))
+  //...other fields
+}
+```
+
+Clients SHOULD use proxy relays to download events before using it's native event-finding method.
```
staab, line 37: `its`

staab (approved):
I like that the relay selections are encrypted. I would also like to see private relay selections as well (not intended to be used for outbox relay selection), but that can wait for another PR.

vitorpamplona:
> I would also like to see private relay selections as well (not intended to be used for outbox relay selection), but that can wait for another PR.

I added one in the #1124 Generic Draft NIPs. It's a relay for all the private settings and things no one else needs to know, like private drafts.

staab:
That's yet another (useful) category. I was thinking of something more along [these lines](https://github.com/nostr-protocol/nips/discussions/1301#discussioncomment-9744095).

erskingardner:
Love this idea. 👍

mikedilger:
I'm in favor of this PR, for sharing this config between clients that use this model.

Gossip client won't support it, but people pretty much know that already; I don't think they will be surprised.

vitorpamplona:
Superseded by https://github.com/nostr-protocol/nips/pull/1985

Issue #1216, opened 2024-05-03
Template and Guidance for writing NIPs


xeruf:
As I am looking to write an NIP for https://github.com/nostr-protocol/nips/discussions/1019, I realized there is no guidance on how to actually write an NIP, and thus a lot of inconsistency.
There seems to be no common template for how to structure new NIPs - for example some NIPs specify event kinds as a list of properties, others through an exemplary JSON structure, what is preferred?

Follow-up to https://github.com/nostr-protocol/nips/issues/106 which was closed prematurely in my opinion, because having a template is not a restriction - we can still decide to allow people to deviate from the template, but it makes the process a lot easier to at least have one available.

This can then easily solve:
- https://github.com/nostr-protocol/nips/issues/1057
- https://github.com/nostr-protocol/nips/issues/1071 / https://github.com/nostr-protocol/nips/issues/1070

So far I had to repeatedly flip through existing NIPs to get an idea of what is required, as well as the meaning of some kinds and tags - for example, I had no idea kinds had agreed-upon number ranges until stumbling upon that in NIP-01 (https://github.com/nostr-protocol/nips/blob/master/01.md#tags).
This incoherence creates unnecessary burdens for NIP writers, reviewers and implementers.
The only guidance so far is in the (arguably convoluted, I think things like the list of kinds should be in a separate file) README, and very sparse: https://github.com/nostr-protocol/nips?tab=readme-ov-file#criteria-for-acceptance-of-nips

Some aspects to consider here:
- A template markdown file to copy to fill in things so no important section is forgotten - there may still be a conscious choice to omit some section
- Is there a rule to how event kind ids are chosen, apart from the general ranges, or any rule of thumb?
- How do I even choose something as simple as the number of my NIP? Is numbering even ideal, wouldn't few-letter abbreviations be better, something like NIP-CAL rather than NIP-52?

xeruf:
This is still vital for me as I want to start contributing to nostr but abhor messes

pablof7z:
cNACK

> This is still vital for me as I want to start contributing to nostr but abhor messes

mess is good. chaos is good. needless standardization is hurtful in a system like nostr.

The protocol is chaotic because is permissionless and NIPs are descriptive, not prescriptive: there is no
> we can still decide to allow people to deviate from the template

people can do what they want, the can write NIPs, not write them, publish them here, publish them as wiki entries, or do whatever they want; a template will only imply at best, inaccuracies (I would call them _lies_).

[...]

Issue #1071, opened 2024-02-22
Should we have standard tags document that detail what each tag means?


basantagoswami:
The description, structure etc related to standard tags should be moved out of the NIPs to a different document. Currently the tags are defined in NIPs that might not even be related to the tag at all.

If there are tags that are used differently or mean different things in different kind events, then we can have something like this in their related NIPs: "This NIP redefined the tag `xyz` to mean lorem ipsum when it is used in a `kind:xxx` event." Or: "When used with `kind:yyy` events, the tag `abc` doesn't need to have a 4th element in the array and can be and empty string"

I am not sure if it's even desired to have the same tag mean different thing in different event kinds. If not, then standardizing them out of the optional NIPs makes sense.

AsaiToshiya:
[README Standardized Tags](https://github.com/nostr-protocol/nips?tab=readme-ov-file#standardized-tags) and [NIP-24 tags](https://github.com/nostr-protocol/nips/blob/master/24.md#tags) are not enough?

basantagoswami:
> README Standardized Tags and NIP-24 tags are not enough?
The tags on the README aren't really standard though. For example, it defines `r` as the relay URL tag from NIP-65, but some events use `r` to mean resource or some other web URL. 

Currently the only truly standard tags are `e`, `p`, `a` and `d`.

vitorpamplona:
Tags change from NIP to NIP. I wouldn't code anything (not even e tags) as a general Nostr tag.

Issue #1070, opened 2024-02-22
Why have "r" and "relay" both refer to relays?


basantagoswami:
[NIP-42 (Authentication of clients to relays)](https://github.com/nostr-protocol/nips/blob/master/42.md) and [NIP-51 (Lists)](https://github.com/nostr-protocol/nips/blob/master/51.md) use the `relay` tag to refer to a relay. But [NIP-65 (Relay List Metadata)](https://github.com/nostr-protocol/nips/blob/master/65.md) uses the `r` tag.

Any reason for that? Do relays mean two different thing conceptually here or is it kept just for backwards compatibility? Or is it related to not wanting to index these tags in some places?

fiatjaf:
Tags can have different meanings in different kinds.

basantagoswami:
> Tags can have different meanings in different kinds.

Does it not break this philosophy from the NIPs readme: *"There should be no more than one way of doing the same thing."* in this particular case?
Also this issue probably relates to issue #1071 about having a different doc for standard tags.

basantagoswami:
Content almost always has different meaning based on the kind. Tags almost always have the same meaning even in different kinds. These tags: `e`, `p`, and `a`are defined in NIP-01 to have a very specific meaning for all kinds. The `d` tag also has a very specific meaning for parameterized replaceable events.

So I don't think they are the same thing at all. The list of tags are also called standard tags in the NIPs readme. I think it would be worth it to have the same tags across kinds and then let new NIPs redefine them for new kinds if they want or need to.

staab:
NIPs were written by just a bunch of people, based on sub-protocols they made up without asking other people, and with little concern for consistency. What exists is normative, tags are inconsistent, everything works, therefore consistent use of tags isn't a constraint for protocol design. Of course, tags should be used conventionally, but that's a different question.

mekefly:
A single label has an indexed eg r p e, and words without an indexed eg relays relay are not searchable

filter
`#r`   yes
`#relay` no

PR #1304, opened 2024-06-12
README: subdivide tags table

AsaiToshiya:
This is a small suggestion for #1070, #1071, and #1216. Of course, I know it is not essential.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,9 @@ They exist to document what may be implemented by [Nostr](https://github.com/nos
 - [Message Types](#message-types)
   - [Client to Relay](#client-to-relay)
   - [Relay to Client](#relay-to-client)
-- [Standardized Tags](#standardized-tags)
+- [Tags](#tags)
+  - [Standardized Tags](#standardized-tags)
+  - [Conventional Tags](#conventional-tags)
 - [Criteria for acceptance of NIPs](#criteria-for-acceptance-of-nips)
 - [Is this repository a centralizing factor?](#is-this-repository-a-centralizing-factor)
 - [How this repository works](#how-this-repository-works)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -224,14 +226,21 @@ They exist to document what may be implemented by [Nostr](https://github.com/nos
 | `AUTH`   | used to send authentication challenges                  | [42](42.md) |
 | `COUNT`  | used to send requested event counts to clients          | [45](45.md) |
 
-## Standardized Tags
+## Tags
```
alexgleason, line 229:
```suggestion
## Tags

Different kinds may use the same tag differently.
```

README.md
```diff
@@ -224,14 +226,21 @@ They exist to document what may be implemented by [Nostr](https://github.com/nos
...
+
+### Standardized Tags
```
alexgleason, line 231:
```suggestion
### Common Tags

These tags are widely used across many different event kinds with the exact same structure.
```

README.md
```diff
@@ -224,14 +226,21 @@ They exist to document what may be implemented by [Nostr](https://github.com/nos
...
+
+| name | value                   | other parameters                | NIP                      |
+|------|-------------------------|---------------------------------|--------------------------|
+| `e`  | event id (hex)          | relay URL, marker, pubkey (hex) | [01](01.md), [10](10.md) |
+| `p`  | pubkey (hex)            | relay URL, petname              | [01](01.md), [02](02.md) |
+| `a`  | coordinates to an event | relay URL                       | [01](01.md)              |
+| `d`  | identifier              | --                              | [01](01.md)              |
+
+### Conventional Tags
```
alexgleason, line 240:
```suggestion
### Conventional Tags

These tags are defined by at least one NIP and may vary slightly between different kinds.
```

README.md
AsaiToshiya: How about adding it to NIP-01?

README.md
AsaiToshiya: I think the meaning of these tags is explained in NIP-01.

https://github.com/nostr-protocol/nips/blob/d251ca0da5372ba0ae0df76754d96068266a6778/01.md?plain=1#L76

README.md
AsaiToshiya: It is needed.

README.md
alexgleason: Oh! Good point. 😃

AsaiToshiya:
@basantagoswami @xeruf

README.md
xeruf: Then I would call them `Standard` rather than `Standardized`

README.md
xeruf: These explanations, duplicating them?
Doesn't sound like a good idea to me.

AsaiToshiya: No, they will not be explained here.

README.md
AsaiToshiya: Agreed.

xeruf (approved):

PR #1305, opened 2024-06-13
NIP-24: fix missing rename

AsaiToshiya:
Related PR: #921


AsaiToshiya:
@staab thank you always.

PR #1306, opened 2024-06-13
Adds key aliases to NIP-17

vitorpamplona:
A simple way to: 
1. Remove the final metadata leak from NIP-17 (the public key of the receiver)
2. Distribute a stream of messages through multiple relays that are not publically advertised by the user. 
3. Control disappearing messages from others to you (just delete/rotate the alias private key). You can already control disappearing messages from you to others by adding the `expiration` tag to wraps.

Questions: 
- Should we add this to all Gift Wraps or just DMs?
- Should we offer more relay URL options per key or is one enough?

### DIFF
--- a/17.md
+++ b/17.md
@@ -108,6 +108,50 @@ It's advisable that users choose relays that conform to these practices.
 
 Clients SHOULD guide users to keep `kind:10050` lists small (1-3 relays) and SHOULD spread it to as many relays as viable.
 
+## Key Aliases
+
+Key Aliases hide the user's main key from the public gift wrap. 
+
+Key Alias Events use `kind:10059` replaceable events to privately inform peers of the set of pub keys a user is ALSO subscribing their GiftWrap filters with. The event MUST include a `p` tag for each active public key with an optional relay url to be used with that key. These keys work in addition to the user's main public key. The `relay` url MAY or MAY NOT be included in the user's private inbox relay list (i.e. kind `10050`),
+
+```js
+{
+  "pubkey": "<Author's Main PubKey>",
+  "kind": 10059,
+  "created_at": now(),
+  "tags": [
+    ["p", "<Pubkey Alias 1>", "<preferred relay url for alias 1>"]
+    ["p", "<Pubkey Alias 2>", "<preferred relay url for alias 2>"]
+    ["p", "<Pubkey Alias 3>", "<preferred relay url for alias 3>"]
+    ["expiration", "1600000000"] // optional
+  ]
+}
+```
+
+`kind:10059` MUST be unsigned, sealed, and gift-wrapped to each receiver individually. They MAY or MAY NOT contain the same aliases for every receiver.
+
+Receiving Clients SHOULD pick one of the keys listed as `p` tags in this event to use as `receiverPublicKey`. Both the wrap and the seal SHOULD be encrypted with this key.
+
+Receiving Clients SHOULD expect a rotation of these aliases and MUST not store past versions of this event.
+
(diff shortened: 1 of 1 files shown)


17.md
```diff
@@ -108,6 +108,50 @@ It's advisable that users choose relays that conform to these practices.
...
+  "tags": [
+    ["p", "<Pubkey Alias 1>", "<preferred relay url for alias 1>"]
+    ["p", "<Pubkey Alias 2>", "<preferred relay url for alias 2>"]
+    ["p", "<Pubkey Alias 3>", "<preferred relay url for alias 3>"]
+    ["expiration", "1600000000"] // optional
+  ]
+}
+```
+
+`kind:10059` MUST be unsigned, sealed, and gift-wrapped to each receiver individually. They MAY or MAY NOT contain the same aliases for every receiver.
```
erskingardner, line 131: I don't like this. The idea of having to know in advance who you're going to receive messages from ahead of receiving the messages doesn't function without first performing part of the conversation using GWs and normal pubkeys.

vitorpamplona: You don't need to know in advance. The base is the raw NIP-17 protocol: just send the message to the pubkey. Once the conversation gets going you can switch to other keys if clients support it. 

The key alias event is very similar to your "Conversation Request" event on #1206. The only difference is that key alias events can happen (or changed) at any point during a regular chat.

staab:
> I also have implemented a secret chat feature in 0xchat, combining nip17 with alias key exchange(NIP 101).

We should use this rather than creating something entirely new, if it's suitable.

The problem I have with adding key exchange to NIP 17 is it hard forks DMs again. Clients that don't support aliases won't receive messages. NIP 17 was "good enough" to displace nip 04, but I don't think the same is true of this NIP. I'd rather see a complete solution come together that is yet another order of magnitude better than 17.

vitorpamplona:
So, the key creation requires more thinking, but for clients that don't want to support aliases, all they need to download the `10058` event and add all those keys to their GiftWrap filter, which has only the p-tag of the user right now. Then, when receiving the GiftWraps, use the private key of the p-Tag, and they will be able to see all messages from the aliases another client created. 

To me is not that much different than adding a specific relay list to the DM NIP. 

But I agree that full support for creating, rotating, and publishing aliases is more intricate to code.

> I'd rather see a complete solution come together that is yet another order of magnitude better than 17.

The only thing that offers an order of magnitude over NIP-17 right now is MLS. And it's going to be very complicated to code that one.

erskingardner:
MLS is coming. 😉

But you're right, it's going to take a bit longer. 

Honestly, I think we should stick with simple nip-17 until MLS based messaging lands. The idea will be that in addition to the spec I'll work on library code and directly with clients to help them implement. It's not going to be super hard but it's really important to do it right so that you aren't breaking the security benefits on accident.

arthurfranca:
You people are smart but aren't focusing on the right thing. Key exchange by itself isn't the most important thing nostr DMs lack.

What we desperately need is new device usage detection - to detect private key leaks - like: 
- Alice-device`#3` talks to Bob-device`#any` for the first time.
- Since Bob-device`#any` never talked to Alice-device`#3`, he should alert Alice devices `#1` and `#2` so these would be notified of new device usage when they get online, possibly revealing that Alice privkey leaked.
- Also, Alice-device`#3` (that could be a hacker) should only see messages exchanged with Bob from now on, not from the past.

Key exchange flow may need to take that into account! This PR nor mentioned NIP-101, nor quadruple^2-ratchet have that.

PR #1309, opened 2024-06-16
metadata clarity.md

ArmanTheParman:
This is not a correction but a suggestion...

When I first read it, it wasn't clear that metadata is specifically for a user and not inclusive of other things (eg not metadata about a photo).

Adding this one extra word here would have helped me understand that immediately.


AsaiToshiya:
Can you also update `metadata` in README.md and 05.md?

ArmanTheParman:
> Can you also update `metadata` in README.md and 05.md?

Done

AsaiToshiya (approved):

ArmanTheParman:
Hey, I'm wondering if "use" is technically incorrect when referring to the
escaped characters? I don't want to introduce inaccuracies.

PR #1310, opened 2024-06-18
tags explanation revision

ArmanTheParman:
The explanation was a bit "jumpy" going from explaining element names, then to values, then back to names.

I made it clearer and flow better.

Also corrected 'keypair' to 'key pair'

### DIFF
--- a/01.md
+++ b/01.md
@@ -71,7 +71,7 @@ Each tag is an array of one or more strings, with some conventions around them. 
 }
 ```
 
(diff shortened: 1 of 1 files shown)


AsaiToshiya:
LGTM

staab:
Maybe it would be better to say that the conventional names depend on the tag/kind they're attached to. `mark`, `hint`, `petname`, etc are all "conventional", just not for every tag.

ArmanTheParman:
> Maybe it would be better to say that the conventional names depend on the tag/kind they're attached to. `mark`, `hint`, `petname`, etc are all "conventional", just not for every tag.

I understand. Wording adjusted.

PR #1315, opened 2024-06-18
Clarify which kinds have lifetime of 'regular' 

ArjenStens:
When trying to learn about the protocol I got confused as to wether or not kind 1, 2, 4 up to (and including) 44 are regular lifetime or not. After some asking around it seems that they do, please correct me if i'm wrong.

### DIFF
--- a/01.md
+++ b/01.md
@@ -92,7 +92,7 @@ Kinds specify how clients should interpret the meaning of each event and the oth
 
 And also a convention for kind ranges that allow for easier experimentation and flexibility of relay implementation:
 
-- for kind `n` such that `1000 <= n < 10000 || 4 <= n < 45 || n == 1 || n ==2`, events are **regular**, which means they're all expected to be stored by relays.
+- for kind `n` such that `1000 <= n < 10000 || 4 <= n < 45 || n == 1 || n == 2`, events are **regular**, which means they're all expected to be stored by relays.
 - for kind `n` such that `10000 <= n < 20000 || n == 0 || n == 3`, events are **replaceable**, which means that, for each combination of `pubkey` and `kind`, only the latest event MUST be stored by relays, older versions MAY be discarded.
 - for kind `n` such that `20000 <= n < 30000`, events are **ephemeral**, which means they are not expected to be stored by relays.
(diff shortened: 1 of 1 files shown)


01.md
```diff
@@ -92,7 +92,7 @@ Kinds specify how clients should interpret the meaning of each event and the oth
 
 And also a convention for kind ranges that allow for easier experimentation and flexibility of relay implementation:
 
-- for kind `n` such that `1000 <= n < 10000`, events are **regular**, which means they're all expected to be stored by relays.
+- for kind `n` such that `1000 <= n < 10000 || 4 <= n < 45 || n == 1 || n == 2`, events are **regular**, which means they're all expected to be stored by relays.
```
AsaiToshiya, line 95: `kind: 41` is a replaceable event according to NIP-28 and nostr-rs-relay.

https://github.com/nostr-protocol/nips/blob/1728f93d17fe399fa6e0911e7bb69d2e159ca572/28.md?plain=1#L40

https://github.com/scsibug/nostr-rs-relay/blob/4f518fd0e765d20a8a8d07bc0a17ef9fbb7bfbb6/src/event.rs#L171-L178

AsaiToshiya (changes requested):

01.md
ArjenStens: Thanks for your response, I went through all the other `kind`s and this seems to be the only one mentioning a different lifetime. I now included kind 41.

Looking at kind 40 this leaves me wondering about its purpose. Why does kind 40 (create channel) contain the same metadata that can be modified in kind 41 (set metadata)? Doesn't it make more sense to have just kind 41 fullfill both purposes similar to how [NIP-52](https://github.com/nostr-protocol/nips/blob/master/52.md#format) (calendar event, kind 31923) is set up?

**Couple other points:**
- Just to verify: kind 40 is just of lifecycle regular, correct?
- I don't see any definition of kind 33. But it looks like deleted one given your [recent PR](https://github.com/nostr-protocol/nips/pull/1318).

AsaiToshiya (approved):

AsaiToshiya:
@ArjenStens Sorry, I was wrong. kind 41 is also a regular event. Can you please revert eead2f5 because I want to merge your changes into.

PR #1316, opened 2024-06-18
public key extra compression info

ArmanTheParman:
Important notes about extra pubkey compression used in Nostr


staab (approved):

ArmanTheParman:
Correction made about Schnorr signatures

PR #1318, opened 2024-06-19
NIP-28: simplify and fix kind 41

AsaiToshiya:


ArjenStens (approved):

28.md
```diff
@@ -37,7 +37,7 @@ In the channel creation `content` field, Client SHOULD include basic channel met
 
 Update a channel's public metadata.
 
-Clients and relays SHOULD handle kind 41 events similar to kind 33 replaceable events, where the information is used to update the metadata, without modifying the event id for the channel.Only the most recent kind 41 is needed to be stored.
+Clients and relays SHOULD handle replaceable event kind 41, where the information is used to update the metadata, without modifying the event id for the channel.
```
vitorpamplona, line 40:
```suggestion
Kind 41 is used to update the metadata without modifying the event id for the channel. Clients and relays MAY handle kind 41 as a replaceable event: only the most recent kind 41 MAY be available.
```

vitorpamplona (approved):
Almost no relay treats kind 41 as replaceable, so we should drop it from SHOULD to MAY.

AsaiToshiya:
Maybe kind 41 is not a replaceable event.

staab (approved):

staab:
I never treated it as replaceable when I had chats in coracle. I believe this nip was pre-replaceables.

AsaiToshiya:
Yes, the change in #526 confused me.

PR #1321, opened 2024-06-20
NIP-64 - Trust Score

arthurfranca:
[Read here](https://github.com/arthurfranca/nips/blob/trust-score/64.md)

### DIFF
--- a/64.md
+++ b/64.md
@@ -1,0 +1,85 @@
+NIP-64
+======
+
+Trust Score
+-----------
+
+A trust score is a public way to individually assign a quantifiable level of trust on specifc topics to a nostr identity.
+Ranks can be used to discover trusted people on a common topic.
+
+## Event Structure
+
+The `kind:30382` is a NIP-61 ["Set Item Reference"](https://github.com/arthurfranca/nips/blob/bunch-of-events/61.md#tagging-into-event-sets) event used to assign public information about a pubkey (set to the `d` tag).
+
+This NIP makes use of one or more trust scores added to such events as `T` tags using the following structure:
+
+```js
+  [
+    "T",
+    "<integer_score_ranging_from_-3_to_3><colon_separator><topic_name>",
+    "<+_or_?_or_-><space_separator><review>" // optional and may appear many times
+  ]
+```
+
+Topic names should be preferably in English, lower case and their words are dash separated.
+Topic names are separated from the score by a colon (`:`).
+Some topic names include a context prefix, also separated by a colon (`:`).
+
+After the tag value, there may be any number of reviews, each one with a prefix followed by a space separator
+and the actual review text.
+The prefix denote a postive review (`+`), a negative review (`-`) or a neutral one (`?`).
+
+The `?` prefix is usually assigned for free form user generated reviews. On the other hand, `+` and `-` are usually
+related to clickable fixed options made available by a client.
+
+Example:
+
+```js
+{
+  "kind": 30382,
+  "tags": [
+    ["d", "<pubkey>"],
+    ["T", "3:wiki:role-playing-video-game"], // has written a based wiki article about RPG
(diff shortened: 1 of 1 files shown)


alexgleason:
We really need to get kind 30382 and 30383 in the NIPs. This is just sugar on top.

arthurfranca:
Seeing https://github.com/Trustroots/nostroots/issues/20 from https://github.com/nostr-protocol/nips/pull/1208#issuecomment-2180204382 comment, I rewrote this trust score proposal .

Imo the above is better than the NIP-77 from #1208. For example, using NIP-77 in the current form a client can't fetch just "good hosts" directly from relays. While NIP-64 make it possible by requesting events with this filter: `{ "kinds": [30382], "#T": ["3:inn:host", "2:inn:host", "1:inn:host"], authors: ["people", "i-became-friends-with", "at-trust-roots"] }`

Another good thing about NIP-64 is it reuses `kind:30382` event that is meant to enable an user adding public information about another user, not just trust/review data.

huumn:
Stray recommendation: it might be worth providing a design that's compatible with float, rather than int, trust scores. 

I haven't seen floats used much in the NIPs last time I read them, so maybe there's a bend against them, but for something like trust, it can help to have more granularity, especially if trust isn't entirely explicit. 

So long as the float is constrained to `[-(2**53)+1, (2**53)-1]` it should be interoperable with most JSON implementations.[^1]

[^1]: https://www.rfc-editor.org/rfc/rfc7159.html#section-6

64.md
```diff
@@ -0,0 +1,85 @@
...
+    ["T", "3:buyer", "? No problems when selling to this person"],
+    ["T", "-2:seller", "? Had a bad experience when buying from this person."],
+    ["T", "0:inn-host"], // This person hosted me but I assigned no trust score yet
+    ["T", "3:inn-guest"] // I hosted this person and had no problem
+  ],
+  // ...other fields
+}
+```
+
+If a client doesn't recognize a topic name, it should leave the corresponding `T` tag untouched when editing other entry.
```
dluvian, line 53: If a client doesn't recognize a topic name it will probably be discarded when updating the event. We shouldn't rely on devs being altruistic enough to write extra code just for the sake of not nuking content of other clients.
We also shouldn't overload events. A wiki client doesn't care about `car` reviews, so they shouldn't be forced to download it when fetching wiki reviews.

Maybe add the topic to the d tag to separate reviews like `["d", "<top-level-topic>:<pubkey>"]` ?

franzaps:
@arthurfranca what's your npub? Can't seem to find you

arthurfranca:
@franzaps npub1l3cgtsurhfchg4cyhhqudm70074sr96srhje330xc5m6czej5n9s9q6vs2

64.md
arthurfranca: This all is yet to be implemented, feel free to test other ideas derived or not from this.

`["d", "<top-level-topic>:<pubkey>"]` works too. Most pubkeys wouldn't have multiple `T` tags, in fact if using `kind:30382` most would have none. Usually an user won't add that many metadata about another user. That's why currently I think stuffing diverse metadata inside the same kind with `d-tag=another-user` isn't a problem.

arthurfranca:
@huumn agree some use cases may need more precision. Problem with floats is you can't ask relays for all events with trust within a specific range of floats, that's why i used (a small set of) integers.

A `["trust", "context", "1.33"]` tag, to be read client-side, could be included in addition to `["T", "1:context"]`

huumn:
> @huumn agree some use cases may need more precision. Problem with floats is you can't ask relays for all events with trust within a specific range of floats, that's why i used (a small set of) integers.

Great point. afaik all queries use equality and set inclusion because there hasn't been cause for range queries yet.

PR #1322, opened 2024-06-20
NIP-60: Record Events

alexgleason:
This is a new take on kind 30382 and 30383 events, originally defined in Event Sets: https://github.com/nostr-protocol/nips/pull/784 (Related: #761 #1321)

But it is a spec for generic records of pubkeys and event IDs, and it does not define `n` tags.

Read here: https://github.com/nostr-protocol/nips/blob/619ff20a871e4316ec59c2b211f20b4f8df26853/60.md

cc @arthurfranca @vitorpamplona

### DIFF
--- a/60.md
+++ b/60.md
@@ -1,0 +1,57 @@
+NIP-60
+======
+
+Record Events
+-------------
+
+`draft` `optional`
+
+This NIP defines a way for users to keep editable records about other people and events.
+
+## Pubkey Record
+
+A pubkey record is a kind 30082 event where the `d` tag points to the pubkey of another user.
+
+```js
+{
+  "kind": 30382,
+  "tags": [
+    ["d", "<pubkey>"],
+    ["n", "colleague"],
+    ["summary", "This is my boss"],
+    // other tags
+  ],
+  // other fields
+}
+```
+
+Any tags can be used. Specific tag uses may be formalized in other NIPs.
+
+The `content` field may be an empty string, or a NIP-44 encrypted tag list of private tags.
+
+## Event Record
+
+An event record is a kind 30083 event where the `d` tag points to the id of another event.
+
+```js
+{
+  "kind": 30383,
+  "tags": [
+    ["d", "<event-id>"],
+    ["k", "1984"],
+    ["p", "<event-pubkey>"],
+    ["n", "open"],
+    // other tags
+  ],
+  // other fields
+}
+```
+
+This event MUST include the following tags:
+
+- `k` - the kind of the referenced event.
+- `p` - the pubkey of the referenced event (for outbox support).
+
+Any other tags can also be used. Specific tag uses may be formalized in other NIPs.
+
+The `content` field may be an empty string, or a NIP-44 encrypted tag list of private tags.
--- a/README.md
+++ b/README.md
@@ -73,6 +73,7 @@ They exist to document what may be implemented by [Nostr](https://github.com/nos
 - [NIP-57: Lightning Zaps](57.md)
 - [NIP-58: Badges](58.md)
 - [NIP-59: Gift Wrap](59.md)
+- [NIP-60: Record Events](60.md)
 - [NIP-65: Relay List Metadata](65.md)
 - [NIP-71: Video Events](71.md)
 - [NIP-72: Moderated Communities](72.md)
@@ -177,6 +178,8 @@ They exist to document what may be implemented by [Nostr](https://github.com/nos
 | `30078`       | Application-specific Data  | [78](78.md)              |
 | `30311`       | Live Event                 | [53](53.md)              |
 | `30315`       | User Statuses              | [38](38.md)              |
+| `30382`       | Pubkey Record              | [60](60.md)              |
+| `30383`       | Event Record               | [60](60.md)              |
 | `30402`       | Classified Listing         | [99](99.md)              |
 | `30403`       | Draft Classified Listing   | [99](99.md)              |
 | `30617`       | Repository announcements   | [34](34.md)              |


60.md
```diff
@@ -0,0 +1,57 @@
...
+
+The `content` field may be an empty string, or a NIP-44 encrypted tag list of private tags.
+
+## Event Record
+
+An event record is a kind 30083 event where the `d` tag points to the id of another event.
+
+```js
+{
+  "kind": 30383,
```
vitorpamplona, line 38: Can you choose another number? 30383 was already used for the private relationship between two keys https://github.com/nostr-protocol/nips/pull/761

vitorpamplona: I will change it on

arthurfranca:
I think it is better to try merging #784. Although 30382/30383 are important events, I think the others should also be listed, along with the standard `n` values, to enable migration from NIP-51 where it fits (be it some of the current NIP-51 lists or future ones).

For example, currently a mute list can include "p" (pubkeys), "t" (hashtags), "word" (lowercase string) or "e" (threads). So to mimick this we need events 30382, 30385, 30389 and 30383 with `n-tag=mute`.

60.md
arthurfranca: @vitorpamplona Why are you still invested in #761 when Event Sets supersedes it?
Though I used 31382 instead of 30383, it is the same thing.

alexgleason: I'm already publishing kind 30383 events in Ditto.

alexgleason:
@arthurfranca I want to get these kinds in the nips sometime this decade, and I think there's a better chance when it's kept simple. We can always expand this document later.

I have actually implemented kind 30382/30383 events as spec'd and am using it in a real project. I haven't implemented any of the others things in #784

60.md
vitorpamplona: > @vitorpamplona Why are you still invested in #761 when Event Sets supersedes it? Although I used 31382 instead of 30383, it is the same thing.

Because you kept changing it and I needed a stable version to ship. 

Regardless, #761 is about defining all tags that can be used to clarify the relationship between two keys. It's not about just putting users into sets.

arthurfranca: At #784, 30382..30389 range is used for public sets, then I added 1000 to the private versions: 31382..31389.

@vitorpamplona @alexgleason you could tell me new numbers to put there instead if you think it is better.

@vitorpamplona your #761 tags are present in the event sets example. The other things like the mapping part and the obfuscating thing with sha256+hkdf are also already there at #784 to make it more generic to be used for any set of referenced items, not just the ones referencing pubkeys.

vitorpamplona:
I like simple NIPs that solve a specific problem, like this or #761, as opposed to a spec describing the basic framework to tag things like #784. It allows these NIPs to be expanded to support new tags for each kind. 

However, #784 is a great base if people need to tag URLs, words, or other things. Maybe #784 mixed two ideas: 
- Putting events into sets. 
- "Labeling" things with replaceable events. 

I think if we want Event Sets to become a thing, #784 could abandon the definition of standard d-tags, and just introduce the `n` tag with the `nset`. Then we can move each one of the "Standard" sets to it's own NIP and add more tags to help annotate these relationships. 

In some ways, this NIP is more of a replacement for #761 than for #784. It just doesn't care about the private stuff on #761.

alexgleason:
I decided to strip it down because of https://github.com/nostr-protocol/nips/pull/1264

> They should be fully implemented in at least two clients and one relay -- when applicable.

I have not actually implemented private event sets, and don't have any immediate need to, even though I think they're a nice idea.

With this version we can actually get two implementations going and merge this.

arthurfranca:
I think the event sets nip is very similar to NIP-51 in that lists/sets should be inside the same NIP.

Though relationship status does could be a separate thing by defining specifc tags (although these tags are present on event sets NIP as examples, I don't really explain them). The same way as #1321 defines a new tag and how to use them.

arthurfranca:
> I decided to strip it down because of https://github.com/nostr-protocol/nips/pull/1264

This is just for looks. Instead of two implementations, a NIP proposal needs two **blessings** in order to be merged.
One from @fiatjaf and another one from one of these: @staab, @pablof7z or @jb55.
😁️

staab:
I still stand by my comments [here](https://github.com/nostr-protocol/nips/pull/784#issuecomment-2181384288), but this is much better. As a primitive with far less semantics attached to it, it could be useful in novel situations. I do think it still violates the "do things one way" rule, and could lead to fragmentation, but it's at least concise.

alexgleason:
The idea proposed here is simple. I would compare it to NIP-78. It is an emergent property of parameterized replaceable events. It seems obvious to me to have a generic event kind for `d` tagging a pubkey or event ID. Every developer thinks of this when they first read NIP-01. It WANTS to exist.

Event Sets would use this, but this NIP is specifically not Event Sets, even though it is inspired by it (I updated the OP to clarify). The NIP-61 proposal does not actually _name_ the event kinds they propose, and only in #1321 did I first see the name "Set Item Reference". What I am proposing is more generic: "Record Event".

This NIP would not make Event Sets a reality. It would simply define and reserve these event kinds, because they are already being used.

vitorpamplona:
I would change the name from Event Record (I don't know what it is meant to mean) to something like "Editable Stickers" because "Editable" is the most important part of this spec and you are sticking things into existing events and pubkeys.

arthurfranca:
Now that its clear we don't want to use event sets to migrate from NIP-51, I see it is a good idea to remove the "Set Item Reference" event from there and add them here.

Starting with references for pubkeys and events may be the best approach. Event kinds to reference other things could be added as people see the need. But I think 30384 for referencing a replaceable event seems like a good addition?

I think it should also include the corresponding private kind numbers and the part [from here](https://github.com/nostr-protocol/nips/blob/1f06fb10741d8e1ab22d5de5f9d80b46a675c7a0/61.md#private-set-item-reference-event) that teaches how to obfuscate one-letter-tags (that can't be encrypted to .content to remain searchable) for the private use case.

arthurfranca:
Well, in fact #784 already teaches the obfuscating part and introduces one of the private "record events".

I understand you want to keep this lean with just the part that has already been implemented. No need to change the PR.

alexgleason:
NIP-85 does what I want. Same kinds.

PR #1323, opened 2024-06-20
Add contact info field to profile

GuardianPlasma:


AsaiToshiya:
I think these things can be written in the `about` field.

PR #1326, opened 2024-06-23
NIP-122: Request For Events

cameri:


vitorpamplona:
Nice! 

You should allow a tags as well.

cameri:
For direct requests?

vitorpamplona:
Yep, you can ask using e tags and/or a tags for the last version of a replaceable.

mikedilger:
If this becomes widely implemented, I could see someone posting an ephemeral public request for missing events, and then thousands of clients all responding with those events flooding the relay they are communicating with.

I'm not sure by what mechanism that can be avoided.

cameri:
> If this becomes widely implemented, I could see someone posting an ephemeral public request for missing events, and then thousands of clients all responding with those events flooding the relay they are communicating with.
> 
> 
> 
> I'm not sure by what mechanism that can be avoided.

Clients can respond to a sample of the incoming requests, rate limit, have quotas for this function or data caps, require PoW, respond only from their Web of Trust.

There's plenty of solutions to mitigate that scenario.

cameri:
> If this becomes widely implemented, I could see someone posting an ephemeral public request for missing events, and then thousands of clients all responding with those events flooding the relay they are communicating with.
> 
> 
> 
> I'm not sure by what mechanism that can be avoided.

Relays also mitigate the effect by not forwarding if the event ID is stored.

vitorpamplona:
It might also be a hint to the relay or to a relay service to find those events because your users want them. 

There could be services that just connect to relays and re-broadcast events from other relays based on these requests.

fabianfabian:
Interesting approach. Could maybe also eventually be extended to respond with wrapped events if events are not accepted (maybe too old or other reason).

PR #1335, opened 2024-06-28
NIP-39 cryptographic identities

franzaps:
Following the discussion on https://github.com/nostr-protocol/nips/pull/1182 I have:
 - Integrated my changes onto NIP-39 (from what was NIP-69)
 - Updated the PGP type according to https://metacode.biz/openpgp/openpgp4fpr
 - Normalized the proof message across platforms and cryptographic identities

### DIFF
--- a/39.md
+++ b/39.md
@@ -13,25 +13,33 @@ Nostr protocol users may have other online identities such as usernames, profile
 ## `i` tag on a metadata event
 
 A new optional `i` tag is introduced for `kind 0` metadata event contents in addition to name, about, picture fields as included in [NIP-01](https://github.com/nostr-protocol/nips/blob/master/01.md):
-```json
+
+```jsonc
 {
+  "kind": 0,
   "tags": [
     ["i", "github:semisol", "9721ce4ee4fceb91c9711ca2a6c9a5ab"],
     ["i", "twitter:semisol_public", "1619358434134196225"],
-    ["i", "mastodon:bitcoinhackers.org/@semisol", "109775066355589974"]
-    ["i", "telegram:1087295469", "nostrdirectory/770"]
+    ["i", "mastodon:bitcoinhackers.org/@semisol", "109775066355589974"],
+    ["i", "telegram:1087295469", "nostrdirectory/770"],
+    ["i", "openpgp4fpr:1A04E0F1A78D982BD8885B7EB325A9C5F70849D0", "LS0tLS1CRUdJTiBQR1AgTUVTU0FHRS0tLS0tCgpvd0did012TXdDVzJXWFhsMGU4Y25oY1lUeTlJWWtpVGw0bHpxbFFvemt6UHk4eExWeWpKeUN4V3lFMHRMazVNClQxWHdWRWpPejB2TExNb0ZDaWVXUUxnbFJmazVRRzZxUWtGUlpsbGlTYXBDZG1xbFFscCtFVmpNTDcrNHBFaWgKb0RRcEp6TVpMSkVIWkJ1V3A1a1VsS1lWbHlhbkZwbVdKVnFrV3hZWUZHV1pwdVJsbEtXbFpwaWxXSlFicEp1bApKbFltcGhhWXBtUVVGSnNWRmFjWG14aW5wR2RZY25XVXNqQ0ljVEhJaWlteVNMRTgrTGk4ZDRiMmpZN29PcGdQCldKbEF6bWZnNGhTQWlTU0pNVExjL2E3YmRtZUx3S0dMZWs2NkRTcDYvNmRmMzl6Nk5pN1E1dmcxLytCU0NRY1oKUm9ZOUwvN2NhWW05WnRZN2FkNm1GMy95ak83OVZqbWhZZmF4bm1mNlFaYldDM0lNQUE9PQo9ZXkyagotLS0tLUVORCBQR1AgTUVTU0FHRS0tLS0tCg", "LS0tLS1CRUdJTiBQR1AgUFVCTElDIEtFWSBCTE9DSy0tLS0tCgptRE1FWmdIZE94WUpLd1lCQkFIYVJ3OEJBUWRBNnBaZ2RzbUR2aVJGNDVDYys5TUJiK3VkUURZZ2ZHRjZrQkJVCk1CeUd4S2EwRjJaeVlXNTZZWEFnUEY5QVpuSmhibnBoY0M1amIyMCtpSk1FRXhZS0FEc1dJUVFhQk9EeHA0MlkKSzlpSVczNnpKYW5GOXdoSjBBVUNaZ0hkT3dJYkF3VUxDUWdIQWdJaUFnWVZDZ2tJQ3dJRUZnSURBUUllQndJWApnQUFLQ1JDekphbkY5d2hKMEw5R0FRQ3VFUFF0aW9EdGlKaWNVSGw4S0RqU0dmRU1laEFWdjZsTGJuZDBTRjN6Ci9BRCtQZWJCcFk3YkwvakY4QnViaU8rVm9nbUhsa2NGZ3h5V2ZEMDYwMDAwb0FvPQo9OWRBUAotLS0tLUVORCBQR1AgUFVCTElDIEtFWSBCTE9DSy0tLS0tCg"]
   ],
-  ...
+  // ...
 }
 ```
 
-An `i` tag will have two parameters, which are defined as the following:
-1. `platform:identity`: This is the platform name (for example `github`) and the identity on that platform (for example `semisol`) joined together with `:`.
-2. `proof`: String or object that points to the proof of owning this identity.
+An `i` tag will have two parameters and an optional third parameter, which are defined as the following:
+1. `type:identity`: This is the platform name (for example `github`) or the linked cryptographic identity type (for example `openpgp4fpr`), and the identity on that platform (for example `semisol`) or public key fingerprint (for example `1A04E0F1A78D982BD8885B7EB325A9C5F70849D0`) joined together with `:`.
+2. `proof`: String that points to the proof of owning this identity or is the proof itself (crytographic signature).
+3. The full public key in the case of a cryptographic identity.
 
-Clients SHOULD process any `i` tags with more than 2 values for future extensibility.
+Clients SHOULD process any `i` tags with more than 3 values for future extensibility.
 Identity provider names SHOULD only include `a-z`, `0-9` and the characters `._-/` and MUST NOT include `:`.
 Identity names SHOULD be normalized if possible by replacing uppercase letters with lowercase letters, and if there are multiple aliases for an entity the primary one should be used.
+
+## Revocation
+
+The event MAY be updated by removing `i` tags that no longer apply.
 
 ## Claim types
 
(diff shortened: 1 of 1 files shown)


39.md
```diff
@@ -13,26 +13,34 @@ Nostr protocol users may have other online identities such as usernames, profile
...
 }
 ```
 
-An `i` tag will have two parameters, which are defined as the following:
-1. `platform:identity`: This is the platform name (for example `github`) and the identity on that platform (for example `semisol`) joined together with `:`.
-2. `proof`: String or object that points to the proof of owning this identity.
+An `i` tag will have two parameters and an optional third parameter, which are defined as the following:
+1. `type:identity`: This is the platform name (for example `github`) or the linked cryptographic identity type (for example `openpgp4fpr`), and the identity on that platform (for example `semisol`) or public key fingerprint (for example `1A04E0F1A78D982BD8885B7EB325A9C5F70849D0`) joined together with `:`.
+2. `proof`: String that points to the proof of owning this identity or is the proof itself (crytographic signature).
+3. The full public key in the case of a cryptographic identity.
```
caesar, line 34: OpenPGP keys can be very large, depending on the number of subkeys, annotations, signatures, etc – mine is 38kB and I have seen much larger.
Rather than embedding the whole key, it would make more sense to provide a URL where it can be downloaded.
Either way, I think it should be optional, and if absent clients can fetch it from keys.openpgp.org or another default keyserver of their choice.

39.md
```diff
@@ -62,3 +70,79 @@ Identity: A Telegram user ID.
 
 Proof: A string in the format `<ref>/<id>` which points to a message published in the public channel or group with name `<ref>` and message ID `<id>`. This message should be sent by user ID `<identity>` and have the text `Verifying that I control the following Nostr public key: "<npub encoded public key>"`.
 This can be located at `https://t.me/<proof>`.
+
+### `openpgp4fpr`
+
+Identity: A PGP fingerprint as described in the [openpgp4fpr URI scheme](https://metacode.biz/openpgp/openpgp4fpr), uppercase.
```
caesar, line 76: This should be case insensitive as described in the linked document.

caesar (changes requested):
Thanks @franzaps! A couple of suggestions below.

39.md
caesar: Also, ideally, mentions of `PGP` should instead be `OpenPGP`, though I'll admit I'm being pedantic here.

39.md
dipunm: I think at such a size, and for the sake of simplicity, the key should be elsewhere. Maybe an optional bloomserver hash is fine, but raw key may break clients if the key is so large.

Is there a bloomserver uri scheme?

franzaps: What is a bloomserver?

39.md
franzaps: Are relay queries case-insensitive? Probably not. That's all I care about.

39.md
franzaps: @caesar My God, PGP is painful. Is there any way to compress and throw away any subkeys and annotations and just leave the meat of it?  I want to avoid having to ping keyservers as much as possible, they are a security hole.

caesar: Hmm, I'm not sure. I don't really see a problem with just providing an external URI for the key though, all the existing proofs are provided via external links anyway. (As for keyservers, I agree, but it's part of the PGP system and not really our problem to solve in Nostr in my opinion.)

39.md
caesar: Interesting point, I don't know the answer.

39.md
franzaps: Okay. For people who think their key is too heavy (we could suggest a threshold, TBD) we should recommend adding multiple keyservers to mitigate both DNS and keyserver operator rugging risks

39.md
franzaps: There are a lot of mentions of "lowercase" in this repo, and also this NIP contained:
> Identity names SHOULD be normalized if possible by replacing uppercase letters with lowercase letters, and if there are multiple aliases for an entity the primary one should be used.

I'll update to reflect lowercase. It costs nothing to lowercase everything before publishing and makes it easy to query.

39.md
dipunm: +1 on the threshold. Good compromise.

dipunm: @caesar It was late, but I meant blossom server: https://github.com/hzrd149/blossom

It gives some DNS issue resistance, but I guess so does listing keyservers. I just like the idea of referencing and downloading files by their hash.

Maybe it is a bad idea, the wiki suggests that key servers are built more for purpose and have evolved over time: https://en.m.wikipedia.org/wiki/Key_server_(cryptographic)
(See problems with keyservers section)

franzaps: I will add the ability to specify URLs as a fallback when public key material is too heavy. Blossom is great but we should not be prescriptive: if people share a Blossom-like URL then it can be assumed as such.

39.md
```diff
@@ -13,26 +13,36 @@ Nostr protocol users may have other online identities such as usernames, profile
...
+    ["i", "openpgp4fpr:1A04E0F1A78D982BD8885B7EB325A9C5F70849D0", "LS0tLS1CRUdJTiBQR1AgTUVTU0FHRS0tLS0tCgpvd0did012TXdDVzJXWFhsMGU4Y25oY1lUeTlJWWtpVGw0bHpxbFFvemt6UHk4eExWeWpKeUN4V3lFMHRMazVNClQxWHdWRWpPejB2TExNb0ZDaWVXUUxnbFJmazVRRzZxUWtGUlpsbGlTYXBDZG1xbFFscCtFVmpNTDcrNHBFaWgKb0RRcEp6TVpMSkVIWkJ1V3A1a1VsS1lWbHlhbkZwbVdKVnFrV3hZWUZHV1pwdVJsbEtXbFpwaWxXSlFicEp1bApKbFltcGhhWXBtUVVGSnNWRmFjWG14aW5wR2RZY25XVXNqQ0ljVEhJaWlteVNMRTgrTGk4ZDRiMmpZN29PcGdQCldKbEF6bWZnNGhTQWlTU0pNVExjL2E3YmRtZUx3S0dMZWs2NkRTcDYvNmRmMzl6Nk5pN1E1dmcxLytCU0NRY1oKUm9ZOUwvN2NhWW05WnRZN2FkNm1GMy95ak83OVZqbWhZZmF4bm1mNlFaYldDM0lNQUE9PQo9ZXkyagotLS0tLUVORCBQR1AgTUVTU0FHRS0tLS0tCg", "LS0tLS1CRUdJTiBQR1AgUFVCTElDIEtFWSBCTE9DSy0tLS0tCgptRE1FWmdIZE94WUpLd1lCQkFIYVJ3OEJBUWRBNnBaZ2RzbUR2aVJGNDVDYys5TUJiK3VkUURZZ2ZHRjZrQkJVCk1CeUd4S2EwRjJaeVlXNTZZWEFnUEY5QVpuSmhibnBoY0M1amIyMCtpSk1FRXhZS0FEc1dJUVFhQk9EeHA0MlkKSzlpSVczNnpKYW5GOXdoSjBBVUNaZ0hkT3dJYkF3VUxDUWdIQWdJaUFnWVZDZ2tJQ3dJRUZnSURBUUllQndJWApnQUFLQ1JDekphbkY5d2hKMEw5R0FRQ3VFUFF0aW9EdGlKaWNVSGw4S0RqU0dmRU1laEFWdjZsTGJuZDBTRjN6Ci9BRCtQZWJCcFk3YkwvakY4QnViaU8rVm9nbUhsa2NGZ3h5V2ZEMDYwMDAwb0FvPQo9OWRBUAotLS0tLUVORCBQR1AgUFVCTElDIEtFWSBCTE9DSy0tLS0tCg"]
```
alexgleason, line 25: It's not that big. I don't see the problem. Sometimes it can be big due to choices made by the user? Well the user should just not do that. Fortunately the user is always a developer in this case, so the UX of that doesn't seem important.

alexgleason: Have you seen the size of kind 3 events?

franzaps: That's a great point, actually. By extension kind 0s could also potentially be quite heavy but probably not the norm. It was mentioned earlier in the thread that some PGP keys can take well over 40kb

39.md
franzaps: `gpg --edit-key` shows different options to reduce size like `clean`, `minimize`, `delphoto` etc

franzaps:
> I don't like changing events that are already well defined and supported. Kind 0 events have content to stringified JSON (yuck) instead of tags. This specifies adding tags which then breaks all the current clients that don't look at tags in kind 0 events.

Confused by your comment. NIP-39 specifies the usage of `i` tags and is well established. Here I'm simply suggesting a few extra types. Adding tags does not break clients.

> I would rather you make a new kind in the 1xxxx range.

I did (this used to be kind 10069) and was asked to open a new PR to NIP-39. https://github.com/nostr-protocol/nips/pull/1182

franzaps:
@Semisol and @staab - as requested by in https://github.com/nostr-protocol/nips/pull/1182 here's the PR to NIP-39. Can we get this merged?

mikedilger:
Oh I didn't know NIP-39 already did this.  What I meant by "break" was that they wouldn't automatically do anything with this data, and they would be considered by some to be non-compliant with kind=0 even though they were totally compliant when they were programmed.

erskingardner (approved):

franzaps:
Updated the spec (and examples) to use binary format instead of hex/ASCII before converting to base64. Double-checked and verified the results and space optimization appears to be significant.

39.md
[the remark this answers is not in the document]
franzaps: How about no?

https://github.com/nostr-protocol/nips/pull/1182#pullrequestreview-2061353840

aaccioly: Trying to resurrect this conversation, given that the Keyoxide folks are implementing Nostr claims.

Aside from the "PGP keys can be big" issue, I also don’t think embedding signed texts is the best form of proof. I.e., if my nsec is compromised there isn't much that I can do with my PGP key to "revoke" the proof.

We can follow Keyoxide’s approach and store the proof in the key itself using notation data:
https://datatracker.ietf.org/doc/html/rfc4880#section-5.2.3.16
For example, to keep things symmetric with Keyoxide, my key could include a notation like:

```
proof@ariadne.id=nostr:npub1a6we08n7zsv2na689whc9hykpq4q6sj3kaauk9c2dm8vj0adlajq7w0tyc
```

Then my kind 0 could simply have an i tag with:

```
openpgp4fpr:OPENPGP_FINGERPRINT
```

Just like Keyoxide, we can retrieve the key itself via WKD or a standard keyserver (Keyoxide supports both WKD and keys.openpgp.org out of the box), and validate the claim by checking for a notation that matches my npub.

Of course, we can support a second optional field to fetch from alternative keyservers or even a direct link to a .asc or .gpg file containing the public key. But this is mostly optional, Keyoxide has been working for ages using just WKD and keys.openpgp.org.

If I lose my nsec, "revoking" is as simply as editing my key to remove the notation and re-uploading to my WKD or key server.

Further details on Nostr (because I didn’t know this proposal was a thing and started my own conversation there):

https://njump.me/nevent1qqs8zr5mpn5tfajspc0uz6pf6t67uzp4ketahh4wtk4jk750h9cwsfsprdmhxue69uhksctkv4hzuctrvd5k7mre9eek7cmfv9kz7q3qa6we08n7zsv2na689whc9hykpq4q6sj3kaauk9c2dm8vj0adlajqxpqqqqqqzypf5dn

franzaps: What do you think are the drawbacks of this approach compared to embedding the signed text? How common are WKDs?

I care much more about x509 at the moment (will start implementing quite soon). Would you be interested in modifying this PR? Or perhaps I could remove the PGP part and let you submit a different PR

aaccioly: Well, IMO, the counterpoint here is a bit more complexity in the validation, with network calls to services outside of Nostr that may become unavailable. On the other hand, this is also what makes it a more suitable proof that the user has control over the key.
WKD is not yet widely used, but it is slowly gaining adoption (and many keyservers are dying).

I’m happy to take a stab at it. I think it makes sense to group this together; otherwise the PGP proof stuff will just be another invisible PR in this repo. What would you prefer me to do? Would you like to grant me write access to your fork/branch, or would you prefer me to submit editable suggestions using GitHub’s interface?

franzaps: Makes sense.

Most PRs are invisible, I don't put much importance on these, they never get merged anyway, implementation is what counts. If you can submit changes here great

franzaps:
Added expiry.

x509 has been implemented in the new publishing tool and be rolled out to developers

Semisol (changes requested):
The public key should not be included in the `i` tag proof. If this is to be published, it should use a dedicated event kind.

franzaps:
> The public key should not be included in the `i` tag proof. If this is to be published, it should use a dedicated event kind.

Why?

Semisol:
> > The public key should not be included in the `i` tag proof. If this is to be published, it should use a dedicated event kind.
> 
> Why?

It is a large blob that clogs up the event more than is needed. If anything, I'd say move the signature as well into an event.

A key can be obtained using WKD, or with a special event kind, or from key servers, so on... and for X.509, you need the cert not the public key.

Also: I'd recommend committing to SPKIFP and allowing multiple certs that way.

franzaps:
Decided not to bother with PGP for now.

Good idea about SPKIFP, I redesigned the NIP around that and using a new kind.

No we don't need certs - we assume the user encountered an APK (thus a certificate) and wants to verify it was signed by the Nostr pubkey claiming it.

franzaps:
NIP home: https://nostrhub.io/naddr1qvzqqqrcvypzq7xwd748yfjrsu5yuerm56fcn9tntmyv04w95etn0e23xrczvvraqy28wumn8ghj7un9d3shjtnyv9kh2uewd9hsq8trwfuhqar0vaexzurgd93j66tyv4h8g6t50ykhqun0dan8xmjazhm

First event: https://njump.me/naddr1qpqrqc3k8yckydmyxvcxzdr9893nqvtzxyuxgvryxfjxgdf3v5enjdt9xqmkzdrpxpnrgvt9xcckycnyvguxvetpvy6kvefsx5erjdmrxgq3vamnwvaz7tmjv4kxz7fwwpexjmtpdshxuet5qgs83nn04fezvsu89p8xg7axjwye2u67errat3dx2um725fs7qnrqlgrqsqqqaedrmtgvw

Semisol:
> No we don't need certs - we assume the user encountered an APK (thus a certificate) and wants to verify it was signed by the Nostr pubkey claiming it.

Not anymore, yep. But when it used the cert fingerprint, it was not possible to verify the public key in the claim is the one in the certificate.

I'd also recommend not using replaceable events for revocations of proofs, but a dedicated regular kind. This means no relay changes, and prevents the weird handling logic required for the following case of replacements:
- event at time 1, initial proof received by relay
- event at time 3, extension of expiration received by relay
- event at time 2, revocation received

This would mean a special path would be needed to do replacements when revocation happens.

C1.md
```diff
@@ -0,0 +1,189 @@
...
+
+| Kind  | Type  | Description                                           |
+|-------|-------|-------------------------------------------------------|
+| 30509 | SPKI  | SubjectPublicKeyInfo (SPKI) extracted from code-signing certificates (e.g., Android APK signing) |
+
+Future versions may add support for additional types such as OpenPGP.
+
+## Event Format
+
+A cryptographic identity proof is published as a parameterized replaceable event:
```
kehiy, line 23:
```suggestion
A cryptographic identity proof is published as a addressable event:
```

I think it's the correct name now(?).

PR #1336, opened 2024-06-29
Software applications events

franzaps:
[Rendered](https://github.com/franzaps/nips/blob/applications/82.md)

This event represents an application as used in [Zapstore](https://zapstore.dev). It is actively being used in a production setting for many months already.

This relates to NIP-51 (kind 30063), those artifact sets point to their application (this event, kind 32267).

Feedback welcome


vitorpamplona:
hum.. isn't it possible to merge this with kind 31990? Kind 31990 is insufficiently weirdly defined on NIP but most apps already have their entry due to https://nostrapp.link/

We could just add the additional tags needed by zap.store to that definition.

franzaps:
NIP-89 is for apps that handle nostr events, isn't it?  It's not at all what I'm defining here. In fact, 90%+ of the apps on zap.store do not handle nostr events.

It feels very contrived to try to fit this simple definition on the complexity of NIP-89.

vitorpamplona:
Yes, that's what NIP-89 needed it for, but those app definitions don't need to exist to solely respond to Nostr events. Actually, I believe most apps don't event that "response" field setup correctly.

DVMs also reuse the same kind to define their  applications. 

Reusing the kind doesnt mean you have to implement NIP-89 to support yours.

franzaps:
Are you saying I can keep my schema as-is and change 32267 to 31990? In such case I would have to change `content` which will break other NIP-89 clients.

vitorpamplona:
Hum.. yeah I missed the `content` discrepancy. 

@pablof7z, do you see a possibility of merging those two? 

Or are we going to have to define our apps in two places?

franzaps:
Do you know what is the rationale for putting stringified JSON in `content` (unless it has to be encrypted)? I've noticed the pattern in several places

vitorpamplona:
> Do you know what is the rationale for putting stringified JSON in content

Dump decisions. They probably just copied how kind 0 works :(

franzaps:
I also need data to be in tags for full text indexing purpose.

31990 seems built for something else which is fine: all apps can sign a 32267 and if they want nostr interop also sign a 31990.

I will keep using kind 32267 unless there's a better alternative to describe an application

fiatjaf (approved):

DanConwayDev:
Love to see this and look forward to adding releases to nip34 clients.

Instead of referencing a nip34 announcement event in `repository` tag it should be referenced with an `a` tag so it will be indexed by relays. clients should look for an a tag with `30617` kind if repository is omitted.

The release event:
1. should be moved into this nip from the example section of nip51.
2. should be immutable and not a replaceable event. then trust attestations could be placed on the event  eg. validation that the checksums of a reproducible build match. 
3. should include an optional `["commit-id","<SHA1-commit-id-used-to-build-release>"]` tag

From a style point of view: I see most nips describe events using jsonc rather than a table of tags, which is easier to read. Is this now the standard?

lastly, the app event could support the maintainers tag like nip34 announcements  This means clients could optionally interpret app events by the pubkeys listed with identical `d` tags as referring to the same app. `["maintainers", "<other-recognized-maintainer>", ...]`

franzaps:
> Love to see this and look forward to adding releases to nip34 clients.

Looking forward too!

> Instead of referencing a nip34 announcement event in `repository` tag it should be referenced with an `a` tag so it will be indexed by relays. clients should look for an a tag with `30617` kind if repository is omitted.

Sounds great

> The release event:
> 
>     1. should be moved into this nip from the example section of nip51.

Sure, I can do that

>     2. should be immutable and not a replaceable event. then trust attestations could be placed on the event  eg. validation that the checksums of a reproducible build match.

We're probably talking about different things. The release set is a NIP-51 replaceable event (kind 30063) that is simply a pointer to a set of software releases (binaries). Trust attestations should be applied to NIP-94 (kind 1063) events because they carry a checksum of the binaries. Moreover, if a developer made a mistake in the file collection or a typo in the release notes, it's useful that they can update the event.

>     3. should include an optional `["commit-id","<SHA1-commit-id-used-to-build-release>"]` tag

Sure - in kind 1063s

> From a style point of view: I see most nips describe events using jsonc rather than a table of tags, which is easier to read. Is this now the standard?

I don't think we have a standard for whether to include tables or not. Other NIPs use bullet points.

> lastly, the app event could support the maintainers tag like nip34 announcements This means clients could optionally interpret app events by the pubkeys listed with identical `d` tags as referring to the same app. `["maintainers", "<other-recognized-maintainer>", ...]`

This is currently done via attribution `p`/`zap` tags. I think a better idea is to include maintainers in releases (not apps).

bezysoftware:
> > Do you know what is the rationale for putting stringified JSON in content
> 
> Dumb decisions. They probably just copied how kind 0 works :(

Sorry to go off topic, but what is so bad about putting stringified json in content? Why are tags better?

franzaps:
@bezysoftware I would ask what is _good_ about putting stringified JSON in `content`. It makes no sense to encode JSON inside JSON for no gain (unless it needs to be encrypted) since values of that object are strings anyway. Tags can be indexed.

82.md
```diff
@@ -0,0 +1,35 @@
...
+The `.content` of these events is a full description of the application in Markdown format.
+
+And tags as follows:
+
+| tag         | description                                 | required | example               |
+| ---         | ---                                         | ---      | ---                   |
+| `d`         | App identifier (preferably in reverse domain notation)   | yes      | com.example.app       |
+| `name`      | Name of the app                             | yes      | Example App           |
+| `description` | Short description of the app              | no       | An app that exemplifies    |
+| `repository`| Source code repository URL or [NIP-34](34.md) repository (if open source) | no       | https://github.com/example/android or `30617:e4ee4fceb9721ce4ee4fceb91c9711ca2a6c9a5ab:example-repo` |
```
DanConwayDev, line 22:
```suggestion
| `repository`| Source code repository URL | no       | https://github.com/example/android |
| `a`| address pointer to [NIP-34](34.md) repository | no       | `30617:e4ee4fceb9721ce4ee4fceb91c9711ca2a6c9a5ab:example-repo` |
| `commit-id`| SHA1 commit id used to build the release  | no       | `42981750c3ab3bacaf242bf05a623ae1b299e0a1` |

```

DanConwayDev (changes requested):
here is a suggested update reflecting the comments you were receptive to

franzaps:
I rewrote good part of this NIP.

As suggested by @NielLiesmons application description should be a wiki article. With that, developers also get access to better tools for updating information.

Presented the tags in `jsonc` example format as suggested by @DanConwayDev 

Lastly, made a clarification on NIP-89.

Events on relay.zap.store do not yet conform to this spec but will be migrated shortly.

franzaps:
Updated the NIP to reflect the usage on Zapstore.

Since it has been used for months now and it's pretty stable, can we get this merged?

staab:
Is this in use by anyone other than zap.store?

franzaps:
@staab Not that I'm aware of.

Other than, maybe, @Giszmo ?

staab:
Normally NIPs are merged once there's some compatibility that needs to be ensured. It's good to have it documented in a draft, but update here when there are other apps using the standard

franzaps:
This data structure will probably end up in [NDK](https://github.com/relaystr/ndk) and used in [Yana](https://github.com/frnandu/yana) and probably in a relay-based community started by @NielLiesmons

vitorpamplona:
This should be merged.

staab:
Are there 2+ clients?

vitorpamplona:
I am putting in on Amethyst right now. We can wait for the next release if we don't have any other client using this yet.

DanConwayDev:
I intend to with gitworkshop and ngit but haven't got around to it yet.

-------- Original Message --------

staab:
> I am putting in on Amethyst right now.

That's good enough for me. Are you implementing as is?

vitorpamplona:
Just to display right now. I am not sure it makes sense to create it yet.

franzaps:
@kehiy Absolutely, I am using it already for the Zapstore package manager (works on MacOS and Linux): https://zapstore.dev/docs/cli/manage/ -- it will get way more packages in the coming months

v0l:
There is no reason for the release sets to be replaceable, can that be changed now?

v0l:
Built a cli tool in rust to do publishing, no dependencies, no shell commands, only works for `APK` atm: https://git.v0l.io/Kieran/nap

franzaps:
> There is no reason for the release sets to be replaceable, can that be changed now?

The first question would be what do we gain by changing it

v0l:
> > There is no reason for the release sets to be replaceable, can that be changed now?
> 
> The first question would be what do we gain by changing it

Cant be rugged, small benefit

franzaps:
> Break things while you can. If there is no need for a replaceable event, let's get rid of it and clean up the spec.
> 
> Also, does `30063` only keep the latest version? Or is the d-tag used as versions?

I keep events of kind 30063 around for every version. I only use the latest one in the UI (for now). Yes the d-tag is used for identifier and version (`com.greenart7c3.nostrsigner@v2.0.1`)

I recall asking developers to use `--overwrite-release` in zapstore-cli multiple times over the past few months (it replaces the kind 30063). It happened for different reasons; the point is they make mistakes and need it.

I know what will happen if we make this a regular event: they will end up asking me to delete the bad event from the relay so they don't have to bump and re-release

vitorpamplona:
Can you add more guidance to the `30063`? Right now it is just a line in the spec. This d-tag for instance should be defined so that other stores can use the same.

franzaps:
> Can you add more guidance to the `30063`? Right now it is just a line in the spec. This d-tag for instance should be defined so that other stores can use the same.

Actually I think it's there simply for convenience because `version` should be picked up from the linked `e` 1063 event and the identifier (`d`-tag) from the linked `a` 32267 event. So other stores could use another string (if I was doing that though, I would observe 30063 events in the wild first)

vitorpamplona:
It would be nice to have a common way of doing. I could imagine switching back one version and having to "search" for what each store/publisher is using as dtag.

v0l:
Also missing some documentation about the NIP-94, i events as per this PR but it doesnt open in Zapstore app,

What is `apk_signature_hash` for?

franzaps:
> Also missing some documentation about the NIP-94, i published events as per this PR but it doesnt open in Zapstore app,
> 
> What is `apk_signature_hash` for?

Should be `apk_certificate_hash` really (used in Zapstore to detect cert mismatches early which for sure will result in failed package updates). None of the NIP-94 extensions are documented. I don't know whether it should be another NIP or what.

vitorpamplona:
Pictures and Videos were on NIP-94 and recently moved away from it into their own event kinds. That seems to have been a good move. 

The same can happen here if software applications need their own NIP-94-like kind where not only specialized tags (like `apk_certificate_hash`) can be defined just for these file types but also the procedures to validate those hashes can be discussed.

ZigBalthazar:
+1 relay implementation:
https://github.com/dezh-tech/ddsr/tree/main/zapoli

dluvian (commented):
Software releases being replaceable have multiple downsides:
- Breaks chronological order: **Impossible** to figure out which is the latest release because the latest timestamp may be an edit. Querying 10 releases will not yield the 10 latest releases when an old release has been edited. 
- Can't differentiate between initial release and update: A Github-style activity feed may only want to display new releases and not their edits.
- Minor downside of getting rugged by edits.

I would prefer a non-replaceable release event and a replaceable edit event for fixing bad releases. I we keep the replaceability then edits should add an initial-time timestamp tag to keep the `created_at` of the initial release event. This tag aids in ordering releases and differentiating edit from initial release.

franzaps:
> Software releases being replaceable have multiple downsides:
> 
>     * Breaks chronological order: **Impossible** to figure out which is the latest release because the latest timestamp may be an edit. Querying 10 releases will not yield the 10 latest releases when an old release has been edited.

I understand your point, however, I have never seen this happen in practice. Replaceable releases (kind 30063) have been live on Zapstore for over a year now.

When developers figure out mistakes and need to edit a release, it's very likely they do *before* cutting a newer release.

Creating events in the past to edit an old release is possible, easy and enforceable by publishing tools like zapstore CLI - which again, in practice, most developers use. So in my experience this is a non-issue.

>     * Can't differentiate between initial release and update: A Github-style activity feed may only want to display new releases and not their edits.

I'm fine with adding an `original_created_at` though I don't see how it would differ from `created_at` in value. Having it could be useful to show an "edited" label.

>     * Minor downside of getting rugged by edits.

Not sure what this means exactly. In Zapstore for example zaps (and eventually comments and other) refer to the asset, which is a regular non-replaceable event, not the release. In general if you think a developer is rugging you stop trusting them.

dluvian:
Releases are replaceable to fix typos and broken artifacts but you're only supposed to fix the latest release or otherwise you will break the release order. Or you use specialized software that hack around this design flaw.

> I have never seen this happen in practice

Yeah, but to be fair there are not many people that are publishing on nostr right now.

What about using 30063 as the edit event for a non-replaceable release event? Both with same structure described in your proposal. So we can have sane ordering with editable releases.

franzaps:
> Releases are replaceable to fix typos and broken artifacts but you're only supposed to fix the latest release or otherwise you will break the release order. Or you use specialized software that hack around this design flaw.

Does a similar mechanism exist for long-form articles (kind 30023)? Or any other replaceable for that matter? Why are we singling this one out?

If you are a developer who modifies a 6-month old release (unlikely but possible) to presumably improve correctness of your event, why wouldn't you also take care of the correctness of the `created_at` date? It's not like you are forced to timestamp it with the current date.

An app that wishes to show latest releases could also query and sort client side by `version` (or even `version_code` if Android).

Also, if you really wanted to rely on non-replaceables, with a little creativity you can work around it. Query for the latest assets (kind 3063) for a specific platform (`f` tag) and you will get non-replaceable events that have a `version` tag in them. Sort accordingly.

> What about using 30063 as the edit event for a non-replaceable release event? Both with same structure described in your proposal. So we can have sane ordering with editable releases.

While I agree your proposal is technically more correct, there is a tradeoff of one additional query **per release** just to check if there were any edits. For something low-risk and for which there are more than decent alternatives as I laid out. Tradeoffs in favor of leaving it unchanged in my opinion.

dluvian:
> Why are we singling this one out?

Because it's much more important to figure out the latest release than the latest blog post.

> While I agree your proposal is technically more correct, there is a tradeoff of one additional query per release just to check if there were any edits. For something low-risk and for which there are more than decent alternatives as I laid out. Tradeoffs in favor of leaving it unchanged in my opinion.

That's fair 🤝

51.md
```diff
@@ -58,7 +58,6 @@ Aside from their main identifier, the `"d"` tag, sets can optionally have a `"ti
 | Kind mute sets        | 30007 | mute pubkeys by kinds<br>`"d"` tag MUST be the kind string                                   | `"p"` (pubkeys)                                                                     |
 | Interest sets         | 30015 | interest topics represented by a bunch of "hashtags"                                         | `"t"` (hashtags)                                                                    |
 | Emoji sets            | 30030 | categorized emoji groups                                                                     | `"emoji"` (see [NIP-30](30.md))                                                     |
-| Release artifact sets | 30063 | group of artifacts of a software release                                                     | `"e"` (kind:1063 [file metadata](94.md) events), `"a"` (software application event) |
```
dluvian, line 62: Is this used by anyone?
30063 is already reserved in rust-nostr as `Kind::ReleaseArtifactSet` @yukibtc

franzaps:
> > Why are we singling this one out?
> 
> Because it's much more important to figure out the latest release than the latest blog post.
> 
> > While I agree your proposal is technically more correct, there is a tradeoff of one additional query per release just to check if there were any edits. For something low-risk and for which there are more than decent alternatives as I laid out. Tradeoffs in favor of leaving it unchanged in my opinion.
> 
> That's fair 🤝

I appreciate your input.

51.md
franzaps: You remind me that we should update NIP-51 with changes here (mention that kind 30063 is covered here, and that it should point to kind 3063 events)

dluvian:
I'm already using `Software Release kind 30063` to show new software releases in the activity feed in [Gitplaza](https://codeberg.org/dluvian/gitplaza) and I would like to support listing and creating repo releases with the same kind. Can you add an optional `a` tag to the documentation so we can point releases to a repo? @franzaps

NielLiesmons:
Very cool!

Yeah, both a Repo event and a Docs event (I need a place to link to the NIPs and other specs, tools, ... that were used in the App) would be useful. 

Don't know at what level is best, Release I guess.

franzaps:
> I'm already using `Software Release kind 30063` to show new software releases in the activity feed in [Gitplaza](https://codeberg.org/dluvian/gitplaza) and I would like to support listing and creating repo releases with the same kind. Can you add an optional `a` tag to the documentation so we can point releases to a repo? @franzaps

@dluvian Very cool!

The tag is available in the application kind 32267 (not in 30063):

```jsonc
    [
      // OPTIONAL
      // Pointer to NIP-34 repository
      "a", "30617:<destination-pubkey>:<repo-id>", "<relay-url>"
    ],
```

FYI it will take 6+ weeks for Zapstore to start the migration to this new format. There are currently too many signers and we need to do this carefully.

dluvian:
No `title` tag for releases?

franzaps:
> No `title` tag for releases?

Sure can add that

82.md
```diff
@@ -0,0 +1,316 @@
...
+      // REQUIRED, MULTIPLE
+      // Platforms
+      // NOTE: See Appendix A
+      "f", "<platform-id>"
+    ],
+    [
+      // OPTIONAL
+      // Minimum operating system version (minimum compatibility)
+      // NOTE: On Android this would be obtained from min_sdk_version
+      "min_os_version", "<min-os-version>"
```
basantagoswami: Wouldn't it have made sense to call it something like `min_platform_version` instead of `min_os_version`. Same for the target tag. Some applications can be things like browser extensions. Or mini apps within the new Damus Notedeck for example. It's probably too late to change it now but considering we already have the f tag for platforms, I thought it makes sense to call it platform instead of os

franzaps: Yes good call

82.md
```diff
@@ -0,0 +1,316 @@
...
+    ],
+    [
+      // REQUIRED
+      // See Appendix C for channel IDs
+      "c", "<channel-id>"
+    ],
+    [
+      // REQUIRED, MULTIPLE
+      // Asset ID
+      "e", "<asset-id>"
```
dluvian: I'm having trouble finding the assets, because there is no pubkey hint for a relay list look up. Can you add it?

basantagoswami:
@franzaps did you decide to not migrate to kind 3063 from 1063? I don't see any 3063 events anywhere

82.md
franzaps: Sure we can add relay hints, but what trouble exactly? How did you even get to the asset IDs?

franzaps:
> @franzaps did you decide to not migrate to kind 3063 from 1063? I don't see any 3063 events anywhere

I did decide; not an easy thing when too many people use a standard. I will be taking care of this because I need it for the new app but the migration will take a bit

82.md
dluvian: Your zapstore npub publishes releases with e-tags. These are asset IDs, right?

franzaps:
Just pushed an update @dluvian @basantagoswami 

- Add title to releases
- Add relay hints to asset IDs
- Use url for release URL
- Use "platform" instead of "operating system"
- Rename to apk_certificate_hash

Migration will start just after the next release with this version of the NIP

franzaps:
Updates:
  - Removed release `title` as it's redundant. Use a header in the release notes markdown if you want one
  - Removed release `url` as it's redundant. Pull release notes from the URL and copy to content field
  - Moved `commit` to SoftwareAsset since it needs to be next to the hash (useful for reproducible builds)
  - Introduce `min_allowed_version` used for security-focused version pinning: installer application can enforce disallowing installation of older versions of same asset

alltheseas:
### suggested backwards compatible language extension 

For Fran review:

```
 Convention: d tag with @lang suffix                                                                                                                                                                            
                                                                                                                                                                                                                 
  // Canonical app (default language)                                                                                                                                                                            
  {                                                                                                                                                                                                              
    "kind": 32267,                                                                                                                                                                                               
    "tags": [                                                                                                                                                                                                    
      ["d", "com.example.app"],           // No suffix = canonical                                                                                                                                               
      ["name", "Calculator"],                                                                                                                                                                                    
      ["L", "ISO-639-1"],                                                                                                                                                                                        
      ["l", "en", "ISO-639-1"]            // NIP-32 language label                                                                                                                                               
    ],                                                                                                                                                                                                           
    "content": "A fast calculator..."                                                                                                                                                                            
  }                                                                                                                                                                                                              
                                                                                                                                                                                                                 
  // Spanish translation (same or different publisher)                                                                                                                                                           
  {                                                                                                                                                                                                              
    "kind": 32267,                                                                                                                                                                                               
    "tags": [                                                                                                                                                                                                    
      ["d", "com.example.app@es"],        // @lang suffix = translation                                                                                                                                          
      ["name", "Calculadora"],                                                                                                                                                                                   
      ["L", "ISO-639-1"],                                                                                                                                                                                        
      ["l", "es", "ISO-639-1"],                                                                                                                                                                                  
      ["a", "32267:<pubkey>:com.example.app", "<relay-hint>"]  // Points to canonical                                                                                                                            
    ],                                                                                                                                                                                                           
    "content": "Una calculadora rápida..."                                                                                                                                                                       
  }                                                                                                                                                                                                              
                         
```
### Context & Reasoning                                                                                                                                                                                        
  

```                                                                                                                                                                                            
  Why This Is The Best Path                                                                                                                                                                                      
  ┌─────────────────────────────┬────────────────────────────────────────────────────────────────────────────────────────┐                                                                                       
  │          Criterion          │                                      How It's Met                                      │                                                                                       
  ├─────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────┤                                                                                       
  │ Original NIP spirit         │ Each Kind 32267 = one app description. No new kinds.                                   │                                                                                       
  ├─────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────┤                                                                                       
  │ Future-proof                │ @lang scales to 1 or 500 languages. Add more anytime.                                  │                                                                                       
  ├─────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────┤                                                                                       
  │ Backwards compatible        │ Old clients see "separate apps" (graceful degradation). New clients group by d prefix. │                                                                                       
  ├─────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────┤                                                                                       
  │ Uses existing NIPs          │ NIP-32 (l tag), NIP-01 (a tag for reference)                                           │                                                                                       
  ├─────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────┤                                                                                       
  │ Same publisher multi-lang   │ ✓ Different d tag = different event identity                                           │                                                                                       
  ├─────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────┤                                                                                       
  │ Community translations      │ ✓ Different pubkey with same d@lang pattern                                            │                                                                                       
  ├─────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────┤                                                                                       
  │ No NIP changes needed later │ Convention is extensible without protocol changes                                      │                                                                                       
  └─────────────────────────────┴────────────────────────────────────────────────────────────────────────────────────────┘                                                                                       
  ---                                                                                                                                                                                                            
  Spec Addition (add to NIP-82)                                                                                                                                                                                  
                                                                                                                                                                                                                 
  ## Localized Application Metadata                                                                                                                                                                              
                                                                                                                                                                                                                 
  Applications MAY provide metadata in multiple languages using the `d` tag suffix convention:                                                                                                                   
                                                                                                                                                                                                                 
  - `["d", "<app-id>"]` - Canonical application (default/primary language)                                                                                                                                       
  - `["d", "<app-id>@<lang>"]` - Localized variant where `<lang>` is an ISO 639-1 code                                                                                                                           
                                                                                                                                                                                                                 
  Localized events SHOULD include:                                                                                                                                                                               
  - `["L", "ISO-639-1"]` and `["l", "<lang>", "ISO-639-1"]` tags per NIP-32                                                                                                                                      
  - `["a", "32267:<canonical-pubkey>:<app-id>", "<relay-hint>"]` referencing the canonical app                                                                                                                   
                                                                                                                                                                                                                 
  Clients SHOULD:                                                                                                                                                                                                
  - Query with `#l` filter to fetch user's preferred language                                                                                                                                                    
  - Group events by `d` tag prefix (before `@`) when displaying apps                                                                                                                                             
  - Prefer canonical (`d` without `@`) for non-localized fields (icon, releases)                                                                                                                                 
                                                                                                                                                                                                                 
  Example query for Spanish apps:                                                                                                                                                                                
  ```json                                                                                                                                                                                                        
  {"kinds": [32267], "#l": ["es"]}                                                                                                                                                                               
                                                                                                                                                                                                                 
  ---                                                                                                                                                                                                            
                                                                                                                                                                                                                 
  ### How It Works In Practice                                                                                                                                                                                   
                                                                                                                                                                                                                 
  **Publisher workflow (50 languages):**                                                                                                                                                                         
  com.example.app       → English (canonical)                                                                                                                                                                    
  com.example.app@es    → Spanish                                                                                                                                                                                
  com.example.app@zh    → Chinese                                                                                                                                                                                
  com.example.app@ar    → Arabic                                                                                                                                                                                 
  ... (47 more)                                                                                                                                                                                                  
                                                                                                                                                                                                                 
  **Client query:**                                                                                                                                                                                              
  ```jsonc                                                                                                                                                                                                       
  // User's locale is Spanish                                                                                                                                                                                    
  {"kinds": [32267], "#d": ["com.example.app", "com.example.app@es"], "#l": ["es", "en"]}                                                                                                                        
                                                                                                                                                                                                                 
  // Or discover all translations of an app                                                                                                                                                                      
  {"kinds": [32267], "#d": ["com.example.app", "com.example.app@*"]}  // If relay supports prefix matching                                                                                                       
                                                                                                                                                                                                                 
  // Or simpler: fetch by app-id prefix pattern in client                                                                                                                                                        
                                                                                                                                                                                                                 
  Client merge logic:                                                                                                                                                                                            
  1. Fetch canonical: com.example.app                                                                                                                                                                            
  2. Fetch translation: com.example.app@{userLang}                                                                                                                                                               
  3. Use translation's name/summary/content                                                                                                                                                                      
  4. Use canonical's icon/images/releases (non-localized fields)                                                                                                                                                 
                                                                                                            

What Gets Added to NIP-82                                                                                                                                                                                      
                                                                                                                                                                                                                 
  1. @lang suffix convention for d tag (~5 lines)                                                                                                                                                                
  2. Recommend NIP-32 labels for language (~3 lines)                                                                                                                                                             
  3. Recommend a tag to canonical (~3 lines)                                                                                                                                                                     
  4. Client behavior for grouping/querying (~10 lines)                                                                                                                                                           
                                                                                                                                                                                                                 
  Total: ~20 lines added to spec. No new kinds. No breaking changes. Scales forever.                                                                                                                             
                                                                                                                                                                                                                 
  ---                                                                                                                                                                                                            
  Optional Extensions (add later without NIP changes)                                                                                                                                                            
  ┌──────────────────────────┬────────────────────────────────────────────────────────┬─────────────────────────┐                                                                                                
  │        Extension         │                          How                           │          When           │                                                                                                
  ├──────────────────────────┼────────────────────────────────────────────────────────┼─────────────────────────┤                                                                                                
  │ Machine translation flag │ ["l", "es", "ISO-639-1", "mt"] (4th element)           │ If MT becomes common    │                                                                                                
  ├──────────────────────────┼────────────────────────────────────────────────────────┼─────────────────────────┤                                                                                                
  │ Translation quality      │ NIP-32 label: ["l", "verified", "translation-quality"] │ If quality tiers needed │                                                                                                
  ├──────────────────────────┼────────────────────────────────────────────────────────┼─────────────────────────┤                                                                                                
  │ Translator attribution   │ ["p", "<translator-pubkey>"]                           │ Already supported       │                                                                                                
  └──────────────────────────┴────────────────────────────────────────────────────────┴─────────────────────────┘      
```

from https://github.com/zapstore/zapstore/issues/298#issuecomment-3835591690

alltheseas:
> Is this in use by anyone other than zap.store?

notedeck implemented for auto-update use case (https://github.com/damus-io/notedeck/pull/1326)

tracker is here: https://github.com/nostrability/nostrability/issues/277

alltheseas:
ditto also added supported today for:

1. rendering kind 32267 app events in-feed
2. fetching releases kind 30063 from ZS relay

https://github.com/nostrability/nostrability/issues/277

cc @alexgleason

alltheseas:
Today learned spec now lives here

https://nostrhub.io/naddr1qvzqqqrcvypzq7xwd748yfjrsu5yuerm56fcn9tntmyv04w95etn0e23xrczvvraqy2hwumn8ghj7un9d3shjtnyd968gmewwp6kyqqz8qeqp2qk5p

DhananjayPurohit:
Great to see `application/vnd.oci.image.manifest.v1+json` in Appendix C. For OCI support to be implementable, a few things need to be defined:

1. **Multi-arch images.** Suggestion: one kind 3063 asset per platform manifest — `m` = the OCI manifest MIME type, `x` = the manifest digest (an OCI digest is the sha256 of the manifest JSON, so it goes into `x` unchanged), and `f` = the platform. A multi-arch image is then N asset events under one release. This works with the existing aggregate-`f` rule as is, and the image index MIME type doesn't need to be added to Appendix C — the release event already serves as the index.

2. **Layer blobs.** A container image is a manifest plus layer blobs, but layers don't need their own events — the manifest already lists their digests. Suggestion: add a line saying clients fetch layer blobs by sha256 from the publisher's Blossom servers (BUD-03 server list), and may also try the host from the asset's `url` tag.

3. **Mutable tags.** Docker-style tags map directly onto Appendix B channels: `myapp:latest` resolves to the highest-version release on channel `main` (using Appendix D version ordering — not the most recent `created_at`, because republishing an old release to repair it refreshes its timestamp, and a recency rule would then wrongly resolve `latest` to that old version). `myapp:1.2.0` resolves to `d = myapp@1.2.0`. Pulls by digest (`@sha256:...`) don't need events at all. One sentence in the spec adopting this mapping would ensure every implementation resolves a given tag to the same release.

Happy to draft the OCI section text for these.

PR #1337, opened 2024-06-29
[NIP-96] Adding optional file_tags to categorize files and listing public files

quentintaranpino:
#### Objective
The objective of this proposal is to enhance the current file handling capabilities by introducing optional query parameters for categorizing and listing **public files**. This enhancement will provide users with more flexibility and control when interacting with files on the server.

#### Proposed Changes

`GET $api_url?page=x&count=y&sort=z&order=w&search=term`

1. **Adding Optional Query Parameters:**

    - `sort` : Determines the sorting criteria of the listed files. It can take values `'created_at'` or `'size'`. The default value is `'created_at'`.
    - `order` : Specifies the order of the listing, either `'ASC'` for ascending or `'DESC'` for descending. The default value is `'DESC'`.
    - `search` : Allows users to search for files based on a search term. The server should search in the `alt` and `file_tags` fields.

2. **Adding Optional `file_tags` Field When Uploading Files:**

    - The `file_tags` field will be an optional field during file upload, allowing users to specify tags for their files. Example usage: `["file_tags", "cat, meme, funny, laugh"]`.

2. **Authorization Requirements:**

    - AUTH **required** for **pubkey** file listing.
    - AUTH **NOT required** for **public** file listing.

#### Benefits

The introduction of these optional fields and parameters will enable more sophisticated file management and search capabilities. Specifically, it will allow clients to:

- **Categorize Files**: By adding `file_tags` during file upload, users can categorize their files more effectively, making it easier to organize and retrieve them later.
- **Enhanced Search**: The `search` parameter will allow users to search files using specific terms, filtering results based on the `alt` and `file_tags` fields. This will make it easier to find specific files, such as gifs of memes, quickly and efficiently.
- **Flexible Listing**: The `sort` and `order` parameters will provide users with the ability to customize the display order of their files, either by creation date or size, in ascending or descending order.

This is inspired by this note from @jb55, where he shares hisconcerns with the GIPHY service. With this small modification, NIP96 will allow media servers to provide this service, easily and with minimal modifications. 

https://nostrudel.ninja/#/n/nevent1qy88wumn8ghj7mn0wvhxcmmv9uq35amnwvaz7tms09exzmtfvshxv6tpw34xze3wvdhk6tcpzemhxue69uhhyetvv9ujumt0wd68ytnsw43z7qghwaehxw309aex2mrp0yh8qunfd4skctnwv46z7qg3waehxw309ahx7um5wghxcctwvshsz9thwden5te0wfjkccte9ejxzmt4wvhxjme0qyghwumn8ghj7mn0wd68ytnhd9hx2tcqyq4c095djx50q4mtz8xhzhmyhgyn3ezd6jclpg2l5hurnvy9j9wky647wzj

What do you think? @arthurfranca @v0l @vitorpamplona

### DIFF
--- a/96.md
+++ b/96.md
@@ -282,9 +282,10 @@ The successful response is a 200 OK one with just basic JSON fields:
 
 ## Listing files
 
-`GET $api_url?page=x&count=y&sort=z&search=w`
+`GET $api_url?page=x&count=y&sort=z&order=w&search=term`
 
 AUTH **required** for pubkey file listing.
+
 AUTH **NOT** required for public file listing.
 
 Returns a list of files. If auth event is present, the server should return only files uploaded by the pubkey in the auth event.
@@ -304,7 +305,7 @@ Example Response:
         ["size", "123456"],
         ["alt", "a meme that makes you laugh"],
         ["expiration",  "1715691139"],
-         ["file_tags", "cat, meme, funny, laugh"],
+        ["file_tags", "cat, meme, funny, laugh"],
         // ...other metadata
       ]
       "content": "haha funny meme", // caption
@@ -319,8 +320,13 @@ Example Response:
 
 ### Query args
 
-- `page` page number (`offset=page*count`)
-- `count` number of items per page
+- `page`   **REQUIRED** Page number (`offset=page*count`)
+- `count`  **REQUIRED** Number of items per page
+- `sort`   **OPTIONAL** Can be 'created_at' or 'size'. Default is 'created_at'.
+- `order`  **OPTIONAL** Page order 'ASC' or 'DESC'. Default is 'DESC'.
+- `search` **OPTIONAL** Search term. The server should search in the `alt` and `file_tags` fields.
+
+The server can or not support optional query args.
 
 ## Selecting a Server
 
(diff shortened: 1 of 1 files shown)


quentintaranpino:
@fishcakeday

jb55:
fwiw I wouldn't use this because it centralizes too much functionality on external providers imo. my idea was just to use a labelling mechanism on the network itself, so that anyone or AI could source and label gifs on the network without any external service. then it would just be a matter of following labeling service(s) that you can query these labels from, since you wouldn't want to trust labels from anyone.

for example, imagine a query such as:

```json
{"kinds": [12345], "#L": ["lol"], "authors": ["label-service-pubkey", "friend-pubkey", "etc..."], "limit": 10 }
```

which returns notes of gifs related to `lol`. The only tricky part would be gathering these labels so that you can query them efficiently, aka converting search queries to labels. or maybe clients could gather these labels and present them as possible things to search.

still someone of a half baked idea but I don't see why it couldn't work.

jb55:
hmm you could even query gifs directly from nip-94! if services started labeling those. you could even search nip-94 contents since those are a description of the file? @vitorpamplona

PR #1340, opened 2024-07-01
NIP 51: Adds Kind 10011 Profile Gallery

believethehype:
A Profile Gallery is predefined List containing "G" tags. ["G", eventid, urltomediafile]. 
A User might add media (images, videos etc) and the eventid where the media appears to the list. Clients can render these for example as galleries in the User's profile and reference to the original Note. 
Capital G as g is taken for geotags

### DIFF
--- a/51.md
+++ b/51.md
@@ -29,7 +29,8 @@ For example, _mute list_ can contain the public keys of spammers and bad actors 
 | Public chats      | 10005 | [NIP-28](28.md) chat channels the user is in                | `"e"` (kind:40 channel definitions)                                               |
 | Blocked relays    | 10006 | relays clients should never connect to                      | `"relay"` (relay URLs)                                                            |
 | Search relays     | 10007 | relays clients should use when performing search queries    | `"relay"` (relay URLs)                                                            |
-| Simple groups     | 10009 | [NIP-29](29.md) groups the user is in                       | `"group"` ([NIP-29](29.md) group ids + mandatory relay URL)                       |
+| Simple groups     | 10009 | [NIP-29](29.md) groups the user is in                       | `"group"` ([NIP-29](29.md) group ids + mandatory relay URL)     
+| Profile Gallery   | 10011 | Profile Media Gallery                                       | `"g"` (gallery entry), (kind:1 notes), (url to media file)|
 | Interests         | 10015 | topics a user may be interested in and pointers             | `"t"` (hashtags) and `"a"` (kind:30015 interest set)                              |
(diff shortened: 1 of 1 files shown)


fiatjaf (approved):

fiatjaf:
(please squash before merging)

believethehype:
I'm not exactly sure which tag to use. I first wanted to introduce a g tag for gallery but that is taken for geotags. I now chose r but I think r is for a direct link to an url. I need something followed by 2 entries, an eventid and an url.

flox1an:
> I'm not exactly sure which tag to use. I first wanted to introduce a g tag for gallery but that is taken for geotags. I now chose r but I think r is for a direct link to an url. I need something followed by 2 entries, an eventid and an url.
> 
> Would some tag like ["gallery", "eventid", "url"] be fitting?
Why not use two different tags for events vs. urls, like `10003` the bookmark list? `"e"` (kind:1 notes), `"r"` (URLs)

believethehype:
I think e and r tags are alternatives in 10003 while here we need both

flox1an:
Ok, now I understand. I'm not quite sure with `gallery` as a tag since those entries point to an event or an image and not to a gallery. So IMHO something like `item`, `image`, `entry` would be more fitting as a tag.

believethehype:
gallery as in gallery entry, but this also might be confusing.

I'm fine with and can adapt to anything the symbolises a pair of eventid and media url.

AsaiToshiya:
It looks like generic pinned list to me.

believethehype:
How about a G tag (large G) ["G", eventId, url]? I like it.

mroxso:
I implemented it for testing purposes on https://beta.lumina.rocks/
Example Profile (dontbelievethehype): https://beta.lumina.rocks/profile/npub1nxa4tywfz9nqp7z9zp7nr7d4nchhclsf58lcqt5y782rmf2hefjquaa6q8
SourceCode (commit) is here: https://github.com/mroxso/lumina-rocks-website/commit/dd8fd14daef89a8ac5885d3bd5048d1b6ab59a7f
Feel free to give some feedback

mroxso:
Am I correct that it's possible to link an image to a random other event/note?

believethehype:
Yes. You could in theory show another Image than the note you're linking to. 

Default Case would be the Image actually is in the note and clients handle it like this.

pablof7z:
does this really need to be indexed? do I understand correctly that the event-id is kinda as a way to "attribute" where the image came from?

in that case, why not just just do something like `[ "url", "<url-of-the-media>", "event-id", "relay" ]` -- it sounds like the default case, the most interesting bit, will be the URL, and the event-id/relay pairs are, for the most part, nice-to-haves.

believethehype:
I'm closing this for now after talking with Pablo. I will submit an alternative approach without lists soon.

PR #1341, opened 2024-07-02
fix some info of nip55 to be same as other nips

haorendashu:
# fix some info of nip55 to be same as other nips

1. Change some fields ```pubKey``` to ```pubkey```, just other nips does.
2. When transmitting information, other nips use pubkey, but here npub is used, and npub is changed to pubkey.
3. Add a function ```get_relays``` just like other nostr signer [NIP-07](https://github.com/nostr-protocol/nips/blob/master/07.md) and [NIP-46](https://github.com/nostr-protocol/nips/blob/master/46.md)

You can see the detail from this [link](https://github.com/haorendashu/nips/blob/nip55-fix/55.md) .

### DIFF
--- a/55.md
+++ b/55.md
@@ -99,10 +99,10 @@ launcher.launch(intent)
     context.startActivity(intent)
     ```
   - result:
-    - If the user approved intent it will return the **npub** in the signature field
+    - If the user approved intent it will return the **pubkey** in the signature field
 
       ```kotlin
-      val npub = intent.data?.getStringExtra("signature")
+      val pubkey = intent.data?.getStringExtra("signature")
       // The package name of the signer application
       val packageName = intent.data?.getStringExtra("package")
       ```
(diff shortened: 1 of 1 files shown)


vitorpamplona:
@greenart7c3

fiatjaf (approved):

greenart7c3 (approved):

greenart7c3:
Changing the return value to hex key breaks some applications like Amethyst

https://github.com/vitorpamplona/amethyst/blob/33b8267de7e2fea6f1c416a52d283222f06eed08/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/screen/AccountStateViewModel.kt#L106

haorendashu:
@vitorpamplona Hello, Vitor. Do you have plan to change Amethyst to suit Amber returning hex pubkey not npub ?

greenart7c3:
I'll make a pull request later for amethyst and test other applications but in my opinion this can be merged

fiatjaf (approved):

vitorpamplona:
@greenart7c3 I migrate the `current_user` from npub to hexkey. That seems to be the last npub used in the external signer communication. Can you check if Amber can receive hex keys as `current_user` today?

PR #1345, opened 2024-07-04
Update language around kind 11 and kind 12 removing microblog and rep…

ismyhc:
After discussing with @fiatjaf it was thought that replacing the term microblog with forum was more indicative of what is meant for kind 11 and kind 12.

Ps. I named my branch wrong, but shouldn't be an issue. This update is around kind 11 and kind 12.

@pablof7z

### DIFF
--- a/29.md
+++ b/29.md
@@ -40,7 +40,7 @@ Relays should prevent late publication (messages published now with a timestamp 
 
 - *text root note* (`kind:11`)
 
-This is the basic unit of a "microblog" root text note sent to a group.
+This is the basic unit of a "forum" root text note sent to a group. These should likely be seperated from `kind:9` & `kind:10` events as mixing the two could be quite confusing. 
 
 ```js
   "kind": 11,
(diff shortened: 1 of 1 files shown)


29.md
```diff
@@ -40,7 +40,7 @@ Relays should prevent late publication (messages published now with a timestamp
 
 - *text root note* (`kind:11`)
 
-This is the basic unit of a "microblog" root text note sent to a group.
+This is the basic unit of a "forum" root text note sent to a group. These should likely be seperated from `kind:9` & `kind:10` events as mixing the two could be quite confusing.
```
fiatjaf, line 43: What is `kind:10`?

ismyhc: Chat message threaded reply?

ismyhc: kinds 9,10 dealing with chat messages and kinds 11,12 dealing with forum messages? no?

fiatjaf: Yes, sorry, I missed that in the text.

fiatjaf (approved):

29.md
```diff
@@ -40,7 +40,7 @@ Relays should prevent late publication (messages published now with a timestamp
 
 - *text root note* (`kind:11`)
 
-This is the basic unit of a "microblog" root text note sent to a group.
```
pablof7z, line 43: not sure about this suggestion of not mixing `kind:{9.10}` with `kind:{11,12}` -- would be normal to have a chat (kind 9/10) about a post within the group so it wouldn't be that weird 🤷‍♂️

ismyhc: Yeah potentially. I guess it's hard to tell without putting into action. I'm okay with removing that suggestion. Feel free to update it if you'd like. The important part was changing from microblog to forum.

PR #1346, opened 2024-07-05
NIP-118, Polls on Nostr 

abhay-raizada:
This NIP takes the ideas evolved from 
- https://github.com/nostr-protocol/nips/pull/1190
- https://github.com/nostr-protocol/nips/pull/1324
- https://github.com/vitorpamplona/amethyst/issues/935
- https://github.com/nostr-protocol/nips/pull/320
- https://github.com/nostr-protocol/nips/pull/148

and tries to combine them in a spec for a  simple polls implementation. 

Working version: https://pollerama.fun/

cc: @staab @vitorpamplona @vicariousdrama @alexgleason @gocedoko  @fiatjaf

### DIFF
--- a/118.md
+++ b/118.md
@@ -4,33 +4,86 @@
 
 `draft` `optional`
 
-The purpose of this nip is to create a mechanism for polling on nostr.
-The NIP aims to have the following properties.
+This NIP defines the event scheme that describe Polls on nostr.
 
-1. Provide a method for gauging eligibility. Only eligible votes should be counted.
-2. Provide a method for maintaining anonymity. Participants should be able to not reveal their identities to everyone, yet their votes should be counted.
-3. Provide a mechanism for verifiability. Participants should be able to verify that their votes have been counted, under an eligibility criteria.
+The NIP makes the following tradeoffs.
+
+1. All votes are public.
+2. All responses are editable (latest by timestamp considered last response)
 
 ## Events
 
 ### Poll Content
 
-The poll event is defined as a `kind:1068` event. This event carries over the `name` , `description` and `field` tags as defined in [nip-101](https://github.com/nostr-protocol/nips/pull/1190/files).
-This event has no other tags and completely defines the poll content, which is uneditable.
+The poll event is defined as a `kind:1068` event. Major tags in the poll event are:
+
+- **label**: The accompanying text of the poll.
+- **option**: The option tags contain an OptionId(any alphanumeric) field, followed by an option label field.
+- **relays**: Set of relays that the poll is expecting respondents to respond on.
+- **settings**: A JSON stringified object that may hold poll level configs and client specific configs.
+
+Example Event
+
+```json
+{
+  "content": "",
+  "created_at": 1719888496,
+  "id": "9d1b6b9562e66f2ecf35eb0a3c2decc736c47fddb13d6fb8f87185a153ea3634",
+  "kind": 1068,
+  "pubkey": "dee45a23c4f1d93f3a2043650c5081e4ac14a778e0acbef03de3768e4f81ac7b",
+  "sig": "7fa93bf3c430eaef784b0dacc217d3cd5eff1c520e7ef5d961381bc0f014dde6286618048d924808e54d1be03f2f2c2f0f8b5c9c2082a4480caf45a565ca9797",
+  "tags": [
+    ["label", "Pineapple on Pizza"],
+    ["option", "qj518h583", "Yay"],
+    ["option", "gga6cdnqj", "Nay"],
+    ["relays", "<relay1 url>", "<relay 2 url>"],
+    ["settings", "{\"PollType\": \"singlechoice\" }"]
+  ]
+}
+```
 
 ### Responses
 
-The response event is the same as the one on [nip-101](https://github.com/nostr-protocol/nips/pull/1190/files)
+The response event is a `kind:1018` event. It contains an e tag referencing the poll event it is referencing, followed by one or more response tags.
 
-### Eligibilty
+- **response** : The tag contains "response" as it's first positional argument followed by the option Id selected.
 
-Eligibility to a vote is determined by another `kind:30180` event that can be created by anyone and is essentially a reference to a list of pubkeys, the author of the eligibility event is interested in. The eligibility event MUST contain an "e" tag to the poll event, and
-"list" tags which references either a `kind:3` (contact list) or a `kind:30000`(follow sets) containing a list of "p" tags.
+Example Response Event
 
-### Counting Results.
+```json
+{
+  "content": "",
+  "created_at": 1720097117,
+  "id": "60a005e32e9596c3f544a841a9bc4e46d3020ca3650d6a739c95c1568e33f6d8",
+  "kind": 1018,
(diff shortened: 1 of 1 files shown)


118.md
```diff
@@ -0,0 +1,89 @@
...
+The NIP makes the following tradeoffs.
+
+1. All votes are public.
+2. All responses are editable (latest by timestamp considered last response)
+
+## Events
+
+### Poll Content
+
+The poll event is defined as a `kind:1068` event. Major tags in the poll event are:
```
AsaiToshiya, line 18: I think defining the new kind is a good point.

118.md
```diff
@@ -0,0 +1,89 @@
...
+
+- **label**: The accompanying text of the poll.
```
AsaiToshiya, line 20: Why not use the `content`?

118.md
```diff
@@ -0,0 +1,89 @@
...
+- **option**: The option tags contain an OptionId(any alphanumeric) field, followed by an option label field.
```
AsaiToshiya, line 21: I don't understand why you need the OptionId. The label can be the ID.

118.md
```diff
@@ -0,0 +1,89 @@
...
+- **relays**: Set of relays that the poll is expecting respondents to respond on.
+- **settings**: A JSON stringified object that may hold poll level configs and client specific configs.
```
AsaiToshiya, line 23: Why the JSON? I haven't seen it as the value of the tag.

118.md
abhay-raizada: what if there are two labels with the same name? i think this is an easy way to get past that. I don't like the idea of tallying same labels together.

118.md
abhay-raizada: it makes it easier if we want to take the same polls format and encrypt it as a private poll, There are no private polls yet however, so i'm open to move this to a label

118.md
abhay-raizada: carried it over from forms where i use it for Ui settings like imageurl etc. Thought I might reuse the same config for the pollType field, but I think the PollType field in the tags also makes sense.

118.md
AsaiToshiya: I feel like a label with the same name but different meaning is not a practical option.

vitorpamplona: I agree with @AsaiToshiya. If clients are only showing the name and not the option ID, the same name should mean the same vote. Not having an ID removes the need to disambiguate names when they duplicate.

118.md
vitorpamplona: I would get rid of it. Add all the extra settings as tags.

118.md
AsaiToshiya: I prefer the `content` unless the specification for private polls is clear.

118.md
abhay-raizada: I disagree, the client is not doing anything in this special case.

Its the poll creator that is just having fun, maybe they want to see how the option order affects the poll results, which they should be allowed to do.

mikedilger: I agree with abhay-raizada.  I might make a poll that says:  Who will you vote for?  1) Trump, 2) Trump, 3) Trump
and it should collect the three answers separately.

118.md
```diff
@@ -0,0 +1,89 @@
...
+  "created_at": 1719888496,
+  "id": "9d1b6b9562e66f2ecf35eb0a3c2decc736c47fddb13d6fb8f87185a153ea3634",
+  "kind": 1068,
+  "pubkey": "dee45a23c4f1d93f3a2043650c5081e4ac14a778e0acbef03de3768e4f81ac7b",
+  "sig": "7fa93bf3c430eaef784b0dacc217d3cd5eff1c520e7ef5d961381bc0f014dde6286618048d924808e54d1be03f2f2c2f0f8b5c9c2082a4480caf45a565ca9797",
+  "tags": [
+    ["label", "Pineapple on Pizza"],
+    ["option", "qj518h583", "Yay"],
+    ["option", "gga6cdnqj", "Nay"],
+    ["relays", "<relay1 url>", "<relay 2 url>"],
```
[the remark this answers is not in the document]
staab: Yeah, variadic tags are usually bad

118.md
```diff
@@ -0,0 +1,89 @@
...
+}
+```
+
+### Poll Types
+
+Poll Types can be configured in the settings object of the poll, the setting dictates how multiple response tags are handled in the `kind:1018` event.
+
+- **PollType: singlechoice**: The first response tag is to be considered the actual response.
+- **PollType: multiplechoice**: The first response tag pointing to each id is considered the actual response, without considering the order.
+- **PollType: rankedchoice**: The first response tag pointing to each id is the considered the actual response, while also taking into account the order of the response tags.
```
staab, line 74: Instead of JSONified `settings`, settings should just be tags

118.md
```diff
@@ -0,0 +1,89 @@
...
+In case of multiple events for a pubkey, the event with the largest timestamp should be considered.
+
+### Curation
+
+The clients may configure fetching results by specific people. This can be achieved by creating `kind:30000` follow sets, and fetching results only from the follow set.
+Clients can also employ other curation algorithms, like Proof Of Work and Web of Trust scores for result curations.
+
+### Private Polls
+
+This nip does not cover private polling, Clients may follow [forms nip](https://github.com/nostr-protocol/nips/pull/1190) for private polling.
```
staab, line 89: I'd omit this section, it's not necessary and nips shouldn't refer to PRs

staab (approved):
This is the best poll nip so far. A few minor changes, but it's the right level of dumb simplicity.

alexgleason:
Does this have 2 implementations? I was just about to comment on it:

- Why is response not a parameterized replaceable event?
- Why not use the tag index for the vote instead of making up random letters?

Not using a replaceable event for the response is especially bad if the response is meant to be edited.

alexgleason:
I will actually implement polls. I have a UI for it and could do it in a couple hours. But counting votes is going to be such a pain because we don't use existing replaceable semantics.

fiatjaf:
> * Why is response not a parameterized replaceable event?
> * Why not use the tag index for the vote instead of making up random letters?

I asked exactly these two questions myself, but I decided it wasn't a big deal.

118.md
```diff
@@ -0,0 +1,92 @@
...
+```json
+{
+  "content": "",
+  "created_at": 1719888496,
+  "id": "9d1b6b9562e66f2ecf35eb0a3c2decc736c47fddb13d6fb8f87185a153ea3634",
+  "kind": 1068,
+  "pubkey": "dee45a23c4f1d93f3a2043650c5081e4ac14a778e0acbef03de3768e4f81ac7b",
+  "sig": "7fa93bf3c430eaef784b0dacc217d3cd5eff1c520e7ef5d961381bc0f014dde6286618048d924808e54d1be03f2f2c2f0f8b5c9c2082a4480caf45a565ca9797",
+  "tags": [
+    ["label", "Pineapple on Pizza"],
```
smolgrrr, line 41: implemented NIP-88 on getwired.app
super easy to implement and works well. don't really see the point of the label tag on the poll event, but other than that no complaints

abhay-raizada: it's an artifact from before the label was moved to the content field. Will update

PR #1347, opened 2024-07-05
Add nip136 code packages kind 1036

brugeman:
Proposing this as part of Nostr Sites effort. A code package event is basically npm package on nostr, where files are stored on Blossom or other sha256-addressed servers, and the package event lists those files along with their relative paths within the package. Being able to fetch individual files using http (instead of downloading the whole package as zip archive) is useful for Nostr sites where specific assets are included selectively into the web pages. NIP136 text [here](https://github.com/brugeman/nips/blob/nip/136/136.md).

### DIFF
--- a/136.md
+++ b/136.md
@@ -1,0 +1,108 @@
+Code packages
+=============
+
+This NIP handles distribution of code packages (akin to npm packages) in a decentralized way. 
+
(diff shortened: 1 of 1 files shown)


alexgleason:
See also: #1112

erskingardner:
This is the same as #1112 right? Guessing it's out of scope here as well?

brugeman:
> This is the same as #1112 right? Guessing it's out of scope here as well?

No this is not the same, it doesn't suggest relays to host anything.

All it does is describe a new event kind, it can't be out of scope.

PR #1348, opened 2024-07-05
Add nip512 nostr sites kind 30512

brugeman:
Nostr-based websites (Nostr Sites) are kind 30512 events. The NIP's goal is to describe web-specific attributes of a site and to augment published nostr events with web-specific attributes (meta tags, url etc). The NIP should also make switching "clients" possible (switching website rendering engines), without breaking the website metadata, structure, etc. An early draft, read full text [here](https://github.com/brugeman/nips/blob/nip/512/512.md).

### DIFF
--- a/512.md
+++ b/512.md
@@ -1,0 +1,173 @@
+Nostr Sites
+===========
+
+This NIP describes a way to render a selection of Nostr events on a website.
+
+A web page that is part of the Nostr Site must include `nostr:site` meta tag:
+```
+<meta property="nostr:site" content="<naddr>" /> 
+```
+
+The `site` event, referenced by `naddr` from `nostr:site` meta tag, is a parameterized replaceable event of kind `30512`. The `content` field MAY contain a rich description of the site in markdown syntax (NIP-23), all tags are optional (except `d`).
+
+Tags:
+```
+{
+  "tags":[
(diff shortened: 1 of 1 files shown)


AsaiToshiya:
Is there an implementation that can help to understand this?

nostrband:
> Is there an implementation that can help to understand this?

Yes the (still partial) implementation is npub.pro.

512.md
```diff
@@ -0,0 +1,173 @@
...
+To override default meta tags of a hashtag page, `hashtag` event may include meta info tags of the `site` event (`title`, `image`, `meta_title`, `og_title` etc). 
+
+If `r` tag starts with `/` then it is a static hashtag webpage, and this hashtag should not be listed under posts and in other hashtag lists.
+
+If several `hashtag` events published by contributors have the same `r` tag then the most recent one should be preferred.
+
+Themes and Plugins
+==================
+
+The `theme` event is parameterized replaceable event of kind `30514`, it's `content` field may have a "rich" description in the same syntax as NIP-23 long-form notes. It has the following tags:
```
RetricSu, line 125: is there an example content to help understanding how the theme works?

brugeman: Try fetching 30514 events from relay.npubpro.com, and also 1036 code package events linked by themes from the same relay.

PR #1349, opened 2024-07-05
Fix filter JSON comments to clarify that timestamp matches are inclusive

tyiu:
The current wording could either read ambiguously or mislead the reader to think it's an exclusive (not inclusive) time filter.


AsaiToshiya:
Related: #1284

tyiu:
> Related: #1284

Happy to close this PR in favor of that one.

PR #1354, opened 2024-07-10
Add optional event kind to encrypt/decrypt

greenart7c3:
This adds an optional event kind parameter to the encrypt/decrypt functions, allowing signers to show the user what the app is attempting to encrypt/decrypt.


vitorpamplona:
Is this because of the private zap stuff?

greenart7c3:
No, it's to improve how we show messages on signers
Right now if you send a decrypt message the only thing the signer can show the user is the encrypted text
This would allow to show something like "Amethyst would like to decrypt a mute list event"

vitorpamplona:
ohhh that's nice! lgtm!

greenart7c3:
Im closing this for now
There's some discussion on better approaches in this issue:
https://github.com/nostr-protocol/nips/issues/1439

PR #1355, opened 2024-07-10
README: add NIP-70 and `"-"` tag

AsaiToshiya:


theborakompanioni:
[NIP-70](https://github.com/nostr-protocol/nips/blob/master/70.md) is a breaking change, isn't it?

> The default behavior of a relay MUST be to reject any event that contains ["-"].

Should probably have an entry in [BREAKING.md](https://github.com/nostr-protocol/nips/blob/master/BREAKING.md), right?

AsaiToshiya:
It can also be interpreted as "a relay that supports this NIP".

PR #1356, opened 2024-07-10
[NIP93] Curatable Profile Gallery

believethehype:
This is a small NIP to define Kind 1163 events: Profile gallery entries.

I had withdrawn my earlier pull request #1340 which would have been an addition to NIP51 to add a List for a profile gallery.
After consulting with several people, this new version uses Kind 1163 Events (Gallery entries) instead of a "replaceable list with gallery entries". The reasoning is that lists might have size limits on relays and might easily be wiped by clients and clients can filter for these events instead. 

Feedback is welcome.

### DIFF
--- a/93.md
+++ b/93.md
@@ -1,0 +1,32 @@
+NIP-93
+======
+
+Profile Gallery
+-----------------
+
+This NIP defines curatable profile galleries that might be used across clients. The intention is that users might curate a gallery in their profile (or somewhere else) to show some of their visual content of choice, such as art, photographs etc. 
+
+The purpose of this NIP is to define an event Kind for Profile Galleries. A Gallery entry is a Kind 1163 Event consisting of an `e` tag and an `url` tag. The `url` tag must direct to a media file (such as an image or video) that will be shown in the gallery. The `e` tag should direct to a Kind 1 note containing this media file.
+
+
+
+## Example
+
+```json
+{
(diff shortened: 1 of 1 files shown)


pablof7z:
I'd remove the mention to kind:1; just say that, if there's another event this event is coming from, to tag it, could be a kind:1 or something else.

Also, relay hints and pubkey hints apply, worth being explicit about them if you're going to include an e tag example.

I would mention that an accompanying text could be added on the content for display alongside the image (like a caption)

believethehype:
Updated with relay hint, optional caption in content and removed reference to Kind 1

flox1an:
- Is the `e` tag required or optional? Can there be `1163` with just an url without an event reference? 
- I don't know if anyone uses them but maybe the `blurhash` from `1063` may be useful for a quick preview? Also it might be helpful to have `width` and `height` (or `dim`)  for rendering the gallery. 
- Should we add an optional `x` tag for the content sha256 content hash to make it blossom compatible?

believethehype:
An e tag is mandatory, as it references to a note (and the social aspects like comments, reactions, zaps around it).

I think other (potentially optional) parameters  from kind 1063 could be added if that helps clients rendering. I think for example from Amethyst it's not a problem to add x, dim and blurhash. Should I add these as optional params?

PR #1357, opened 2024-07-10
NIP-01: clarify single filter match

patrickReiis:
I didn't understand it first time when I read the current explanation, I think with this merge request it'll be easier for people with no nostr experience to understand it first time.

### DIFF
--- a/24.md
+++ b/24.md
@@ -40,4 +40,5 @@ tags
 These tags may be present in multiple event kinds. Whenever a different meaning is not specified by some more specific NIP, they have the following meanings:
 
   - `r`: a web URL the event is referring to in some way
+  - `i`: an external id the event is referring to in some way - see [NIP-73](73.md)
   - `title`: name of [NIP-51](51.md) sets, [NIP-52](52.md) calendar event, [NIP-53](53.md) live event or [NIP-99](99.md) listing
--- a/73.md
+++ b/73.md
@@ -1,0 +1,48 @@
+NIP-73
+======
+
+External Content IDs
+-------------------------
+
+`draft` `optional`
+
+There are certain established global content identifiers that would be useful to reference in nostr events so that clients can query all events assosiated with these ids.
+
+- Book [ISBNs](https://en.wikipedia.org/wiki/ISBN)
+- Podcast [GUIDs](https://podcastnamespace.org/tag/guid)
+- Movie [ISANs](https://en.wikipedia.org/wiki/International_Standard_Audiovisual_Number)
+
+Since the `i` tag is already used for similar references in kind-0 metadata events it makes sense to use it for these content ids as well.
+
+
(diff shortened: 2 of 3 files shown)


01.md
```diff
@@ -139,7 +139,7 @@ The `ids`, `authors`, `#e` and `#p` filter lists MUST contain exact 64-character
 
 The `since` and `until` properties can be used to specify the time range of events returned in the subscription. If a filter includes the `since` property, events with `created_at` greater than or equal to `since` are considered to match the filter. The `until` property is similar except that `created_at` must be less than or equal to `until`. In short, an event matches a filter if `since <= created_at <= until` holds.
 
-All conditions of a filter that are specified must match for an event for it to pass the filter, i.e., multiple conditions are interpreted as `&&` conditions.
+All different fields of a filter, if present, must match for an event for it to pass the filter. For example, if a filter has `{"kinds": [1, 3], "authors": [<pubkey>]}` it'll be interpreted as `(kind === 1 || kind === 3) && author === <pubkey>`.
```
AsaiToshiya, line 142:
```suggestion
All different fields of a filter, if present, must match for an event for it to pass the filter. For example, if a filter has `{"kinds": [1, 3], "authors": [<pubkey>]}` it'll be interpreted as `(kind == 1 || kind == 3) && author == <pubkey>`.
```

I think the equality operator is `==` in many languages.

AsaiToshiya (approved):

01.md
patrickReiis: Sorry, I mostly Deno. For me it doesn't matter, committing your suggestion.

patrickReiis:
What do you think? @staab

alexgleason:
The original text is concise. An example helps, but I'm not sure it belongs there. 

Maybe it would make more sense to put a bulleted list of some examples at the end of the section, similar to the relay "OK" message examples.

alexgleason:
I was thinking the same @Semisol

patrickReiis:
How would tests look like? Using this merge request as an example.

bezysoftware:
I would loooove to see gherkin tests for NIPs (either along each NIP or maybe in a separate repo?) This is a sample test from nostream:
```gherkin
Feature: NIP-01
  Scenario: Alice gets an event by ID
    Given someone called Alice
    And someone called Bob
    When Bob sends a text_note event with content "why is nobody talking about this?"
    And Alice subscribes to last event from Bob
    Then Alice receives a text_note event from Bob with content "why is nobody talking about this?"
```

I'm sure this could be adapted to cover the filter matching

bezysoftware:
I wrote a few NIP-01 gherkin tests for my Netstr project, check them out: https://github.com/bezysoftware/netstr/blob/main/test/Netstr.Tests/NIPs/01.feature

PR #1358, opened 2024-07-11
Update 90.md: Add additional commonly used NIP89 parameters to example

believethehype:
Extend NIP89 announcement example with commonly used parameters like picture, lud16 and encryptionSupported to signal if the dvm supports encrypted jobs


staab (approved):

vitorpamplona:
This is great. And looks like a few clients are already using the listed params. 

Next time just try to avoid changing format and content at the same time. It's just harder to review.

staab (approved):
Seems fine to me, although could we not expand all the json? It makes sense in codebases for easier editing/diffing, but is much harder to read in specs.

PR #1359, opened 2024-07-11
Set correct a-tag with the provided d-tag value in the example code (NIP-23)

Sebastix:
In the example code the d-tag value is `lorum-ipsum` so I adjusted the a-tag value with this value instead of `lorum`.

### DIFF
--- a/23.md
+++ b/23.md
@@ -54,7 +54,7 @@ References to other Nostr notes, articles or profiles must be made according to 
     ["published_at", "1296962229"],
     ["t", "placeholder"],
     ["e", "b3e392b11f5d4f28321cedd09303a748acfd0487aea5a7450b3481c60b6e4f87", "wss://relay.example.com"],
-    ["a", "30023:a695f6b60119d9521934a691347d9f78e8770b56da16bb255ee286ddf9fda919:ipsum", "wss://relay.nostr.org"]
+    ["a", "30023:a695f6b60119d9521934a691347d9f78e8770b56da16bb255ee286ddf9fda919:lorem-ipsum", "wss://relay.nostr.org"]
   ],
   "pubkey": "...",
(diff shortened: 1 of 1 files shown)


AsaiToshiya:
The `a` tag is a link, so I think it represents a different event.

Sebastix:
The a-tag should "link" to the current event as reference / identifier?

From [NIP-01](https://github.com/nostr-protocol/nips/blob/master/01.md?plain=1#L81):

```md
- The `a` tag, used to refer to a (maybe parameterized) replaceable event
    - for a parameterized replaceable event: `["a", <kind integer>:<32-bytes lowercase hex of a pubkey>:<d tag value>, <recommended relay URL, optional>]`
```

AsaiToshiya:
No. The `a` tag is a reference to another event, and the identifier for the current event is the `d` tag.

Sebastix:
Ok I'm confused 🤔

What is the purpose of the a-tag on a kind 30023 event (and probably for other kinds too)?

AsaiToshiya:
Maybe `kind:30023`'s `a` tags are used to search for backlinks.

Sebastix:
Anyone else (NIP-23 / a-tags experts hehe) who could clarify this? @fiatjaf @pablof7z ?

erskingardner:
@Sebastix - @AsaiToshiya is right. `a` tags on all events are meant to link to other events. Whether it's an event that is being referenced (as in a reply to or linked to) doesn't matter. Most other event kinds stipulate how the `a` tags are used but it's always another event, not the event itself.

Sebastix:
Thanks :)

Could we clarify this more? Or is it documented but I haven't seen / found it (yet)? ☺️ 

I've updated the PR with a proposal to make it more clear.

PR #1360, opened 2024-07-11
Mention Protected Events in NIP-01

bezysoftware:
The default behavior of rejecting protected events when client isn't authenticated should be at least mentioned in NIP-01 with links to NIP-42 and NIP-70.

### DIFF
--- a/01.md
+++ b/01.md
@@ -73,13 +73,14 @@ Each tag is an array of one or more strings, with some conventions around them. 
 
 The first element of the tag array is referred to as the tag _name_ or _key_ and the second as the tag _value_. So we can safely say that the event above has an `e` tag set to `"5c83da77af1dec6d7289834998ad7aafbd9e2191396d75ec3cc27f5a77226f36"`, an `alt` tag set to `"reply"` and so on. All elements after the second do not have a conventional name.
 
-This NIP defines 3 standard tags that can be used across all event kinds with the same meaning. They are as follows:
+This NIP defines 4 standard tags that can be used across all event kinds with the same meaning. They are as follows:
 
 - The `e` tag, used to refer to an event: `["e", <32-bytes lowercase hex of the id of another event>, <recommended relay URL, optional>]`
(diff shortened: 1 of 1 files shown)


fiatjaf:
I don't know, isn't this an inversion of control?

erskingardner:
Yeah, I agree with @fiatjaf, we should add the `-` tag to the reference list of tags in the readme (might already be done) but I don't think this belongs in `01.md`.

bezysoftware:
Thanks for your comments. I personally don't have a problem with making NIP-01 more of a "signpost" outlining default behaviors (mandatory) with links to other NIPs (optional). But if you think this isn't a good place then I'll defer to your judgement,

Upon reflecting on this I realized that the reason I wanted to add this here was because I miss a single place I could watch for NIPs which are relevant to me. I'm currently developing a relay (still early with NIP-01 stuff) and I only stumbled across NIP-70 by accident. I find it very hard to filter NIPs which are relevant to me (as a relay dev). Sure I can just go one-by-one but (which is time consuming) but that's still not enough - I have to actively watch this entire repo for new NIPs / changes to existing ones and determine if they impact me. 

What I would appreciate would be some categorization(grouping) of individual NIPs / "meta NIP document" / or some other way to filter them. 

Does it make sense? Or do you think I'm just lazy when I don't want to scan thru a 100 documents (honestly interested in your thoughts)?

fiatjaf:
It makes sense that you would want that, but consider that we're all lazy here and having such a meta document would require a lot of work. I have begun writing a high-level documentation of all things Nostr that I think should take care of that, but, again, it's a lot more work than it sounds, so who knows what it's going to happen.

bezysoftware:
> It makes sense that you would want that, but consider that we're all lazy here and having such a meta document would require a lot of work. I have begun writing a high-level documentation of all things Nostr that I think should take care of that, but, again, it's a lot more work than it sounds, so who knows what it's going to happen.

Maybe use some kind of tags for each NIP and then filter by those tags?

Issue #1291, opened 2024-06-07
AND Filters


dskvr:
There have been a number of discussions about this topic, and it's even been implemented in a PR to `nostr-rs-relay`. I published a gist 6 months ago with the hopes someone would pick up where I left off, but nobody has.

# tldr 
```
filters: {
  "kinds": [1],
  "&t": ["meme", "cat"],
  "#t": ["black", "white"]
}
// returns kind `1` events with `t` tags that have both "meme" and "cat" that have the tag "black" or "white"
```

# Resources 
- The [gist](https://gist.github.com/dskvr/abccd012f90f9771fc53fc173e005f33/edit)
- The [PR](https://github.com/scsibug/nostr-rs-relay/pull/178)

# Rationale 
1. Makes filters more powerful/intuitive
2. Optimizes bandwidth 
3. Potentially optimizes relays
4. Potentially optimizes client development

vitorpamplona:
Making AND available only for tag queries makes sense in my mind.

PR #1365, opened 2024-07-15
NIP-119: AND operator for filters

dskvr:
Filter modifier that has optimization benefits for relays, users and developers. _Likely ### contentious._

## Discourse
- [issue](https://github.com/nostr-protocol/nips/issues/1291)

## Implementations
- [nostr-rs-relay (draft PR)](https://github.com/scsibug/nostr-rs-relay/pull/178)


alexgleason:
Have you implemented a relay? I can't even make tag queries fast as they are.

dskvr:
I have not, no. But there is a PR to `nostr-rs-relay` as listed above. 

> I can't even make tag queries fast as they are.

`AND` queries are objectively faster for `btree` and equivalent for all other indexing methods, so if you are using `btree` indices `AND` filters are faster.

rtxs34 (approved):

mikedilger:
I cannot think of a use case where I want to ask a relay for only events that have the same tag twice with two different values.

dskvr:
@mikedilger @fiatjaf Really? I can think of an infinite number of cases. Here's one: `{ kinds: [2003], "#t": [ "movie", "4k" } }`

fiatjaf:
I think this proposal is reasonable if we limit `&` to 2 values -- more than that and it becomes infeasible for filtering on the relay side.

But even if it's reasonable I also think it's a slippery slope and soon people will be demanding full SQL support on relays.

alexgleason:
Postgres has a `@>` operator that is almost identical in functionality to Nostr filters, except it uses AND instead of OR for array values. If anything supporting OR was actually harder, because I had to break the query up into multiple `@>` statements for each value in the list.

We have been able to work around Nostr filter limitations by changing the events themselves in many cases. I think the torrent category example (or Amazon product filtering as another hypothetical), is still basically unsolved, and we worked around that by simply not supporting the ability to do it. :smiley: 

I'm not sure it's worth changing NIP-01 filters. I've been shoehorning any extra functionality I need into NIP-50 [search extensions](https://github.com/nostr-protocol/nips/blob/master/50.md#extensions).

dskvr:
> I think this proposal is reasonable if we limit & to 2 values

Why not just leave it to the operator to decide? Maybe i run a relay that stores a unique event kind and I have optimized it to support up to 20 and filters. Why shouldn't I be able to allow a client to run `REQ`s up to my determined limit? 

> But even if it's reasonable I also think it's a slippery slope and soon people will be demanding full SQL support on relays.

Possible, and fair point. However, as proposed it's an **optimization** that **just so happens to be a feature**. If data is provided to prove that there is optimization benefits for all parties (relays, clients, users) this would establish a baseline that could be used as a measure against filter creep.

> I'm not sure it's worth changing NIP-01 filters.

As proposed it's additive and does not "change" NIP-01 filters.

119.md
```diff
@@ -0,0 +1,18 @@
...
+  "#t": ["black", "white"]
+}
+// returns kind `1` events with `t` tags that have both "meme" and "cat" that have the tag "black" or "white"
+```
+
+## Rules
+
+- `AND` **MUST** take precedence over `OR`
+- Tags used in `AND` **SHOULD NOT** be used in standard `OR` tags [`#`]
+- Any tag used in `AND` **SHOULD** be ignored in `OR`
```
bezysoftware, line 18: Can you elaborate on these rules? They say

```Tags used in `AND` **SHOULD NOT** be used in standard `OR` tags```

and yet in your example filter you specifically use the ```t``` tag in both AND and OR filters.

**Comment by bezysoftware** [comment]
How would current relay implementations signal supporting this if it becomes part of 01? I think making this a separate NIP is better, with NIP 11 signaling. Much like NIP 70, that feels pretty similar. 

And perhaps make NIP 11 required.

**Comment by bezysoftware** [comment]
My question was more about the transition period after merging. The current implementation wouldn't magically start supporting this, you would need a mechanism to check for support. 

I agree the current state isn't great, but frankly that's always going to be a problem. You'll always have to code defensively and have fallbacks when a relay says one thing (or nothing), but behaves differently.

**Comment by vitorpamplona** [comment]
Agree. We can have: 
1. A minimum required set of features for the base protocol + poor signaling support and a few very rare add-ons
2. An expanded required set of features for the base protocol + poor signaling support and a few very rare add-ons
3. A minimum required set of features for the base protocol + great signaling support and great add-ons
4. An expanded required set of features for the base protocol + great signaling support and great add-ons

Today we have 1. If we think this PR is just for VERY RARE use cases where the client and the relay are probably coming from the same developer, then this makes sense. 

If we want to see more usage of this, then we have to go for 2 or 3. Since we like simplicity, 4 is not desired on Nostr.

**Comment by staab** [comment]
Feature support signaling is a hassle, but that's sort of the only downside. It would be nice to make any additions signal-able by putting them in a new NIP, so relays can signal support if they want to (even though many don't).

**Comment by fiatjaf** [comment]
The most important downside of feature signaling is that it makes changes very easy.

**Comment by bezysoftware** [comment]
> The most important downside of feature signaling is that it makes changing things a much easier process than currently.

But currently feature signaling is done by adding new NIPs, so.. I don't follow. Do you mean we should have a different way to signal feature support?

**Comment by bezysoftware** [comment]
FYI I have this implemented in [netstr](https://github.com/bezysoftware/netstr/) relay and deployed to a dev instance: `wss://relay-dev.netstr.io` 
Feel free to play with it

**Comment by dskvr** [comment]
What's worse than support signaling? A NIP that is merged too quickly and changes `n` times.

IMO Just leave this open and let relays signal support for it, like it already happening. If it works and is found helpful, the benefits of a merge will be implicit, the need for discussion will be reduced and the effort of merging will be minimized. 

I see no reason to rush this NIP (or really, most NIPs).

File: 119.md (hunk 1 above)
Comment (bezysoftware) [reply 119.md]: Got it, second point is asking clients "please don't do this", third is telling relays "if they do it, ignore it"

**Comment by bezysoftware** [comment]
I'm personally happy with this NIP and deployed it to production: https://relay.netstr.io/

**Comment by lez** [comment]
I want to add a use-case for the possible use of this NIP.
If someone was to create a deli.cio.us or pinboard.in like public bookmarking tool over nostr, this NIP would be practical to filter other people's bookmarks based on tags, like `#nostr #messaging #android`

Is there any objection left here? When is it going to be merged?

**Comment by 1l0** [comment]
That's natural.

**Comment by dskvr** [comment]
I already use it every day without it being merged. Would be best IMO if adding this to NIP-01  a majority of relay devs to weighed in first. Otherwise, merging it as a separate NIP would cause fragmentation issues and make client-dev more complicated.

@mikedilger Could you give an example? As interpreted it seems both more complex and less intuitive than this proposal, but it's possible (and likely) I am misunderstanding your idea.

**Comment by mikedilger** [comment]
Example:  Maybe you want to ask for "(Movie and 4k) or (Audio and 256kbps)".  This could be represented as:

```
"#t": [["Movie", "4k"], ["Audio","256kbps"]]
```

You could mix in other tag values as we already do (and it means what it already means, it is an OR term) like this:

```
"#t": [["Movie", "4k"], ["Audio","256kbps"],"memes"]
```

That is, this only introduces the sub-array with the meaning that the terms are ANDed within the subarray.

To get full boolean expressions you could lead terms with "!":

```
"#t": [["Movie", "!18+"]]
```

I am not making a proposal, I really don't know if it is a good idea. I'm just saying we are quite close to having full boolean logic already.  And I know from my studies at university half a lifetime ago that all possible boolean expressions can be put into this form.

dskvr:
Oh, ok, I understand. I can see benefits to it. 

I suppose my main argument against that format would be that it breaks the fundamental typing of tags. So instead of `tuple of strings` it becomes `tuple of strings OR arrays`, which isn't backwards compatible.

The proposed format is non-breaking, whereas a schema change would be breaking.

vitorpamplona:
`&` tags as proposed here is the way to go. The array-based construction doesn't seem that useful and it is more confusing/hard to code.

mikedilger:
> Oh, ok, I understand. I can see benefits to it.
> 
> I suppose my main argument against that format would be that it breaks the fundamental typing of tags. So instead of `tuple of strings` it becomes `tuple of strings OR arrays`, which isn't backwards compatible.
> 
> The proposed format is non-breaking, whereas a schema change would be breaking.

Of course I agree with that. The particular encoding wasn't really the message, and it wasn't a proposal.  Carry on.

dskvr:
NIP-119 -> NIP-ND

hzrd149:
`applesauce-core` v4.4 will have support for NIP-ND https://github.com/hzrd149/applesauce/pull/32

This will allow for client-side AND tag filters

hzrd149:
@dskvr since this NIP was renamed from NIP-119 to NIP-ND, how are relays supposed to signal support for this in their NIP-11 document?

fiatjaf:
Pick a number like 91.

dskvr:
NIP-ND was a troll to the artifical capping of nips and of using hexadecimal as a workaround. It's been updated.

hzrd149:
Opened PR to nostr-tools for JavaScript https://github.com/nbd-wtf/nostr-tools/pull/518

dskvr:
NIP-ND -> NIP-91

hzrd149:
Built an example in applesauce for https://hzrd149.github.io/applesauce/examples/#hashtags/explore

vitorpamplona:
I just realized this is great for decrypted NIP-17 group chats where we need to filter by the presence of many p-tags in local dbs. 

@jb55 since you are putting them on nostrdb, have you thought about indexing in this way?

vitorpamplona:
Added to Amethyst/Quartz

Semisol:
Next Nostr.land relay software release will support this

dskvr:
This NIP now has wider support than the majority of approved NIPs did before they were merged, and more support than a handful of the "approved" NIPs do years after they were merged.

vitorpamplona:
@dskvr have you thought about a similar filter but limited by only the mentioned tags? 

For instance, I want all the kind 1s that cited Myself and Amethyst's key in the same event, but no other p tags.

dskvr:
@vitorpamplona I will not be making any functional amendments to this NIP.

fiatjaf:
Sorry, this shouldn't have been merged. Rewriting history now.

dskvr:
HAHAHAHAHAHAHAHAHAHAHAHA

@staab @vitorpamplona @eskema could you reopen it please?

staab:
I don't have the button, you'll have to open a new one.

dskvr:
@fiatjaf for all the "centralization slippery slope" talk you are definitely not leading by example by rewriting history because you don't like a widely-adopted NIP :1st_place_medal:

staab:
Is it widely adopted though? I see one relay signaling support on nostr.watch

hzrd149:
@fiatjaf @staab if this wasn't supposed to be merged then it at least should have been left open so the conversation could continue. I don't see why it was necessary to merge then revert / force push

staab:
@hzrd149 yes, I agree. I merged because this looks good and technically meets the requirements in the readme, but fiatjaf is right that it isn't really "widely adopted" (at least I don't see any evidence that it is). I don't want to argue about it, I'm not interested in this NIP so I'm just not going to participate any more, but I can't reopen because github won't let me.

dskvr:
@staab @fiatjaf 

> Is it widely adopted though?

# NIP-32 "OK!"

swappy-20260304_234718

# NIP-91 "not widely adopted"

swappy-20260304_234835

# Final Outcome
- merge
- rewrite history
- "not interested in this NIP, not participating" 
- _PR buried_

staab:
That's what draft PRs are for you melon. There were several implementations by the time it was merged.

dskvr:
gfy you tyrant.

alexgleason:
It's a good idea that should have been part of NIP-01 to begin with. But the problem with it is that it changes the NostrFilter interface, one of the 2 only core interfaces to the Nostr protocol. Meaning that every Nostr library in existence would have to be updated. At this point I think it would be exciting to see a big change like this, though. Give Nostr devs something to do lol.

dskvr:
@alexgleason The way applesauce did it was simple and elegant. Clients can support `AND` without relays supporting `AND`, the library just does the intersections locally when relay doesn't have support, not different from how client-developer has to do right now. Instead of `AND` being implemented by hand by each developer, it can be signaled by a protocol level operand, offloaded to the library's post-filters; something that already exists in every single library. If relays implement NIP-01 correctly, then they just ignore the filter. It's **voluntary**. It breaks nothing. Find me a case that breaks anything, for any properly and defensively implemented client, relay or library; I will happily eat my words. As libraries and relays support it, clients adopt it, and everbody wins. Every. Single. Stakeholder: Relay Operators, Client Developers and Users.

lez:
> But the problem with it is that it changes the NostrFilter interface

No, this is our *job* to change the interface in all implementations.

mikedilger:
> > But the problem with it is that it changes the NostrFilter interface
> 
> No, this is our _job_ to change the interface in all implementations.

Please submit PRs for chorus and relay-tester, neither of which support this.  I don't have the luxury of having time to keep up with nostr changes.  But if the community wants these, I'll accept the PRs.  Chorus is a bit tricky as it does a lot of low-level optmized json handling and elides (skips past) the '#' character early on, which will have to be undone.

PR #1367, opened 2024-07-17
NIP-55: Add intent flags to open the signer once when sending multiple events

greenart7c3:


greenart7c3:
@vitorpamplona can you review this?
It was missing asking for multiple signatures in this nip

vitorpamplona:
The guidance seems right. And we are already supporting this, right? This PR just adds information on how to do it correctly.

greenart7c3:
Yes, we already support this

vitorpamplona:
I merged. I am not sure this level of detail should be in the NIP, but we can always remove it later if it becomes too big.

dluvian:
Check spelling before merging :D

PR #1369, opened 2024-07-18
Cashu wallet + Nutzaps

pablof7z:
This PR introduces two independent but highly related NIPs.

Either NIP can be implemented separately and has independent benefits.

## **NIP-60**: Cashu wallets
Store Cashu proofs in relays so that a wallet balance can follow you around nostr and can be easily spent from any client. Just like you load a client and your follow list is there, so can your pocket-money.

## **NIP-61**: Nutzaps
Scheme to send nutzaps -- these are P2PK-locked cashu transfers that are NOT encrypted, but require a signature from a private key to be redeemed. Anyone can see the payload but only the holder of the private key can spend them.

Later on, I'll add DLEQ proofs to the nutzaps to make it possible to validate the validity of the proofs without talking to the mints.

Nutzaps have no _zapper pubkey_, where a third-party has to sign-off that a zap occurred.

NIP-60 money can be used to send NIP-57 zaps (by instructing the mint to pay the recipient's bolt11) or NIP-61 nutzaps.

# Important considerations
Because ecash is such a contentious topic, here's a TL;DR:
* you must OPT-IN to receive NIP-61 nutzaps -- if you don't signal with a bunch of mints you want to receive nutzaps on then people can't nutzap you.
* nutzaps can provide a way around the issues that hodl invoices (e.g. zaps to zeus users) cause with force-closes -- a zeus user *could* use a nutzap as a temporary hold while they are offline and the moment they come online sweep everything into their zeus wallet.


60.md
```diff
@@ -0,0 +1,188 @@
...
+
+## Token Event
+Token events are used to record the unspent proofs that come from the mint.
+
+There can be multiple `kind:7375` events for the same mint, and multiple proofs inside each `kind:7375` event.
+
+```jsonc
+{
+    "kind": 7375,
+    "content": nip44_encrypt({
```
ebrakke, line 67: Curious, why not use the cashu token encoding here, rather than the raw json?
https://github.com/cashubtc/nuts/blob/main/00.md#v4-tokens

dluvian (commented):
I feel like there is too much friction on the sender side of a nut zap. A sender has to fetch the Nutzap informational event, check if listed relays and mints are online and then mint new tokens in one of these mints. This complexity discourages client devs to adopt this standard and nut-zap-senders are discouraged from zapping if almost all of their attempts fail because the intended recipient hasn't published a `10019` event yet or the listed relays and mints are offline.

I haven't dabbled in the Cashu protocol yet but isn't it possible to timelock Cashu tokens with conditions? A sender could then timelock tokens in a way that allows the recipient to claim them anytime and the sender to reclaim them if they haven't been claimed after n days. Nuts will never get lost, relays don't have to be trusted and the sender can always immediately send those nutzaps.

60.md
calvadev: It would add an unnecessary extra encryption/decryption step.

pablof7z:
> I feel like there is too much friction on the sender side of a nut zap. A sender has to fetch the Nutzap informational event, check if listed relays and mints are online and then mint new tokens in one of these mints. This complexity discourages client devs to adopt this standard and nut-zap-senders are discouraged from zapping if almost all of their attempts fail because the intended recipient hasn't published a `10019` event yet or the listed relays and mints are offline.
> 
> 
> 
> I haven't dabbled in the Cashu protocol yet but isn't it possible to timelock Cashu tokens with conditions? A sender could then timelock tokens in a way that allows the recipient to claim them anytime and the sender to reclaim them if they haven't been claimed after n days. Nuts will never get lost, relays don't have to be trusted and the sender can always immediately send those nutzaps.

No, this is as simple as it gets; anyone can nutzap without checking the 10019 of the recipient, but the recipient might never see those nutzaps if they don't check for them; this spec is actually incredibly simple and adding a timelock component would only make it more complex.

Checking if the infrastructure is online is a downside of a decentralized protocol; sure, it would be more straightforward to assume that you can publish to api.twitter.com and not even bother with error handling, but any nostr client needs must properly handle trying to speak with a few relays and having one of them be offline.

Fwiw, I wrote and implemented the entire spec in a few hours, this is not hard stuff whatsoever.

Also, you might have misunderstood the spec, relays don't need to be trusted in the same way NIP 65 relays don't need to be trusted, the listing in the 10019 is just a shorthand to make it easier to know where to publish in a way the recipient will see it. It's literally;y the exact same thing as nip 65 or as DM relays

60.md
pablof7z: Yup, what calvadev said

vitorpamplona:
Can the user's relay push its user by deleting all 7375 and 7377 events? I suppose you cannot recover the tokens if you don't have these events anymore.

dluvian:
> relays don't need to be trusted
 
You need to trust the relay to not delete the token event and to deliver the event in the future. If you nutzap a 1-of-2 multisig nut then you don't have to trust the relay anymore as you can reclaim the funds if something goes wrong on the relay side. And if you can reclaim then you're free to zap anyone, making it less restrictive and simpler.

pablof7z:
@vitorpamplona 
> Can the user's relay push its user by deleting all 7375 and 7377 events? I suppose you cannot recover the tokens if you don't have these events anymore.

I don't understand the comment -- "relay push its user"? what does that mean? like kick out its user from using the relay?

At one point I had this appendix, that might shed some light to this concern, if that's what you meant. I removed it from the NIP because it sounded too paternalistic and clients can do whatever they choose to ensure their users are ok (i.e. like choosing good and enough relays, etc)


> ## Appendix 1: Wallet health
> Clients can check the resilience of the relays the are using to make sure tokens are not getting lost. Clients should regularly check if they are receiving the tokens they expect to receive across the set of relays where their proofs are being stored, and if a relay is not responding with the right tokens, the client can warn the user and/or roll the tokens to a new relay and discard the offending relay.

pablof7z:
> > relays don't need to be trusted
> 
> You need to trust the relay to not delete the token event and to deliver the event in the future. If you nutzap a 1-of-2 multisig nut then you don't have to trust the relay anymore as you can reclaim the funds if something goes wrong on the relay side. And if you can reclaim then you're free to zap anyone, making it less restrictive and simpler.

you're mixing two separate concerns: 1) token availability -- 2) clawback possibility

when you talk about "you need to trust the relay" you are exclusively talking about 1) because if you don't have the token you can't claw it back, regardless of what script you had in it, so the multisig comment is irrelevant here.

keep in mind that the recipient WANTS to receive the money, so they wouldn't choose relays that are going to delete their incoming money, that makes no sense. And ultimately the sender can publish to the relays the recipient specified and to their own relays or somewhere else.

vitorpamplona:
@pablof7z sorry, I meant to punish the user. We have seen relays just going bust or deleting events at their will, either reacting to something the user uploaded or just as a suspicion of malfeasance. The warning you mention is good but it can be clearer: 

In the current version of this NIP, the user can be rug pulled by the mint AND by the relay he/she decides to use. 

I know the same risk is true for all nostr events, but losing money is very different than losing Twitter posts.

pablof7z:
> @pablof7z sorry, I meant to punish the user. We have seen relays just going bust or deleting events at their will, either reacting to something the user uploaded or just as a suspicion of malfeasance. The warning you mention is good but it can be clearer: 
> 
> In the current version of this NIP, the user can be rug pulled by the mint AND by the relay he/she decides to use. 
> 
> I know the same risk is true for all nostr events, but losing money is very different than losing Twitter posts.

Ok, perhaps it should be explicit, but yeah, this is dollars to be used with multiple relays, particularly ones you run or trust; much like the drafts nip, I would hurt more if I lost my drafts than if I lose a few thousands sats. (Not kidding, I have tweet drafts I've only saved in typefully and have gone back to reading them for years, if typefully goes bust it'll suck; I need to move thise to nostr drafts asap)

I'll update the nip to make this explicit and perhaps add the Appendix as is if you think it's helpful.

60.md
```diff
@@ -0,0 +1,188 @@
...
+        [ "a", "37375:<pubkey>:my-wallet" ]
+    ]
+}
+```
+
+`.content` is a [[NIP-44]] encrypted payload storing the mint and the unencoded proofs.
+* `a` an optional tag linking the token to a specific wallet.
+
+### Spending proofs
+When one or more proofs of a token are spent, the token event should be [[NIP-09]]-deleted and, if some proofs are unspent from the same token event, a new token event should be created rolling over the unspent proofs and adding any change outputs to the new token event.
```
davidcaseria, line 88: This is challenging to maintain on the client since each spend requires the client to determine which events need to be deleted and what proofs from the deleted event should roll over to the new event while possibly adding more proofs.

What if the `Spending History Event` had inputs of the spent proofs and outputs of the new proofs instead? This event would now be mandatory, but it would make wallet tracking easier on the client since all the client has to do is replay all the events to get the current state. When the wallet makes a transaction, it is an easy-to-construct single event.

Maybe optionally, there can be a wallet "snapshot" or backup event that has all the unspent proofs at a specified point in time. The wallet can then replay the spending events since then to get the current state.

water783:
NIP-61 still doesn't seem to address my previous concern: 

If the content(nuts) of kind7337 is obtained by someone else and they also send a kind7337 event with this content(nuts), then who is the real person sent this nuts zap? How does the client determine this?

callebtc:
> If the content(nuts) of kind7337 is obtained by someone else and they also send a kind7337 event with this content(nuts), then who is the real person sent this nuts zap? How does the client determine this?

Very good point! I think this can be addressed: you can commit to arbitrary data by adding [tags](https://github.com/cashubtc/nuts/blob/main/10.md#well-known-secret) to the `Secret` field of a Cashu proof, just like you would with a nostr note.

callebtc:
> A sender could then timelock tokens in a way that allows the recipient to claim them anytime and the sender to reclaim them if they haven't been claimed after n days. Nuts will never get lost, relays don't have to be trusted and the sender can always immediately send those nutzaps.

This is certainly possible from a Cashu perspective and maybe even compliant with this spec? A P2PK-locked token can have a `locktime` tag which would allow the sender to redeem it back. Clients should at least be aware of this (and treat the zap as valid only if `local time < locktime`) or regard timelocked zaps as invalid.

pablof7z:
> > If the content(nuts) of kind7337 is obtained by someone else and they also send a kind7337 event with this content(nuts), then who is the real person sent this nuts zap? How does the client determine this?
> 
> Very good point! I think this can be addressed: you can commit to arbitrary data by adding [tags](https://github.com/cashubtc/nuts/blob/main/10.md#well-known-secret) to the `Secret` field of a Cashu proof, just like you would with a nostr note. If this included the note ID that is being zapped, replays should be impossible.

Awesome, yeah, this is a good point -- I'll update the NIP to include tags for the sender and an optional "e" of the zapped note.

So the proofs would have an additional `"tags": [ [ "p", "<sender-pubkey>" ], [ "e", "<optional-nutzapped-event>" ] ]`

davidcaseria:
In order for a client to implement a complete wallet, there needs to be a way to store [NUT-04](https://github.com/cashubtc/nuts/blob/main/04.md) and [NUT-05](https://github.com/cashubtc/nuts/blob/main/05.md) quotes. I believe a new expirable event for each quote is probably best suited to handle the wallet's storage needs. Note that the quote ID **must** remain private, so it needs to be in an encrypted content payload.

davidcaseria:
Another case that should be handled is the different statuses a proof can be in. For example, when sending a token, CDK puts a proof in a `Reserved` status. This status is essential so that a wallet does not double-spend a proof. Only later does the CDK wallet check if the sent proofs have been claimed and remove them from the database.

pablof7z:
> Another case that should be handled is the different statuses a proof can be in. For example, when sending a token, CDK puts a proof in a `Reserved` status. This status is essential so that a wallet does not double-spend a proof. Only later does the CDK wallet check if the sent proofs have been claimed and remove them from the database.

yeah, the idea I had for this was to moved the proofs that are in reserved status to a new kind, so the same operation of rolling over proofs from a token event after they're spent, but move them to a different event kind, exact same format; upon initialization a wallet can check for this event kind and check if the proofs have been spent or not and either delete them or roll them back in into the available tokens kind.

davidcaseria:
What if instead of having to manage deleting events, the token event was structured in such a way that the action for the proof was embedded:
```
{
  "a": [<proof>], // added proofs
  "d": [<proof>], // deleted proofs
  "r": [<proof>]  // reserved proofs
}
```

I want to refrain from relying on deleted events because that can cause issues if mishandled.

pablof7z:
> What if instead of having to manage deleting events, the token event was structured in such a way that the action for the proof was embedded:
> 
> ```
> {
>   "a": [<proof>], // added proofs
>   "d": [<proof>], // deleted proofs
>   "r": [<proof>]  // reserved proofs
> }
> ```
> 
> I want to refrain from relying on deleted events because that can cause issues if mishandled.

Oh yeah, that's an interesting idea; I'm not sure we need the deleted proofs one, the state of these events shouldn't grow too much as that would make syncing a bit more expensive -- is there a benefit to keeping that state?

> can cause issues if mishandled.
What do you mean by "mishandled"?

Mishandled by the relays, as in, you instruct the relay to delete and it doesn't? Or what do you mean exactly?

60.md
pablof7z: > This is challenging to maintain on the client since each spend requires the client to determine which events need to be deleted and what proofs from the deleted event should roll over to the new event while possibly adding more proofs.
> 
This is not problematic at all, or maybe I'm not understanding your point; but if the client fetches, say, 4 events, which each has, say, 3 proofs inside, total of 12 proofs.

If the client spends, say, 10 of those proofs, it's trivial to know it has to rollover the 2 unspent proofs into a new token event and delete all 4 token events.

Which equates to the client publishing 2 events:
Kind 9 (event deletion) e-tagging 4 events
Kind 7375 storing two proofs.

Just to be very explicit, the relays won't send the 4 deleted events any more when a client REQs for 7375s, it would now just reply with a single event, the one that contains the two proofs.

Again, perhaps I misunderstood your comment 😂

60.md
```diff
@@ -0,0 +1,188 @@
...
+}
+```
+
+* `direction` - The direction of the transaction; `in` for received funds, `out` for sent funds.
+* `a` - The wallet the transaction is related to.
+
+Clients MUST add `e` tags to create references of destroyed and created token events along with the marker of the meaning of the tag:
+* `created` - A new token event was created.
+* `destroyed` - A token event was destroyed.
+* `redeemed` - A [[NIP-61]] nutzap was redeemed.
```
believethehype, line 113: Would it make sense to also log outgoing zaps? When balance changes because of "out" it could be due to the user e.g. melting tokens to lightning, but it could also be dude to sending a zap. (as in "redeemed" but in the other direction)

water783:
NIP-61 uses `p2pk-pubkey` to receive eCash, which is great for enhancing security. However, NIP-60 with `kind 37375` places the `privkey` within the content. If a user's `privkey` is compromised, the `p2pk-pubkey` would also be exposed. My point is, instead of storing the `privkey`, could we allow users to securely manage their account keys on their own? Alternatively, could we store `ncryptsec` rather than the `privkey` to mitigate this risk?

60.md
ticruz38: I tend to agree with @davidcaseria 
The annoying part with deleting events is if you have two nip60 wallet implementations on your device or different device.
You would store those events/proofs in an indexdb or local storage. 
To get these 2 wallets in sync means replaying all the unspent proof events from the wallet creation (to account for the deleted one that once existed) Which you probably don’t want to do each time you open your app.

the append only method sounds simpler and more robust to sync the wallet accross platform/app, you just take all the unspent proofs events minus the spent proofs event and you should be fine.

it doesn’t change anything for that NIP except that the spent proofs feed/event would now be mandatory. One can still delete the unspent proofs event but this would now be optional.

ticruz38: Also if the wallet wants to show an history for the user, which most neobanks do (wise, Revolut etc) keeping the full event history is a must

rodant: I have the feeling the NIP is getting too deep into implementation details at this point, which isn't the goal of a NIP. It should be an implementation decision how to track the balance of the wallet. I'd find great to make the `Spending History Event` mandatory and thus wallets could choose an event sourcing approach as proposed by @davidcaseria too.

60.md
```diff
@@ -0,0 +1,188 @@
...
+        ]
+    }),
+    "tags": [
+        [ "a", "37375:<pubkey>:my-wallet" ]
+    ]
+}
+```
+
+`.content` is a [[NIP-44]] encrypted payload storing the mint and the unencoded proofs.
+* `a` an optional tag linking the token to a specific wallet.
```
rodant, line 85: Why define this optional? I think this could lead to some weird issues. Imagine an user having two wallets and both picking this event. But the proofs belong to only one of the wallets.

ticruz38:
Might be asking in the wrong place entirely

Don't you think it would be possible to somehow use a Nostr extension, NIP-07, to let the mint know you are indeed in possession of the private key to the P2PK proofs that this mint has emitted?
The step that want you to encrypt the wallet nsec in the Nostr don't seems super safe and smooth, the beauty of those nips is to use your Nostr profile as your wallet.

You would just need to sign some sort of secret, that's probably possible already...

@callebtc ?

ekzyis (commented):
This looks cool! I haven't read through the details yet but I am already intrigued by this. I was wondering about this though:

> NIP-60 money can be used to send NIP-57 zaps (by instructing the mint to pay the recipient's bolt11) or NIP-61 nutzaps.

Can't NIP-60 money be used to pay _any_ bolt11 (like any other cashu wallet), no matter how we fetched it? As this is written with specific mentions of NIP-57 and NIP-61, this suggests that might actually not be the case.

60.md
```diff
@@ -0,0 +1,188 @@
...
+
+The purpose of this NIP is:
+* ease-of-use: new users immediately are able to receive funds without creating accounts with other services.
+* interoperability: users' wallets follows them across applications.
+
+This NIP doesn't deal with users' *receiving* money from someone else, it's just to keep state of the user's wallet.
+
+# High-level flow
+1. A user has a `kind:37375` event that represents a wallet.
+2. A user has `kind:7375` events that represent the unspent proofs of the wallet. -- The proofs are encrypted with the user's private key.
```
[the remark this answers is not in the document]
pablof7z: you can encrypt (NIP-44) with your nostr private key

61.md
```diff
@@ -0,0 +1,132 @@
...
+
+## Final Considerations
+
+1. Clients SHOULD guide their users to use NUT-11 (P2PK) compatible-mints in their `kind:10019` event to avoid receiving nut zaps anyone can spend
+
+2. Clients SHOULD normalize and deduplicate mint URLs as described in NIP-65.
+
+3. A nut zap MUST be sent to a mint the recipient has listed in their `kind:10019` event or to the NIP-65 relays of the recipient, failure to do so may result in the recipient donating the tokens to the mint since the recipient might never see the event.
+
+## Appendix 1: Alternative P2PK pubkey
```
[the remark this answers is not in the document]
pablof7z: this is already there, the only caveat is that to spend those p2pk proofs that were locked to your main pubkey the client needs access to the private key (i.e. not a nip-07 or nip-46 interface) as you need to sign a non-nostr payload.

60.md
pablof7z: it doesn't matter, only one of them will be able to spend it; it's up to the client to differentiate and not list the funds as the same.

The reason it's optional is that a user might choose to completely skip the creation of a wallet or association of tokens with a specific wallet.

60.md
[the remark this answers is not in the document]
ekzyis: No you don't: For NIP-44, you calculate a shared secret out of sender's private key + recipient's public key, then you calculate a message key out of that, then you encrypt with that. Saying "you can encrypt (NIP-44) with your nostr private key" is confusing. You need to state the sender and recipient for NIP-44.

My suggestion would be to simply write "the proofs are encrypted with NIP-44" if the above is what you mean. Encrypting with a private key is an oxymoron.

However, maybe the confusion arises from this: I can't tell who the recipient is since there's never any `p` tag in this spec (by design) from which the recipient's public key could be deduced. Therefore, I don't know which sk+pk I should use for NIP-44 for the wallet and token event. Does the following mean you're using sk+pk of the same key pair since you're essentially sending these events to yourself according to `<my-pubkey>`:

> **Fetch wallet and token list**
>
> From those relays, the client should fetch wallet and token events.
>
> `"kinds": [37375, 7375], "authors": ["<my-pubkey>"]`

? (not sure if that's even possible)

ekzyis:
> Are nutzaps so much better than NIP-57 to justify creating a competing standard for transfering sats within nostr? If it is better than ok, let NIP-57 die with time, but else, this may confuse users a lot for little gain.

NIP-60 is always custodial since it's based on cashu whereas NWC does not have to use cashu. I don't think we should limit sending sats on nostr via cashu only.

Egge21M:
> Don't you think it would be possible to somehow use a Nostr extension, NIP-07, to let the mint know you are indeed in possession of the private key to the P2PK proofs that this mint has emitted? The step that want you to encrypt the wallet nsec in the Nostr don't seems super safe and smooth, the beauty of those nips is to use your Nostr profile as your wallet.
> 
> You would just need to sign some sort of secret, that's probably possible already...

NIP-07 does not allow you to sign arbitrary data, only nostr events. Every attempt to change this got shut down quite quickly

pablof7z:
> > Don't you think it would be possible to somehow use a Nostr extension, NIP-07, to let the mint know you are indeed in possession of the private key to the P2PK proofs that this mint has emitted? The step that want you to encrypt the wallet nsec in the Nostr don't seems super safe and smooth, the beauty of those nips is to use your Nostr profile as your wallet.
> > You would just need to sign some sort of secret, that's probably possible already...
> 
> NIP-07 does not allow you to sign arbitrary data, only nostr events. Every attempt to change this got shut down quite quickly

yes, the `privkey` and the different `p2pk` is to get around this limitation

ticruz38:
Correct me if I'm wrong, but cashu and nostr do not share the same encryption. That means cashu and Nostr pubkey cannot be the same.
nutszap are only possible to profile that have previously disclosed their cashu pubkey?

If this above is correct, I would rather have NIP-61 NIP-44 encrypt the C part of the proofs a sender wants to zap, without p2pk lock the proofs at all, at least you can now zap anyone on Nostr without any previous configuration from the recipient

pablof7z:
> Correct me if I'm wrong, but cashu and nostr do not share the same encryption. That means cashu and Nostr pubkey cannot be the same. nutszap are only possible to profile that have previously disclosed their cashu pubkey?

incorrect, they are the same curve

> 
> If this above is correct, I would rather have NIP-61 NIP-44 encrypt the C part of the proofs a sender wants to zap, without p2pk lock the proofs at all, at least you can now zap anyone on Nostr without any previous configuration from the recipient

that would make them unverifiable by others, which is a major benefit over LN zaps.

You can still nutzap users that haven't done any setup, by zapping their nostr pubkey; but they will need to enter the nsec in the client to redeem it -- still completely reasonable, but would in real-world limit the number of clients a user can redeem from.

In my view, the main reason to have a user do the initial nutzap setup (a byproduct of which can be establishing this other pubkey for the `p2pk` lock), is the selection of the mints they are willing to work with. Making this explicit is, I believe, a good flow.

Of course, nothing prevents you from nutzapping someone that has not done their nutzap setup, but clients should not display those nutzaps as valid since perhaps even the receiver user is not even checking for them.

This is all open design space; different clients/users will make different decisions.

ticruz38:
From my experience, p2pk lock a cashu token using my Nostr pubkey and then unlocking it with my Nostr private key just did not work on the mainstream mint, had an error where it said my key was not 32 or 64 Byte long. Probably an error on my end.

As long as using a Nostr pubkey to p2pk an ecash is feasible, I agree with this flow, where not only the recipient but a Nostr client can validate a zap. I just wasn't able to do so using minibit or lnserver mints.

For me the main selling point for this NIP is to enable ecash to unaware user. Most people will not configure their Nostr as an ecash wallet, they will come to ecash because their Nostr identity is ecash compatible and it is just more convenient than paying an LN invoice.

fiatjaf:
Is this ready to merge? I see there are like 5 clients using it already. Is it working?

davidcaseria:
> Is this ready to merge? I see there are like 5 clients using it already. Is it working?

I suggest waiting until a consensus is reached on using deleted events to manage proofs. I am firmly in favor of *not* using deleted events. Also, there needs to be a way to handle reserved proofs; otherwise, double-spend issues will be common.

60.md
```diff
@@ -0,0 +1,205 @@
...
+}
+```
+
+The wallet event is a parameterized replaceable event `kind:37375`.
+
+Tags:
+* `d` - wallet ID.
+* `mint` - Mint(s) this wallet uses -- there MUST be one or more mint tags.
+* `relay` - Relays where the wallet and related events can be found. -- one ore more relays SHOULD be specified. If missing, clients should follow [[NIP-65]].
+* `unit` - Base unit of the wallet (e.g. "sats", "usd", etc).
```
callebtc, line 48: `sat` would be consistent with the denomination of the mint.
```suggestion
* `unit` - Base unit of the wallet (e.g. "sat", "usd", etc).
```

60.md
```diff
@@ -0,0 +1,205 @@
...
+
+## Spending History Event
+Clients SHOULD publish `kind:7376` events to create a transaction history when their balance changes.
+
+```jsonc
+{
+    "kind": 7376,
+    "content": nip44_encrypt([
+        [ "direction", "in" ], // in = received, out = sent
+        [ "amount", "1", "sats" ],
```
callebtc, line 98: Like above, "sat" is consistent with the units of the mint.
```suggestion
        [ "amount", "1", "sat" ],
```

60.md
```diff
@@ -0,0 +1,205 @@
...
+}
+```
+* MUST delete event `event-id-1`
+* SHOULD create a `kind:7376` event to record the spend
+```jsonconc
+{
+    "kind": 7376,
+    "content": nip44_encrypt([
+        [ "direction", "out" ],
+        [ "amount", "4", "sats" ],
```
callebtc, line 177: Like above, "sat" is consistent with the units of the mint.
```suggestion
        [ "amount", "4", "sat" ],
```

pablof7z: my code is full of `s/sats/sat/g` -- this is a welcome addition 😂

fiatjaf:
There have been enough discussions and most people seem to be in consensus about this, and there are like 10 implementations out there, so it's time to merge. That doesn't mean we can't still change things.

PR #1372, opened 2024-07-19
NIP-01: sort events by id after created_at

alexgleason:
See discussion in https://github.com/nbd-wtf/nostr-tools/pull/398

It's a small detail, but can matter a lot for syncing relays, pagination, and ultimately UX.

The order is: `created_by DESC, id ASC`


staab (approved):

mikedilger:
Looks good to me

We probably have software out there that doesn't do this since it wasn't specified before, so beware.

fiatjaf (approved):

PR #1375, opened 2024-07-20
nip19: remove `nrelay`

fiatjaf:
It's better to just use relay hostnames or full URLs, more readable. This encoding just adds unnecessary bloat and confusion and obfuscates things. Is anyone using it? I know @verbiricha's https://nostrrr.com/ has support for it, but now it has transitioned to using hostnames primarily, which is much better.

### DIFF
--- a/19.md
+++ b/19.md
@@ -34,8 +34,8 @@ These are the possible bech32 prefixes with `TLV`:
 
   - `nprofile`: a nostr profile
   - `nevent`: a nostr event
-  - `nrelay`: a nostr relay
   - `naddr`: a nostr _replaceable event_ coordinate
+  - `nrelay`: a nostr relay (deprecated)
 
 These possible standardized `TLV` types are indicated here:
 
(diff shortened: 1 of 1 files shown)


mattn:
If it mean an encoding for group chat that pointed to relay and group (h tag), I want it.

staab:
I use it in a number of place in Coracle, since urls have to be encoded somehow to be included in the paths of another URL. It doesn't strictly have to be interoperable, but it would be nice to be able to link to `njump.me/nrelay1...` for example.

fiatjaf:
Isn't `encodeURIComponent()` good enough? In most cases you can just use the domain name -- if there is no path and the scheme is "https", in other cases you do URL encoding like a normal person. That's what https://nostrrr.com/ does.

fiatjaf:
@mattn NIP-29 has a suggested encoding there, it's `<relay-domain>'<h-tag>`.

staab:
> Isn't encodeURIComponent() good enough?

Uh, yeah, it is. I used to use base64 which broke with some urls, but uri encoding would work better. Anyway, looking over the history of nrelay, it looks like it was so you could register handlers for relay links, which I think is still a good idea.

https://github.com/nostr-protocol/nips/pull/196#issuecomment-1406673890

dluvian:
> Is anyone using it?

It's de- and encoding is implemented in rust-nostr and Voyage renders `nostr:nrelay...` as clickable in case someone posts it.

fiatjaf:
@yukibtc what do you think?

yukibtc:
> @yukibtc what do you think?

Agree about removing it.

mikedilger:
I wouldn't remove it just because nostr has a hell of a lot of projects, some we dont even know of, who might be using it.

I would mark it as deprecated or "not recommended".

AsaiToshiya (approved):
LGTM

PR #1378, opened 2024-07-20
Fix #1377: NIP-46: It's not obvious to the end-user that optional secrets are not temporary

### LINKED ISSUE #1377: NIP-46: It's not obvious to the end-user that optional secrets are not temporary
Consider the following flow:
1. Uncle Normie (UN) initiates a [direct connection](https://github.com/nostr-protocol/nips/blob/master/46.md#direct-connection-initiated-by-remote-signer) using [Amber](https://github.com/greenart7c3/Amber#readme) signer by generating a `bunker://` URI with optional secret
2. UN is aware it's a secret, but not that it's not a *temporary* secret
- UN might think that the entire URI is something like 2FA auth token, just too long to type manually
3. UN transmits it over untrusted proprietary messenger to his PC and pastes into a non-`type="password"` [wnj](https://git.njump.me/wnj) input field and successfully establishes connection after clicking on Amber notification
4. Amber doesn't show the URI anymore, so it appears to UN that the same URI is not designed to be used multiple times, it's probably impossible
5. Stranger uses the same URI to connect their application and sign their message
- Stranger succeeds because UN clicks on the granting notification again because it's not clear what's the normal behavior. UN might think that he received the notification because:
    - it was a requirement of permission for new kind of event he was about to sign
    - or perhaps it's because client recreated a session due to who knows what, IP change for instance, and for some reason signer requires granting this session again
    - or perhaps it's a result of some client/[device bug](https://www.reddit.com/r/GalaxyS23/comments/18arwsc/old_notifications_constantly_get_repeated_new/) UN got used to.

That's unlikely, but perhaps UN has actually read entire NIP-46 multiple times and he carefully reads every mobile notification. But even then he could accidentally click on the malicious notification if it appeared just before attempt to click on some other one.

Ideally I'd never want to see any signer notifications if Stranger tries to establish a new session using the same URI with the same optional secret key. I just don't expect it; I expect a single endless session for a single app I granted for the single device. Perhaps "app" is the wrong concept here, it's actually a session (and "secret" is actually a token).

If it's not possible, at least I expect a signer implementation requirement for explicit warning that the long line of the copied text is not safe to transmit over untrusted channels.

alopatindev:
Not sure whether I phrased it unambiguously enough and correctly in terms of this NIP, feel free to make any changes.

### DIFF
--- a/46.md
+++ b/46.md
@@ -28,7 +28,7 @@ The remote signer would provide a connection token in the form:
 bunker://<remote-user-pubkey>?relay=<wss://relay-to-connect-on>&relay=<wss://another-relay-to-connect-on>&secret=<optional-secret-value>
 ```
 
-This token is pasted into the client by the user and the client then uses the details to connect to the remote signer via the specified relay(s). Optional secret can be used for single successfully established connection only, remote signer SHOULD ignore new attempts to establish connection with old secret key.
+This token is pasted into the client by the user and the client then uses the details to connect to the remote signer via the specified relay(s). Optional secret can be used for single successfully established connection only, remote signer SHOULD ignore new attempts to establish connection with old optional secret.
 
 ### Direct connection initiated by the client
 
--- a/46.md
+++ b/46.md
@@ -28,7 +28,7 @@ The remote signer would provide a connection token in the form:
 bunker://<remote-user-pubkey>?relay=<wss://relay-to-connect-on>&relay=<wss://another-relay-to-connect-on>&secret=<optional-secret-value>
 ```
 
-This token is pasted into the client by the user and the client then uses the details to connect to the remote signer via the specified relay(s).
+This token is pasted into the client by the user and the client then uses the details to connect to the remote signer via the specified relay(s). Optional secret can be used for single successfully established connection only, remote signer SHOULD ignore new attempts to establish connection with old secret key.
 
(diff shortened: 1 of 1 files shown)


46.md
```diff
@@ -28,7 +28,7 @@ The remote signer would provide a connection token in the form:
 bunker://<remote-user-pubkey>?relay=<wss://relay-to-connect-on>&relay=<wss://another-relay-to-connect-on>&secret=<optional-secret-value>
 ```
 
-This token is pasted into the client by the user and the client then uses the details to connect to the remote signer via the specified relay(s).
+This token is pasted into the client by the user and the client then uses the details to connect to the remote signer via the specified relay(s). Optional secret can be used for single successfully established connection only, remote signer SHOULD ignore new attempts to establish connection with old secret key.
```
alexgleason, line 31:
```suggestion
This token is pasted into the client by the user and the client then uses the details to connect to the remote signer via the specified relay(s). Optional secret can be used for single successfully established connection only, remote signer SHOULD ignore new attempts to establish connection with old optional secret.
```

alexgleason (approved):

alexgleason:
I'm going to merge this next time I see it unless someone tells me not to.

brugeman:
ACK

Issue #1380, opened 2024-07-21
Reactions to a website


nikolat:
I would like to react to a website.  
There are some services that can do it, but they are centralized.  
https://s.hatena.ne.jp/  
I would like to achieve it with Nostr.  

Am I allowed to use a `kind 7` event that contains the `r` tag?  
Or should I use an event other than `kind 7`, such as `kind 17`?  

```json
{
  "kind": 7,
  "content": "⭐",
  "tags": [
    ["r", "https://example.com/"]
  ],
  ...other fields
}
```

[NIP-25](https://github.com/nostr-protocol/nips/blob/master/25.md) states "The reaction event SHOULD include `e` and `p` tags."

alexgleason:
You are allowed to do anything you set your heart to.

vitorpamplona:
I prefer to just reuse kind 7 with the `r` tag.

staab:
I think repurposing kind 7 with an r tag makes perfect sense. A new kind might be better though, since you mightvwant to query these reactions without getting all the normal ones.

nikolat:
Thanks.
I think a new kind is better.
I will create a pull request with `kind 17`.

nikolat:
I could make what I wanted.
https://nikolat.github.io/makibishi/
And many people used it.
https://zenn.dev/nikolat/articles/800b19c6b6cc2e#%E8%AC%9D%E8%BE%9E
Thank you!

PR #1381, opened 2024-07-22
NIP-25: reactions to a website

nikolat:
Define `kind 17` event to react to a website.
It includes an `r` tag with the website's URL.

I want to react to a website with a decentralized system.

#1380

### DIFF
--- a/25.md
+++ b/25.md
@@ -52,6 +52,22 @@ func make_like_event(pubkey: String, privkey: String, liked: NostrEvent) -> Nost
 }
 ```
 
+Reactions to a website
+---------------------
+
+If the target of the reaction is a website, the reaction SHOULD be a `kind 17` event and SHOULD include an `r` tag with the website's URL.
+
+```json
+{
+  "kind": 17,
+  "content": "⭐",
+  "tags": [
+    ["r", "https://example.com/"]
+  ],
+  ...other fields
+}
+```
+
 Custom Emoji Reaction
 ---------------------
 
--- a/25.md
+++ b/25.md
@@ -55,7 +55,7 @@ func make_like_event(pubkey: String, privkey: String, liked: NostrEvent) -> Nost
 Reactions to a website
 ---------------------
 
-If the target of the reaction is a website, the reaction SHOULD be a `kind 17` event and SHOULD include an `r` tag with the website's URL.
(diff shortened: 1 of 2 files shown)


fiatjaf (approved):

staab (approved):

25.md
```diff
@@ -52,6 +52,22 @@ func make_like_event(pubkey: String, privkey: String, liked: NostrEvent) -> Nost
...
+---------------------
+
+If the target of the reaction is a website, the reaction SHOULD be a `kind 17` event and SHOULD include an `r` tag with the website's URL.
+
+```json
+{
+  "kind": 17,
+  "content": "⭐",
+  "tags": [
+    ["r", "https://example.com/"]
```
vitorpamplona, line 65: Normalize the URL, like on https://datatracker.ietf.org/doc/html/rfc3986#section-6. 

You might also want to add guidance on fragments (can you like a section of a page?)

Also, keep in mind that there is no `startsWith` filter in Nostr, which means that if you want to get all reactions to a domain name (or path), you have to create individual tags for each.

25.md
```diff
@@ -52,6 +52,22 @@ func make_like_event(pubkey: String, privkey: String, liked: NostrEvent) -> Nost
 }
 ```
 
+Reactions to a website
+---------------------
+
+If the target of the reaction is a website, the reaction SHOULD be a `kind 17` event and SHOULD include an `r` tag with the website's URL.
```
vitorpamplona, line 58: SHOULD -> MUST

nikolat:
Thanks for your review.
I have updated the points you indicated.

nikolat:
Is there anything else that needs to be corrected?

staab:
LGTM, just waiting for two client implementations to merge.

25.md
fiatjaf: Can we add this? And some examples maybe?

penpenpng:
@staab 

> just waiting for two client implementations to merge.

There are two implementations:

- https://nikolat.github.io/makibishi-demo/
- https://penpenpng.github.io/makibishi-component/

And some websites using them:

- https://jiftechnify.github.io/motherfucking-nostr-client/
- https://nostr.compile-error.net/
- https://koteitan.github.io/nostr-post-checker/
- https://bewith.bz/

PR #1384, opened 2024-07-24
NIP-17: Add tag for file attachments

Vap0r1ze:
I'm working on something that uses Nostr as a WebRTC signaling server. I tried to create a solution that is as general as possible, so it can apply to more than just my use case.

My project could use this tag, where a "call invite" could be a NIP-17 Private DM with an `attachment` tag with an SDP offer as a `data:application/sdp;base64,` URL, and a "call answer" could be the corresponding SDP answer.

This could also be used for normal DM file sending without having to put the files in the content itself. It seems like a pretty extendable approach. For example, if a service wanted to add Discord-like embeds to messages: they could send it as something like `data:application/discord-embed+json;base64,`.

I can imagine this tag could be relevant to event kinds other than just NIP-17's kind `14`, let me know if you think this is missing something!

### DIFF
--- a/17.md
+++ b/17.md
@@ -23,6 +23,7 @@ Kind `14` is a chat message. `p` tags identify one or more receivers of the mess
     ["p", "<receiver-2-pubkey>", "<relay-url>"],
     ["e", "<kind-14-id>", "<relay-url>", "reply"] // if this is a reply
     ["subject", "<conversation-title>"],
+    ["attachment", "<attachment-url>", "<file-name>"], // URL to an attachment, small files could be data: URIs
     ...
   ],
   "content": "<message-in-plain-text>",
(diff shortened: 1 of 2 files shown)


staab:
Are you aware of the [`imeta`](https://github.com/nostr-protocol/nips/blob/master/92.md) tag? I think it should cover this use case. A couple other notes:

- Links should generally be included in content, and additional metadata about the link in `imeta` tags. This way, clients that don't support `imeta` aren't missing information when they just show the content.
- Base64 encoded images can get bulky, and so shouldn't be added to events directly. Better is to use a solution like nip 96 or blossom and reference it by url or hash. I'm not sure how big sdp data gets, so it could be fine to include inline.
- Overloading kinds is usually a bad idea, I can't tell here if the semantics of NIP 17 messages make sense here, or if you'd want to create a separate message kind.

Vap0r1ze:
Oh wow I did miss the imeta tag, I guess a new event kind is probably ideal here

PR #1164, opened 2024-04-06
NIP-47 notifications

rolznz:
This allows the wallet service to asynchronously send notifications to connected clients as an alternative to polling.

Example usecases:
- Realtime balance, invoice status, and transaction list updates

Currently two notification types are defined - `payment_received` and `payment_sent`.

This has been implemented in:
- Alby's Nostr Wallet Connect app ("Next" version) https://github.com/getAlby/nostr-wallet-connect-next/pull/183
- Alby JS SDK https://github.com/getAlby/js-sdk/pull/217
- BTCPayServer (Nostr Plugin) https://github.com/Kukks/NNostr/blob/master/NNostr.Client/Protocols/NIP47.cs#L182
- http-nostr https://github.com/getAlby/http-nostr
- [YANA](https://yana.do/)
- [Alby Go](https://github.com/getAlby/go)

There are also some minor copy updates to communicate that NWC is not just for sending payments, but a protocol for connecting apps to lightning wallets.

Currently the notification type is encrypted in the payload, similar to existing responses. The downside is that clients cannot filter notifications - they would get all or none (or would have to be filtered on the wallet service level which is currently outside the spec).

PR #1385, opened 2024-07-24
Add multi_invoice_lookup to NWC / NIP47

blackcoffeexbt:
Performing individual lookups for multiple invoices can be very network heavy and potentially a DOS vector. Adding a `multi_invoice_lookup` method to NIP47 will reduce network usage and reduce the potential for a wallet service / relay DOS attack.

### DIFF
--- a/47.md
+++ b/47.md
@@ -305,6 +305,69 @@ Response:
 Errors:
 - `NOT_FOUND`: The invoice could not be found by the given parameters.
 
+### `multi_lookup_invoice`
+
+Description: Requests information for multiple invoices simultaneously.
+
+Request:
+```jsonc
+{
+    "method": "multi_lookup_invoice",
+    "params": {
+        "invoices": [
+            {"invoice": "lnbc1...", "payment_hash": "abc123..."}, // Optional payment hash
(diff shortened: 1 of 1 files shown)


rolznz:
@blackcoffeexbt have you implemented this somewhere? I'm interested to hear what are you working on!

blackcoffeexbt:
> @blackcoffeexbt have you implemented this somewhere? I'm interested to hear what are you working on!

Not yet but I will. We're adding NWC support to LNbits. With LNbits there's likely to be multiple payments that aren't settled so having multi invoice lookup support would be really useful. I imagine other projects would benefit from it too.

rolznz:
Would notifications possibly solve this? https://github.com/nostr-protocol/nips/pull/1164

vitorpamplona:
Do we need this or has #1164  addressed it?

rolznz:
I think #1164 has resolved it :pray: - and there is an issue in LNBits to implement it: https://github.com/lnbits/lnbits/issues/3187 (which would be a big improvement over polling too)

vitorpamplona:
closing this for now then

PR #1388, opened 2024-07-26
Binary NIP

mikedilger:
This is just here for discussion.  I'm not currently working on it.

### DIFF
--- a/binary.md
+++ b/binary.md
@@ -8,7 +8,7 @@ Nostr uses websockets but only utilizes websocket text messages, using JSON form
 
 This NIP proposes an optional additional websocket binary messaging to improve this situation, and also introduces a number of other protocol level improvements, given that the protocol is new and therefore additional changes wont break existing software.
 
-# Backwards Compatibility
+## Backwards Compatibility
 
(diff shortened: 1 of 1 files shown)


arthurfranca:
If you introduce a new killer feature that requires this NIP to work, it has a great chance of adoption. Yeah I'm talking about fiiile hosting!!

What do you think of expanding the "Binary Event Structure" to support an optional extra data that comes after the event itself as a way to add custom commands through ephemeral events instead of introducing new opcodes? E.g.: An event of kind 20094 tells relay that the extra data is a file chunk of fixed max size defined at a hypothetical binary file hosting NIP. The event may include a challenge tag to prove who is the uploader and even a cashu_token tag the relay can claim as payment for storage.

This way, "ClientEvent" message handling is gonna be implemented by relays that want to enable file hosting and other future features.

mikedilger:
I would prefer that to be a different opcode, otherwise everybody has to buy in.  The whole reason for the feature flags in RelayHello is so people could add this kind of thing.  The relay signals support in those flags, and clients who don't support it just use the regular ClientEvent opcode.  Clients that do support it use whatever new opcode you design.

kristapsk:
There was criticism of Nostr not using binary encoding some year ago by @dr-orlovsky. https://dr.orlovsky.ch/blog/thoughts-on-nostr

PR #1400, opened 2024-07-29
NIP-95 (FROST signature scheme for signing Nostr events)

zmeyer44:
This is a very-very-rough draft for a FROST signature scheme implementation on Nostr via a similar approach to the NIP-46 remote signing flow.

Reference video of FROST implementation: 
https://v.nostr.build/eK0wNZ9PkKaqLx5z.mp4

### DIFF
--- a/95.md
+++ b/95.md
@@ -1,0 +1,135 @@
+# NIP-95 - Nostr Frost Remote Signing
+
+`draft` `optional`
+
+## Rationale
+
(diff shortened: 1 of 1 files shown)


arthurfranca:
I saw the "Coming Soon" section. Can FROST really be used to create a shared nip04/44 key or it is yet to be investigated?

zmeyer44:
> I saw the "Coming Soon" section. Can FROST really be used to create a shared nip04/44 key or it is yet to be investigated?

Yes it can!

PR #1405, opened 2024-07-31
Read status 2

staab:
Another attempt to solve interoperable read statuses. This one combines the timestamp-based approach with the event-id based approach in #933. It uses e tags, not a tags to allow clients to differentiate versions of replaceable events. It uses replaceable events of a few different kinds to reduce storage requirements. It allows for only private read receipts in order to avoid leaking conversation metadata.

### DIFF
--- a/97.md
+++ b/97.md
@@ -16,9 +16,13 @@ This NIP combines both approaches, tracking categories of things a user might wa
 both timestamp and event id. This results in small event payloads, only a few replaceable events,
 and granular seen status for recent events.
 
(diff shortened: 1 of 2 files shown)


staab:
I had this in production for a while, but it was pretty error prone, and resulted in notification badge whack-a-mole with old events that came in late. I think it could still work, but maybe a simpler timestamp-based solution would be better. Removed from coracle in https://github.com/coracle-social/coracle/pull/504/files

PR #1407, opened 2024-08-02
[NIP90 ] - replace NIP04 encryption with NIP44 encryption for encrypted jobs

believethehype:
As there are considerations to fully replace nip04 with nip44 encryption and major libraries used for DVMs like rust-nostr and ndk already support nip44 en/decryption this PR suggests replacing nip04 encryption for "encrypted jobs" with nip44 encryption. This only concerns jobs that are encrypted aka use the "encrypted" tag.

### DIFF
--- a/90.md
+++ b/90.md
@@ -69,7 +69,7 @@ All tags are optional.
 
 ## Encrypted Params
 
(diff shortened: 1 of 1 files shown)


believethehype:
We're gonna do this or nah?

vitorpamplona:
Amethyst is following the change of https://github.com/nostr-protocol/nips/pull/1248#issuecomment-2437731316. 

Right now we test if the encrypted payload is nip04 or nip44 and decrypt it either way. But we still encrypt in NIP-04 to wait for DVMs to catch up. Once DVMs are ready with NIP-44, we can encrypt to nip-44 and disable nip-04 decryption.

believethehype:
My DVMs could do Nip44 but I kept them on Nip04 because Nip90 says Nip04. Thats why I was adding this PR back then to move to Nip44. If clients support both, its good, but for future development, the question is what we want to move on with and the Nip should point both dvm and client devs to it.

PR #1409, opened 2024-08-03
Proposal for mnemonic key encoding of nostr secret keys

mleku:
Pretty straight forward, see the reference implementation [here](https://github.com/mleku/wordstr).

This is a very simple implementation that should be cheap to implement on any device that has enough memory for the word list.

My reference implementation precalculates the place values and performs the decoding from left to right but it could be implemented in reverse, it is just easier when the memory is available.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,112 +89,114 @@ They exist to document what may be implemented by [Nostr](https://github.com/nos
 - [NIP-99: Classified Listings](99.md)
 
 ## Event Kinds
-| kind          | description                | NIP                      |
-| ------------- | -------------------------- | ------------------------ |
-| `0`           | User Metadata              | [01](01.md)              |
-| `1`           | Short Text Note            | [01](01.md)              |
-| `2`           | Recommend Relay            | 01 (deprecated)          |
-| `3`           | Follows                    | [02](02.md)              |
-| `4`           | Encrypted Direct Messages  | [04](04.md)              |
-| `5`           | Event Deletion             | [09](09.md)              |
-| `6`           | Repost                     | [18](18.md)              |
-| `7`           | Reaction                   | [25](25.md)              |
-| `8`           | Badge Award                | [58](58.md)              |
-| `9`           | Group Chat Message         | [29](29.md)              |
-| `10`          | Group Chat Threaded Reply  | [29](29.md)              |
-| `11`          | Group Thread               | [29](29.md)              |
-| `12`          | Group Thread Reply         | [29](29.md)              |
-| `13`          | Seal                       | [59](59.md)              |
-| `14`          | Direct Message             | [17](17.md)              |
-| `16`          | Generic Repost             | [18](18.md)              |
-| `40`          | Channel Creation           | [28](28.md)              |
-| `41`          | Channel Metadata           | [28](28.md)              |
-| `42`          | Channel Message            | [28](28.md)              |
-| `43`          | Channel Hide Message       | [28](28.md)              |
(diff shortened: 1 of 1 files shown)


alexgleason:
Why do we need this instead of just using NIP-06 as it is?

mleku:
> Why do we need this instead of just using NIP-06 as it is?

- vanity keys

- existing keys you don't want to change and deal with losing followers/follow list

- not losing your nsec because you lost all your digital copies of your nsec

fiatjaf:
27 words for a key? That's way too many.

mleku:
the implementation is 24

yes i revised it, that's what the code i refer to uses

it has to exactly reproduce the bits so 24 is minimal

fiatjaf:
I ran your code, it was painful, forced me to download some niche Golang version.

What if we use a bigger word pool? How many would be necessary to reduce that word count to 8?

mleku:
64 bits per cipher that is, that's 19 digit decimal number

there isn't that many words in all the human languages

the math is neater with 12 so you need to find an appropriate further 2048 words that are similarly unambiguous

you can use 12 words with BIP-39 but you can't encode an exact key without 24

i'll dial back the Go version i suppose :)

mleku:
nobody cares about having a way to store an exact key more easily on paper?

fiatjaf:
What if instead of words we used arbitrary combinations of syllables? That could be easy to write and read on paper, and maybe even memorizeable.

vitorpamplona:
I really like the idea of easy to write down nsecs, but I am not sure we need a new encoding scheme.

1Password generates a Emergency Kit with a master key in the form of a dashed sequence of 6 uppercase letters, starting with A3, like this: 



If we just use Bech32, then a 6-letter nsec would be like: 
```
NSEC1-AGZ4Q7-LGWN8V-4K5MFA-4KTYK8-GVY0RA-HV2CPL-2KJQ6M-SP8XPZ-V2PQX7-8ZKF
```

Using 5/6-letter words in multiple lines for ease of typing
```
NSEC1-AGZ4Q-7LGWN-8V4K5M
FA4KT-YK8GV-Y0RAH-V2CPL2
KJQ6M-SP8XP-ZV2PQ-X78ZKF
```

Using 5/4-letter words in multiple lines for even easier typing memorization
```
NSEC1-AGZ4-Q7LG-WN8V-4K5M
FA4KT-YK8G-VY0R-AHV2-CPL2
KJQ6M-SP8X-PZV2-PQX7-8ZKF
```

PS: Bech32 includes a checksum, so it's a little bigger than Base32

mleku:
i don't see the point of making another variant of bech32 when we already have bech32 with checksum and all... you can just add a spacer scheme to that and it's the same thing

THESE ARE NOT EASY TO TYPE OR READ FOR NORMIES

straight decimal numbers would be easier, and they aren't even that much longer than hex

and furthermore you don't need to make any standard spec for putting spaces or hyphens into a bech32 encoding more than a line of change, and it doesn't help anyone

- this 24 word scheme is human readable, 
- it has a check, 
- it's unlikely you are going to mistranscribe it because it's designed that way, 
- it uses the same bitwise values as BIP-39 words, it just doesn't sacrifice a whole word to the check, so you can retain the full 256 bits, 
- the prefix 1 bit ensures that an implementation is simple, you just divide/modulo repeatedly with 2048 and it all falls out as a set of numbers that match a set of words,
- lastly, its check encoding is almost impossible to get confused with BIP-39 keys, i forget how long the checks are but i think it's nearly all of the last two words

PR #1410, opened 2024-08-04
Add shouldAutoLogin flag to NIP-07

arthurfranca:
Reason: user should be able to configure the NIP-07 extension to skip the step of manually having to click a "Login with Extension" button.

### DIFF
--- a/07.md
+++ b/07.md
@@ -24,6 +24,12 @@ async window.nostr.nip44.encrypt(pubkey, plaintext): string // returns ciphertex
 async window.nostr.nip44.decrypt(pubkey, ciphertext): string // takes ciphertext as specified in nip-44
 ```
 
+### Automatic Login
+
+An optional `window.nostr.shouldAutoLogin` boolean may be added.
+
+When set to `true`, clients with login functionality should automatically login using the pubkey returned from `await window.nostr.getPublicKey()` call, i.e, without the need to manually click a "Login with Extension" or similar button.
+
 ### Recommendation to Extension Authors
 To make sure that the `window.nostr` is available to nostr clients on page load, the authors who create Chromium and Firefox extensions should load their scripts by specifying `"run_at": "document_end"` in the extension's manifest.
 


staab:
This seems more like a client UX decision to me than a user preference. Users are generally going to have to grant permission to a new site anyway, so if a client wants a fast login, they should just go ahead and ask.

arthurfranca:
First, let's start from the premise that it isn't "polite" for an app to call `window.nostr.getPublicKey()` on load, because user would see an unexpected extension prompt.

With that said, this flag would enable micro web apps with no complex login flow, which can be a pain.

For example, consider a simple app with a single feature: it shows a kind:1 feed depending on the user pubkey. No login logic in the app besides a global currentPubkey variable (no messing with local storage, no idb nor inter browser tabs communication).

The variable is set if A) user clicks "Login with Extension" or B) `window.nostr.shouldAutoLogin=true` is set by the extension.

If extensions showed an "Auto Login Next Time" marked toggle along with the "Allow microapp.xyz to see your identity (pubkey)", the microapp can remain with zero login logic/state besides that simple global variable. Currently, user would have to click "Login with Extension" every time user opens the app - a bad UX, that is, a simple micro app isn't possible.

The extension's exact flow can vary. It could be a simple "Auto Login" toggle the user can manually press that isn't scoped for specific apps.

lez:
The end goal is exactly the same here as in #1599, the implementation differs only.
Here are my thoughts on that:
- Implementing setting a variable involves some complexity. Either the extension needs to render the value into the script that is executed in page (right now it's a static script) OR the script needs to do a call right after loading, which adds a possibility for race condition (the client might want to read the variable before it's set)

For this reason, I prefer the callback way.

As a note to your proposal: If we expect the extension to set a variable on `window.nostr`, it sould be `nostr.pubkey`, so we save one extra call.

lez:
> > ...possibility for race condition
> 
> Although extensions can optionally access a synchronous storage api (localStorage), this PR would make it a requirement. I agree an async method call would be better, closing in favor of #1599.

I don't fully understand your argument. Extensions already needs access to the localStorage to store keys and configuration. This would not add any requirement in my opinion.

The race condition I was referring to only applies to one particular implementation of setting `window.nostr.pubkey`. Namely, the critical part is when the extension inserts a <script> tag into the page and executes that.
If a static script is injected to create `window.nostr`, it must use an async call to fetch the pubkey from the extension, and during this time `pubkey` would not be accessible to the client.

What I think the optimal solution would be (I know, my opinions changed):
- The extension, before injecting the script into the page, renders the currently active pubkey or `null` *into the script*. So when the script is executed in the page (e.g. when `window.nostr` is created), it already has the `pubkey` set to the proper value. And `getPublicKey` can be left to only be called when the user presses the Login button.

This doesn't need any additional permission for the extension, and there is basically noting to remember because `nostr.pubkey` will hold the ... pubkey. And there is no need to call an async method during first render of the page, which can be uncomfortable as it might block rendering the page / part of the page in some cases.

I really like this approach, I actually did implement such thing in my life, and now I think it's worth the extra programming to make it as good as possible. (The big part of the work is not implementation, it's getting to a social consensus)

lez:
Thanks for clarifying, now I see the problem.
I tested two extensions:
* Alby registers `window.nostr` object before any script on the page runs, but it has a member saying it's not ready at that point. Probably it blocks any async call while being initialized and then it lets async calls pass. This implementation will have problems implementing `window.nostr.pubkey`.
* nos2x-fox I had to wait a few millisecs before `window.nostr` is set. It is `undefined` before, so clients need to implement an initial wait loop for `window.nostr` to appear anyways. Adding `window.nostr.pubkey` in this implementation probably wouldn't break anything, even if it takes a few async calls on the extension side to find out and set the `pubkey` and then send the whole `window.nostr` initialization code in one piece to the client. So when `window.nostr` appears, `window.nostr.pubkey` is already set.
I checked the implementation in nos2x-fox, and to be honest I don't understand why it is initialized right on page load. My firefox might be outdated though (122)

PR #1411, opened 2024-08-06
New NIP to lock user

anurag-l1nt:
This PR defines a new NIP that allows users to block their keys when they suspect or are certain that their keys have been stolen.

Relays implementing this NIP will not accept events from blocked keys, and Clients will hide or visually mark events emitted by blocked keys.

This will mitigate the impact of key theft by preventing identity impersonation.

### DIFF
--- a/100.md
+++ b/100.md
@@ -1,0 +1,78 @@
+
+NIP-100
+======
+
+Lock user
+-------------
+
+`draft` `optional`
+
+### Abstract
+
+In Nostr, users are defined by a pair of public and private keys. These keys are what allow the user to be identified (through the known public key) and ensure that the content is published by the owner of the key pair (through signing with the private key).
+
+This cryptography-based mechanism is very robust and secure, but it is not immune to being attacked. One of the main attack vectors is the theft of the private key itself.
+
+Unlike other protocols like Bitcoin, the theft of a private key does not imply a direct theft of the user's assets, but it can have other implications such as identity theft and consequent fraud attempts. This has a special impact when the stolen key belongs to a highly relevant user.
+
+Although there are currently mechanisms to protect and safeguard our keys, this does not provide complete reliability, and they can always be stolen. To try to mitigate the impact of this problem, this NIP provides a mechanism by which once the key owner suspects or has certainty that the keys have been stolen, they can lock their user and thus prevent (or at least partially prevent) events from being emitted in their name.
+
+### User Lock Event
+
+#### Field `kind`
+
+The `kind:1000` event is defined, which will be sent to the Relays and will indicate that this user should be locked.
+
+#### Field `pubkey`
+
+This is the public key of the event signer and the corresponding public key associated with the stolen private key that wants to be locked.
+
+#### Field `content`
+
+The `content` field may include a description of the reason why the user wants to be locked, although it can be left empty.
+
+##### Example
+
+```json
+{
+  "kind": 1000,
+  "pubkey": <pubkey to be locked>,  
+  "content": "Someone has stolen my password",
+  ...
+}
+```
+### Relay Usage
+
+Relays that implement this NIP, once they receive the `1000` type event, MUST NOT accept any other event issued by the user who has been locked.
+
+Previous events CAN still be queried, but NO operations on them, such as deletion ([NIP-09](./09.md)) or replacement (in the case of replaceable events), will be allowed.
+
+### Client Usage
+
+To lock a user, clients will have an option that allows performing this action. This action will emit a `kind:1000` event signed with the private key that is to be locked.
+
+Clients that implement this NIP MUST check if a `kind:1000` event has been issued. If so, they MUST either hide the events of that user or indicate through some visual mechanism that the user authoring those events has been locked.
+
+Optionally, clients CAN display the `1000` locking event, indicating the reason included in the `content` field. If this field is empty, they can display a generic message. They can also indicate in the user's profile that the user has been locked.
+
+#### Optional
+
+Optionally, clients CAN prevent the sending of events signed by a locked user, as well as decrypt direct encrypted messages using [NIP-04](./04.md) or [NIP-17](./17.md).
+
+For the key thief, this is easily avoidable by using another client or developing a custom one, which is why it is defined as an optional feature in this NIP. However, any difficulty in the illicit use of a key will be welcomed.
+
+
+
+### NIP-59 Usage
+
+The operation of the Relays and C
+lients in the event of receiving an event using [NIP-59](./59.md) will be the same as described in the previous sections, except that the signer of the `seal` event (`kind:13`) will be used to validate whether the event should be accepted (in the case of Relays) or hidden/marked (in the case of Clients).
+
+### Considerations
+
+1. Once a user is locked, **this action is irreversible**, so the keys would remain locked in the Relays and clients that implement this NIP forever.
+2. It is possible that the thief of the key may continue to emit events in the name of the affected user by sending them to Relays that do not implement this NIP. However, these will not be replicated to Relays that do implement them. Eventually, users will tend to use Relays that implement this NIP since they provide more security and reliability.
+3. Although events continue to be retransmitted to Relays that do not implement this NIP, using clients that do implement it provides reliability in the event query.
+4. It is up to the clients to decide how to handle the events of a locked user. They can hide them, mark them in some way, or even allow the user to configure how these events are handled.
+5. The process proposed in this NIP implies that in the event of key theft, the thief can lock the user before the legitimate key owner does. This is assumed since the main problem in the event of Nostr key theft is identity theft, and not so much the ability to continue using them.
+6. As explained for NIP-59, any future NIP developed where the signature is not made by the event author, the control performed by Relays and clients should keep it in mind.
--- a/README.md
+++ b/README.md
@@ -87,6 +87,7 @@ They exist to document what may be implemented by [Nostr](https://github.com/nos
 - [NIP-96: HTTP File Storage Integration](96.md)
 - [NIP-98: HTTP Auth](98.md)
 - [NIP-99: Classified Listings](99.md)
+- [NIP-100: Lock User](100.md)
 
 ## Event Kinds
 
@@ -114,6 +115,7 @@ They exist to document what may be implemented by [Nostr](https://github.com/nos
 | `43`          | Channel Hide Message            | [28](28.md)                            |
 | `44`          | Channel Mute User               | [28](28.md)                            |
 | `818`         | Merge Requests                  | [54](54.md)                            |
+| `1000`        | Lock User                       | [100](100.md)                          |
 | `1021`        | Bid                             | [15](15.md)                            |
 | `1022`        | Bid confirmation                | [15](15.md)                            |
 | `1040`        | OpenTimestamps                  | [03](03.md)                            |


100.md
```diff
@@ -0,0 +1,78 @@
...
+Optionally, clients CAN prevent the sending of events signed by a locked user, as well as decrypt direct encrypted messages using [NIP-04](./04.md) or [NIP-17](./17.md).
+
+For the key thief, this is easily avoidable by using another client or developing a custom one, which is why it is defined as an optional feature in this NIP. However, any difficulty in the illicit use of a key will be welcomed.
+
+
+
+### NIP-59 Usage
+
+The operation of the Relays and C
+lients in the event of receiving an event using [NIP-59](./59.md) will be the same as described in the previous sections, except that the signer of the `seal` event (`kind:13`) will be used to validate whether the event should be accepted (in the case of Relays) or hidden/marked (in the case of Clients).
```
staab, line 69: This is impossible for relays to enforce if they don't have the key to decrypt the wrapper.

staab (commented):
I actually quite like this, but because it relies on wide adoption to be effective, I'd like to see a version that covers key rotation and maybe delegation as well. @pablof7z specced one a while back #829, but it didn't get traction (probably because of the OTS requirement). This is a huge need with nostr, so if this is an area of interest for you I encourage you to push forward on it.

vitorpamplona:
This is a weird one because it can be used in the attacker's favor. If I have somebody's private key, I can create events in the past and in the future and send to the relay before locking the account. I can lock the account in some relays, but keep creating events in the past and future on other relays, generating some confusion of what's locked and what's not. 

That's the main reason I wrote on https://github.com/nostr-protocol/nips/pull/1056 that clients should disqualify anything from that pubkey as soon as one of these events are seen anywhere and followers should repost with a stringified version of the lock/migration event to make sure it is spread everywhere and, most importantly, never gets deleted.

fiatjaf:
Pablo's proposal is still better. I don't think OTS had anything to do with it not getting traction. It's just that it won't be priority until a catastrophic event occurs, but the OTS requirement isn't that hard to solve, we just need some nice service doing it on our behalf. It's in my list of things to do.

mleku:
if the cancel event includes a reasonably long link of recent event IDs it could prove precedence, but the concept of chaining events altogether hasn't seen very much use yet, except as necessary for replies, this is an example of how it could be useful, as proof of causality

vitorpamplona:
OTS isn't hard, but it does create a dependency on a Blockchain lookup service and the need to wait for confirmations from the request to the response time. Requesting OTS events and waiting for confirmations could be better solved in a DVM.

alexgleason:
This almost the same as https://github.com/nostr-protocol/nips/pull/377, just different semantics.

anurag-l1nt:
> I actually quite like this, but because it relies on wide adoption to be effective, I'd like to see a version that covers key rotation and maybe delegation as well. @pablof7z specced one a while back #829, but it didn't get traction (probably because of the OTS requirement). This is a huge need with nostr, so if this is an area of interest for you I encourage you to push forward on it.

I like Pablo's proposal. In fact, when I wrote this NIP, I had thought about including key replacement and some kind of OTS proof, but I ultimately preferred to keep it simpler in this version and didn't work much on those points.

100.md
anurag-l1nt: Yes, you're right. My mistake. I will remove this part.

anurag-l1nt:
> This is a weird one because it can be used in the attacker's favor. If I have somebody's private key, I can create events in the past and in the future and send to the relay before locking the account. I can lock the account in some relays, but keep creating events in the past and future on other relays, generating some confusion of what's locked and what's not, what's real and what's not.

Yes, that can happen. That's why I also wrote that clients should discard (hide or visually mark in some way) the events of a key that has been blocked. I like the idea of clients republishing the events on their relays. This would make propagation faster.

anurag-l1nt:
> Pablo's proposal is still better. I don't think OTS had anything to do with it not getting traction. It's just that it won't be priority until a catastrophic event occurs, but the OTS requirement isn't that hard to solve, we just need some nice service doing it on our behalf. It's in my list of things to do.

Key replacement requires action from the user before the key is compromised. Key blocking could help mitigate issues for those users who haven't taken that action.

I believe the best alternative would be a combination of both: blocking the key with the option to perform a replacement.

anurag-l1nt:
> This almost the same as #377, just different semantics.

Yes, it’s very similar. The main difference with mine is that you propose deleting all previous events, while I only suggest hiding/marking them. I'm not sure which option is more desirable, so possibly the best approach would be to let the user choose what they prefer to do.

100.md
anurag-l1nt: Done in 45b863d2c3b6bd8c2f7808d71f1bdf2fb987f53f

anurag-l1nt: > Yes, you're right. My mistake. I will remove this part.

Done in 45b863d2c3b6bd8c2f7808d71f1bdf2fb987f53f

100.md
```diff
@@ -0,0 +1,70 @@
...
+
+The `kind:1000` event is defined, which will be sent to the Relays and will indicate that this user should be locked.
+
+#### Field `pubkey`
+
+This is the public key of the event signer and the corresponding public key associated with the stolen private key that wants to be locked.
+
+#### Field `content`
+
+The `content` field may include a description of the reason why the user wants to be locked, although it can be left empty.
```
dluvian, line 32: I think `content` should be blank. Clients and relays can't distinguish between your `kind 1000` event and one that has been created by an attacker. Better not give the attacker a chance to publish a malicious message.

100.md
```diff
@@ -0,0 +1,70 @@
...
+
+Previous events CAN still be queried, but NO operations on them, such as deletion ([NIP-09](./09.md)) or replacement (in the case of replaceable events), will be allowed.
+
+### Client Usage
+
+To lock a user, clients will have an option that allows performing this action. This action will emit a `kind:1000` event signed with the private key that is to be locked.
+
+Clients that implement this NIP MUST check if a `kind:1000` event has been issued. If so, they MUST either hide the events of that user or indicate through some visual mechanism that the user authoring those events has been locked.
+
+Optionally, clients CAN display the `1000` locking event, indicating the reason included in the `content` field. If this field is empty, they can display a generic message. They can also indicate in the user's profile that the user has been locked.
```
dluvian, line 56: Same concern as above

dluvian (commented):
I prefer this over https://github.com/nostr-protocol/nips/pull/377. The latter reuses `kind 5 ` which makes it harder to request all my friends who nuked themselves as I have to set `authors` and `#p` in the subscription filter and then filter the responses to make sure that `p tag` == `pubkey`. With this nip I only have to set `authors` in the filter which is much more efficient.

100.md
anurag-l1nt: Yes, I defined it as optional in case the user wanted to indicate a message to their followers, but you are right. If the attacker, for example, locks the key before the real user, and specifies in the `.content` a message such as a new fake public key, it can be a new attack vector.

I change it

100.md
anurag-l1nt: Done

100.md
```diff
@@ -0,0 +1,70 @@
...
+
+The `content` field must be empty to prevent the attacker from publishing a malicious message that could result in a new attack vector.
+
+##### Example
+
+```json
+{
+  "kind": 1000,
+  "pubkey": <pubkey to be locked>,  
+  "content": "Someone has stolen my password",
```
dluvian, line 40:
```suggestion
  "content": "",
```

pablof7z:
> OTS isn't hard, but it does create a dependency on a Blockchain lookup service and the need to wait for confirmations from the request to the response time. Requesting OTS events and waiting for confirmations could be better solved in a DVM.

I already wrote that DVM when I was working on the NIP-41 stuff.

https://github.com/pablof7z/dvm-references/blob/d3b2405b12e967c07bc5345f11b8d23956187e4e/apps/dvms/src/dvms/notary.ts#L4

You send an event id as the input, it resolves when the stamp has been confirmed with the event id of the NIP-03 that was published

dluvian:
We shouldn't use `kind 1000` for this because it's used by users like [snort](https://next.nostrudel.ninja/#/u/84de35e2584d2b144aae823c9ed0b0f3deda09648530b93d1a2a146d1dea9864/about) and [waterblower](https://next.nostrudel.ninja/#/u/6b9da920c4b6ecbf2c12018a7a2d143b4dfdf9878c3beac69e39bb597841cc6e/about).

100.md
```diff
@@ -0,0 +1,70 @@
...
+
+Unlike other protocols like Bitcoin, the theft of a private key does not imply a direct theft of the user's assets, but it can have other implications such as identity theft and consequent fraud attempts. This has a special impact when the stolen key belongs to a highly relevant user.
+
+Although there are currently mechanisms to protect and safeguard our keys, this does not provide complete reliability, and they can always be stolen. To try to mitigate the impact of this problem, this NIP provides a mechanism by which once the key owner suspects or has certainty that the keys have been stolen, they can lock their user and thus prevent (or at least partially prevent) events from being emitted in their name.
+
+### User Lock Event
+
+#### Field `kind`
+
+The `kind:1000` event is defined, which will be sent to the Relays and will indicate that this user should be locked.
```
dluvian, line 24: Random number I checked
```suggestion
The `kind:398` event is defined, which will be sent to the Relays and will indicate that this user should be locked.
```

dluvian:
FYI I implemented this in the latest Voyage release with `kind 398`

anurag-l1nt:
> We shouldn't use `kind 1000` for this because it's already used by users like [snort](https://next.nostrudel.ninja/#/u/84de35e2584d2b144aae823c9ed0b0f3deda09648530b93d1a2a146d1dea9864/about) and [waterblower](https://next.nostrudel.ninja/#/u/6b9da920c4b6ecbf2c12018a7a2d143b4dfdf9878c3beac69e39bb597841cc6e/about).
> 
> I'm using `kind 398`

Used for what? I don't know the implementation for which they are using `kind: 1000`

Thank you!

dluvian:
> Used for what? I don't know the implementation for which they are using `kind: 1000`

I have no idea but there are many `kind:1000` events

@anurag-l1nt Can you replace `kind:1000` with `kind:398` ?

anurag-l1nt:
> > Used for what? I don't know the implementation for which they are using `kind: 1000`
> 
> I have no idea but there are many `kind:1000` events
> 
> @anurag-l1nt Can you replace `kind:1000` with `kind:398` ?

OK. Yes, I change it

anurag-l1nt:
> > > Used for what? I don't know the implementation for which they are using `kind: 1000`
> > 
> > 
> > I have no idea but there are many `kind:1000` events
> > @anurag-l1nt Can you replace `kind:1000` with `kind:398` ?
> 
> OK. Yes, I change it

Done @dluvian

bezysoftware:
Should relays also not allow AUTH for given pubkey? Ideally relays should only serve DMs to authed clients who are part of the conversation, but when a user is locked they probably should never serve these events? Therefore AUTH should fail for a locked account

anurag-l1nt:
> Should relays also not allow AUTH for locked user? Ideally relays should only serve DMs to authed clients who are part of the conversation, but when a user is locked they probably should never serve these events? Therefore AUTH should fail for a locked account

I'm sorry for the delay in responding. You are right. Chaged in 4d58a56b027e8f3cc9a5f6714238ed51e45bee97.

PR #1414, opened 2024-08-07
Amend NIP-52 to include e and p tags to calendar event RSVPs

tyiu:
This PR attempts to address two use cases with calendar event RSVPs:
1. I want a way to have my RSVP not hold true in perpetuity if I know that the calendar event I'm responding to has a chance of changing in a way that makes me not want to attend anymore.
2. The author of the calendar event being responded to may want an easy way of pulling all responses to any event that they authored in one query.


tyiu:
@zmeyer44 @vicariousdrama @staab please review this NIP-52 calendar event RSVP amendment.

52.md
```diff
@@ -178,17 +178,25 @@ This NIP is intentionally not defining who is authorized to attend a calendar ev
 
 This NIP is also intentionally not defining what happens if a calendar event changes after an RSVP is submitted.
 
+The RSVP MUST tag the calendar event it is in response to with an `e` tag on the calendar event's id.
+
+The RSVP MAY tag use an `a` tag to the calendar event's event coordinates to indicate that the RSVP response could hold trueregardless of any future revisions to the calendar event. If the `a` tag is omitted, the author of the RSVP is explicitly indicating that its response holds true for only that specific revision of the calendar event from the `e` tag.
```
tyiu, line 183: I'm actually not sure about this part. Allowing RSVP authors to omit `a` means that it's harder for the calendar event author to pull responses. What do you all think?

staab:
I think we should limit this to adding optional `e` and `a` tags, since making `a` optional and `e` required isn't backwards-compatible. Adding language that clients SHOULD prefer the e tag if available basically gets us the pinning behavior you're looking for if clients start publishing `e` tags.

vicariousdrama:
I concur with @staab in that I'd prefer them to be optional.

52.md
tyiu: Resolving this comment as I've updated the PR to reflect the review comments.

tyiu:
@staab @vicariousdrama Updated.

staab:
LGTM, is this implemented anywhere?

tyiu:
> LGTM, is this implemented anywhere?

No, I'll implement this now in Nostr SDK for Apple Platforms, which Comingle uses. Happy to hold off on merging until it's implemented.

staab:
Ah yeah, good point, that's what I originally meant

AsaiToshiya:
If clients prioritize `e`, there is no need to change `a`.

tyiu:
Thank you all for helping me iterate on this PR. I've pushed a commit to address your feedback.

AsaiToshiya (approved):

PR #1416, opened 2024-08-07
Create fifth parameter in zap tag

patrickReiis:
A normal zap happens when the lightning address is got from the kind 0 of the recipient being zapped. 
A zap split takes into account the lightning addresses of the kind 0s of the pubkeys specified in the zap tags instead of the recipient's pubkey.

If I am zapping someone directly, I am sure of what I'm doing.
If I am zapping someone directly and 5% of the total amount goes to a random pubkey, I may get skeptical and may refuse paying it.

Therefore, one way to prevent such cases is to explain to the user who are the extra pubkeys.
I don't think using the `about` field from the kind 0 is a good idea because it's more intended for general cases, the message in the zap tag is more straightforward, example: "Designer of Snort", "Creator of Amethyst", "Relay X", etc...

I don't think this is a breaking change because it adds a parameter to the end of the array. This feature seem to be done separately by some clients, but if done separately then no interoperability is achieved.

### DIFF
--- a/57.md
+++ b/57.md
@@ -169,12 +169,12 @@ A client can retrieve `zap receipt`s on events and pubkeys using a NIP-01 filter
 
 ### Appendix G: `zap` tag on other events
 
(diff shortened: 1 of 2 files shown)


staab:
I think this makes sense. Nitpick in the PR, I would leave the blank ones off the tags, to indicate that they're optional. Other than that, we just need 2+ clients to implement in order to merge.

patrickReiis:
> Nitpick in the PR, I would leave the blank ones off the tags, to indicate that they're optional  

Ok, I'll do it once 2+ clients implement it, then I'll come back here

patrickReiis:
Ditto implements this: https://gitlab.com/soapbox-pub/ditto/-/merge_requests/468

PR #1418, opened 2024-08-09
rename "parameterized replaceable event" to "addressable event"

fiatjaf:
Because it's impossible to write and takes two minutes to say "parameterized replaceable".

This makes more sense specially given that we already use `naddr` and `a` for referring to these.


AsaiToshiya (approved):
I feel it is an appropriate name.

vitorpamplona:
ack

staab:
NACK, non-parameterized replaceables are addressable too. I tried to fix this in welshman, it's impossible, we will just have to talk faster.

```
import {kinds} from 'nostr-tools'

export const isRegularKind = kinds.isRegularKind
export const isEphemeralKind = kinds.isEphemeralKind
export const isPlainReplaceableKind = kinds.isReplaceableKind
export const isParameterizedReplaceableKind = kinds.isParameterizedReplaceableKind
export const isReplaceableKind = (kind: number) =>
  isPlainReplaceableKind(kind) || isParameterizedReplaceableKind(kind)

```

fiatjaf:
This NACK makes no sense. "addressable" is a better name. Names don't need to contain a perfect an unambiguous definition of things, they just have to be good enough to be said.

10000-19999 are also addressable, but in a different way (in a 2d plane, while 30000-39999 events are addressable in a 3d space), and they already have a good name. Normal events are also addressable, by the way, in a 1d line, and yet we don't use `naddr` or `a` to refer to them (nor we use these for normal replaceables except in some weird cases that are mostly unnecessary to me).

mattn:
What is important is what we have to tell for readers of this spec. It is not the replaceable. not the parameterized. It is possible to point or address to events.

PR #1419, opened 2024-08-09
NIP-74: addressable chat

fiatjaf:
This generalizes [NIP-53](../53.md) live chat events to any other addressable events that specifies a `"relays"` tag.

@v0l @pablof7z

### DIFF
--- a/74.md
+++ b/74.md
@@ -1,0 +1,29 @@
+NIP-74
+======
+
+Addressable Chat
+----------------
+
+`draft` `optional`
+
+Every time an addressable event (kinds 30000-39999) contains a `"relays"` tag with any number of relay URLs specified in it (`["relays", "wss://myrelay.com", "wss://addressable.chat"]`) clients have the option of toggling a chat box about the topic of that event.
+
+Chat messages are expected to be read and published from and to the specified relays, which will _somehow_ provide moderation and consistency.
+
+(Clients can also use different relays other than those specified by the addressable event author, but this would be a power/optional/manual feature until other means of signaling this are invented.)
+
+### Chat Message
+
+Chat messages must have `kind:1311`. Clients MUST include the `a` tag of the activity with a `root` marker. Other Kind-1 tags such as `reply` and `mention` can also be used. An `e` tag containing the specific ID of the current addressable event version being replied to SHOULD also be included.
+
+```json
+{
+  "kind": 1311,
+  "tags": [
+    ["a", "<addressable-event-kind>:<addressable-author-pubkey>:<d-identifier>", "<optional relay url>", "root"],
+    ["e", "<addressable-event-id>", "<optional relay url>", "root"]
+  ],
+  "content": "hello fans of this article",
+  ...
+}
+```


pablof7z:
It should also include the `d` tag of the addressable event, at least for some kinds (e.g. `kind:30818`).

pablof7z:
It should also probably p-tag the author of the addressable event.

v0l:
Whats the use case for this?

pablof7z:
> Whats the use case for this?

A chat where the scope is any other event.

* A chat around a wiki article
* A chat around a livestream
* A chat around a coinjoin round

pablof7z:
- [x] Highlighter.com implements this on wiki entries

https://highlighter.com/pablof7z/cashu-zap-nip/chat

v0l (approved):
LGTM

arthurfranca:
Can you spot any difference between a comment section and a chat?

Despite the "chat" name, this event kind is just a disguised "generic comment", like #1233, but worse. Worse because it is NIP-10 all over again.

As seen on https://highlighter.com/pablof7z/cashu-zap-nip/chat, this event kind is just a regular comment that won't appear in the "feed" because it isn't `kind:1`.


...Oh, the bubbles!

pablof7z:
ok

pablof7z:
sure

pablof7z:
I mean, people do use comments and chats in different ways

pablof7z:
but yeah, that's like, my opinion, bro

pablof7z:
haha

pablof7z:
😀

fiatjaf:
> Can you spot any difference between a comment section and a chat?

I guess a chat is purely sequential, not threaded, and consistency is very important, that's why the events can be simpler but the relay indications must be followed strictly if people want the chat to be usable.

arthurfranca:
It has 'root', 'reply', 'mention' markers, isn't this "threaded", like the main topic can deviate from the OP?

Ended youtube transmissions have both comments and (replayed) chat.
However, you can't comment while the event is live, you can just chat.

In this case it is debatable if chat and comments are to be treated differently or they are just messages of the same kind grouped by created_at ranges, separated by the moment the live finishes.

arthurfranca:
Even if there should exist a kind for generic comments and another kind for chat messages, they should differ just on the kind number.

Imo merging this is the same as electing NIP-10 as the de-facto standard for both, forever.

fiatjaf:
Well, this NIP is just too broad, we can make it better.

I think we can specify just a single `reply` tag thing because in group chats like Telegram and Discord that's the most that happens: you send a message, it goes in the sequential feed of other messages, but it references a previous one for clarity of context. It's definitely not full threading like we want for async events.

I agree with you that for the generic threaded comment there are more complex things involved and https://github.com/nostr-protocol/nips/pull/1233 is better, but just chat is simpler, dumber.

I disagree that chat and comments should work the same with just a different kind number.

Also this NIP just takes the experience of NIP-53 and zap.stream and generalizes it.

arthurfranca:
> [...] you send a message, it goes in the sequential feed of other messages, [...] I disagree that chat and comments should work the same with just a different kind number.

I see, if limited to this narrower scope feed UI, disregarding some discoverability funcionality, it does not need #1233's K/k tags nor its always present r (reply) tag, thus saving bytes. I bless this NIP with my ACK \o/.

Though, a `k` tag set to the root event kind could help with discovering things of a specific kind that are trending chat-wise.

A bit odd that the first implementation has no `relays` tag (on the wiki article). A fallback to the author's NIP-65 read relays feels like a good idea.

PR #1420, opened 2024-08-10
NIP-15 - Replaced NIP-04 (deprecated) with NIP-17: Private Direct Messages

amunrarara:
### Replaced NIP-04 (deprecated) with NIP-17: Private Direct Messages

Cheers,
Any opposition to an outright update of this protocol to replace the use of NIP-04 (deprecated) with NIP-17: Private Direct Messages?

### DIFF
--- a/15.md
+++ b/15.md
@@ -139,7 +139,7 @@ Fields that are not self-explanatory:
 
 ## Checkout events
 
-All checkout events are sent as JSON strings using [NIP-04](04.md).
+All checkout events are sent as JSON strings using [NIP-17](17.md).
 
 The `merchant` and the `customer` can exchange JSON messages that represent different actions. Each `JSON` message `MUST` have a `type` field indicating the what the JSON represents. Possible types:
 
@@ -150,7 +150,7 @@ The `merchant` and the `customer` can exchange JSON messages that represent diff
 | 2            | Merchant | Order Status Update |
 
 ### Step 1: `customer` order (event)
-The below JSON goes in content of [NIP-04](04.md).
+The below JSON goes in content of [NIP-17](17.md).
 
 ```json
 {
@@ -182,7 +182,7 @@ _Open_: is `contact.nostr` required?
 
 Sent back from the merchant for payment. Any payment option is valid that the merchant can check.
 
-The below JSON goes in `content` of [NIP-04](04.md).
+The below JSON goes in `content` of [NIP-17](17.md).
 
 `payment_options`/`type` include:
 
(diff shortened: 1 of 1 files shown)


vitorpamplona:
Since we are changing, how about instead of using the same DM kind, we go for a new special kind for the JSONs that can be transferred on NIP-15? You could use the same NIP-17 mechanism and double-wrap a new kind with the usual seal and wrap. In that way, the users will never see the raw json in the chats of clients that don't support NIP-15.

vitorpamplona:
Actually, does NIP-15 requires the new extreme privacy for DMs? You might not need the whole thing. You could just GiftWrap the new kind directly (so, no Seals... one less event to sign and decrypt).

amunrarara:
> Since we are changing, how about instead of using the same DM kind, we go for a new special kind for the JSONs that can be transferred on NIP-15? You could use the same NIP-17 mechanism and double-wrap a new kind with the usual seal and wrap. In that way, the users will never see the raw json in the chats of clients that don't support NIP-15.

You're totally right; these events have no use outside of NIP-15 enabled clients. It would be great to have each of these events sent with their own kind, instead. 

### Checkout Events

- customer order (event)
- merchant request payment (event)
- merchant verify payment/shipped (event)

Currently, the events above generate JSON that is placed in the content field of a NIP-04 / kind 4 Encrypted Direct Message, sent between customer and merchant. 

> Actually, does NIP-15 requires the new extreme privacy for DMs? 
That would mean the difference between encrypting the rumor, or not encrypting the rumor? Did I understand that correctly?

Having the rumor encrypted could be a good addition to the protocol, I'd think; I'm sure there's a large percentage of transactions that the participants would prefer to have as hyper-private as possible. I personally wouldn't like anyone to have possible knowledge of my participation in a transaction, unless I want them to have that knowledge; a good argument for using extreme privacy with the new kinds. 

### Customer Support 
I still believe these messages could / should be sent via NIP-17, since they're a bit more conversational in nature; though likely related to a specific product, stall, or transaction, it might be nice to continue communications on other non-NIP-15 clients. 

Whaddya think, @vitorpamplona? Awesome thinkin' here, thanks for jumping on the PR. Congrats BTW on the OpenSats LTS grant!!

amunrarara:
Ah I see your logic, that is an excellent solution @vitorpamplona 

I'll update the PR with these recommendations shortly!

amunrarara:
@vitorpamplona Alright, finished an edit that includes the new kinds. Want to give it a review?

I'll pull-and-squash the commits before I reopen the PR again, FYI.

vitorpamplona:
You don't want to use https://github.com/nostr-protocol/nips/pull/1078 to embed the event instead of just pasting a JSON in the message? Then it would be easier for clients to find which parts of the message to render (just look for `nembed1`..)

That would replace all JSONs inside the .content.

PR #1423, opened 2024-08-14
NIP-113: Activity Events

ebrakke:
This nip aims to define activity events on NOSTR. The goal is to enable developers to create applications like Strava or Garmin Connect.

Example protocotol implementation: https://github.com/ebrakke/nostr-activities-sdk

Example client implementation:
https://github.com/ebrakke/nostr-ava
https://nostrava.xyz

View the changes [here](https://github.com/nostr-protocol/nips/blob/c299b749600d8715e6b8004e70bc6100fdb0713a/113.md)


113.md
```diff
@@ -0,0 +1,141 @@
...
+```json
+{
+  "kind": 30101,
+  "content": "JSON.stringify(summary)", 
+  "tags": [
+	 ["d", "UUID generated by client"],
+	 ["r", "<UUID-of-activity>"],
+	 ["t", "general-summary"],
+	 ["g", "OPTIONAL geohash"],
+	 ["unit", "OPTIONAL imperial | metric"],
```
fiatjaf, line 64: should stick to one (metric) and clients do the conversion on their side, makes it easier for everybody

113.md
```diff
@@ -0,0 +1,141 @@
...
+	 ["published_at", "<unix-timestamp>"],
+	 ["updated_at", "<unix-timestamp>"],
+	 ["encrypted"] // Optional, summary content is encrypted
```
fiatjaf, line 67: is this necessary? I feel like encrypted is not necessary and just complicates things. if you feel like encrypted is important, make it a different kind so clients can opt out of fetching those if they want to display all public activities and save everybody's bandwidth.

113.md
```diff
@@ -0,0 +1,141 @@
...
+`t` - A summary MUST include a type which clients can specifically look for. These types must be lowercase with no spaces. e.g. `sports-lib-summary`  
+`g` - Optional geohash of a summary  
+`encrypted` - Optional, if summary is encrypted.  
+`content` - stringified payload of the summary - the structure of which is defined by the type.  
+`unit` - Optional, defines the unit types that are present in a summary. Either `imperial | metric`.
+
+```json
+{
+  "kind": 30101,
+  "content": "JSON.stringify(summary)",
```
fiatjaf, line 58: I don't get what goes here. if it's structured data isn't it better to turn it into non-indexed tags?

fiatjaf:
Aside from the comments above this looks pretty good. Thank you for your work!

113.md
ebrakke: Agreed, I was going back and forth on this one but it seems most activity watches report things in meters anyways.

113.md
ebrakke: Yeah perhaps a different kind for encrypted things might be better. The main dilemma as I see it is a lot of this data is quite personal (gpx files run the risk of doxxing your exact locations), so I wanted the ability for clients to still publish summaries so they're synced across devices, but without revealing the contents of the summary.

I different kind could work as well, and definitely will make it easier for clients to filter them out.

113.md
ebrakke: So it's structured, but it really depends on the contents of the `t` tag for how to interpret it. 
I figured that given the sheer amount of activities that exist, there isn't one structure of a summary that makes sense for every activity type. So the contents would be stringified JSON, and the type will tell the client how to interpret the results.

For instance, one summary type my client publishes right now is `geojson`, where the content is a url to blob storage where the geojson file can be downloaded (since they can be quite large).

But other summaries of activities could be distance metrics for running, or lap metrics for swimming / track, or strokes for rowing. I'm not sure if it's better to have all those values as non-indexed tags, or to just stringify it in the content. Open to suggestions here.

fiatjaf: got it, this is a good solution then -- but maybe the NIP can explain it a little better and also standardize what goes here for each different activity otherwise everything is opaque and impossible to implement.

113.md
fiatjaf: so clients could publish both events, each of a kind, one encrypted for personal use and one for public consumption. when fetching the user personal history you just fetch encrypted stuff, when fetching stuff from others you fetch public.

ebrakke:
@pablof7z my intention with making it one generic summary event is that there are a ton of ways one could actually want their event or events summarized, and I figured it would be annoying to have to create a new kind for every single summary view. 

The two I currently use in my client is
1) The summary stats that the JS library I'm using spits out for gpx files. It's useful for a lot of aerobic activities (like running and cycling), but wouldn't make much sense for an activity like surfing or paddleboarding. 
2) geoJSON summary of the GPX file. Used to display a map of the activity. 

My hope is that others create DVMs that can further summarize the activities. Meta summaries about groups of your activities, summaries about specific routes (like heatmaps). All useful things that strava does in their app. 

So I wanted to leave it flexible at the protocol level to allow for these things. I guess the alternative would be reserving a range of types for summaries that could pop up?

Sebastix:
@ebrakke Just checking in, are you still working on https://nostrava.xyz/ and this NIP?
I have some plans to build a PoC (in https://github.com/nostrver-se/nuxstr) using your NIP proposal. As a fanatic Strava user / cyclist I really would like to have a Nostr native client for sharing my activities ;)

ebrakke:
@Sebastix I haven't touched it in a little while. Been busy / other projects have pulled my interest, but I do plan on getting back to it at some point.
Would love more input / POCs of this nip as well!

113.md
```diff
@@ -0,0 +1,141 @@
...
+  "kind": 30100,
+  "content": "<description-of-activity>",
+  "tags": [
+	 ["title", "Title of Activity"],
+	 ["t", "<activity-type-from-nip>"]
+	 ["d", "UUID generated by client"],
+	 ["r", "url to download activity file"],
+	 ["published_at", "<unix-timestamp>"],
+	 ["updated_at", "<unix-timestamp>"],
+	 ["recorded_at", "<unix-timestamp>"],
```
Sebastix, line 36: Is this the timestamp of when the activity has started to record? Or saved / ended? As the whole recording is in most cases a timedelta.

Sebastix:
@ebrakke FYI while I'm fetching `30100` kind events (from my personal relay wss://nostr.sebastix.dev) I noticed this kind is also used by the Yakihonne client (or was used in the past), for example:

```json
{
  "kind": 30100,
  "id": "2dc8fef6b470b6de26707b5af7abecea9a1bee4f12bb9069c52193cbda9cf3a6",
  "pubkey": "06639a386c9c1014217622ccbcf40908c4f1a0c33e23f8d6d68f4abf655f8f71",
  "created_at": 1698328201,
  "tags": [
    [
      "d",
      "MyFavoriteTopicsInYakihonne"
    ],
    [
      "t",
      "Webdevelopment"
    ],
    [
      "t",
      "Drupal"
    ],
    [
      "t",
      "PHP"
    ],
    [
      "t",
      "Nostr"
    ]
  ],
  "content": "",
  "sig": "7ac91fb2bad98a6ec31c04af0d33f81faf5eff7f80e43adcb101b6c1215fa65138ab05608fea5d332a4ebb9131badd1e2e63b7d014c87742dde139cb0bebbb7e"
}
```
The latest created `30100` event with this structure I could fine is from `1714230783` (2024-04-27) so my guess Yakihonne is not using this event kind anymore since that period.

113.md
```diff
@@ -0,0 +1,141 @@
...
+}
+```
+### Summaries
+`kind:30101` defines an event which is meant to summarize activities, users, places, etc...
+Summaries are intended to capture a wide range things that users may wish to summarize in relation to activities. 
+Users will likely want a summary of a specific event (e.g. milage, time, pace, etc...), a summary for a group of events (e.g. miles run this month), a summary of locations (e.g. heatmaps based on a geohash). 
+
+This NIP does not attempt to describe all the different summaries that users could want, as this list will be ever evolving. Rather this event type should facilitate clients to retrieve relevant summaries for a user.
+#### Summary Metadata
+`r` - An ID that this summary pertains to. It could be an activity ID or a user pubkey
```
Sebastix, line 48: In case of a user pubkey, a p-tag can be used.
In case of a activity ID, is this the ID of the 30100 event? If so, a a-tag can be used to reference to an event.

vitorpamplona:
We should consider renaming NIP-53 from "Live Activities" to "Live Event" or something similar to avoid confusion with this one.

Sebastix:
@vitorpamplona These types of activity are always physical based, so I think it's even better also include that in the name in someway for this NIP. Like 'sport activities'.

Sebastix:
@ebrakke Work in progress, this PoC: https://nostrver.se/nostrides

Sebastix:
@ebrakke nostrava.xyz is giving me 308 http responses. Any chance to fix this? I would like to fetch the GPX files attached to the events which are stored on https://blossom.nostrava.xyz

ebrakke:
@Sebastix the infrastructure that I was using I ended up repurposing for something else since nostrava.xyz wasn't getting used much. I can probably get it back up and running at some point, maybe this weekend.

nostrides looks sweet! Excited to see another alternative out there to strava.

113.md
ebrakke: if it's a replaceable event and you've since replaced it, wouldn't you still want a reference to the most up to date activity ID, rather than the eventID that might be stale?

ebrakke:
@Sebastix I was doing some thinking on this nip and wondering if I was initially trying to do too much with the meta "summary" types. I made a pass to try to vastly simplify the NIP down to the essentials of what you would need for a strava alternative. One that could still evolve to encompass more functionality, but stays simple to implement.  Curious your thoughts on something like this.
https://github.com/ebrakke/nip113/blob/main/NIP.md

PR #1426, opened 2024-08-15
Update 01.md clarification added: BIP0340 tagged hashes are not used

GregTonoski:


fiatjaf:
Why not? They are definitely used.

GregTonoski:
The tags in signature hashes in the BIP-340 are designed to be context specific so "NIP01/challenge" tag in a signature hash would be more appropriate than "BIP0340/challenge" for Nostr.

fiatjaf:
It's too late for that, but I don't see a problem with it. Nostr is defined to use BIP-340, so we inherit the BIP-340 context.

GregTonoski:
No, it's not too late. The change (upgrade) is possible at any moment. It is simple and beneficial for both Bitcoiners and Nostriches (and open source in general) although not critial, I think. BIP-340 should be adapted instead of copied/inherited.  Also, BIP-340 "key prefixing" may be considered for removal.

vitorpamplona:
The text would be better if we described only what we use from the standard instead of the whole standard minus tags, forced even y coordinates, etc

PR #1427, opened 2024-08-15
Add NIP-104: E2EE messaging using MLS

erskingardner:
Take 2! 

This NIP outlines a way to do secure direct and group messaging on Nostr using the MLS protocol.

[Easy to read](https://github.com/erskingardner/nips/blob/50101643f1326bf4e83167f6c415e8f1907819ee/104.md)

### DIFF
--- a/104.md
+++ b/104.md
@@ -1,0 +1,226 @@
+# NIP-104
+
+## E2EE Messaging using the Messaging Layer Security (MLS) Protocol
+
+`draft` `optional`
+
+This NIP standardizes how to use the [MLS Protocol](https://www.rfc-editor.org/rfc/rfc9420.html) with Nostr for efficient and E2EE (end-to-end encrypted) direct and group messaging.
+
+## Context
+
+Originally, one-to-one direct messages (DMs) in Nostr happened via the scheme defined in [NIP-04](04.md). This NIP is not recommended because, while it encrypts the content of the message (provides decent confidentiality), it leaks significant amounts of metadata about the parties involved in the conversation (completely lacks privacy).
+
+With the addition of [NIP-44](44.md), we have an updated encryption scheme that improves confidentiality guarantees but stops short of defining a new scheme for doing direct messages using this encryption scheme. Hence, makes little to no difference to privacy.
+
+Most recently, [NIP-17](17.md) combines [NIP-44](44.md) encryption with [NIP-59](59.md) gift-wrapping to hide the encrypted direct message inside another set of events to ensure that it's impossible to see who is talking to who and when messages passed between the users. This largely solves the metadata leakage problem; while it's still possible to see that a user is receiving gift-wrapped events, you can't tell from whom and what kind of events are within the gift-wrap outer event. This gives some degree of deniability/repudiation but doesn't solve forward secrecy or post compromise security. That is to say, if a user's private key (or the calculated conversation key shared between two users used to encrypt messages) is compromised, the attacker will have full access to all past and future DMs sent between those users.
+
+In addition, neither [NIP-04](04.md) or [NIP-17](17.md) attempt to solve the problem of group messages.
+
+### Why is this important?
+
+Without proper E2EE, Nostr cannot be used as the protocol for secure messaging clients. While clients like Signal do a fantastic job with E2EE, they still rely on centralized servers and as a result can be shut down by a powerful (i.e. state-level) actor. The goal of Nostr is not only to protect against centralized entities censoring you and your communications, but also protect against the ability of a state-level actor to stop these sorts of services from existing in the first place. By replacing centralized servers with decentralized relays, we make it nearly impossible for a centralized actor to completely stop communications between individual users.
+
+### Goals of this NIP
+
+1. Private _and_ Confidential DMs and Group messages
+   1. **Private** means that an observer cannot tell that Alice and Bob are talking to one another, or that Alice is part of a specific group. This necessarily requires protecting metadata.
+   2. **Confidential** means that the contents of conversations can only be viewed by the intended recipients.
+2. Forward secrecy and Post-compromise security
+   1. **Forward secrecy** means that encrypted content in the past remains encrypted even if a key material is leaked.
+   2. **Post compromise security** means that leaking key material doesn't allow an attacker to continue to read messages indefinitely into the future.
+3. Scales efficiently for large groups
+4. Allows for the use of multiple device/clients in a single conversation/group.
+
+### Why MLS?
+
+This scheme adapts the Message Layer Security (MLS) protocol for use with Nostr. You can think of MLS as an evolution of the Signal Protocol. However, it significantly improves the scalability of encryption operations for large group messaging significantly (linear -> log), is built to accommodate federated environments, and also allows for graceful updating of ciphersuites and versions over time. In addition, it's very flexible and agnostic about the message content that is sent.
+
+It's beyond the scope of this NIP to explain the MLS protocol but you can read more about it in it's [Architectural Overview](https://www.ietf.org/archive/id/draft-ietf-mls-architecture-13.html) or the [RFC](https://www.rfc-editor.org/rfc/rfc9420). MLS is on track to become an internet standard under the IETF so the protocol itself is extremely well vetted and researched. This also means there is the potential for cross network messaging interoperability in the future as MLS gains more adoption.
+
+## Core MLS Concepts
+
+From the [MLS Architectural Overview](https://www.ietf.org/archive/id/draft-ietf-mls-architecture-13.html):
+
+> MLS provides a way for clients to form groups within which they can communicate securely. For example, a set of users might use clients on their phones or laptops to join a group and communicate with each other. A group may be as small as two clients (e.g., for simple person to person messaging) or as large as hundreds of thousands. A client that is part of a group is a member of that group. As groups change membership and group or member properties, they advance from one epoch to another and the cryptographic state of the group evolves.
+> 
+> The group is represented as a tree, which represents the members as the leaves of a tree. It is used to efficiently encrypt to subsets of the members. Each member has a state called a LeafNode object holding the client's identity, credentials, and capabilities.
+
+The MLS protocol's job is to manage and evolve the cryptographic state of a group. This includes managing the membership of a group, the cryptographic state of a group (ratchet tree, keys, and encryption/decryption/authentication of messages), and managing the evolution of the group over time.
+
+### Groups 
+
+Groups are created by their first member, who then invites one or more other members. Groups evolve over time in blocks called `Epochs`. New epochs are proposed via one ore more `Proposal` messages and then committed to via a `Commit` message.
+
+### Clients
+
+The device/client pair (e.g. Primal on iOS or Coracle on web) with which a user joins the group is represented as a `LeafNode` in the tree. The terms `Client` and `Member` are interchangeable in this regard. It is not possible to share group state across multiple `Clients`. If a user joins a group from 2 separate devices, their state is separate and they will be tracked as 2 separate members of the group.
+
+### Messages
+
+There are several different types of messages sent within a group. Some of these are control messages that are used to update the group state over time. These include `Welcome`, `Proposal`, and `Commit` messages. Others are the actual messages that are sent between members in a group. These include `Application` messages.
+
+Messages in MLS are "framed". Meaning that they are wrapped in a data structure that includes information about the sender, the epoch, the message index within the epoch and the message content. This framing makes it possible to authenticate and decrypt messages correctly, even if they arrive out of order.
+
+MLS is agnostic to the "content" of the messages that are sent. This is a key feature of MLS that allows for the use of MLS for a wide variety of applications.
+
+MLS is also agnostic to the transport protocol that is used to send messages. Obviously for us, we'll be using websockets, Nostr events and relays.
+
+## The focus of this NIP
+
+This NIP focuses on how to use Nostr to perform the Authentication Service and Delivery Service functions required by the MLS protocol. Most clients will choose to use an MLS implementation to handle keys, ratcheting, group state management, and other aspects of the MLS protocol itself. [OpenMLS](https://github.com/openmls/openmls) is the most actively developed library that implements MLS.
+
+This NIP specifies the following:
+
+1. A standardized way that Nostr clients should [create MLS groups](#creating-groups).
+2. The required format of the MLS [`Credential`](#mls-credentials) that Nostr clients should use to represent a Nostr user in a group.
+3. The structure of [KeyPackage Events](#keypackage-event-and-signing-keys) published to relays that allow Nostr users to be added to a group asynchronously.
+4. The structure of [Group Events](#group-events) published to relays that represent the evolution of a group's state and the contents of the messages sent in the group.
+
+## Creating groups
+
+When creating a new group clients MUST create a random 32-byte hex-encoded group ID value. This identifier can be changed over the life of the group with an MLS re-init `Proposal`. See [Group Messages](#group-events) for more information.
+
+Clients must also ensure that the ciphersuite, capabilities, and extensions they use when creating the group are compatible with those advertised by the users they'd like to invite to the group. They can check this info via the user's published KeyPackage Events.
+
+Changes to an MLS group are affected by first creating one or more `Proposal` events and then committing to a set of proposals in a `Commit` event. These are MLS events, not Nostr events. However, for the group state to properly evolve the Commit events (which represent a specific set of proposals - like adding a new user to the group) must be published to relays for the other group members to see. See [Group Messages](#group-events) for more information.
+
+## MLS Credentials
+
+A `Credential` in MLS is an assertion of who the user is coupled with a signing key. When constructing `Credentials` for MLS, clients MUST use the `BasicCredential` type and set the `identity` value as the 32-byte hex-encoded public key of the user's Nostr identity key.
+
+A `Credential` also has an associated signing key. The initial signing key for a user is included in the KeyPackage event. The signing key MUST be different from the user's Nostr identity key. This signing key will be rotated over time to provide improved post-compromise security.
+
+## KeyPackage Event and Signing Keys
+
+Each user that wishes to be reachable via MLS-based messaging MUST first publish at least one KeyPackage event. The KeyPackage Event is used to authenticate users and create the necessary `Credential` to add members to groups in an asynchronous way. Users can publish multiple KeyPackage Events with different parameters (supporting different ciphersuites or MLS extensions, for example).
+
+KeyPackages SHOULD be used only once. Reuse of KeyPackage Events can lead to replay attacks. In most cases, clients that implement this NIP will manage the creation and rotation of KeyPackage Events.
+
+The signing key (the public key included in the KeyPackage Event) is used for signing within the group that adds a new user via the KeyPackage Event. Therefore, clients implementing this NIP MUST ensure that they retain access to the private key material of the signing key for each group they are a member of.
+
+### Example KeyPackage Event
+
+```json
+  {
+    "id": <id>,
+    "kind": 443,
+    "created_at": <unix timestamp in seconds>,
+    "pubkey": <main identity pubkey>,
+    "content": "",
+    "tags": [
+        ["mls_protocol_version", "1.0"],
+        ["ciphersuite", <MLS CipherSuite ID value e.g. "0x0001">],
+        ["extensions", <An array of MLS Extension ID values e.g. "0x0001, 0x0002">],
+        ["signing_key", <signing key pubkey>],
+        ["client", <client name>, <handler event id>, <optional relay url>],
+        ["relays", <array of relay urls>],
+        ["-"]
+    ],
+    "sig": <signed with main identity key>
+}
+```
+
+- The `content` field is not required.
+- The `mls_protocol_version` tag is required and MUST be the version number of the MLS protocol version being used. For now, this is `1.0`.
+- The `ciphersuite` tag is the value of the MLS ciphersuite that this KeyPackage Event supports. [Read more about ciphersuites in MLS](https://www.rfc-editor.org/rfc/rfc9420.html#name-mls-cipher-suites).
+- The `extensions` tag is an array of MLS extension IDs that this KeyPackage Event supports. [Read more about MLS extensions](https://www.rfc-editor.org/rfc/rfc9420.html#name-extensions).
+- The `signing_key` tag is the signing key public key.
+- The `client` tag helps other clients manage the user experience when they receive group invites but don't have access to the signing key.
+- The `relays` tag identifies each of the relays that the client will attempt to publish this KeyPackage event. This allows for deletion of KeyPackage Events at a later date.
+- The `-` tag is optional, but can be used to ensure that KeyPackage Events are only published by their authenticated author. Read more in [NIP-70](70.md)
+ 
+### Deleting KeyPackage Events
+
+Clients MUST delete the KeyPackage Event on all the listed relays any time they successfully process a group request event for a given KeyPackage Event. Clients MAY also create a new KeyPackage Event at the same time.
+
+If clients cannot process a Welcome message (e.g. because the signing key was generated on another client), clients MUST not delete the KeyPackage Event and SHOULD show a human-understandable error to the user.
+
+### Rotating Signing Keys
+
+Clients MUST regularly rotate the user's signing key in each group that they are a part of. The more often the signing key is rotated the stronger the post-compromise security. This rotation is done via `Proposal` and `Commit` events and broadcast to the group via a Group Event. [Read more about forward secrecy and post-compromise security inherent in MLS](https://www.rfc-editor.org/rfc/rfc9420.html#name-forward-secrecy-and-post-co).
+
+### KeyPackage Relays Event
+
+A `kind: 10051` event indicates the relays that a user will publish their KeyPackage Events to. The event MUST include a list of relay tags with relay URIs.
+
+```json
+{
+  "kind": 10051,
+  "tags": [
+    ["relay", "wss://inbox.nostr.wine"],
+    ["relay", "wss://myrelay.nostr1.com"],
+  ],
+  "content": "",
+  //...other fields
+}
+```
+
+### Welcome Event
+
+When a new user is added to a group via an MLS `Commit` message. The member who sends the `Commit` message to the group is responsible for sending the user being added to the group a Welcome Event. This Welcome Event is sent to the user as a [NIP-59](59.md) gift-wrapped event. The Welcome Event gives the new member the context they need to join the group and start sending messages.
+
+Clients creating the Welcome Event SHOULD wait until they have received acknowledgement from relays that their Group Event with the `Commit` has been received before publishing the Welcome Event.
+
+```json
+{
+   "id": <id>,
+   "kind": 444,
+   "created_at": <unix timestamp in seconds>,
+   "pubkey": <nostr identity pubkey of sender>,
+   "content": <serialized MLSMessage object>,
+   "tags": [
+      ["relays", <array of relay urls>],
+   ],
+   "sig": <NOT SIGNED>
+}
+```
+
+- The `content` field is required and is a serialized MLSMessage object containing the MLS `Welcome` object.
+- The `relays` tag is required and is a list of relays clients should query for Group Events.
+
+Welcome Events are then sealed and gift-wrapped as detailed in [NIP-59](59.md) before being published. Like all events that are sealed and gift-wrapped, `kind: 444` events MUST never be signed. This ensures that if they were ever leaked they would not be publishable to relays.
+## Group Events
+
+Group Events are all the messages that are sent within a group. This includes all "control" events that update the shared group state over time (`Proposal`, `Commit`) and messages sent between members of the group (`Application` messages).
+
+Group Events are published using an ephemeral Nostr keypair to obfuscate the number and identity of group participants. Clients MUST use a new Nostr keypair for each Group Event they publish.
+
+```json
+{
+   "id": <id>,
+   "kind": 445,
+   "created_at": <unix timestamp in seconds>,
+   "pubkey": <ephemeral sender pubkey>,
+   "content": <serialized MLSMessage object>,
+   "tags": [
+      ["h", <group id>]
+   ],
+   "sig": <signed with ephemeral sender key>
+}
+```
+- The `content` field is a [tls-style](https://www.rfc-editor.org/rfc/rfc9420.html#name-the-message-mls-media-type) serialized [`MLSMessage`](https://www.rfc-editor.org/rfc/rfc9420.html#section-6-4) object.
+- The `pubkey` is the hex-encoded public key of the ephemeral sender.
+- The `h` tag is the group ID value
+
+### Application Messages
+
+Application messages are the messages that are sent within the group by members. These are contained within the `MLSMessage` object. The format of these messages should be unsigned Nostr events of the appropriate kind. For example, if a user sends a text note to the group, it would be a `kind: 1` event. If the user reacts to a message, it would be a `kind: 7` event.
+
+This means that once the application message has been deserialized, clients can store those events and treat them as any other Nostr event, effectively creating a private Nostr feed of the group's activity and taking advantage of all the features of Nostr.
+
+The Nostr event MUST use the member's Nostr identity key for the `pubkey` field.
+
+These Nostr events MUST remain **unsigned** to ensure that if they were to leak to relays they would not be published publicly.
+
+### `Commit` Message race conditions
+
+The MLS protocol is resilient to almost all messages arriving out of order. However, the order of `Commit` messages is important for the group state to move forward from one epoch to the next correctly. Given Nostr's nature as a decentralized network, it is possible for a client to receive 2 or more `Commit` messages all attempting to update to a new epoch at the same time.
+
+Clients sending commit messages MUST wait until they receive acknowledgement from at least one relay that their Group Message Event with the `Commit` has been received before applying the commit to their own group state.
+
+If a client receives 2 or more `Commit` messages attempting to change same epoch, they MUST apply only one of the `Commit` messages they receive, determined by the following:
+
+1. Using the `created_at` timestamp on the kind `445` event. The `Commit` with the lowest value for `created_at` is the message to be applied. The other `Commit` message is discarded.
+2. If the `created_at` timestamp is the same for two or more `Commit` messages, the `Commit` message with the lowest value for `id` field is the message to be applied.
+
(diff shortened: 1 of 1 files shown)


104.md
```diff
@@ -0,0 +1,226 @@
...
+
+The MLS protocol's job is to manage and evolve the cryptographic state of a group. This includes managing the membership of a group, the cryptographic state of a group (ratchet tree, keys, and encryption/decryption/authentication of messages), and managing the evolution of the group over time.
+
+### Groups 
+
+Groups are created by their first member, who then invites one or more other members. Groups evolve over time in blocks called `Epochs`. New epochs are proposed via one ore more `Proposal` messages and then committed to via a `Commit` message.
+
+### Clients
+
+The device/client pair (e.g. Primal on iOS or Coracle on web) with which a user joins the group is represented as a `LeafNode` in the tree. The terms `Client` and `Member` are interchangeable in this regard. It is not possible to share group state across multiple `Clients`. If a user joins a group from 2 separate devices, their state is separate and they will be tracked as 2 separate members of the group.
```
abhay-raizada, line 56: I may be misunderstanding but can't clients share keypackages with the user or other clients? Making it interoperable within the same device?

104.md
```diff
@@ -0,0 +1,226 @@
...
+
+### `Commit` Message race conditions
+
+The MLS protocol is resilient to almost all messages arriving out of order. However, the order of `Commit` messages is important for the group state to move forward from one epoch to the next correctly. Given Nostr's nature as a decentralized network, it is possible for a client to receive 2 or more `Commit` messages all attempting to update to a new epoch at the same time.
+
+Clients sending commit messages MUST wait until they receive acknowledgement from at least one relay that their Group Message Event with the `Commit` has been received before applying the commit to their own group state.
+
+If a client receives 2 or more `Commit` messages attempting to change same epoch, they MUST apply only one of the `Commit` messages they receive, determined by the following:
+
+1. Using the `created_at` timestamp on the kind `445` event. The `Commit` with the lowest value for `created_at` is the message to be applied. The other `Commit` message is discarded.
```
abhay-raizada, line 223: Malicious clients can chose to send lower created_at values on purpose, making sure their commits are the ones that are used. 
 Maybe we can use NIP-03 (timestamp attestation) to make sure these timestamps are valid?

104.md
```diff
@@ -0,0 +1,226 @@
...
+- The `ciphersuite` tag is the value of the MLS ciphersuite that this KeyPackage Event supports. [Read more about ciphersuites in MLS](https://www.rfc-editor.org/rfc/rfc9420.html#name-mls-cipher-suites).
+- The `extensions` tag is an array of MLS extension IDs that this KeyPackage Event supports. [Read more about MLS extensions](https://www.rfc-editor.org/rfc/rfc9420.html#name-extensions).
+- The `signing_key` tag is the signing key public key.
+- The `client` tag helps other clients manage the user experience when they receive group invites but don't have access to the signing key.
+- The `relays` tag identifies each of the relays that the client will attempt to publish this KeyPackage event. This allows for deletion of KeyPackage Events at a later date.
+- The `-` tag is optional, but can be used to ensure that KeyPackage Events are only published by their authenticated author. Read more in [NIP-70](70.md)
+ 
+### Deleting KeyPackage Events
+
+Clients MUST delete the KeyPackage Event on all the listed relays any time they successfully process a group request event for a given KeyPackage Event. Clients MAY also create a new KeyPackage Event at the same time.
```
ok300, line 134: The RFC says[^1] that
> In order to avoid replay attacks and provide forward secrecy for messages sent using the initial keying material, KeyPackages are intended to be used only once.

However, reliable message deletion on Nostr is not possible. Especially when using a 3rd party relay, even moreso when using a bunch of relays.

Doesn't this mean MLS on Nostr can't guarantee forward secrecy? Or am I missing something?

[^1]: https://www.ietf.org/archive/id/draft-ietf-mls-architecture-13.html#name-key-storage-and-retrieval

104.md
```diff
@@ -0,0 +1,226 @@
...
+In addition, neither [NIP-04](04.md) or [NIP-17](17.md) attempt to solve the problem of group messages.
+
+### Why is this important?
+
+Without proper E2EE, Nostr cannot be used as the protocol for secure messaging clients. While clients like Signal do a fantastic job with E2EE, they still rely on centralized servers and as a result can be shut down by a powerful (i.e. state-level) actor. The goal of Nostr is not only to protect against centralized entities censoring you and your communications, but also protect against the ability of a state-level actor to stop these sorts of services from existing in the first place. By replacing centralized servers with decentralized relays, we make it nearly impossible for a centralized actor to completely stop communications between individual users.
+
+### Goals of this NIP
+
+1. Private _and_ Confidential DMs and Group messages
+   1. **Private** means that an observer cannot tell that Alice and Bob are talking to one another, or that Alice is part of a specific group. This necessarily requires protecting metadata.
```
ok300, line 26: Is this still the case for a malicious or compromised Delivery Service (relay)?

The RFC says that
> the DS may be able to determine group membership

rabble:
Is this the code implementing it?

https://github.com/erskingardner/openmls_nostr_crypto

 I feel like this is only part of it.

reyamir:
I'm working on implement this NIP for https://github.com/lumehq/coop
let see how it go 😄

104.md
erskingardner: So, I had been operating under the assumption that this private key couldn't/shouldn't be shared between devices but given these key packages should only be used once there is the option of encrypting the private key (to yourself) in the content field using NIP-44 encryption. This would mean the following: 

- Any client could respond to a group welcome message (the invite to a new group) provided they had access to your nostr identity private key.
- The client that responded to the welcome message would be the client that would be added to the group. Any other clients that you wanted to have access to the group would have to be added separately.
- If your identity key leaks then the private key to these key package keys also leaks. Given these events are meant to be used once and deleted, I don't think this is actually that big a deal tbh. 

Would love thoughts from others on this.

104.md
erskingardner: I think that the time it takes to get the commit OTS attested will cause more troubles than it solves. 

Fundamentally, a malicious client that is part of the group was added to the group by someone in the group. Anyone in the group can evict that malicious client as soon as they notice bad behavior. I think this is probably not something the spec can solve and rather is the domain of the applications that are implementing this spec to decide.

104.md
erskingardner: This is why users or the clients implementing this should select the relays that they publish key packages to carefully based on advertised support of event deletion and (optionally) support of the `-` tag to ensure that the even isn't rebroadcast to other relays.

Another mitigation strategy for this (that resides at the client/application level) is that clients could require some level of freshness for the Key Package events that they use and refuse to use Key Package events that are older than a certain time in the past.

104.md
erskingardner: This is why we publish all the Group events from ephemeral keys. A sophisticated malicious relay could use IP addresses to triangulate a give user but if users are using a VPN then this is mitigated (and this is the case across all of nostr).

erskingardner:
> I feel like this is only part of it.

@rabble yes, this is definitely only part of it. I started that library and intend for it to be complete but then got sucked into updating upstream dependencies to support Schnorr signatures over `secp256k1`. Ultimately though, it's possible for clients to be built using the other ciphersuites that are already supported by OpenMLS though so I wanted to get the spec out here for review.

The future benefit of that library you liked above will be that nostr clients will have fewer dependencies on underlying crypto primitives, that's basically it.

erskingardner:
> I'm working on implement this NIP for https://github.com/lumehq/coop let see how it go 😄

🤙 Definitely feel free to DM me if you run into anything or have questions.

104.md
```diff
@@ -0,0 +1,226 @@
...
+   "created_at": <unix timestamp in seconds>,
+   "pubkey": <ephemeral sender pubkey>,
+   "content": <serialized MLSMessage object>,
+   "tags": [
+      ["h", <group id>]
+   ],
+   "sig": <signed with ephemeral sender key>
+}
+```
+- The `content` field is a [tls-style](https://www.rfc-editor.org/rfc/rfc9420.html#name-the-message-mls-media-type) serialized [`MLSMessage`](https://www.rfc-editor.org/rfc/rfc9420.html#section-6-4) object which is then encrypted according to [NIP-44](44.md) but using the MLS [`exporter_secret`](https://www.rfc-editor.org/rfc/rfc9420.html#section-8.5) and the group ID values to calulate the `conversation key` value. The rest of the NIP-44 encryption process is the same. The `exporter_secret` value should be generated with a 32-byte length and labeled `nostr`. This `exporter_secret` value is rotated on each new epoch in the group. Clients should generate a new 32-byte value each time they process a valid `Commit` message.
```
erskingardner, line 201: Calling out a small update here. The serialized `MLSMessage` object isn't always encrypted and even when encrypted would leak header information. 

To remedy this, I'm proposing that we use NIP-44 encryption for the content field with a twist. Since all members of the group need to be able to decrypt the message content we need to use a shared secret value that they all know. 

MLS provides a mechanism to export a secret key from the current state of the ratchet tree. This `exporter_secret` value is labeled and can be generated to be any length. We'll generate a 32-byte value and use it along with the group ID value (also 32 bytes) to generate a value that we'll use for the `conversation key` in NIP-44's encryption. This will allow all participants to encrypt and decrypt these content fields and get to the messages. 

The `exporter_secret` is rotated on each group epoch so the key will be rotated after each `Commit`, providing forward and post-compromise security for this content field encryption. In other words, even if an attacker got the `exporter_secret` value, they would only be able to decrypt group messages from a single epoch.

paulmillr:
Farcaster implemented Signal protocol in DMs. Then they've dropped it because it was too complex to support on different devices (old / new / different libraries).

It would be great to have some kind of demo app first. I suspect it'll be quite complicated.

kngako:
Nice, how can one extend the Group creation to use FROST (Maybe even have the FROST signers as group admins)?

erskingardner:
> Nice, how can one extend the Group creation to use FROST (Maybe even have the FROST signers as group admins)?

I've had a bit of a look at FROST but haven't really thought about how it might apply. For now, it's completely out of scope.

erskingardner:
> It would be great to have some kind of demo app first. I suspect it'll be quite complicated.

@paulmillr working on it :)

Sjors:
Some very initial thoughts on a potential choice of cipher suite: https://njump.me/nevent1qvzqqqqqqypzpp59a0hkv5ecm45nrckvmu7pnk0sukssvly33u3wwzquy4v037hcqyg8wumn8ghj7mn0wd68ytnddakj7qgwwaehxw309ahx7uewd3hkctcqyprq4w327fynj7wa6lha6pktmer8c0979ds0sc89cgsmh5af4pet60p3dcc

104.md
```diff
@@ -0,0 +1,228 @@
...
+  "content": "",
+  //...other fields
+}
+```
+
+### Welcome Event
+
+When a new user is added to a group via an MLS `Commit` message. The member who sends the `Commit` message to the group is responsible for sending the user being added to the group a Welcome Event. This Welcome Event is sent to the user as a [NIP-59](59.md) gift-wrapped event. The Welcome Event gives the new member the context they need to join the group and start sending messages.
+
+Clients creating the Welcome Event SHOULD wait until they have received acknowledgement from relays that their Group Event with the `Commit` has been received before publishing the Welcome Event.
```
EthnTuttle, line 164: What does an "acknowledgement from relays" look like?

104.md
```diff
@@ -0,0 +1,228 @@
...
+- The `pubkey` is the hex-encoded public key of the ephemeral sender.
+- The `h` tag is the group ID value
+
+### Application Messages
+
+Application messages are the messages that are sent within the group by members. These are contained within the `MLSMessage` object. The format of these messages should be unsigned Nostr events of the appropriate kind. For example, if a user sends a text note to the group, it would be a `kind: 1` event. If the user reacts to a message, it would be a `kind: 7` event.
+
+This means that once the application message has been deserialized, clients can store those events and treat them as any other Nostr event, effectively creating a private Nostr feed of the group's activity and taking advantage of all the features of Nostr.
```
EthnTuttle, line 211: This "private Nostr feed" makes me think of the local relay ideas. And the various *box models that have surfaced.

104.md
EthnTuttle: Is there someway to have an "authority" identified for these Commits? like a group admin or similar.

EthnTuttle:
> Nice, how can one extend the Group creation to use FROST (Maybe even have the FROST signers as group admins)?

This makes me wonder how Signal does group admin stuff. They have a nice features to tweak within groups. For these different MLS events, is it expected that all participants are publishing all event types for the group?

104.md
jmanero: > I think that the time it takes to get the commit OTS attested will cause more troubles than it solves.
> 
> Fundamentally, a malicious client that is part of the group was added to the group by someone in the group. Anyone in the group can evict that malicious client as soon as they notice bad behavior. I think this is probably not something the spec can solve and rather is the domain of the applications that are implementing this spec to decide.

Could a malicious client use a similar technique to preempt attempts to remove it?

paulmillr:
Signal groups is just a bunch of 1-to-1 messages. Group settings are stored in the server in enclaves.

104.md
erskingardner: Just that you get an ack from a relay when sending the event. Same as any other event that you publish.

github-tijlxyz:
This looks super great, I have been thinking about something like this for a while and will maybe try building a client for it. Any update on the more specific example process/app?

erskingardner:
> This looks super great, I have been thinking about something like this for a while and will maybe try building a client for it. Any update on the more specific example process/app?

I'm working on a reference client right now @github-tijlxyz I'll let everyone know here and on nostr when it's ready. :)

erskingardner:
This is ready for review. I've made some updates based on what I'm doing in White Noise and while I don't have fully completed libraries, I would love to start working with anyone that is interested in implementing this to help find more places where we need supporting library code. 

Obviously, I'd love to merge this NIP once we get some more eyes on it. 

@fiatjaf @staab @vitorpamplona @pablof7z @paulmillr @Sjors @v0l @hzrd149

paulmillr:
@erskingardner do you have a fully working MLS over nostr implementation in place?

erskingardner:
Yes. https://github.com/erskingardner/whitenoise

I'm still having some platform specific issues (nothing related to MLS) before it's ready for distribution but it works end to end.

erskingardner:
> A brief description of how this is solved would be useful. I know MLS is supposed to solve it, but an ELI5 version would  nonetheless be helpful. Like, what exactly happens after user's messaging key is leaked. Or if their private key is leaked.

@paulmillr do you think it's worth having a section in the NIP about the threat model and various compromises possible? Or are you just looking for an explanation here? MLS has a specific section I can link to that goes into lots of detail on this for MLS specifically. There are some nostr specific things I could add though.

paulmillr:
MLS spec is hard to read. I think a simple section describing how things improve over the status-quo would be useful, including compromises. Best to keep it concise.

erskingardner:
Added a section on Security considerations with some basic details and links to MLS docs on the more in-depth stuff.

alexfj0:
I mean if I should try to run whitenoise now, because I think there are some problems like a release tag missing for https://github.com/erskingardner/nostr-openmls which is specified here https://github.com/erskingardner/whitenoise/blob/master/src-tauri/Cargo.toml and some of the dependencies there point to local directories.

erskingardner:
Ah yeah, good catch. I've been making changes to those libraries locally. Let me try and get a buildable version done tomorrow. Fair warning. I've been building locally on a mac and things seem to be working well but once I try and run the app on any other platform I get a lot of strange behavior which I'm working on fixing now.

erskingardner:
@alexfj0 you should be able to build the app now, I've updated the dependencies to point to GH. 

To test out groups and messages, it's easiest to do the following: 

1. Sign in with an account that is following a few users (hopefully at least one other account that is yours). 
2. Go to settings, sign in with the second account.
3. Publish a key package event from the settings page.
4. Switch to the first user. 
5. Go to the chats screen, click the plus icon at the top right, select your other user and invite them to start a chat. 
6. Once the group is created, you can try sending a few messages. 
7. Then switch to the other user. You should see an invite waiting for you.
8. Accept that invite and then you can start replying to the other user.

Let me know if you have any troubles.

water783:
I’m encountering some issues while implementing NIP-104:

- I just want to confirm if the content of the kind 443 event is an encodedKeyPackage string? I couldn’t find this in the spec, but the whitenoise implementation seems to do it this way.

- Does the kind 445 group event need to include the current epoch? Otherwise, do we need to iterate through the local group keys to try decrypting?

ok300:
The "easy to read" link in the PR description points to an older commit.

A better link, always pointing to the latest version, is
```
https://github.com/erskingardner/nips/blob/master/104.md
```

Can you please update it?

ok300:
In the "Example KeyPackage Event" snippet, the `pubkey` and `sig` comments refer to the "main identity key".

However, a few paragraphs above warn that:
> KeyPackages include a signing key that is used for signing MLS messages within a group. This signing key MUST not be the same as the user's Nostr identity key.

Is the "main identity key" (from the snippet comments) the same as the user's Nostr identity key? I assume not, but they sound very similar. Could you add a bullet point under that snippet explaining what the "main identity key" is?

erskingardner:
> I just want to confirm if the content of the kind 443 event is an encodedKeyPackage string? I couldn’t find this in the spec, but the whitenoise implementation seems to do it this way.

The content of the kind 443 is a hex encoded MLS welcome message object (from OpenMLS). You can see everything I'm doing in the [create_group](https://github.com/erskingardner/whitenoise/blob/e290773802bf780239cf514a835fa5ab84afd7c3/src-tauri/src/commands/groups.rs#L154) command on White Noise. You'll also want to look at the [nostr-openmls](https://github.com/erskingardner/nostr-openmls/blob/94e4612963b4c3b1f1f0332c7142620ee771ddc3/src/groups.rs#L87) library for the actual call to the OpenMLS methods.

> Does the kind 445 group event need to include the current epoch? Otherwise, do we need to iterate through the local group keys to try decrypting?

No it doesn't need to include the epoch in the actual event or tags. That's included in the `MlsMessageOut` object you get from the OpenMLS library.

> Is the "main identity key" (from the snippet comments) the same as the user's Nostr identity key? I assume not, but they sound very similar. Could you add a bullet point under that snippet explaining what the "main identity key" is?

The "main identity key" is the nostr pubkey (npub in hex format). The "signing key" is the key that you use to generate the `KeyPackageBundle`. You can see how that is getting done in [nostr-openmls](https://github.com/erskingardner/nostr-openmls/blob/94e4612963b4c3b1f1f0332c7142620ee771ddc3/src/key_packages.rs#L57).

I can add a bit more context to the spec for that.

water783:
> No it doesn't need to include the epoch in the actual event or tags. That's included in the MlsMessageOut object you get from the OpenMLS library.

Sorry, maybe I’m wrong, but shouldn’t we decrypt the event first before getting the `MlsMessageOut` object? Nip44 decrypting requires knowing the epoch to obtain the corresponding group key. I’m not sure if I understand this correctly

erskingardner:
Ah ok, I misunderstood what you were trying to do. Yes, you do need the epoch for decrypting. In the nostr-openmls library there is a method `export_secret_as_hex_secret_key_and_epoch` that gives you the secret from the MLS group and the current epoch. In White Noise I'm storing that in the secrets store for the app and then using it when needed. 

I've not added epoch rotation (processing commits) to the app yet (there was a bug in the underlying OpenMLS library I'd started to work on fixing in addition to the more standard group changes I'll be implementing soon) but my idea was this: 

1. When a new message arrives, try to decrypt using the current exporter secret and epoch. 
2. If that fails then try to use the past  exporter secret and epoch - this should be limited to some number of epochs in the past though to ensure that we're not keeping old secrets around indefinitely.
3. If that fails, then it means we should have a commit message that rolls the group into the next epoch. If we haven't seen that we need to wait and store this message so we can decrypt later.

Any change to the group (add/remove member, change of signing key of any member, change of permissions, etc.) will require a proposal and commit message to be created.

We really don't want to put the epoch as a value in the tags of the 445 event because that will give away metadata that tells you roughly how old the group is.

erskingardner:
@water783 are you using OpenMLS and my nostr-openmls library or are you doing something else?

water783:
> @water783 are you using OpenMLS and my nostr-openmls library or are you doing something else?

I am using your nostr-openmls library, but since I’m not very familiar with Rust, I bridged it to Dart language

https://github.com/0xchat-app/nostr-mls-package

> When a new message arrives, try to decrypt using the current exporter secret and epoch.
If that fails then try to use the past exporter secret and epoch - this should be limited to some number of epochs in the past though to ensure that we're not keeping old secrets around indefinitely.
If that fails, then it means we should have a commit message that rolls the group into the next epoch. If we haven't seen that we need to wait and store this message so we can decrypt later.

I think I understand your idea. I have a stupid question(I’m not very familiar with MLS): Is it possible that the local storage might not have the corresponding old epoch, which could prevent decrypting a message? For example, if the local storage has group keys for epoch 1 and epoch 2, but the epoch retrieved through `export_secret_as_hex_secret_key_and_epoch` is 4, and the message received encrypted at epoch 3, how would one decrypt the message in this case?

erskingardner:
> I am using your nostr-openmls library, but since I’m not very familiar with Rust, I bridged it to Dart language

Would Dart language bindings be better or is that the same thing as what you've built? I was planning on building out language bindings for my library once the API settled a bit more.

erskingardner:
To answer your question on a missing older epoch: I don't think your scenario could happen in the way you're describing.

From the point you've joined a chat, the way that you move forward in epochs with the ratchet tree is by applying commit messages. When each commit is applied, the tree moves forward and you get the new exporter secret. I guess there's a chance that a client could miss getting a commit from relays and then receive a later commit without ever finding the earlier commit but that would be quite a strange situation. And, in theory, the relay would still have your missing commit, you just need to sift through all the messages for the group to find it.

Another thing to keep in mind; the keys from older epochs (and older messages within those epochs) are thrown away as soon as they are used. This preserves FS. So you need to make sure that you hang on to all the decrypted messages in your own storage in the app (ideally in an encrypted at rest format). 

in order to keep the chat here about the spec itself, feel free to ping me directly on nostr DM and we can connect to talk more about the implementation details.

104.md
```diff
@@ -0,0 +1,288 @@
...
+        ["extensions", <An array of MLS Extension ID values e.g. "0x0001, 0x0002">],
+        ["client", <client name>, <handler event id>, <optional relay url>],
+        ["relays", <array of relay urls>],
+        ["-"]
+    ],
+    "sig": <signed with main identity key>
+}
+```
+
+- The `content` field is not required.
```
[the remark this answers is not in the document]
erskingardner: Great catch!

fiatjaf:
We need a new NIP number for this, please.

Maybe NIP-EE (aka _ernesty ernest_)?

erskingardner:
> We need a new NIP number for this, please.
> 
> Maybe NIP-EE (aka _ernesty ernest_)?

🫡

erlend-sh:
Very cool work! Any chance of harmonizing with this standardization effort?

https://germ-mark.github.io/distributed-mls-id/draft-xue-distributed-mls.html

https://www.germnetwork.com/blog/decentralized-pairwise-mls

erskingardner:
@erlend-sh thanks for pinging those links. I had only a very cursory look but it looks like that project is mostly focused on pairwise DMs rather than groups (of which DMs are a subcategory - just a group of 2). At first glance, I don't know why you'd modify the base MLS protocol to do pairwise as efficient groups are one of MLS's design goals. Maybe you can give me a bit more of an overview of why that is?

erlend-sh:
I’ve encouraged the author of that spec to open a new [discussion](https://github.com/parres-hq/whitenoise/discussions) in the whitenoise repo to begin an exchange of note. Mark is one of a handful of people investigating [MLS for group DMs in atproto.](https://wiki.atprotocol.community/en/working-groups/e2ee)

erskingardner:
Sounds good @erlend-sh - I'll keep an eye out for that.

EE.md
```diff
@@ -0,0 +1,288 @@
...
+   2. **Confidential** means that the contents of conversations can only be viewed by the intended recipients.
+2. Forward secrecy and Post-compromise security
+   1. **Forward secrecy** means that encrypted content in the past remains encrypted even if a key material is leaked.
+   2. **Post compromise security** means that leaking key material doesn't allow an attacker to continue to read messages indefinitely into the future.
+3. Scales efficiently for large groups
+4. Allows for the use of multiple device/clients in a single conversation/group.
+
+### Why MLS?
+
+This scheme adapts the Message Layer Security (MLS) protocol for use with Nostr. You can think of MLS as an evolution of the Signal Protocol. However, it significantly improves the scalability of encryption operations for large group messaging significantly (linear -> log), is built to accommodate federated environments, and also allows for graceful updating of ciphersuites and versions over time. In addition, it's very flexible and agnostic about the message content that is sent.
```
raphaelrobert, line 36:
```suggestion
This scheme adapts the Messaging Layer Security (MLS) protocol for use with Nostr. You can think of MLS as an evolution of the Signal Protocol. However, it significantly improves the scalability of encryption operations for large group messaging significantly (linear -> log), is built to accommodate federated environments, and also allows for graceful updating of ciphersuites and versions over time. In addition, it's very flexible and agnostic about the message content that is sent.
```

wcat7:
Hi! I recently had a discussion with Keychat regarding the MLS-related NIP:
[https://njump.me/nevent1qqsptah4ud4zhn9c596erx72x8k06qj3myvf89f4rrf4ks4xkf2u0ycpp4mhxue69uhkummn9ekx7mqpzemhxue69uhhyetvv9ujumn0wd68ytnzv9hxgq3qh0uj825jgcr9lzxyp37ehasuenq070707pj63je07n8mkcsg3u0q8cds9q](https://njump.me/nevent1qqsptah4ud4zhn9c596erx72x8k06qj3myvf89f4rrf4ks4xkf2u0ycpp4mhxue69uhkummn9ekx7mqpzemhxue69uhhyetvv9ujumn0wd68ytnzv9hxgq3qh0uj825jgcr9lzxyp37ehasuenq070707pj63je07n8mkcsg3u0q8cds9q)

I’d like to suggest a few potential improvements to the design:

1. **KeyPackage Event**
   Could we consider using an *addressable event* instead? This would allow us to utilize a `d` tag to identify different clients and MLS versions. It would ensure that, for a given client and MLS version, only one KeyPackage is stored on relays — simplifying lookup and avoiding duplication.

2. **Kind 445 Group Event (`h` tag)**
   Would it be feasible to derive the group public key using the `exporter_secret` as a private key, and computing the corresponding public key from that? This way, as epochs advance, the public key evolves as well, and we wouldn’t need to update the Nostr group metadata repeatedly.
   Additionally, perhaps the `created_at` field could be randomized (similar to NIP-17), using a value within a -2 day window, with the real timestamp stored in the inner event payload.

wcat7:
> 2. **Kind 445 Group Event (`h` tag)**
>    Would it be feasible to derive the group public key using the `exporter_secret` as a private key, and computing the corresponding public key from that? This way, as epochs advance, the public key evolves as well, and we wouldn’t need to update the Nostr group metadata repeatedly.
>    Additionally, perhaps the `created_at` field could be randomized (similar to NIP-17), using a value within a -2 day window, with the real timestamp stored in the inner event payload.

And how about using a **giftwrapped event** (`kind1059`) directly to encapsulate group messages — just like how welcome messages are handled? For example:
```
kind1059(nip44(kind445(mls_message(kind9("hello")))))
```
Instead of the current format:
```
kind445(nip44(mls_message(kind9("hello"))))
```

erskingardner:
> KeyPackage Event
Could we consider using an addressable event instead? This would allow us to utilize a d tag to identify different clients and MLS versions. It would ensure that, for a given client and MLS version, only one KeyPackage is stored on relays — simplifying lookup and avoiding duplication.

I thought about that originally, but it doesn't work. You actually want to have multiple key packages on relays at any given time for several reasons: 
1. The private key of the key package only exists on one client/device - so only that client/device can accept an welcome that was created with that given key package. This poses some UI challenges to clients but it's required because...
2. Groups can run different ciphersuites. Right now, I'm specifying that groups should use a single ciphersuite out of simplicity to start but the expectation is that this will change (and must for quantum resistance). 

> Kind 445 Group Event (h tag)
Would it be feasible to derive the group public key using the exporter_secret as a private key, and computing the corresponding public key from that? This way, as epochs advance, the public key evolves as well, and we wouldn’t need to update the Nostr group metadata repeatedly.
Additionally, perhaps the created_at field could be randomized (similar to NIP-17), using a value within a -2 day window, with the real timestamp stored in the inner event payload.

That's not a good idea. it means that an attacker that gains access to a group (or just the exporter secret) can then derive all later group id values and follow the group going forward. It also means that you can't run multiple group_id values at the same time (which is another trick for further obfuscating group activity that this design allows)

> And how about using a giftwrapped event (kind1059) directly to encapsulate group messages — just like how welcome messages are handled? For example:

That doesn't work, because then you have to send each message event to each member of the group individually. Completely destroying the efficiency of groups on MLS.

wcat7:
> I thought about that originally, but it doesn't work. You actually want to have multiple key packages on relays at any given time for several reasons: 
> 
> 1. The private key of the key package only exists on one client/device - so only that client/device can accept an welcome that was created with that given key package. This poses some UI challenges to clients but it's required because...
> 
> 2. Groups can run different ciphersuites. Right now, I'm specifying that groups should use a single ciphersuite out of simplicity to start but the expectation is that this will change (and must for quantum resistance). 

I’m not opposed to using multiple KeyPackages. What I’m suggesting is using an addressable event, where we can differentiate between different clients/devices, and even ciphersuites, using the d tag

> That's not a good idea. it means that an attacker that gains access to a group (or just the exporter secret) can then derive all later group id values and follow the group going forward. It also means that you can't run multiple group_id values at the same time (which is another trick for further obfuscating group activity that this design allows)

I mean using the export_secret derived public key. I believe exposing this public key should not be a security risk at this point. Additionally, as the group’s epoch advances, this public key (group ID) will change accordingly

> That doesn't work, because then you have to send each message event to each member of the group individually. Completely destroying the efficiency of groups on MLS. 

Sorry if I wasn’t clear. As for the kind 1059 event, I’m proposing that the p tag could represent the group ID, which would be the public key derived from the export_secret

wcat7:
Here is a more detailed proposal:

### **KeyPackage Event**

We could use an addressable event kind, e.g. `kind: 33443`, where the `d` tag is a SHA-256 hash of the JSON stringified array:

```json
["1.0", "0x0001", "0x0001, 0x0002", "whitenoise", "...other fields"]
```

**Example:**

```json
{
  "id": <id>,
  "kind": 33443,
  "pubkey": <main identity pubkey>,
  "content": "",
  "tags": [
    ["d", "sha256(jsonstring([\"1.0\", \"0x0001\", \"0x0001, 0x0002\", \"whitenoise\"]))"],
    ["mls_protocol_version", "1.0"],
    ["ciphersuite", "0x0001"],
    ["extensions", "0x0001, 0x0002"],
    ["client", "whitenoise", "<handler event id>", "<optional relay URL>"]
  ],
  "sig": <signed with main identity key>
}
```

**Benefits:**

* Ensures that only one KeyPackage per MLS version, ciphersuite, and client is stored on relays — identified uniquely by the `d` tag.
* Avoids manual relay cleanup of outdated KeyPackages.

---

### **Group Event**

We propose to derive the group ID by treating the `exporter_secret` as a private key, and deriving the corresponding public key.

**Group message event (`kind: 445`):**

```json
{
  "id": <id>,
  "kind": 445,
  "pubkey": <main identity pubkey>,
  "content": <serialized MLSMessage object>,
  "tags": [
    ["h", <pubkeyof(exporter_secret)>]
  ],
  "sig": <signed with main identity pubkey>
}
```

**Giftwrapped group message event (`kind: 1059`):**

```json
{
  "id": <id>,
  "kind": 1059,
  "created_at": <randomTimeUpTo2DaysInThePast>,
  "pubkey": <ephemeral sender pubkey>,
  "content": <nip44Encrypt(JSON.stringify(kind445 event))>,
  "tags": [
    ["p", <pubkeyof(exporter_secret)>]
  ],
  "sig": <signed with ephemeral sender key>
}
```

---

### **Advantages:**

1. Keeps consistency with how welcome messages are handled using `kind: 1059`, reducing confusion between `kind: 445` and `1059`.
2. Supports randomized timestamps (e.g., up to 2 days in the past), similar to NIP-17
3. As group epochs change, the group ID evolves naturally via `exporter_secret`, eliminating the need for an additional group metadata commit.
4. No need to send individual messages to all group members — clients just need to listen for `kind: 1059` events with `p` tag equal to the group ID.

erskingardner:
Thanks for the further context, but I still disagree with your logic. I don't think that the key package events should be replaceable. I've checked with a good few other devs on the topic and every one agreed that the (sometimes) odd behavior of replaceable events doesn't warrant changing just for the sake of one less req to delete an event. 

Re: message events. I also still disagree and will keep the kind as 445 instead of using a traditional GW. I don't want to couple the epoch to the secret or the id of the group. this has several privacy and security implications and further limits our ability to obfuscate groups by using multiple IDs. Basically, it's something I thought a lot about many months ago and decided against.

wcat7:
> Thanks for the further context, but I still disagree with your logic. I don't think that the key package events should be replaceable. I've checked with a good few other devs on the topic and every one agreed that the (sometimes) odd behavior of replaceable events doesn't warrant changing just for the sake of one less req to delete an event.
> 
> Re: message events. I also still disagree and will keep the kind as 445 instead of using a traditional GW. I don't want to couple the epoch to the secret or the id of the group. this has several privacy and security implications and further limits our ability to obfuscate groups by using multiple IDs. Basically, it's something I thought a lot about many months ago and decided against.

Thank you for your detailed explanation. I respect your considerations and choice.

That said, I have another question: would it be possible to consider changing the `h` tag to something else? This could help avoid confusion with the NIP-29 group tag, since NIP-29 groups accept events of any kind. Using the same `h` tag might prevent sending kind 445 messages within NIP-29 groups.

jab-r:
I am implementing the MLS kinds in a client and relay. Some points:

The MLS keypackage is essentially an MLS public key. No big security issue exposing it judiciously. I use then as one-time.

Shared Key packages **do not** contain private keys. Sharing key packages is essential to MLS group management.

SubatomicPlanets:
Hey I'm new to nostr and I would really love to see this merged! Why isn't it merged yet? Isn't it entirely thought through yet? I started making a simple messaging app a few days ago and this would be really nice to add! Do some relays already support it? And when will it be merged? I just want to know when I can fully use this and if it has been tested before. Thanks!

erskingardner:
@SubatomicPlanets thanks for the kick. We probably should merge this. The spec hasn't changed in many months so I think it's pretty stable. I believe we have two implementations now with @wcat7's 0xchat and my whitenoise.

@staab what do you think?

PR #1428, opened 2024-08-16
Define the `t` tag in NIP-24

alexgleason:


staab (approved):
I agree with this change, but is lowercasing the only normalization we want to do? I think yes, but would more aggressive slugification make sense?

vitorpamplona:
Lowercase is good. I am not sure about the requirement to have the tags listed in content as well. Sounds like overkill.

alexgleason:
@vitorpamplona The only time I've seen `t` tags with no hashtag in the text is spam. Blocking that blocked a lot of spam. I'm not sure there's a legitimate reason to have invisible hashtags in a note.

vitorpamplona:
Kind 1s have hidden tags all the time. I know because Amethyst displays hidden tags in an extra section below the post. I don't know which client creates them, but it is fairly common.

fiatjaf (approved):

fiatjaf:
I feel like having `t` tags without the corresponding text is a legitimate use case, but in practice I don't know how it would work because I never use them, so I'm ok with trying to remove that practice from earth.

I guess if you're forced to put the tags in the text it makes it clear that you're trying to spam topics if you are.

dluvian:
Voyage lets you set these hidden hashtags and I use them all the time. Writing hashtags in the content is ugly imo and setting them only in the metadata to improve discoverability should not be discouraged. 

If the goal is to prevent spam I would set a limit on the number of `t` tags.

alexgleason:
Alright, I modified it. Now it just says:

>   - `t`: a hashtag. The value MUST be a lowercase string.

PR #1429, opened 2024-08-17
Nip 01 escaping

mleku:
A proposal for further clarification regarding string escaping to help implementors avoid problems with generating incorrect event IDs.

Happy to trim it back a bit if others feel it is too verbose, certainly a couple of points are a bit repetitive, but I think they should stay to help reinforce the understanding of what a correct escape/unescape should actually do.


mleku:
this included commits from my other PR, revising

mleku:
see this PR for the revised version that has only the intended changes

https://github.com/nostr-protocol/nips/pull/1430

PR #1430, opened 2024-08-17
additional notes about escaping to ensure correct event IDs

mleku:
might be a bit long-winded but i think it makes it clearer how to avoid issues with incorrect re-encoding for implementers and ensuring that whatever clients will make and use won't get altered by a relay between decoding and encoding

it also makes a section header which will help in directing readers to this section in future

### DIFF
--- a/01.md
+++ b/01.md
@@ -42,6 +42,8 @@ To obtain the `event.id`, we `sha256` the serialized event. The serialization is
 ]
 ```
 
+### String Escapes
+
 To prevent implementation differences from creating a different event ID for the same event, the following rules MUST be followed while serializing:
 - UTF-8 should be used for encoding.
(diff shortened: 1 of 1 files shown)


mikedilger:
For taking a signature, you must
(1) convert the original JSON strings into UTF-8 strings by changing all escape sequences into their UTF-8 codepoints
(2) re-render these as canonical JSON strings using the escaping specified in NIP-01  (IMHO if there are codepoints below 20 not specified in NIP-01, then render them with \uXXXX), and take the signature.

The reason for this is that ANY character can be legally escaped, including all letters of the alphabet, meaning there are a large number of different representations mapping to the same UTF-8 string.  And as you mention there are octal escapes, etc, which should not be used when taking the signature.

I think changing to a scheme where we retain original escape sequences would break nostr far more than doing what I just explained, which is IMHO how most nostr software currently operates.

PR #1437, opened 2024-08-20
Revert "rename "parameterized replaceable event" to "addressable event""

fernandolguevara:
Reverts nostr-protocol/nips#1418

### DIFF
--- a/01.md
+++ b/01.md
@@ -78,8 +78,8 @@ This NIP defines 3 standard tags that can be used across all event kinds with th
 - The `e` tag, used to refer to an event: `["e", <32-bytes lowercase hex of the id of another event>, <recommended relay URL, optional>]`
 - The `p` tag, used to refer to another user: `["p", <32-bytes lowercase hex of a pubkey>, <recommended relay URL, optional>]`
 - The `a` tag, used to refer to a (maybe parameterized) replaceable event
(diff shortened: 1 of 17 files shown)


jb55:
nack

kehiy:
I think we need to make a final decision here. at the moment there are places where we used both of these names. this concept is confusing enough itself.

bezysoftware:
I find the renaming to "addressable" to make no sense. Now the NIPs, including 01, are polluted with stuff like
`a (maybe parameterized) replaceable event` and "normal replaceable events" 
Screenshot_20240928-143356
Bring back "replaceable" and "parametrized replaceable", stop this madness.

vitorpamplona:
I don't mind either way, but "normal replaceable" is indeed weird. 

Maybe it should be just `replaceable` then. Or just remove the world replaceable and make everything addressable (since "normal replaceables" are also addressable with d="")

fiatjaf:
The real madness is "parameterized replaceable", that is not a reasonable name. Come up with a better name, or a better set of names, instead, I'm all for that.

"Parameterized replaceable" is not descriptive, like what is the "parameter" that determines its replaceability, or what the hell all that means? "Addressable" matches `naddr` and `"a"` tags, that's it. Everything is just convention and getting used to.

But sure I'll clean the NIPs from these broken references, changes that should have gone in that other PR.

bezysoftware:
@fiatjaf still not a fan, but much better, thx

vitorpamplona:
Looks like this was settled.

PR #1438, opened 2024-08-22
Quoted highlights

jb55:
This is in preparation for Damus' highlight implementation. We have a new quote highlight variation. This is to prevent the creation and multiple notes (highlight + kind 1) for a single highlight action that wishes to add a comment along side a highlight, which looks bad in micro-blogging clients where these notes may appear in succession.

This is backwards compatible with existing highlights, except for the situation where clients might not properly handle the mention attribute.

Since @pablof7z has already mentioned he does not like this approach, we will create a [DIP] if/when this is not accepted.

[DIP]: https://github.com/damus-io/dips

### DIFF
--- a/84.md
+++ b/84.md
@@ -40,3 +40,10 @@ last value of the tag.
 ### Context
 Clients MAY include a `context` tag, useful when the highlight is a subset of a paragraph and displaying the
 surrounding content might be beneficial to give context to the highlight.
+
+## Quote Highlights
+A `comment` tag may be added to create a quote highlight. This should be rendered like a quote repost with the highlight as the quoted note.
+
+This is to prevent the creation and multiple notes (highlight + kind 1) for a single highlight action, which looks bad in micro-blogging clients where these notes may appear in succession.
+
+p-tag mentions should have a `mention` attribute to distinguish it from authors and editors.


vitorpamplona:
The tricky part is determining whether Highlights should be on a kind1 feed or not. Amethyst has always shown them, but it hasn't been super well received. Highlights are seen as personal notes + notes for others following the original page/post. It's not about letting my followers know all the highlights of the book I am reading as I mark them. Users might be highlighting lots of things for their future selves and those just pollute their feed. 

Derek posted 20+ highlights from an article he was reading last week. All highlights showed up to his followers. It was all fine, but it felt very weird. It looks like the user forgot that each highlight is a note and they are just spamming their followers with notes.

Otherwise, this looks ok to me.

staab:
> This is backwards compatible with existing highlights, except for the situation where clients might not properly handle the mention attribute.

This seems like the main problem to me. I agree with Vitor's UX critique as well. Maybe a new kind would be better? I do agree that having a comment go with a highlight would be an improvement.

vitorpamplona:
> I do agree that having a comment go with a highlight would be an improvement.

Agree as well. I just don't know if this is a comment for future me, for public view in this site or for my followers. 

> Maybe a new kind would be better?

To me, it's either this or a regular kind1 quote with the highlight's nevent inside of it. I am not sure if there is a need for a new kind.

staab:
> To me, it's either this or a regular kind1 quote with the highlight's nevent inside of it. I am not sure if there is a need for a new kind.

The problem with this is that you get duplicate notes, which is the original problem Will is pointing out. A new kind that is "highlight with comment for public viewing" vs the current "highlight without comment for private use" has pretty different semantics.

vitorpamplona:
> The problem with this is that you get duplicate notes, which is the original problem Will is pointing out. 

Yeah, but if we all agree that the Highlight is supposed to be quoted by a kind1 to display on Kind1 feeds then we would never show the highlight event itself (or only show it as a "reply")

pablof7z:
I'm not vehemently against it, I mainly don't see the issue this addresses as a big problem, as clients already need to be aware of the other stuff they're rendering in a feed (i.e. reposts/quotes).

The drawback I see is that if you repost, are you reposting the comment? the highlight? I mean -- you're reposting both, but you can't repost just the highlight.

84.md
```diff
@@ -40,3 +40,10 @@ last value of the tag.
 ### Context
 Clients MAY include a `context` tag, useful when the highlight is a subset of a paragraph and displaying the
 surrounding content might be beneficial to give context to the highlight.
+
+## Quote Highlights
+A `comment` tag may be added to create a quote highlight. This should be rendered like a quote repost with the highlight as the quoted note.
+
+This is to prevent the creation and multiple notes (highlight + kind 1) for a single highlight action, which looks bad in micro-blogging clients where these notes may appear in succession.
+
+p-tag mentions should have a `mention` attribute to distinguish it from authors and editors.
```
danieldaquino, line 49: This looks good! In addition to this, I also had to add attributes for `r` tags as well in our implementation, to allow us to distinguish between a source highlight URL _(I named that attribute `"source"`)_ and a regular a URL attachment/mention _(I named it `"mention"`)_

staab:
Coracle now supports rendering `comment` (in next release)

84.md
danieldaquino: @jb55, should we add the above to the spec PR?

jb55: just saw this, updating

jb55:
published DIP04 to document damus' implementation of this:

https://github.com/damus-io/dips/blob/master/04.md

fabianfabian:
I'm curious to know solution @pablof7z had in mind with clients already being aware of stuff to render? If a highlight event and a quote of that highlight event come in within 1 second its a guess that they should be linked together for rendering? DIP04 seems like a simple solution, I'll probably implement it also.

jb55:
The feedback I got was that you were supposed to filter the highlight if there was a quote repost of the highlight. I prefer something more self contained vs a spec that requires you to filter your feed.

fiatjaf:
https://git.fiatjaf.com/lantern implements this. It fits well with the code inherited from hypothes.is

Originally I went out of my way to create a cumbersome solution to do the kind:1 combo, but after chasing way too many bugs I've decided to just implement this proposal.

It's somewhat sad that the comment is in a tag though, it would be better as the main content while the highlight quote would go in a tag, but it's too late now.

PR #1446, opened 2024-08-26
Add `subject` and `t` tags to git issues

dluvian:
Issues usually have headers and self reported labels.

### DIFF
--- a/34.md
+++ b/34.md
@@ -104,7 +104,9 @@ The first patch in a series MAY be a cover letter in the format produced by `git
 
 ## Issues
 
-Issues are Markdown text that is just human-readable conversational threads related to the repository: bug reports, feature requests, questions or comments of any kind. Like patches, these SHOULD be sent to the relays specified in that repository's announcement event's `"relays"` tag.
+Issues are Markdown text that is just human-readable conversational threads related to the repository: bug reports, feature requests, questions or comments of any kind. Like patches, these SHOULD be sent to the relays specified in that repository's announcement event's `"relays"` tag. 
+
+Issues may have a `subject` tag, which clients can utilize to display a header. Additionally, one or more `l` tags may be included to provide labels for the issue.
 
 ```jsonc
 {
@@ -113,6 +115,9 @@ Issues are Markdown text that is just human-readable conversational threads rela
   "tags": [
     ["a", "30617:<base-repo-owner-pubkey>:<base-repo-id>"],
     ["p", "<repository-owner>"]
+    ["subject", "<issue-subject>"]
+    ["l", "<issue-label>"]
+    ["l", "<another-issue-label>"]
   ]
 }
 ```
--- a/34.md
+++ b/34.md
@@ -106,7 +106,7 @@ The first patch in a series MAY be a cover letter in the format produced by `git
 
 Issues are Markdown text that is just human-readable conversational threads related to the repository: bug reports, feature requests, questions or comments of any kind. Like patches, these SHOULD be sent to the relays specified in that repository's announcement event's `"relays"` tag. 
 
-Issues may have a `subject` tag, which clients can utilize to display a header. Additionally, one or more `l` tags may be included to provide labels for the issue.
+Issues may have a `subject` tag, which clients can utilize to display a header. Additionally, one or more `t` tags may be included to provide labels for the issue.
 
 ```jsonc
 {
@@ -116,8 +116,8 @@ Issues may have a `subject` tag, which clients can utilize to display a header. 
     ["a", "30617:<base-repo-owner-pubkey>:<base-repo-id>"],
     ["p", "<repository-owner>"]
     ["subject", "<issue-subject>"]
-    ["l", "<issue-label>"]
-    ["l", "<another-issue-label>"]
+    ["t", "<issue-label>"]
+    ["t", "<another-issue-label>"]
   ]
 }
 ```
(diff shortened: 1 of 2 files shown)


fiatjaf:
Instead of `l` wouldn't these be `t` tags?

dluvian:
> Instead of `l` wouldn't these be `t` tags?

A repo owner could choose to assign/overwrite labels with nip32 `l` tags. I chose `l` instead of `t` for consistency but I'm not  opposed to `t` if that feels more natural.

fiatjaf:
@DanConwayDev

DanConwayDev:
see [this comment on nostr](https://njump.me/nevent1qvzqqqqqqypzpgqgmmc409hm4xsdd74sf68a2uyf9pwel4g9mfdg8l5244t6x4jdqythwumn8ghj7un9d3shjtnwdaehgu3wvfskuep0qy88wumn8ghj7mn0wvhxcmmv9uqzqcysfuf2h2q4x9jx4kmxgymtmlv4py6lrw238a75hqh30cp786skm39ksm) and the surrounding thread for where @dluvian and I were discussing this.

@dluvian have another look at NIP-32. the `l` tag in NIP-32 is used to apply a label, in a verity of forms to an existing event.

The original intention was to use the `t` tag and allow application of these labels after the fact using NIP-32 events as follows:
```json
{
  "kind": 1985,
  "tags": [
    ["L", "#t"],
    ["l", "feature", "#t"],
    ["e", <issue_or_root_patch_event_id>, <relay_url>]
  ],
}
```

This might h have been intuitive for you without any changes to the NIP if this was actually implemented in gitworkshop.dev and issues / patches actually included `t` tags. The use of placeholder 'label' in the gitworkhop.dev UI probably thrown you off. I used this term as it leverages the ontology (terms used and understood) of github users.

I considered suggesting referecningn it in the NIP but it would grow large and unnecessarily complex when you start referencing all the other NIPs that could be used as is in combination with it. When it comes to implementing it though it probably would make life easier.

Custom stuff on top of vanilla NIP-32 that I have considered to improve tagging / labeling to improve the experience:
1. referencing the git repository announcement event aslo in the kind `1985` event so that all issues and any NIP-32 labels related to a repository could be fetched in a single filter. the problem with this is that some clients might not implement it so it would be best to fetch them individually also. this has a couple of problems: 1) the repository itself would no longer be able to use hashtags 2) adding this to NIP-34 would add complexity without a big improvement.
2. the ability to remove an incorrectly applied label. I thought issuing a more recent kind `1985` event with the same hashtag with a `-` prefix. @fiatjaf would special character prefix break how relays index tags?

I actually think removing labels (2) is important but I'm not sure how to elegantly implement it. Do you have any ideas about this?

fiatjaf:
You could send a a kind 5 in order to delete the label maybe?

dluvian:
Ok I changed it to `t` tags.

I was thinking that the repo announcement could have a `labels` tag containing predefined valid labels which issue creators can choose from. Clients would recognize these labels as valid and would render them if they are used in the issue event.

If the repo owner wants to change/remove labels from an issue he will publish a `kind:1985` event with a list of labels that overrides the labels defined in the issue itself. If he wants to change it again he will delete the old `kind:1985` and publishes a new one.

DanConwayDev:
> You could send a a kind 5 in order to delete the label maybe?

what if the maintainer wants to delete a label added by the issue author?

DanConwayDev:
> Ok I changed it to `t` tags.
> 
> I was thinking that the repo announcement could have a `labels` tag containing predefined valid labels which issue creators can choose from. Clients would recognize these labels as valid and would render them if they are used in the issue event.
> 
> If the repo owner wants to change/remove labels from an issue he will publish a `kind:1985` event with a list of labels that overrides the labels defined in the issue itself. If he wants to change it again he will delete the old `kind:1985` and publishes a new one.

No way! I actually planned to do that too but I didn't include it my comment above as it was getting too long.

I thought about having some general standard labels that a the client would choose. eg. good-first-issue, feature, bug, etc and then pull in a custom list of labels that a repository could specify.

this repositories would benefit from some solid default choices but could overwrite the defaults with their own if they wanted.

fiatjaf:
> what if the maintainer wants to delete a label added by the issue author?

It's the relay's choice to decide if they will accept a label from the issue opener or the maintainer.

And it's the client's choice to decide if they want to display labels from the maintainer or from the issue creator or both.

It's the relay's choice to decide they will delete a label from the issue creator on request from the maintainer.

In Nostr people are allowed to have different views over the same content, we don't have to decide that upfront.

DanConwayDev:
I actually meant that the maintainer may want to delete a label casually but inaccurately added by the issue author in a scenario that the author wouldn't object to.

the `-` prefix may not be a bad way of doing this clients could interpret this as removing a hashtag in the original event.

it never occurred to me that NIP-09 deletion events could be issued by anyone other than the original event author. If we only had to contend with hash tags via NIP-32 labels then perhaps deletion events might have been the way to achieve this (so long as they wouldn't be rejected by relays). Was this considered when drafting NIP-09? the only line which seems to suggest this isn't supported is:

> https://nips.nostr.com/9
A client MUST validate that each event pubkey referenced in the e tag of the deletion request is identical to the deletion request pubkey, before hiding or deleting any event.

DanConwayDev:
I think it is worth merging this as is as it is. It's implemented by voyage and gitworkshop.dev. 
a rough consensus emerged for the use of `subject` in this context. see https://github.com/nostr-protocol/nips/pull/1505.

Clarity around adding and removing labels via NIP-32 and kind 5 deletion events can be added to NIP-34 later.

PR #1448, opened 2024-08-27
nchat

v0l:
To make it possible to create a link to a DM chat with another user we should add a new TLV HRP for chats

### DIFF
--- a/17.md
+++ b/17.md
@@ -123,6 +123,11 @@ This NIP offers the following privacy and security features:
 
 The main limitation of this approach is having to send a separate encrypted event to each receiver. Group chats with more than 100 participants should find a more suitable messaging scheme.
 
+## NIP-19 entity
+To create an `nchat` link:
+  - `kind`: `17`
+  - `special`: one or more pubkeys concatenated
+
 ## Implementation
 
 Clients implementing this NIP should by default only connect to the set of relays found in their `kind:10050` list. From that they should be able to load all messages both sent and received as well as get new live updates, making it for a very simple and lightweight implementation that should be fast.
--- a/28.md
+++ b/28.md
@@ -133,6 +133,11 @@ Clients MAY hide event 42s for users other than the user who sent the event 44.
 }
 ```
 
+## NIP-19 entity
+To create an `nchat` link:
+  - `kind`: `28`
+  - `special`: `id` from `kind: 40`
+
 ## Relay recommendations
 
 Clients SHOULD use the relay URLs of the metadata events. 
--- a/19.md
+++ b/19.md
@@ -36,6 +36,7 @@ These are the possible bech32 prefixes with `TLV`:
   - `nevent`: a nostr event
   - `naddr`: a nostr _replaceable event_ coordinate
   - `nrelay`: a nostr relay (deprecated)
+  - `nchat`: a nostr chat conversation
 
 These possible standardized `TLV` types are indicated here:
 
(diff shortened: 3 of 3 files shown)


17.md
```diff
@@ -123,6 +123,11 @@ This NIP offers the following privacy and security features:
 
 The main limitation of this approach is having to send a separate encrypted event to each receiver. Group chats with more than 100 participants should find a more suitable messaging scheme.
 
+## NIP-19 entity
+To create an `nchat` link:
+  - `kind`: `17`
```
vitorpamplona, line 128: Correct kind is 14

17.md
```diff
@@ -123,6 +123,11 @@ This NIP offers the following privacy and security features:
...
+  - `special`: one or more pubkeys concatenated
```
vitorpamplona, line 129: I think we should not concatenate but add all pubkey hexes in a TLV list (which is already been used for relays).

May need to note that the order of keys doesn't matter.

28.md
```diff
@@ -133,6 +133,11 @@ Clients MAY hide event 42s for users other than the user who sent the event 44.
 }
 ```
 
+## NIP-19 entity
+To create an `nchat` link:
+  - `kind`: `28`
+  - `special`: `id` from `kind: 40`
```
vitorpamplona, line 139: Do we need nchat for public chats? I just use `nevent`s as usual.

vitorpamplona (approved):
Good idea.

17.md
v0l: `kind` is NIP number not event kind

17.md
v0l: We could have multiple `special` entries but idk if thats a problem for some clients

17.md
vitorpamplona: Why use the NIP number instead of just the chat kind?

v0l: Yea maybe it makes more sense to use the kind number, originally i had the HRP as `nchat17` and `nchat28` but decided to change it last minute to use the kind entry.

fiatjaf:
What is the point of this? Can't you chat with someone just from their `npub` or `nprofile`?

vitorpamplona:
> What is the point of this?

It creates a link that apps can parse and send the user straight into a DM chat with a group of users. Very useful.

17.md
vitorpamplona: It's going to be hard to trust people do this correctly, but we can request the list to be in ascending order by pubkeys.

But I would parse and resort it myself before doing anything with it.

v0l:
> What is the point of this? Can't you chat with someone just from their `npub` or `nprofile`?

Can also be used for NIP-29 chats with relay and `g` tag

vitorpamplona:
> How do you do that?

NIP-17 allows you to p-tag many people and send individual wraps to each of them. They define a group. It's like an email with many `to`'s :)

v0l:
> NIP-29 chats can be referred to with an naddr very easily, I don't think this is necessary

Its probably less important for web apps, but for native apps who want to link to nostr dms you cant really do it with npub because it would just go to profile, `nostr:npub` vs `nostr:nchat` 

I use something like `nchat` on snort already and though i would be useful for others: https://snort.social/messages/nchat171qgsyvrp9u6p0mfur9dfdru3d853tx9mdjuhkphxuxgfwmryja7zsvhq854tnh

17.md
```diff
@@ -123,6 +123,11 @@ This NIP offers the following privacy and security features:
 
 The main limitation of this approach is having to send a separate encrypted event to each receiver. Group chats with more than 100 participants should find a more suitable messaging scheme.
 
+## NIP-19 entity
+To create an `nchat` link:
```
1l0, line 127: A blank line is needed between 126 and 127 for expected rendering.

28.md
```diff
@@ -133,6 +133,11 @@ Clients MAY hide event 42s for users other than the user who sent the event 44.
 }
 ```
 
+## NIP-19 entity
+To create an `nchat` link:
```
1l0, line 137: A blank line is needed between 136 and 137 for expected rendering.

PR #1450, opened 2024-08-27
NIP-102: Subkey Attestation

ynniv:
This NIP defines a way to separate identity from authentication by using separate keys for each while still allowing immediate local validation of identity. If a subkey is compromised, the identity key can publish an event disavowing it, similar to a NIP-09 recovation.

### DIFF
--- a/102.md
+++ b/102.md
@@ -1,0 +1,96 @@
+NIP-102
+=======
+
+Hierarchical Deterministic Key Management
+-----------------------------------------
+
+`draft` `optional`
+
+This NIP defines a way to separate identity from authentication using hierarchical deterministic (HD) keys. This allows people to use one key pair to issue independent key pairs for different apps. If a key is compromised, the root key pair can publish an event revoking that key as of a specific time.
+
+HD keys (BIP-32, BIP-39, BIP-44) provide a means of creating new key pairs from a parent in such a way that the new pubkey can be verified to be a child of the parent's pubkey. A message signed by this new key can be rapidly and locally verified as an authentic subkey of the claimed parent.
+
+NIP-06 declares the default nostr derivation path to be `m/44'/1237'/<account>'/0/0`.
+
+```
+recovery phrase -> seed -> xpriv
+   -> m/44'/1237'/0'     -> Account0
+   -> m/44'/1237'/0'/0/0 -> Account0 Subkey0
+   -> m/44'/1237'/0'/1/0 -> Account0 Subkey1
+   -> m/44'/1237'/0'/2/0 -> Account0 Subkey2
+```
+
+## Management Event
+
+The parent key can publish a `Kind 10102` event for subkey management. This event lists subkeys that have been revoked, as well as those that are currently active, and a preference for how valid subkeys that are not listed should be treated. This is only a preference, as it may not always be immediately available to clients and relays.
+
+Content:
+```json
+{
+  "keys": {
+    "<hex-subkey0-pubkey>": { "active_at": "<epoch-timestamp>" },
+    "<hex-subkey1-pubkey>": { "active_at": "<epoch-timestamp>", "revoked_at": "<epoch-timestamp>" },
+    "<hex-subkey2-pubkey>": { "active_at": "<epoch-timestamp>" }
+  },
+  "default_policy": "allow"
+}
+```
+
+## Signing
+
+When signing events using a subkey, the account pubkey and subkey derivation path will be included as a well known attribute. Relays and clients should validate the subkey->parent relationship, and immediately discard messages with invalid claims.
+
+```json
+{
+  "pubkey": "<hex-subkey1-pubkey>",
+  "kind": 1,
+  "tags": [
+    ["account", "<hex-account0-pubkey>", "<subkey1-derivation-path>"],
+  ],
+}
+```
+
+For clients that don't implement NIP-102, we can have the account key publish a `Kind 10102` key management event using the subkey at the time of creation:
+
+Content:
+```json
+{
+  "account": "<hex-account0-pubkey>",
+  "derivation_path": "<subkey1-derivation-path>"
+}
+```
+
+## Behavior
+
+All searches for the account pubkey should return those signed by any account subkey. If an event is replaceable, it is up to the client or relay as to whether both messages are maintained, or if only the latest across all subkeys is maintained. In any situation where data will be lost, a reasonable effort should be made to locate the most recent revocation list before proceeding.
+
+## Migration
+
+Most current keys do not use BIP-39 derivation and will need to perform a manual key rotation to a new pubkey using a method not defined here.
+
+## Reference Code
+
+```python
+from bip_utils import Bip39SeedGenerator, Bip32Slip10Secp256k1
+
+account0_derivation_path = "m/44'/1237'/0'"
+subkey0_derivation_path = "0/0"
+
+mnemonic = "abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon about"
+master_xpriv = "xprv9s21ZrQH143K3GJpoapnV8SFfukcVBSfeCficPSGfubmSFDxo1kuHnLisriDvSnRRuL2Qrg5ggqHKNVpxR86QEC8w35uxmGoggxtQTPvfUu"
+account0_xpub = "xpub6D6V5EX8HTe95getx2tTH2QApmrA1nPJFEnneAK813RjcDdSc3WaAF7BRNpTF7o7zXjVm3DD3VMX66jhQ7wLaZ9sS6NzyfiwfzqDZbxvpDN"
+subkey0_xpub = "xpub6Gf5o5yEF14TykSmvZBzS9wFSgnqvPsxit1v4CaaNf6S6S5mm169FRN3QkCsVsDm8NNaN8eGbQg9vR43BD9UqQTrfWFmRKoWep2gxQpFh3Q"
+
+seed = Bip32Slip10Secp256k1.FromSeed(Bip39SeedGenerator(mnemonic).Generate())
+account0 = seed.DerivePath(account0_derivation_path)
+subkey0 = account0.DerivePath(subkey0_derivation_path)
+
+def is_subkey(pubkey, subkey, derivation_path):
+    parent = Bip32Slip10Secp256k1.FromExtendedKey(pubkey)
+    if parent.DerivePath(derivation_path).PublicKey().ToExtended() == subkey:
+        return True
+    else:
+        return False
+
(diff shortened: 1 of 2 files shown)


bezysoftware:
FYI NIP-100 is already used in this PR: https://github.com/nostr-protocol/nips/pull/1411

ynniv:
> FYI NIP-100 is already used in this PR: #1411

Moved to NIP-102

102.md
```diff
@@ -0,0 +1,96 @@
...
+```json
+{
+  "keys": {
+    "<hex-subkey0-pubkey>": { "active_at": "<epoch-timestamp>" },
+    "<hex-subkey1-pubkey>": { "active_at": "<epoch-timestamp>", "revoked_at": "<epoch-timestamp>" },
+    "<hex-subkey2-pubkey>": { "active_at": "<epoch-timestamp>" }
+  },
+  "default_policy": "allow"
+}
+```
```
staab, line 37: I'd model this using single-letter tags so that they can be indexed. That would then obviate the need for adding an `account` tag to every event published by a subkey, since client could just fetch any `10102`s for pubkeys they don't recognize.

staab (commented):
Is there a benefit to using HD keys rather than just arbitrary sub-keys? The reason I ask is the sub-keys have to be published anyway, so the deterministic connection doesn't seem to matter much.

The challenge with this sort of thing has always been the soft-fork nature of making clients aware of relationships between keys. Not a terribly difficult problem, but it would be some time before this could be really usable.

102.md
ynniv: I'm not sure that I understand how using a single letter tag would change whether or not we need a tag with the `account`s npub. My suspicion is that the pipeline stalls caused by constantly fetching new `10102`s would outweigh the space savings. Is the efficiency of single-letter vs word tags due to how relays are optimized?

melvincarvalho:
I have been using subkeys for a while, and I love them.  Subkeys should be a standalone nip, and not coupled to use cases such as user locking.  While I personally prefer open ended subkeys (maybe that's still possible, here?) I can live with this solution as a way of working.  I do already have subkeys to my master key, so it would be nice if it would be possible for me to link those still, rather than have to start a new hierarchy.

alexgleason:
I wish I had done this for the Mostr Bridge.

vitorpamplona:
Instead of a list of active keys why not just do a bloomfilter (or equivalent structure)? In that way, users don't need to expose all keys their master accounts use. Yes, people could still collect ALL events and filter all `account` to a given key, but events that are protected in some relays or inside encrypted wraps won't be public. Some keys (your health care data keys, for instance) will never be used in public.

The active dates don't really work. If they are based on `created_at`s then the attacker can post in the past. If they are based on the receipt date, they are not needed (you can just remove keys from the list). Either way, if a key has leaked, nothing from that key can be trusted. There is no way to separate the author's original event set from an attacker's set based on dates because the precise date/time the attacker got the keys is unknown.

ynniv:
> Instead of a list of active keys why not just do a bloomfilter (or equivalent structure)? In that way, users don't need to expose all keys their master accounts use. Yes, people could still collect ALL events and filter all `account` to a given key, but events that are protected in some relays or inside encrypted wraps won't be public. Some keys (your health care data keys, for instance) will never be used in public.

The goal of `active` is really to facilitate DM's. If you want to be able to read them using multiple clients, they'll need to be sent in a way that any of those clients will be able to read them (so, individually, or as some type of group messaging). Any valid subkey claiming to be part of the account (that hasn't been revoked) would still be part of the account.

> The active dates don't really work. If they are based on `created_at`s then the attacker can post in the past. If they are based on the receipt date, they are not needed (you can just remove keys from the list). Either way, if a key has leaked, nothing from that key can be trusted. There is no way to separate the author's original event set from an attacker's set based on dates because the precise date/time the attacker got the keys is unknown.

Agreed. I'll clarify that these attributes are non-authoritative, but I think it's good to include some record of the expected timing.

ynniv:
> I do already have subkeys to my master key, so it would be nice if it would be possible for me to link those still, rather than have to start a new hierarchy.

If they were created using the same derivation (`m/44'/1237'/0'/<subkey-index>/0`), then clients and relays that support this NIP should automatically begin treating them as part of the `subkey0` account.

vitorpamplona:
> The goal of active is really to facilitate DM's.

Make an "active" set just for DMs, then. DMs are not going to be in every client. There will be two/three orders of magnitude more keys than the ones used by DM-enabled clients. 

> I'll clarify that these attributes are non-authoritative, but I think it's good to include some record of the expected timing.

Sounds like wishful thinking. There is nothing anyone can reliably do with those dates. It only adds insecurity to the proposal.

ynniv:
> Make an "active" set just for DMs, then

What would that look like? Renaming `active` to `dm`?

> Sounds like wishful thinking. There is nothing anyone can reliably do with those dates. It only adds insecurity to the proposal.

That's fair. If an account wants to be able to revoke a key while preserving historical content, they could aggressively rotate them. Aside from the hassle of rotating keys securely, there's nothing that should prevent someone from using a new subkey for every message.

vitorpamplona:
> What would that look like? Renaming active to dm?

Sure. I think more key sets will show up as new NIPs arrive. Thus, I would align the name to the thing that the group of keys is supposed to be used for... in this case, to group keys and encrypt to them. 

We discussed key aliases for DMs in the past: https://github.com/nostr-protocol/nips/pull/1306 Your idea is somewhat similar, but instead of senders just picking one, as in #1306, they have to encrypt all of them. #1306 is an OR filter, here it is an AND.

melvincarvalho:
> > I do already have subkeys to my master key, so it would be nice if it would be possible for me to link those still, rather than have to start a new hierarchy.
> 
> If they were created using the same derivation (`m/44'/1237'/0'/<subkey-index>/0`), then clients and relays that support this NIP should automatically begin treating them as part of the `subkey0` account.

I didnt use that no.  There's quite a few different KDFs possible.  Then there are random keys for completely unlinked personae.

mikedilger:
First read just now and I'm kicking myself that I didn't think of this.  I had thought that subkeys required some kind of signed document that tied the master key to the subkey.  I forgot you could derive a subkey algorithmically... so I'm very glad to have seen this NIP.

This is a little bit like NIP-26 delegated event signing, except no conditions, actually the same principal not really delegated.  But it has that tag saying which real identity the event author is.

Semisol (changes requested):
Knowledge of the parent xpub and a child private key being compromised causes compromise of all keys derived from the xpub.

ynniv:
> Knowledge of the parent xpub and a child private key being compromised causes compromise of all keys derived from the xpub.

This is the best reason for code reviews. I did some research, and it is unfortunately correct.

The only property of HD keys that's essential here is the ability to validate their derivation from the claimed identity pubkey without any additional information. This keeps things fast in the common case, though we still need to do a lookup to confirm non-redaction before making destructive changes (eg, replaceable events, or revocation events). We can implement this same property with a non-HD key by including an attestation that this key is part of the main identity, signed by the main key. This adds more data to each message, and makes this NIP look very similar to NIP-26.

So, what's wrong with NIP-26? Adoption has been low enough to soft-deprecate it and remove the code from some clients, even though it seems to solve an important problem that still doesn't have a satisfying solution. An obvious criticism is that it's too complicated, and adds a lot of data to each message. This comes from the idea that keys aren't merely revokable tokens, but have specific authorizations based the attributes they are allowed to publish: kind, created_at, etc. @vitorpamplona pointed out the limited value of making decisions on created_at. Both these rules and the attestation need to be included in every message, even though they are of limited use.

If we've given up on NIP-26, does it make sense to re-imagine a lighter version? Specifically one that looks exactly like this proposal, with the addition of the smallest possible attestation from the already included `I` keypair saying that this pubkey is equivalent to `I (until revoked), and removal of "other_keys" from `Kind 10102` since this is now the common case.

mikedilger:
NIP-26 had 3 problems.  It had conditions of questionable use.  It had a non-searchable tag.  And it required both relays and clients to support it before it became useful... there was no easy path through partial support. Solve those three problems and you'll have a winner.

BTW, you can prove part of the BIP-32 tree belongs to your pub without releasing your xpub.

paulmillr:
Just my 5 cents here.

HD keys are terrible. Anything is better. For example, `child_key = hkdf(main_key, salt: "nip-102", info: "reason-5")` is better security-wise and complexity-wise.

> Knowledge of the parent xpub and a child private key being compromised causes compromise of all keys derived from the xpub.

That quote by @Semisol is true. Only **hardened** hd keys don't suffer from this.

102.md
```diff
@@ -0,0 +1,107 @@
...
+
+When signing events using a subkey, the account pubkey will be included as a well-known attribute. The event will also include an attestation of the subkey's authorization signed by the main key. Relays and clients should validate this attestation, and discard messages that make invalid claims.
+
+```json
+{
+  "pubkey": "<hex-subkey1-pubkey>",
+  "kind": 1,
+  "tags": [
+    ["I", "<hex-account0-pubkey>"],
+    ["Ia", "<subkey1-attestation>"]
```
ynniv, line 41: I think this is correct because we won't search for the attestation. If there are other reasons to use single characters I was considering `J`.

devsou-com:
Just a reminder: I'm a newbie Nostr developer and I have no experience with public/private key infrastructure.

I understand that NIP-26 only deals with delegation, that is, it does not deals with token revocation or token management (it is quite similar to JWT), and worst, its attestation token is valid forever! In my opinion, it can be deprecated right now https://github.com/nostr-protocol/nips/pull/1051

I was sad to know that using hierarchical deterministic subkeys is problematic. It would be very important to validate that the public key is derived from the parent's public key. But at the same time I'm happy, because this opens up the possibility for an existing account to be able to manage its own events and events on behalf of another account at the same time!

@ynniv I liked your proposal, simple and direct (aligned with the Fiatjaf's philosophy/methodology). It deals with key management and revocation. In my opinion, it is ready to be approved. Just a small change, I would love to use the `I` and `J` tags (I didn't like `Ia`).

fiatjaf:
This introduces absurd complexity that makes Nostr unusable.

ynniv:
> This introduces absurd complexity that makes Nostr unusable

While I agree that it could be confusing without significant client adoption, I think backwards compatibility allows for some clients to lead the way. Impact on the relays is minimal, and primarily about replaceable events and deletion requests.

To me, key management is critical, and coming up with a way forward is worth the effort.

devsou-com:
@ice-orestes I analyzed your proposal. @vitorpamplona asked you about the use of timestamps. You said that this would prevent replay attacks. @ynniv proposal does not use timestamps and it seems to work very well. Since these are replaceable events, you just need to rely on the "created_at" attribute.

Am I wrong in my reasoning? (reminder, I'm a newbie Nostr developer).

PR #1452, opened 2024-08-28
Key Migration and Revocation

braydonf:
"This NIP defines the protocol that SHOULD be implemented by clients and relays to handle a key migration and revocation of a key. At a minimum this specification defines a protocol for a compromised private key to be revoked. Clients give warning that the key is compromised. Relays and clients delete events and reject future events from a revoked key. Also defined is a protocol to migrate to a new successor key with multiple optional methods for users of clients to verify and accept or reject the change. This includes methods such as: a social graph with attestations, recovery keys and side-channel identity anchors."

[Read in Full](https://github.com/braydonf/nips/blob/key-migration-and-revocation/xx.md)

### DIFF
--- a/xx.md
+++ b/xx.md
@@ -1,0 +1,153 @@
+NIP-X
+=====
+
+Key Migration and Revocation
+------
+
+`draft` `optional` `author:braydonf`
+
+This NIP defines the protocol that SHOULD be implemented by clients and relays to handle a key migration and revocation of a key. At a minimum this specification defines a protocol for a compromised private key to be revoked. Clients give warning that the key is compromised. Relays and clients delete events and reject future events from a revoked key. Also defined is a protocol to migrate to a new successor key with multiple optional methods for users of clients to verify and accept or reject the change. This includes methods such as: a social graph with attestations, recovery keys and side-channel identity anchors.
+
+There are four new events introduced:
+
+* [Recovery Keys Setup](#recovery-keys-setup-event)
+* [Recovery Keys Attestation](#recovery-keys-attestation-event)
+* [Key Migration and Revocation](#key-migration-and-revocation-event)
+* [Key Migration Attestation](#key-migration-attestation-event)
+
+## Recovery Keys Setup Event
+
+This is an event that is non-replaceable. Its primary purpose is to define a set of recovery keys that can be used to migrate to a new key in the future, if it becomes necessary. The event can assign anywhere from `1` to `n` recovery keys assigned to be able to sign the _Key Migration and Revocation Event_. A threshold number of keys (`m` of `n`) can be assigned to verify this event.
+
+```json
+{
+  "kind": "<tbd>",
+  "pubkey": "<user-pubkey>",
+  "tags": [
+    ["p", "<recovery-pubkey-1>"],
+    ["p", "<recovery-pubkey-2>"],
+    ["p", "<recovery-pubkey-3>"],
+    ["threshold", "2"],
+    ["recovery-key-setup"]
+  ],
+  "content": ""
+}
+```
+
+* A `p` tag MUST be included and be the hex encoded public keys for the recovery keys.
+* The `recovery-key-setup` MUST be included, it's otherwise ignored, however can help prevent making this event by chance or accident.
+* If `threshold` tag is NOT included, the threshold MUST default to `1`.
+* The content MAY include a comment, most clients can ignore this field.
+
+### Behaviors
+
+* Clients MUST only consider one to be valid.
+* If multiple events exist for a public key, a user interface SHOULD provide a means to pick one. There can be various means to help select the key including; displaying _public_ attestations from within a social graph or NIP-03 timestamp associated with the event. Please see [External References](#external-references) for UX examples.
+* Any future events of this kind MUST NOT be automatically accepted and considered verified as it could be from an attacker due to a compromised private key, it could ALSO be the honest event.
+* Every client SHOULD make either a _private_ or _public_ attestation when receiving a _Recovery Keys Setup Event_ for public keys that they follow. The default SHOULD be _private_. The client SHOULD provide user interaction to make attestations _public_ or _private_.
+* The attestations SHOULD be stored locally to verify a possible future _Key Migration and Revocation Event_.
+* Relays SHOULD store multiple events from a public key of this kind.
+* Clients SHOULD implement a user interface to help prevent accidental broadcasts of this event.
+* Clients MAY provide a manual verification process that can be verified through a side-channel to be able to independently replace the _Recovery Keys Setup Event_.
+* Clients are ENCOURAGED to use hardware devices and NIP-06 seed phrases to backup the recovery keys.
+
+## Recovery Keys Attestation Event
+
+This is an event that is non-replaceable. This is a means to save a single valid _Recovery Keys Setup Event_ for a public key. The attestation can be either _private_ or _public_. By making a _public_ attestation, others in the network can see that they are able to verify the recovery keys for a profile; this can help built a robust fault-tolerant network. The default option SHOULD be _private_. The primary purpose for this event is for clients to be able to verify the recovery keys for a later _Key Migration and Revocation Event_.
+
+```json
+{
+  "kind": "<tbd>",
+  "pubkey": "<user-pubkey>",
+  "tags": [
+    ["p", "<pubkey-of-friend>"],
+    ["recovery-key-attestation"]
+  ],
+  "content": "<recovery-key-setup-event>"
+}
+```
+
+* The `p` tag for a _public_ attestation MUST include a public key and MUST match the public key for the Recovery Key Setup Event.
+* A _private_ attestation MUST NOT include the `p` tag.
+* The `content` field MUST include the Recovery Key Setup Event either encrypted and _private_ or unencrypted and _public_.
+* The `recovery-key-attestation` MUST be included, it's otherwise ignored, however can help prevent making this event by chance or accident.
+* A NIP-03 timestamp attestation MAY be included for this event and clients can use this to help with verification.
+
+## Key Migration and Revocation Event
+
+### Event Details
+
+This is an event that is non-replaceable. It will revoke a public key and all events for the key will be deleted. It will also provide a key migration for clients and users to accept or reject the key migration based on various means of verification including; recovery keys, social graph and external identities.
+
+```json
+{
+  "kind": "<tbd>",
+  "pubkey": "<user-pubkey>",
+  "tags": [
+    ["new-key", "<new-pubkey>"],
+    ["e", "<event-id-of-recovery-key-setup>"],
+    ["key-migration-and-revocation"]
+  ],
+  "content": "<recovery-key-signatures-and-optional-comment>"
+}
+```
+
+* The `content` MUST include the signatures for `m` of `n` public keys as specified in the referenced _Recovery Keys Setup Event_. The signature SHOULD be for the entire event as similar to the event signature.
+* If a `new-key` IS provided, the `key-migration-and-revocation` tag MUST be included, it's otherwise ignored, however can help prevent making this event by chance or accident.
+* If a `new-key` IS NOT provided, the `key-revocation` tag MUST be included, this is also to help prevent mistakes.
+* The `new-key` tag SHOULD be included, otherwise it will be a key revocation without a successor key and all events will be deleted.
+* Clients SHOULD implement a means to verify that users are aware of the account and behaviors of the event to avoid accidental broadcast.
+* A NIP-03 timestamp attestation MAY be included for this event and clients can use this to help with verification.
+* Clients MAY use a NIP-18 Generic Repost with a copy of the event in the content to help relay distribution.
+
+### Event Handling for Clients
+
+For a client, this event is both a revocation and a migration. The revocation MUST be handled by verifying only the event signature and MUST be automatic. The migration, on the other hand, MUST be presented and verified by the user to accept or reject the migration key change.
+
+#### Behaviors
+* Signature verification of the event and NOT the recovery keys MUST determine the revocation validity.
+* Upon a valid key revocation:
+  * All events MUST display a warning that the event is from a revoked public key.
+  * All events of a revoked public key SHOULD be deleted, this MAY be after a duration of time depending on the client or preferences.
+* A user interface MUST display to the user an option to accept or reject a key migration. This SHOULD include the recovery keys and how many have valid signatures, social graph verification of those that they follow that are now following the new key, and an other side-channel pinned identities such as with NIP-05. At least one method MUST be provided. The existence of a _Recovery Keys Setup Event_ is NOT REQUIRED to be able to handle a _Key Migration and Revocation Event_. Please see [External References](#external-references) for UX examples.
+* Upon a valid key migration:
+  * The old key MUST be unfollowed and the new key MUST be followed.
+  * The old key MUST be stored, with a Key Migration Attestation Event, for future reference to able to block and delete future events from the old key.
+  * The old key MAY be added to other mute lists.
+
+### Event Handling for Relays
+
+For a relay, this event is primarily a key revocation, and storing the necessary information for clients to verify the key migration.
+
+#### Behaviors
+* Signature verification of the event and NOT the recovery keys MUST determine the revocation validity.
+* Upon a valid key revocation:
+  * All events of a revoked public key MUST be deleted. The time-frame that events are deleted MAY be defined by an agreed upon terms between client and relay.
+  * All future events, as determined when it was received and not the date on the event, of a revoked public key MUST be rejected except for another _Key Migration and Revocation Event_. This is to ensure that if a key is compromised and a fraudulent event is made, an honest event can also be made and broadcast. Each client can then verify which is honest.
+* The recovery keys and signatures SHOULD NOT be verified, all key migration verification is handled by the client.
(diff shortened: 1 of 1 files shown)


braydonf:
For those familiar with similar proposals by @pablof7z, @fiatjaf, @nvk and @vitorpamplona on the topic, I've written a summary of the differences.

In comparison with: https://github.com/nostr-protocol/nips/pull/829

- This proposal doesn't rely upon a bitcoin protocol opentimestamp service, though it can be used as an option. Keeping a network consensus wide agreed upon order of events isn't mandatory, instead the client keeps track of the order and validity. This is done through attestations that are either private or public, and verified through a social graph for others. Migration is not automatic, however revocation can be automatic.
- "Whitelisting a pubkey" is similar to "Recovery Key Setup" in this proposal, however it also includes the ability to have multiple keys and assign `m` of `n` signatures to create a migration event later. Also as mentioned, a NIP-3 timestamp isn't required.
- "Migrating to a pubkey" is similar to "Key Migration and Revocation" in this proposal. The event can also be used purely to revoke a key without a successor key. The most minimal implementations SHOULD implement revocation. The 60-day wait isn't necessary for migration, this is determined by the user of the client based on information presented, which could include a time component.

In comparison with: https://github.com/nostr-protocol/nips/pull/637

- This proposal is similar in that it includes both a simple revocation of a key and a social means to migrate by means of multiple keys and signatures. However instead of a predetermined set of friends (of which relationships could change in the future), a social graph of mutual friends, can be used to verify a migration. Attestations of a migration or backup keys setup can be made by mutual friends that will help others in determining the truth.
- The recommended switch of a new pubkey is done through attestations in this proposal, this applies to both a recovery key setup as well as migration events.

In comparison with: https://github.com/nostr-protocol/nips/pull/1056

- This proposal is similar in that it uses a social graph as defense against an attack from a compromised private key. 
- It differs in that it also includes an ability to provide recovery keys that can help to facilite the social verification of the migration of a key, though these are not required on other means can be used. 
- Also, the event for migration of a key is also a revocation of the key and can just be a revocation. Revocation is verified using the event signature, migration is verified using several options and the user of the client decides.
- Similarly, ideas such as using Generic Reposts to help propogate migration and revocation events are also used.

This proposal is, more or less, a merger of all three of the above proposals.

fiatjaf:
One thing that isn't clear is how this protects against an attacker stealing your key, then publishing a new recovery key event with an older date, then proceeding to do the key rotation with his own recovery key and now he owns your identity.

braydonf:
If an attacker steals a private key and publishes a new and fraudulent _Recovery Keys Setup Event_ with an older date, it can be detected as fraudulent by everyone that has published their own dated and signed copy in a _Recover Keys Attestation Event_ of the previously seen and honest _Recovery Keys Setup Event_. Attestations can be either _private_ or _public_, public attestations can help those that have not made such attestations to be able to detect if it is fraudulent _(in addition to the other methods)_.

pablof7z:
> If an attacker steals a private key and publishes a new and fraudulent _Recovery Keys Setup Event_ with an older date, it can be detected as fraudulent by everyone that has published their own dated and signed copy in a _Recover Keys Attestation Event_ of the previously seen and honest _Recovery Keys Setup Event_. Attestations can be either _private_ or _public_, public attestations can help those that have not made such attestations to be able to detect if it is fraudulent _(in addition to the other methods)_.

You mean my client has to cache forever all fiatjaf's key setup event to try to ascertain time order?

braydonf:
@pablof7z The _Recovery Keys Attestation Event_ is saved on relays, for both _private (encrypted)_ and _public_.

nvk:
This might become an issue, relays are not designed to offer data relation guarantees

pablof7z:
> @pablof7z The _Recovery Keys Attestation Event_ is saved on relays, for both _private (encrypted)_ and _public_.

Ok, yeah, I assumed so, but how can OTS be optional, if you're not requesting clients to keep track of all setup events; how do you order by date?

braydonf:
@nvk Can you clarify what you mean by "data relation guarantees" and how that is relevant?

braydonf:
@pablof7z OTS is optional because ordering by date of all _Recovery Keys Setup Events_ for a pubkey isn't necessary for determining which one is valid.

Here is an example:
- Bob creates 3 keys to be used for a recovery, and broadcasts a _Recovery Keys Setup Event_ (2 of 3 keys).
- Alice follows Bob and makes a private _Recovery Keys Attestation_ for Bob's _Recovery Keys Setup Event_. Both Bob and Alice have now signed the _Recovery Keys Setup Event_.
- An attacker steals Bob's private key and broadcasts a fraudulent _Recovery Keys Setup Event_ with a different set of recovery keys.
- Alice sees this new Recovery Keys Setup Event event and can know that it fraudulent because she already has attested to a different one.

nvk:
Sorry, autocorrect. "Retention"

braydonf:
@nvk Okay, that makes more sense. Data retention by relays depends on the relay and the terms of service. Redundancy of relays, paid relays and private/personal relays all help to backup ALL notes and events. Some relays will delete older events to free disk space; in that process, it should be emphasized that relays SHOULD NOT delete _Recovery Key Attestation Events_ and _Key Migration Attestations Events_.

nvk:
> @nvk Okay, that makes more sense. Data retention by relays depends on the relay and the terms of service. Redundancy of relays, paid relays and private/personal relays all help to backup ALL notes and events. Some relays will delete older events to free disk space; in that process, it should be emphasized that relays SHOULD NOT delete _Recovery Key Attestation Events_ and _Key Migration Attestations Events_.

The problem is user expectations and guarantees,

braydonf:
If anyone is interested in helping to implement in clients and relays and flush out details the specification, please reach out to me! Thanks!

braydonf:
@nvk If it is a problem, it would be for the Nostr protocol and all events; if events can disappear from all relays then Nostr isn't a censorship-resistant network!

nvk:
> @nvk If it is a problem, it would be for the Nostr protocol and all events; if events can disappear from all relays then Nostr isn't a censorship-resistant network!

Doesn't feel right, happy to be proven wrong. Also aside from the previous proposal I'm not sure I have any better ideas.

vitorpamplona:
> if events can disappear from all relays then Nostr isn't a censorship-resistant network!

They can and they do every day. Not your relay, not your data. 

Nostr is a censorship resistance protocol, not network. It allows users to use backup relay networks at their will, but if they put everything in a single relay and that relay goes out of business, all their data is gone.

pablof7z:
> > if events can disappear from all relays then Nostr isn't a censorship-resistant network!
> 
> 
> 
> They can and they do every day. Not your relay, not your data. 
> 
> 
> 
> Nostr is a censorship resistant protocol, not network. It allows users to use backup relay networks at their will, but if they put everything in a single relay and that relay goes out of business, all their data is gone.

Ok, hear me out..

Nostr... but on a blockchain

MIC-DROPPED

melvincarvalho:
IMHO better to use [single use seals](https://petertodd.org/2017/scalable-single-use-seal-asset-transfer#) for the on-chain commitments, rather than OTS.  Peter Todd made both OTS and single-use seals, in that order.  I dont think I would implement this when single-use seals offers imho a vastly superior path to data, with history, audit trail, and recoverable profiles.  No issues with clients opting into this, but I think it may lead to a fractured landscape of solutions (which itself is not the end of the world).

melvincarvalho:
> @nvk If it is a problem, it would be for the Nostr protocol and all events; if events can disappear from all relays then Nostr isn't a censorship-resistant network!

It's not a censorship-resistant network.  Spam is censored all the time.  Relays are not data stores, they just transmit notes, and other stuff, from one user, to another.  If you want to store events, it is much better not to use nostr for that, but rather, the cloud, which was designed for storage, and has super mature tooling.

mikedilger:
First, if I am Alice, how do I know that I should attest to Bob's key recovery setup event?  Do I call him on the phone?  Is it just because I have never seen one before?  Because maybe an Imposter sends a second one, but that is the first one I have seen and so I attest to the wrong one.

Second, in your scenario Alice becomes aware of an imposter recovery key setup event.  But how does the rest of the network learn?  Most people don't follow Alice, and even if they could they don't trust her.

Third, what if Bob creates one and thinks it didn't work, so he creates a second one.  Isn't he now seen as an attacker on himself?

mikedilger:
> if events can disappear from all relays then Nostr isn't a censorship-resistant network!

They don't have to disappear from ALL relays, just the ones I happen to talk to.

Nostr is censorship resistant only in so much as it doesn't have a single point of failure.

braydonf:
> First, if I am Alice, how do I know that I should attest to Bob's key recovery setup event? Do I call him on the phone? Is it just because I have never seen one before? Because maybe an Imposter sends a second one, but that is the first one I have seen and so I attest to the wrong one.
> 
> Second, in your scenario Alice becomes aware of an imposter recovery key setup event. But how does the rest of the network learn? Most people don't follow Alice, and even if they could they don't trust her.
> 
> Third, what if Bob creates one and thinks it didn't work, so he creates a second one. Isn't he now seen as an attacker on himself?

@mikedilger Great questions. 

1) This specification doesn't specify how Alice determines which key is truthful, it does provide several ways to help to determine which one is honest (NIP-05 DNS identifiers, NIP-03 opentimestamps and attestations in this NIP). Offline and side-channel verification should be considered too! Especially if it's not clear and there are multiple key setup or key migration and revocation events.

2) Attestations from Alice will only be known and relevant to people who follow Alice. It would be important for there to be many attestations from different identities to have greater reach in different social graphs. This is a benefit, an attacker could try to create many false identities and make fraudulent attestations to a false recovery key, however if nobody follows these false identities (aside from the attacker), it has zero relevance.

3) If Bob accidentally (or willfully) creates multiple honest key setup events or key migration and revocation events then it will be important for him to communicate and help clarify which one is the honest one. In this case the NIP-03 opentimestamp would be less relevant and the NIP-05 DNS identifier would be more relevant. Furthermore, the more attestations there are from shared contacts for one over the other should also help. If all else fails, then falling back to offline or side-channel verification can and should be used.

vitorpamplona:
I might be too dumb for these things but I don't feel very comfortable implementing this. 

The reason I thought #1056 would be good is exactly because it didn't touch any of the key attestation and verification procedures. It just passed ALL that responsibility to users directly: "Nostr cannot help you. Relays and clients do not have the information you need. You need to pick up a phone and find person X to ask what is the new key they are going to use. And you cannot use Nostr for that. It's all on you."

Here, because there is a "nostr" way of doing it, an attacker can use any of these events to confuse users into believing a key is migrating. Because there are so many kinds, relay checks, clients with varying levels of compliance, replaceable events and all of which can happen in the past, present and future, with expiration and event deletions, in separate relays that don't talk to each other or relay that are fully controlled by the attacker, etc... it creates more problems than what it might solve. 

Even with this, in the end, it boils down to Web of Trust. This PR just codifies that into a sequence of Nostr events that apps can verify without the User having to call a friend. That's the delta improvement here. 

Is the complexity worth that delta? And can we expect Clients to implement all of this without bugs and in a UX that match each other to avoid further confusion and a potential form of attack?

I don't know. It might be that this NIP is trying to do too much.

braydonf:
I've started implementing the events for this NIP in go-nostr and can be found at:

https://github.com/braydonf/go-nostr/tree/key-migration-and-revocation/nipxx

This NIP has been updated to reflect changes based from that implementation. Several clarifications and changes have been made.

braydonf:
> "Nostr cannot help you." 

@vitorpamplona I think it can help. If someone that I follow has their private key compromised, I can still trust by own events.

As far as implementation, it helps me to better understand the problem and to discover a solution that I may not have previously seen.

For example, here are some things that I'm thinking about after my first round of implementation and will be addressing in another set of updates:

1) Is it necessary to have the _Key Migration Attestation Event_ for a _Key Migration and Revocation Event_? This could be done, as similar to [#1056](https://github.com/nostr-protocol/nips/pull/1056) by changing the follow list (`kind 3`) and mute list (`kind 10000`). While it isn't codified as an attestation, it is functionally similar. It might be more simple to have an event for it rather than it being necessary to pull follow and mute lists. However, the best solution should clear up with further implementation.

2) In the event that a private key is compromised, the NIP-05 in the profile, the name, the profile image and etc are all likely to change to something evil. It may become nearly impossible to even remember who the original profile was, I certainly don't have npubs memorized. Therefore, I think the _Recovery Keys Attestation_ is not quite enough and it needs to expand to be not only an attestation of the Recovery Keys, but an attestation of the profile name, NIP-05, domain, and other contact points, such that I can remember who the profile was before the key was compromised.

braydonf:
I've made several simplifications and updates to this NIP and it feels much closer to a solution!

Next, I'll work on implementing it more in go-nostr and updating the UX design wireframes.

braydonf:
Okay, here are some of the UX wireframes for secure identities, as now included in this NIP.

ux-storyboard-animation-layout

Link to animation:
https://raw.githubusercontent.com/braydonf/nostr-key-migration-and-revocation/89a9437e958f44bf748f145b85e603d83cf7ef2f/graphics/ux-storyboard-animation.webp

braydonf:
I've realized that the user metadata attestation portion of his drafted NIP can have a lot of other uses beyond the scope defined in this NIP and have separated it into another (see [#1499](https://github.com/nostr-protocol/nips/pull/1499)). I will leave this NIP, as drafted, for the time being.

The remaining portions, the revocation piece can likely merge with @vitorpamplona [#1056](https://github.com/nostr-protocol/nips/pull/1056). It would rename it Key Revocation and not Key Migration (and mostly leave out the migration/transfer portion). The migration keys part can be another NIP of it's own that can define the scheme for pubkeys on a user's metadata and methods to verify them.

vitorpamplona:
Yeah I think that is a good move. I updated the text on https://github.com/nostr-protocol/nips/pull/1056 to match the revocation ideas from here.

braydonf:
Cool, sounds good.

I'm convinced this will all work and would like to go through and document how in more detail. Various scenarios and attacks can be enumerated and how each is mitigated. This should also help to find any issues, if they remain. 

Implementation seems fairly straight forward so far, I don't foresee any issues here. It'll still be useful to explore to uncover anything unexpected in both clients and relays, that can likely follow.

xx.md
```diff
@@ -0,0 +1,153 @@
+NIP-X
+=====
+
+Key Revocation and Migration
+------
```
kehiy, line 5:
```suggestion
----------------------------
```

xx.md
```diff
@@ -0,0 +1,153 @@
...
+
+There is one new optional field (`migration_pubkeys`) for a user's metadata (`kind 0`):
+
+* [Migration Keys](#migration-keys)
+
+## Key Revocation Event
+
+This is a regular event with kind `50`. It will revoke a public key _(it has been compromised)_ and stop future events from the key _(as determined as after the event was received)_. It will also provide a means to inform followers of the compromise and suggest a successor public key to follow.
+
+```js
```
kehiy, line 28:
```suggestion
```json
```

xx.md
```diff
@@ -0,0 +1,153 @@
...
+  * The `d` tag MUST be an encrypted and hashed version of the public key (hex encoding of a sha256 hash of an encrypted, with NIP-44, of the public key).
+  * The `p`, `metadata` and `attestations` tags, as the same as the public attestation, MUST be JSON stringified and NIP-44 encrypted in the content field.
+
+## Migration Keys
+
+This is a new field on a users metadata (`kind 0`) event with a key of `migration_keys`. Its purpose is to define a set of migration keys that can be used to help migrate to a successor key in the future. The event can assign anywhere from `1` to `n` migration keys. A threshold number of keys (`m` of `n`) can be assigned to verify this event.
+
+The value should be as follows:
+
+```js
```
kehiy, line 131:
```suggestion
```json
```

or just remove it. since most nips don't use a ```json for arrays.

xx.md
```diff
@@ -0,0 +1,153 @@
...
+* The array SHOULD be all strings, including the threshold.
+* Clients MAY present a user interface to make an attestation, if this field is available on the metadata.
+* Clients MAY use hardware devices and NIP-06 seed phrases to store the migration keys.
+
+### Revocation Event Signing
+
+A _Key Revocation Event_ MAY be signed with `m` of `n` of these migration keys. This can help assist the migration process for those that have previously attested to the `migration_keys`.
+
+The following `migration-sigs` tag MAY be included:
+```js
```
kehiy, line 144:
```suggestion
```jsonc
```

xx.md
```diff
@@ -0,0 +1,153 @@
...
+{
+	"kind": 50,
+	"tags": [
+		// other tags
+		["migration-sigs", "<index-0-sig", "<index-1-sig>", "<index-2-sig>"]
+	]
+	// other fields
```
kehiy, line 151:
```suggestion
	// other fields...
```

xx.md
```diff
@@ -0,0 +1,153 @@
...
+### Revocation Event Signing
+
+A _Key Revocation Event_ MAY be signed with `m` of `n` of these migration keys. This can help assist the migration process for those that have previously attested to the `migration_keys`.
+
+The following `migration-sigs` tag MAY be included:
+```js
+{
+	"kind": 50,
+	"tags": [
+		// other tags
```
kehiy, line 148:
```suggestion
		// other tags...
```

kehiy (commented):
some small changes. makes everything look better.

xx.md
braydonf: I'll look at updating some of the formatting in another update. Some of the examples could also be expanded and improved. :)

PR #1455, opened 2024-08-29
NIP-05: add identification vs verification note

dtonon:
One clarification that I think it is useful to add in the NIP, as a result of [this thread](https://njump.me/nevent1qqsrpvvhnu5fu8gaudnlnuk7cdp3m5yezl6w88x9t38c3zdhaju52tcpzamhxue69uhky6t5vdhkjmn9wgh8xmmrd9skctcpr9mhxue69uhkxcfwwp6hyurvv4ex2mrp0yhxxmmd9uqjqamnwvaz7tmrdphhyatn9ekkj6m9v35kcem9wghxxmmd8g6rgdp0wukeu8).

### DIFF
--- a/05.md
+++ b/05.md
@@ -58,6 +58,15 @@ A client may implement support for finding users' public keys from _internet ide
 
 ## Notes
 
+### Identification, not verification
+
+The NIP-05 is not intended to _verify_ a user, but only to _identify_ them, for the purpose of facilitating the exchange of a contact or their search.  
+Exceptions are people who own (e.g., a company) or are connected (e.g., a project) to a well-known domain, who can exploit NIP-05 as an attestation of their relationship with it, and thus to the organization behind it, thereby gaining an element of trust.
+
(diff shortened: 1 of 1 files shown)


fiatjaf (approved):

melvincarvalho:
NIP-05 shouldnt be changed in 2 days based on a very selective, opinionated nostr thread.  It's one of the most stable NIPs in nostr.  NIP-05 absolutely does provide 2 way verification.  Strongly against.  Please revert.

fabianfabian:
> NIP-05 shouldnt be changed in 2 days based on a very selective, opinionated nostr thread. It's one of the most, multi-year, stable NIPs in nostr. NIP-05 absolutely does provide 2 way verification. Strongly against. Please revert.

over that same period when someone mentions its use as verification there is always someone correcting them that its not verification, the NIP is still stable as in there is no technical change, not sure what your goal is in reverting this, do you want to keep people using it the wrong way?

PR #1458, opened 2024-08-30
NIP-06: adding multi account management

kehiy:
This PR adds a new section to nip-06 which introduces an implementation possibility for the client and signer to manage multiple accounts related or unrelated.


dluvian:
This is already described in line 13

kehiy:
@dluvian In this section I explained this potential implementation in detail. maybe we can remove line 13?
Line 13 not clear enough I think. We need more opinion and votes.

PR #1459, opened 2024-08-30
NIP-65: rename "Relay List Metadata" to "Outbox model"

AsaiToshiya:
The "Outbox model" is familiar to many nostriches. And for beginners who don't know this NIP, there is no significant difference between the names "Outbox model" and the "Relay List Metadata".

### DIFF
--- a/65.md
+++ b/65.md
@@ -1,8 +1,8 @@
 NIP-65
 ======
 
-Relay List Metadata
--------------------
+Outbox model
+------------
 
 `draft` `optional`
 
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ They exist to document what may be implemented by [Nostr](https://github.com/nos
 - [NIP-58: Badges](58.md)
 - [NIP-59: Gift Wrap](59.md)
 - [NIP-64: Chess (PGN)](64.md)
-- [NIP-65: Relay List Metadata](65.md)
+- [NIP-65: Outbox model](65.md)
 - [NIP-70: Protected Events](70.md)
 - [NIP-71: Video Events](71.md)
 - [NIP-72: Moderated Communities](72.md)


vitorpamplona:
LGTM

staab (approved):

kehiy:
@AsaiToshiya Why do we have to rename it while the `Relay List Metadata` is a more self-describing name?

AsaiToshiya:
No one calls NIP-65 "Relay List Metadata". A commonly recognized name is more important than a self-describing name. Besides, either way, the name doesn't tell everything.

kehiy:
No one means the current users and developers on the nostr. but what about someone new who checks these NIPs or see this names in clients and docs???

AsaiToshiya:
Will someone new not read the document? No, I think they will read it if they are interested.

pablof7z:
NACK

Outbox model and NIP-65 are not the same thing; NIP-65 is a tool, among multiple others (like relay/pubkey hints) that inform the outbox model.

erskingardner:
I agree on the NACK. Shouldn't change.

dluvian:
What about something like "In- and Outbox Relay List" ?

The word "Outbox model" is being used a lot but people can't find anything about it in this repo. So I agree that we should rename this nip to something that includes "Outbox", as it is the main tool for implementing the outbox model.

monlovesmango:
NACK for me too.

>A commonly recognized name is more important than a self-describing name. 

I dunno. kind 0 is called user metadata but colloquially its referred to as the user profile, but I guess you might feel that this should also be renamed.

>Besides, either way, the name doesn't tell everything.

sure, but imo "Relay List Metadata" tells more than "Outbox Model".

mikedilger:
I'm also NACK.

I think it is fine as is.  I get the point though.  If we wanted these titles to be ELI5 we could just call this "How a user declares which relays they use, and therefore where you can find them" or for short "User Relay List".  The term "Metadata" was added to the original title "Relay List" as a compromise for people who wanted to put this data into kind-0, but it just makes the title more confusing IMHO.   Outbox model is a common phrase but doesn't describe what the hell it is.

So my preference is "User Relay List" or just "Relay List" or just keep "Relay List Metadata".  Or something like that.

melvincarvalho:
NACK Outbox is not quite the same things as a Relay List.  Outbox needs a lot of testing to even see if it is going to be the optimal model.  Relay Lists are future proofed because we will always need lists of relays.  Furthermore, relay lists are in documentation and code in dozens or places, and would require many pull requests to rebrand, which is a pain.

fabianfabian:
I prefer "User Relay List" too, or "User Preferenced Relay List". Metadata is the confusing part, I consider metadata created_at/pubkey/kind etc.

staab:
I agree metadata is a stupid word. I do think outbox is equivalent to NIP 65, general relay selection being a broader problem, so the change is semantically fine in my book. But I don't feel strongly either way. Closing, since consensus is clearly to keep this as is.

AsaiToshiya:
OK. Alright, thanks everyone for the discussion!

PR #1460, opened 2024-08-30
NIP-69 Nostr Offer STRings

shocknet-justin:
Static payment codes to replace LNURL-P

edit: This NIP now has a companion, NIP-68, to allow the inverse operation of ad-hoc debits

https://github.com/nostr-protocol/nips/pull/1529


69.md
```diff
@@ -0,0 +1,220 @@
...
+  0: <receiver_public_key_in_hex>
+  1: <relay_url>
+  2: <offer_id_string> 
+  3: <pricing_type_flag> (optional, 0 for fixed, 1 for variable, 2 for spontaneous)
+  4: <price_in_sats> (optional)
+```
+
+### NIP-05 Integration
+
+To support trust-optimized Lightning Addresses, services can add a `nip69` field in the NIP-05 content to contain an offer for spontaneous payments. This field will provide the offer on name-based lookups.
```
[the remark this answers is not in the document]
shocknet-justin: I think i'm following... pushed changes lmk if that's what you meant

staab:
So I'm not well versed in the details of lightning, but it looks like this at least partly overlaps with bolt12. Why not use that instead?

shocknet-justin:
> So I'm not well versed in the details of lightning, but it looks like this at least partly overlaps with bolt12. Why not use that instead?

Bxlt12 is effectively a re-implementation of Tor which has proven too unreliable with Lightning as bouncing onion messages across hops adds latency and failure probability. It also can't be used trivially in web apps, where Nostr excels and most economic activity happens. 

There's also the separation of concerns in that Lightning is a payments network, not a communications network. Muddling the two like bxlt12 is scope creep for implementations that now would need to provide account related hooks, something that's much more straightforward in a Nostr middleware. Nostr offers are also portable across nodes unlike bxlt12 (and it's keysend predecessor). Reality is that the prevailing implementation has had keysend for years and that coupling of LN and Identity is not optimal. Same reason LNURL-Auth had so many issues. Bxlt12 is just a bad design and manufactured hype by the minor implementations looking to gain some share as they trapped themselves into trying to turn phones into servers.

69.md
```diff
@@ -47,12 +47,26 @@ noffer1...
...
+  "pubkey": "hex_pub",
+  "kind": 0,
+  "content": "{\"name\": \"bob\", \"nip05\": \"bob@example.com\", \"nip69\": \"noffer1...\"}"
+  ...
+}
+```
+
+Example NIP-05 Service Response with Offers
+
+
```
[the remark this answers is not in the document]
shocknet-justin: Restructured all that now for nip 01, 05 and 57

shocknet-justin:
**This is now implemented for additional testing in [ShockWallet](https://github.com/shocknet/wallet2/pull/268) and [Lightning.Pub](https://github.com/shocknet/Lightning.Pub/pull/727)**

A simpler demo client is also available at [demo.nip69.dev](https://demo.nip69.dev)

**How we are using this to provide self-custodial Lightning Addresses:**

Shockwallet will automatically claim on users behalf a shockwallet.app Lightning Address from a LNURL bridge server. The bridge server will serve legacy LNURL clients by using NIP69 as the method of communication for invoicw fetching from self-hosted Lightning Pubs. NIP69 aware clients will simply see this value and fetch the invoice directly as if the offer were NIP05 metadata.

69.md
```diff
@@ -0,0 +1,293 @@
...
+   - This step is out of scope for this NIP, but potential receipt considerations include:
+     - Nostr Event as Receipt: Emitting a public Nostr event as a "zap" receipt for verifiable record-keeping.
+     - Lightning Pre-Image: The Lightning pre-image is proof of payment.
+     - External / WebHook: The receiver may finalize the flow out-of-band.
+
+### Nostr Events
+
+This NIP specifies the use of event kind `21001` with the following structure:
+
+- `content`: NIP-44 encrypted payment details.
```
jklein24, line 131: Is it worth gift-wrapping these events (NIP-59/17) to increase privacy and reduce metadata leaks? How are the privacy comparisons here to bolt-12 with blinded paths? Is the thinking that the sender could always just use a new anonymous keypair for a payment if they want to?

69.md
```diff
@@ -0,0 +1,293 @@
...
+### Static Payment Code Format
+
+The static payment code is a bech32 (per [NIP-19](19.md)) encoded string prefixed with `noffer`. The encoded string will include the following TLV (Type-Length-Value) items:
+
+- `0`: The 32 bytes of the receiving service's public key, encoded in hex.
+- `1`: The relay URL where the receiving service subscribes to payment requests.
+- `2`: The offer identifying string.
+- `3`: A flag indicating the pricing type:
+  - `0`: Fixed price
+  - `1`: Variable price
```
jklein24, line 28: Maybe I'm missing something, but what does this mean? Should it be defined in here somewhere?

69.md
```diff
@@ -0,0 +1,293 @@
...
+  - `2`: Spontaneous payment (payer specifies the amount)
+- `4`: The price in sats (optional for display purposes).
+
+If neither the price nor the pricing type flag is present, the sender may assume it is a spontaneous payment offer.
+
+Example static payment code structure:
+
+```
+noffer1...
```
jklein24, line 37: It might be worth considering adding requested payerdata fields here like LUD-18 does. This allows the receiver to specify that they want certain information about the payer for UX reasons, etc.

69.md
```diff
@@ -0,0 +1,293 @@
...
+}
+```
+
+The `content` field, after NIP-44 decryption, should contain an object with the following structure:
+
+```json
+{
+    "offer": "<offer_string>",
+    "amount": "<conditional_amount_in_sats>",
+    "payer_data": "<optional_payer_data>"
```
jklein24, line 158: Is this structures as a json blob with field name strings like LUD-18?

69.md
```diff
@@ -0,0 +1,293 @@
...
+}
+```
+
+Example response event with `bolt11` invoice:
```
jklein24, line 162: Check out this [LUD-22 payee data](https://github.com/lnurl/luds/pull/252) proposal. It may be useful for the receiver to be able to dynamically communicate free-form data back to the sender if they request certain fields, etc.

69.md
```diff
@@ -0,0 +1,293 @@
...
+
+Example static payment code structure:
+
+```
+noffer1...
+  0: <receiver_public_key_in_hex>
+  1: <relay_url>
+  2: <offer_id_string> 
+  3: <pricing_type_flag> (optional, 0 for fixed, 1 for variable, 2 for spontaneous)
+  4: <price_in_sats> (optional)
```
jklein24, line 42: As a future extension, it would be awesome to think about specifying amounts in other currencies (although payment is still over lightning). Bolt-12 has a mechanism for this and LNURL has [LUD-21](https://github.com/lnurl/luds/pull/251) which has some decent usage now.

jklein24 (commented):
Love this! Very cool stuff @shocknet-justin

69.md
shocknet-justin: Exactly yea... for those not also on Telegram:

I think its much better, bolt12 has to onion things because its correlating the comms and the lightning node, which with the combined metadata can reveal a lot

With nostr, the comms are separate from the Lightning Node, a nostr key could service many nodes or one lightning node could have many nostr keys.

The nostr keys for applications/wallet services aren't inherently linked to social either, they're meant to be quasi-ephemeral. The giftwrapping/seal nips don't quite make sense to me as far as solving the handshake problem, even simplex is just doing that out of band with ephemeral keys

69.md
shocknet-justin: ty I Just pushed comments on these

69.md
shocknet-justin: I considered that quite a bit and decided to compromise at just calling it optional payer data on the request side and added the marketplace example with addressable events. I wanted to keep this in scope specifically for wallets, with the apps that use being expected to layer in more shop-level functionality.

69.md
shocknet-justin: stringified per typical event payloads, just pushed a clarification

69.md
shocknet-justin: Yea as mentioned above I leaned into keeping the scope wallets only, but also leaving room for upgrades... particularly in the context of nostr as opposed to lnurl, nostr itself is a better place to tie those requirements to a product/service addressable event rather than link encoding (will probably break QR's etc)

69.md
shocknet-justin: Agreed, I thought through a bit already but haven't actually worked through an implementation...

Thinking is doable in this version, you would make a variable price offer, set a `price_usd` (or whatever unit) as an optional value, the service would calculate that on invoice generation, and then the wallet could take the response and check exchange rates to ensure its in the ballpark of what the offer had quoted

shocknet-justin: Added reference to fiat units in relation to variable pricing

69.md
vitorpamplona: Agree, I don't think that level of privacy is needed. But if you are concerned with relays tracking users by IP then you could use the gift wrapping spec to route the message through a second trusted relay or DVM as a native onion routing technique.

69.md
```diff
@@ -0,0 +1,293 @@
...
+The static payment code is a bech32 (per [NIP-19](19.md)) encoded string prefixed with `noffer`. The encoded string will include the following TLV (Type-Length-Value) items:
+
+- `0`: The 32 bytes of the receiving service's public key, encoded in hex.
+- `1`: The relay URL where the receiving service subscribes to payment requests.
+- `2`: The offer identifying string.
+- `3`: A flag indicating the pricing type:
+  - `0`: Fixed price (Price stated in 4 will not change)
+  - `1`: Variable price (Price must be calculated and will be reflected in the invoice, useful for products priced in fiat)
+  - `2`: Spontaneous payment (Payer specifies the amount in request payload)
+- `4`: The price in sats (optional for display purposes).
```
ekzyis, line 30: Afaict, this means that one can only specify invoices in sats. It would be nice if you could consider to use msats here. Explanation:

[LUD-06](https://github.com/lnurl/luds/blob/luds/06.md), [NIP-47](https://github.com/nostr-protocol/nips/blob/master/47.md) and [NIP-57](https://github.com/nostr-protocol/nips/blob/master/57.md) also specify the amount in msats. It would also be good for us at [Stacker News](https://stacker.news) if this spec supports msats out of the box because we forward payments between users and take 10% as a fee for sybil resistance and rewards. This means that if someone zaps 25 sats, 22.5 sats will be received which requires being able to specify invoices in msats.

Ideally would be amount + unit like [BOLT11](https://github.com/lightning/bolts/blob/master/11-payment-encoding.md) though but msats is probably good enough and would be consistent with other specs as mentioned.

I have to mention thought that most LNURL providers seem to simply round to nearest sat so don't really support msats; contrary to the spec.

shocknet-justin: > LNURL providers seem to simply round to nearest sat 

Yea I considered this but steered away from assuming any baggage from LNURL... the issue with millisats is that they can't actually be settled in payments (technically they are shitcoins and thus rounding happens regardless of whether we use them or not). The scope operates on the assumption that offers will be used in sales reports and things of that nature which do not go to decimal places of a penny etc and so I didn't re-introduce that here.

69.md
shocknet-justin: marketplace addressable events

shocknet-justin:
> Do you have a source for this? I don't know anyone working on Lightning that thinks this.

You can find hundreds, maybe thousands of people debugging their Lightning node across socials over the years where the ultimate solution was disabling Tor. There's too much latency and failure probability to make it reliable enough for real world use. 

> This seems like a pretty poor reimplementation of BOLT 12 over nostr that loses some of the important features of BOLT 12 (like recipient privacy!). If you're insistent on not using Lightning onion messages (and getting the privacy that comes with them), why reinvent the wheel? Just shove BOLT 12 messages over Nostr?

It's not a re-implementation of BOLT12 at all, and that's intentional. This also makes it far more private because the keys are completely decoupled from the payment layer, Nostr ensures that comms and payment concerns are properly separated. If a recipient wants to use blinded paths/rendezvous/randomized scids in the bolt11 invoice they can already do that. 

Nesting OM's into yet another overlay network would make performance an order of magnitude worse than it already is while at the same time delivering only what's needed where it is needed: the web. 

That said, optional fields mean a wallet service COULD return a bolt12 offer just as they would a bolt11 invoice, but there's no logical reason why they would.

vitorpamplona:
> optional fields mean a wallet service COULD return a bolt12

We hate optional fields. Just pick one and go for it. If it doesn't work, we can always create a new PR/NIP.

shocknet-justin:
> arguments around how Tor sucks largely don't apply to lightning onion messages

Except that they do, what matters is proximity and availability of peers between you and the payer. Channel peers in Lightning vs. arbitrary peers in Tor is better, sure. But optimizing for those as your reverse proxy adds yet another consideration for users who are struggling to grasp channel management already. 

Back and forth aside, still need to get the response out to a browser in most cases, and offers with any external utility will have already been orchestrated by middleware.

> Nostr and privacy are....not a thing. If the nostr relay wants to deanonymize you, they can.

Bolt12 isn't a tool for IP obfuscation, you have to connect to other things on the internet in either case. At least here, your keys aren't linked to anything, and are still optionally gift-wrappable as Vitor pointed out. If traffic level inspection is level of concern we're talking then in that context Nostr has the added benefit of blending in with everything else that uses https. Nostr can do anything onions can do, but better.

>  learn which nodes are online and ready to forward the payment

I agree this is a very good byproduct but of what is otherwise useless traffic. LND is doing something similar iiuc by flagging channels as inactive. Not sure about others. This does lend credit to the theory that, as Lightning and Nostr move closer together, information from the Nostr social graph will pull Lightning route construction up a layer.

shocknet-justin:
Video demo 

https://cdn.lightning.video/bd926eaa77a9a1046afb12faabb268be925b01d952edc3ce19d092f258ea3572/4WyiE6SuFQ.mp4

dhalsim:
This looks very promising, but let me summarize what I understood. 
Please correct me if I'm wrong.

Does it mean I can:

1. Run my own LND node locally not connected to public network
2. Run a service locally that listens to NIP-69 events for zaps, talks with my LND node
3. Creates and sends bolt11 invoices over a nostr relay to communicate

So that means, I can have self custody lightning without using tor and a LNURL service, a wallet can pay directly to a NIP-05 address by checking its nip69 field in the content.

But we still need to support Lightning Addresses because it is already popular by so many applications and wallets. So you built a bridge, that when there is a "nip69" field, supported wallets can upgrade the payment to nostr. Unsupported wallets can still pay using lnurl-p but underneath the connection between the LND node happens over nostr.

And I have some questions, too

1. what are the privacy implications over tor connections? Is NIP-44 encryption enough?
2. The bridge itself is still a burden. What is the advantage of using NWC over your bridge? 
3. Do you think people stop using LNURL altogether and switch to nostr?

shocknet-justin:
This NIP now has a companion, NIP-68, to allow the inverse operation of ad-hoc debits

https://github.com/nostr-protocol/nips/pull/1529

Also a fork of nostr-tools has been provided

shocknet-justin:
Spec has moved to https://github.com/shocknet/clink

an SDK is available at https://www.npmjs.com/package/@shocknet/clink-sdk

PR #1462, opened 2024-08-31
Add functionality to exclude specified relays from deletion requests

0ceanSlim:
This PR proposes an update to NIP-09 by introducing a new exclude tag to the Event Deletion Request specification. The exclude tag allows users to specify relays that should not delete or stop publishing the events referenced in a deletion request.

### DIFF
--- a/09.md
+++ b/09.md
@@ -6,7 +6,7 @@ Event Deletion Request
 
 `draft` `optional`
 
-A special event with kind `5`, meaning "deletion request" is defined as having a list of one or more `e` or `a` tags, each referencing an event the author is requesting to be deleted. Deletion requests SHOULD include a `k` tag for the kind of each event being requested for deletion.
+A special event with kind `5`, meaning "deletion request" is defined as having a list of one or more `e` or `a` tags, each referencing an event the author is requesting to be deleted. Deletion requests SHOULD include a `k` tag for the kind of each event being requested for deletion. The exclude tag is used to specify relays that are excluded from the deletion request. Relays listed in this tag SHOULD NOT delete the referenced events and SHOULD continue to publish them.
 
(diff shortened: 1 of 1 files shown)


0ceanSlim:
Could also be a new nip with a new kind to specify a list of relays that shouldn't store your events and delete any that they have from you.

fiatjaf:
NIP-70 should address this in a more elegant way, I think.

0ceanSlim:
> NIP-70 should address this in a more elegant way, I think.

This is good but it also requires your relays implement NIP42. The motivation for this change was so I can request all relays delete all my previous notes but keep them in a handful  of relays. I don't want to delete them off certain private relays. I'm sure a lot of people are in the same boat as me with having added a lot of relays when first using the protocol. I don't want my old notes stored on old relays I don't use anymore and as an added bonus, those operators can free up some space. This change would allow me to make some sort of tool to send a delete request for all previous events from my Pubkey but keep them on my private relay. Then going forward it's easier to keep track of where all my data could potentially be. NIP70 wouldn't let me go back and change non replaceable events as protected. Of course the longer I think about the issue less sure I am about the best solution. There's just no way to selectively delete events from some relays and not others.

vitorpamplona:
> The motivation for this change was so I can request all relays delete all my previous notes but keep them in a handful of relays.

That will be better achieved with something like this https://github.com/nostr-protocol/nips/pull/1256

0ceanSlim:
> > The motivation for this change was so I can request all relays delete all my previous notes but keep them in a handful of relays.
> 
> That will be better achieved with something like this https://github.com/nostr-protocol/nips/pull/1256 

Ah yes, I was thinking of the right to vanish last night aswell shortly after making the PR. I'll need to review that a little later today to see how it might fit there.

PR #1466, opened 2024-08-31
nip44: clarify ecdh hashing

paulmillr:
Community member wasted time because of that.

### DIFF
--- a/44.md
+++ b/44.md
@@ -142,6 +142,7 @@ validation rules, refer to BIP-340.
     The operation produces a shared point, and we encode the shared point's 32-byte x coordinate, using method
     `bytes(P)` from BIP340. Private and public keys must be validated as per BIP340: pubkey must be a valid,
     on-curve point, and private key must be a scalar in range `[1, secp256k1_order - 1]`.
+    Some libraries, like libsecp256k1, apply sha256 on the output. The hashing is unnecessary in NIP44.
 - Operators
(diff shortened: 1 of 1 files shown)


staab:
I don't have the audit in front of me, but I believe it said not hashibg was ok, but it would be better to hash.

paulmillr:
audit https://github.com/paulmillr/nip44/blob/main/audit-2023.12.pdf

I don't think there was anything like that. In any case we don't hash it and we don't plan to.

prolic:
@paulmillr Thanks for the assitance

staab:
Ok, thanks, I skimmed the report as well and didn't see anything.

vitorpamplona:
Ack. I remember having the same question when I was coding and having to look at the raw code to find which one matched what @paulmillr was asking for. Tweak_mul was it. 

That being said, since this error seems common and the secp lib is where people are going to look, I do think the PR could mention the function names to avoid any additional misunderstanding of this. If function names change, we could change the NIP note as well.

mikedilger:
I would change "unnecessary" to "unnecessary and incompatible" to be even more clear.

Other than that,
ACK

paulmillr:
updated

staab (approved):

vitorpamplona (approved):

prolic:
ACK

prolic (approved):

PR #1467, opened 2024-09-01
Add clarification to NIP44

prolic:


paulmillr:
Already submitted it in #1466. I don't think we should mention detailed function names because they can change and the library itself is not part of nostr or nip44.

Issue #1442, opened 2024-08-25
Adding parse mode tag


kehiy:
I think events can have a parsing mode like: `html5`, `markdown`, `pure/plain`, and more.

So we can use different parsing modes and embed different stuff on an event or even make webpages with an event and more.

This is the basic idea, If it's acceptable, I can work on NIP related to it.

fiatjaf:
This is against criteria 4 for acceptance of NIPs and a big mistake, I think.

https://fiatjaf.com/27598e6f.html

kehiy:
@fiatjaf where can I find list of criterias for acceptance of NIPs?

I read this article and it was logical for most cases. But in the case of parse mode, I think it don't need to be arbitrary parse modes from any client. Parse modes will be defined in the NIP and the is a low chance to add a new parse mode. But on the other hand we can support webpages and MD stuff in notes integrated on every client that implements this NIP.

What do you think? 

Over all that's not something for arbitrary value. That's a field which helps client to understand how to parse the note and at most it's some specific values such as HTML and MD.

fiatjaf:
It's on the README. It's not a super hard criteria, but still.

I think I see what is your idea here and I can imagine scenarios in which we could have such tag, but they would be worst-case scenarios in which the protocol has already been bloated by added flexibility (i.e. some big corporate entity is capturing the network and has started producing HTML content and others are rushing to follow). Just adding this tag here is inviting extra work it for no reason and a move that can't be justified for me.

Another interesting read: https://this.how/standards/

kehiy:
This post about standards was intresting and makes sense. But I still think we can have more disscustions on this. The current reason are logical but I think I can put more time on this and work on this idea details. I prefer to keep this open for a while and work on that more.

vitorpamplona:
Different event kinds can have different formats. For instance, https://github.com/nostr-protocol/nips/pull/811 hosts HTML, CSS and JS files in the `.content` of each kind. 

Mixing formats in a single kind is not a good idea.

kehiy:
@vitorpamplona So, do you think specific kinds with specific parse modes are good enough to use all the potential behind this idea?

Otherwise, that's logical to close this for me.

vitorpamplona:
Yes, We have not seen the need for multiple formats in the same NIP/event kind.

kehiy:
@vitorpamplona Got it. Thanks a lot.

PR #1468, opened 2024-09-01
NIP-22 Smart Widgets

vir2alexport:
Introducing Smart Widgets


22.md
```diff
@@ -0,0 +1,65 @@
+# NIP-22 
+
+## Smart Widgets
```
kehiy, line 3: It's better to set the NIP header the same as other NIPs.

```suggestion
Smart Widgets
-------------
```

22.md
```diff
@@ -0,0 +1,65 @@
+# NIP-22
```
kehiy, line 1:
```suggestion
NIP-22
======
```

22.md
```diff
@@ -0,0 +1,65 @@
+# NIP-22 
+
+## Smart Widgets
+`draft` `optional`
+
+Smart Widgets are customizable, interactive components that can be embedded within Nostr events and allow to create rich, dynamic content. This proposal outlines the structure, implementation, and potential use cases of Smart Widgets.
+
+## Motivation
+Traditional static content limits user interaction and engagement. Smart Widgets provide a way to embed dynamic components, such as buttons, images, and polls, directly into Nostr events, enabling more interactive and responsive content.
```
kehiy, line 9: To support such a thing we can simply use HTML. put some HTML5 in `.content` and then the client can parse it like a webpage or built-in Telegram browser that we can see a lot nowadays.

But overall about parsing HTML, we have a closed issue which I suggest you to check.

- #1442

It can bring the potential of embedding games/web pages on events as well. A cool idea. But based on the above issue we have reasons to avoid that. IMO we can simply support this on some specific clients with no standards. If they find important use cases we can make them standard.

kehiy (changes requested):

22.md
vir2alexport: This pr is all about letting clients decide what they need: `html` or `html5` or `xml` or `svg` or `json` or pretty much whatever. For example, if yakihonnes wants stringified json they can use it without forcing everyone to accept it as a standard.

Regarding that issue, I don’t understand why whould you need a separate kind for each, instead ordering clients to redirect in case of no support for a specific mime type by using `naddr` and the `client` used for creating smart widget.

Beyond that, I completely agree with you.

vir2alexport: Few thoughts to clarify my view and avoid misundertanding. IMO this pr is different from #811 . Adding scripts is unnecessary for this specific pr NO SCRIPTS. Adding games and webpages with mixed formats is too much. The idea is great, no doubt, but smart widgets are by no mean a decentralized web hosting on Nostr. Smart widgets don't need variables like css3 or javascript etc. These are just dynamic interactive images that user can click on. Forms that user can submit and other parsable content.

PR #1477, opened 2024-09-03
format(all): JSON formatting

kehiy:
The JSON format is used commonly in different NIPs and the correction of JSONs and the correct Markdown format is important when someone reads them or uses them for implementation, testing, and more.

In this PR I've just fixed some broken JSON formats and used `jsonc` for JSON examples with comments to make all of them integrated with NIP-01!

### DIFF
--- a/96.md
+++ b/96.md
@@ -143,7 +143,7 @@ The upload response is a json object as follows:
   // and, optionally, all file metadata the server wants to make available
   //
   // nip94_event field is absent if unsuccessful upload
-  "nip94_event2":{
+  "nip94_event2": {
     // Required tags: "url" and "ox"
     "tags": [
       // Can be same from /.well-known/nostr/nip96.json's "download_url" field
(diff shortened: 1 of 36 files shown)


96.md
```diff
@@ -143,9 +143,9 @@ The upload response is a json object as follows:
   // and, optionally, all file metadata the server wants to make available
   //
   // nip94_event field is absent if unsuccessful upload
-  nip94_event: {
+  "nip94_event2":{
```
fiatjaf, line 146: Is this correct?

fiatjaf:
This is pretty good. Thank you very much.

96.md
kehiy: Based on online formatters it seems yes. But it won't affect how encoders interpret it. so I think we can go with the same way like all other JSONs in this repo.

kehiy: Fixed.

kehiy:
@fiatjaf You'r welcome. Changes were applied and one missed fixed corrected.

kehiy:
Some notes on this PR and #1471  :

These kind of changes make NIPs more clean and easy to read. but we have to consider we should keep in mind to review this stuff to avoid huge refactors like this in the future. A small change request on PRs with an invalid header or JSON format for example.

staab (approved):

PR #1478, opened 2024-09-03
NIP-XX - Style Tags

arthurfranca:
After reading [this comment](https://github.com/nostr-protocol/nips/pull/1233#issuecomment-2326586471), I remembered this idea I had to add styles to unstyled `.content` text by using tags referencing character index ranges.

Read [here](https://github.com/arthurfranca/nips/blob/styles/xx.md)

This is a draft cause I have no time atm to make a POC parser implementation.


fiatjaf:
Although this is similar to https://docs.bsky.app/docs/advanced-guides/post-richtext I think it is not a bad idea, but I definitely think we already have way too many standards to flesh out and implement already, I hope we can postpone this for 12 months.

PR #1479, opened 2024-09-03
NIP-09: Adding relay implementation potential

kehiy:
This paragraph can be used in lightweight, personal, or economic relays. It helps them keep their storage cleaner and lighter. reactions are not heavy but in the Nostr we have to wrap them on an event object and they are also easy to send by clients so these to conditions can make the cleaning of them more logical.


staab:
Any particular reason to include this recommendation? It's sort of in the nature of a relay to be able to do whatever they want, so I think this is implied and unnecessary.

kehiy:
@staab Besides being lightweight, this helps to remove dead stuff from the network. The kind 7 events are intended to point to some other events always. when they are pointing to void, why should we keep them?


That's the only reason I have for that at the moment.

PR #1482, opened 2024-09-06
NIP-0b - On-Behalf of (Simple Sub-Key Management)

ice-orestes:
#### Prior art:

- [NIP-26](https://github.com/nostr-protocol/nips/blob/master/26.md): Multiple key delegation and no revocation (just time bound)
- [NIP-41](https://github.com/nostr-protocol/nips/blob/master/41.md): More complete single key delegation and revocation (complex)

This NIP defines a simple way for an Identity Sub-Key management system, including active, inactive and revoked public keys, for publishing On-Behalf of a master identity.

#### Why?

Each of these methods has their own pros and cons, but there's a very common use case that is not covered by any of these methods completely, which is having a master identity that is secure (possibly in cold storage), that can whitelist and blacklist multiple public keys (one for each device or app, for example), that can be used individually (less common but interesting for media management companies) and can publish on-behalf of one or several master identities.

Using [NIP-26](https://github.com/nostr-protocol/nips/blob/master/26.md) for this purpose doesn't allow for blacklisting (revocation), or alternatively requires time bound delegation, with new delegation signing at every expiry. On the other hand [NIP-41](https://github.com/nostr-protocol/nips/blob/master/41.md) is a more complete and complex identity management solution, but only allows for one active sub-key at any point in time, and that sub-key's only purpose is bound to the main identity.

### DIFF
--- a/0b.md
+++ b/0b.md
@@ -1,5 +1,5 @@
-NIP-XXX
-=======
+NIP-0B
+======
 
 On-Behalf of - Simple Sub-Key Management
 -----
--- a/51.md
+++ b/51.md
@@ -33,6 +33,7 @@ For example, _mute list_ can contain the public keys of spammers and bad actors 
 | Interests         | 10015 | topics a user may be interested in and pointers             | `"t"` (hashtags) and `"a"` (kind:30015 interest set)                              |
 | Emojis            | 10030 | user preferred emojis and pointers to emoji sets            | `"emoji"` (see [NIP-30](30.md)) and `"a"` (kind:30030 emoji set)                  |
 | DM relays         | 10050 | Where to receive [NIP-17](17.md) direct messages            | `"relay"` (see [NIP-17](17.md))                                                   |
+| On-Behalf of | 10100 | [NIP-0b](0b.md) attestations list            | `"p"` (pubkeys, with [NIP-0b](0b.md) attestation string)                                                   |
 | Good wiki authors | 10101 | [NIP-54](54.md) user recommended wiki authors               | `"p"` (pubkeys)                                                                   |
 | Good wiki relays  | 10102 | [NIP-54](54.md) relays deemed to only host useful articles  | `"relay"` (relay URLs)                                                            |
 
--- a/README.md
+++ b/README.md
@@ -31,6 +31,7 @@ They exist to document what may be implemented by [Nostr](https://github.com/nos
 - [NIP-07: `window.nostr` capability for web browsers](07.md)
 - [NIP-08: Handling Mentions](08.md) --- **unrecommended**: deprecated in favor of [NIP-27](27.md)
 - [NIP-09: Event Deletion Request](09.md)
+- [NIP-0b: On-Behalf of (Simple Sub-Key Management)](0b.md)
 - [NIP-10: Conventions for clients' use of `e` and `p` tags in text events](10.md)
 - [NIP-11: Relay Information Document](11.md)
 - [NIP-13: Proof of Work](13.md)
@@ -160,6 +161,7 @@ They exist to document what may be implemented by [Nostr](https://github.com/nos
 | `10050`       | Relay list to receive DMs       | [51](51.md), [17](17.md)               |
 | `10063`       | User server list                | [blossom]                              |
 | `10096`       | File storage server list        | [96](96.md)                            |
+| `10100`       | On-Behalf of Attestations List  | [0b](0b.md)                            |
 | `13194`       | Wallet Info                     | [47](47.md)                            |
 | `21000`       | Lightning Pub RPC               | [Lightning.Pub][lnpub]                 |
 | `22242`       | Client Authentication           | [42](42.md)                            |
@@ -247,10 +249,11 @@ They exist to document what may be implemented by [Nostr](https://github.com/nos
 | name              | value                                | other parameters                | NIP                                   |
 | ----------------- | ------------------------------------ | ------------------------------- | ------------------------------------- |
 | `e`               | event id (hex)                       | relay URL, marker, pubkey (hex) | [01](01.md), [10](10.md)              |
-| `p`               | pubkey (hex)                         | relay URL, petname              | [01](01.md), [02](02.md)              |
+| `p`               | pubkey (hex)                         | relay URL, petname / attestation             | [01](01.md), [02](02.md), [0b](0b.md)              |
 | `a`               | coordinates to an event              | relay URL                       | [01](01.md)                           |
 | `d`               | identifier                           | --                              | [01](01.md)                           |
 | `-`               | --                                   | --                              | [70](70.md)                           |
+| `b`               | pubkey (hex)                         | --                              | [0b](0b.md)                           |
 | `g`               | geohash                              | --                              | [52](52.md)                           |
 | `h`               | group id                             | --                              | [29](29.md)                           |
 | `i`               | external identity                    | proof, url hint                 | [39](39.md), [73](73.md)              |
--- a/XXX.md
+++ b/XXX.md
@@ -1,21 +1,31 @@
 NIP-XXX
 =======
 
-On Behalf of - Delegation (and revocation) of responsibility to other accounts
+On-Behalf of - Simple Sub-Key Management
 -----
 
 `draft` `optional`
 
-This NIP defines how a publisher public key (a media manager, another device, a sub account) can create events on-behalf of an author public key (the content author). Unlike NIP-26, here the attestations live in a repository controlled by the author, which is the `kind 0` replaceable profile event.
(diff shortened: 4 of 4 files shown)


0b.md
```diff
@@ -0,0 +1,273 @@
...
+The **Inactive** or **Revoked Attestations** should be a string in the following formats:
+
+```
+"inactive:<timestamp>"
+```
+
+or 
+
+```
+"revoked:<timestamp>"
```
vitorpamplona, line 104: What's the purpose of the revoked timestamp if it is going to invalidate everything? 

It's easier to just delete the key from the list.

0b.md
```diff
@@ -0,0 +1,273 @@
...
+
+#### Active Attestation String
+
+The **Active Attestation** should be a string in the following format:
+
+```
+"active:<timestamp>:<kinds comma separated list, optional>"
+```
+
+Last `:` is only required if an optional kinds list is present. Without a kinds list, the attestation is valid for all event kinds besides `kind 10100`.
```
vitorpamplona, line 91: Kinds MUST be required, because otherwise subkeys can delete posts of the main account and can also change `kind:10100` of the main account. 

There are many kinds that don't work well with delegation or that the delegation gets ambiguous. For instance, what should kinds that use encryption, like DMs, do? Encrypt to the main account or keep working as if there is no delegation?

Can subkeys have their own subkeys ad infinitum by having multiple `b`s to build the stack? And if so, do verifiers need to verify the complete chain?

0b.md
```diff
@@ -0,0 +1,273 @@
...
+
+`draft` `optional`
+
+This NIP defines a simple way for an Identity Sub-Key management system, including active, inactive and revoked public keys, for publishing On-Behalf of a master identity.
+
+### Motivation
+
+The state of art now has two methods for enabling sub-key management:
+- [NIP-26](https://github.com/nostr-protocol/nips/blob/master/26.md): Multiple key delegation and no revocation (just time bound)
+- [NIP-41](https://github.com/nostr-protocol/nips/blob/master/41.md): More complete single key delegation and revocation (complex)
```
vitorpamplona, line 15: All links to NIP-41 are broken. This doesn't exist.

0b.md
```diff
@@ -0,0 +1,273 @@
...
+```
+
+Last `:` is only required if an optional kinds list is present. Without a kinds list, the attestation is valid for all event kinds besides `kind 10100`.
+
+#### Inactive and Revoked Attestation String
+
+The **Inactive** or **Revoked Attestations** should be a string in the following formats:
+
+```
+"inactive:<timestamp>"
```
vitorpamplona, line 98: This should not be a timestamp as the inactive key can still post in the past, during the active period. It should contain a bloomfilter with all the valid event_ids made during the time the key was active. 

The main account gathers all valid events and merges their IDs in a bloom filter and posts it, likely as a separate event because this will get big very quicky (Imagine corporate accounts)

Verification will check if the event id is in the list or not.

0b.md
```diff
@@ -0,0 +1,273 @@
...
+    ...
+    "sig": "a9a4e2192eede77e6c9d24ddfab95ba3ff7c03fbd07ad011fff245abea431fb4d3787c2d04aad001cb039cb8de91d83ce30e9a94f82ac3c5a2372aa1294a96bd"
+  }
+  ```
+
+
+### Relay & Client Support
+
+Relays should answer requests such as `["REQ", "", {"authors": ["A"]}]` by querying both the `pubkey` and on-behalf `#b` tags `[1]` value.
+
```
vitorpamplona, line 272: NIP-42 MUST allow the use of the subkey to login if authorized.

vitorpamplona (commented):
Few comments. In general, this does create significant burdens for signers and verifiers as well as broken user experiences when clients and relays don't implement it. The only way to solve the UX is to make this mandatory for all of Nostr, which significantly increases the startup time of any Nostr project. 

It also does not solve how to deal with encryption and shared secrets for event kinds that need it, like DMs and private group chats. Meaning that subkeys don't see what other subkeys are doing even when authorized.

More importantly, this type of delegation changes the indexing of replaceable events. Since subkeys can change replaceables of the main key, queries by `a` tag must be broken down to also search by changes from authorized `b` keys. Relays that keep only the last event per pubkey in the 10000-20000 range must now also account for subaccounts that can dynamically change at any time, past and future (by rebroadcasting past kind 10100 events or deleting them). 

One can also create an elaborate attack by using a past kind 10100 when the stolen subkey was still authorized. Attacker gets the past version, sends it to a relay the attacker controls and doesn't allow that event to be replaced by newer versions. Now new posts that make clients think that relay is the main's account's relay will see the stolen key as still valid.

fiatjaf:
Vitor said it all.

0b.md
ice-orestes: fixed 👍

vitorpamplona:
> This should be similar to NIP-26 

Yeah. Some/many of these issues are also on NIP-26 and that's why no one implements NIP-26 and why I proposed to deprecate it entirely #1051

ice-orestes:
Addressing the other points of @vitorpamplona initial comment:

>It also does not solve how to deal with encryption and shared secrets for event kinds that need it, like DMs and private group chats. Meaning that, subkeys don't see what the main account and other subkeys are encrypting/decrypting even when authorized.

True for all ECDH Key Exchange based functionality, as they rely on participant accounts private keys to compute the shared secret. This limitation is also valid to any sub-account mechanism in nostr. But we had a few attempts like PR #59, or PR #580, to have a single shared secret effectively shared, by DM or replaceable event, encrypted to each participant. Something alike would solve this problem and allow sub-accounts to coexist peacefully with DMs and private group chats in nostr.

> More importantly, this type of delegation changes the indexing of replaceable events. Since subkeys can change replaceables of the main key, queries by a tag must be broken down to also search by changes from authorized b keys.

True. Relays should consider a replaceable with a valid `b` tag as effectively coming from the master account, that way this problem is mitigated.

>  Relays that keep only the last event per pubkey in the 10000-20000 range must now also account for subkeys and those subkeys can be dynamically changed at any time, past and future (by rebroadcasting past kind 10100 events or deleting them, reverting to a previous state).

This is why we have these requirements for kind `10100`:
- `This list is owned by the master account, and can only be updated by an event signed by the master account itself`
- `This list is ever growing, so each update has to include all previous attestations unchanged, and add one or more attestations to the list, otherwise they should be considered invalid and ignored by clients and relays.`

> A replaceable that has a b tag cannot affect the indexing of the sub key's own replaceables and should not be returned when a client requests the replaceables of the subkey.

Exactly, relays that receive a replaceable event with a valid `b` tag should apply it to the master account and ignore it on the sub-account level. Maybe this should be added to the last section of the NIP?

> One can also create an elaborate attack by using a past kind 10100 when the stolen subkey was still authorized. The attacker gets the past version, sends it to a relay the attacker controls, and doesn't allow that event to be replaced by newer versions. Now new posts by the sub key can make clients think the attacker's relay is the main account's relay, which will make users see the stolen key as still valid.

Not sure what you mean by `and doesn't allow that event to be replaced by newer versions`, as I am not aware of a way for a single relay to block other relays from accepting new versions, or `make clients think the attacker's relay is the main account's relay`, as I don't know what that means, do you mean NIP-65 relay list metadata?

Thanks again for your comments and looking forward to improve this NIP with all the input...

0b.md
ice-orestes: The list has to keep all history to keep track of revocations, exactly to avoid replay attacks of past events

0b.md
ice-orestes: > Kinds MUST be required, because otherwise subkeys can delete posts of the main account and can also change `kind:10100` of the main account.

Not against sub-keys deleting posts by master or any other sub-account. The kind `10100` is not allowed to be signed by any delegation mechanism, only the master account can sign those events.

> There are many kinds that don't work well with delegation or that the delegation gets ambiguous. For instance, what should kinds that use encryption, like DMs, do? Encrypt to the main account or keep working as if there is no delegation?

Those kinds shouldn't take `b` tag into account... see my reply below about DM's etc.

> Can subkeys have their own subkeys ad infinitum by having multiple `b`s to build the stack? And if so, do verifiers need to verify the complete chain?

Don't like this idea, I think that only one level is enough. Maybe that should be more explicit in the NIP?

0b.md
ice-orestes: The idea of the `inactive` attestation is for sub-accounts that were not compromised, just inactivated. Imagine the use case of an iPhone that you want to replace, you request to the master account to inactivate that sub-account as the key is removed from the device, effectively destroying the key. If the key was leaked by any means, than the revocation is needed, not the inactivation.

0b.md
ice-orestes: This is cool. So you mean that kind `22242` ephemeral event on the `AUTH` message should use the `b` tag to reference the master account, and the relays should authenticate the main account when such `b` tag is present?

PR #1484, opened 2024-09-07
nip-29: human-readable, url friendly group names.

kehiy:
Here is a better, and **known** way to identify groups for NIP-29 while sharing both group _id_ and host.

Also, the current model: `host'id`, is not URL friendly.


fiatjaf:
If we were to do that it would be better to use it for readable names instead of the canonical group ids, which are meant to be random.

kehiy:
@fiatjaf So, you mean something like that: `<random_id>@<host>`?

kehiy:
@fiatjaf Yes, got it. By _id_m here I mean a readable name of course. Should I add it to the spec?

kehiy:
@fiatjaf @water783 I've updated the spec. please check. I used NIP-05 this time to point to group _id_, which is not human-readable and random generated string, but the name itself is human-readable. I've added the detail on how clients can find out how to behave when they see a nip-05 address. but if you think it's not logical we can use another similar thing. this is URL-friendly, easy to share, and human-readable now.

water783:
Perhaps we can add kind 9009, and authorized admins can request a unique name(human-readable)?

kehiy:
@water783 you mean admins request human-readable names from relays? it can work as well. and it's a good idea. then we don't need to resolve them like nip-05.

kehiy:
@staab same for this?

IngwiePhoenix:
I just randomly saw this NIP and I know this PR is merged

But why exactly apostrophe? This feels... a little random, honestly. A very personal take, I am aware, but I would like to know what the idea is regardless. :) My gut would've told me to use a common URI-esque pattern like `host/name` or even `host#name`, since URL fragments are usually not sent and perfectly cut between host, path, query and the anchor.

Thanks!

fiatjaf:
It's arbitrary, we can change it. I pick the apostrophe specifically so it wouldn't conflict with web characters that already have their own meaning like `/` and `#` and could interfere with how web apps are made. `@` was also avoided since it has other meanings. Other characters were also avoided because they tend to be URL-escaped.

Unfortunately I don't think any apps are using this identifier format anyway, which is a shame. If people can agree on all using the same format, doesn't matter what format, I'm all for it.

wcat7:
> Unfortunately I don't think any apps are using this identifier format anyway

0xchat has been using the apostrophe from the very beginning.

IngwiePhoenix:
> I pick the apostrophe specifically so it wouldn't conflict with web characters that already have their own meaning

Ah, that makes sense then. :) Hence why I asked. Thank you for elaborating!

> 0xchat has been (...)

Thank you for reminding me to actually finally try it out. I keep forgetting to do so...

PR #1486, opened 2024-09-10
Remove recommendation to map emoji reactions to like/dislike

staab:
See the discussion [here](https://github.com/nostrability/nostrability/issues/88)

Emojis are a high-fidelity medium of expression subject to a significant amount of interpretation, while +/- reactions are a low-fidelity medium of expression subject to much less interpretation. The two are semantically incompatible ways of reacting to content, and should not be conflated.

Previously, the NIP included a recommendation to interpret emojis as likes/dislikes, but this put a burden on receiving clients to map every possible emoji to a binary sentiment. This is not only a huge mapping, it's also impossible because the same emoji reaction can be used in different contexts to mean different things.

Additionally, some clients created emoji reactions on their users' behalf to indicate certain application states, such as a failed zap or a content warning. These concepts should instead be represented as first-class things within the protocol.

This PR also adds a recommendation to add a `c` tag so that reaction content can be indexed. This will allow clients that only support like/dislike or a subset of emojis to fetch only relevant events.

### DIFF
--- a/25.md
+++ b/25.md
@@ -7,20 +7,14 @@ Reactions
 
 `draft` `optional`
 
-A reaction is a `kind 7` event that is used to react to other events.
+A reaction is a `kind 7` event that is used to react to other events. A reaction's `content` field
+MUST include a single character indicating the value of the reaction.
 
-The generic reaction, represented by the `content` set to a `+` string, SHOULD
-be interpreted as a "like" or "upvote".
+A reaction with `content` set to `+` or an empty string MUST be interpreted as a "like" or "upvote".
+A reaction with `content` set to `-` MUST be interpreted as a "dislike" or "downvote".
 
-A reaction with `content` set to `-` SHOULD be interpreted as a "dislike" or
-"downvote". It SHOULD NOT be counted as a "like", and MAY be displayed as a
-downvote or dislike on a post. A client MAY also choose to tally likes against
-dislikes in a reddit-like system of upvotes and downvotes, or display them as
-separate tallies.
-
-The `content` MAY be an emoji, or [NIP-30](30.md) custom emoji in this case it MAY be interpreted as a "like" or "dislike",
-or the client MAY display this emoji reaction on the post. If the `content` is an empty string then the client should
-consider it a "+".
+A reaction with `content` set to an emoji or [NIP-30](30.md) custom emoji SHOULD NOT be interpreted
+as a "like" or "dislike". Clients MAY instead display this emoji reaction on the post.
 
 Tags
 ----
--- a/25.md
+++ b/25.md
@@ -7,8 +7,10 @@ Reactions
 
 `draft` `optional`
 
-A reaction is a `kind 7` event that is used to react to other events. A reaction's `content` field
-MUST include a single character indicating the value of the reaction.
+A reaction is a `kind 7` event that is used to indicate user reactions to other events. A
+reaction's `content` field MUST include a single character indicating the value of the reaction.
+Reactions MUST NOT be used to support application-level semantics, they are for user-generated
+content only.
 
 A reaction with `content` set to `+` or an empty string MUST be interpreted as a "like" or "upvote".
 A reaction with `content` set to `-` MUST be interpreted as a "dislike" or "downvote".
--- a/25.md
+++ b/25.md
@@ -8,9 +8,8 @@ Reactions
 `draft` `optional`
 
 A reaction is a `kind 7` event that is used to indicate user reactions to other events. A
-reaction's `content` field MUST include a single character indicating the value of the reaction.
-Reactions MUST NOT be used to support application-level semantics, they are for user-generated
-content only.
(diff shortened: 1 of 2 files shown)


25.md
```diff
@@ -7,20 +7,16 @@ Reactions
 
 `draft` `optional`
 
-A reaction is a `kind 7` event that is used to react to other events.
+A reaction is a `kind 7` event that is used to indicate user reactions to other events. A
+reaction's `content` field MUST include a single character indicating the value of the reaction.
```
vitorpamplona, line 11: Custom reactions `:name:` are not single chars. Also, many emojis are not single chars. It's common to see emoji blends, which depending on the OS's support, renders as one or more chars.

25.md
```diff
@@ -7,20 +7,16 @@ Reactions
...
-"downvote". It SHOULD NOT be counted as a "like", and MAY be displayed as a
-downvote or dislike on a post. A client MAY also choose to tally likes against
-dislikes in a reddit-like system of upvotes and downvotes, or display them as
-separate tallies.
-
-The `content` MAY be an emoji, or [NIP-30](30.md) custom emoji in this case it MAY be interpreted as a "like" or "dislike",
-or the client MAY display this emoji reaction on the post. If the `content` is an empty string then the client should
-consider it a "+".
+A reaction with `content` set to an emoji or [NIP-30](30.md) custom emoji SHOULD NOT be interpreted
+as a "like" or "dislike". Clients MAY instead display this emoji reaction on the post.
```
vitorpamplona, line 19: Clients SHOULD (not MAY) display the emoji if present.

25.md
```diff
@@ -7,20 +7,16 @@ Reactions
 
 `draft` `optional`
 
-A reaction is a `kind 7` event that is used to react to other events.
+A reaction is a `kind 7` event that is used to indicate user reactions to other events. A
+reaction's `content` field MUST include a single character indicating the value of the reaction.
+Reactions MUST NOT be used to support application-level semantics, they are for user-generated
+content only.
```
vitorpamplona, line 13: Aren't upvotes and downvotes and tallying behind them application-level semantics? Clients should be able to rank by reactions in custom UIs if they want to. There is no need to make this limitation.

vitorpamplona (commented):
The content warning emoji is not an application state and will stay.

25.md
staab: Got it, will update that

25.md
staab: Just keeping the original content of the NIP. I think MAY is actually correct, because you may filter by emoji and display related posts, in which case display would be redundant.

25.md
staab: Sure, that's not what I meant, but I've removed this line anyway.

mikedilger:
I have used negative-sentiment emojis in jest multiple times. I would hate for them to be interpreted as downvotes.

+1

25.md
```diff
@@ -33,7 +28,8 @@ The last `p` tag MUST be the `pubkey` of the event being reacted to.
 
 The `a` tag MUST contain the coordinates (`kind:pubkey:d-tag`) of the replaceable being reacted to.
 
-The reaction event MAY include a `k` tag with the stringified kind number of the reacted event as its value.
+The reaction event SHOULD include a `k` tag with the stringified kind number of the reacted event as
+its value, and a `c` tag with the reaction's `content` so that reactions can be indexed and filtered.
```
dluvian, line 32: I think having `c` as a sentiment category with values  [+, -, 0]  for positive, negative and neutral is more useful than simply copying the `content`.

staab: Where does this mapping come from? If from the client, it doesn't solve anything about assuming user intention. If from the user, then it adds UX friction. If your users want to explain their emojis, that's fine, I just don't think that would be widely adopted.

staab:
I went ahead and removed the `c` tag for now. I still think this is worth merging.

25.md
```diff
@@ -7,20 +7,15 @@ Reactions
...
-A reaction with `content` set to `-` SHOULD be interpreted as a "dislike" or
-"downvote". It SHOULD NOT be counted as a "like", and MAY be displayed as a
-downvote or dislike on a post. A client MAY also choose to tally likes against
-dislikes in a reddit-like system of upvotes and downvotes, or display them as
-separate tallies.
-
-The `content` MAY be an emoji, or [NIP-30](30.md) custom emoji in this case it MAY be interpreted as a "like" or "dislike",
-or the client MAY display this emoji reaction on the post. If the `content` is an empty string then the client should
-consider it a "+".
+A reaction with `content` set to an emoji or [NIP-30](30.md) custom emoji SHOULD NOT be interpreted
```
alexgleason, line 17: Now there is nothing that explicitly says content can be an emoji. It's only implied

alexgleason (approved):

25.md
staab: Added a small clarification.

PR #1488, opened 2024-09-11
make *.md linguist detectable.

kehiy:


alexgleason:
What does that do?

kehiy:
@alexgleason currently all of this repo is `*.md` files. github can't determine and show it by default; there is no top-used language, and it looks like an empty repository from outside. adding this file will show md as the language used here.

for example BOLT [repo](https://github.com/lightning/bolts).

in their pr to add this file they have some more reasons for that as well which we can consider(?).

alexgleason (approved):

PR #1490, opened 2024-09-11
nip-01: clarify response to close message.

kehiy:
we have to clarify how a relay can acknowledge a close request from a client. here is a [draft pr on immortal](https://github.com/dezh-tech/immortal/pull/23/files#diff-65376c52e7093ac631724e7edc57361fc3d6be28d42df033d842d43172d34deeR168) which implements that. more implementation may do that which im not aware of.

### DIFF
--- a/01.md
+++ b/01.md
@@ -149,15 +149,16 @@ The `limit` property of a filter is only valid for the initial query and MUST be
 
 Relays can send 5 types of messages, which must also be JSON arrays, according to the following patterns:
 
-  * `["EVENT", <subscription_id>, <event JSON as defined above>]`, used to send events requested by clients.
-  * `["OK", <event_id>, <true|false>, <message>]`, used to indicate acceptance or denial of an `EVENT` message.
-  * `["EOSE", <subscription_id>]`, used to indicate the _end of stored events_ and the beginning of events newly received in real-time.
-  * `["CLOSED", <subscription_id>, <message>]`, used to indicate that a subscription was ended on the server side.
-  * `["NOTICE", <message>]`, used to send human-readable error messages or other things to clients.
+* `["EVENT", <subscription_id>, <event JSON as defined above>]`, used to send events requested by clients.
+* `["OK", <event_id>, <true|false>, <message>]`, used to indicate acceptance or denial of an `EVENT` message.
+* `["EOSE", <subscription_id>]`, used to indicate the _end of stored events_ and the beginning of events newly received in real-time.
+* `["CLOSED", <subscription_id>, <message>]`, used to indicate that a subscription was ended on the server side.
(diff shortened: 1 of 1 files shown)


kehiy:
@staab in most of protocols (state-full) you can see we need to acknowledge about changing the state that nodes have about each other.

01.md
```diff
@@ -168,8 +169,11 @@ This NIP defines no rules for how `NOTICE` messages should be sent or treated.
...
+
+- `CLOSED` messages MUST be sent in response to a `REQ` when the relay refuses to fulfill it, or in response to `CLOSE` to acknowledge the end of a subscription based on client request. It can also be sent when a relay decides to kill a subscription on its side before a client has disconnected or sent a `CLOSE`. This message uses the same pattern of `OK` messages with the machine-readable prefix and human-readable message. Some examples:
   * `["CLOSED", "sub1", "unsupported: filter contains unknown elements"]`
   * `["CLOSED", "sub1", "error: could not connect to the database"]`
   * `["CLOSED", "sub1", "error: shutting down idle subscription"]`
-- The standardized machine-readable prefixes for `OK` and `CLOSED` are: `duplicate`, `pow`, `blocked`, `rate-limited`, `invalid`, and `error` for when none of that fits.
+  * `["CLOSED", "sub1", "ok: sub1 closed successfully"]`
```
AsaiToshiya, line 177: I feel like `["CLOSED", "sub1"]` or `["CLOSED", "sub1", ""]` is enough.

kehiy: @AsaiToshiya yes, agree. based on my changes, your examples will be valid as well. let me use your examples.

kehiy: example updated, `ok` got removed.

staab:
That's not really an argument. Is there a use case this supports, or is it just busy work?

fiatjaf:
This just adds unnecessary bloat.

kehiy:
how a client can find out if the closing process was successful? relay changes an state related to client which affects their communication, but client can be aware of that how was it? successful or failed?

staab:
> how a client can find out if the closing process was successful?

Why would a client care?

kehiy:
@staab to make sure a relay won't misbehave, otherwise close the connection with them in the WebSocket layer.

AsaiToshiya:
I agree with @staab. I thought some relays send `["CLOSED", "sub1"]`, but it was just my imagination.

staab:
@kehiy doing so would break all current implementations. Also, CLOSED is no guarantee that the relay isn't still misbehaving. If clients want to do that they have to analyze the behavior of the relay, not just blindly trust what it says.

kehiy:
@staab @AsaiToshiya ok, makes sense.

PR #1492, opened 2024-09-12
custom emojis in reactions

ZigBalthazar:
add reaction(kind 7) example.

### DIFF
--- a/30.md
+++ b/30.md
@@ -54,3 +54,19 @@ In kind 1 events, the `content` should be emojified.
   "created_at": 1682630000
 }
 ```
+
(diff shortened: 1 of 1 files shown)


AsaiToshiya (approved):

fiatjaf:
This is just documenting a standard practice, right?

AsaiToshiya:
Yes, I think so. It doesn't necessarily have to be documented.

alexgleason:
It's documented already in NIP-25, but it doesn't hurt to include it in NIP-30 too.

PR #1493, opened 2024-09-12
nip-29: restriction for _id_

kehiy:
I decided to update _id_ details before working on `names` #1484. here are some restrictions added and more details.


ismyhc:
This is okay, but I don't really like "." as a character for an id.

I would stick with a-z,0-9,-,_

kehiy:
@ismyhc I've updated the details. a-z (case-insensitive), 0-9, - and _. which can support UUID as well.

ismyhc:
> @ismyhc I've updated the details. a-z (case-insensitive), 0-9, - and _. which can support UUID as well.

This works for me.

kehiy:
should we close this?
@fiatjaf @ismyhc

staab:
Yes, I think so, since it conflicts with being able to upgrade unmanaged groups now.

kehiy:
@staab thanks a lot. done.

